PR #31, opened 2020-08-31
fix: add updated lockfile

hoorayimhelping:
somewhere along the way, we forgot to add the lockfile after running `yarn add`


drdelambre:
gimme one sec to double check why this fell out of sync

TCL735:
My favorite kind of PR.

PR #33, opened 2020-09-01
fix: allow Raw Data View to parse objects in CSV

TCL735:
Closes idpe #7553

Currently, when an object is returned in a field of the CSV, Raw Data View is unable to parse the CSV correctly especially when the object has commas (multiple properties, nested objects, etc).

The solution is the parse each of the objects correctly, and then re-insert them back into the correct position in the parsed CSV.


src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -30,3 +37,96 @@ export const parseFiles = (responses: string[]): ParseFilesResult => {
 
   return {data, maxColumnCount}
 }
+
+export const parseFilesWithObjects = (
```
drdelambre, line 41: uuuuugh tech debt

TCL735: Well, it is easy to pay.... should i mark it with a comment, something like, "Rip this out when RawDataView comes directly from Giraffe"?

src/timeMachine/components/RawFluxDataTable.tsx
```diff
@@ -23,11 +27,15 @@ class RawFluxDataTable extends PureComponent<Props, State> {
...
 
   private parseFiles = memoizeOne(parseFiles)
+  private parseFilesWithObjects = memoizeOne(parseFilesWithObjects)
 
   public render() {
     const {width, height, files, disableVerticalScrolling} = this.props
+
     const {scrollTop, scrollLeft} = this.state
-    const {data, maxColumnCount} = this.parseFiles(files)
+    const {data, maxColumnCount} = isFlagEnabled('parseObjectsInCSV')
```
drdelambre, line 36: dont forget to update config cat.. THEN update idpe to get this flag into production

src/timeMachine/utils/rawFluxDataTable.ts
TCL735: actually, probably not necessary, since the whole util file will be removed when the change to Giraffe's RawFluxDataTable happens.

PR #35, opened 2020-09-01
feat(aggregate-error): added actionable notification for invalid auto aggregate type selection

Asalem1:
Closes https://github.com/influxdata/influxdb/issues/19344

The purpose of this PR is to address the issue of inactionable error messages when a user's auto aggregation type is incompatible with the data that's being queried.

This PR aims to:

- Add a global context (reducer) to only display the error message when the auto aggregation is selected
- Add a new notification to notify the user of what the issue is with their query when the aggregate type is incompatible
- Update the Notification component to allow an actionable event that will update the UI for the user

aggregate-error


hoorayimhelping:
@Asalem1 do we have any designs for this? was the design team consulted before we decided to add this new interaction?

Asalem1:
@alexpaxton would love your thoughts on this

src/utils/aggregateTypeErrors.ts
```diff
@@ -0,0 +1,15 @@
...
+  errorMessage: string
+): boolean =>
+  errorCode === 'invalid' && errorMessage.includes(unsupportedAggregateType)
+
+export const aggregateTypeError = () => {
+  return {
+    aggregateType: true,
+    message:
+      "It looks like you're trying to apply an incompatible aggregate function. You can fix this by selecting the Last Aggregate Function",
+  }
```
alexpaxton, line 14: Could we explain "incompatible" a bit more? Something like `It looks like you're trying to apply a number-based aggregate function to a string, which cannot be processed. You can fix this by selecting the Aggregate Function "Last"`

alexpaxton (approved):
Good stuff

src/utils/aggregateTypeErrors.ts
Asalem1: Sounds great!

PR #40, opened 2020-09-02
fix(client-lib): fix client library redirect bug

Asalem1:
Updated the redirect so that clicking on the load data client libraries redirects to the new load data page rather than the old client libraries

client-lib-fix


russorat (approved):
LGTM

PR #44, opened 2020-09-08
feat: move cypress over to using env variables

drdelambre:
use this for overrides: https://docs.cypress.io/guides/guides/environment-variables.html#Option-2-cypress-env-json


hoorayimhelping (approved):
nice, this is a huge first step

PR #45, opened 2020-09-08
feat: Readme call from redux

zoesteinkamp:
Closes some of the issues in #19470

We now store the readme in redux, instead of calling it every time. This will prevent lots of calls for the data and will allow us to  display faster. 
 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/src/templates/actions/thunks.ts
+++ b/src/templates/actions/thunks.ts
@@ -3,7 +3,7 @@ import {normalize} from 'normalizr'
 
 // APIs
 import {client} from 'src/utils/api'
-import {fetchStacks} from 'src/templates/api'
+import {fetchStacks, fetchReadMe} from 'src/templates/api'
 import {createDashboardFromTemplate} from 'src/dashboards/actions/thunks'
 import {createVariableFromTemplate} from 'src/variables/actions/thunks'
 import {createTaskFromTemplate} from 'src/tasks/actions/thunks'
@@ -21,6 +21,7 @@ import {
   setExportTemplate,
   setTemplatesStatus,
   setTemplateSummary,
+  setTemplateReadMe,
   Action as TemplateAction,
 } from 'src/templates/actions/creators'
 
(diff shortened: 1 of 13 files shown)


src/templates/actions/thunks.ts
```diff
@@ -289,3 +291,22 @@ export const fetchAndSetStacks = (orgID: string) => async (
...
+
+export const fetchAndSetReadme = (name: string, directory: string) => async (
+  dispatch: Dispatch<Action>
+): Promise<void> => {
+  try {
+    const response = await fetchReadMe(directory)
+    dispatch(setTemplateReadMe(name, response))
+  } catch (error) {
+    reportError(error, {
+      name: 'The community template fetch github readme failed',
```
hoorayimhelping, line 303:
```suggestion
      name: `The community template github readme fetch failed for ${name}`,
```
Just a little bit of extra context to help us if there's a problem

src/templates/components/CommunityTemplateReadme.tsx
```diff
@@ -1,37 +1,39 @@
...
-      reportError(error, {
-        name: 'The community template fetch github readme failed',
-      })
-
-      this.setState({
-        readMeData: "## We can't find the readme associated with this template",
-      })
+class CommunityTemplateReadmeUnconnected extends Component<Props> {
+  componentDidMount = () => {
+    if (!this.props.communityTemplateReadmeCollection) {
```
hoorayimhelping, line 22: ahh, i told you to change it everywhere, but this is a place where it should be called something else. see below.

src/templates/components/CommunityTemplateReadme.tsx
```diff
@@ -41,3 +43,25 @@ export class CommunityTemplateReadme extends PureComponent<Props> {
     )
   }
 }
+
+const mstp = (state: AppState, props: any) => {
+  const templateDetails = getTemplateNameFromUrl(props.url)
+  return {
+    directory: templateDetails.directory,
+    name: templateDetails.name,
+    communityTemplateReadmeCollection:
```
hoorayimhelping, line 52: i told you to change it everywhere, but this is a place where it should be slightly different
```suggestion
    readme:
```
you'll have to change it elsewhere in the file, but i figured i'd keep from suggestion all of those

hoorayimhelping (changes requested):
this is great work! just a couple small suggestions

hoorayimhelping (approved):
🤙 🤙 🤙 🤙 🤙 🤙 🤙

PR #46, opened 2020-09-08
feat: add x axis domain manipulation to scatter plot

drdelambre:
https://github.com/influxdata/influxdb/issues/18981

was a bit unsure about the acceptance criteria for this one, but it seems like we wanted to add the ability to manually set the domain on the x axis for histograms and scatter plots. it exists on histograms, so i just added it onto the scatter plots. also went through and fixed whatever was making the giant auto domain range as well. Problem seemed rooted in mimicking the XYPlot interface to giraffe and assuming that time was significant on the scatter plot, when it's not (more of an explicit declaration of the requirement, whereas XY is more implicit)


zoesteinkamp (approved):
Looks good to me

PR #51, opened 2020-09-09
fix(dashboard-var-hydration): limiting variable hydration of the dashboard to those within the view

### LINKED ISSUE #53: Limit Dashboard Variable Hydration
### Problem

Currently navigating to a dashboard will load and hydrate all the query variables available. This is a slow, tedious, and expensive process that is inefficient and unnecessary. 

### Steps to Reproduce

1. Create a few query variables
2. Create an empty dashboard
3. Check the network tab to see all the variables being queried

Asalem1:
Closes #53 

### Problem

Variable hydration is currently only context aware when the view has been set. If no view has been set, then all of the variables will be hydrated. 

### Solution

This issue stems from the fact that the way we are limiting variable hydration is dependent upon the view. Since this current implementation depends on the concept of a view to limit variable hydration, simply changing the order with which things are dispatched (i.e. first the view, then the variables) allows the dashboard to limit variable hydration based on its current implementation. 

Below is an example of three different dashboards:

dashboard-var-hydration

The first example shows 2 queries being issued (1 for the cell, 1 for the buckets query variable)
The second example shows 3 queries being issued (1 for each cell, none for the query variables since none are used)
The final example shows 5 queries being issues (1 for each cell, 1 for each query variable being used / nested)

The last example also shows the impact of variable caching, since cell 1 and cell 2 each depend upon overlapping variables

### DIFF
--- a/src/dashboards/actions/thunks.ts
+++ b/src/dashboards/actions/thunks.ts
@@ -366,9 +366,6 @@ export const getDashboard = (
       throw new Error(resp.data.message)
     }
 
-    const skipCache = true
-    dispatch(hydrateVariables(skipCache, controller))
-
     const normDash = normalize<Dashboard, DashboardEntities, string>(
       resp.data,
       dashboardSchema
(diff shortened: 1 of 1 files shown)


drdelambre (approved):
i dont quite follow the reasoning, but i also dont see anything obviously wrong with the change warranting further investigation. make sure to screenshot this message for when tools falls over.

PR #54, opened 2020-09-09
chore: rename reportError => reportErrorThroughHoneyBadger

hoorayimhelping:
### What

Renames `reportError` to `reportErrorThroughHoneyBadger` since that's what it's doing.

### Why

Work ahead of adding a `reportError` function which will only write an error point, not alert HoneyBadger.

### Future Usage
`reportErrorThroughHoneyBadger`: there has been an error that affects the user's experience, and they're aware of it. We should likely look into these kinds of errors immediately.

`reportError`: an error has been triggered, most likely by user interaction, but it's not worth addressing at the moment. We want to measure it and track it and look at it as part of a larger trend, not something to respond to in the moment.

### DIFF
--- a/src/checks/actions/thunks.ts
+++ b/src/checks/actions/thunks.ts
@@ -14,7 +14,7 @@ import {checkSchema, arrayOfChecks} from 'src/schemas/checks'
 
 // Utils
 import {incrementCloneName} from 'src/utils/naming'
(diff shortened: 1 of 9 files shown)


Asalem1 (approved):
Forward thinking. Nothing wrong with this in my eyes

PR #58, opened 2020-09-10
fix: cleanup some metric selector stuff

drdelambre:
i was a good engineer and kept all of these changes real clean and scoped behind a commit in the git log. prolly first time in my career. real proud of myself.

JUST SOME POLISH TO THE LIFECYCLE METHODS
and my inner russ yelling "how is it so hard to have everything alphabetically ordered"

### DIFF
--- a/src/notebooks/context/notebook.current.tsx
+++ b/src/notebooks/context/notebook.current.tsx
@@ -42,6 +42,8 @@ const getHumanReadableName = (type: string): string => {
   switch (type) {
     case 'data':
       return `Bucket ${GENERATOR_INDEX}`
+    case 'queryBuilder':
+      return `Metric Selector ${GENERATOR_INDEX}`
     case 'visualization':
       return `Visualization ${GENERATOR_INDEX}`
     case 'markdown':
--- a/src/notebooks/pipes/Data/BucketSelector.tsx
+++ b/src/notebooks/pipes/Data/BucketSelector.tsx
@@ -30,7 +30,8 @@ const BucketSelector: FC = () => {
       event('Updating Bucket Selection in Flow Query Builder', {
         bucket: updatedBucket.name,
       })
-      update({bucketName: updatedBucket.name})
+
+      update({bucket: updatedBucket})
     },
     [update]
   )
@@ -49,7 +50,7 @@ const BucketSelector: FC = () => {
             key={bucket.name}
             value={bucket}
             onClick={updateBucket}
-            selected={bucket.name === data.bucketName}
+            selected={bucket.name === data.bucket?.name}
             title={bucket.name}
             wrapText={true}
           >
@@ -60,10 +61,10 @@ const BucketSelector: FC = () => {
     )
   }
 
-  if (loading === RemoteDataState.Done && !data.bucketName) {
+  if (loading === RemoteDataState.Done && !data.bucket) {
     buttonText = 'Choose a bucket'
-  } else if (loading === RemoteDataState.Done && data.bucketName) {
-    buttonText = data.bucketName
+  } else if (loading === RemoteDataState.Done && data.bucket) {
+    buttonText = data.bucket.name
   }
 
   const button = (active, onClick) => (
--- a/src/notebooks/pipes/Data/BucketSelector.tsx
+++ b/src/notebooks/pipes/Data/BucketSelector.tsx
@@ -13,7 +13,6 @@ import {
 // Contexts
 import {BucketContext} from 'src/notebooks/context/buckets'
 import {PipeContext} from 'src/notebooks/context/pipe'
-import {SchemaContext} from 'src/notebooks/context/schemaProvider'
 
 // Utils
 import {event} from 'src/cloud/utils/reporting'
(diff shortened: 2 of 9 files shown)


drdelambre:
note.. dont try to click on the demo data bucket. it will explode. i'm 95% sure that this is expected behavior as we are bypassing the sample data service mock when we run a query.

src/notebooks/context/notebook.current.tsx
```diff
@@ -42,6 +42,8 @@ const getHumanReadableName = (type: string): string => {
   switch (type) {
     case 'data':
       return `Bucket ${GENERATOR_INDEX}`
+    case 'queryBuilder':
+      return `Metric Selector ${GENERATOR_INDEX}`
```
Asalem1, line 46: oh dang. Good catch

src/notebooks/pipes/Data/BucketSelector.tsx
```diff
@@ -64,10 +61,10 @@ const BucketSelector: FC = () => {
     )
   }
 
-  if (loading === RemoteDataState.Done && !selectedBucketName) {
+  if (loading === RemoteDataState.Done && !data.bucket) {
     buttonText = 'Choose a bucket'
-  } else if (loading === RemoteDataState.Done && selectedBucketName) {
-    buttonText = selectedBucketName
+  } else if (loading === RemoteDataState.Done && data.bucket) {
+    buttonText = data.bucket.name
```
Asalem1, line 67:
```suggestion
    buttonText = data.bucket?.name
```

src/notebooks/pipes/Data/FieldsList.tsx
```diff
@@ -11,31 +11,29 @@ import Selectors from 'src/notebooks/pipes/Data/Selectors'
...
+    return (
       <div className="data-source--list__empty">
         <p>Could not fetch schema</p>
       </div>
     )
   }
 
-  if (loading === RemoteDataState.Done && Object.keys(data).length > 0) {
-    body = <Selectors />
+  if (loading === RemoteDataState.Done) {
```
Asalem1, line 32: not sure how I feel about removing the data condition here.

src/shared/utils/flowSchemaNormalizer.ts
```diff
@@ -133,10 +135,30 @@ export const normalizeSchema = (
     data?.field
   )
 
-  const filteredTags = filterTags(dedupedTags, lowerCasedSearchTerm)
+  const dedupedMeasurements = dedupeArray(measurements)
+
```
Asalem1, line 139: I'm so confused, how is it even possible to have duplicate measurements if the parser is assigning the measurement as a key in an object

src/shared/utils/flowSchemaNormalizer.ts
```diff
@@ -133,10 +135,30 @@ export const normalizeSchema = (
...
+  dedupedMeasurements.sort((a, b) => a.localeCompare(b))
+
+  const filteredTags = filterTags(dedupedTags, lowerCasedSearchTerm).map(
```
Asalem1, line 142: Yup, not sure why I thought this would be handled by the previous filter but thanks

Asalem1 (approved):
Thanks for cleaning up after me!

src/notebooks/pipes/Data/FieldsList.tsx
Asalem1: The reason being, what happens in the event that there isn't any data loaded?

drdelambre: yeah, it's a bit of a regression not gonna lie. but the same functionality of that guard should be carried out within the `Selectors` and i didn't know exactly how to map the intent of the data field with the measurements/fields/tags concept. i'd be happy to add the guard back in if you wouldn't mind explaining that part

src/shared/utils/flowSchemaNormalizer.ts
drdelambre: lol it's your code man! i just moved it up a line. lemme know what you want me to do with it

Asalem1: Let’s keep it for safety and the assumption that past Ariel wasn’t as out of it as current Ariel

src/notebooks/pipes/Data/FieldsList.tsx
Asalem1: makes sense

src/notebooks/pipes/Data/BucketSelector.tsx
drdelambre: handled by line 66 in the conditional right above it

src/shared/utils/flowSchemaNormalizer.ts
russorat: this should be done in the flux code IMO.

src/shared/utils/flowSchemaNormalizer.ts
```diff
@@ -133,10 +135,30 @@ export const normalizeSchema = (
...
+      const key = Object.keys(tag)[0]
+
+      return {
+        [key]: tag[key].sort((a, b) => a.localeCompare(b)),
+      }
+    }
+  )
+
+  filteredFields.sort((a, b) => a.localeCompare(b))
+  filteredTags.sort((a, b) => {
```
russorat, line 153: 🥇

src/shared/utils/flowSchemaNormalizer.ts
Asalem1: Unfortunately I don't think the query builder will work properly if we return unique values with the flux results, since the schema is currently mapping out associated results in order to only allow users to build a valid query. Put simply, if we were to return unique values from the flux query, we might run the risk of losing the context of the existing associations between a measurement, field, and tag

PR #60, opened 2020-09-14
chore: groundwork for adding visualization options

drdelambre:
This PR is getting too crazy. I guess I should make a clean break at the cleanup stage before going off into the feature work. So, what this PR does:

1. moves all the types that were defined in `src/notebooks/index.ts` and moves them to the `src/types/notebooks.ts` file that @Asalem1 dropped in a few PRs ago. This is done mostly to follow convention within our codebase and depart from the "notebooks is an island" philosophy we were operating off of during prototyping.

1. inverts the pipe registration mechanism. this helps keep pipes more issolated from notebooks in that they no longer need awareness of notebooks to register themselves. theorettically, you can now mimic the pipe import mechanism in other parts of the app, and it should still work.

1. decouple the view types and better encapsulate them. right now there are, 4 or 5 large switch statements with correspondingly large state objects that all need to be kept in sync with giraffe changes. determining which types were enabled based on feature flags or implementation status was fragmented between different view implementations, and their state definitions were all tangled together. Now, they're following the pipe registration mechanism to keep them isolated into their own folders, with a strict interface for how they interact within the system. The idea is to migrate this mechanism into timemachine, and make the individual definitions lighter as we move more responsibility into giraffe.


Asalem1 (approved):
Our little notebook is growing up so fast

PR #61, opened 2020-09-14
chore: template parameterization bucket follow up

hoorayimhelping:
- Adds a width, style property to the bucket dropdown
- When installing a template or removing a template, re-fetch buckets.
  - This might cause performance issues on installs with lots of buckets

### DIFF
--- a/src/dataLoaders/components/BucketsDropdown.tsx
+++ b/src/dataLoaders/components/BucketsDropdown.tsx
@@ -14,6 +14,7 @@ interface Props {
   selectedBucketID: string
   buckets: Bucket[]
   onSelectBucket: (bucket: Bucket) => void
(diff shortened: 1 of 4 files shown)


Asalem1 (approved):
Looks good to me. Would've loved to have seen some pics of the before/after styles

src/dataLoaders/components/BucketsDropdown.tsx
```diff
@@ -14,6 +14,7 @@ interface Props {
   selectedBucketID: string
   buckets: Bucket[]
   onSelectBucket: (bucket: Bucket) => void
+  style?: {[key: string]: string}
```
alexpaxton, line 17: There's a built in type for this called `CSSProperties`

PR #64, opened 2020-09-16
test: Sn tests missing in Cypress added

Sciator:
Added cyprest test for:

- deep navigation home/left panel expanded
- homepage tests
- data explorer
   - zoom
   - refresh manual/automatic
   - saving as cell/task/variable
   - tooltip on graph hover
- dashboard edit note (Note cell)
- dashboard edit description
- ~~dasboards filter by name~~ (already exists)
- importing dasboards 
   - as file
   - paste as json

> - [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
> - [x] Rebased/mergeable
> - [ ] Tests pass
> - [x] Signed [CLA](https://influxdata.com/community/cla/)


TheRWe:
Moved from [PR influxdb#19541](https://github.com/influxdata/influxdb/pull/19541) and [PR influxdb#19552](https://github.com/influxdata/influxdb/pull/19552)

cypress/e2e/buckets.test.ts
```diff
@@ -394,5 +431,97 @@ describe('Buckets', () => {
       // mymeasurement comes from fixtures/data.txt
       cy.getByTestID('selector-list mymeasurement').should('exist')
     })
+
+    it('create scraper', () => {
```
kristinarobinson, line 435: Scrapers are not supported for cloud.  We need to have this changed to support OSS only.   Check with one of the developers to find out how to turn this turn off for cloud v2 applications.

kristinarobinson: Here's how to determine whether the app is running in cloud mode or not:
Check the following env variable `CLOUD_URL`,  if it's set to `\auth` then DO NOT run the scrapers tests (you are in the cloud environment).  Please make these changes to all scrapers tests.

desa: @drdelambre is it possible to detect if `CLOUD_URL` has that value from the cypress tests? If not, what do we do instead?

drdelambre: not from the cypress tests. we'd have to run an oss version of the cluster, then run the tests against it. there's been informal talk around having "cloud only", "oss only", and "shared" subsets of the test that we'd run in their respective environments, but nothing has been formalized / spec'd to coincide with our GA strategy. once all that's done, then we can introduce that concept into the test runner.

121watts:
what is the status of this PR?

121watts:
@Sciator could you rebase this with the latest changes and I'll personally make sure this gets merged in.  Thank you.

Cubone21:
@121watts I have rebased the branch and fixed the failing tests

121watts:
@Cubone21 great.  ive got a PR up https://github.com/influxdata/ui/pull/308

PR #65, opened 2020-09-16
refactor: add some polish to flows

alexpaxton:
- Make flows index view centered
- Increase size of aggregate dropdown in metric selector to avoid text-wrapping
- Integrate selector with focus interactions on search bar


TCL735:
🔥

TCL735 (approved):

PR #66, opened 2020-09-17
fix: save 'mainColumn' properly and set 'upperColumn' and 'lowerColumn' explicitly

TCL735:
Part of idpe #8337

- save 'mainColumn' properly so that Dashboards can render something in the cell for Band Charts.
- require the user to explicitly set or remove 'upperColumn' and 'lowerColumn'
- no automatic removal of upper and lower columns from Band Options even when they are not included in the response
- Query Builder mode: user selects from dropdowns
- Script Editor mode: user types into text input field

### DIFF
--- a/src/shared/components/BandPlot.tsx
+++ b/src/shared/components/BandPlot.tsx
@@ -32,7 +32,6 @@ import {
   BAND_SHADE_OPACITY,
   LEGEND_OPACITY_DEFAULT,
   QUERY_BUILDER_MODE,
-  SCRIPT_EDITOR_MODE,
   VIS_THEME,
   VIS_THEME_LIGHT,
 } from 'src/shared/constants'
@@ -101,12 +100,8 @@ const BandPlot: FC<Props> = ({
   theme,
 }) => {
   const mainColumnName = useMemo(() => {
-    const editMode = get(
-      queries,
-      `${activeQueryIndex}.editMode`,
-      QUERY_BUILDER_MODE
-    )
-    if (editMode === SCRIPT_EDITOR_MODE) {
+    const editMode = get(queries, `${activeQueryIndex}.editMode`, 'unknown')
+    if (editMode !== QUERY_BUILDER_MODE) {
       return mainColumn
     }
 
--- a/src/shared/components/BandPlot.tsx
+++ b/src/shared/components/BandPlot.tsx
@@ -31,6 +31,8 @@ import {
   BAND_LINE_WIDTH,
   BAND_SHADE_OPACITY,
   LEGEND_OPACITY_DEFAULT,
+  QUERY_BUILDER_MODE,
+  SCRIPT_EDITOR_MODE,
   VIS_THEME,
   VIS_THEME_LIGHT,
 } from 'src/shared/constants'
@@ -99,6 +101,15 @@ const BandPlot: FC<Props> = ({
   theme,
 }) => {
   const mainColumnName = useMemo(() => {
+    const editMode = get(
+      queries,
+      `${activeQueryIndex}.editMode`,
+      QUERY_BUILDER_MODE
+    )
+    if (editMode === SCRIPT_EDITOR_MODE) {
+      return mainColumn
+    }
+
     const aggregateFunctions = get(
       queries,
       `${activeQueryIndex}.builderConfig.functions`,
--- a/src/shared/components/BandPlot.tsx
+++ b/src/shared/components/BandPlot.tsx
@@ -2,6 +2,7 @@
 import React, {FC, useMemo} from 'react'
 import {connect, ConnectedProps} from 'react-redux'
 import {Config, Table} from '@influxdata/giraffe'
+import {get} from 'lodash'
 
 // Components
 import EmptyGraphMessage from 'src/shared/components/EmptyGraphMessage'
@@ -22,7 +23,7 @@ import {
   defaultXColumn,
   defaultYColumn,
 } from 'src/shared/utils/vis'
-import {getActiveQuery} from 'src/timeMachine/selectors'
+import {getActiveQueryIndex} from 'src/timeMachine/selectors'
 
 // Constants
 import {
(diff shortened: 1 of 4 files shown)


TCL735:
Sorry, a quick force push to get rid of an unnecessary variable. All good now. TIA for the review.

TCL735:
Based on feedback from a demo, I will be updating the way 'upper' and 'lower' in the Band Options updates. Since only user interaction can save 'upper' and 'lower', it follows that only user interaction should change what is displayed in the dropdown.

Putting this back into WIP temporarily.

TCL735:
Read for review.

src/shared/components/BandPlot.tsx
```diff
@@ -89,20 +90,24 @@ const BandPlot: FC<Props> = ({
...
-        mainColumn,
-        lowerColumnName
-      ),
-    [selectedFunctions, upperColumnName, mainColumn, lowerColumnName]
-  )
-
+  const mainColumnName = useMemo(() => {
+    const aggregateFunctions = get(
+      queries,
+      `${activeQueryIndex}.builderConfig.functions`,
```
drdelambre, line 100: what happens when a band plot is made using raw flux and not the query builder? feels like these wont be defined?

src/shared/components/BandPlot.tsx
```diff
@@ -208,11 +213,9 @@ const BandPlot: FC<Props> = ({
   return children(config)
 }
 
-const mstp = (state: AppState) => {
-  const {builderConfig} = getActiveQuery(state)
-  const {functions} = builderConfig
-  return {selectedFunctions: functions.map(f => f.name)}
-}
+const mstp = (state: AppState) => ({
+  activeQueryIndex: getActiveQueryIndex(state),
```
drdelambre, line 217: why use the index and not keep with the active query?

src/shared/components/BandPlot.tsx
TCL735: Thanks for pointing out this case. Yes, good point. I will make a change. When selectedFunctions are empty (no aggregate functions are found in the builderConfig) and a "mainColumn" is specified, in which case it should just be the "mainColumn"

src/shared/components/BandPlot.tsx
TCL735: Because in Dashboard view, the active query only keeps a single aggregate function in the draft queries. This causes the upper and lower columns to disappear from the cells. The upper and lower columns are available in the props to Band Plot, and so we use that to render upper and lower in cells of Dashboard view.

TCL735: Keeping the index and using it to looking the query in props is the secret sauce

TCL735:
The force push was only to fix merge conflicts.

TCL735:
There is a limitation to Band Options, as pointed out to me by the Docs team. Using the query editor will not populate the correct bands in Band Options. Currently, that is extremely difficult to do and is out of scope for this PR.

I will add helper text to the Band Options indicating that the selectable bands are only based on the query builder and not the query editor.

TCL735:
Better yet: only when in Query Builder, the Band Options will allow selecting the bands. When in Query Editor mode, Band Options will not display the selectors.

src/shared/components/BandPlot.tsx
TCL735: This is actually a mode problem, Query Builder vs Script Editor. The latest commit allows Band Plot to work in all modes.

src/timeMachine/components/view_options/BandOptions.tsx
```diff
@@ -291,17 +347,37 @@ class BandOptions extends PureComponent<Props> {
 
   private onChangeLowerColumn = selectedLowerColumnName =>
     this.onChangeColumn(this.props.onSetLowerColumn, selectedLowerColumnName)
+
+  private onTextChangeUpperColumn = (e: ChangeEvent<HTMLInputElement>) => {
```
hoorayimhelping, line 351: there's very little cost to adding `vent` to `e` (to make it `event`). `e` is really ambiguous in javascript - it often represents `event` or `error`. being able to search for instances of event handlers is much easier when they're named `event` or `evt`. try doing a global search for `e` or `e:`.

src/timeMachine/selectors/index.ts
```diff
@@ -65,6 +68,14 @@ export const getActiveQuery = (state: AppState): DashboardDraftQuery => {
   return draftQueries[activeQueryIndex]
 }
 
+export const getActiveQueryIndex = (state: AppState): number => {
+  const activeTimeMachine = getActiveTimeMachine(state)
+  if (activeTimeMachine) {
+    return activeTimeMachine.activeQueryIndex
+  }
+  return 0
```
hoorayimhelping, line 76: is the active query index always going to default to 0? are the instances when it might be `undefined` or some other value?

drdelambre: theoretically yes, practically no. As in there's no guards against it in the system, but timemachine always starts up with one in the default state and doesn't allow you to delete all of the queries. seems flimsy, but it an ism that holds up a bunch of other logic as well

src/shared/components/BandPlot.tsx
drdelambre: imma need your help validating my port of this to notebooks then, cause they dont share that same schema. later though, and unrelated to this PR

src/timeMachine/components/view_options/BandOptions.tsx
TCL735: Originally, i wanted event, but for some reason my VS Code put a strikethrough over `event`, which puzzled me. That's when I changed it to `e`. Let me see if changing it back now has the same issue.

TCL735: okay it worked this time, no complaints from VS Code.

drdelambre (approved):
sorry man got distracted

PR #70, opened 2020-09-18
feat: filling in the view options on notebooks

drdelambre:
This adds the ability for view options to self register their modules to a view like flow pipes and decouples a lot of the interfaces between them.

now we can use view options outside of a time machine redux context, and changing a property on one view option doesn't need to be cross referenced with the others. we can add and remove these options without fear of breaking something unrelated or having dark code branches floating through the code.

this is the first step in allowing giraffe to be the source of truth for view options from the ui side. following steps would be to move time machine over to using the same code to consolidate the definition of how a view option is rendered to reduce duplication of effort when updating the view layer, consolidating and normalizing view property data schemes (axes.x.label vs xAxisLabel for example) so that we can re-implement view layer code sharing to implement these structures without all of the "if this is a single stat, treat the axis property differently", and finally use those definitions to generate the view based off of the type definitions that come from giraffe (an xy plot has: 1 axes called x and 1 axes called y and some color.... add those normalized interfaces automagically to the view option)


src/flows/components/header/Submit.tsx
```diff
@@ -91,6 +91,10 @@ export const Submit: FC = () => {
           } else if (pipe.type === 'queryBuilder') {
             const {aggregateFunction, bucket, field, measurement, tags} = pipe
 
+            if (!bucket) {
```
Asalem1, line 94: shouldn't we be setting a default bucket? Also, what happens if a bucket's name is `''`

src/flows/pipes/Visualization/types/Gauge/GaugeOptions.tsx
```diff
@@ -0,0 +1,143 @@
...
+                  ? AutoInputMode.Custom
+                  : AutoInputMode.Auto
+              }
+              onChangeMode={() => {}}
+              inputComponent={
+                <Input
+                  name="decimal-places"
+                  placeholder="Enter a number"
+                  onFocus={() => {}}
+                  onChange={() => {}}
```
Asalem1, line 117: For my sanity, can I get a `// TODO(alex): do something`

Asalem1 (approved):
Mama mia, that's a whole lotta pasta

src/flows/components/header/Submit.tsx
drdelambre: the setting of the default bucket should be happening somewhere else (like in a dropdown or something). is it legal to have a bucket with no name? i feel like there would be more issues than this with that idea

Asalem1: 🤷🏽

PR #71, opened 2020-09-18
feat(ui): Add GA tracking and calendly contact

mavarius:
- Changes the messaging for free and PAYG series cardinality limit alerts
- Adds an option for PAYG users to schedule a call in Calendly to resolve series cardinality rate limit breaches
- Adds class names for GA tracking which upgrade CTA buttons were clicked

**FREE**
image

**PAYG**
image


TCL735:
Looks good. I don't think the e2e failures have anything to do with the changes here. So i am okay with no e2e changes.

TCL735 (approved):

PR #77, opened 2020-09-21
chore: bump the docker cache

drdelambre:
turns out "waiting for the cache to expire" would take months. SO. doing it explicitly here. I think the "build ingress on every test" if more of a hold over to the last time we had to bump the cache, and just forgot about it, but the next coarse of action after this would be to remove the debug bump as well.... and maybe just creating this PR will fix it

### DIFF
--- a/cypress/e2e/tasks.test.ts
+++ b/cypress/e2e/tasks.test.ts
@@ -39,7 +39,7 @@ from(bucket: "${name}"{rightarrow}
 
     cy.getByTestID('notification-error').should(
       'contain',
-      'error calling function "to": missing required keyword argument "bucketID"'
+      'missing required keyword argument "bucketID"'
     )
   })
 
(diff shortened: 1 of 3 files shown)


cypress/e2e/telegrafs.test.ts
```diff
@@ -1,7 +1,7 @@
 import {Organization} from '../../src/types'
 
 // a generous commitment to delivering this page in a loaded state
-const PAGE_LOAD_SLA = 10000
+const PAGE_LOAD_SLA = 20000
```
drdelambre, line 4: idk why, but sometimes it takes forever for cypress to load a page. it just sits there with the dark grey background... doing nothing

cypress/e2e/tasks.test.ts
```diff
@@ -39,7 +39,7 @@ from(bucket: "${name}"{rightarrow}
 
     cy.getByTestID('notification-error').should(
       'contain',
-      'error calling function "to": missing required keyword argument "bucketID"'
+      'missing required keyword argument "bucketID"'
```
drdelambre, line 42: an api change seemed to break this test.

hoorayimhelping (approved):

PR #78, opened 2020-09-21
fix: 404 not rendering

121watts:
Closes influxdata/influxdb#19283

### The Problem
When navigating to a non-existing route a blank page displayed.  

### The Solution
Remove the top level `react-router` `<Switch/>` component that returns the first matching component.   In this case that would be `<Setup/>`.  Didn't think it would be this easy but seems to not interfere with `/onboarding` and works with root and nested routes.

Kapture 2020-09-21 at 14 56 27

### DIFF
--- a/src/shared/components/NotFound.tsx
+++ b/src/shared/components/NotFound.tsx
@@ -1,7 +1,7 @@
 import React, {FC} from 'react'
 
 const NotFound: FC = () => (
-  <div className="container-fluid" data-testid="404">
+  <div className="container-fluid" data-testid="not-found">
     <div className="panel">
       <div className="panel-heading text-center">
         <h1 className="deluxe">404</h1>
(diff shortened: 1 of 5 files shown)


src/shared/components/NotFound.tsx
```diff
@@ -1,7 +1,7 @@
-import React, {SFC} from 'react'
+import React, {FC} from 'react'
 
-const NotFound: SFC = () => (
-  <div className="container-fluid">
+const NotFound: FC = () => (
+  <div className="container-fluid" data-testid="404">
```
hoorayimhelping, line 4: small suggestion to keep the naming consistent: the component and the concept is called `NotFound`, not 404. let's not mix terms / levels of abstraction.

```suggestion
  <div className="container-fluid" data-testid="component-not-found">
```

hoorayimhelping (approved):
nice find

kristinarobinson (approved):

PR #79, opened 2020-09-21
test: update community template e2e test

hoorayimhelping:
closes https://github.com/influxdata/influxdb/issues/19606

There are two issues that needed to be fixed to the code under test that using the test templates revealed. Those were fixed, and then the test was committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ node_modules
 vendor
 
 # files
+cypress.env.json
 junit.xml
 npm-error.log
 src/client/generatedRoutes.ts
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
     "tsc": "tsc -p ./tsconfig.json --noEmit --pretty --skipLibCheck",
     "tsc:cypress": "tsc -p ./cypress/tsconfig.json --noEmit --pretty --skipLibCheck",
     "cy": "CYPRESS_baseUrl=http://localhost:9999 cypress open",
-    "cy:dev": "CYPRESS_baseUrl=http://localhost:8080 cypress open",
+    "cy:dev": "CYPRESS_baseUrl=http://kubernetes.docker.internal:8080 cypress open",
```
drdelambre, line 49: whoa!

drdelambre (approved):

.gitignore
```diff
@@ -15,6 +15,7 @@ node_modules
 vendor
 
 # files
+cypress.env.json
```
drdelambre, line 18: thanks for that

PR #80, opened 2020-09-21
feat: metric selector for the future

drdelambre:
this removes the bucket pipe and removes the need for the feature flag for the metric selector

closes influxdata/idpe#8650


alexpaxton (approved):
for glory!

PR #85, opened 2020-09-22
feat(table-fromFlux-parser): adds a data transformation layer to the UI that uses fromFlux to parse CSVs for Tables

### LINKED ISSUE #73: Replace parseResponse in the Raw Data Table to use fromFlux
For more on the background surrounding this issue, please see:

https://github.com/influxdata/ui/issues/39
https://github.com/influxdata/ui/issues/72

### Problem

Unlike the other usages of the `parseResponse` function, the Raw Data Table component relies upon a different schema than what is currently provided by the `fromFlux` function. 

### What needs doing

Since our end goal is still to maintain 1 parser as our source of truth, we will want to create a data transformation layer in the UI that transforms the data returned from the `fromFlux` function into something that the Raw Data Table can ingest

### Criteria for Success

Feature flag and use a data transformation layer that relies upon the `fromFlux` function in order to output the results in the format the is used in the Raw Data Table

Asalem1:
Closes #73 

### Problem
There are a bunch of parsers, which means maintaining a bunch of code. This PR aims to unify the parsers by replacing the usage of parseResponse with fromFlux

### Solution
Created a data transformation layer on the UI that is dependent upon `fromFlux` to parse and return the data. This data transformation layer outputs a FluxTable in accordance with the requirements of the `Table` visualization. 

Swapped where parseResponse was being used for the Table and put it behind a feature flag. The feature flag PR is up in IDPE and the feature flag has been set in configCat


Asalem1:
I'd rather wait for #91 to get merged in so we can reuse some of that functionality here before merging this in

src/shared/parsing/flux/response.ts
```diff
@@ -148,3 +149,172 @@ export const parseTables = (responseChunk: string): FluxTable[] => {
...
+  const dataTypes = {
+    '': '#datatype',
+  }
+  let fluxData
+  const groupSet = new Set(['result', ...fluxGroupKeyUnion])
+  // build out the dataTypes here based on the columnKeys
+  for (let i = 0; i < columnKeys.length; i++) {
+    const column = columnKeys[i]
+    fluxData = table.getOriginalColumnType(column)
+    dataTypes[column] = fluxData
```
drdelambre, line 179: that's a lot of lines to say:
```dataTypes[columnKeys[i]] = table.getOriginalColumnType(columnKeys[i])```

File: src/shared/parsing/flux/response.ts [hunk 2]
```diff
@@ -148,3 +149,172 @@ export const parseTables = (responseChunk: string): FluxTable[] => {
...
+  }
+  let columnType: any
+  let columnValues
+  // build out the columnHeaders
+  let columnHeaders = columnKeys.filter(col => {
+    columnType = table.getColumnType(col)
+    columnValues = table.getColumn(col, columnType)
+    return columnValues[valueIndex] !== undefined
+  })
```
drdelambre, line 185: maybe readability could be helped with the format of
```
})


// description of what i'm doing after some whitespace
let variables = { be: decided, down: here }
{
 transforming some data in here
}

```
then operations take on some visual weighting scheme. might not work, idk

drdelambre (approved):

PR #98, opened 2020-09-23
feat(raw-csv-parser): created a data transformation layer based around fromFlux to use as the Raw CSV parser

### LINKED ISSUE #91: Replace parseFiles with fromFlux-based data parser
For more on the background surrounding this issue, please see:

#39
#72
#73 

### Problem
Unlike the previous issues, the Raw Data Table component relies upon a different schema than what is currently provided by the fromFlux function.

### What needs doing
Since our end goal is still to maintain 1 parser as our source of truth, we will want to create a data transformation layer in the UI that transforms the data returned from the fromFlux function into something that the Raw Data Table can ingest

### Criteria for Success
Feature flag and use a data transformation layer that relies upon the fromFlux function in order to output the results in the format the is used in the Raw Data Table

Asalem1:
Closes #91 

### Problem
There are a bunch of parsers, which means maintaining a bunch of code. This PR aims to unify the parsers by replacing the usage of parseFiles with a fromFlux-based data transformation layer for raw CSVs

Solution
Created a data transformation layer on the UI that is dependent upon fromFlux to parse and return the data. This data transformation layer outputs a `ParseFilesResult` in accordance with the requirements of the Raw CSV Table view.

Swapped where parseFiles was being used for the Raw CSV Table and put it behind a feature flag. The feature flag PR is up in IDPE and the feature flag has been set in configCat


Asalem1:
I'd rather get this in before #85 so that we can reuse some of these functions there

src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -14,6 +15,107 @@ export interface ParseFilesResult {
...
+export const parseFilesWithFromFlux = (
+  responses: string[]
+): ParseFilesResult => {
+  const chunks = parseChunks(responses.join('\n\n'))
+  const parsedChunks = chunks.map(c => fromFluxTableTransformer(c))
+  let data = []
+  parsedChunks.forEach((chunk, i) => {
+    data = data.concat(chunk)
+    if (i !== parsedChunks.length - 1) {
+      data.push([])
```
Asalem1, line 27: an empty array signifies that the following data set should be treated as its own table

src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -14,6 +15,107 @@ export interface ParseFilesResult {
   maxColumnCount: number
 }
 
+export const parseFilesWithFromFlux = (
+  responses: string[]
+): ParseFilesResult => {
+  const chunks = parseChunks(responses.join('\n\n'))
```
drdelambre, line 21: this throws off a million red flags

Asalem1: It’s the basis for the current parsers. I can take a look at removing it since I now feel distant enough from it to see that we can probably just iterate through the results and achieve the same outcome

src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -14,6 +15,165 @@ export interface ParseFilesResult {
...
+  column: string,
+  table: Table,
+  index: number
+): boolean => {
+  const columnType: any = table.getColumnType(column)
+  const values = table.getColumn(column, columnType)
+  return values[index] !== undefined
+}
+
+const buildGroupHeaders = (
```
drdelambre, line 135: not a blocker, but functions don't come for free. when dealing with large datasets, you wanna minimize the amount of memory duplication in temporary allocations (like how the reduce function duplicates acc every time instead of operating on memory in place) as these dont get removed immediately (i forget the default chrome garbage collection frequency) and the number of function calls as these create similar temporary [memory allocations](https://en.wikipedia.org/wiki/Call_stack).

for loops > helper functions, but only when performance matters more than ergonomics

drdelambre (approved):

PR #102, opened 2020-09-24
fix: prevent invisible tooltips for the zero value of legendOpacity from the api

TCL735:
The initial value from the backend api comes back as 0 for opacity, which doesn't change in props, causing legends to appear invisible. Prevent this by ensuring minimum opacity is always used.

### DIFF
--- a/src/shared/components/XYPlot.tsx
+++ b/src/shared/components/XYPlot.tsx
@@ -12,7 +12,10 @@ import {
 import EmptyGraphMessage from 'src/shared/components/EmptyGraphMessage'
 
 // Utils
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {
+  useLegendOpacity,
+  useLegendOrientationThreshold,
+} from 'src/shared/utils/useLegendOrientation'
 import {
   useVisXDomainSettings,
   useVisYDomainSettings,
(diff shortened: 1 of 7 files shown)


alexpaxton (approved):
LGTM

src/shared/components/XYPlot.tsx
```diff
@@ -85,12 +86,15 @@ const XYPlot: FC<Props> = ({
...
-  }, [legendOpacity])
+  const tooltipOpacity = Math.max(
+    useMemo(() => {
+      if (isFlagEnabled('legendOrientation')) {
+        return legendOpacity
+      }
+      return LEGEND_OPACITY_DEFAULT
+    }, [legendOpacity]),
+    LEGEND_OPACITY_MINIMUM
+  )
```
121watts, line 97: All of these are doing the exact same thing.  I would pull them out into their own hook with a `useCallback` whose dependency array is contains `legendOpacity`.  Then each of these files would read something like: 

`const tooltipOpacity = useLegendOpacity(legendOpacity)`

121watts (changes requested):

src/shared/components/XYPlot.tsx
TCL735: Ok, will do.

Also, unrelated. A better default is a solid looking tooltip rather than a barely visible 0.2

121watts (approved):
if this gets much larger we might want to pull all of this into something like `useLegendSettings`.  Looks great.

TCL735:
The force push was to fix merge conflicts.

PR #105, opened 2020-09-25
feat: fixing the template.md link

zoesteinkamp:
closes #19599
This fixes the template.md file, originally it tried to search for it on influxdb, but now it routes to github properly

### DIFF
--- a/cypress/e2e/communityTemplates.test.ts
+++ b/cypress/e2e/communityTemplates.test.ts
@@ -211,6 +211,8 @@ describe('Community Templates', () => {
     it.skip('takes you to github readme when you click on the Community Templates button', () => {
       cy.getByTestID('community-template-readme-overlay-button').click()
       cy.get('.markdown-format').should('contain', 'Setup Instructions')
+
+      //we should confirm the link works
     })
 
     it('deletes templates', () => {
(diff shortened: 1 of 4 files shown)


cypress/e2e/communityTemplates.test.ts
```diff
@@ -211,6 +211,8 @@ describe('Community Templates', () => {
     it.skip('takes you to github readme when you click on the Community Templates button', () => {
       cy.getByTestID('community-template-readme-overlay-button').click()
       cy.get('.markdown-format').should('contain', 'Setup Instructions')
+
+      //we should confirm the link works
```
hoorayimhelping, line 215: what's this? is this a legit todo?

src/templates/utils/index.test.ts
```diff
@@ -239,4 +240,14 @@ describe('validating template URLs', () => {
       ).toBe(TEMPLATE_URL_VALID)
     })
   })
+
+  describe('validating template readme Formatter', () => {
```
hoorayimhelping, line 244: might consider adding another example, maybe with some text below the `## Setup Instructions`.

hoorayimhelping (approved):
great work!

cypress/e2e/communityTemplates.test.ts
zoesteinkamp: should have added a better note, but yes its to check the link works and leads to the proper template.md file. Though we have this as a unit test, i guess we should do it in the e2e possibly

PR #114, opened 2020-09-26
fix: keep formatStatValue in sync with same file in Giraffe

TCL735:
Closes https://github.com/influxdata/giraffe/issues/295

Stay in sync with Giraffe


hoorayimhelping:
Why is this necessary?

TCL735:
We will, very soon. The next step in the migration of all visualizations into Giraffe is the actually consume the ones that have been moved over, including Single Stat, Gauge (these two will use `formatStatValue`), RawFluxDataTable, TableGraph.

In the meantime, this little bit of housekeeping insures that whoever is involved in the work of the migration would not have to spend time figuring out which `formatStatValue` is the correct version. Keeping them in lockstep ensures that any choice is the correct choice.

hoorayimhelping (approved):

PR #116, opened 2020-09-28
fix(redux): export store singleton, rather than configureStore factory function

### LINKED ISSUE #124: Never import from index.tsx
### Background
The redux store is created in `index.tsx` the entry point of the app. The store is exported, and a couple of modules import the store from `index.tsx`. This is extremely confusing to reason about. It's probably not a problem from an prod execution standpoint, as ES6 modules [only evaluate a module once.](https://stackoverflow.com/a/37325716/92446) This is probably an issue in testing environments

### How to Fix
Move the central point where the store is created out of index.tsx and make sure it's only created once.

### Acceptance Criteria
- `index.tsx` exports no modules
- `configureStore.ts` configures the store a single time and returns copies of the store
- code that imported the store from `index.tsx` should now import it from `configureStore.ts`

hoorayimhelping:
Closes #124

* Previously, `index.tsx` exported `store` after configuring it. Modules would then import `store` from `index.tsx.`
  * I just want to reiterate here: modules **included the entry point of the app which automatically runs and includes the module that included `index.tsx` as part of the code that runs**
  * Other modules would import `configureStore` and re-configure the store, which overwrote the store willy nilly.
* Now: `store` is a true [singleton.](https://en.wikipedia.org/wiki/Singleton_pattern) It's created exactly once, and the instance of the created `store` is returned via an exported function. Any place that imported `store` from `index.tsx` or the function `configureStore` now imports the `getStore` function.
* This also includes a change to `pageTitles` and a test for it. This PR initially started as change for how `pageTitles` handles there not being an org in store (it crashed before). 

### Testing
I got paranoid that this code would literally ruin everything, so to test it, I updated `configureStore` to look something like:

```ts
export const getStore = () => {
  if (!storeSingleton) {
    console.log('NO SINGLETON')
    storeSingleton = configureStore()
  }

  console.log('configureStore.ts getStore', {...storeSingleton.getState().resources.orgs.org})

  return storeSingleton
}
```
The console does in fact show `NO SINGLETON` only once.

### DIFF
--- a/src/store/configureStore.ts
+++ b/src/store/configureStore.ts
@@ -9,6 +9,8 @@ import {queryStringConfig} from 'src/shared/middleware/queryStringConfig'
 import sharedReducers from 'src/shared/reducers'
 import persistStateEnhancer from './persistStateEnhancer'
 
+import {loadLocalStorage} from 'src/localStorage'
+
 // v2 reducers
 import meReducer from 'src/shared/reducers/me'
 import flagReducer from 'src/shared/reducers/flags'
@@ -122,22 +124,9 @@ export const rootReducer = (history: History) =>
 const composeEnhancers =
   (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose
 
-let _store
-
-// NOTE: just used to reset between tests
-export function clearStore() {
-  _store = null
-}
-
-export default function configureStore(
-  initialState?: LocalStorage
+function configureStore(
+  initialState: LocalStorage = loadLocalStorage()
 ): Store<AppState & LocalStorage> {
-  // NOTE: memoizing helps keep singular instances of the store
-  // after initialization, or else actions start failing for reasons
-  if (_store) {
-    return _store
-  }
-
   const routingMiddleware = routerMiddleware(history)
   const createPersistentStore = composeEnhancers(
     persistStateEnhancer(),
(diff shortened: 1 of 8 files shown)


src/store/configureStore.ts
```diff
@@ -152,6 +141,18 @@ export default function configureStore(
   // https://github.com/elgerlambert/redux-localstorage/issues/42
   // createPersistentStore should ONLY take reducer and initialState
   // any store enhancers must be added to the compose() function.
-  _store = createPersistentStore(rootReducer(history), initialState)
-  return _store
+  return createPersistentStore(rootReducer(history), initialState)
+}
+
+let storeSingleton
```
Asalem1, line 147: I'm curious why the storeSingleton is scoped outside the `getStore` function

Asalem1 (approved):
Way to go. Super impressed

121watts:
If we really want to test that `getStore` is a singleton we could `expect(getStore() === getStore()).toBe(true)`

src/store/configureStore.ts
121watts: How do you propose to define it within `getStore`?

Asalem1: Ya, no. I see why now. Singleton 🎉

121watts (approved):

PR #117, opened 2020-09-28
feat: move raw results view to all notebook panels

### LINKED ISSUE #110: Flows: Add Raw data view to notebook panels
we want this ish to be darn near universal in the interface. make the raw table view interaction in the flux script pipe show up on the bottom of any panel if the type registration for the pipe belongs to the one of the following families: inputs, test, transform. seems like a small subset, but we dont have a lot of pipe types yet.

the verbage around the loading states is gonna be all messed up as we drop the component into a new context, just ping paxton before you ship it and he'll fill in the blanks

drdelambre:
closes #110 

took a product change of direction and decided that the ability to view the results of each panel is an affordance required of all panels within the system in order to view the transformations you're making. this moves the responsibility of ownership of this info away from the pipe and onto the notebook. I'm not certain if this meta data should persist refresh cycles, seems like it would be weird if two users were editing the same notebook, so i'm keeping that state local.


src/flows/components/panel/FlowPanel.tsx
```diff
@@ -140,17 +142,27 @@ const FlowPanel: FC<Props> = ({id, children, controls}) => {
...
   ) {
     return null
   }
+
   return (
     <ClickOutside onClickOutside={handleClickOutside}>
       <div className={panelClassName} onClick={handleClick} ref={panelRef}>
         <FlowPanelHeader id={id} controls={controls} />
         <div className="flow-panel--body">{children}</div>
+        {showResults && (
```
Asalem1, line 161: It feels like this could exist in `Results` and make the component itself responsible for rendering?

Asalem1: JK that might make it confusing since the `Results` component is responsible for some pagination visibility

src/flows/pipes/TestFlux/view.tsx
```diff
@@ -91,6 +91,10 @@ const TestFlux: FC<PipeProp> = ({Context}) => {
         emptyText="This cell will visualize results from uploaded CSVs"
         emptyIcon={IconFont.BarChart}
         toggleVisibilityEnabled={false}
+        height={data.panelHeight}
+        onUpdateHeight={panelHeight => update({panelHeight})}
+        visibility={data.panelVisibility}
```
Asalem1, line 96: why not make the height and visibility component-based like you did here:

https://github.com/influxdata/ui/pull/117/files#diff-f62f29439ccd6eea91c6b82414e62b0cR82-R84

Asalem1 (approved):
Few comments, but looks good.

src/flows/pipes/TestFlux/view.tsx
drdelambre: because i totally forgot about that component. good catch!

drdelambre: oh nope, i found out why. it's connected to the display height of the panel itself, not just the results of the panel. it makes more sense on the visualization component this was cloned from, but the user should be able to resize the panel contents, and have that persist across reloads

Asalem1: That makes sense. I’m assuming it’s the same reason as other view file 👍🏽

PR #118, opened 2020-09-29
fix: removing focus logic to help with usability a bit

drdelambre:
there was a change added that tried to remove a scroll trap within the flows interface when using the metrics selector. it... introduced it's own set of issues, so this PR reverts those changes

### DIFF
--- a/src/notebooks/pipes/Data/SearchBar.tsx
+++ b/src/notebooks/pipes/Data/SearchBar.tsx
@@ -1,15 +1,8 @@
 // Libraries
-import React, {FC, useContext, useState, useRef, KeyboardEvent} from 'react'
+import React, {FC, useContext} from 'react'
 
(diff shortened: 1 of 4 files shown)


Asalem1 (approved):
If you say so. I totally thought we wanted to conditionally set results based on whether the search bar was being focused but what'd I know 🤷🏽

drdelambre:
bigger picture, we dont want to introduce scroll traps on the interface, as it's meant to be one long interaction, so scrolling should be quick and defined to the page. if you break that paradigm, you find yourself trying to scroll on the page, and just seeing the results scrolling forever instead. the searchbar focus was a way to try and reduce the effect of that scroll trap, but it introduced a state where clicking on anything around the page, made the page jump around and forget what you clicked on.

PR #121, opened 2020-09-29
fix(cancel-query): using hotkeys to submit a query are now cancelable

Asalem1:
Closes #https://github.com/influxdata/idpe/issues/8328

### Problem

Using the command + enter keys would not render the cancel button, rendering different functionality for different methods of submitting a query

### Solution

Updated the way the cancel query button is being rendered by setting a timeout on a reducer state so that the property can be toggled outside of the component. Also updated the abort controller so that we can make use of the existing cancellation infrastructure to cancel an existing query.

This PR also addresses cancellation of queries that were issued using the refresh query button. This PR **does not** address the issue where queries are issued when a timerange is changed

cancel-fix

### DIFF
--- a/src/mockState.tsx
+++ b/src/mockState.tsx
@@ -18,6 +18,7 @@ export const localState: LocalStorage = {
     ephemeral: {
       inPresentationMode: false,
       hasUpdatedTimeRangeInVEO: false,
+      shouldShowCancelBtn: false,
     },
     persisted: {
       autoRefresh: 0,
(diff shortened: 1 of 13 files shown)


Asalem1:
Gonna pump the brakes on this since there's an edge case to consider for styling when a user repeatedly hits command + enter (and also some issues to consider when integrating this functionality with flows)

Asalem1:
Gonna close this PR since it's confusing and is no longer relevant

PR #122, opened 2020-09-29
fix: set default hoverDimension for Line Options to 'auto'

TCL735:
Closes https://github.com/influxdata/influxdb/issues/19178

Previously, hoverDimension as a variable was defaulted to have 'auto as a value. But this doesn't actually set the redux state with the correct default hoverDimension. Default view properties should be set in **src/views/helpers/index.ts** which will reflect in redux state.


russorat (approved):
yesssss! LGTM

PR #128, opened 2020-09-29
fix: allows one to deselect tags individually

### LINKED ISSUE #81: Flows: Metric selector deleteing one tag deletes all tags
WHEN i click the delete option on a tag in metrics selector
THEN it should delete only that tag and not all tags

Kapture 2020-09-21 at 16 03 45

drdelambre:
closes #81 

cleans up the filtering branching

### DIFF
--- a/src/notebooks/pipes/Data/FilterTags.tsx
+++ b/src/notebooks/pipes/Data/FilterTags.tsx
@@ -64,36 +64,33 @@ const constructFilters = (value: string, type: string) => {
   }
 }
 
+interface SelectedTags {
+  [key: string]: string[]
+}
+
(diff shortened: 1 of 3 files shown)


Asalem1 (commented):
I'm a little confused. This should've already been doing that. Is this just to simplify the code a bit or was there a bug when command + clicking to deselect?

Asalem1 (approved):
Super weird. Just saw the issue. Not sure how this broke but 🤷🏽

drdelambre:
the cmd + clicking to toggle a selection from within the schema window was working, just not when you clicked on the delete from the top list of applied filters.

PR #129, opened 2020-09-30
Fix: Add event for template install and edit event for template rename

zoesteinkamp:


src/templates/components/CommunityTemplateInstallOverlay.tsx
```diff
@@ -122,6 +122,7 @@ class CommunityTemplateInstallOverlayUnconnected extends PureComponent<Props> {
         this.props.resourcesToSkip,
         this.props.stagedTemplateEnvReferences
       )
+      event('template_install', {templateID: this.props.org.id})
```
hoorayimhelping, line 125: this code is calling the org id the `templateId`. I think the url might be appropriate: 

```ts
event('template_install', {templateUrl: this.props.stagedTemplateUrl})
```

hoorayimhelping (changes requested):

hoorayimhelping (approved):

PR #130, opened 2020-10-01
fix: moving definition of query into the pipe definition

drdelambre:
it was weird that a button hiding somewhere in the header defined the transformation of notebook data into flux queries, but the final straw was when a type was deleted, but it's data transformation code lingered past it's depreciation because of this disconnect. This pushes the onus onto the pipe definition to determine how the query is being generated. the ultimate solution pushes the generation of this flux to the backend, so that we can decouple the flux aspect, but this should isolate it in the mean time.


Asalem1 (approved):
🎉

src/flows/context/query.tsx
```diff
@@ -41,6 +51,58 @@ export const QueryProvider: FC<Props> = ({children, variables, org}) => {
...
+        } else {
+          append()
+        }
+
+        return stages
+      }, [])
+      .map(queryStruct => {
+        const queryText =
+          Object.entries(queryStruct.requirements)
+            .map(([key, value]) => `${key} = (\n${value}\n)\n\n`)
```
Asalem1, line 96: I heard you like new lines

drdelambre: tell me about it

PR #142, opened 2020-10-02
refactor(redux): move me actions with side effects to meThunks

### LINKED ISSUE #140: Tech Debt: remove side effects from me action creators
### Background
The `shared/actions/me.ts` action has a function with side effects in it. This should be moved to the `shared/actions/meThunks.ts` thunks file, which needs to be created.

See https://github.com/influxdata/ui/issues/131.

### Acceptance Criteria
- `shared/actions/meThunks.ts` is created
- `getMe` is moved to `meThunks`
  - any files that imported this function from `shared/action/predicates` will need to update its import
- `madge --webpack-config webpack.common.ts src/store/configureStore.ts --image configure_store_new.svg --circular` produces a dependency graph where `me` is absent or points between reducers and actions

hoorayimhelping:
closes #140

Makes `me` action creators pure by moving the single function with side effects into a thunk

#### Old dependency graph
configure_store_new

#### New dependency graph
configure_store_new

```sh
madge --webpack-config webpack.common.ts src/store/configureStore.ts --image graph.svg --circular
```

#### How does this break dependencies?
- `configureStore` builds the empty store based off the initial state the reducers describe
- to do this, `configureStore` imports all the reducers.
- reducers pull in the action creator functions and the name of the actions
  - if the actions have outside dependencies (say on `event`, which indirectly depends on `featureFlag`), those are imported and executed before `configureStore` is executed
- by moving the side effects (and more importantly, their dependencies) outside the dependency chain of reducers and action creators,  circular dependencies that flow back to `configureStore` are removed.

### DIFF
--- a/src/shared/actions/me.ts
+++ b/src/shared/actions/me.ts
@@ -1,7 +1,4 @@
 import {MeState} from 'src/shared/reducers/me'
-import {client} from 'src/utils/api'
-import HoneyBadger from 'honeybadger-js'
-import {updateReportingContext, gaEvent} from 'src/cloud/utils/reporting'
 
 export const SET_ME = 'SET_ME'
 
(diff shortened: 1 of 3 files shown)


zoesteinkamp (approved):
Love the graph!

PR #143, opened 2020-10-02
refactor(redux): remove side effects from predicates actions

### LINKED ISSUE #131: Tech Debt: Untangle the web of circular redux dependencies
### Background
The circular dependencies of our redux store imports are causing odd loading behavior where `featureFlag` loads before `configureStore` (which it depends on) has been executed. `featureFlag` being executed before `configureStore` seems to point to a [cyclic dependency.](https://exploringjs.com/impatient-js/ch_modules.html#esms-transparent-support-for-cyclic-imports-advanced)

**The graph above represents the circular dependencies found in the imports of `configureStore.ts`. To see them, follow every arrow that flows into `configureStore.ts` outwards to see which components produce circular dependencies.**

The graph was created by using [madge](https://github.com/pahen/madge) and running `node_modules/madge/bin/cli.js --webpack-config webpack.common.ts src/store/configureStore.ts  --image configure_store_new.svg --circular`

#### Reproduction Instructions
* add `console.log('featureFlag')` to `featureFlag.ts` anywhere in the file
* add `console.log('configureStore')` to `configureStore.ts` anywhere in the file
* load any page and check your JavaScript console. it should show featureFlag loading before configureStore
  ```
     featureFlag
     configureStore
  ```

### What to actually do
Consider `demoData`. To find it, find `configureStore.ts` in the above image, and follow it to `cloud/reducers/demoData.ts`. There is a circular dependency there that is imported from `cloud/reducers/demoData.ts` via `cloud/actions/demoData.ts` via `shared/utils/errors.ts`. `shared/utils/errors.ts` depends on `cloud/utils/reporting.ts`, which includes `cloud/apis/reporting.ts` which includes `shared/utils/featureFlag.ts` which is loaded before `configureStore.ts` (that is a pretty gnarly dependency chain). So we look at `cloud/actions/demoData.ts` and find the dependency on `shared/utils/errors.ts`. [There is a direct one](https://github.com/influxdata/ui/blob/master/src/cloud/actions/demodata.ts#L29) and [there is an indirect one.](https://github.com/influxdata/ui/blob/master/src/cloud/actions/demodata.ts#L27)

To actually fix this, we have to break the dependency. There are a couple of issues here - action creators shouldn't really have side effects like calling `event` or `reportError`. Those belong in components, or if they need to be repeated in multiple places, thunks. So the solution here would be to either create a `demoData/thunks.ts` thunk and move the side effects like `event` and `reportErrorThroughHoneybadger` into there, or skip the thunk and move the side effects to the components that call these action creators. But the overall solution is to remove the side effects that create a circular dependency from the action creator.

### Acceptance Criteria
* `configureStore` logs its message before `featureFlag` meaning it is properly loaded and evaluated before `featureFlag`, which depends on it.
* the circular dependencies in redux are resolved
* action creators are pure and have no side effects

### LINKED ISSUE #138: Tech Debt: remove side effects from predicates action creators
### Background
The `shared/actions/predicates.ts` actions have functions with side effects in them. These should be moved to the `shared/actions/predicateThunks.ts` thunks file, which needs to be created.

See https://github.com/influxdata/ui/issues/131.

### Acceptance Criteria
- `shared/actions/predicateThunks.ts` is created
- `deleteWithPredicate`, `executePreviewQuery`, `setBucketAndKeys`, `setValuesByKey`, are all moved to `predicateThunks`
  - any files that imported these four functions from `shared/action/predicates` will need to update its import
- `madge --webpack-config webpack.common.ts src/store/configureStore.ts --image configure_store_new.svg --circular` produces a dependency graph where `predicates` is absent or points between reducers and actions

hoorayimhelping:
closes #131
closes #138

Makes `predicates` action creators pure by moving the functions with side effects into `predicatesThunks`

#### Old dependency graph
configure_store_new

#### New (and final) dependency graph - note the lack of `cofingureStore.ts`
new

```sh
madge --webpack-config webpack.common.ts src/store/configureStore.ts --image graph.svg --circular
```

#### How does this break dependencies?
- `configureStore` builds the empty store based off the initial state the reducers describe
- to do this, `configureStore` imports all the reducers.
- reducers pull in the action creator functions and the name of the actions
  - if the actions have outside dependencies (say on `event`, which indirectly depends on `featureFlag`), those are imported and executed before `configureStore` is executed
- by moving the side effects (and more importantly, their dependencies) outside the dependency chain of reducers and action creators,  circular dependencies that flow back to `configureStore` are removed.

### DIFF
--- a/src/dataExplorer/components/DeleteDataOverlay.tsx
+++ b/src/dataExplorer/components/DeleteDataOverlay.tsx
@@ -17,11 +17,8 @@ import {convertTimeRangeToCustom} from 'src/shared/utils/duration'
 import {AppState, ResourceType} from 'src/types'
 
 // Actions
-import {
-  resetPredicateState,
-  setTimeRange,
-  setBucketAndKeys,
-} from 'src/shared/actions/predicates'
+import {resetPredicateState, setTimeRange} from 'src/shared/actions/predicates'
+import {setBucketAndKeys} from 'src/shared/actions/predicatesThunks'
 
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = RouteComponentProps<{orgID: string}> & ReduxProps
--- a/src/shared/actions/predicates.ts
+++ b/src/shared/actions/predicates.ts
@@ -1,31 +1,5 @@
-// Libraries
-import {Dispatch} from 'react'
-import {extractBoxedCol} from 'src/timeMachine/apis/queryBuilder'
-import moment from 'moment'
-
-// Utils
-import {postDelete} from 'src/client'
-import {runQuery} from 'src/shared/apis/query'
-import {getWindowVars} from 'src/variables/utils/getWindowVars'
-import {buildVarsOption} from 'src/variables/utils/buildVarsOption'
-import {getVariables, asAssignment} from 'src/variables/selectors'
-import fromFlux from 'src/shared/utils/fromFlux'
-import {getOrg} from 'src/organizations/selectors'
-
-// Actions
-import {notify} from 'src/shared/actions/notifications'
-
-// Constants
-import {
-  predicateDeleteFailed,
-  predicateDeleteSucceeded,
-  setFilterKeyFailed,
-  setFilterValueFailed,
-} from 'src/shared/copy/notifications'
-import {rateLimitReached, resultTooLarge} from 'src/shared/copy/notifications'
-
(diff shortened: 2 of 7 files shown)


Asalem1 (approved):
Considering this isn't actively being used in our app, I feel confident that we can merge this in ASAP

PR #146, opened 2020-10-05
chore(queryCache): remove dashboard queryCache from the UI

Asalem1:
The Dashboard Query Cache has been on for a few weeks without any reported issues. This is just some clean-up to remove the feature flag from the UI and establish this cache as the norm


drdelambre (approved):
:fire:

PR #149, opened 2020-10-06
feat: deploy with circle ci

drdelambre:
working on moving as much stuff as possible off of jenkins and over to circle, because our neglect of jenkins as gotten to the point where it's upsetting day to day operations.
this extends our current testing pipeline to add a conditional workflow that only runs on master. it builds the production image in tandem with the testing image, waits for the tests to pass, then deploys the production image.

currently not gating on e2e passing until that gets more reliable (mimics current behavior in production pipeline)

depends on https://github.com/influxdata/monitor-ci/pull/99, but like.. real close to each other

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,10 @@
 version: 2.1
 
 jobs:
-  cloud-e2e:
+  build-image:
     machine:
-      image: ubuntu-1604:202007-01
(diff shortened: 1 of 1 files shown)


drdelambre:
ugh.. merge seemed to have failed. fixing

121watts (approved):

PR #151, opened 2020-10-07
fix: fixing the 404 not displaying

### LINKED ISSUE #119: 404 page has an accessibility issue with its H1
### Issue
The 404 page has an `h1` that says 404. The `h1` here is an accessibility issue since it conflicts with another `h1` that should take precedence.

### Acceptance Criteria
- the NotFound page's heading is an H2 or smaller
- the NotFound page is hidden from screen readers and automated tools
  - [`aria-hidden`](https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/hidden/) might be viable
  - changing the logic to render nothing, rather than a hidden 404 page might be better

zoesteinkamp:
closes #119 

This fixes the 404 page so it now properly routes and does not get called everytime the page loads

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -14,7 +14,6 @@ import {history} from 'src/store/history'
 
 // Components
 import Setup from 'src/Setup'
-import NotFound from 'src/shared/components/NotFound'
 import GetLinks from 'src/shared/containers/GetLinks'
 
 // Utilities
@@ -70,9 +69,9 @@ class Root extends PureComponent {
     return (
       <Provider store={getStore()}>
         <ConnectedRouter history={history}>
+          {/* <Switch></Switch> */}
           <Route component={GetLinks} />
           <Route component={Setup} />
-          <Route component={NotFound} />
         </ConnectedRouter>
       </Provider>
     )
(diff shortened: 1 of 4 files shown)


src/shared/components/NotFound.tsx
```diff
@@ -4,7 +4,7 @@ const NotFound: FC = () => (
   <div className="container-fluid" data-testid="not-found">
     <div className="panel">
       <div className="panel-heading text-center">
-        <h1 className="deluxe">404</h1>
```
drdelambre, line 7: dont need this change. if the 404 is the only thing rendered, then we want to keep the header, as it is now the page

src/index.tsx
```diff
@@ -72,7 +71,6 @@ class Root extends PureComponent {
         <ConnectedRouter history={history}>
           <Route component={GetLinks} />
           <Route component={Setup} />
-          <Route component={NotFound} />
```
drdelambre, line 75: we still need to figure out a way to add this in nicely for when i go to `/cool/kids` instead of `/org/:orgID/cool/kids`. or does this still work in that example?

src/shared/components/NotFound.tsx
zoesteinkamp: fixed

src/index.tsx
zoesteinkamp: this should work for this example, i got it to work for both types of invalid urls we discussed

drdelambre (approved):

PR #154, opened 2020-10-07
feat(cancel-btn): cancel button can now cancel queries no matter how they are issued

Asalem1:
Closes https://github.com/influxdata/idpe/issues/8328

### Problem

The cancel query button was previously being toggled onClick, making it not cancellable if a query is triggered some other way

### Solution

Created a cache to store queryIDs and abortControllers that can be referenced by the cancel button.

cancel-btn

### DIFF
--- a/src/flows/context/query.tsx
+++ b/src/flows/context/query.tsx
@@ -13,6 +13,7 @@ import {fromFlux as parse} from '@influxdata/giraffe'
 import {event} from 'src/cloud/utils/reporting'
 import {FluxResult} from 'src/types/flows'
 import {PIPE_DEFINITIONS} from 'src/flows'
+import {generateHashedQueryID, setQueryByHashID} from 'src/timeMachine/actions/queries'
 
 interface Stage {
   text: string
@@ -106,9 +107,12 @@ export const QueryProvider: FC<Props> = ({children, variables, org}) => {
   const query = (text: string) => {
     const windowVars = getWindowVars(text, vars)
     const extern = buildVarsOption([...vars, ...windowVars])
+    const queryID = generateHashedQueryID(text, variables, org.id)
 
     event('runQuery', {context: 'flows'})
-    return runQuery(org.id, text, extern)
+    const result = runQuery(org.id, text, extern)
+    setQueryByHashID(queryID, result)
+    return result
       .promise.then(raw => {
         if (raw.type !== 'SUCCESS') {
           throw new Error(raw.message)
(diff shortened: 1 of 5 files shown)


drdelambre:
first

drdelambre:
how do we wrap this in a feature flag?

Asalem1:
> how do we wrap this in a feature flag? other than that and some angry robots, it all looks good.

Good call. Gonna get on that

src/flows/context/query.tsx
```diff
@@ -106,10 +110,13 @@ export const QueryProvider: FC<Props> = ({children, variables, org}) => {
   const query = (text: string) => {
     const windowVars = getWindowVars(text, vars)
     const extern = buildVarsOption([...vars, ...windowVars])
+    const queryID = generateHashedQueryID(text, variables, org.id)
 
     event('runQuery', {context: 'flows'})
-    return runQuery(org.id, text, extern)
-      .promise.then(raw => {
+    const result = runQuery(org.id, text, extern)
```
Asalem1, line 116: not feature flagging here since the entire feature is flagged

src/timeMachine/components/SubmitQueryButton.tsx
```diff
@@ -61,15 +79,19 @@ class SubmitQueryButton extends PureComponent<Props> {
...
   public render() {
     const {text, queryStatus, icon, testID, className} = this.props
-
-    if (queryStatus === RemoteDataState.Loading && this.state.timer === true) {
+    if (queryStatus === RemoteDataState.Loading && this.state.timer) {
       return (
         <Button
           text="Cancel"
           className={className}
-          icon={icon}
```
Asalem1, line 72: not being used by the cancel button

src/flows/context/query.tsx
drdelambre: flagged by the same flag?

Asalem1: It's flagged under the flows feature flag, which basically means we're enabling the feature for anyone who has access to flows

drdelambre: that's not the right way of thinking about flags. we shouldn't have to turn off flows to disable the side effects of the cancel button changing. they should be able to be toggled independently from each other, not enabled / disabled by convention

Asalem1: k

drdelambre (approved):

PR #157, opened 2020-10-08
feat(bucket-selector): added a bucket selector to flows

Asalem1:
Added a generic bucket selector to flows as the `output to bucket` option

output-to-bucket


src/flows/components/header/Submit.tsx
```diff
@@ -25,9 +25,15 @@ export const Submit: FC = () => {
   const time = timeContext[id]
   const tr = !!time && time.range
 
+  const hasQueries = flow.data.all
```
drdelambre, line 28: what does this do?

alexpaxton (approved):
NAIS

src/flows/components/header/Submit.tsx
alexpaxton: I was gonna ask the same

src/flows/pipes/Data/index.ts
```diff
@@ -17,7 +17,10 @@ export default register => {
     },
     generateFlux: (pipe, create, _append) => {
       const {aggregateFunction, bucket, field, measurement, tags} = pipe
-
+      if (!bucket) {
+        create('')
```
drdelambre, line 21: does this send off a network request query of ``''`?

src/flows/components/header/Submit.tsx
Asalem1: So glad you asked. We are using it to disable the `Submit` button down below (where it was moved from), but in this case it's being used to prevent the `submit` function from being called when it's mounted:

https://github.com/influxdata/ui/pull/157/files#diff-d80db76fbd0677f114efeb2cd097c4aeR32-R36

Basically, creating an empty metric selector pipe and refreshing the page was causing an empty query to run (which is an added charge), and was also returning an error because the query was empty

src/flows/pipes/Data/index.ts
Asalem1: I believe so. This was more to guard against times when a bucket isn't selected, do you think we should just return in those cases?

drdelambre: yeah try that out. there's also this example for making a query pipe a passthrough when there's no data: https://github.com/influxdata/ui/blob/master/src/flows/pipes/Query/index.ts#L35

drdelambre (approved):

PR #161, opened 2020-10-12
chore(remove-fromFluxParseReponse): removed parser feature flag

Asalem1:
This PR is part of a series of PRs aim to reduce the number of parsers in use in the UI. In particular, this PR aims to remove a parser that has been in use on tools and in production behind a feature flag for the past week without raising any red flags


drdelambre (approved):
GOOD NEWS!!!

PR #163, opened 2020-10-12
chore(tests): added some tests to the data transformation layer

Asalem1:
This PR adds a few tests to the parsing data transformation layer to account for wonky data formats


121watts (changes requested):
Can you reduce these CSVs to the smallest possible and useful strings?

Asalem1:
> Can you reduce these CSVs to the smallest possible and useful strings?

Totally

121watts (approved):

PR #164, opened 2020-10-12
test(e2e): fix intermittent test failure

121watts:
Closes influxdata/monitor-ci#98

### The Problem

Our e2e tests would fail intermittently.  Most of the failures produced a screenshot that resembled something like this:

Screen Shot 2020-10-12 at 2 56 24 PM

Just a blank black screen.  Cool...

#### Part 1
Turns out docker uses [`/dev/shm`](https://www.cyberciti.biz/tips/what-is-devshm-and-its-practical-usage.html) as a shared memory space to run programs.  The problem is it defaults to 64mb.  Which, for the memory hog that is Chrome, is just not enough for our use case ~1% of the time.  This was a bug in Chromium [identified three years ago](https://bugs.chromium.org/p/chromium/issues/detail?id=736452). 

Luckily for us, there is a simple fix for this `--disable-dev-shm-usage`.  This tells Chrome to write to `/tmp` instead of `/dev/shm`.  The downside is that writing to disk is "slower".  However, with these changes tests seem to run at similar if not faster speeds.  

#### Part 2
To save resources in CI I also reduced the amount of tests saved in memory to 1.  Cypress saves tests to memory to allow [time traveling" previously run tests](https://docs.cypress.io/guides/getting-started/writing-your-first-test.html#Time-travel).  We can't do this in CI so we might as well reduce it.   

#### Part 3
There were a couple of legitimate flaky tests that mostly all boiled down to this: 

```
CypressError: cy.click() failed because this element is detached from the DOM.

<element>

Cypress requires elements be attached in the DOM to interact with them.

The previous command that ran was:

  cy.get()

This DOM element likely became detached somewhere between the previous and current command.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You typically need to re-query for the element or add 'guards' which delay Cypress from running new commands.

https://on.cypress.io/element-has-detached-from-dom
```
There's an [open issue on Cypress' gituhub](https://github.com/cypress-io/cypress/issues/7306) to have retrying for detached elements be the default behavior of Cypress actions.  Until that happens I've introduced two workarounds.  The first is to [introduce a command that retries "clicks"](https://github.com/influxdata/ui/pull/164/files#diff-6c107e2390c5ad84e123f3be09d7b91ba35de8f8cc550d1f358c75806df47645R536).  The second is to introduce test retrying in circleci.  This way, if there's a 1/100 test run race condition the test will just re-run and likely pass.  This will save someone the time from having to run the entire suite over again.  

### Part 4
There were a few legitimate test failures that I either just deleted because they should not have existed to begin with or updated to more idiomatic Cypress.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,5 +1,5 @@
 {
-  "numTestsKeptInMemory": 25,
+  "numTestsKeptInMemory": 0,
   "integrationFolder": "cypress/e2e",
   "defaultCommandTimeout": 10000,
   "chromeWebSecurity": false,
--- a/cypress.json
+++ b/cypress.json
@@ -2,6 +2,7 @@
   "numTestsKeptInMemory": 1,
   "integrationFolder": "cypress/e2e",
   "defaultCommandTimeout": 10000,
+  "videoUploadOnPasses": false,
   "chromeWebSecurity": false,
   "env": {
     "username": "dev_user",
--- a/cypress.json
+++ b/cypress.json
@@ -1,5 +1,5 @@
 {
-  "numTestsKeptInMemory": 1,
+  "numTestsKeptInMemory": 25,
   "integrationFolder": "cypress/e2e",
   "defaultCommandTimeout": 10000,
   "videoUploadOnPasses": false,
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,8 @@
 {
+  "retries": {
+    "runMode": 2,
+    "openMode": 0
+  },
   "numTestsKeptInMemory": 25,
   "integrationFolder": "cypress/e2e",
   "defaultCommandTimeout": 10000,
@@ -9,9 +13,5 @@
     "password": "CHECKMEOUTASUPERSECRETPASSWORD",
     "org": "InfluxData",
     "bucket": "defbuck"
-  },
-  "retries": {
-    "runMode": 2,
-    "openMode": 0
   }
 }
--- a/cypress.json
+++ b/cypress.json
@@ -9,5 +9,9 @@
     "password": "CHECKMEOUTASUPERSECRETPASSWORD",
     "org": "InfluxData",
     "bucket": "defbuck"
+  },
+  "retries": {
+    "runMode": 2,
+    "openMode": 0
   }
 }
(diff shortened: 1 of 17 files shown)


cypress.json
```diff
@@ -2,11 +2,16 @@
   "numTestsKeptInMemory": 25,
   "integrationFolder": "cypress/e2e",
   "defaultCommandTimeout": 10000,
+  "videoUploadOnPasses": false,
```
drdelambre, line 5: :raised_hands: no one cares about these

cypress/e2e/explorer.test.ts
```diff
@@ -28,7 +28,7 @@ function getTimeMachineText() {
 }
 
 // NOTE: this crashes circle FOR NO REASON
```
drdelambre, line 30: lol.. like that you're keeping the comment. JUST IN CASE

cypress/plugins/index.js
```diff
@@ -1 +1,9 @@
-module.exports = on => {}
+module.exports = (on, config) => {
+  on('before:browser:launch', (browser = {}, launchOptions) => {
+    if (browser.family === 'chromium' && browser.name !== 'electron') {
+      launchOptions.args.push('--disable-dev-shm-usage')
```
drdelambre, line 4: here's the magic

cypress/support/commands.ts
```diff
@@ -530,6 +530,24 @@ export const createEndpoint = (
   return cy.request('POST', 'api/v2/notificationEndpoints', endpoint)
 }
 
+// helpers
+// Re-query elements that are found 'detached' from the DOM
+// https://github.com/cypress-io/cypress/issues/7306
+export const clickAttached = (subject?: JQuery<HTMLElement>) => {
```
drdelambre, line 536: and some more magic. is there a forum where we can tell everyone "if xyz starts breaking, use clickAttached instead"?

drdelambre (approved):
good work!

Asalem1:
🔥

cypress/support/commands.ts
121watts: specifically, if xyz starts breaking _because_ its been detached from the DOM for sure.  I'm hosting a lunch and learn to disseminate all this info.  Also, this PR can act as an artifact on what changed and why.

PR #167, opened 2020-10-14
fix: pass the correct env vars to the deploy script

drdelambre:
watch me learn bash and circle. live. on master.

when i ssh'd into the deploy step and ran the new command as outlined in the script changes, it pushed the correct image all the way to argo, so i think it's in a good spot


Asalem1 (approved):
🧍🏽‍♂️

PR #171, opened 2020-10-14
feat: keep the ui image around for local development

drdelambre:
we're using an intermediary image for testing, or whenever you use `make update` in local development. currently this image is generated on all pushes to master and built over on jenkins.

but the theme of the day is "DOWN WITH JENKINS"

so... this builds that image on circle instead


121watts (approved):

121watts:
please rebase with master so this is green before merge

121watts (approved):

PR #180, opened 2020-10-16
fix(parsing-NaN): fixed NaN time conversion bug

Asalem1:
Closes https://github.com/influxdata/idpe/issues/8927

### Problem

NaN is still a `typeof === number` so parsing a NaN was causing the parser to crash.

### Solution

Make sure that the value is also not `NaN`

### DIFF
--- a/src/shared/parsing/flux/response.test.ts
+++ b/src/shared/parsing/flux/response.test.ts
@@ -114,7 +114,7 @@ describe('parseResponseWithFromFlux', () => {
 ,,0,gauge,kube_pod_container_resource_requests_cpu_cores,2020-10-16T15:41:21.798083141Z,2020-10-16T16:41:21.798083141Z,,0.1
 `.trim()
 
-    const expectedData = [
+    const expected = [
       [
         '',
         'result',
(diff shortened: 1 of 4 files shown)


src/shared/parsing/flux/response.test.ts
```diff
@@ -104,6 +104,60 @@ describe('parseResponseWithFromFlux', () => {
     const result = parseResponseWithFromFlux(MULTI_SCHEMA_RESPONSE)
     expect(result).toHaveLength(4)
   })
+  test('should not convert NaN timestamps', () => {
```
drdelambre, line 107: i want you to know how happy this makes me

src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -105,10 +105,13 @@ export const fromFluxTableTransformer = (
       rowType = table.getColumnType(column)
       originalType = table.getOriginalColumnType(column)
       columnData = table.getColumn(column, rowType)[i]
+      console.log('columnDataOutside: ', columnData)
```
drdelambre, line 108: WHAT

Asalem1: 😢

drdelambre (approved):

PR #187, opened 2020-10-19
feat(export-to): allow queries to be exported as tasks from flows

Asalem1:
Closes https://github.com/influxdata/influxdb/issues/19603

This PR adds functionality to the `Output to Bucket` flows pipe option by allowing users to create queries and export them as tasks.

export-to

### DIFF
--- a/src/flows/components/ExportTaskOverlay/ExportTaskOverlay.tsx
+++ b/src/flows/components/ExportTaskOverlay/ExportTaskOverlay.tsx
@@ -1,0 +1,192 @@
+// Libraries
+import React, {ChangeEvent, FC, useEffect, useState} from 'react'
+import {useDispatch, useSelector} from 'react-redux'
+import {useHistory} from 'react-router-dom'
+
+// Components
+import {
+  Form,
+  InputType,
+  Input,
+  Button,
+  Grid,
+  JustifyContent,
+  AlignItems,
+  Alignment,
+  Columns,
+  ComponentSize,
+  ButtonType,
+  ComponentColor,
+  ComponentStatus,
+  Overlay,
+  Panel,
+  Gradients,
+  Tabs,
+  Orientation,
+  IconFont,
+} from '@influxdata/clockface'
+
+// Utils
+import {saveNewScript} from 'src/tasks/actions/thunks'
+import {getOrg} from 'src/organizations/selectors'
+import {getTasks} from 'src/tasks/actions/thunks'
+import {getAllTasks as getAllTasksSelector} from 'src/resources/selectors'
+
+enum ExportAsTask {
+  Create = 'create',
+  Update = 'update',
+}
+
+const ExportTaskOverlay: FC = () => {
+  const [activeTab, setActiveTab] = useState(ExportAsTask.Create)
+  const [taskName, setTaskName] = useState('')
+  const [interval, setInterval] = useState('')
+  const history = useHistory()
+  const tasks = useSelector(getAllTasksSelector)
+  const org = useSelector(getOrg)
+
+  const dispatch = useDispatch()
+  useEffect(() => {
+    dispatch(getTasks())
+  }, [dispatch])
+
+  const onDismiss = () => {
+    history.goBack()
+  }
+
+  const onSubmit = () => {
+    const task: string = `option task = { \n  name: "${taskName}",\n   every: ${interval},\n offset: 0s\n`
+    const trimmedOrgName = org.name.trim()
+    const trimmedBucketName = org.name.trim()
+    const script: string = `${task}\n  |> to(bucket: "${trimmedBucketName}", org: "${trimmedOrgName}")`
+    const preamble = `${task}`
+
+    dispatch(saveNewScript(script, preamble))
+  }
+  const canSubmit = true
+
+  return (
+    <Overlay visible={true}>
+      <Overlay.Container maxWidth={600}>
+        <Overlay.Header
+          title="Export As Task"
+          onDismiss={onDismiss}
+          testID="export-as-overlay--header"
+        />
+        <Overlay.Body>
+          <Tabs.Container orientation={Orientation.Horizontal}>
+            <Tabs alignment={Alignment.Left} size={ComponentSize.Small}>
+              <Tabs.Tab
+                id={ExportAsTask.Create}
+                text="Create New"
+                testID="task--radio-button"
+                onClick={() => setActiveTab(ExportAsTask.Create)}
+                active={activeTab === ExportAsTask.Create}
+              />
+              <Tabs.Tab
+                id={ExportAsTask.Update}
+                text="Update Existing"
+                testID="variable-radio-button"
+                onClick={() => setActiveTab(ExportAsTask.Update)}
+                active={activeTab === ExportAsTask.Update}
+              />
+            </Tabs>
+            <Tabs.TabContents>
+              <Form>
+                <Grid>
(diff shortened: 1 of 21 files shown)


src/flows/components/Flow.tsx
```diff
@@ -32,6 +34,13 @@ const FlowContainer: FC = () => {
       <ResultsProvider>
         <RefProvider>
           <ScrollProvider>
+            <Switch>
+              {/* wrapping in a switch in anticipation of export to dashboard */}
```
drdelambre, line 38: :+1:

src/flows/components/panel/FlowPanel.tsx
```diff
@@ -36,14 +37,16 @@ import {RefContext} from 'src/flows/context/refs'
 
 export interface Props extends PipeContextProps {
   id: string
+  shouldShow?: boolean
 }
 
 export interface HeaderProps {
   id: string
   controls?: ReactNode
+  shouldShow?: boolean
```
drdelambre, line 46: this should be a ReactNode, not a boolean, so that we can pass arbitrary buttons that are shown all the time, and not "show the control buttons all the time"

src/flows/components/ExportTaskOverlay/ExportTaskOverlay.tsx
```diff
@@ -0,0 +1,344 @@
+// Libraries
+import React, {ChangeEvent, FC, useEffect, useState} from 'react'
```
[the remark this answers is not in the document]
Asalem1: Totally hear you on that. I felt the same way. My concern was updating the parent state from the child felt wrong, but if we were to set the state directly within the child we would've lost context (and since this exists outside of `context` I wasn't sure how to get the two to work together)

src/flows/components/ExportTaskOverlay/ExportTaskOverlay.tsx
```diff
@@ -0,0 +1,344 @@
...
+                <Grid>
+                  <Grid.Row>
+                    <ExportTaskBody
+                      setEveryInterval={handleSetEveryInterval}
+                      activeTab={activeTab}
+                      interval={interval}
+                      setTaskName={handleSetTaskName}
+                      taskName={taskName}
+                      formattedQueryText={formattedQueryText}
+                      tasks={tasks}
```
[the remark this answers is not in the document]
Asalem1: I know right. This was definitely my biggest issue with all of this

src/flows/components/panel/FlowPanel.tsx
Asalem1: Gotcha.

src/flows/components/panel/FlowPanel.tsx
```diff
@@ -100,6 +102,7 @@ const FlowPanelHeader: FC<HeaderProps> = ({id, controls}) => {
               active={canBeMovedDown}
             />
           </FeatureFlag>
+          {shouldShow && <ExportTaskButton />}
```
drdelambre, line 105: this should turn into whatever that shouldShow turns into. Follow the `controls` pattern here, just that it's not hidden. And for next time, if you find yourself making panel specific changes is the flows pattern, and not just on the pipe, that should shoot off some red flags

src/flows/context/pipe.tsx
```diff
@@ -35,6 +38,12 @@ interface PipeContextProps {
 export const PipeProvider: FC<PipeContextProps> = ({id, children}) => {
   const {flow} = useContext(FlowContext)
   const results = useContext(ResultsContext)
+  const {generateMap} = useContext(QueryContext)
+
+  const stages = useMemo(() => generateMap(), [generateMap, flow.data.get(id)])
+  const queryText =
+    stages.filter(stage => stage.instances.includes(id))[0]?.text || ''
```
drdelambre, line 45: lol i was about to yell about this being illegible, then remembered i told you to write it. so it looks perfect now

src/flows/pipes/Data/index.ts
```diff
@@ -51,8 +51,7 @@ export default register => {
       }
 
       if (aggregateFunction?.name) {
-        text += `  |> aggregateWindow(every: v.windowPeriod, fn: ${aggregateFunction.name}, createEmpty: false)
-            |> yield(name: "${aggregateFunction.name}")`
+        text += `  |> aggregateWindow(every: v.windowPeriod, fn: ${aggregateFunction.name}, createEmpty: false)|> yield(name: "${aggregateFunction.name}")`
```
drdelambre, line 54: how do we get this out of here and only get included at the end?

src/tasks/actions/thunks.ts
```diff
@@ -193,6 +193,39 @@ export const updateTaskName = (name: string, taskID: string) => async (
...
+}
+
+export const updateTask = ({
+  script,
+  interval,
+  preamble,
+  task,
+}: UpdateTaskParams) => async (dispatch: Dispatch<Action>) => {
+  try {
+    const fluxScript = await insertPreambleInScript(script, preamble)
```
drdelambre, line 210: did you get clarification from russ on the concept of system variables in task queries? they still dont make any sense to me

Asalem1: I didn't, let me run that by him right now.

src/flows/components/ExportTaskOverlay/QueryTextPreview.tsx
```diff
@@ -0,0 +1,39 @@
+import React, {FC} from 'react'
+import {Form} from '@influxdata/clockface'
+import MonacoEditor from 'react-monaco-editor'
+import THEME_NAME from 'src/external/monaco.toml.theme'
+import TOMLLANGID from 'src/external/monaco.toml.syntax'
```
drdelambre, line 5: oh.. this isn't what i meant when i linked to the toml editor. Add a readonly prop to the flux editor component so we can use that

src/flows/components/ExportTaskOverlay/UpdateTaskBody.tsx
```diff
@@ -0,0 +1,71 @@
...
+import WarningPanel from 'src/flows/components/ExportTaskOverlay/WarningPanel'
+import QueryTextPreview from 'src/flows/components/ExportTaskOverlay/QueryTextPreview'
+import {OverlayContext} from 'src/flows/context/overlay'
+
+type Props = {
+  formattedQueryText: string
+}
+
+const UpdateTaskBody: FC<Props> = ({formattedQueryText}) => {
+  const {interval, handleSetEveryInterval, hasError, tasks} = useContext(
```
drdelambre, line 22: instead of pulling tasks in like this from the context, as we're just passing it around in the backend, just to check it's length later, use a redux `useSelector` in here instead that returns a boolean of hasTasks

src/flows/context/overlay.tsx
```diff
@@ -0,0 +1,122 @@
...
+  handleSetActiveTab: (tab: ExportAsTask) => void
+  handleSetError: (value: boolean) => void
+  handleSetEveryInterval: (value: string) => void
+  handleSetTask: (task: Task) => void
+  handleSetTaskName: (value: string) => void
+  hasError: boolean
+  interval: string
+  selectedTask: Task | undefined
+  taskName: string
+  tasks: Task[]
```
drdelambre, line 24: all of this data contract is good, except this one `task` property.

drdelambre (approved):
:fire:

PR #191, opened 2020-10-19
fix(de): View Type Dropdown will use the defined types

Palakp41:
- Moved the visualization types and loading of those files from flows to shared folder
- Updated the View type dropdown in data explorer to use the defined types in shared folder


src/timeMachine/components/visualization/ViewTypeDropdown.tsx
```diff
@@ -0,0 +1,66 @@
+import React, {FC} from 'react'
+import {TYPE_DEFINITIONS} from 'src/shared/visualization'
+
+// Components
+import {Dropdown, DropdownMenuTheme} from '@influxdata/clockface'
+import {ComponentStatus} from 'src/clockface'
+
+// Utils
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
```
drdelambre, line 9: why is this file here instead of using the `src/shared/visualization` version?

src/timeMachine/components/visualization/view.tsx
```diff
@@ -0,0 +1,38 @@
+// Libraries
+import React, {FC, useCallback} from 'react'
+import {useDispatch, useSelector} from 'react-redux'
+
+// Components
+import ViewTypeDropdown from 'src/timeMachine/components/visualization/ViewTypeDropdown'
```
drdelambre, line 6: import this component from `src/shared/visualization` instead

src/timeMachine/components/visualization/ViewTypeDropdown.tsx
Palakp41: Good catch, thanks!

drdelambre (approved):
:crossed_fingers: i think it looks good

PR #193, opened 2020-10-20
chore: remove legacy context menu

alexpaxton:
Taking steps towards eliminating the legacy `src/clockface` directory

- Always show context menus on resource cards
- Refactor every component that uses `<Context>` to use components from the Clockface package instead
- Update affected e2e tests
- Delete `Context` component and associated styles from `src/clockface`

Screen Shot 2020-11-03 at 3 42 43 PM
Screen Shot 2020-11-03 at 3 42 31 PM
Screen Shot 2020-11-03 at 3 42 19 PM
Screen Shot 2020-11-03 at 3 42 10 PM
Screen Shot 2020-11-03 at 3 42 00 PM


cypress/e2e/tasks.test.ts
```diff
@@ -242,30 +242,22 @@ http.post(
 
     it('can delete a task', () => {
       cy.getByTestID('task-card')
-        .first()
-        .trigger('mouseover')
+        // .first()
```
alexpaxton, line 245: I don't entirely understand why removing `.first()` makes this work except for there being only 1 card and it's not an array, so trying to grab the first item doesn't actually make sense

alexpaxton:
@hoorayimhelping this is actually ready for review now, was fixing some e2e tests

src/buckets/components/BucketContextMenu.tsx
```diff
@@ -23,42 +19,24 @@ interface Props {
...
-            label="Confirm"
-            action={onDeleteBucket}
-            value={bucket}
-            testID={`context-delete-bucket ${bucket.name}`}
-          />
-        </Context.Menu>
-      )
-    } else {
+
+    if (bucket.type !== 'user') {
```
hoorayimhelping, line 24: ❤️  early return >>>>>>>>>>> `if/else`

src/authorizations/components/TokenRow.tsx
```diff
@@ -42,6 +41,7 @@ class TokenRow extends PureComponent<Props> {
     return (
       <ResourceCard
         contextMenu={this.contextMenu}
+        contextMenuInteraction="alwaysVisible"
```
hoorayimhelping, line 44: i feel like this string `"alwaysVisible"` should be made into an enum so we get some type safety and consistency. what do you think?

src/notifications/endpoints/components/EndpointCardMenu.tsx
```diff
@@ -1,38 +1,84 @@
...
+              size={ComponentSize.ExtraSmall}
+              testID="context-clone-endpoint"
+              onClick={handleCloneAndHide(onHide)}
+            >
+              <List.Icon glyph={IconFont.Duplicate} />
+              Clone
+            </List.Item>
+          </List>
+        )}
+      />
```
hoorayimhelping, line 70: this `View History` and `Clone` behavior seems similar to the other card. is there enough repeated commonality to consider this a pattern?

src/scrapers/components/ScraperRow.tsx
```diff
@@ -2,13 +2,19 @@
...
+  ComponentColor,
+  ButtonShape,
+  ComponentSize,
+} from '@influxdata/clockface'
 import {Scraper} from 'src/types'
 
 // Constants
 import {DEFAULT_SCRAPER_NAME} from 'src/dashboards/constants'
-import {IconFont, ComponentColor} from '@influxdata/clockface'
+import {} from '@influxdata/clockface'
```
hoorayimhelping, line 17: ✂️  the whole line

hoorayimhelping (commented):
A couple of comments. not blocking, but I'd like to hear your thoughts on them before approving

src/authorizations/components/TokenRow.tsx
alexpaxton: It's typed as allowing specific strings to be passed in, which is somewhere between an enum and any string:
https://github.com/influxdata/clockface/blob/master/src/Components/ResourceList/Card/ResourceCard.tsx#L20-L21

Changing it seems out of scope for this PR, but def worth considering

src/notifications/endpoints/components/EndpointCardMenu.tsx
alexpaxton: Yeah it's def a pattern. Not all resources (currently) have a clone button, but it's pretty common. I'd love to standardize it in the code so it's not up to each resource list to remember including it

hoorayimhelping (approved):
not anywhere near the worst thing imaginable

genehynson:
@hoorayimhelping This seems like an awesome PR...do we want to revive it? Any idea where Alex left off?

hoorayimhelping:
@genehynson i think it was pretty much good to go? seemed like it was ready

genehynson:
So I've tried to rebase and merge this PR to get it updated and it's not going super well. I pushed the merge commit to this PR but I'm not confident that all the necessary changes are in here. The [rebase](https://github.com/influxdata/ui/compare/genehynson/rebase-remove-legacy-context-menu?expand=1) has more changes to tests I think 🤔  I did a [compare](https://github.com/influxdata/ui/compare/chore/remove-context-menu...genehynson/rebase-remove-legacy-context-menu?expand=1) between the rebase and merge and the differences are all over the place, not helping build my confidence. And both methods are failing for typescript issues _le sigh_

genehynson:
Sadly this is too old to recover. Closing

PR #195, opened 2020-10-20
chore: update github metadata stuff

hoorayimhelping:
#### Semantic Commit Changes:
- allows revert commits (`"Revert "feat: ride unicorns""`)
- allows any commit to have a semantic commit
  - the idea behind this is if you have three or four commits on a branch, and those commits are going to be squash merged into a single semantic commit, it's not really worth making someone force pushing commit changes to be semantic if they're going to be squashed anyway. it's causing more frustration than it's solving

- adds a github pull request template

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+Closes #
+
+### Testing Steps
+
+### Merge Checklist
+- [ ] User facing change?
+  - [ ] CHANGELOG.md is updated and links to this PR
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,3 +1,4 @@
 # docs: https://github.com/probot/semantic-pull-requests#configuration
-titleAndCommits: true
 allowMergeCommits: true
+allowRevertCommits: true
+titleOnly: true


drdelambre:
it does suck to have to rename a commit. is there a way to say "i dont care about any commit you made, but that squash commit better be perfect"?

glinton:
seemed like there was a 'title only' option https://github.com/influxdata/idpe/pull/5641/files. great for squashing `titleOnly: true`

hoorayimhelping:
>is there a way to say "i dont care about any commit you made, but that squash commit better be perfect"?

unfortunately not, which is kind of dumb, ya know? didn't notice any issues about it either

glinton:
since squashing generally uses the pr title, `titleOnly: true` will do what you seek

hoorayimhelping:
@glinton I can't seem to request your review but I updated this to be `titleOnly`

glinton (approved):
Nice! No problem in leaving them, but with `titleOnly`, merge/revert commits are automatically allowed as well because all commits are ignored.

drdelambre (approved):

genehynson:
@hoorayimhelping good to merge I think?

PR #196, opened 2020-10-20
fix: nightly e2e

121watts:
Looks like yet _another_ `dev/shm` issue with Firefox.  This time had to adjust `monitor-ci` and increase the `dev/shm` size on the `cypress` image.


drdelambre (approved):
great to see everything passes on firefox as well

PR #199, opened 2020-10-21
chore(remove-flag): removing cancelQueryUiExpansion flag after being on for two weeks

Asalem1:
The feature flag for cancelQueryUiExpansion has been enabled on production and tools for two weeks without any complaints. This is just a clean-up to remove it from the UI. Once the UI has been updated, I'll remove the flag from IDPE, followed by config cat


drdelambre (approved):
congrats!

PR #200, opened 2020-10-21
refactor: improve notification buttons

### LINKED ISSUE #183: Enable notifications to optionally include buttons
- Notifications should allow for a button to be displayed
- Button text should be configurable
- Button color should be configurable
- Button destination should be configurable
- Button should inherit size from notification dialog (parent)

alexpaxton:
closes #183 

### Changes

- `Notification` has an optional `buttonElement` prop instead of `link` and `linkText`
  - Since this is an element it is slightly more flexible and composable
  - `buttonElement` gets access to the specific dismiss function of its parent notification via render prop
- Refactored `Notifications` component to no longer care about aggregate functions
- Introduced a central place to store notification button components
  - Don't love it, but it does work
- Refactored notifications that previously used `link` and `linkText` to adopt new pattern

### Testing

We manually tested the aggregate type notification by hand and verified it works properly:

Screen Shot 2020-10-21 at 2 42 53 PM

You can reproduce by submitting a query that tries to aggregate (mean) to a string


alexpaxton:
@Palakp41 just fixed the e2e tests in here

Palakp41 (approved):
lgtm

PR #202, opened 2020-10-22
feat: make telegraf configs editable

alexpaxton:
Closes https://github.com/influxdata/influxdb/issues/19772

#### Changes

- Create a wrapper around the overlay that uses `GetResources` in case the page loads with the overlay visible and telegrafs have not been fetched yet
- Store the "draft" in temporary state until user cancels or saves
- Ask for confirmation before saving
- Wrapped changes in flag `editTelegrafs` which is `false` by default

Screen Shot 2020-10-22 at 2 37 31 PM


src/shared/selectors/flags.ts
```diff
@@ -24,6 +23,7 @@ export const OSS_FLAGS = {
   'load-data-flux-sources': false,
   'load-data-integrations': false,
   'notification-endpoint-telegram': false,
+  editTelegrafs: false,
```
Asalem1, line 26: we should aim to set these in IDPE and sync it with configcat

src/telegrafs/components/TelegrafConfigOverlayForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+}) => {
+  const match: Match = useRouteMatch({
+    path: '/orgs/:orgID/load-data/telegrafs/:id/view',
+    exact: true,
+    strict: false,
+  })
+
+  let telegraf
+
+  if (match && match.params && match.params.id) {
```
Asalem1, line 59:
```suggestion
  if (match?.params?.id) {
```

src/telegrafs/components/TelegrafConfigOverlay.tsx
```diff
@@ -1,109 +1,55 @@
...
-type Props = OwnProps & StateProps
+type ReduxProps = ConnectedProps<typeof connector>
+type Props = OwnProps & ReduxProps
 
-@ErrorHandling
-class TelegrafConfigOverlay extends PureComponent<Props> {
-  public render() {
-    return <>{this.overlay}</>
+const TelegrafConfigOverlay: FC<Props> = ({onUpdateTelegraf, onClose}) => {
+  const handleUpdateTelegraf = (telegraf: Telegraf): void => {
```
Asalem1, line 26: Since we're just updating the telegraf and closing the overlay (and we have access to the `onClose` function in the `TelegrafConfigOverlayForm`, let's move this into that component to localize the logic. That way we can remove redux from this component altogether

src/telegrafs/components/TelegrafConfigOverlayForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+const mstp = (state: AppState) => {
+  return {
+    telegrafs: getAll<Telegraf>(state, ResourceType.Telegrafs),
+  }
+}
+
+const connector = connect(mstp)
+
+export default connector(withRouter(TelegrafConfigOverlayForm))
```
Asalem1, line 138: Nit: seems like a lot of work to set this FC up when we could use hooks and make this component a bit more contained and isolated.

Nit2: Maybe instead of drilling `onClose` all the way from the `OverlayController` we can use a hook to instantiate it here?

Asalem1 (commented):
Love the + to - ratio on this one. Would love to see some things moved around before this gets merged in

alexpaxton:
Feature flags business: https://github.com/influxdata/idpe/pull/8992

src/shared/selectors/flags.ts
alexpaxton: https://github.com/influxdata/idpe/pull/8992

Asalem1: Awesome, these can be removed since they now exist on IDPE

Asalem1 (approved):
Just need to remove the feature flag from being set in the UI and it's all good

PR #210, opened 2020-10-26
fix(flows-ui-pagination): fixed the UI pagination for flows and feature flagged it

### LINKED ISSUE #201: Metric Selector Raw Data Table isn't working
### Problem

Not entirely sure what's happening here but it could be related to the changes made in the following PR:

https://github.com/influxdata/ui/pull/168/files

Here's how the bug looks:

### Steps to Reproduce

1. Create a flow with a metric selector
2. Generate & run a query with the metric selector
3. Try and navigate the paginated raw data results

### Expected Behavior

The results of the metric selector should be paginated, and the number of rows that are being paginated through should not be updated with every click of the next button

Asalem1:
Closes #201 

### Problem

The way that the RawDataTable is currently built out, the view layer can't parse partial CSVs and just renders blank data

### Solution

Update RawDataTable to accept parsed CSVs rather than raw CSVs, and render the data with an optional offset that is applied for UI pagination. Feature flagged this here:

https://github.com/influxdata/idpe/pull/9011

UI-pagination

### DIFF
--- a/src/timeMachine/components/RawFluxDataTable.tsx
+++ b/src/timeMachine/components/RawFluxDataTable.tsx
@@ -2,22 +2,21 @@
 import React, {PureComponent} from 'react'
 import memoizeOne from 'memoize-one'
 import RawFluxDataGrid from 'src/timeMachine/components/RawFluxDataGrid'
+import {FromFluxResult} from '@influxdata/giraffe'
 
 // Utils
 import {
   parseFiles,
   parseFilesWithObjects,
   parseFilesWithFromFlux,
+  parseFromFluxResults,
 } from 'src/timeMachine/utils/rawFluxDataTable'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {DapperScrollbars, FusionScrollEvent} from '@influxdata/clockface'
 
-// TODO(ariel): remove files and make data and maxColumnCount mandatory when
-// turning the flowsUiPagination feature flag off
 interface Props {
   files?: string[]
-  data?: string[][]
-  maxColumnCount?: number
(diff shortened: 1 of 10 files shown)


src/timeMachine/components/RawFluxDataTable.tsx
```diff
@@ -27,25 +32,67 @@ interface State {
...
+      startRow,
+    } = this.props
+    let {data, maxColumnCount} = this.props
 
     const {scrollTop, scrollLeft} = this.state
+
+    const tableWidth = width
+    const tableHeight = height
+
+    if (isFlagEnabled('flowsUiPagination')) {
```
drdelambre, line 55: i can't put my finger on it, but triggering this off of a flag instead of by props feels wrong. what about those other flags below? why aren't they included in this solution?

src/timeMachine/components/Vis.tsx
```diff
@@ -83,16 +90,41 @@ const TimeMachineVis: SFC<Props> = ({
         >
           {isViewingRawData ? (
             <AutoSizer>
-              {({width, height}) =>
-                width &&
-                height && (
+              {({width, height}) => {
+                if (isFlagEnabled('flowsUiPagination')) {
```
drdelambre, line 94: maybe i'm kind of getting it. the flag has nothing to do with the feature, but instead the "is data csv or fromfluxresult" and i'd argue that the migration strategy in giraffe mandates that we allow for both, instead of overloading the data prop with some modality

so add a new prop called something else? parsedData?

drdelambre (approved):
this makes more sense to me. lemme go hunt down your idpe one too

PR #213, opened 2020-10-26
feat: simple query interface

drdelambre:
removes the state requirements from the existing query cache that ari made, and takes a hard line on parsing the results right off the bat. _should_ provide for a cleaner separation of concerns, to allow us to generalize the solution, and roll out the query caching functionality to all requests made in the system.

don't know if there's value in exposing the response as an event architecture so that we can visualize results as they're loaded from the request, or if waiting for the promise to resolve is fine.... i think we should go with the later as it's less work, and the changes to the view layer that we would be building to capitalize on that event streaming interface would be wiped out and replaced once we have apache arrow support for our data views. they're opposite architectures from one another.


Asalem1 (approved):

genehynson:
@drdelambre do you plan to revive this at some point?

drdelambre:
i've moved on from this. there's a better way, let usecontext set us free

PR #214, opened 2020-10-26
feat(annotations): annotations control bar

### LINKED ISSUE #197: Annotations: As a user I can access the annotations control bar
As a user I want to access controls and tools relating to annotations from a dashboard

#### Toggle Controls ON

GIVEN I am viewing a dashboard
WHEN I click the [annotations button](https://influxdata.github.io/clockface/?path=/story/components-buttons-composed--standardbutton) in the page header
THEN I see a set of controls appear below

[Design Mockup](https://www.figma.com/file/AQk0tUthLFepRhFUibvzMI/Annotations?node-id=131%3A452)

#### Toggle Controls OFF

WHEN I click the annotations button again
THEN I see the controls disappear

[Design Mockup](https://www.figma.com/file/AQk0tUthLFepRhFUibvzMI/Annotations?node-id=131%3A64)

#### Navigate to Settings

GIVEN I am viewing a dashboard
AND the annotations controls are visible
THEN I see a [settings button](https://influxdata.github.io/clockface/?path=/story/components-buttons-composed--squarebutton) on the rightmost side of the controls

WHEN I click the settings button
THEN I am taken to the Settings page with the Annotations tab selected

[Design Mockup](https://www.figma.com/file/AQk0tUthLFepRhFUibvzMI/Annotations?node-id=131%3A502)

---

#### Clockface Components

[Button with Icon](https://influxdata.github.io/clockface/?path=/story/components-buttons-composed--standardbutton)
[SquareButton](https://influxdata.github.io/clockface/?path=/story/components-buttons-composed--squarebutton)

#### UI Components / Technical Notes

- Annotations control bar should re-use styles and markup from the [VariablesControlBar](https://github.com/influxdata/ui/tree/master/src/dashboards/components/variablesControlBar)
  - Might want to abstract out and make a generic "Control Bar" component for use in both
- Annotations control bar should have its state stored in local storage the same way as the VariablesControlBar, so that refreshing the page does not affect it being open or not and it is treated (roughly) like a user preference

#### Design Notes

- Control bar does not have to fade in/out like in the mockup

alexpaxton:
Closes #197 

#### Toolbars

- Introduce `Toolbar.tsx` as a common component for variables and annotations control bars

### Annotations Control Bar

Screen Shot 2020-10-29 at 2 29 52 PM
Screen Shot 2020-10-29 at 2 30 11 PM

- Add state in user settings redux for visibility of annotations control bar
- Display annotations controls in Dashboards & Data Explorer
- User can see which streams are currently enabled
- User can search through, enable, and disable annotation streams
- User can enter annotation mode via the `Annotate Button`

### Add/Edit Annotation Overlays

Screen Shot 2020-10-29 at 2 30 43 PM

- Accessible via `/add-annotation` or `/edit-annotation` from dashboards and the Data Explorer
- `Add Annotation` uses url search params to pass `timeStart` and `timeStart` from anywhere to the overlay
  - For example `/add-annotation?timeStart=oct29&timeStop=oct30` will show the overlay prepopulated with those times


src/annotations/components/AnnotationsToggleButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+// Actions
+import {toggleShowAnnotationsControls} from 'src/userSettings/actions'
+
+// Components
+import {Button, ComponentColor, IconFont} from '@influxdata/clockface'
+
+const AnnotationsToggleButton: FC = () => {
+  const dispatch = useDispatch()
+  const isVisible = useSelector(getAnnotationControlsVisibility)
```
drdelambre, line 16: i saw that you're following the variable example, but does it make sense for the visibility of this component to exists between dashboards? same applies to variables as well, just wondering how to scope this state.

src/annotations/components/EditAnnotationOverlay.tsx
```diff
@@ -0,0 +1,39 @@
...
+  streamID: 'anno2',
+  bucketName: 'defBuck',
+  measurement: 'annotations',
+  tags: {
+    _field: 'snacks',
+  },
+}
+
+const EditAnnotationOverlay: FC = () => {
+  // const handleSubmit = (annotation: Annotation): void => {
```
drdelambre, line 25: dont keep around commented out code. you can get around the linting error by prepending an underscore to annotation

src/annotations/components/AnnotationsToggleButton.tsx
alexpaxton: The UX I am going for is: User turns on a stream of annotations, they are overlaid on all the graphs regardless of where the user is. So it's a global thing

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+} from 'src/annotations/reducers/annotationReducer'
+
+// Types
+import {AnnotationStream} from 'src/annotations/constants/mocks'
+
+// Contexts
+import {OverlayContext} from 'src/overlays/components/OverlayController'
+
+interface Props {
+  title: 'Edit' | 'Add'
```
drdelambre, line 40: bit of a red flag here that the typing might be wrong. why isn't the title derived from a boolean or enum?

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+    const formIsValid = annotationFormIsValid(state, dispatch)
+
+    if (formIsValid) {
+      onSubmit(getAnnotationFromDraft(state))
+      onClose()
+    }
+  }
+
+  const handleSummaryChange = (e: ChangeEvent<InputRef>): void => {
+    dispatch({type: 'updateSummary', payload: e.target.value})
```
drdelambre, line 88: these actions should follow the convention used in the rest of the app. but before you do that, why have individual actions for every property on the store and not have a single action that takes a Partial<StateShape> ?

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+              <Grid.Column
+                widthXS={state.type === 'range' ? Columns.Six : Columns.Twelve}
+              >
+                <AnnotationTimeStartInput
+                  {...state.timeStart}
+                  type={state.type}
+                  onChange={handleTimeStartChange}
+                />
+              </Grid.Column>
+              {state.type === 'range' && (
```
drdelambre, line 141: my inner watts is yelling about this not being included in a component that self guards

drdelambre: and it is strange that it's not following the convention of the AnnotationTimeStartInput above?

src/annotations/components/annotationForm/AnnotationTimeStopInput.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+const AnnotationTimeStopInput: FC<Props> = ({
+  value,
+  status,
+  error,
+  onChange,
+}) => {
+  return (
+    <Form.Element label="Stop Time" required={true} errorMessage={error}>
```
drdelambre, line 21: why is this a different component?

src/annotations/selectors/index.ts
```diff
@@ -0,0 +1,35 @@
...
+export const getAnnotationControlsVisibility = (state: AppState): boolean => {
+  return state.userSettings.showAnnotationsControls || false
+}
+
+export const getVisibleAnnotationStreams = (
+  state: AppState
+): AnnotationStream[] => {
+  const visibleStreams = state.annotations.visibleStreamsByID
+  const filtered = MOCK_ANNOTATION_STREAMS.filter(stream =>
+    visibleStreams.includes(stream.id)
```
drdelambre, line 20: this feels expensive computationally (for every item in one array, go though all the items in another array). we should add visibility as client side meta data on the stream state. it might not come in from the api layer, but when we normalize the data, we append that default visibility state to it

alexpaxton: oh that sounds rad. Probably cleaner than having a whole different area in redux for it

drdelambre (approved):
thanks for proving out the reducer interface, it looks super slick!
I'd like to decouple from the route based triggering entirely, as it seems to make the app rigid in weird ways for a popup, but given that the pattern is in use in other places around the app, that's a bigger discussion outside of this PR

PR #219, opened 2020-10-27
test: new dashboards tests

Cubone21:
- new test for dashboard cell time format saving
- new test for sorting values in a dashboard cell


drdelambre:
@Cubone21 we're working on figuring out how to merge this stuff after moving our deployments to circle. They currently don't allow for us to take in PRs from forked repos, so we're all talking about what that means

Issue #192, opened 2020-10-20
Scatterplot's x axis is configurable


Palakp41:
When rendering the scatterplot in a dashboard, if a customized time range is selected in the ui, it will handle what properties to send to Giraffe. Giraffe will be able to control and decide what the auto domain for x axis should be.

Functionality in UI should be:
- if a user sets the domain by hand, use that.
- if they select _time for the x axis, set the default domain to the page's time window (timemachine / dashboard / whateves).
- anything else, use the auto domain. and make sure it works with the example samantha put at the bottom of that ticket (https://github.com/influxdata/influxdb/issues/18981)

Steps to fix:
1.) Update the giraffe such that the current functionality for controlling the x axis in UI is calculated in scatter plot in giraffe
2.) Update ui in a way where we we send customize properties for the x axis to giraffe's scatter plot.

121watts:
@Palakp41 referenced in other issue.  closing

PR #229, opened 2020-10-28
fix: use time range in Scatterplot

zoesteinkamp:
closes issue #192

### DIFF
--- a/src/shared/components/ScatterPlot.tsx
+++ b/src/shared/components/ScatterPlot.tsx
@@ -76,7 +76,7 @@ const ScatterPlot: FunctionComponent<Props> = ({
     legendOrientationThreshold
   )
 
-  let timerange = null
+  let timerange
 
   if (xColumn === '_time') {
     timerange = timeRange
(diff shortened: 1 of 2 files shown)


src/shared/components/ScatterPlot.tsx
```diff
@@ -74,14 +76,22 @@ const ScatterPlot: FunctionComponent<Props> = ({
...
+
+  if (xColumn === '_time') {
+    timerange = timeRange
+  }
+
   const [xDomain, onSetXDomain, onResetXDomain] = useVisXDomainSettings(
     storedXDomain,
-    table.getColumn(xColumn, 'number')
+    table.getColumn(xColumn, 'number'),
+    ...timerange
```
TCL735, line 88: I think this will throw an error when `timerange` is null because you cannot spread a null value into the function call arguments.

TCL735 (changes requested):

src/shared/components/ScatterPlot.tsx
zoesteinkamp: alright i went ahead and fixed this, so now it defaults not to null

TCL735 (approved):

TCL735 (approved):

PR #230, opened 2020-10-28
fix: resolved issue where selecting a second bucket in the metric selector crashed the app

### LINKED ISSUE #227: can't have multiple metric selectors in a flow
we would like to have multiple metric selectors in a flow

Asalem1:
Closes #227 

bug-fix


drdelambre (approved):
you are crushing it today!

PR #232, opened 2020-10-28
feat: data exploder

### LINKED ISSUE #112: Flows: Data Exploder
As an early access flows user, I want to start playing with data as soon as possible and don't yet know the available cell types or how to add them.

GIVEN I am on the flows index page
WHEN I click `Create Flow`
THEN I see a new flow that already has a `Input from Bucket` cell and a `Visualization Cell`

We got feedback from user testing!

there's a bit of a hurdle when we drop a user into a new flow. once they see how it works, they seem to get it, it's just that first jump is too big of a divide. There's two things that need to happen for this.

1) when a user creates a flow, prepopulate it with a metric selector with nothing selected in it, and a visualization pipe with all the defaults it already contains.

[Design Mock](https://www.figma.com/file/8S8IH8mimYIwanZ6keIQ6U/Notebooks-Prototype?node-id=885%3A6)

drdelambre:
Kapture 2020-10-28 at 15 16 38

closes #112 

Still dont have a data contract, which means i have no idea where the ids of secondary resources are going to come from, so some of the code is a bit weird. Should clean up after we iron all that out


src/flows/context/flow.list.tsx
```diff
@@ -48,51 +50,133 @@ export const FlowListContext = React.createContext<FlowListContextType>(
...
+        meta: flow.meta,
         readOnly: flow.readOnly,
       }
     }
 
-    setFlows({
-      ...flows,
-      [id]: _flow,
-    })
+    // console.log('add to the api', serialize(data))
```
drdelambre, line 145: i know i shouldn't leave in commented code. i'm using it here to show where the api plugs into the context mutations, lemme know it there's a better way to communicate this

Asalem1 (approved):
LGTM

PR #234, opened 2020-10-29
fix: remove that variable nonsense

### LINKED ISSUE #162: Metric selector kills previous results
while scoping out influxdata/idpe#8802, turns out that the metric selector has created a condition where we can't use previous results. stems from the aggregate window being applied.

drdelambre:
Kapture 2020-10-28 at 17 30 03

closes #162 
closes influxdata/idpe#8802

### DIFF
--- a/src/flows/pipes/Data/index.ts
+++ b/src/flows/pipes/Data/index.ts
@@ -51,7 +51,7 @@ export default register => {
       }
 
       if (aggregateFunction?.name) {
-        text += `|> aggregateWindow(every: v.windowPeriod, fn: ${aggregateFunction.name}, createEmpty: false)`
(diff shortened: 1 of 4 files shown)


src/flows/pipes/Data/index.ts
```diff
@@ -51,7 +51,7 @@ export default register => {
       }
 
       if (aggregateFunction?.name) {
-        text += `  |> aggregateWindow(every: v.windowPeriod, fn: ${aggregateFunction.name}, createEmpty: false)|> yield(name: "${aggregateFunction.name}")`
```
Asalem1, line 54: i think we need a space prepended for formatting when we preview the text

Asalem1 (approved):
a small thing, might wanna see how this looks when we export it as a task

PR #243, opened 2020-11-02
fix: filtering based on view name

drdelambre:
after all of our clever filtering to determine the valid subset of schema values, we apply another filter to ensure that, of that subset, the normalized / flattened structure only returns names that match the filter


src/shared/utils/flowSchemaNormalizer.ts
```diff
@@ -135,7 +135,9 @@ export const normalizeSchema = (
     data?.field
   )
 
-  const dedupedMeasurements = dedupeArray(measurements)
+  const dedupedMeasurements = dedupeArray(measurements).filter(a =>
```
Asalem1, line 138: Oh man, I recently went through this code on my branch simplified it. Long story short, it would be more readable and understandable if we did something like this:

```
const setMeasurements = new Set(measurements)
const dedupedMeasurementsArray = [...setMeasurements]

```

Asalem1 (commented):
Primary question: This doesn't necessarily address the issue of having chips (labels, tags, whatever you want to call it) on after selecting a `measurement`, `field`, `tag`. If the objective of the current filter functionality (including adding values as chips at the top of the input field after selecting a measurement, etc...) is to allow users to create a valid query, how would this filtration affect that?

Asalem1 (approved):

PR #246, opened 2020-11-02
fix(markdown): update Markdown Renderer Library to render the components correctly

Palakp41:
Closes influxdata/idpe#8810
- Updated from `ReactMarkdown` to `marked` library
- Updated the renderer code to render literal string with image url if the markdown contains `` tag
- Created a test for markdown renderer when markdown has an img tag it should render literal string

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,6 +164,7 @@
     "intersection-observer": "^0.7.0",
     "jsonlint-mod": "^1.7.5",
     "lodash": "^4.3.0",
+    "marked": "^1.2.2",
     "memoize-one": "^4.0.2",
     "moment": "^2.13.0",
     "monaco-editor": "^0.19.2",
--- a/src/shared/components/views/MarkdownRenderer.tsx
+++ b/src/shared/components/views/MarkdownRenderer.tsx
@@ -28,7 +28,8 @@ export const MarkdownRenderer: FC<Props> = ({
           if (cloudRenderers?.image) {
             return cloudRenderers.image
           }
-          return `${html.match(/"([^"]+)"/)[0]}`
+          const url = html.match(/"([^"]+)"/)[0]
+          return `![](${url})`
         }
         return html
       },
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -164,6 +164,7 @@
     "intersection-observer": "^0.7.0",
     "jsonlint-mod": "^1.7.5",
     "lodash": "^4.3.0",
+    "marked": "^1.2.2",
```
hoorayimhelping, line 167: i think we also want to run `yarn remove react-markdown` to get rid of the older library, right?

121watts (approved):

src/shared/components/views/MarkdownRenderer.tsx
```diff
@@ -27,16 +22,42 @@ export const MarkdownRenderer: FC<Props> = ({
...
   // don't parse images in cloud environments to prevent arbitrary script execution via images
   if (CLOUD) {
-    const renderers = {image: imageRenderer, imageReference: imageRenderer}
+    const renderer = {
+      html(html) {
+        if (html.includes('<img')) {
+          if (cloudRenderers?.image) {
+            return cloudRenderers.image
+          }
+          return `${html.match(/"([^"]+)"/)[0]}`
```
hoorayimhelping, line 31: is it possible to return the literal string
```
const url = html.match(/"([^"]+)"/)[0]
<img src="`${url}`" />
```
and not render the image? it's a little odd entering `` and seeing `"http://www.example.com/image.gif"`

hoorayimhelping (changes requested):
Good work! 

One question about returning the literal representation of an image

package.json
Palakp41: Oh yes, forgot that i hadn't removed it. Good catch!

src/shared/components/views/MarkdownRenderer.tsx
Palakp41: I had to instead return "``" for it to display as a literal string.

hoorayimhelping (approved):

PR #250, opened 2020-11-02
test: select all buckets in Generate Read/Write token

121watts:
Closes #217 

Bonitoo has written some e2e tests for us. Until we get forked repo credentials worked out I'm just going to copy the branches and push them up myself.

This is one of those tests.

### DIFF
--- a/cypress/e2e/tokens.test.ts
+++ b/cypress/e2e/tokens.test.ts
@@ -439,4 +439,238 @@ describe('tokens', () => {
         })
       })
   })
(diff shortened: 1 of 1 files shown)


drdelambre (approved):
for my conscience: we have no way of validating this test suite or if we'll ever be able to turn it back on (been off for 5mo already right?), so i dont think we should merge it. /giphy shot in the foot /giphy time bomb /giphy pray its ok

PR #253, opened 2020-11-02
fix: comment spacing

121watts:
adds a rule enforcing whitespace after starting a comment line:
`// this is a comment`
vs
`//this is a comment`

99% of the time we used the former.  the 1% of the time I saw the latter it bugged me.  a computer should be enforcing this.


drdelambre (approved):
i hate it

PR #254, opened 2020-11-03
feat(idpe-ui-e2e): Add circle config and cypress fixtures

ebb-tide:
Add circle config and cypress task to make UI e2e tests work

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -154,11 +154,12 @@ jobs:
           no_output_timeout: 30m
           command: |
             cd ./monitor-ci; docker-compose -f compose/fig.cloud.yml -f compose/fig.cypress.yml run cypress ./node_modules/.bin/cypress run --browser chrome --config numTestsKeptInMemory=1 --spec "$(cd ../ui; circleci tests glob "./cypress/e2e/**/*.test.ts" | circleci tests split | paste -sd "," -)"
-      - run:
-          name: Push IDPE UI E2E images
-          command: |
-            docker push quay.io/influxdb/ui-e2e:${CIRCLE_SHA1}
-            docker push quay.io/influxdb/ui-e2e:latest
+      - when:
+          condition:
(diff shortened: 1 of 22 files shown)


ebb-tide:
This isn't gonna work until https://github.com/influxdata/monitor-ci/pull/104 is merged

drdelambre (dismissed):
We are building the cypress image during the cloud-e2e stage. It is available after the `make build NODE=cypress` line. Push that up after build or refactor the e2e task to look more like the other prebuilds in the config (using the cache tar ball then load into docker directly technique) to ensure that we are all running the same tests. As is, it doesn't provide us with the operational safe guards we are looking for.

drdelambre (commented):
the circle stuff looks great. are you and watts talking about the user incremented part?

cypress/support/commands.ts
```diff
@@ -16,10 +16,20 @@ export const signin = (): Cypress.Chainable<Cypress.Response> => {
     })
   \*/
 
+  let userName = Cypress.env('username')
+  const runningAgainstIDPE = Cypress.env('runningAgainstIDPE') || false
```
wojciechka, line 20:
```suggestion
  const runningAgainstIDPE = Cypress.env('runningAgainstIDPE') !== undefined
```

.circleci/config.yml
```diff
@@ -152,6 +154,11 @@ jobs:
           no_output_timeout: 30m
           command: |
             cd ./monitor-ci; docker-compose -f compose/fig.cloud.yml -f compose/fig.cypress.yml run cypress ./node_modules/.bin/cypress run --browser chrome --config numTestsKeptInMemory=1 --spec "$(cd ../ui; circleci tests glob "./cypress/e2e/**/*.test.ts" | circleci tests split | paste -sd "," -)"
+      - run:
+          name: Push IDPE UI E2E images
+          command: |
```
ebb-tide, line 159: // need to add something like this
      - when:
          condition:
            equal: [ "master", "<< pipeline.git.branch >>" ]

PR #258, opened 2020-11-03
chore: update giraffe to 0.42.0 the one that switched to webpack builds

hoorayimhelping:
Updates to the latest giraffe. Should be no different to the application.

See: https://github.com/influxdata/giraffe/pull/349


TCL735 (approved):

TCL735:
"Automatic publicPath is not supported in this browser" - never seen this before. Does a dependency's webpack.config have any effect on this repo's unit tests?

hoorayimhelping:
@TCL735 i think it has to do with the difference in versions. this repo uses webpack 4, giraffe is on webpack 5.

PR #259, opened 2020-11-04
feat: derive orgid from bucket selection in query

drdelambre:
cause.. you can't "own" a demo data bucket?

uses the flux query parser provided by the flux team to generate an ast of the query, then we walk the tree to find a bucket name in there, and try to cross reference it with the buckets that are existing in the system. Same technique that is used in time machine.

i'm not happy about waiting for buckets to be loaded before we can run a query, feels like a hydrate variables situation, but it is what it is.


Asalem1 (approved):
Man this is depressing

PR #283, opened 2020-11-06
test: turn off skip

### LINKED ISSUE #172: Find out why e2e/buckets.test.ts is failing
Bucket e2e tests fail in the middle of execution because the app routes the user back to login.  Only happens when testing buckets.

zoesteinkamp:
closes #172 

So when i turned off the skip nothing failed locally, or in PR so far -_- .

### DIFF
--- a/cypress/e2e/buckets.test.ts
+++ b/cypress/e2e/buckets.test.ts
@@ -1,6 +1,6 @@
 import {Bucket, Organization} from '../../src/types'
 
-describe.skip('Buckets', () => {
+describe('Buckets', () => {
   beforeEach(() => {
     cy.flush()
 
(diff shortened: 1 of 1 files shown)


cypress/e2e/buckets.test.ts
```diff
@@ -173,17 +173,19 @@ describe.skip('Buckets', () => {
     })
 
     // TODO: ZOE Come back and fix race condition
```
121watts, line 175: do we need this comment anymore?

121watts (approved):

cypress/e2e/buckets.test.ts
zoesteinkamp: nope nope!

PR #285, opened 2020-11-09
feat: add tick generation options to view options

### LINKED ISSUE #238: Integrate Giraffe tick generation options into View Options
Integrate Giraffe's new [tick generation](https://github.com/influxdata/giraffe/issues/79) options into the View Options. This will be feature flagged behind the flag `axisTicksGenerator`.

A user should be able to specify any of the following: total number of ticks, tick interval (step), and starting point of the first tick. User's selections should be saved per cell. If a user does not specify any of these properties, or if missing properties leads to a failure to determine ticks for rendering, then the ticks should be auto-populated or determined by Giraffe similar to behavior prior to this feature. However, a deliberate "0 ticks rendered" scenario should be allowed through overtly specified options.

The following views should have options for tick generation:
 - XYView
 - XY + Single Stat View
 - BandView
 - HeatmapView
 - ScatterView
 - MosaicView (x axis only)

TCL735:
Closes #238 

Adds options to View Options to allow user to create ticks on each axis for Line, Band, Heatmap, and Scatter plots. Mosaic plot gets x-axis only.

The options are total ticks, tick start, and tick step. They may be used in any combination. The options look like this:
Screen Shot 2020-11-08 at 6 12 50 PM


src/shared/components/axisTicks/TimeTickInput.tsx
```diff
@@ -0,0 +1,236 @@
...
+      <Grid.Column widthXS={Columns.Six}>
+        <Form.Element label="Date Picker">
+          <Popover
+            appearance={Appearance.Outline}
+            position={PopoverPosition.ToTheLeft}
+            triggerRef={triggerRef}
+            visible={isDatePickerOpen}
+            showEvent={PopoverInteraction.None}
+            hideEvent={PopoverInteraction.None}
+            distanceFromTrigger={8}
```
hoorayimhelping, line 181: magic number! what's 8 mean?

src/shared/components/axisTicks/TimeTickInput.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const onClickOutside = () => {
+    if (isOnClickOutsideHandlerActive) {
+      hideDatePicker()
+    }
+  }
+
+  const allowOnClickOutside = () => setIsOnClickOutsideHandlerActive(true)
+  const suppressOnClickOutside = () => setIsOnClickOutsideHandlerActive(false)
+
+  const style: CSSProperties = isDatePickerOpen
```
hoorayimhelping, line 149: i think conventionally it's called `styles`

src/shared/components/axisTicks/TimeTicksOptions.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const {
+    axisName,
+    tickOptions,
+    totalTicks,
+    tickStart,
+    tickStep,
+    setOptions,
+  } = props
+  return (
+    <>
```
hoorayimhelping, line 28:
```suggestion
```

this isn't needed right? `Grid.Row` contains all the `ValueTickInputs`

src/shared/components/axisTicks/TimeTicksOptions.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <ValueTickInput
+          axisName={axisName}
+          tickOptions={tickOptions}
+          initialTickOptionValue={tickStep}
+          label="Tick Step"
+          placeholder="milliseconds"
+          setOptions={setOptions}
+        />
+      </Grid.Row>
+    </>
```
hoorayimhelping, line 53:
```suggestion
```

src/shared/components/axisTicks/ValueTickInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+      : []
+    setOptions(tickOptionNameWithoutAxis, convertedValue)
+    if (convertedValue === convertedValue) {
+      setOptions('GenerateAxisTicks', [
+        ...filteredTickOptions,
+        tickOptionNameWithAxis,
+      ])
+    } else {
+      setOptions('GenerateAxisTicks', filteredTickOptions)
+      if (tickOptionInput !== '') {
```
hoorayimhelping, line 63: same comment here about the structure of this `if` statement

src/shared/components/axisTicks/ValueTicksOptions.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const {
+    axisName,
+    tickOptions,
+    totalTicks,
+    tickStart,
+    tickStep,
+    setOptions,
+  } = props
+  return (
+    <>
```
hoorayimhelping, line 27:
```suggestion
```
same thing here with `Grid.Row` being a container, right?

src/shared/utils/convertUserInput.test.ts
```diff
@@ -67,3 +70,67 @@ describe('convertUserInputToNumOrNaN', () => {
     expect(convertUserInputToNumOrNaN(event)).toEqual(0)
   })
 })
+
+describe('convertUserInputValueToNumOrNaN', () => {
+  it('should convert undefined to NaN', () => {
+    let result = convertUserInputValueToNumOrNaN()
+    expect(typeof result === 'number')
+    expect(result !== result)
+
```
hoorayimhelping, line 79: i think you should add an expect for `isNan` like 
```ts
expect(Number.isNan(result)).toBeTruthy()
```
to these tests

(i also prefer `toBeTruthy` because it's more explicit)

src/shared/utils/convertUserInput.ts
```diff
@@ -1,5 +1,8 @@
 import React from 'react'
 
+export const convertUserInputValueToNumOrNaN = (value?: any): number =>
+  value === '' ? NaN : Number(value)
```
hoorayimhelping, line 4: Why does `''` get converted to `NaN` and not `Number('') // 0`?

src/shared/utils/useAxisTicksGenerator.ts
```diff
@@ -0,0 +1,31 @@
...
+export const useAxisTicksGenerator = ({
+  xTotalTicks,
+  xTickStart,
+  xTickStep,
+  yTotalTicks,
+  yTickStart,
+  yTickStep,
+}) =>
+  useMemo(() => {
+    if (!isFlagEnabled('axisTicksGenerator')) {
```
hoorayimhelping, line 13: we have to be very careful about how this is deployed if this is done through config cat. the way `isFlagEnabled` is written, you can't be sure that the results will be correct, depending on when the function is called in request's lifecycle

hoorayimhelping (changes requested):
Great work! have a few suggestions / questions

src/shared/utils/convertUserInput.ts
TCL735: This is specifically used when we have user input. An empty string means the user has deleted all of the input field. We should interpret that case as "not a number" rather than 0.

src/shared/components/axisTicks/AxisTicksGenerator.tsx
```diff
@@ -0,0 +1,203 @@
...
+import {AppState} from 'src/types'
+import {ButtonShape} from '@influxdata/clockface'
+
+const getGenerateAxisTicksOptionsNames = (axisName: string) => {
+  const result = []
+  result.push(`generate${axisName.toUpperCase()}AxisTicks`)
+  result.push(`${axisName.toLowerCase()}TotalTicks`)
+  result.push(`${axisName.toLowerCase()}TickStart`)
+  result.push(`${axisName.toLowerCase()}TickStep`)
+  return result
```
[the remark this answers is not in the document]
TCL735: Oops, yeah no reason to do this here. I think this was the remnants of console.log-ing the return value and not using the shorter syntax after removing it.

I will update it to the shorter syntax.

src/shared/components/axisTicks/AxisTicksGenerator.tsx
```diff
@@ -0,0 +1,203 @@
...
+      })
+      return total
+    }, 0) > 0
+  )
+
+  const isTimeColumn = columnType === '_time' ? true : false
+
+  if (!isFlagEnabled('axisTicksGenerator')) {
+    return null
+  }
```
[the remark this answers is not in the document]
TCL735: It is about as high as it can be. It needs to come after all hooks, as early return is not allowed with hooks.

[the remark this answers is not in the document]
TCL735: i can move it above `isTimeColumn`

src/shared/components/axisTicks/AxisTicksGenerator.tsx
```diff
@@ -0,0 +1,203 @@
...
+    onSetGenerateAxisTicks,
+    onSetTotalTicks,
+    onSetTickStart,
+    onSetTickStep,
+    totalTicks,
+    tickStart,
+    tickStep,
+  } = props
+
+  const [showInputs, setShowInputs] = useState(
```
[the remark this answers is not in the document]
TCL735: Sounds good. Done.

src/shared/components/axisTicks/AxisTicksGenerator.tsx
```diff
@@ -0,0 +1,203 @@
...
+}
+
+type ReduxProps = ConnectedProps<typeof connector>
+type Props = AxisTicksGeneratorProps & ReduxProps
+
+const AxisTicksGenerator: FC<Props> = props => {
+  const {
+    axisName,
+    columnType,
+    generateAxisTicks,
```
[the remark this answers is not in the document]
TCL735: yeah, it does sound like a function. However, "generated" is not correct either. Because it really is just a list of options to use for generating the ticks, not the ticks themselves.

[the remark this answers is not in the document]
TCL735: I wanted to match the name in the api, so that's how it ended up as "generateAxisTicks".

The bigger issue is that it might be misnamed in the api as well, if "generateAxisTicks" sounds too much like a function.

src/shared/components/axisTicks/AxisTicksGenerator.tsx
```diff
@@ -0,0 +1,203 @@
...
+        hasTickStep = true
+      }
+    })
+  }
+
+  return {
+    generateAxisTicks: Array.isArray(generateAxisTicks)
+      ? generateAxisTicks
+      : [],
+    totalTicks: hasTotalTicks ? totalTicks : '',
```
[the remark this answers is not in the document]
TCL735: This will be feed into a user input field. An empty string represents "no user input".

Even if we don't want to use an empty string here, there is a subtle distinction between "lack of value" versus 0. "lack of value" means, to Giraffe, please calculate the ticks as you see fit. 0 means exactly that -- 0. For example, 0 for total ticks would mean don't label the axis with any ticks. Or, for tick start, 0 would mean start at the value 0 on the axis. (Tick step can also be 0 -- Giraffe handles that gracefully.)

And, so, we use an empty string here to denote "lack of value" which is not the same as 0.

src/shared/components/axisTicks/TimeTickInput.tsx
```diff
@@ -0,0 +1,236 @@
...
+  setOptions: (optionName: string, arg: string[] | number) => void
+}
+
+export const TimeTickInput: FC<TimeTickInputProps> = props => {
+  const {
+    axisName,
+    tickOptions,
+    initialTickOptionValue,
+    label,
+    placeholder = 'RFC3339',
```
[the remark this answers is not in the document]
TCL735: Ah yes, good call. Done.

src/shared/components/axisTicks/TimeTickInput.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  const handleKeyPress = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      updateTickOption()
+    }
+  }
+
+  const getDatePickerDateTime = () => {
+    const date = new Date(tickOptionInput)
+    if (date.valueOf() === date.valueOf()) {
```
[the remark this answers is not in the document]
TCL735: if `date` is an invalid date, then calling .valueOf() will return `NaN` and `NaN` is never equal to itself.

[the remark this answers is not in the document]
TCL735: I will update this to use `Number.isNaN()` to be more explicit

src/shared/components/axisTicks/TimeTickInput.tsx
TCL735: According to clockface, this is the pixel distance between the trigger (our button) and the popover's dialog box.

All of the `Popover`s in ui have `8` as the value. I just copied it. Perhaps we can clean that up all in one pass rather than piecemeal. Can we leave this for another PR?

src/shared/components/axisTicks/TimeTickInput.tsx
TCL735: `styles` it is. Done.

src/shared/components/axisTicks/TimeTicksOptions.tsx
TCL735: Correct, not needed. I had changed back and forth between a few designs. Forgot to remove the fragment tag. Done now.

src/shared/utils/convertUserInput.test.ts
TCL735: oh yeah good call. Done.

TCL735: And... actually, `expect(result !== result)` doesn't assert anything. It always passes even if the expression inside is false 😅 

So this change was definitely needed, just to get the tests to run correctly.

src/shared/components/axisTicks/TimeTickInput.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  const handleInput = (event: ChangeEvent<HTMLInputElement>) => {
+    setTickOptionInput(event.target.value)
+    setTickOptionInputStatus(ComponentStatus.Default)
+  }
+
+  const handleBlur = () => updateTickOption()
+
+  const handleKeyPress = (event: React.KeyboardEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
TCL735: Maybe.

Here, we need it because we want to wait until the user types the whole input before attempting to generate the ticks. Without this, as soon as the user types `1` in Tick Step, it would attempt to generate ticks on a step of 1 millisecond, for example. This helped catch a performance issue in giraffe. But we also do not want eager input fields.

So, a long way of saying: yes, in certain cases, we want this.

src/shared/components/axisTicks/TimeTickInput.tsx
```diff
@@ -0,0 +1,236 @@
...
+    if (convertedValue === convertedValue) {
+      setOptions('GenerateAxisTicks', [
+        ...filteredTickOptions,
+        tickOptionNameWithAxis,
+      ])
+    } else {
+      setOptions('GenerateAxisTicks', filteredTickOptions)
+      if (tickOptionInput !== '') {
+        setTickOptionInputStatus(ComponentStatus.Error)
+      }
```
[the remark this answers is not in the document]
TCL735: Yes but with a small change. I will need to call `setOptions` with the computedTickOptions no matter what, so that call goes last.

And i think checking convertedValue might be confusing. I changed it to a check of `Number.isNaN` because that is basically what i want to know here.

Also, I changed it to an if - else if structure. i'd rather not check `NaN`ness twice. Just once and the `else if` can check the other condition.

TCL735:
This also requires changes to make it work properly in the Dashboard page.

TCL735:
@hoorayimhelping in addition to addressing the review comments, i also added another commit to fix dashboard cells. Now, they correctly display the user's selected values from the `<AxisTicksGenerator>`

src/shared/components/axisTicks/TimeTickInput.tsx
hoorayimhelping: >Can we leave this for another PR?

sure, sounds like a good idea

src/shared/components/axisTicks/AxisTicksGenerator.tsx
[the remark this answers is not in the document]
hoorayimhelping: >It needs to come after all hooks, as early return is not allowed with hooks.

🤦 thanks, didn't know this

hoorayimhelping (approved):

Issue #287, opened 2020-11-09
Create draft of unity swagger for users page


121watts:
- [x] Delete User
   - [x] Get Users
   - [x] Delete Invitation
   - [x] Resend Invitation
   - [x] Create Invitation

121watts:
I've already done this once in: https://github.com/influxdata/influxdb/pull/18167

This will be a standalone swagger that can live in `influxdata/ui` for now.

121watts:
- [ ] Delete User
   - [ ] Delete Invitation
   - [ ] Resend Invitation
   - [ ] Create Invitation
   - [ ] Get Users
   - [ ] Get Invites

PR #293, opened 2020-11-09
feat(rfc): unity swagger.yml Users and Invites draft

121watts:
Closes: #287 

This is a draft for the `Users` and `Invites` unification API client.  If we can knock out a contract between Quartz and the `ui` then e-commerce and ui can start working in parallel 😸 .

   - [x] Delete User
   - [x] Get Users
   - [x] Delete Invitation
   - [x] Resend Invitation
   - [x] Create Invitation

### DIFF
--- a/src/client/unity.yml
+++ b/src/client/unity.yml
@@ -61,6 +61,41 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  "/orgs/{orgID}/invites":
+    post:
+      operationId: PostOrgsIDInvites
+      tags:
+        - Invites
+        - Organizations
+      summary: Creates an invite to an organization
+      parameters:
+        - $ref: "#/components/parameters/TraceSpan"
+        - in: path
+          name: orgID
+          schema:
+            type: string
+          required: true
+          description: The organization ID.
+      requestBody:
+        description: Invite to be sent
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Invite"
+      responses:
+        "201":
+          description: Invite sent
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Invite"
+        default:
+          description: Unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
 components:
   parameters:
     TraceSpan:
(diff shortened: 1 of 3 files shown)


subirjolly:
I will give it a 👍 instead of approving. Don't consider myself qualified to approve, yet.

randycoulman (approved):
Overall, this looks about right, though I haven't had time to do a detailed comparison against what's actually implemented in Quartz right now, so we might need a couple of minor changes later.

This looks like what used to live in the influxdb swagger.  It looks like most of that was removed by https://github.com/influxdata/influxdb/pull/19102, but there's still some cruft left over there that we might want to remove.  I also noticed that the above-mentioned PR removed the `/orgs/{orgID}/owners/{userID}` endpoint, but it hasn't been added back in this PR.  However, I think that might be an error in the previous PR because I think that's still a valid endpoint in the Influx API.

My only other thought is that I think this Swagger spec should ultimately live in the Quartz repo because that's where it will be implemented, but I realize that makes it harder to generate code from it. If it stays here, I'd want to be sure that we're OK with publishing this stuff in an OSS repo.

src/client/unity.yml
```diff
@@ -0,0 +1,280 @@
+openapi: "3.0.0"
+info:
+  title: Unity API Service
+  version: 0.1.0
+servers:
+  - url: /unity/v2 # to be determined
```
gunnaraasen, line 6: Maybe simply `/ui/v2`.

src/client/unity.yml
```diff
@@ -0,0 +1,280 @@
...
+paths:
+  "/orgs/{orgID}/invites":
```
gunnaraasen, line 8: Should there also be a `get:` method to fetch the list of invites?

gunnaraasen (commented):
Thanks @121watts for taking a first run at this spec. 

Since this will be an API primarily for UI use in InfluxDB Cloud, I agree with Randy's suggestion this spec should live in the quartz repo until we're ready for the UI API routes to be publicly consumable.

121watts:
@gunnaraasen before this code is live we can horse trade and decide where stuff should live.  for now, im going to keep it here as this is the most convenient place for the `ui` team to consume it.

src/client/unity.yml
randycoulman: When this moves to Quartz, it will be `/api/v2private`.

src/client/unity.yml
randycoulman: The list of invites comes back when fetching the user list: you get back both users and invites in that response.

PR #294, opened 2020-11-09
refactor(flows): misc polish

alexpaxton:
- Change icons in upgrade buttons from `Star` to `Upgrade`
- Change icons on bucket selector and aggregate selector
- Show empty state in Metric Selector when nothing matches search term
- Round add cell buttons
- Improve legibility of cell focus state
- Polish visualization options
  - Guard against error when no data exists
- Fix jitter when resizing visualization for the first time
- Remove `Minimap`, `ScrollContext`, and state in redux (not in use, no plans to bring back any time soon)
- Remove legacy export to dashboard components


src/flows/components/PipeList.tsx
```diff
@@ -23,16 +20,7 @@ const PipeList: FC = () => {
     return <FlowPipe key={`pipe-${id}`} id={id} />
   })
 
-  return (
```
drdelambre, line 26: if you're gonna kill this, kill the scroll context as well

alexpaxton: Burn it all!!

drdelambre (approved):
super rad

PR #296, opened 2020-11-10
feat: flows in annotations

drdelambre:
it still needs a way to move to run the queries and set the time.. moving the header stuff straight over felt weird, but this gets most of the flow stuff up and running outside of the flow page.


alexpaxton (approved):
Sickkkkkk. Would love to introduce some concept of "modes" to flows to facilitate some guide rails on this overlay

PR #299, opened 2020-11-10
fix(de): cancel button cancels all running queries

Palakp41:
After clicking the cancel on data explorer, we don't get into the componentDidUpdate life-cycle because we are not updating the state of the query status in redux. The query status in the props is from the redux state query results which connects to all the query. Adding a dispatch to update the query status for query result to done when cancelled is clicked, fixes this issue: #idpe/7683 (https://app.zenhub.com/workspaces/applicationmonitoring-5e3b05772922e316b3a210a4/issues/influxdata/idpe/7683) 

Kapture 2020-11-10 at 14 02 44

### DIFF
--- a/src/flows/components/header/Submit.tsx
+++ b/src/flows/components/header/Submit.tsx
@@ -24,6 +24,7 @@ import {notify} from 'src/shared/actions/notifications'
 
 // Utils
 import {event} from 'src/cloud/utils/reporting'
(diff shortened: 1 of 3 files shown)


drdelambre (approved):

hoorayimhelping (approved):
awesome work, palak!

i edited the title from `Fix(de):` to `fix(de):` to make the semantic bot shut it.

PR #301, opened 2020-11-11
feat(uie2e): Remove the dependence of tests on what is returned from /debug/provision

ebb-tide:
Tests will no longer depend on what is returned from `/debug/provision` endpoint.

### DIFF
--- a/cypress/e2e/variables.test.ts
+++ b/cypress/e2e/variables.test.ts
@@ -1,11 +1,14 @@
+import {Organization} from '../../src/types'
+
 describe('Variables', () => {
   beforeEach(() => {
     cy.flush()
 
-    cy.signin().then(({body}) => {
-      cy.wrap(body.org).as('org')
-      cy.createQueryVariable(body.org.id)
-      cy.visit(`orgs/${body.org.id}/settings/variables`)
+    cy.signin().then(() => {
+      cy.get('@org').then(({id}: Organization) => {
+        cy.createQueryVariable(id)
+        cy.visit(`orgs/${id}/settings/variables`)
+      })
     })
 
     cy.location('pathname').should('match', /\/variables$/)
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -1,6 +1,5 @@
 import {NotificationEndpoint} from '../../src/types'
 import 'cypress-file-upload'
-import {Organization} from '../../src/types'
 
 export const signin = (): Cypress.Chainable<Cypress.Response> => {
   /* \ OSS login
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -1,5 +1,6 @@
 import {NotificationEndpoint} from '../../src/types'
 import 'cypress-file-upload'
+import {Organization} from '../../src/types'
 
 export const signin = (): Cypress.Chainable<Cypress.Response> => {
   /* \ OSS login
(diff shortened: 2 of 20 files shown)


cypress/support/commands.ts
```diff
@@ -16,15 +17,31 @@ export const signin = (): Cypress.Chainable<Cypress.Response> => {
...
+    .then(() =>
+      cy.request({
+        method: 'GET',
+        url: '/api/v2/orgs',
+      })
+    )
+    .then(response => {
+      cy.wrap(response.body.orgs[0]).as('org')
+    })
+    .then(() =>
```
ebb-tide, line 36: it might be a better idea to get `@bucket` only when it is needed by the test.

drdelambre:
how are we keeping tokens in sync with secondary services (ie: telegraf) between oss and cloud across testing cycles without /debug/provision?

drdelambre:
actually, what does this have to do with /debug/provision at all? i dont see anything wrong with the changes, just confused

cypress/support/commands.ts
```diff
@@ -16,15 +17,31 @@ export const signin = (): Cypress.Chainable<Cypress.Response> => {
...
   \*/
 
-  return cy.setupUser().then(body => {
-    return cy
-      .visit('/api/v2/signin')
-      .then(() => cy.get('#login').type(Cypress.env('username')))
-      .then(() => cy.get('#password').type(Cypress.env('password')))
-      .then(() => cy.get('#submit-login').click())
-      .then(() => cy.get('.theme-btn--success').click())
-      .then(() => cy.wrap(body))
```
ebb-tide, line 26: Previously--- all tests depend on what is returned from this 'body' returned from setupUser which is the response to a 'GET' request to `/debug/provision`

cypress/e2e/variables.test.ts
```diff
@@ -1,11 +1,14 @@
+import {Organization} from '../../src/types'
+
 describe('Variables', () => {
   beforeEach(() => {
     cy.flush()
 
-    cy.signin().then(({body}) => {
-      cy.wrap(body.org).as('org')
-      cy.createQueryVariable(body.org.id)
-      cy.visit(`orgs/${body.org.id}/settings/variables`)
```
ebb-tide, line 8: All tests depend on org id returned by `debug/provision`.

ebb-tide:
I removed the dependence of tests on what is returned from `debug/provision`. 
Previously, tests were getting orgID from response from `debug/provision`, but that was unnecessary because orgID can be got soo many other ways (that are actually in our API, even!)

drdelambre:
gotcha.. i guess the confusion is in the mention of /debug/provision

here's the OSS specific version that defined the contract for reference:
https://github.com/influxdata/influxdb/blame/master/ui/cypress/support/commands.ts#L427-L436

/debug/provision is an endpoint that serves to abstract the implementation specific details of generating that intent so that we can run the same test suite against our other environment requirements. the aforementioned contract is forwarded the same as is for oss, and recreated using the same APIs as this PR for cloud e2e testing.

so this is effectively adding another set of api calls for oss, and moving the implementation on the cloud side into the cypress tests so that we can reduce the contract for what /debug/provision returns?

ebb-tide:
You no longer need to have a contract for what `debug/provision` returns that different test environments need to fulfill. The test suite now depends entirely on what our actual API returns.

cypress/support/commands.ts
ebb-tide: done

drdelambre (approved):
please change the title to something more fitting before merging, other than that the changes look alright

PR #308, opened 2020-11-12
test(e2e): more bonitoo tests

121watts:
Closes: #64 

bonitoo has created some e2e tests for us.  these are some of those tests.


Cubone21:
@121watts So I fixed the failing tests. The problem was that the tests work locally but fail in circleci. I've fixed some of them but some are still failing so I've put them on skip just so the working tests can finally be merged in and the failing ones can be fixed later.

TCL735 (approved):

PR #309, opened 2020-11-12
fix(variables): improve drag & drop variables in dashboards interaction 

### LINKED ISSUE #275: Dashboards: Variables can no longer be re-arranged by clicking and dragging
Kapture 2020-11-05 at 08 51 43

you used to be able to rearrange these variables on the dashboard, but now you can't

alexpaxton:
Closes #275 

### Changes

- Refactor touched components to `FunctionComponent`
- Add `react-beautiful-dnd` dependency as an eventual replacement for `react-dnd`
  - More composable, more customizable, and easier to reason about
  - `react-dnd` is still being used by the table options component (out of scope for this PR)
- Refactor move variable reducer to move a variable (instead of swapping two variables)
- Add `removeView` action
- Call `removeView` when a cell is deleted so that redux doesn't have a stale list of views until next refresh
  
re-order-variables


src/variables/reducers/index.ts
```diff
@@ -102,11 +102,10 @@ export const variablesReducer = (
...
         newOrder = newOrder.slice(0)
 
-        const idToMove = newOrder[originalIndex]
-        const idToSwap = newOrder[newIndex]
-
-        newOrder[originalIndex] = idToSwap
-        newOrder[newIndex] = idToMove
+        // Pull out variable and insert at new location
+        const variable = newOrder[originalIndex]
+        newOrder.splice(originalIndex, 1)
```
Asalem1, line 107:
```suggestion
        newOrder.splice(originalIndex, 1)
```
```suggestion
        const variable = newOrder.splice(originalIndex, 1)
```

Asalem1 (approved):
A couple of small things but looks great!

PR #310, opened 2020-11-12
feat: give pipes more meaningful names

drdelambre:
also renames the project back to `notebooks` and adds a new pipe in there as well for all of those times we feel it pertinent to change a type definition

Screenshot from 2020-11-12 12-30-59


src/annotations/components/overlay/CreateAnnotationStreamOverlay.tsx
```diff
@@ -28,8 +28,8 @@ const DEFAULT_FLOW = JSON.stringify({
     {
       title: 'Select a Metric',
       visible: true,
-      type: 'queryBuilder',
-      ...JSON.parse(JSON.stringify(PIPE_DEFINITIONS['queryBuilder'].initial)),
+      type: 'metricSelector',
```
Asalem1, line 31: RIP queryBuilder

src/flows/components/header/index.tsx
```diff
@@ -30,7 +31,7 @@ const FlowHeader: FC = () => {
         <RenamablePageTitle
           onRename={handleRename}
           name={flow.name}
-          placeholder="Name this Flow"
+          placeholder={`Name this ${PROJECT_NAME}`}
```
Asalem1, line 34: nit: `Flow` is capitalized, but `PROJECT_NAME` isn't. Is that gonna be a problem?

121watts (approved):
can we please go 💯 on this and also rename the directories to "notebooks".  it gets really confusing for me when i need to search for "foos" but they are all named "bar"

src/flows/components/header/index.tsx
drdelambre: that is the product request, to specifically make it "notebooks with a lower case n"

src/annotations/components/overlay/CreateAnnotationStreamOverlay.tsx
drdelambre: that's why i ended up doing this... started to migrate the DE query builder, and... the name was taken

src/flows/components/header/index.tsx
121watts: 🚲 🏠 .  looks odd for it to be the only lowercased thing in the sidenav.  idk not a designer.

drdelambre:
@121watts no. i am totally with you, and it requires the dev to have context of a history that may not have been communicated to them, but product has made it painfully clear that we're going to be renaming this thing again in month. once we have a real project name, and less of this "working title" back and forth, remind me and i'll do that cleanup pass

121watts:
@drdelambre 🤦

PR #333, opened 2020-11-17
feat(reporting): enable variant experiments in UI

mavarius:
This PR adds a new component and utility function to enable creating Google Optimize experiments directly in the UI. It also adds the first experiment we'll be testing with this method: changing the upgrade icon.

### DIFF
--- a/src/cloud/utils/experiments.ts
+++ b/src/cloud/utils/experiments.ts
@@ -1,0 +1,23 @@
+const pushDataLayerEvent = async (dataLayerEvent: string) => {
+  if (window.dataLayer) {
+    await window.dataLayer.push({event: dataLayerEvent})
+  }
+}
+
+const getVariant = (experimentID: string, attempts: number = 10): string => {
+  if (attempts <= 0) {
+    return ''
+  }
+  if (window['google_optimize']) {
+    return window['google_optimize'].get(experimentID) || null
+  }
+  setTimeout(() => getVariant(experimentID, attempts - 1), 100)
+}
+
+export const getExperimentVariantId = (
+  experimentID: string,
+  activationEvent: string = 'optimize.activate'
+): string => {
+  pushDataLayerEvent(activationEvent)
+  return getVariant(experimentID)
+}
--- a/src/cloud/utils/reporting.ts
+++ b/src/cloud/utils/reporting.ts
@@ -185,25 +185,26 @@ export const useLoadTimeReporting = (measurement: string) => {
   }, [measurement, loadStartTime])
 }
 
+const pushDataLayerEvent = async (dataLayerEvent: string) => {
+  if (window.dataLayer) {
+    await window.dataLayer.push({event: dataLayerEvent})
+  }
+}
+
+const getVariant = (experimentID: string, attempts: number = 10): string => {
+  if (attempts <= 0) {
+    return ''
+  }
+  if (window['google_optimize']) {
+    return window['google_optimize'].get(experimentID) || ''
+  }
+  setTimeout(() => getVariant(experimentID, attempts - 1), 100)
+}
+
 export const getExperimentVariantId = (
   experimentID: string,
   activationEvent: string = 'optimize.activate'
 ): string => {
-  const activateOptimize = async (activationEvent: string) => {
-    if (window.dataLayer) {
-      await window.dataLayer.push({event: activationEvent})
-    }
-  }
-  activateOptimize(activationEvent)
-
-  const getVariant = (): string => {
-    const googleOptimize = window['google_optimize']
-    if (googleOptimize) {
-      const variantID = googleOptimize.get(experimentID)
-      return variantID
-    } else {
-      setTimeout(getVariant, 100)
-    }
-  }
-  return getVariant()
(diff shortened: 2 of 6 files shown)


src/shared/components/CloudUpgradeNavBanner.tsx
```diff
@@ -62,7 +63,22 @@ const CloudUpgradeNavBanner: FC<StateProps> = ({inView}) => {
             href={`${CLOUD_URL}${CLOUD_CHECKOUT_PATH}`}
             target="_self"
           >
-            <Icon glyph={IconFont.Upgrade} />
+            <GoogleOptimizeExperiment
+              experimentID="e44rY7GjQN-ASmGeWLs_pA"
```
mnorriswalker, line 67: Love how easy this is to read. This is just for A/B test with one variant, but it looks like we could add multiple variants too for a multi-variate test. Niiice.

mnorriswalker (approved):

src/shared/components/CloudUpgradeNavBanner.tsx
mavarius: I wanted to start with a simple test; this one is 2 variants plus the original.

src/cloud/utils/reporting.ts
```diff
@@ -184,3 +184,26 @@ export const useLoadTimeReporting = (measurement: string) => {
...
+  activationEvent: string = 'optimize.activate'
+): string => {
+  const activateOptimize = async (activationEvent: string) => {
+    if (window.dataLayer) {
+      await window.dataLayer.push({event: activationEvent})
+    }
+  }
+  activateOptimize(activationEvent)
+
+  const getVariant = (): string => {
```
hoorayimhelping, line 199: any reason not to move the definition `activateOptimize` and `getVariant` to outside of `getExperimentVariantId`? right now, every time this function is called, those functions are being re-created then called. looks to me like they only need to be created once

src/shared/components/CloudUpgradeButton.tsx
```diff
@@ -44,11 +47,23 @@ const CloudUpgradeButton: FC<StateProps & OwnProps> = ({
     [`${className}`]: className,
   })
 
+  const [iconVariant, setIconVariant] = useState<IconFont>(IconFont.Upgrade)
```
hoorayimhelping, line 50: i see what you're going for here, but the name `iconVariant` makes it seem like it is *the* variant, not one of the variants.

what about `iconVariantForUser` or `usersIconVariant` to make it the specific variant for this user?

src/shared/components/CloudUpgradeNavBanner.tsx
hoorayimhelping: so this will work for this PR and experiment, but in general, having a key or an id hard coded into code like this is an indicator we need a little more abstraction. right now, it'll only work for the specific experiment id we provide in code, which means it'll only work in prod for a specific container, right? if we abstract this `experimentId` into an environment variable, we can inject the correct value per environment (e.g. tools, staging, prod), which would let us test in non-production environments (or any arbitrary environment).

using component props for experiment handling is fine for a couple of variants and for simple components, but it might get unwieldy really quickly for complex components or component trees and hierarchies.

hoorayimhelping (changes requested):
great work! 

have two requests - one to move the function declarations out of `getExperimentVariantId` and add some kind of failure condition for fetching the variant from optimize, in case it doesn't work.

src/cloud/utils/reporting.ts
```diff
@@ -184,3 +184,26 @@ export const useLoadTimeReporting = (measurement: string) => {
...
+    const googleOptimize = window['google_optimize']
+    if (googleOptimize) {
+      const variantID = googleOptimize.get(experimentID)
+      return variantID
+    } else {
+      setTimeout(getVariant, 100)
```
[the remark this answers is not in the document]
mavarius: Good catch

src/shared/components/CloudUpgradeNavBanner.tsx
mavarius: Let's chat a bit more about how we can abstract this in the future.

src/cloud/utils/experiments.ts
```diff
@@ -0,0 +1,23 @@
...
+    await window.dataLayer.push({event: dataLayerEvent})
+  }
+}
+
+const getVariant = (experimentID: string, attempts: number = 10): string => {
+  if (attempts <= 0) {
+    return ''
+  }
+  if (window['google_optimize']) {
+    return window['google_optimize'].get(experimentID) || null
```
hoorayimhelping, line 12: fyi, i changed this from `''` to `null` because `null` is slightly more explicit for how it's being used:

```ts
if (variantId) {
...
```

hoorayimhelping (approved):
lgtm!

i made a couple of small structural changes - i made an `experiments` directory in `cloud/utils` and put the optimize code you wrote there. also changed the Google Optimize template directory from `googleOptimize` to `experiments`. removed a default export in favor of a named export.

tested it locally and the fallback worked for me

PR #336, opened 2020-11-18
feat(unify-users): move Users page from Quartz with mocked data

### LINKED ISSUE #304: Move Quartz Users Page to influxdata/ui
Stub the routes w/ mock data

121watts:
Closes #304 

Kapture 2020-11-17 at 16 02 09


cypress/e2e/cloud/users.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    cy.getByTestIDSubStr('invite-list-item').should(
+      'have.length',
+      invites.length
+    )
+
+    cy.getByTestID(`user-list-item iris@influxdata.com`).within(() => {
+      cy.getByTestID('delete-user--button').trigger('mouseover')
+      cy.getByTestID('delete-user--button').should('be.visible')
+      cy.getByTestID('delete-user--button').click()
+    })
```
randycoulman, line 72: 👎  Please don't delete @ischolten from our org! 😛

src/client/unity.yml
```diff
@@ -6,6 +6,31 @@ servers:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Invites"
+        default:
+          description: Unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
```
randycoulman, line 33: So separate endpoint for pulling invites rather than getting both users and invites with the same endpoint? That makes a lot of sense! 👍

src/pageLayout/components/UserWidget.tsx
```diff
@@ -78,6 +79,16 @@ const UserWidget: FC<Props> = ({
             />
           )}
         />
+        <FeatureFlag name="unity">
+          <TreeNav.UserItem
+            id="unity-users"
+            label="Unity Users"
+            testID="user-nav-item-unity-users"
+            linkElement={className => (
+              <a className={className} href={`/orgs/${org.id}/unity-users`} />
```
randycoulman, line 88: Do we want `unity` in our route name and on the Nav Item? Or is that temporary because we have both `users` and `unity users` right now?

src/types/unity.ts
```diff
@@ -0,0 +1,119 @@
...
+export interface InviteErrors {
+  email?: string[]
+}
+
+export interface InviteErrorsJSON {
+  errors: InviteErrors
+}
+
+export type GetOrgsUsersResult =
+  | GetOrgsUsersOKResult
```
randycoulman, line 24: Oh, cool! Does TS allow a leading `|` here with nothing before it?  That's a nice trailing-comma-like feature.

src/unity/thunks/index.ts
```diff
@@ -0,0 +1,127 @@
...
+// Types
+import {Invite, CloudUser as User} from 'src/types'
+
+// Constants
+import {GTM_USER_REMOVED} from 'src/unity/constants'
+
+export const getUsersAndInvites = async (dispatch: Dispatch<Action>) => {
+  try {
+    const userResp = await getOrgsUsers()
+    const inviteResp = await getOrgsInvites()
```
randycoulman, line 36: Should this be `const [userResp, inviteResp] = await Promise.all([getOrgsUsers(), getOrgsInvites()])` to parallelize the two requests?

randycoulman (approved):
This looks like a great start! A few comments below, but nothing major.

src/unity/thunks/index.ts
121watts: Great catch!  I'll change that.

joelfried (approved):

PR #341, opened 2020-11-18
fix: improve metric selector perf

drdelambre:
the parser in giraffe is super slow for high cardinality result sets. instead of relying on that to generate the table schema, i reduced the number of returned tables by joining all of the results together in flux with a `|> group()`, and recreating the schema generation functionality inside of ui.

little bit slower than if we could have grabbed the schema during parsing, but the results are as such:

BEFORE
Screenshot from 2020-11-18 14-07-57

AFTER
Screenshot from 2020-11-18 14-08-56

next bottle neck is that we're rendering too many refs for the individual line items (everything after that vertical cliff), but lets see if this is good enough for now


src/shared/actions/schemaThunks.ts
```diff
@@ -47,25 +46,78 @@ export const fetchSchemaForBucket = async (
...
+    const len = out.length
+    const measurements = out.columns._measurement.data
+    const fields = out.columns._field.data
+    const columns = out.columnKeys.filter(key => {
+      return filtered.reduce((acc, curr) => {
+        return acc && !curr.test(key)
+      }, true)
+    })
+    const colLen = columns.length
+    const schema = {} as any
```
drdelambre, line 81: sorry for all of the anys... needed to be done using the same memory space for large datasets, so i couldn't just copy between types

alexpaxton:
Yoooo `20000ms` to `1000ms` is a pretty massive bump

PR #345, opened 2020-11-19
chore: update giraffe and use name "Band" in plot dropdown

TCL735:
Closes idpe 9144

- update giraffe
- use "Band" instead of "Band Plot" in the dropdown for consistency


hoorayimhelping (approved):
huge

PR #347, opened 2020-11-19
fix(de): add condition to return empty string for empty values

Palakp41:
Closes #19216
The issue of when group  unselection happens, our queryBuidler doesn't update the flux query stored in the state. The same issue was occurring for the selection from filter to group. Now in the video below: we have updated the code where the flux query in the editor corresponds to the exact selections in the ui.

Kapture 2020-11-19 at 14 23 32


Palakp41:
Closing this pull request. Need to take a look again at the bug fix

PR #348, opened 2020-11-20
feat(experiment): add Customer Success link to home & load data

mavarius:
A Google Optimize experiment that adds a Customer Success link to the home page and the Load Data page.

### DIFF
--- a/src/cloud/components/RateLimitAlert.tsx
+++ b/src/cloud/components/RateLimitAlert.tsx
@@ -15,6 +15,8 @@ import {
   BannerPanel,
 } from '@influxdata/clockface'
 import CloudUpgradeButton from 'src/shared/components/CloudUpgradeButton'
+import {GoogleOptimizeExperiment} from './experiments/GoogleOptimizeExperiment'
+import {CustomerSuccessLinkHeader} from './experiments/variants/CustomerSuccessLinkHeader'
 
 // Utils
 import {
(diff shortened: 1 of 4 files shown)


src/cloud/components/experiments/variants/CustomerSuccessLinkPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+      </Panel.Header>
+      <Panel.Body>
+        <LinkButton
+          style={{textAlign: 'center'}}
+          className="home-page--contact-button"
+          color={ComponentColor.Primary}
+          shape={ButtonShape.Default}
+          size={ComponentSize.Small}
+          text="Speak with an Expert"
+          href="https://calendly.com/c/EDCUW4IQ7Z4ZWYSB"
```
mnorriswalker, line 27: Should we put the name of the location of the test into the Calendly Event Header so that when it gets booked into their calendars the onboarding team can see the source? Ror example: Speak with InfluxDB Cloud Expert (Load Data)

mnorriswalker (approved):

src/cloud/components/experiments/variants/CustomerSuccessLinkPanel.tsx
mavarius: There is now a seperate event for home page vs load data page.

src/cloud/components/RateLimitAlert.tsx
```diff
@@ -15,6 +15,8 @@ import {
   BannerPanel,
 } from '@influxdata/clockface'
 import CloudUpgradeButton from 'src/shared/components/CloudUpgradeButton'
+import {GoogleOptimizeExperiment} from './experiments/GoogleOptimizeExperiment'
```
drdelambre, line 18: please use aliasing and /or absolute paths

drdelambre (approved):

PR #353, opened 2020-11-20
fix: perist metrics on refresh

drdelambre:
for some reason, refreshing the notebooks page would make the metric selector clear the filters you put on there. undid that.


alexpaxton (approved):
Terminal!

PR #365, opened 2020-11-25
fix(notificationRule): add operator dropdown value for TagRuleDropdown

Palakp41:
Closes EAR#1932

TagRuleDropdown didn't include `notequal` text in the dropdown. So, when an api is used to update notificationRule to have a tagRule with `notEqual`, ui was erroring out because the dropdown couldn't find `!=` operator in the operators array.

- [x] Well-formatted commit messages
- [x] Rebased/mergeable
- [x] Tests pass


hoorayimhelping (approved):
this one line represents like two days of digging into things. excellent work!

PR #369, opened 2020-11-30
fix: trying to fix race issue

drdelambre:


121watts:
this was resolved with another fix

PR #370, opened 2020-11-30
chore(notebooks): metric selector ux polish

alexpaxton:
Closes #

- Use proper case in searchbar placeholder text
- Show human friendly error messages for aggregate type errors
  - Using a fragile string match implementation for now, as they say good enough for government
  - Will need to refactor once flux sends back error codes alongside messages
- Add button in index view that links to feedback form
- Anytime the user changes which aggregate function is selected, the notebook re-runs

- [x] Rebased/mergeable
- [x] Tests pass


src/flows/shared/Resizer.tsx
```diff
@@ -157,7 +170,11 @@ const Resizer: FC<Props> = ({
   let body = children
 
   if (error) {
-    body = <div className="panel-resizer--error">{error}</div>
+    body = (
+      <div className="panel-resizer--error">
+        {makeErrorHumanFriendly(error, selectedFunction.name)}
```
Asalem1, line 175: should we make this actionable like the other aggregate type notification we return in the data explorer? Like, rather than telling the user what to do - shouldn't we render a bottom for them to click that'll do that for them?

alexpaxton: That is a great idea. I'll add that in

src/flows/components/FlowsIndex.tsx
```diff
@@ -12,6 +12,13 @@ import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
 import {PROJECT_NAME_PLURAL} from 'src/flows'
 
 const FlowsIndex = () => {
+  const handleFeedbackClick = (): void => {
+    window.open(
+      'https://w2.influxdata.com/feedback-notebooks-early-access/',
```
Asalem1, line 17: is this something we want to exist elsewhere as a const? Does that matter?

src/flows/components/header/Submit.tsx
```diff
@@ -60,11 +60,21 @@ export const Submit: FC = () => {
       .length
   }, [flow.data])
 
+  // Returns a string to avoid array deep comparison woes
+  // Triggers a re-run any time any aggregate function is changed
+  const aggregateOfAggregates = useMemo(() => {
```
Asalem1, line 65: Would love to hear @drdelambre 's opinion on this

Asalem1 (commented):
Looks great to me. Would love @drdelambre's opinion on this so that we can get this across the line

src/flows/components/FlowsIndex.tsx
alexpaxton: Probs more semantic to just use an anchor tag, but ran into some styling jank and this was the path of least resistance. This button will only exist so long as we're in early access so probs not something that needs more permanence

Asalem1: Gotcha. Totally chill with leaving it there if it's gonna get cut out later

src/flows/components/header/Submit.tsx
drdelambre: so a notebook is a collection of pipes, and to keep everything loosely coupled, we need to ensure that one pipe type has no concept of the internal logic of any other pipe type. as an antithesis: if Pipe A relies on Pipe B having a function in it that transforms a string into a number, then anytime we change Pipe B, we must also ensure that Pipe A still works, and we wont be able to delete Pipe B without rewriting Pipe A. within notebooks, a pipe shouldn't even know that there exists a list of pipes somewhere, it should think of itself as an island at the center of a black hole of loneliness. The notebook itself should know that there is a list of pipes, but shouldn't expect there to be a difference between the implementation of Pipe A vs Pipe B vs... they're all just pipes.

A good rule of thumb for testing coupling is that, in a pure absolutist fantasy, you should be able to delete any module without having to change anything. just go nuclear, delete the file, and the app should keep running just fine.

this is the ideal to chase, but usually some concessions must be made.

the only sore point i might see in this one is that it forces the notebook to know that there is a list of pipes, that there is a pipe of type `metricSelector` (all pipes have the property `type`, but we shouldn't expect any particular value to exist in that space), and that a pipe of type metricSelector implements a property of `agregateFunction`. To decouple this, i would probably flip the filter an map to look like:
```
flow.data.all
  .map(p => p.aggregateFunction?.name)
  .filter(p => !!p)
  .join('')
```

then it doesn't expect any implementation to exist from a pipe, but instead uses any aggregate function that is defined.

drdelambre (approved):

PR #372, opened 2020-11-30
fix: locking idpe to Nov10th until resolved

drdelambre:
should unblock all the things. also makes sure that firefox is running the same image of the current branch as the chrome version


121watts (approved):
merge when green

PR #381, opened 2020-12-01
fix(experiment): restrict to Load Data page

mavarius:
This is a fix to prevent the header experiment from showing on any page except the Load Data page.


drdelambre:
lame that we can't make that a parameter on the experiment from google's side

drdelambre (approved):

mavarius:
We can have targeting rules, and originally did but it only works for changes made through the Optimize interface.

PR #383, opened 2020-12-02
chore(dependency): upgrade clockface to 2.3.7

alexpaxton:
Needed to unblock notebooks early access work 

- [x] Rebased/mergeable
- [x] Tests pass


hoorayimhelping (changes requested):
Is changing Flows to Notebooks required to implement this update of Clockface? i'd prefer to keep these as two separate PRs, since they seem like logically different issues, and for posterity, and potential bug fixing / test breaking issues, it's helpful to separate these things out. We've had situations where we combine code in PRs, then a library update breaks a test and we revert the whole PR, and people are confused as to why the the old behavior is there (in this case, why are Notebooks still called Flows?)

alexpaxton:
@hoorayimhelping I've made a ton of PRs that include both clockface upgrades and other changes -- not sure why this is being treated any differently. As a clean code practice I agree with you but in this specific case there are no breaking changes in the new version of clockface, and the non-dependency changes are purely presentational. Feels like splitting hairs to make this into 2 PRs

hoorayimhelping:
>I've made a ton of PRs that include both clockface upgrades and other changes -- not sure why this is being treated any differently. 

I get it, but we have to start somewhere. I've been making this request to many people over the past few months - Timmy has heard it a few times with Giraffe stuff. I understand that it's annoying to have to start doing it, but like you said, it's a cleaner from a code perspective and we have to start somewhere.

I'm sorry to disrupt your workflow. I know that it's less risky, but part of being consistent and disciplined about process is following it even when it feels a little bit like overkill. If follow it selectively, it's too easy to end up slipping up and not following it when it matters.

hoorayimhelping (approved):

PR #391, opened 2020-12-04
fix: close the task on completion

drdelambre:
close a task when you export it in flows


drdelambre:
paxton rolled this into another PR

PR #396, opened 2020-12-04
feat: fix label popup close on esc key 

jrenee42:
Closes https://github.com/influxdata/influxdb/issues/17853

fix for popups not closing when esc is pressed.  

clockface will close the popover on the esc key if it is not controlled; but in this case it is controlled by the InlineLabelPopover.

 since the input takes all the focus and all the keys, listen on the input for the escape,  and the container of the inline labelpopover controls the popover visible state, so added another handler there to hide it.


TCL735 (approved):

TCL735:
@jrenee42 I approved this. You might need to rebase this with master branch (again) to get the ci checks to pass.

joelfried (approved):

PR #397, opened 2020-12-04
fix: dont use stale flow data

drdelambre:
found an error on the demo in which data was "being lost" when changing the query period dropdown. seems to be that we were reapplying stale state


alexpaxton (approved):
Nais

PR #398, opened 2020-12-04
fix: less aggressive whitespace trimming

### LINKED ISSUE #395: Notebooks: Flux map not parsed correctly
When using `map()` in a Flux script cell, it appears spaces in the mapped key value pairs are removed, causing the results to be incorrect.

image

**Steps to reproduce:**

1. Build a basic Flux query
2. Add a new Flux script cell
3. Add the following Flux:
    ```js
    __PREVIOUS_RESULT__
      |> map(fn: (r) => ({ r with _value: r._value * 100.0 }))
    ```

**Expected behavior:**
This `map()` call should just update the existing `_value` column and leave all other columns intact.

**Actual behavior:**
This actually returns data with only a `rwith_value` column. This indicates that spaces are being removed from the map call before the script is sent to the Flux query engine.

**Environment info:**

- In quartz.

drdelambre:
Closes #395

### DIFF
--- a/src/flows/pipes/RawFluxEditor/index.ts
+++ b/src/flows/pipes/RawFluxEditor/index.ts
@@ -28,7 +28,7 @@ export default register => {
     generateFlux: (pipe, create, append) => {
       const text = pipe.queries[pipe.activeQuery].text
         .replace(COMMENT_REMOVER, '')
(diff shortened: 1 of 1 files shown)


alexpaxton (approved):
Sweeeeet, good catch

PR #399, opened 2020-12-04
test: refactor fridays; added a test and comments (code is clearer, renamed a method for clarity)

jrenee42:
refactor fridays.  part of the tokentab ecosystem

### DIFF
--- a/src/authorizations/utils/permissions.test.ts
+++ b/src/authorizations/utils/permissions.test.ts
@@ -1,4 +1,4 @@
-import {allAccessPermissions} from './permissions'
+import {allAccessPermissions, toggleSelectedBucket} from './permissions'
 import {Permission} from 'src/types'
 
 const hvhs: Permission[] = [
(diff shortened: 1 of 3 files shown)


TCL735:
@jrenee42 @Palakp41 - i did a force push to rebase it with master so that the ci checks can pass. They seem to pass when rebased with master for other PRs.

src/authorizations/utils/permissions.test.ts
```diff
@@ -252,3 +252,17 @@ const hvhs: Permission[] = [
 test('all-access tokens/authorizations production test', () => {
   expect(allAccessPermissions('bulldogs', 'mario')).toMatchObject(hvhs)
 })
+
+test('toggleSelectedBucket test', () => {
+  const myBuckets: string[] = ['bucket1', 'bucket2', 'bucket3']
+
+  expect(toggleSelectedBucket('bucket4', myBuckets)).toEqual([
+    ...myBuckets,
```
Palakp41, line 260: I like these tests! new use of spread operators. Small and tidy :)

src/authorizations/utils/permissions.ts
```diff
@@ -94,6 +94,8 @@ export const allAccessPermissions = (
   return allPermissionTypes.flatMap(withOrgID)
 }
 
+// add a permission string ('read' or 'write' is the second argument)
```
Palakp41, line 97: (nit): should this comment be here?

src/authorizations/utils/permissions.ts
```diff
@@ -110,6 +112,8 @@ export const specificBucketsPermissions = (
   })
 }
 
+// assign permission string ('read' or 'write' is the second argument)
```
Palakp41, line 115: (nit): should we leave this comments here?

Palakp41 (approved):
Looks good to me

src/authorizations/utils/permissions.ts
TCL735: I agree with @Palakp41 that we do not need comments in here. But i will defer to @jrenee42 on this because she said it made it easier for her to understand.

src/authorizations/utils/permissions.ts
TCL735: I agree with @Palakp41 that we do not need comments in here. But i will defer to @jrenee42 on this because she said it made it easier for her to understand.

PR #401, opened 2020-12-04
chore: Refactor Friday! Refactor the me reducer

joelfried:
Refactored the me reducer:

- Moved from shared folder to src/me/reducers and cleaned up other paths as needed.
- Added a test for the me reducer.
- Added a type for me.


121watts (approved):

src/me/reducers/index.test.ts
```diff
@@ -0,0 +1,17 @@
...
+// Actions
+import {setMe} from 'src/me/actions/creators'
+
+// Mocks
+import {me} from 'src/me/mockUserData'
+
+describe('me reducer', () => {
+  it('can set me', () => {
+    const expected = me
+    const actual = reducer(undefined, setMe(expected))
```
hoorayimhelping, line 13: shouldn't this be `me` getting passed in and not `expected`?

joelfried: It doesn't make a difference in this particular case but I like that better too.  That way both expected and actual start out with the data form the import something is done to them (even though in expected's case it's very simple) and compared at the end.

Changed!

PR #426, opened 2020-12-11
fix: handle Infinity return value from window period function

### LINKED ISSUE #422: Using a zero or negative range for a query causes the UI to generate an invalid Flux AST
**Steps to reproduce:**

Reproduction behavior is detailed in influxdata/flux#3014.
It only happens in the UI.

**Expected behavior:**
An error message from Flux saying that the range was empty.
The JSON in the extern that is submitted to Flux should have an integer value instead of `null`.

**Actual behavior:**
`compilation failed: could not get handle from JSON AST: invalid type: null, expected i64 at line 1 column 857`

This seems to be caused because the UI generations `{"type":"DurationLiteral","values":[{"magnitude":null,"unit":"ms"}]}`.

ShmuelLotman:
Closes #422 

Previously, when the time range for the flux query was identical in start/stop time, or the stop time was before the start time, the query would fail with a 500. This is because the windowPeriod retrieval function would return a value of `Infinity`, which our code wasn't handling. This PR addresses that, while also removing a heinous quote mark 

Some before and after:
image
image


121watts (approved):
Nice catch!

src/variables/utils/getWindowVars.test.ts
```diff
@@ -21,6 +21,26 @@ describe('getWindowPeriodVariable', () => {
...
+
+  test('should return null when timeRange is negative', () => {
+    const query = `from(bucket: "Go Pack Go")
+    |> range(start: 2020-12-10T17:00:00Z, stop: 2020-12-09T17:00:00Z)
+    |> filter(fn: (r) => r["_measurement"] == "query_request_duration")
+    |> filter(fn: (r) => r["_field"] == "success_rate")`
+
+    const actual = getWindowPeriodVariable(query, defaultVariableAssignments)
+    expect(actual).toEqual(null)
+  })
```
hoorayimhelping, line 42: do these two cases trigger `Infinity` to be returned?

src/variables/utils/getWindowVars.test.ts
```diff
@@ -21,6 +21,26 @@ describe('getWindowPeriodVariable', () => {
     expect(actual).toEqual(null)
   })
 
+  test('should return null when timeRange is identical', () => {
```
hoorayimhelping, line 24: one thing you might do for SUPER EXTRA PROTECTION AND DOCUMENTATION™ is to add a comment linking to the issue this test is guarding against. so if i'm bumbling around in tests and i see this one and i see a link to a bug, i can get more context if i'm so inclined.

hoorayimhelping (approved):
nice

src/variables/utils/getWindowVars.test.ts
ShmuelLotman: Yes, hence the tests. Before, that was causing the error to be returned ultimately.

src/variables/utils/getWindowVars.test.ts
ShmuelLotman: Your wish is my command, and so it is done

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #431, opened 2020-12-14
feat: makes CSV times more deterministic

### LINKED ISSUE #428: Make Line Protocol Deterministic
### Problem

Line protocol requires users to explicitly set the precision of the timestamp that the data is formatted in in order to upload line protocol. This might lead to misconfigurations if the user:

1. Doesn't know what the line protocol configuration time is 
2. Doesn't realize they need to make a selection when uploading line protocol

### Solution

The UI shouldn't have to do that, but we should probably make this process determinstic

Asalem1:
Closes #428

### Problem

The current CSV uploader is set to the assumption that the incoming CSVs are Ms formatted times rather than basing it off of something more determinsitic

### Solution

Calculate the time based on length to determine whether the time is `ms`, `s`, `us`, or `ns`

### DIFF
--- a/src/buckets/components/context/csvUploaderProvider.tsx
+++ b/src/buckets/components/context/csvUploaderProvider.tsx
@@ -42,7 +42,7 @@ const MAX_CHUNK_SIZE = 1750
 
 export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
   const [value, setValue] = useState(0)
-  const {bucketID, orgID} = useParams()
+  const {bucketID} = useParams()
   const [hasFile, setHasFile] = useState(false)
   const [uploadFinished, setUploadFinished] = useState(false)
 
@@ -50,13 +50,13 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
     useSelector((state: AppState) =>
       getByID<Bucket>(state, ResourceType.Buckets, bucketID)
     )?.name ?? ''
-  const org = useSelector(getOrg).name
+  const org = useSelector(getOrg)
 
   const history = useHistory()
 
   const handleDismiss = useCallback(() => {
-    history.push(`/orgs/${orgID}/load-data/buckets`)
-  }, [history, orgID])
+    history.push(`/orgs/${org.id}/load-data/buckets`)
+  }, [history, org.id])
 
   const handleDrop = useCallback(
     (csv: string) => {
@@ -116,7 +116,7 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
             }
             const resp = postWrite({
               data: chunk,
-              query: {org, bucket, precision},
+              query: {org: org.name, bucket, precision},
             }).then(() => {
               const percent = (++progress / counter) * 100
               setValue(Math.floor(percent))
@@ -156,7 +156,7 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
           }
           const resp = postWrite({
             data: chunk,
-            query: {org, bucket, precision},
+            query: {org: org.name, bucket, precision},
           }).then(() => {
             const percent = (++progress / counter) * 100
             setValue(Math.floor(percent))
--- a/src/buckets/components/context/csvUploaderProvider.tsx
+++ b/src/buckets/components/context/csvUploaderProvider.tsx
@@ -1,7 +1,7 @@
 // Libraries
 import React, {FC, useCallback, useState} from 'react'
 import {fromFlux} from '@influxdata/giraffe'
-import {useHistory, useParams} from 'react-router-dom'
+import {useParams} from 'react-router-dom'
 import {useSelector} from 'react-redux'
 
 // Utils
@@ -19,19 +19,17 @@ export type Props = {
 }
 
 export interface CsvUploaderContextType {
-  handleDismiss: () => void
-  handleDrop: (csv: string) => void
   hasFile: boolean
+  progress: number
+  uploadCsv: (csv: string) => void
   uploadFinished: boolean
-  value: number
 }
 
 export const DEFAULT_CONTEXT: CsvUploaderContextType = {
-  handleDismiss: () => {},
-  handleDrop: (_: string) => {},
   hasFile: false,
+  progress: 0,
+  uploadCsv: (_: string) => {},
   uploadFinished: false,
-  value: 0,
 }
 
 export const CsvUploaderContext = React.createContext<CsvUploaderContextType>(
@@ -41,7 +39,7 @@ export const CsvUploaderContext = React.createContext<CsvUploaderContextType>(
 const MAX_CHUNK_SIZE = 1750
 
 export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
-  const [value, setValue] = useState(0)
+  const [progress, setProgress] = useState(0)
   const {bucketID} = useParams()
   const [hasFile, setHasFile] = useState(false)
   const [uploadFinished, setUploadFinished] = useState(false)
@@ -52,13 +50,12 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
     )?.name ?? ''
   const org = useSelector(getOrg)
 
-  const history = useHistory()
+  const normalizeTimes = (time: number): BigInt => {
+    const t = `${time}` + Array(19 - `${time}`.length + 1).join('0')
+    return BigInt(t)
+  }
 
-  const handleDismiss = useCallback(() => {
-    history.push(`/orgs/${org.id}/load-data/buckets`)
-  }, [history, org.id])
-
-  const handleDrop = useCallback(
+  const uploadCsv = useCallback(
     (csv: string) => {
       setHasFile(true)
       setTimeout(() => {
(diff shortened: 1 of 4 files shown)


src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+export const CsvUploaderContext = React.createContext<CsvUploaderContextType>(
+  DEFAULT_CONTEXT
+)
+
+const MAX_CHUNK_SIZE = 1750
+
+export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
+  const [value, setValue] = useState(0)
+  const {bucketID, orgID} = useParams()
```
drdelambre, line 45: grab orgID from the redux state. it's better

Asalem1: Why's it better?

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const [hasFile, setHasFile] = useState(false)
+  const [uploadFinished, setUploadFinished] = useState(false)
+
+  const bucket =
+    useSelector((state: AppState) =>
+      getByID<Bucket>(state, ResourceType.Buckets, bucketID)
+    )?.name ?? ''
+  const org = useSelector(getOrg).name
```
drdelambre, line 53: what. just have org be the org object. you have 3 different versions of "what is the current org" in one component

Asalem1: Makes sense

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  const history = useHistory()
+
+  const handleDismiss = useCallback(() => {
+    history.push(`/orgs/${orgID}/load-data/buckets`)
+  }, [history, orgID])
```
drdelambre, line 59: org.id

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  const bucket =
+    useSelector((state: AppState) =>
+      getByID<Bucket>(state, ResourceType.Buckets, bucketID)
+    )?.name ?? ''
+  const org = useSelector(getOrg).name
+
+  const history = useHistory()
+
+  const handleDismiss = useCallback(() => {
```
drdelambre, line 57: it'd be better not to use any route based eventing and instead use a popup controller

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+          measurement = table.getColumn('_measurement')[i]
+          field = table.getColumn('_field')[i]
+          time = table.getColumn('_time')[i]
+          tags = columns
+            .filter(col => !!table.getColumn(col)[i])
+            .map(col => `${col}=${table.getColumn(col)[i]}`)
+            .join(',')
+            .trim()
+            .replace(/(\r\n|\n|\r)/gm, '')
+          line = `${measurement},${tags} ${field}="${field}" ${time}`
```
drdelambre, line 137: does the spec allow for multiple time values to come down from queries?

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+        let counter = 0
+        let progress = 0
+
+        const pendingWrites = []
+        let precision = WritePrecision.Ms
+
+        for (let i = 0; i < length; i++) {
+          if (i !== 0 && i % MAX_CHUNK_SIZE === 0) {
+            // second range
+            if (`${time}`.length === 10) {
```
drdelambre, line 102: why are you expecting time to survive the loop like this?

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+            // millisecond range
+            if (`${time}`.length === 13) {
+              precision = WritePrecision.Ms
+            }
+            // microsecond range
+            if (`${time}`.length === 16) {
+              precision = WritePrecision.Us
+            }
+            // nanosecond range
+            if (`${time}`.length === 19) {
```
drdelambre, line 114: lot of copy pasta being defined within a loo[

src/buckets/components/context/csvUploaderProvider.tsx
drdelambre, line 137: instead of all this, normalize the time value to nano seconds instead, then always send nanoseconds in the write command
``` line = `${measurement},${tags} ${field}="${field}" ${normalize(time)}` ```

and have normalize do something like ` (time) => time + Array(13 - time.length + 1).join('0')`

File: src/buckets/components/context/csvUploaderProvider.tsx [hunk 8]
```diff
@@ -0,0 +1,191 @@
...
+  )
+
+  return (
+    <CsvUploaderContext.Provider
+      value={{
+        handleDismiss,
+        handleDrop,
+        hasFile,
+        uploadFinished,
+        value,
```
drdelambre, line 183: make this progress, `value` is kinda meaningless

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+      }, 0)
+    },
+    [bucket, org]
+  )
+
+  return (
+    <CsvUploaderContext.Provider
+      value={{
+        handleDismiss,
+        handleDrop,
```
drdelambre, line 180: drop only makes sense in the context of a view layer that implements drag and drop, where this is more of a `uploadCSV` function

src/buckets/components/context/csvUploaderProvider.tsx
Asalem1: No

src/buckets/components/context/csvUploaderProvider.tsx
Asalem1: I'm not. i'm basing it off the assumption that the incoming CSV should be homogenized, but I can rework that assumption

src/buckets/components/context/csvUploaderProvider.tsx
Asalem1: Totally, but better than creating a new function am i right?

src/buckets/components/context/csvUploaderProvider.tsx
Asalem1: That's a great idea

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,191 @@
...
+        })
+      }, 0)
+    },
+    [bucket, org]
+  )
+
+  return (
+    <CsvUploaderContext.Provider
+      value={{
+        handleDismiss,
```
drdelambre, line 179: handleDismiss is a view layer concern, it should be defined there

src/buckets/components/context/csvUploaderProvider.tsx
drdelambre: see the next comment about "what is the current org". url params should be triggers and not statefull stores

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -0,0 +1,156 @@
...
+  DEFAULT_CONTEXT
+)
+
+const MAX_CHUNK_SIZE = 1750
+
+export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
+  const [progress, setProgress] = useState(0)
+  const {bucketID} = useParams()
+  const [hasFile, setHasFile] = useState(false)
+  const [uploadFinished, setUploadFinished] = useState(false)
```
drdelambre, line 45: not a big deal, but isn't hasFile and uploadFinished the same as RemoteDataState (loading and done respectively). could join those two if it makes sense

drdelambre (approved):

src/buckets/components/context/csvUploaderProvider.tsx
Asalem1: It'd be more apt to say `NotStarted` and `Done` but considering it's a binary outcome, might be quicker to use a bool rather than an enum?

drdelambre: hasFile = false => notstarted
hasFile = true => loading
uploadFinished = true => done

Asalem1: Aw, I see what you're saying. Totally can consolidate those

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -40,19 +41,23 @@ const MAX_CHUNK_SIZE = 1750
 
 export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
   const [progress, setProgress] = useState(0)
-  const {bucketID} = useParams()
+  const [bucketID, setBucket] = useState('')
```
drdelambre, line 48: you're making new sources of truth again, this time around "what is the currently selected bucket". this is bad. i'd either figure out why useParam didn't work, or pass it in as a function parameter for uploadCsv, so that you can fetch it someplace else.

Asalem1: I'm pretty sure the reason it didn't work was because we were calling useParams in the provider, which is wrapping the route. We could nest things down?

PR #436, opened 2020-12-15
chore: fix breaking change from giraffe 2.0.0

hoorayimhelping:
Moves types that were previously in Giraffe into ui

- [x] Rebased/mergeable
- [x] Tests pass


121watts (approved):

Asalem1 (approved):
Good catch on the TS exports

PR #438, opened 2020-12-15
feat(notebooks): enable inline buckets creation

### LINKED ISSUE #390: Notebooks: Add a bucket from dropdown
when i'm adding a task from a notebook, i should be able to create a bucket from the dropdown so that i dont have to go find where buckets are created.

as a first time user, i only have the bucket that has my data in it and it is unclear what the interaction is supposed to do if i can only put the data back into the same bucket

alexpaxton:
Closes #390 

- Allow users to create a bucket from the `BucketSelector` component used in the metric selector and output to bucket cells
- Show the asset limit overlay if the user has no more available buckets
- Automatically select the recently created bucket in the dropdown

create-bucket-inline

- [x] Rebased/mergeable
- [ ] Tests pass


src/buckets/components/CreateBucketOverlay.tsx
```diff
@@ -24,19 +27,15 @@ import {
...
-type Props = OwnProps & ReduxProps
+type Props = ReduxProps
 
 const CreateBucketOverlay: FC<Props> = ({
   org,
   isRetentionLimitEnforced,
   createBucket,
-  onClose,
 }) => {
+  const {onClose} = useContext(OverlayContext)
```
drdelambre, line 38: is this available for the existing paradigm?

src/flows/shared/BucketSelector/index.tsx
```diff
@@ -44,6 +65,8 @@ const BucketSelector: FC = () => {
   if (loading === RemoteDataState.Done && buckets.length) {
     menuItems = (
       <>
+        <CreateBucketDropdownItem onClick={handleCreateNewBucket} />
```
drdelambre, line 68: why is this something special and not another Dropdown.Item?

src/buckets/components/CreateBucketDropdownItem.tsx
```diff
@@ -0,0 +1,74 @@
+// Libraries
+import React, {FC, useEffect} from 'react'
```
drdelambre, line 2: this component has a lot of presumptions about the context it's initialized in, and none of that context has to do with src/buckets

src/buckets/components/CreateBucketDropdownItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+import {connect, ConnectedProps, useDispatch} from 'react-redux'
```
drdelambre, line 3: use hooks and not connect

src/flows/shared/utils.ts
```diff
@@ -1,6 +1,34 @@
+import {Bucket} from 'src/types'
+
 export const formatQueryText = (queryText: string): string => {
   return (queryText || '')
     .trim()
     .split('|>')
     .join('\n  |>')
 }
+
+export const getNewestBucket = (
```
drdelambre, line 10: this just signifies that the eventing isn't being passed around correctly right?

src/buckets/components/CreateBucketOverlay.tsx
alexpaxton: Yop yop yop

src/flows/shared/BucketSelector/index.tsx
alexpaxton: It had a lot of things going on specifically for it, and it made the parent component feel messy so I isolated it

src/buckets/actions/thunks.ts
```diff
@@ -112,6 +115,8 @@ export const createBucket = (bucket: OwnBucket) => async (
 
     dispatch(addBucket(newBucket))
     dispatch(checkBucketLimits())
+
+    callback && callback(resp.data as Bucket)
```
drdelambre, line 119: how come a callback and not stringing together a bunch of awaits like:
```
await dispatch(addBucket(newBucket))
await dispatch(checkBucketLimits())
return newBucket
```

you also most def want to use the normalized bucket and not resp.data

src/overlays/reducers/overlays.ts
```diff
@@ -20,7 +20,7 @@ export type OverlayID =
   | 'edit-annotation'
 
 export interface OverlayParams {
-  [key: string]: string
+  [key: string]: any
```
alexpaxton, line 23: This feels weird but not sure how else to deal with it

src/buckets/actions/thunks.ts
```diff
@@ -92,7 +92,10 @@ export const getBuckets = () => async (
   }
 }
 
-export const createBucket = (bucket: OwnBucket) => async (
+export const createBucket = (
+  bucket: OwnBucket,
+  callback?: (bucket: Bucket) => void
```
121watts, line 97: create bucket does one thing really well.  let's not add an optional callback to get the data from this.

121watts: bucket names are unique and you know the name of the bucket since you passed it in.  why not retrieve the bucket that way?

src/cloud/utils/limits.ts
```diff
@@ -10,6 +11,11 @@ export const isLimitError = (error): boolean => {
 export const extractBucketLimits = (limits: LimitsState): LimitStatus => {
   return get(limits, 'buckets.limitStatus')
 }
+
+export const getBucketLimitStatus = (state: AppState): LimitStatus => {
```
drdelambre, line 15: it's strange that this is scoped to app state, where as the other selectors are scoped to the subset of LimitState. I'm totally for using AppState (i view too large of a global namespace as a code smell), but you might want to reach out to watts to see why that jailing was made in the first place.

src/buckets/actions/thunks.ts
alexpaxton: @121watts suggested not returning a bucket from the thunk, so I opted for a generic callback approach.

At first I was using the normalized bucket but every time I tried to extract the `bucket` property from it, it returned `undefined` so I gave up on that

src/cloud/utils/limits.ts
alexpaxton: Everywhere the limit selectors are used they are passed `state.cloud.limits` which I could not use with the `useSelector` hook

src/flows/shared/BucketSelector/CreateBucketDropdownItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+interface Props {
+  onCreateBucket: (bucket: Bucket) => void
+}
+
+const CreateBucketDropdownItem: FC<Props> = ({onCreateBucket}) => {
+  const limitStatus = useSelector(getBucketLimitStatus)
+  const dispatch = useDispatch()
+  useEffect(() => {
+    // Check bucket limits when component mounts
+    dispatch(checkBucketLimits())
```
drdelambre, line 30: is there some kind of caching behind this? You should totally be able to call it like this in theory, but it doesn't feel right to lean on the convention that this will only render once. it's always one dom level change from exploding a bunch of network requests, like a bomb waiting for some misplaced input field to set it off

src/buckets/actions/thunks.ts
drdelambre: what was the reasoning behind not returning the bucket?

you should totally revisit the normalized form.

 after chasing around a lot of "sometimes this returns null instead of a string" bugs, we dropped that layer in to decouple the frontend from the api, and lean pretty heavy on it. normalizing also creates a guarantee that, if you start passing objects around in the view layer, the next component that tries to reference the state will have the same data. imagine: you take resp.data, and it has a property `name` that you use. a month later, you see that code in the view layer, and try to make a selector that looks up the bucket by name in the redux resources in order to grab it's id. except the normalized form doesn't use `name` anymore, it uses `display` for some reason. now it's all broken because we've created two definitions of what a bucket is in two places.
I wish this was just an esoteric thought experiment, but creating multiple sources of truth is at the core of a lot of our reusability / defect issues within the current system

src/cloud/utils/limits.ts
drdelambre: i think you're missing the point here. everywhere you see those existing selectors, someone passed `state.cloud.limits` into it right? why?

src/flows/shared/BucketSelector/CreateBucketDropdownItem.tsx
alexpaxton: I didn't write it, but I think it's some insurance that bucket limits exist before trying to render this component

alexpaxton: Open to suggestions here

src/buckets/components/CreateBucketOverlay.tsx
```diff
@@ -101,22 +100,4 @@ const CreateBucketOverlay: FC<Props> = ({
...
-  }
-}
-
-const mdtp = {
-  createBucket,
-}
-
-const connector = connect(mstp, mdtp)
-
-export default connector(CreateBucketOverlay)
```
121watts, line 122: nice

121watts (approved):
I would love to see an e2e test for this

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | 22 | :red_circle:

drdelambre (approved):

PR #445, opened 2020-12-17
chore: remove schema reducer in favor of the query cache

### LINKED ISSUE #317: Remove schema reducer
currently the metric selector queries the db, transforms the results, and drops that transformation into a reducer. we should lean in on the query cache handling these kinds of things instead of redux so that we can leverage the mutex queuing on page.

since the query cache only deals with flux results, we will need to move that transformation piece to the other side of the state management to look more like a selector, where the required data format is transformed by the view layer before being used, as opposed to the current implementation that does the transformation before setting it into redux to double dip on the action and use redux as a memoization layer for that transformation

this also removes any source of truth for the query's loading state, so we should use this as an opportunity to add that to the query caching layer instead of keeping it in a context / reducer / view layer state

Asalem1:
Closes #317 

This PR removes the schema reducer and leans into the queryCache as a means of consolidating some of our existing caches. This is part of an ongoing process to lean more heavily upon existing infrastructure and consolidate repetitively functionality. In doing so, a few updates were made to the queryCache, like adding the `status` (RemoteDataState) and `error` state. Although these aren't currently used in the implementation, they will be an important piece to lean up when consolidating and expanding the usage of the queryCache.


src/flows/components/panel/Results.tsx
```diff
@@ -29,7 +29,8 @@ const Results: FC = () => {
   const resultsExist =
     !!results && !!results.raw && !!results.parsed.table.length
 
-  const rows = useMemo(() => results.raw.split('\n'), [results.raw])
+  const rows = useMemo(() => results?.raw?.split('\n') ?? '', [results?.raw])
```
Asalem1, line 32: added this in since running an empty query was crashing the app

src/flows/context/query.tsx
```diff
@@ -41,44 +42,6 @@ export const QueryContext = React.createContext<QueryContextType>(
 
 const PREVIOUS_REGEXP = /__PREVIOUS_RESULT__/g
 
-const findOrgID = (text, buckets) => {
```
Asalem1, line 44: moved this to a central locale to be used for the metric selector as well

src/flows/context/schemaProvider.tsx
```diff
@@ -69,22 +130,41 @@ export const SchemaProvider: FC<Props> = React.memo(({children}) => {
...
+    const orgID = findOrgID(query, [data.bucket])
+    const result = getCachedResultsOrRunQuery(orgID, query, [])
+    result.promise
+      .then((res: RunQueryResult) => {
+        const schemaForBucket = csvToSchema(res)
+        setSchema(schemaForBucket)
+        setLoading(RemoteDataState.Done)
+      })
+      .catch((error: RunQueryResult) => {
+        console.error('error: ', error)
```
Asalem1, line 156: wasn't quite sure what to do if there was an error here

src/shared/apis/queryCache.ts
```diff
@@ -44,8 +54,22 @@ export const hashCode = (rawText: string): string => {
...
 
+type CacheValue = {
+  dateSet: number
+  hashedVariables: string
+  isCustomTime: boolean
+  mutex: ReturnType<typeof RunQueryPromiseMutex>
+  status: RemoteDataState
+  error?: string
+  values?: RunQuerySuccessResult | null
+}
```
Asalem1, line 65: As mentioned in the PR comments, I added `status` and `error` here in anticipation of future work. Both are being set in this PR but have no ramifications on the view layer

src/shared/apis/queryCache.ts
```diff
@@ -155,11 +191,10 @@ const hasWindowVars = (variables: VariableAssignment[]): boolean =>
 export const getCachedResultsOrRunQuery = (
   orgID: string,
   query: string,
-  state: AppState
+  usedVars: Variable[]
 ): CancelBox<RunQueryResult> => {
   const queryID = `${hashCode(query)}`
   event('Starting Query Cache Process ', {context: 'queryCache', queryID})
-  const usedVars = filterUnusedVarsBasedOnQuery(getAllVariables(state), [query])
```
Asalem1, line 162: made this a parent concern to make the function more isolated

src/shared/apis/queryCache.ts
```diff
@@ -192,21 +227,30 @@ export const getCachedResultsOrRunQuery = (
...
+      // TODO(ariel): handle custom time range
+      // if the timeRange is non-relative (i.e. a custom timeRange or the query text has a set time range)
+      // we will need to pass an additional parameter to ensure that the cached data is treated differently
+      // set the resolved promise results in the cache
+      queryCache.setCacheByID(queryID, hashedVariables, res)
+      // non-variable start / stop should
+      return res
+    })
+    .catch((error: RunQueryErrorResult | RunQueryLimitResult) => {
+      queryCache.setCacheByID(queryID, hashedVariables, error)
```
Asalem1, line 241: added in that we're setting the error

src/shared/utils/flowSchemaNormalizer.ts
```diff
@@ -65,6 +65,13 @@ export const normalizeSchema = (
   data: PipeData,
   searchTerm: string
 ): NormalizedSchema => {
+  if (!schema) {
```
Asalem1, line 68: We shouldn't ever hit this, but if for some reason the schema is undefined, the page shouldn't crash

src/timeMachine/actions/queries.ts
```diff
@@ -299,7 +299,10 @@ export const executeQueries = (abortController?: AbortController) => async (
       if (isCurrentPageDashboard(state)) {
         // reset any existing matching query in the cache
         resetQueryCacheByQuery(text)
-        const result = getCachedResultsOrRunQuery(orgID, text, state)
+        const usedVars = filterUnusedVarsBasedOnQuery(getAllVariables(state), [
```
drdelambre, line 302: seems like this needs to go inside of `getCachedResultsOrRunQuery` if you're making copy pasta with it. the interface being `nameThingsBetter(orgID, text, getAllVariables(state))` and having the function filter out any of those variables not referenced in the query before making the hash

src/shared/apis/queryCache.ts
drdelambre: the fetching of the variables should totally be a parent concern, but i think there's value in having an interface where you can throw it some dirty values and it'll clean them up for you

src/shared/apis/queryCache.ts
```diff
@@ -120,10 +145,21 @@ class QueryCache {
     values: RunQueryResult
   ): void => {
     event('Query Cache was Set', {context: 'queryCache', queryID})
-    this.cache[queryID] = {
-      ...this.initializeCacheByID(queryID, hashedVariables),
-      dateSet: Date.now(),
-      values,
+    if (values.type === 'SUCCESS') {
```
drdelambre, line 148: i'm not normally one to say something here, but the duplication in these conditions rubs me the wrong way for some reason

src/timeMachine/actions/queries.ts
Asalem1: I see what you're saying. So we should be passing all the variables and then filtering them within the child component? What's the benefit of that as opposed to just determining what's used in the parent, and passing it along

src/shared/apis/queryCache.ts
Asalem1: Sounds good

src/flows/context/query.tsx
drdelambre: why not use the query context instead?

src/flows/context/schemaProvider.tsx
```diff
@@ -42,17 +42,78 @@ export const SchemaContext = React.createContext<SchemaContextType>(
...
+    /^result$/,
+  ]
+  if (!result) {
+    return
+  }
+  if (result?.type !== 'SUCCESS') {
+    throw new Error(result.message)
+  }
+
+  const out = fromFlux(result.csv).table as any
```
drdelambre, line 64: shouldn't the result already have a parsed result?

src/flows/context/query.tsx
Asalem1: Do you man to use the `query` from QueryContext in order to pass the query in and run it?

src/flows/context/schemaProvider.tsx
Asalem1: Nah, currently we're just setting the value that's returned from the query

src/flows/context/schemaProvider.tsx
```diff
@@ -69,22 +127,40 @@ export const SchemaProvider: FC<Props> = React.memo(({children}) => {
...
+        const schemaForBucket = parsedResultToSchema(response.parsed)
+        setSchema(schemaForBucket)
+        setLoading(RemoteDataState.Done)
+      })
+      .catch((error: RunQueryResult) => {
+        console.error('error: ', error)
+        setLoading(RemoteDataState.Error)
+        setSchema({})
+      })
+  }, [data?.bucket?.name, data?.bucket, flow?.range, query])
```
drdelambre, line 156: just need data.bucket.name here right? should make the one on line 130 look for that as well

drdelambre: since flow.range is an object, this might be called more than it needs to be because of the shallow compare and tendency to overwrite using partials in the contexts. formatTimeRangeArguments returns a string, which should be more stable / intentional than the object reference within the shallow compare, so pull range out of this useEffect and reference it from within

src/flows/context/schemaProvider.tsx
```diff
@@ -69,22 +127,40 @@ export const SchemaProvider: FC<Props> = React.memo(({children}) => {
...
 
-  const {fields, measurements, tags} = normalizeSchema(schema, data, searchTerm)
+  useEffect(() => {
+    const normalized = normalizeSchema(schema, data, searchTerm)
```
drdelambre, line 159: this gets run everytime a tag is selected within the metric selector. intentional?

drdelambre: also, since these are all derived values, wouldn't it make more sense to use a `useMemo` here?

drdelambre (approved):
i think i tied a bow on the last thing? lemme know.

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #459, opened 2020-12-22
chore: added guarding on load-data page

Asalem1:
This PR closes an issue that @kristinarobinson came across where a user without buckets trying to access the `load-data` page made the app display an error in the UI. This PR adds some guarding against some of the assumptions that the `load-data` page was making


121watts (approved):

src/writeData/components/WriteDataDetailsContext.tsx
```diff
@@ -80,7 +80,7 @@ const WriteDataDetailsContextProvider: FC<Props> = ({
 }
 
 const mstp = (state: AppState) => {
-  const buckets = getAll<Bucket>(state, ResourceType.Buckets)
+  const buckets = getAll<Bucket>(state, ResourceType.Buckets) ?? []
```
ShmuelLotman, line 83: Just curious - Should this be something thats done here, or in the getAll response so as to potentially guard against this elsewhere? Or does the function response necessitate potentially returning a null value?

Asalem1: Good point. moved it up the pipeline

ShmuelLotman (approved):
Approved. LGTM. Thanks for moving the check for empty array upstream.

PR #462, opened 2020-12-22
fix: point at my branch for circleci oss test verification

joelfried:
This will let us see for certain that the monitor-ci branch in question passes oss e2e tests


Asalem1 (approved):
🤞🏽 here's to hoping

PR #464, opened 2020-12-23
feat: fixed some bugs and allowed users to query from the pipe icon

### LINKED ISSUE #417: Notebooks:  I can preview/run a notebook
## User Problem
As a notebooks early access user, I don't know how to preview/run the notebook and try clicking the prompt below the cell that says "Click Preview to See Results"

Example from @abalone23, "I tried clicking it multiple times before reading the text that said to click Preview located at the top of the notebook."

## Proposed solution
Given: I have a notebook with at least one cell
When: I click the Play button or the text "Click Preview/Run to See Results"
Then: The notebook is Previewed/Ran

Asalem1:
Closes #417

The primary purpose of this PR is to allow users to run queries when clicking the `play` icon in the pipe:

run-query

This PR also addresses some random issues that were causing Notebooks to be wonky, namely:

- The run button never being disabled due to a miscalculation in identifying metric selector queries
- removing aggregateOfAggregates as a hack to run queries explicitly when updating an aggregate type and just running the query when an aggregate type is changed
- Only changes the loading state of pipes that are being run

### DIFF
--- a/src/flows/components/header/Submit.tsx
+++ b/src/flows/components/header/Submit.tsx
@@ -18,17 +18,20 @@ import {FlowContext} from 'src/flows/context/flow.current'
 
 // Utils
 import {cancelAllRunningQueries} from 'src/timeMachine/actions/queries'
-import {PIPE_DEFINITIONS} from 'src/flows'
+import {event} from 'src/cloud/utils/reporting'
 
 // Styles
 import 'src/flows/components/header/Submit.scss'
+
+// Types
+import {RemoteDataState} from 'src/types'
 
 const fakeNotify = notify
 
 export const Submit: FC = () => {
   const {flow} = useContext(FlowContext)
   const {runMode, setRunMode} = useContext(RunModeContext)
-  const {isLoading, queryAll} = useContext(QueryContext)
+  const {generateMap, queryAll} = useContext(QueryContext)
 
   const _range = useMemo(
     () => `${flow.range.lower} to ${flow.range.upper || 'now'}`,
@@ -36,24 +39,34 @@ export const Submit: FC = () => {
   )
 
   const hasQueries = useMemo(() => {
-    return !!flow.data.all
-      .filter(p => {
-        // this condition ensures that empty metric selectors are filtered out
-        if (p?.type === 'metricSelector') {
-          return !!p?.bucket
-        }
-        return true
-      })
-      .map(p => PIPE_DEFINITIONS[p.type])
-      .filter(p => p && (p.family === 'transform' || p.family === 'inputs'))
-      .length
-  }, [flow.data])
+    return generateMap().length > 0
+  }, [generateMap])
 
   useEffect(() => {
     if (hasQueries) {
       queryAll()
     }
   }, [_range, hasQueries]) // eslint-disable-line react-hooks/exhaustive-deps
+
+  const handleSubmit = () => {
+    event('Notebook Submit Button Clicked')
+    queryAll()
+  }
+
+  const status = flow.meta.all.reduce((_, curr) => {
+    switch (curr?.loading) {
+      case RemoteDataState.Error:
+        return RemoteDataState.Error
+      case RemoteDataState.NotStarted:
+        return RemoteDataState.NotStarted
+      case RemoteDataState.Loading:
+        return RemoteDataState.Loading
+      case RemoteDataState.Done:
+        return RemoteDataState.Done
+      default:
+        return RemoteDataState.NotStarted
+    }
+  }, RemoteDataState.NotStarted)
 
   const DropdownButton = (
     active: boolean,
@@ -66,8 +79,8 @@ export const Submit: FC = () => {
           className="flows--submit-button"
           icon={IconFont.Play}
           submitButtonDisabled={hasQueries === false}
-          queryStatus={isLoading}
-          onSubmit={queryAll}
+          queryStatus={status}
+          onSubmit={handleSubmit}
           onNotify={fakeNotify}
           queryID=""
           cancelAllRunningQueries={cancelAllRunningQueries}
--- a/src/flows/components/header/Submit.tsx
+++ b/src/flows/components/header/Submit.tsx
@@ -53,20 +53,17 @@ export const Submit: FC = () => {
     queryAll()
   }
 
-  const status = flow.meta.all.reduce((_, curr) => {
-    switch (curr?.loading) {
-      case RemoteDataState.Error:
-        return RemoteDataState.Error
-      case RemoteDataState.NotStarted:
-        return RemoteDataState.NotStarted
-      case RemoteDataState.Loading:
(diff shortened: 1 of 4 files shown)


src/flows/components/header/Submit.tsx
```diff
@@ -16,130 +9,51 @@ import {
...
+  const _range = useMemo(
+    () => `${flow.range.lower} to ${flow.range.upper || 'now'}`,
+    [flow.range]
+  )
 
   const hasQueries = useMemo(() => {
     return !!flow.data.all
+      .filter(p => {
+        // this condition ensures that empty metric selectors are filtered out
+        if (p?.type === 'metricSelector') {
```
drdelambre, line 42: the submit button shouldn't know about the implementation details of a self registering type. can we use generateMap somehow instead? like "if stages.length" or something

src/flows/components/header/Submit.tsx
```diff
@@ -151,9 +65,9 @@ export const Submit: FC = () => {
           text={runMode}
           className="flows--submit-button"
           icon={IconFont.Play}
-          submitButtonDisabled={!hasQueries}
+          submitButtonDisabled={hasQueries === false}
```
drdelambre, line 68: this change isn't needed right? just the filter above to make the button disabled?

src/flows/context/query.tsx
```diff
@@ -81,11 +96,14 @@ const findOrgID = (text, buckets) => {
...
 export const QueryProvider: FC = ({children}) => {
   const {flow} = useContext(FlowContext)
+  const {runMode} = useContext(RunModeContext)
+  const {add, update} = useContext(ResultsContext)
   const variables = useSelector((state: AppState) => getVariables(state))
   const buckets = useSelector((state: AppState) => getSortedBuckets(state))
   const bucketsLoadingState = useSelector((state: AppState) =>
     getStatus(state, ResourceType.Buckets)
   )
+  const [isLoading, setLoading] = useState(RemoteDataState.NotStarted)
```
drdelambre, line 106: this isn't it's own state, but a derived state. it's not that `queryAll` sets the state and keeps track of it, but that it's a conglomeration of all of the stages loading state. instead of a useState, it's more of a `flows.meta.all.reduce(meta => if any are loading, show the loading state), RemoteDataState.NotStarted)'

src/flows/context/query.tsx
```diff
@@ -191,12 +209,72 @@ export const QueryProvider: FC = ({children}) => {
...
+      return
+    }
+
+    const map = generateMap(runMode === RunMode.Run)
+
+    if (!map.length) {
+      return
+    }
+
+    event('Flow Submit Button Clicked')
```
drdelambre, line 231: this event should exist on the submit button only, and not on the underlying implementation of the submit logic. then if we wanna change how we submit something, the events in GA keep ticking

src/flows/context/query.tsx
drdelambre: and yes i know that's not how it worked before. but i think we're in a better place now with this

src/flows/pipes/MetricSelector/AggregateWindowSelector.tsx
```diff
@@ -31,8 +34,9 @@ const AggregateFunctionSelector: FC = () => {
         function: aggregateFunction.name,
       })
       update({aggregateFunction})
+      queryAll()
```
drdelambre, line 37: omg so much better

src/flows/shared/ResizerHeader.tsx
```diff
@@ -34,16 +40,22 @@ const ResizerHeader: FC<Props> = ({
...
   const {readOnly} = useContext(PipeContext)
+  const {queryAll} = useContext(QueryContext)
   const glyph = visibility === 'visible' ? IconFont.EyeOpen : IconFont.EyeClosed
   const className = classnames('panel-resizer--header', {
     'panel-resizer--header__multiple-controls':
       toggleVisibilityEnabled || additionalControls,
     [`panel-resizer--header__${visibility}`]: resizingEnabled && visibility,
   })
 
+  const handleSubmit = () => {
```
drdelambre, line 51: :exploding_head:

src/flows/components/header/Submit.tsx
Asalem1: It isn't necessary, but i think it makes it a bit more understandable

src/flows/components/header/Submit.tsx
Asalem1: I hear ya, I'll give the stages one a shot

src/flows/context/query.tsx
Asalem1: Gotcha. SO then this should be obtainable in the submit context

src/flows/context/query.tsx
Asalem1: Totally

src/flows/components/header/Submit.tsx
```diff
@@ -16,130 +9,64 @@ import {
...
-    if (isLoading === RemoteDataState.Loading) {
-      return
-    }
-
-    const map = generateMap(runMode === RunMode.Run)
-
-    if (!map.length) {
-      return
+  const status = flow.meta.all.reduce((_, curr) => {
+    switch (curr?.loading) {
```
drdelambre, line 57: we gotta get a little more complicated with this one. as is, this'll only return the loading state of the last pipe. instead, lets change it back to an `isLoading` boolean, but write the reduce like this:
```
const isLoading = flow.meta.all.reduce((acc, curr) => {
  return acc || curr.loading === RemoteDataState.Loading
}, false)
```
with this, as long as any of them are still loading, no mater how they were triggered or how long they take, the submit button stays loading

src/flows/context/query.tsx
```diff
@@ -191,12 +206,83 @@ export const QueryProvider: FC = ({children}) => {
...
 
+  const forceUpdate = (id, data) => {
+    try {
+      update(id, data)
+    } catch (_e) {
+      add(id, data)
+    }
+  }
+
+  const status = flow.meta.all.reduce((_, curr) => {
```
drdelambre, line 217: see the note above for how to change this up

drdelambre (approved):
:fire:

src/flows/components/header/Submit.tsx
```diff
@@ -16,130 +9,51 @@ import {
...
 
   const hasQueries = useMemo(() => {
     return !!flow.data.all
+      .filter(p => {
+        // this condition ensures that empty metric selectors are filtered out
+        if (p?.type === 'metricSelector') {
+          return !!p?.bucket
+        }
+        return true
+      })
```
genehynson: nit: could be a ternary if you want. `return p?.type === 'metricSelector' ? !!p?.bucket : true`

genehynson (approved):
one nit, but lgtm!

PR #473, opened 2020-12-30
chore: bump that number up

Asalem1:
Since most browsers only allow for 6 concurrent requests to the same API (queueing every one thereafter), we want to increase the chunk size of LP that's being written to the API in order to reduce the number of API requests that are being made. 

This PR also adds some minor error handling in the event that a user has uploaded some malformed CSVs. In its current implementation, the user will just see a loading screen that doesn't do anything. This allows the user to receive actionable feedback


src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -35,17 +39,16 @@ export const CsvUploaderContext = React.createContext<CsvUploaderContextType>(
...
 
 /**
- * This is an arbitrary number that should be capped out at 1750.
- * For now, the number is set to 500 since there were significant performance issues
- * that had cropped up when writing 1750 lines on tools that took ~20s for a write query
- * to resolve
+ * This is an arbitrary number that should be capped out at 1750 due to max data sizes
  */
-const MAX_CHUNK_SIZE = 500
+const MAX_CHUNK_SIZE = 1750
```
drdelambre, line 44: not denying that this will help alleviate the problem in tools, but changing a magic number to fix it is the wrong solution. there are no guarantees that this number works in both tools and production, and if it's determined by the deployment environment of the api, that also means there's an undocumented constant out there that can drastically affect the user experience, leaving this interface susceptible to being disabled as a side effect of a configuration file that has no direct connection to the interface.

if there's too much infrastructure required to make this constant derive it's value from the api option that alters it, at least providing some documentation here on why tools is different that local development and what mechanism affects that performance should help mitigate the tight coupling.

drdelambre (approved):
i think there's some incorrect assumptions about the system when we look into how requests are handled with http/2, but matches observation and fits a logical framework

hercules-influx:
Metric | Score | Level
performance | 26 | :red_circle:
seo | 80 | :yellow_circle:
pwa | 22 | :red_circle:

PR #475, opened 2021-01-04
feat: better isolation

drdelambre:
To create a visualization within our code, we currently have to create a component that generates a giraffe config from a combination of view properties and results in a folder `src/shared/components`. We then have to register that component to a giant switch statement in `src/shared/components/ViewSwitcher`.
To edit those view properties, we need to have a view option interaction defined in `src/timeMachine/components/view_options' that is then registered to `src/shared/visualization`, with defaults defined in `src/helpers` somewhere.

As each instance of the visualization rendering pipeline has it's own implementation, any changes then need to be validated in the dashboard page, the data explorer page, the checks page, and notebooks to determine if the change introduces any bug into the system.

In short: it's complicated, and it's difficult to pull off a feature in the core of our app. Try and open a band chart in flows to see how we're failing.

What this PR aims to do is create a single rendering pipeline that can be used across the app. Dashboard cells, time machine, flows... all using a single component, with a unified interface. That interface is defined in `src/visualization/index.tsx` and implemented through a self registration scheme similar to the one introduced in notebooks. Only three components need to be exported from the `src/visualization` in order to render a graph on a page: `View`, `ViewOptions`, and `ViewTypeDropdown`. Since this uses the self registration mechanism, each visualization type is now strongly decoupled and can be iterated on independently without fear of breaking other types, or implicit relationships between modules that might cause rendering artifacts.

While doing this, i found it hard to see what modules we were no longer using based on the simplification of the system. So i ran a webpack plugin that shows all of the unused files within our repo. And then I deleted them. The next step would be to remove all exported modules that aren't being consumed in the application, but that seemed to require the development of custom tooling.


package.json
```diff
@@ -195,13 +186,9 @@
     "react-monaco-editor": "^0.33.0",
     "react-redux": "^7.2.0",
     "react-router-dom": "^5.2.0",
-    "react-scrollbars-custom": "^4.0.0-alpha.8",
     "react-virtualized": "^9.18.5",
     "redux": "^4.0.0",
-    "redux-auth-wrapper": "^1.0.0",
     "redux-thunk": "^2.3.0",
-    "reselect": "^4.0.0",
-    "rome": "^2.1.22",
```
Asalem1, line 204: let it burn

hercules-influx:
Metric | Score | Level
performance | 34 | :red_circle:
seo | 80 | :yellow_circle:
pwa | 22 | :red_circle:

TCL735:
This works for me locally. Band chart is more consistent when toggling between Query Builder and Script Editor. And I was able to get it to work in Notebooks using Flux. Still needs work on the metric selector specifically for Band chart.

src/visualization/README.md
```diff
@@ -0,0 +1,27 @@
...
+- `graphic` (_required_): currently these are svg images, stored as react components, that get used as icons for the `ViewTypeDropdown` to differentiate the visualization types with pictures
+- `initial` (_required_): the default values for the ViewProperties that are being passed in. this is used to normalize data that comes in from the API and as a means of ensuring we dont get any errors when switching between visualization types. This was previously done through `src/views/helpers` with some strange rules.
+- `component` (_required_): The visualization to render for the `View`. These defintions were previously found in `src/shared/components` and controlled by the `ViewSwitcher`. The majority of these components take ViewProperties and FluxResults, normalize them, and translate them into a giraffe interface.
+- `options` (_optional_): this is a component that represents the interface used to configure the ViewProperties of a visualization. The defintions of this aspect of a visualization used to be everywhere, but mostly in `src/timeMachine/components/view_options`.
```
TCL735, line 27: :fire:

I like this. Very clear to me. Thank you for writing it.

TCL735 (approved):

hoorayimhelping (commented):
Before anything, I want to call attention to your initiative and vision on this. It's not entirely clear in hindsight that this was a big source of pain, and I think cleaning it up and making it easier to understand is a really valuable and wise thing to do. Thank you for taking it on!

I want to make sure I understand your intent, so I'm going to try to restate what I got out of your words to make sure we're on the same page:

Previously, the view rendering code was spread across many locations, and full of implicit assumptions and inconsistent interfaces. You've taken the actual code that was in disparate areas of the app and consolidated it into a single location under `visualizations`. You've also examined the concepts, and pulled out the relevant assumptions and made them explicit with the TypeScript interface `VisTypeRegistration`. There are now just a handful of container components that handle the concepts of a visualization's `View`, `ViewOptions`, and `ViewTypeDropdown`.

There's also a concept of `TYPE_DEFINITIONS` which is a map of all the view types the app recognizes, and is filled in as a side effect the first time the `src/visualization/index.ts` file is loaded/executed. `TYPE_DEFINITIONS` is intrinsically linked to `VisTypeRegistration` - it is the map of instantiated `VisTypeRegistration` s. These are ultimately used by `View` to render the correct visualization type with the correct metadata.

---

Its the relationship between `VisTypeRegistration`, `TYPE_DEFINITIONS`, and `visualization/types/` that jumps out to me. To use your own words, it's a bit magical. The `TYPE_DEFINITIONS` data structure is required to render visualizations, but it doesn't actually exist in the source code. It's not real until the app is run. It's encoded implicitly in the directory structure of `visualizations/types`.

A consequence of this is that our visualization rendering pipeline relies on a data structure existing implicitly and being filled in properly by the time it executes to work. That could potentially be a very heinous souce of bugs.

Another consequence of this is that it's almost impossible to understand what is happening just by reading code. The code needs to be run to understood.

This magic implicitness is my main sticking point, especially in context of the work you've already done to remove chaotic implicitness and bring explicitness to this pipeline. I'd really like to see this be more explicit and clear, even if it means making a coarse data structure like this so that all the information about view types exists in a single place that can act as the definitive source of truth:

```js
const SUPPORTED_VIEW_TYPES = [{
    type: 'xy',
    name: 'Graph',
    graphic: XYIcon,
    component: XYView,
    initial: xyProperties,
    options: xyOptions,
  }, {
    type: 'geo',
    name: 'Map',
    graphic: GeoIcon,
    featureFlag: 'mapGeo',
    initial: geoPoperties,
    component: GeoView,
  }
]
```

Another thought: since feature flags are involved and this is actually stateful data and dependent upon the user, could this be solved by putting this into redux?

---

There's a separate question about whether we should prefer default exports or named exports. I'm bringing it up in this PR because the architecture seems dependent on modules being exported default in a particular format and I'd like to make sure we're all consistent on what we're doing as a team going forward.

I've worked under both styles and I prefer named exports for a number of reasons. The two main ones are that default exports have a tendency to make the link between an export and its import unclear because a default export doesn't have a name; and because default exports interfer with dead code elimination / tree shaking, which we get for free with webpack 2+. Here are a couple of articles that make other points about it:

  - https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/
  - https://basarat.gitbook.io/typescript/main-1/defaultisbad
  - https://blog.piotrnalepa.pl/2020/06/26/default-exports-vs-named-exports/

What are the reasons you prefer default exports other than React code splitting requires it?

---

Very solid work overall.

drdelambre:
seems to be some confusion:
The building of the `TYPE_DEFINITIONS` object happens at build time and not at run time. All of the tree shaking / module loading steps are implemented upon the dependency graph that is generated, including these modules. I guess, all this stuff should be available for inspection in a deterministic way without having to instantiate the module itself, unless obfuscated by some other means unrelated to this PR (ie: lazy loading). This mechanism also allows us to implement new visualization types and collaborate on them locally, but completely isolate them from the customer's bundle. Think of the workflow mosaic went through, multiple devs working in the same place, trying to keep a shared branch from getting stale against master, while having their own PR process against that branch... we can remove a lot of frustrating overhead by using a `disabled` flag in the data contract

The "magic" part is intentional. By requiring implementations to adhere to the data contract and not code convention, it forces loose coupling of those implementations to the system (everything comes through one door) and adds friction to the subversion of it. What comes out the other side, after getting used to the ergonomics, is better encapsulated modules that adhere to tree shaking... friendly? patterns. If i want to go and delete the `Gauge` visualization, i just delete that folder; the rendering component, it's style rules, it's view property definition, any sub components, and all of the libraries that it references in isolation, are all removed.

 Regarding readability, i'd argue that by creating a single door into the system, we've removed the requirement that a developer has a working knowledge of the entire application and can now focus on writing to one spec. Currently within our code base, when adding a new visualization type, we expect the developer to know or uncover the 8 touch points that must all be satisfied before their visualization type can be seen by the user interface. Maintenance is an even worse story, as you have to now ensure that each of these disparate touch points are implemented without any side effects in step, without knowledge of how the original functionality is _supposed_ to work in the first place. Commonly, we find bugs arising from the developer having an incomplete picture of the surface area of a module; fixing an issue in one area only to find that their solution breaks something elsewhere. Here, a module's surface area is documented and defined within the module's registration function. Defining it externally just creates another source of truth.. another touch point for the developer to find. Do i think we should use this pattern everywhere? No. But we should consider it any time we find ourselves with a giant switch statement within the system.
 
I dont understand your redux question, as these are all functional transforms of redux state until they hit Giraffe.
 
Default vs named exports feels like bike shedding to me. i dont have any quantifiable, or even anecdotal, evidence that it has any impact to a reduction of bugs that are released to the customer, conversion of users from free accounts to payed accounts, or the speed at which new functionality is released. I also think that we should have 4 spaces and not 2, but the juice isn't worth the squeeze. Even with my caveman development workflow tooling of grep + vim, we have a habit of naming the default export the same as if it were a named export when importing it, and namespace collisions of named exports create the same issue you're trying to prevent, ie: `import {Beans as Foo} from 'src/Beans`. Seems like we got bigger fish to fry.

src/visualization/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export interface VisProps {
+  properties: ViewProperties
+  result: FromFluxResult
+  theme?: Theme
+  timeZone?: TimeZone
+  timeRange?: TimeRange
+}
+
+export interface VisTypeRegistration {
```
hoorayimhelping, line 19:
```suggestion
export interface Visualization {
```

this is the guy. it's what represents a visualization in our system. call it out loudly

src/visualization/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  name: string // the name that shows up in the dropdown
+  graphic: JSX.Element // the icon that shows up in the dropdown
+  component?: FunctionComponent<VisProps> | ComponentClass<VisProps> // the view component for rendering the interface
+  disabled?: boolean // if you should show it or not
+  featureFlag?: string // designates a flag that should enable the panel type
+  initial: ViewProperties // the default state
+  options?: FunctionComponent<VisOptionProps> // the view component for rendering the interface
+}
+
+interface VisTypeRegistrationMap {
```
hoorayimhelping, line 30:
```suggestion
interface Visualizations {
```

src/visualization/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  [key: string]: VisTypeRegistration
+}
+
+export const TYPE_DEFINITIONS: VisTypeRegistrationMap = {}
```
hoorayimhelping, line 34:
```suggestion
export const SUPPORTED_VISUALIZATIONS: VisTypeRegistrationMap = {}
```

```
SUPPORTED_VISUALIZATIONS
AVAILABLE_VISUALIZATIONS
ACTIVE_VISUALIZATIONS
REGISTERED_VISUALIZATIONS
USER_VISUALIZATIONS
```
etc

src/visualization/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const context = require.context('./types', true, /index\.(ts|tsx)$/)
```
hoorayimhelping, line 36:
```suggestion
const visualizationTypeModules = require.context('./types', true, /index\.(ts|tsx)$/)
```

src/visualization/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  graphic: JSX.Element // the icon that shows up in the dropdown
+  component?: FunctionComponent<VisProps> | ComponentClass<VisProps> // the view component for rendering the interface
+  disabled?: boolean // if you should show it or not
+  featureFlag?: string // designates a flag that should enable the panel type
+  initial: ViewProperties // the default state
+  options?: FunctionComponent<VisOptionProps> // the view component for rendering the interface
+}
+
+interface VisTypeRegistrationMap {
+  [key: string]: VisTypeRegistration
```
hoorayimhelping, line 31:
```suggestion
  [visualizationType: string]: VisTypeRegistration
```

`key` is kind of vague, hinting that it's a map of objects keyed on the type makes it so much easier for me conceptualize

src/visualization/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  timeRange?: TimeRange
+}
+
+export interface VisTypeRegistration {
+  type: string // a unique string that identifies a visualization
+  name: string // the name that shows up in the dropdown
+  graphic: JSX.Element // the icon that shows up in the dropdown
+  component?: FunctionComponent<VisProps> | ComponentClass<VisProps> // the view component for rendering the interface
+  disabled?: boolean // if you should show it or not
+  featureFlag?: string // designates a flag that should enable the panel type
```
hoorayimhelping, line 25: can you add a little note around `disabled` and `featureFlag` explaining that `disabled` takes precedence please?

cypress/e2e/shared/explorer.test.ts
```diff
@@ -12,6 +11,20 @@ import {
...
+  'xy',
+  'heatmap',
+  'histogram',
+  //  'map',
+  'mosaic',
+  'scatter',
+  'single-stat',
+  'line-plus-single-stat',
+  'table',
+]
```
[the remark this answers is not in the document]
drdelambre: not quite. this is just for tests, and coupling the testing framework (that's defined nearly completely separately, with little aprori knowledge of the application configuration) to the application build system seems... antithetical to the attempt to isolate subsystems. It's mostly used to hoist definitions required for line 758 to the top, where it's easier to inspect

PR #489, opened 2021-01-07
feat(geo): Add Map to data explorer's view type dropdown

Palakp41:
Closes #9213

Describe your proposed changes here:
Added a view type for Map Geo and put it behind the feature flag. Right now, we are able to select Map in the dropdown but not rendering map plot from giraffe. That implementation will be in another pr.

Kapture 2021-01-07 at 12 47 07

Screen Shot 2021-01-13 at 10 46 40 AM

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/shared/components/GeoPlot.tsx
+++ b/src/shared/components/GeoPlot.tsx
@@ -1,0 +1,54 @@
+// Libraries
+import {FunctionComponent} from 'react'
+import {Config, Table} from '@influxdata/giraffe'
+
+// Types
+import {GeoViewProperties} from 'src/types'
+
+interface Props {
+  children: (config: Config) => JSX.Element
+  fluxGroupKeyUnion?: string[]
+  table: Table
+  viewProperties: GeoViewProperties
+}
+
+const GeoPlot: FunctionComponent<Props> = ({
+  children,
+  table,
+  viewProperties,
+}) => {
+  const {
+    layers,
+    zoom,
+    allowPanAndZoom,
+    detectCoordinateFields,
+    mapStyle,
+  } = viewProperties
+  const tileServerConfiguration = {
+    tileServerUrl: '',
+    bingKey: '',
+  }
+
+  const {lat, lon} = viewProperties.center
+
+  const config: Config = {
+    table,
+    showAxes: false,
+    layers: [
+      {
+        type: 'geo',
+        lat,
+        lon,
+        zoom,
+        allowPanAndZoom,
+        detectCoordinateFields,
+        mapStyle,
+        layers,
+        tileServerConfiguration: tileServerConfiguration,
+      },
+    ],
+  }
+  return children(config)
+}
+
+export default GeoPlot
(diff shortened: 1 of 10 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/shared/components/GeoPlot.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const {
+    layers,
+    zoom,
+    allowPanAndZoom,
+    detectCoordinateFields,
+    mapStyle,
+  } = viewProperties
+  const tileServerConfiguration = {
+    tileServerUrl:
+      'https://api.mapbox.com/styles/v1/influxdata/ckhl79okh00o919npquotuqxp/tiles/256/{z}/{x}/{y}@2x?access_token=pk.eyJ1IjoiaW5mbHV4ZGF0YSIsImEiOiJja2hndnM0eWQwNTVtMzRwZnd5cmQxNHhoIn0.IEAjIguiM67ql4mhiD88rw',
```
hoorayimhelping, line 30: what's up with this being hardcoded here?

src/shared/utils/vis.ts
```diff
@@ -376,3 +377,37 @@ export const getMainColumnName = (
...
+  if (typeof value !== 'string') {
+    return null
+  }
+  return value
+}
+
+export const PRECISION_TRIMMING_TABLE = [BigInt(1)]
+for (let i = 1; i < 17; i++) {
+  PRECISION_TRIMMING_TABLE[i] = PRECISION_TRIMMING_TABLE[i - 1] * BigInt(16)
+}
```
hoorayimhelping, line 396: What's up with this `for` loop? what is `17`? should `PRECISION_TRIMMING_TABLE` be initialized with the proper values when it's exported?

src/shared/utils/vis.ts
```diff
@@ -376,3 +377,37 @@ export const getMainColumnName = (
...
+
+export const getLatLon = (table, index: number) => {
+  const cellId = getS2CellID(table, index)
+  if (cellId === null || cellId.length > 16) {
```
hoorayimhelping, line 400: what is `16` other than something that relates to `17`? what do these magic numbers mean?

src/shared/components/GeoPlot.tsx
Palakp41: good catch, i removed it, then rebased and pushed and forgot to check

src/shared/components/GeoPlot.tsx
```diff
@@ -0,0 +1,54 @@
...
+}) => {
+  const {
+    layers,
+    zoom,
+    allowPanAndZoom,
+    detectCoordinateFields,
+    mapStyle,
+  } = viewProperties
+  const tileServerConfiguration = {
+    tileServerUrl: 'mapboxurlhere',
```
hoorayimhelping, line 28: maybe just leave this an empty string 🤷‍♂️

hoorayimhelping (approved):
Great work! much cleaner!

src/shared/components/GeoPlot.tsx
Palakp41: Updated it :)

PR #496, opened 2021-01-08
test: adding refactoring to simplify graph configurations.

### LINKED ISSUE #453: reduce code duplication when creating graph configurations
for ui/src/views/helpers/index.ts:

   there is a lot of code duplication when creating the properties.

so:  refactor out duplication and write unit tests alongside them.

question:  why do we need all these null properties?  empty arrays and properties with default values are understandable, but do we need all these empty/null properties?

partnered with @joelfried

jrenee42:
put all the 'tick' properties together
   also testing the 'legend properties which were grouped together when doing the colorize rows (for legend) feature

Closes #453

- [ ] added test file
- [ ] Tests pass

### DIFF
--- a/src/views/helpers/index.test.ts
+++ b/src/views/helpers/index.test.ts
@@ -1,0 +1,87 @@
+import {createView, defaultBuilderConfig} from 'src/views/helpers/index'
+import {
+  XYViewProperties,
+  ScatterViewProperties,
+  BandViewProperties,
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | 22 | :red_circle:

TCL735 (approved):

PR #498, opened 2021-01-11
feat: Auto Scale Time Toggle For X Axis On Scatter Plot

ShmuelLotman:
Closes #19867 

This PR introduces a toggle for the auto domain scaling along the x axis. When active, it will scale to the data set, when inactive it defaults to the current behavior. 

- [x] Rebased/mergeable
- [x] Tests pass


hercules-influx:
Metric | Score | Level
performance | 34 | :red_circle:
seo | 80 | :yellow_circle:
pwa | 22 | :red_circle:

src/timeMachine/components/view_options/ScatterOptions.tsx
```diff
@@ -175,6 +176,11 @@ const ScatterOptions: FunctionComponent<Props> = props => {
         availableColumns={numericColumns}
         axisName="x"
       />
+      {xColumn === '_time' && (
```
hoorayimhelping, line 179: I just searched around and noticed that we don't have any kind of typed description of `'_time'` which is odd, since it's so important.

https://github.com/influxdata/ui/issues/499

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {useSelector} from 'react-redux'
+import {SlideToggle, ComponentSize} from '@influxdata/clockface'
+
+import {getXColumnTimeDomainBounds} from 'src/timeMachine/selectors'
+
+interface Props {
+  onSetDomain: (xDomain: [number, number]) => void
+}
+
+const TimeDomainAutoToggle: FC<Props> = ({onSetDomain}) => {
```
hoorayimhelping, line 11:
```suggestion
export const TimeDomainAutoToggle: FC<Props> = ({onSetDomain}) => {
```

default exports aren't great in a team environment. They make searching a large codebase like ours difficult, they lead to inconsistent naming on a 12 person team like ours, and they make tree shaking and code splitting less effective.

https://basarat.gitbook.io/typescript/main-1/defaultisbad

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,38 @@
...
+        active={isActive}
+        onChange={handleClick}
+        disabled={!timeDomain}
+        testID="time-domain-toggle-slide"
+      />
+    </div>
+  )
+}
+
+export default TimeDomainAutoToggle
```
hoorayimhelping, line 38:
```suggestion
```

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {getXColumnTimeDomainBounds} from 'src/timeMachine/selectors'
+
+interface Props {
+  onSetDomain: (xDomain: [number, number]) => void
+}
+
+const TimeDomainAutoToggle: FC<Props> = ({onSetDomain}) => {
+  const [isActive, setIsActive] = useState<boolean>(false)
+
+  const timeDomain = useSelector(getXColumnTimeDomainBounds)
```
hoorayimhelping, line 14: what is `timeDomain`? can you describe this with TypeScript?

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,38 @@
+import React, {FC, useState} from 'react'
+import {useSelector} from 'react-redux'
+import {SlideToggle, ComponentSize} from '@influxdata/clockface'
+
+import {getXColumnTimeDomainBounds} from 'src/timeMachine/selectors'
+
+interface Props {
+  onSetDomain: (xDomain: [number, number]) => void
```
hoorayimhelping, line 8:
```suggestion
  setDomain: (xDomain: [number, number]) => void
```

this function sets the domain, right? it's not something that happens in reaction to the domain being set. It's doing the setting, yes?

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,38 @@
...
+    }
+    setIsActive(!isActive)
+  }
+  return (
+    <div data-testid="time-domain-toggle">
+      <SlideToggle
+        style={{height: 'auto'}}
+        size={ComponentSize.ExtraSmall}
+        active={isActive}
+        onChange={handleClick}
```
hoorayimhelping, line 30: kind of oddly named function. 'onChange, execute handleClick'. `handleClick` seems more like a function on a component, something like a function that handles someone clicking on the component. think about it like this: what if we wanted to add super power user keyboard shortcuts and one of those shortcuts let them toggle this component. `handleClick` makes not much sense in that case.

what is the function doing? to me it looks like it's toggling the component being `active` or not. See if you can name it something more along those lines

src/timeMachine/selectors/index.ts
```diff
@@ -147,6 +148,16 @@ export const getXColumnSelection = (state: AppState): string => {
   return defaultXColumn(table, preferredXColumnKey)
 }
 
+export const getXColumnTimeDomainBounds = (
+  state: AppState
+): [number, number] => {
```
hoorayimhelping, line 153: what happens if the checks fail and `null` is returned?

src/timeMachine/selectors/index.ts
```diff
@@ -147,6 +148,16 @@ export const getXColumnSelection = (state: AppState): string => {
...
+  const {table} = getVisTable(state)
+
+  const times = table.getColumn('_time')
+
+  return times ? [times[0] as number, times[times.length - 1] as number] : null
```
hoorayimhelping, line 158: this is a bit unsafe and inferring. it infers that if the `times` is truthy, i'll be an array. That's a bit dangerous.

I think it would improve readability and make the code more explicit to expand this out into one or more guard statements like:

```typescript
if (!Array.isArray(times)) {
  return null
}

if (times.length < 2) {
  return null
}

return [times[0] as number, times[times.length - 1] as number]
```

hoorayimhelping (changes requested):
some code suggestions and nitpicks.

I think you should also include a screenshot or a gif showing this change, so people can quickly understand what's happening

TCL735 (changes requested):
Giraffe also has the ability to find the time range of the entire data set. This is what you want here, correct? When xDomain is not set, Giraffe will set the xDomain by finding the extent (min, max) of the data column.

I guess this works too, since there are cases where the UI sets a time range to a smaller period. But just be aware of what is available.

TCL735 (approved):

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {useSelector} from 'react-redux'
+import {SlideToggle, ComponentSize} from '@influxdata/clockface'
+
+import {getVisTable} from 'src/timeMachine/selectors'
+
+import {useVisXDomainSettings} from 'src/shared/utils/useVisDomainSettings'
+
+interface Props {
+  setDomain: (xDomain: [number, number]) => void
+  xDomain: number[] | null
```
hoorayimhelping, line 11: @TCL735 is this `number [] | null` okay for `xDomain`? i've only seen `xDomain` typed to `number[]` in our code. the documentation in giraffe says it should be a number array or it shouldn't exist, which is `undefined`

src/timeMachine/components/view_options/TimeAutoToggle.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+      const {getByTestId} = render(
+        <TimeDomainAutoToggle setDomain={setDomainMock} xDomain={null} />
+      )
+
+      fireEvent.click(getByTestId('time-domain-toggle-slide'))
+
+      expect(setDomainMock).not.toHaveBeenCalled()
+    })
+    it('fires setDomain call to set time with range if clicked from off to on, then resets when going from on to off', () => {
```
hoorayimhelping, line 96: >fires setDomain call to set time with range if clicked from off to on, then resets when going from on to off

testing nit: that is quite a mouthful to read out loud - i think you can cut some words out or re-arrange some to make it a little bit more clear

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const isValidXDomain = xDomain => {
```
hoorayimhelping, line 14: this is an odd place for this function. does something like this in giraffe @TCL735? even if it doesn't, this seems like it belongs elsewhere.

does it need to happen? it feels like this function existing is related to `xDomain` being `number[] | null`...

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,52 @@
...
+    typeof xDomain[1] === 'number'
+  )
+}
+
+export const TimeDomainAutoToggle: FC<Props> = ({setDomain, xDomain}) => {
+  const [isActive, setIsActive] = useState<boolean>(false)
+
+  const {table} = useSelector(getVisTable)
+
+  const [xDom] = useVisXDomainSettings(
```
hoorayimhelping, line 28: `xDom` is a really basic name. chopping off the end of another variable isn't generally a great strategy for writing readable code. this is the new domain right? why not call it that?

src/timeMachine/components/view_options/TimeAutoToggle.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+    setIsActive(!isActive)
+  }
+  return (
+    <div data-testid="time-domain-toggle">
+      <SlideToggle
+        size={ComponentSize.ExtraSmall}
+        active={isActive}
+        onChange={toggleActiveDomain}
+        disabled={!isValidXDomain(xDom)}
```
hoorayimhelping, line 47: is this needed here? can we instead just check if `props.xDomain` exists, and if it doesn't, just skip all this process and render this disabled?

under what conditions is a good `xDomain` passed in, but a bad `xDom` (or whatever it's called) is produced?

hoorayimhelping (commented):
a couple of questions about the overall design of the component

src/timeMachine/components/view_options/TimeAutoToggle.tsx
TCL735: It doesn't need to happen per se. Giraffe will auto-calculate the domain if null is passed in or if any of the following is not given: xDomain, onSetXDomain, onResetXDomain.

But Scatterplot is trying to do a lot here. It wants to use the time range that appears in a dropdown (**not** in the View Options), so that it can handle something other than __time_ on the x-axis.

And the proposed changes here also allow Scatterplot to handle __time_ but ignore the dropdown and auto-calculate the _time domain. I guess we are passing in the __time_ domain specifically to be consistent with the other behavior described above.

If instead Scatterplot passes it sometimes but not other times, it might be confusing for future Devs. And the type checking here is just in support of being consistent.

src/timeMachine/components/view_options/TimeAutoToggle.tsx
ShmuelLotman: Because this allows us not to worry about what props.xDomain is. Whatever is happening outside this component, we now wont mess up the graph if somehow a bad value does make it in.

PR #501, opened 2021-01-11
feat: add cross network communication

subirjolly:
Part 1: We have some existing weirdness where we load Lighthouse and Cypress images into the compose cloud cluster network and tell it to run things using the internal network routing.

It would be super cool to have these service images built and run outside of the cluster, but point to the cloud cluster (kubernetes.docker.internal or equivalent) when running the tests. Seems pedantic, but the testing infrastructure shouldnt be defined within the "cloud infrastructure" definition and we now have the ability to separate them.

**Story: ** https://app.zenhub.com/workspaces/applicationui-5e3b05772922e316b3a210a4/issues/influxdata/monitor-ci/132

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -396,7 +396,7 @@ jobs:
       - run: docker login -u=$QUAY_USER -p=$QUAY_PASS quay.io
       - run:
           name: Grab compose files
-          command: git clone --branch story132 git@github.com:influxdata/monitor-ci.git
+          command: git clone git@github.com:influxdata/monitor-ci.git
       - run:
           name: Copy over the lighthouse variables
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -396,15 +396,20 @@ jobs:
       - run: docker login -u=$QUAY_USER -p=$QUAY_PASS quay.io
       - run:
           name: Grab compose files
-          command: git clone git@github.com:influxdata/monitor-ci.git
+          command: git clone --branch story132 git@github.com:influxdata/monitor-ci.git
       - run:
           name: Copy over the lighthouse variables
           command: |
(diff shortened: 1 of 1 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

.circleci/config.yml
```diff
@@ -401,10 +401,15 @@ jobs:
           name: Copy over the lighthouse variables
           command: |
             cp ./monitor-ci/env.lighthouse ./monitor-ci/.env;
+      - run:
+          name: Add Ingress Host to /etc/hosts File
+          command: |
+            INGRESS_HOST=$(cat ./monitor-ci/.env | grep INGRESS_HOST= | cut -d '=' -f2);
+            grep -qxF "127.0.0.1 ${INGRESS_HOST}" /etc/hosts || sudo -- sh -c "echo \"127.0.0.1 ${INGRESS_HOST}\" >> /etc/hosts";
```
drdelambre, line 408: you can just use `. ./monitor-ci/.env && grep...` to load in the environment variables instead of doing the cat thing. the `.` is the same as `source` on a mac

drdelambre (approved):
:fire: thanks for breaking these up into migrations. once you prove out the lighthouse mechanism, can we do it for cypress too?

subirjolly:
> 🔥 thanks for breaking these up into migrations. once you prove out the lighthouse mechanism, can we do it for cypress too?

Yes sir! I have enlisted those 2 in the story as steps.

.circleci/config.yml
subirjolly: yeah. I just do this for some reason 🤯

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #502, opened 2021-01-11
feat: Annotations API interface POST route definition with tests

ShmuelLotman:
Closes # 490

This PR introduces an interface for the abstraction of interactions with the annotation API, and includes a POST request to the annotations API. 

- [x] Rebased/mergeable
- [x] Tests pass

Sample response included in console during local testing: 
image


hercules-influx:
Metric | Score | Level
performance | 33 | :red_circle:
seo | 80 | :yellow_circle:
pwa | 22 | :red_circle:

src/annotations/api/index.ts
```diff
@@ -0,0 +1,27 @@
...
+// URL
+const url = `${API_BASE_PATH}api/v2private/annotations`
+
+export const writeAnnotation = async (
+  annotations: PostAnnotationPayload[]
+): Promise<AxiosResponse<PostAnnotationResponse[]>> => {
+  try {
+    const res = await axios.post(url, annotations)
+
+    if (res.status !== 200) {
```
hoorayimhelping, line 19: What happens if this returns a 201, or a 202 or a 204, which are all potentially valid here?

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_success

2xx codes are successes, so we usually do something [like this](https://github.com/influxdata/ui/blob/master/src/templates/api/index.ts#L539)

src/annotations/api/index.ts
```diff
@@ -0,0 +1,27 @@
...
+      throw new Error(res.data?.message)
+    }
+
+    return res.data
+  } catch (err) {
+    throw new Error(err?.message)
+  }
```
hoorayimhelping, line 26: ✂️ 

this is the same as not having a catch block because nothing is being done with the error except re-throwing. i liked, it's actually worse than not having a catch block because this strips the error object of context and information, and it also creates a temporary variable that it copies from the caught error.

since this method throws an error when the status is less than 300, any code using this method will have to handle an error. since it'll have to handle an error, it can handle this error that you're catching and re-throwing, and probably know what to do with the error better than this code does. this is api code, it doesn't know what it **means** to have an error, just that something didn't work.

if you're not going to do anything with an error, don't catch it.

src/annotations/api/index.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        Promise.resolve({status: 200, data})
+      )
+
+      const response = await writeAnnotation([{summary: 'Hey there'}])
+
+      expect(axios.post).toHaveBeenCalledTimes(1)
+
+      expect(response).toEqual(data)
+    })
+    it('makes a post request with annotations in the body of the request', async () => {
```
hoorayimhelping, line 31: this test can be removed. it's testing implementation that will change when we remove axios

hoorayimhelping (changes requested):
One change about how we're throwing an error here.

Also, when you write the description and you link to a ticket, put the pound sign against the ticket number, rather than with a space between them. That way, github will automatically link to the ticket, if it's in this repo's issues list.

src/types/annotation.ts
```diff
@@ -0,0 +1,11 @@
+export interface PostAnnotationResponse {
+  stream: string
+  'start-time': string
+  'end-time': string
+  summary: string
+  stickers?: any
+}
```
hoorayimhelping, line 7: oh, one more thing I forgot to mention: since we're only implementing a single annotation and not ranges for the first part, `end_time` should always equal `start_time`

second: we camelCase our variables here, and we make sure external things conform to our standards. so we don't create variables called `start-time` because that's how they're passed into the API. We create a variable called `startTime` and then set `start-time` into `startTime` at the boundaries of the system when the value enters into our domain.

change `'start-time'` to `startTime`

hoorayimhelping (approved):
AND WE'RE OFF!!!!

src/annotations/api/index.test.ts
```diff
@@ -0,0 +1,52 @@
...
+          stream: 'Lambeau Field',
+        },
+      ]
+
+      const transformedData = [
+        {
+          summary: 'GO PACK GO',
+          startTime: Date.now(),
+          endTime: Date.now(),
+          stream: 'Lambeau Field',
```
hoorayimhelping, line 26: 👍

src/annotations/api/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const transformedResponse = res.data.map(d => {
+    const newData = {
+      ...d,
+      startTime: d['start-time'],
+      endTime: d['end-time'],
+    }
+
+    delete newData['start-time']
+    delete newData['end-time']
```
hoorayimhelping, line 30: an alternative to deleting is to copy all the fields over by hand, rather than splatting out `d`.

counting it out, it's the same amount of lines (5), but without any scary delete mutations. I think that's probably a little less scaryscary.

src/annotations/api/index.ts
```diff
@@ -0,0 +1,34 @@
...
+    throw new Error(res.data?.message)
+  }
+
+  const transformedResponse = res.data.map(d => {
+    const {'start-time': startTime, 'end-time': endTime, ...props} = d
+    const newData = {
+      startTime,
+      endTime,
+      ...props,
+    }
```
hoorayimhelping, line 28: this is fine. the thing is, we know what the structure of `PostAnnotationResponse` is, so we can just create an object by hand:

```typescript
const newData = {
  startTime: d['start-time'],
  endTime: d['end-time'],
  stream: d.stream,
  summary: d.summary,
  stickers: d.stickers
}
```

this is explicit. it lays out in clear, unambiguous detail, what properties from the api are being converted to properties of our domain. this will show up when people search through the codebase for either of these types of variables (e.g. searching for `'startTime'` or `'start-time'`). it clearly documents what is happening in local context without making someone look at the definition of `PostAnnotationResponse` to see what properties are being splatted out in `...props`

we want to optimize code to be dead simple and extremely clear and unambiguous when people read it. sometimes that means adding more lines to an existing piece of code. that is okay, if it's in service of comprehension.

hoorayimhelping (approved):

PR #508, opened 2021-01-12
chore: use react suspense to lazy load components by route

### LINKED ISSUE #509: Route Based Code Splitting
Using React Suspense to do route based code splitting should improve our performance score.

See: https://dev.to/redraushan/route-based-code-splitting-in-reactjs-28ah

genehynson:
Closes #509

Use React Suspense to code split based on routes - improves performance scores by 20 points!

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -40,7 +40,7 @@ const App: FC<Props> = ({inPresentationMode, currentPage, theme}) => {
       <NotesPortal />
       <OverlayController />
       <TreeNav />
-      <Suspense fallback={<PageSpinner/>}>
+      <Suspense fallback={<PageSpinner />}>
         <Switch>
           <Route path="/orgs/new" component={CreateOrgOverlay} />
           <Route path="/orgs/:orgID" component={SetOrg} />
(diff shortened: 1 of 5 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

hoorayimhelping:
This is solid, and an easy performance improvement is super welcome!

But in general, I'm skeptical there's any free lunch or any benefit without a tradeoff. What are the potential downsides to this? Might there be some conditions under which some routes load longer? If so, do we know what those conditions might be?

I guess I'm just trying to get a sense of the known unknowns and maybe try to map out the boundary of where the unknown unknowns exist.

Asalem1 (approved):
Great work!

121watts:
> This is solid, and an easy performance improvement is super welcome!
> 
> But in general, I'm skeptical there's any free lunch or any benefit without a tradeoff. What are the potential downsides to this? Might there be some conditions under which some routes load longer? If so, do we know what those conditions might be?
> 
> I guess I'm just trying to get a sense of the known unknowns and maybe try to map out the boundary of where the unknown unknowns exist.

Lazy loading does not happen instantly hence the need for the `fallback` prop to render some intermediate state as the `lazy` component is loaded.  What's nice about doing this on the route level is that switching between pages are generally expected to have a some loading state already.  

That all said, the reduction of bundle size served greatly outweighs the tiny extra loading time.

PR #519, opened 2021-01-14
feat: Unity usage page

Asalem1:
This PR introduces the usage page as a feature flagged entity within the UI. 

usage-page


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/usage/UsageToday.tsx
```diff
@@ -0,0 +1,119 @@
...
+// Types
+import {TimeRange} from 'src/types'
+
+interface StateProps {
+  selectedUsageID: string
+}
+
+type Props = StateProps
+
+const UsageToday: FC<Props> = ({selectedUsageID}) => {
```
genehynson, line 33: I think you can just do `FC<StateProps>` and get rid of `type Props = StateProps`

src/usage/UsageToday.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  const getTimeRangeLabel = () => {
+    switch (timeRange.type) {
+      case 'selectable-duration':
+        return timeRange.label
+      case 'duration':
+        return `from ${timeRange.lower} to now`
+      default:
+        return `from ${timeRange.lower} to ${timeRange.upper}`
```
genehynson, line 59: might want to format `.lower` and `.upper` to ISO

src/usage/UsageToday.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <Panel className="usage--panel">
+        <Panel.Header>
+          <h4 data-testid="rate-limits-header--timerange">{`Rate Limits ${timeRangeLabel}`}</h4>
+        </Panel.Header>
+        <Panel.Body
+          direction={FlexDirection.Column}
+          margin={ComponentSize.Small}
+          alignItems={AlignItems.Stretch}
+        >
+          {GRAPH_INFO.rate_limits.map(graphInfo => {
```
genehynson, line 98: nit: rename `rate_limits` -> `rateLimits`

genehynson (approved):
lgtm other than a couple nitpick comments 🎉

src/client/unity.yml
```diff
@@ -23,6 +23,60 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/History'
+        default:
+          description: Unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
```
121watts, line 79: should this go with the other `/account` endpoint?

src/types/billing.ts
```diff
@@ -4,85 +4,48 @@
...
 }
 
-export interface BillingNotifySettings extends GenBillingNotifySettings {
+export interface BillingDate extends GenBillingDate {
   status: RemoteDataState
 }
 
 // Current PayAsYouGo Props
 export interface Props {
+  isRegionBeta: boolean // Current FreePage Props
```
121watts, line 43: generally not a fan of booleans like this it locks us into two possible values.  something like `regionStatus` or `regionType` would allow for a `beta` tag or anything else we might want.

src/usage/UsageToday.tsx
```diff
@@ -0,0 +1,119 @@
...
+  )
+}
+
+const mstp = () => {
+  return {
+    selectedUsageID: 'Data In (MB)',
+  }
+}
+
+export default connect<Props>(mstp)(UsageToday)
```
121watts, line 119: why use this at all if it's hard coded?  if you plan on supplying to redux in the future why not use the `useState` hook instead of the `connect` HOC?

src/usage/reducers/index.ts
```diff
@@ -0,0 +1,138 @@
...
+      case 'SET_HISTORY_STATUS': {
+        if (!draftState.history?.status) {
+          draftState.history = {...draftState.history, status: action.status}
+          return
+        }
+        draftState.history.status = action.status
+        return
+      }
+    }
+  })
```
121watts, line 138: test for this reducer and actions might be nice.

121watts (approved):
some small suggestions and nits.  nice work man!  cruising right along

src/usage/UsagePage.tsx
```diff
@@ -0,0 +1,87 @@
...
+import UsageToday from 'src/usage/UsageToday'
+import RateLimitAlert from 'src/cloud/components/RateLimitAlert'
+import AlertStatusCancelled from 'src/billing/components/Usage/AlertStatusCancelled'
+import LimitChecker from 'src/cloud/components/LimitChecker'
+
+// Reducers
+import {
+  Action,
+  UsageState,
+  usageReducer,
```
subirjolly, line 16: @Elders of the InfluxData. I was wondering if we should think about denoting types/interfaces as tSomething or iSomething? `usageReducer` and `UsageReducer` were confusing. Although, there's a difference in the casing. 

Thought I should ask.
@Asalem1 @drdelambre @121watts @genehynson

src/usage/thunks/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  setBillingDateStatus,
+  setHistory,
+  setHistoryStatus,
+  Action,
+} from 'src/usage/reducers'
+
+// Types
+import {RemoteDataState} from 'src/types'
+
+export const getBillingDate = async (dispatch: Dispatch<Action>) => {
```
subirjolly, line 25: Was wondering about your thoughts on extracting the repitition out?
```
const process = async (cbSetStatus, cbGet, cbSet) => {
    try {
        dispatch(cbSetStatus(RemoteDataState.Loading))
        const resp = await cbGet()

        if (resp.status !== 200) {
        }
        dispatch(cbSet({...resp.data, status: RemoteDataState.Done}))
      } catch (error) {
        dispatch(cbSetStatus(RemoteDataState.Error))
        console.error(error)
      }
}
```

src/usage/UsageToday.tsx
Asalem1: Totally. I think this was residual from some of the pairing that I had done earlier. I'll cut it out

src/usage/UsagePage.tsx
Asalem1: Ya, definitely something to think about

drdelambre: i think it's bike shedding, but if you feel strongly about it, figure out a way to automate it. We have a hard time enforcing things like this if we cant. There's been calls to alpha sort all of the imports, and you might occasionally find a // Libraries, // Components, // Types format for imports. And everyone keeps at it for awhile, and then a PR or two without it slip in, and then new people join and copy pasta from one of those "bad" files, and that rule isn't transferred amongst the million other things people are expected to know right off the bat....

if you can automate it, i support your crusade.

src/usage/thunks/index.ts
Asalem1: I hear what you're getting at. I think that we might end up constraining ourselves later on with an API that's increasingly confined towards a specific pattern of expectations.

src/usage/UsagePage.tsx
subirjolly: https://en.wiktionary.org/wiki/bikeshedding

@drdelambre  Learned something. You have a valid point. Not sure if it's worth automating also.

subirjolly:
Resolve my comments when you feel like.

src/client/unity.yml
Asalem1: You bring up a good point. Truthfully I'm not entirely sold on keeping this in the long run since the `history` endpoint in Quartz currently makes a flux request based on certain user meta data (found in quartz) and the selected time range. The big "gotcha" that I have with straight moving the flux query to the UI right now would be because the flux queries query a specific bucket's aggregate data that Quartz is being written to. Figured I'd want to talk to the e-commerce team to try and see where it best fits.

My reasoning for leaving this API separate from the `/billing/account` one is because I'm not sure how large the CSVs being returned from the API would be, so I don't want to overfetch data that isn't being used and potentially slow the browser down if we're getting a couple of 28mb csvs returned.

src/types/billing.ts
Asalem1: I don't think we're actually using this anywhere.

src/usage/UsagePage.tsx
121watts: +1 on the bike shedding. Not worth the effort.

PR #523, opened 2021-01-15
feat: annotations api read update delete interface

### LINKED ISSUE #503: Annotations: API: Wire up RUD (Read, Update, Delete) operations
### Background
We've established a Create pattern for writing annotations. It's time to establish the other aspects of CRUD operations, Read, Update, Delete.

See #490

### Acceptance Criteria
- In the `annotations/api` directory, there are API methods for `read` `update` `delete` operations.
  - Follow a similar structure established by the `write` method.
  - Name them reasonably, you don't have to call them `readAnnotation`.
- There are tests for the `read`, `update`, and `delete`, operations
  - Similar to the tests for creating an annotation

ShmuelLotman:
Closes #503

Introduces the read, update, delete functionalities for the annotations api interface.

- [x] Rebased/mergeable
- [x] Tests pass


src/annotations/api/index.ts
```diff
@@ -1,8 +1,13 @@
...
-import {PostAnnotationResponse, PostAnnotationPayload} from 'src/types'
+import {
+  Annotation,
+  GetAnnotationResponse,
+  GetAnnotationPayload,
+  DeleteAnnotation,
+} from 'src/types'
+
+// Libraries
+import axios from 'axios'
```
Asalem1, line 10: nit: can we get a comment and a link to the issue that will remove axios once the API is developed?

src/annotations/api/index.ts
```diff
@@ -11,25 +16,111 @@ import {API_BASE_PATH} from 'src/shared/constants'
...
+  if (start) {
+    queryString += `${queryString.length ? '&' : '?'}start=${start}`
+  }
+
+  if (end) {
+    queryString += `${queryString.length ? '&' : '?'}end=${end}`
+  }
+
+  if (stickers) {
+    Object.keys(stickers).forEach((s: string) => {
```
Asalem1, line 63: ⭐

Asalem1 (approved):
Minor nit. Would also be curious if the grand plan is to query these API queries directly from the UI or if it'll be queried based on a thunk? My curiosity stems from how the errors and at what point we should be handling the errors

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/annotations/api/index.test.ts
```diff
@@ -1,52 +1,189 @@
...
+    it('retrieves annotations and returns them categorized by annotation stream', async () => {
+      mocked(axios.get).mockImplementationOnce(() => Promise.resolve({data}))
+      const response = await getAnnotation({
+        start: Date.now().toString(),
+        end: Date.now().toString(),
+        stream: 'Lambeau Field',
+      })
 
-      expect(response).toEqual(transformedData)
+      expect(response).toEqual(data)
```
jrenee42, line 118: if it's requesting the "Lambeau field" annotations, it gets the default too?
(that is what the code says)

that seems counter-intuitive to me.

src/types/annotation.ts
```diff
@@ -1,13 +1,27 @@
-export interface PostAnnotationResponse {
-  stream: string
-  endTime: string
-  startTime: string
+export interface Annotation {
+  end: string
+  start: string
```
jrenee42, line 3: giraffe uses startValue and stopValue

here we are using start and end

inconsistencies like this are not good.

src/annotations/api/index.test.ts
ShmuelLotman: No. If it requests Lambeau Field it will only get Lambeau Field annotations. I'll update the test to reflect that, but its because of the value in the mock needing to be changed, not because it doesn't filter by stream

src/types/annotation.ts
ShmuelLotman: Per @hoorayimhelping this is something that will be handled in the view layer not the api layer.

src/annotations/api/index.test.ts
```diff
@@ -1,52 +1,191 @@
...
+  getAnnotation,
+  formatAnnotationQueryString,
+  deleteAnnotation,
+  updateAnnotation,
+} from 'src/annotations/api'
 import axios from 'axios'
 
 jest.mock('axios')
 
+describe('query string formatting function', () => {
```
hoorayimhelping, line 15: this should be the first test in the new `utils` directory

src/annotations/api/index.test.ts
```diff
@@ -1,52 +1,191 @@
...
+            summary: 'blah blah',
+          },
+        ],
+      },
+    ]
 
-      expect(response).toEqual(transformedData)
+    it('retrieves annotations and returns them categorized by annotation stream', async () => {
+      mocked(axios.get).mockImplementationOnce(() =>
+        Promise.resolve({data: [data[0]]})
```
hoorayimhelping: do us a favor and change `data[0]` to something more self documenting, like:

```
const [lambeau] = data
```

also, you already know how I feel about variables called `data`

src/annotations/api/index.test.ts
```diff
@@ -1,52 +1,191 @@
...
+            summary: 'blah blah',
+          },
+        ],
+      },
+    ]
 
-      expect(response).toEqual(transformedData)
+    it('retrieves annotations and returns them categorized by annotation stream', async () => {
+      mocked(axios.get).mockImplementationOnce(() =>
+        Promise.resolve({data: [annotationResponse[0]]})
```
hoorayimhelping, line 112: I think it would make this test a little bit more self documenting and readable if you pulled this out into a variable like:

```ts
const [lambeauResponse] = annotationResponse
```

then down below it would be like:

```ts
expect(response).toEqual(lambeauResponse)
```

src/annotations/utils/formatQueryString.ts
```diff
@@ -0,0 +1,27 @@
...
+import {GetAnnotationPayload} from 'src/types'
+
+export const formatAnnotationQueryString = (
+  annotation: GetAnnotationPayload,
+  requestType?: string
+): string => {
+  const {stream, start, end, stickers} = annotation
+
+  const getAnnotationParams = new URLSearchParams(
+    JSON.parse(JSON.stringify({start, end}))
```
hoorayimhelping, line 11: what's the intent behind this? this doesn't do anything

```ts
const foo = {foo: 'bar'}
const bar = JSON.parse(JSON.stringify({foo: 'bar'}))

console.log(foo, bar) // {foo: "bar"}, {foo: "bar"}
```

src/annotations/utils/formatQueryString.ts
```diff
@@ -0,0 +1,27 @@
...
+    getAnnotationParams.append(streamParam, stream)
+  }
+
+  if (stickers) {
+    Object.keys(stickers).forEach((s: string) => {
+      getAnnotationParams.append(`stickerIncludes[${s}]`, stickers[s])
+    })
+  }
+
+  return `?${getAnnotationParams.toString()}`
```
hoorayimhelping, line 26: I'm not a fan of putting the `?` here. It lowers the reusability of this method. The assumption it makes is that whatever is passed in will always want a `?` prepended to it. This makes it unable to be composed - you always have to pass it the first part or all of the query string. 

```suggestion
  return `${getAnnotationParams.toString()}`
```

src/annotations/api/index.ts
```diff
@@ -1,35 +1,97 @@
...
+      stream,
+    },
+  ]
+}
+
+export const getAnnotation = async (
+  annotation: GetAnnotationPayload
+): Promise<GetAnnotationResponse[]> => {
+  const formattedQueryString = formatAnnotationQueryString(annotation)
+  const appendedURL = url + formattedQueryString
```
hoorayimhelping, line 48:
```suggestion
  const appendedURL = `${url}?${formattedQueryString}`
```

src/annotations/api/index.ts
```diff
@@ -1,35 +1,97 @@
...
+}
+
+export const deleteAnnotation = async (
+  deleteAnnotation: DeleteAnnotation
+): Promise<number> => {
+  const formattedQueryString = formatAnnotationQueryString(
+    deleteAnnotation,
+    'delete'
+  )
+  const res = await axios.delete(url + formattedQueryString)
```
hoorayimhelping, line 90:
```suggestion
  const res = await axios.delete(`${url}?${formattedQueryString}`)
```

hoorayimhelping (changes requested):
good work! two suggestions

hoorayimhelping (approved):

hoorayimhelping:
good work!

PR #527, opened 2021-01-15
feat(e2e-in-KinD): Remove all references to Cypress.baseUrl

ebb-tide:
The `cypress/e2e/shared/checks.test.ts`, `cypress/e2e/shared/notificationEndpoints.test.ts`, and `cypress/e2e/shared/notificationRules.test.ts` tests all make assertions that reference `Cypress.config().baseUrl`. 

Cypress deals with routing intelligently and is agnostic to whether or not there is a trailing `/` when routing. But using string interpolation to construct and assert on a url results in a situation where if the baseUrl ends with a `/` the route string asserted on can contain `//` but the url actually contains just one `/`. I've chosen to remove references to `Cypress.config().baseUrl`.  instead asserting the url string include the rest of the route.

### DIFF
--- a/cypress/e2e/shared/checks.test.ts
+++ b/cypress/e2e/shared/checks.test.ts
@@ -346,9 +346,7 @@ describe('Checks', () => {
             Cypress.minimatch(
               url,
               `
-                ${
-                  Cypress.config().baseUrl
-                }${orgs}/${id}${alerting}${checks}/*/edit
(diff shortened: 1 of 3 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

drdelambre (approved):

Issue #524, opened 2021-01-15
CSV Uploader Error Handling


Asalem1:
### Problem

Currently failures to write data using the CSV Uploader don't trigger an error to be caught and notified by the try / catch boundary. Therefore, when a user comes across an issue uploading their data (i.e. failure to write), the loader keeps going and notifies the user that everything was successful.

### Solution

Add better error handling to the CsvUploader so that a user gets an actionable understanding that:

1. Something went wrong
2. The reason something went wrong
3. (hopeful on this last one): allow for some action to help resolve the issue (link to docs, button to fix the CSV format, whatever)

PR #533, opened 2021-01-20
fix: CSV uploader reports failures and parses special characters

### LINKED ISSUE #525: CSV Uploader Escape White Space
### Problem

The current CSV uploader falls apart if a tag or field have white space included in their values

### Solution

Update the way the LP is constructing tags and fields in order to make the LP valid. For reference on white spacing in LP, please see documentation here:

https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/#examples-of-special-characters-in-line-protocol

genehynson:
Closes #525, #524

Some improvements to the CSV uploader:

1) If the upload fails, the wizard will report the failure. This includes async failures like if the write API requests fails. Some explanation for the failure is provided in the notification.
2) If csv data contains special characters in tag values, these characters are escaped (this only applies to tag columns of string type). 
3) If csv _value column is of string type, it adds quotes around the values

See these example CSVs for trying this out:
[example csv.zip](https://github.com/influxdata/ui/files/5845370/example.csv.zip)

special_chars

error_message

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/buckets/components/context/csvUploaderProvider.tsx
+++ b/src/buckets/components/context/csvUploaderProvider.tsx
@@ -68,13 +68,24 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
   const resetUploadState = (): void =>
     setUploadState(RemoteDataState.NotStarted)
 
-  const handleError = (error): void => {
+  const handleError = (error: Error): void => {
     setUploadState(RemoteDataState.Error)
     reportErrorThroughHoneyBadger(error, {
       name: 'uploadCsv function',
     })
     const message = getErrorMessage(error)
     dispatch(notify(csvUploaderErrorNotification(message)))
+  }
+
+  const parseTagStringValue = (value: string): string => {
+    // Replaces special characters in accordance to Line Protocol standards
+    // https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/#special-characters
+    return value
+      ? value
+          .replace(/,/g, '\\,')
+          .replace(/ /g, '\\ ')
+          .replace(/=/g, '\\=')
+      : '""'
   }
 
   const uploadCsv = useCallback(
@@ -155,7 +166,14 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
             value = table.getColumn('_value')?.[i] ?? field ?? ''
             tags = columns
               .filter(col => !!table.getColumn(col)[i])
-              .map(col => `${col}=${table.getColumn(col)[i]}`)
+              .map(
+                col =>
+                  `${col}=${
+                    table.getColumnType(col) === 'string'
+                      ? parseTagStringValue(table.getColumn(col)[i] as string)
+                      : table.getColumn(col)[i]
+                  }`
+              )
               .join(',')
               .trim()
               .replace(/(\r\n|\n|\r)/gm, '')
@@ -178,12 +196,13 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
             pendingWrites.push(resp)
             counter++
           }
-
           chunk = ''
           Promise.all(pendingWrites)
             .then(values => {
               if (values.find(v => v.status >= 400)) {
-                throw new Error(`The CSV data could not be written to the bucket`)
+                throw new Error(
+                  `The CSV data could not be written to the bucket`
+                )
               }
               setUploadState(RemoteDataState.Done)
             })
--- a/src/buckets/components/context/csvUploaderProvider.tsx
+++ b/src/buckets/components/context/csvUploaderProvider.tsx
@@ -77,26 +77,15 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
     dispatch(notify(csvUploaderErrorNotification(message)))
   }
 
-  const parseTagStringValue = (value: string): string => {
-    // Replaces special characters in accordance to Line Protocol standards
-    // https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/#special-characters
-    return value
-      ? value
-          .replace(/,/g, '\\,')
-          .replace(/ /g, '\\ ')
-          .replace(/=/g, '\\=')
-      : '""'
-  }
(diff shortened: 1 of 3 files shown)


src/buckets/components/csvUploader/CsvUploaderError.tsx
```diff
@@ -0,0 +1,27 @@
...
+  <div className="line-protocol--spinner">
+    <p
+      data-testid="line-protocol--status"
+      className="line-protocol--status error"
+    >
+      Failed to Write Data
+    </p>
+    <SparkleSpinner loading={RemoteDataState.Error} sizePixels={220} />
+    <p className="line-protocol--status error">
+      Make sure your CSV is{' '}
```
genehynson, line 15: Is there a better way to add a space around the text in the `<a/>` tag?

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/buckets/components/csvUploader/CsvUploaderError.tsx
Asalem1: The only other options would be to add a `&nbsp;` instead, or set specific padding onto the `a` tag

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -68,12 +68,37 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
...
+      : '""'
+  }
+
   const uploadCsv = useCallback(
     (csv: string, bucket: string) => {
       setUploadState(RemoteDataState.Loading)
       setTimeout(() => {
         try {
           const {table} = fromFlux(csv)
+          if (!table.length) {
```
Asalem1, line 97: we may also want to handle whether an error is returned from the parser, something like:

```
if (!table.length || error) {
}
```

where error is defined like so:

```
const {table, error} = fromFlux(csv)

File: src/buckets/components/context/csvUploaderProvider.tsx [hunk 3]
```diff
@@ -68,12 +68,37 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
...
+            throw new Error(
+              `The CSV could not be parsed. Make sure to include _measurement and _field columns`
```
Asalem1, line 99: We should make this more actionable. There are several reasons that a CSV parser might fail here, maybe we can be a bit more specific by saying that the CSV must be in annotation format with a link to the docs?

https://docs.influxdata.com/influxdb/cloud/reference/syntax/annotated-csv/

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -155,24 +194,29 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
...
-
           chunk = ''
-          Promise.all(pendingWrites).finally(() => {
-            setUploadState(RemoteDataState.Done)
-          })
+          Promise.all(pendingWrites)
+            .then(values => {
+              if (values.find(v => v.status >= 400)) {
+                throw new Error(
+                  `The CSV data could not be written to the bucket`
```
Asalem1, line 210:
```suggestion
                  `Looks like some of the CSV data could not be written to the bucket. Please make sure that CSV was in Annotated Format https://docs.influxdata.com/influxdb/cloud/reference/syntax/annotated-csv/`
```

Asalem1: Obviously, that's a bit long-winded, but there might be a situation where some of the data successfully writes and some doesn't

src/buckets/components/csvUploader/CsvUploaderError.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <a
+        target="_blank"
+        href="https://docs.influxdata.com/influxdb/v2.0/write-data/developer-tools/csv/"
```
Asalem1, line 18:
```suggestion
        href="https://docs.influxdata.com/influxdb/v2.0/write-data/developer-tools/csv/#csv-annotations"
```

Asalem1 (approved):
A few points but overall this looks great! Would love some screenshots of how it looks & how it handles that CSV

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -138,9 +164,22 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
             }
             time = table.getColumn('_time')?.[i] ?? Date.now()
             value = table.getColumn('_value')?.[i] ?? field ?? ''
+            // Adds quotes to values if _value is of string type
+            // https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/#quotes
+            value =
+              table.getColumnType('_value') === 'string' && value
+                ? '"' + value + '"'
```
Asalem1, line 171: To this point, we can probably just use template strings

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -138,9 +164,22 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
...
+                : value
             tags = columns
               .filter(col => !!table.getColumn(col)[i])
-              .map(col => `${col}=${table.getColumn(col)[i]}`)
+              .map(
+                col =>
+                  `${col}=${
+                    table.getColumnType(col) === 'string'
+                      ? parseTagStringValue(table.getColumn(col)[i] as string)
```
Asalem1, line 179: This is solid, and under normal circumstances I wouldn't bat an eyelash at having an abstracted function that serves a specific purpose (in this case, formatting a value to align with our write endpoint expectations). However, since we are looping over this function N times, we are creating N instances of this function. In this case, readability provides a hit to performance. Would love it if we could simply inline the regex with a comment

genehynson: nice find!

PR #541, opened 2021-01-21
test: turn off TypeScript warnOnly

hoorayimhelping:
closes idpe 9623

TypeScript errors were temporarily turned to warnings. After reversing this back, some of the tests written since `warnOnly` was on have failed. This goes through and fixes the errors in those tests.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/jestSetup.ts
+++ b/jestSetup.ts
@@ -23,7 +23,7 @@ window.flushAllPromises = async () => {
 
 // mocks and stuff
 fetchMock.enableMocks()
-jest.mock('honeybadger-js', () => () => null)
+jest.mock('src/shared/utils/errors')
 
 // cleans up state between @testing-library/react tests
 afterEach(() => {
--- a/src/templates/components/CommunityTemplateInstallOverlay.test.tsx
+++ b/src/templates/components/CommunityTemplateInstallOverlay.test.tsx
@@ -9,7 +9,6 @@ import {mocked} from 'ts-jest/utils'
 jest.mock('src/cloud/utils/reporting')
 jest.mock('src/resources/components/GetResources')
 jest.mock('src/shared/actions/notifications')
-jest.mock('src/shared/utils/errors')
 
 jest.mock('src/templates/selectors/index.ts', () => {
   return {
--- a/jest.config.js
+++ b/jest.config.js
@@ -24,7 +24,6 @@ module.exports = {
       tsConfig: 'tsconfig.test.json',
       diagnostics: {
         ignoreCodes: [6133, 6192], // ignore unused variable errors
-        warnOnly: true
       },
     },
   },
--- a/src/templates/containers/CommunityTemplatesIndex.test.tsx
+++ b/src/templates/containers/CommunityTemplatesIndex.test.tsx
@@ -9,7 +9,6 @@ import {mocked} from 'ts-jest/utils'
 jest.mock('src/cloud/utils/reporting')
 jest.mock('src/resources/components/GetResources')
 jest.mock('src/shared/actions/notifications')
-jest.mock('src/shared/utils/errors')
 
 jest.mock('src/templates/api', () => {
   return {
--- a/src/variables/reducers/index.test.ts
+++ b/src/variables/reducers/index.test.ts
@@ -5,7 +5,7 @@ import {variablesReducer} from 'src/variables/reducers'
 import {moveVariable} from 'src/variables/actions/creators'
 
 // Types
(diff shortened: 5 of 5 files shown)


Asalem1 (dismissed):
That's really strange that the tests are passing after removing the mock data. Basically the intention of those reducer tests was to ensure that specific property types were being set correctly without actually checking each property (besides the status). It stands to reason that the tests still pass since technically they're still equal to the input object, but I guess we should update the tests to check specific property types

src/billing/reducers/mockBillingData.ts
```diff
@@ -62,11 +61,6 @@ export const mockPaymentMethods: PaymentMethod[] = [
 ]
 
 export const mockRegion: Region = {
-  title: 'EU Central',
-  isBeta: false,
-  isAvailable: true,
-  provider: 'aws',
-  region: 'eu-central',
```
Asalem1, line 69: Rather than removing the mock data from the mocks, we should update the test to check for these property types

src/usage/reducers/mockUsageData.ts
```diff
@@ -11,31 +11,13 @@ export const mockAccount: Account = {
   type: 'free',
   updatedAt: '01/01/2021',
   users: [],
-  pricingVersion: 4,
   zuoraAccountId: 'z-id-me',
 }
 
 export const mockBillingStart: BillingDate = {
-  date: '01/01/2021',
-  time: '1/20/2021, 3:15:00 PM',
```
Asalem1, line 20: Same as the previous comment

src/usage/reducers/mockUsageData.ts
```diff
@@ -11,31 +11,13 @@ export const mockAccount: Account = {
...
-  billingStats: `#group,false,false,true,true,false
-#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,long
-#default,query_count,,,,
-,result,table,_start,_stop,query_count
-,,0,2021-01-13T22:00:24.064007058Z,2021-01-15T22:00:24.064007058Z,1954`,
-  usageStats: `#group,false,false,true,false,false
-#datatype,string,long,string,dateTime:RFC3339,long
-#default,limits,,,,
-,result,table,_field,_time,_value
-,,0,limited_write,2021-01-13T22:05:00Z,0`,
```
Asalem1, line 40: Same as the previous comment

Asalem1 (changes requested):

hoorayimhelping:
>we should update the test to check for these property types

@Asalem1 Agreed. I was actually hoping you could help me with that. I tried to add some checks, but I was a bit lost between the tests and the TypeScript between unity and billing and couldn't get it to work. Would you mind taking a look at those tests when you have a minute?

Asalem1:
> > we should update the test to check for these property types
> 
> @Asalem1 Agreed. I was actually hoping you could help me with that. I tried to add some checks, but I was a bit lost between the tests and the TypeScript between unity and billing and couldn't get it to work. Would you mind taking a look at those tests when you have a minute?

Definitely. Let me update the tests on master and push that up. i'll keep you posted on the progress

hoorayimhelping:
@Asalem1 if it helps, these errors should show up in the console when you run `yarn test`, they'll just be `WARN` level.

jestSetup.ts
```diff
@@ -23,7 +23,7 @@ window.flushAllPromises = async () => {
 
 // mocks and stuff
 fetchMock.enableMocks()
-jest.mock('honeybadger-js', () => () => null)
+jest.mock('src/shared/utils/errors')
```
hoorayimhelping, line 26: fixes that one error about annotations on the flux response not existing

src/templates/components/CommunityTemplateInstallOverlay.test.tsx
```diff
@@ -9,7 +9,6 @@ import {mocked} from 'ts-jest/utils'
 jest.mock('src/cloud/utils/reporting')
 jest.mock('src/resources/components/GetResources')
 jest.mock('src/shared/actions/notifications')
-jest.mock('src/shared/utils/errors')
```
hoorayimhelping, line 12: no longer necessary since this call is being made in jestSetup

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

zoesteinkamp (approved):
yay no more typescript warnonly

PR #542, opened 2021-01-22
fix: csv parser understand multi-measurements

### LINKED ISSUE #538: csv uploader: fails to parse csv with multiple measurements
If you use the CSV uploader to upload a CSV file with multiple measurement columns, it will fail. 

Attached is a CSV that recreates this issue. 

[multiple measurements.csv.zip](https://github.com/influxdata/ui/files/5852424/multiple.measurements.csv.zip)

Error from API: 
```
unable to parse 'baywheels_meta,host=ip-192-168-1-6.ec2.internal,station_id=412,lat=37.7328128005294,lon=-122.43414044380188,name=Glen\\ Park\\ BART\\ Station,region_id=3,short_name=SF-W19 external_id=external_id 1611262738549000000': invalid boolean
```

The CSV to LP converter should adhere to line protocol standards for multiple measurements as defined here: https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/

genehynson:
Closes #538 

Allows CSV parser to handle multiple measurement columns when there are multiple types of the `_value` column involved. 

See https://github.com/influxdata/giraffe/blob/master/giraffe/src/utils/fromFlux.ts#L62

Screen Shot 2021-01-21 at 3 56 22 PM

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/buckets/components/context/csvUploaderProvider.tsx
+++ b/src/buckets/components/context/csvUploaderProvider.tsx
@@ -113,6 +113,7 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
           let field: any = ''
           let time: any = ''
           let value: any = ''
+          let valueColumn: string = '_value'
           let tags: any = ''
           let line: any = ''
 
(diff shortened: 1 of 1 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

genehynson:
Giraffe issue to improve the UX of `fromFlux` in this scenario: https://github.com/influxdata/giraffe/issues/455

Asalem1 (commented):
You’re absolutely crushing it. We should totally write some e2e tests on these to make sure all your hard work stays solid

src/buckets/components/context/csvUploaderProvider.tsx
```diff
@@ -152,11 +153,20 @@ export const CsvUploaderProvider: FC<Props> = React.memo(({children}) => {
...
+            table.columnKeys
+              // Matches _value, _value ('number'), _value ('string'), etc. to find value
+              // https://github.com/influxdata/giraffe/blob/master/giraffe/src/utils/fromFlux.ts#L62
+              .filter(c => /_value( \('\w*'\))?/g.test(c))
+              .forEach(c => {
+                if (table.getColumn(c)?.[i]) {
+                  value = table.getColumn(c)[i]
+                  valueColumn = c
+                }
+              })
```
Asalem1, line 165: I think we might be able to simplify this by simply passing in the datatype as the second parameter in the `table.getColumn` function:

https://github.com/influxdata/giraffe/blob/master/giraffe/src/utils/newTable.ts#L25

Asalem1 (approved):

src/buckets/components/context/csvUploaderProvider.tsx
Asalem1: Turns out I was running and giraffe is 💩  with this.  Looks good to me!

Asalem1 (approved):

PR #552, opened 2021-01-25
fix(performance): prevents unneeded renders on identical props

genehynson:
Closes https://github.com/influxdata/idpe/issues/9584

ErrorBoundary and Vis components were re-rendering on every interaction, even though their
props were identical. This caused performance issues with query editor. 

I added React.memo to both of these components and provided a custom comparison function that uses Lodash's `isEqual` to perform a deep comparison between the props. Shallow comparisons provided by `PureComponent` or simply `===` does not do the trick in this case.

To verify the improvement, upload the attached CSV to a bucket, query data from the last 30 days and show the "raw data view". There will be 120k rows. Start typing in the flux query editor and you shouldn't notice any slow down over time. 

Without this improvement, the table was being re-rendered on every keystroke and caused noticeable performance issues when typing in the query editor over time.

[very large dataset.csv.zip](https://github.com/influxdata/ui/files/5869088/very.large.dataset.csv.zip)

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/shared/components/ErrorBoundary.tsx
+++ b/src/shared/components/ErrorBoundary.tsx
@@ -1,5 +1,6 @@
 // Libraries
 import React, {Component, ErrorInfo} from 'react'
+import _ from 'lodash'
 
 // Components
 import DefaultErrorMessage from 'src/shared/components/DefaultErrorMessage'
@@ -35,6 +36,8 @@ class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
       component: parseComponentName(errorInfo),
     })
   }
+
+  public shouldComponentUpdate = next => !_.isEqual(this.props, next)
 
   public render() {
     const {error} = this.state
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/shared/components/ErrorBoundary.tsx
```diff
@@ -1,5 +1,6 @@
 // Libraries
 import React, {Component, ErrorInfo} from 'react'
+import _ from 'lodash'
```
drdelambre, line 3: be more specific on the imports, like `import {isEqual} from 'lodash'` instead. every couple months we get a "LETS REMOVE ALL OF THE LODASH" bug under the skin, and having the explicit mappings makes it easier to whoever takes it on

genehynson: good call!

subirjolly (approved):

PR #558, opened 2021-01-25
fix: for #19216

jrenee42:
took part of palak's fix;  just needed the part that  zeroed out the group clause if values are absent.

Closes https://github.com/influxdata/influxdb/issues/19216

- [ ] Rebased to most recent master
- [ ] Tests pass; they passed on the earlier commit and circle-ci had a problem on the most recent rebase; but they should pass.

### DIFF
--- a/src/timeMachine/utils/queryBuilder.ts
+++ b/src/timeMachine/utils/queryBuilder.ts
@@ -165,6 +165,12 @@ const convertTagsToFluxFunctionString = function convertTagsToFluxFunctionString
   }
 
   if (tag.aggregateFunctionType === 'group') {
+    // if group is selected, but there are no values, don't return anything.
+    // same behavior as when 'filter' is selected but no values are chosen
+    if (!tag.values.length) {
+      return ''
+    }
+
     const quotedValues = tag.values.map(value => `"${value}"`) // wrap the value in double quotes
 
     if (quotedValues.length) {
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/timeMachine/utils/queryBuilder.ts
```diff
@@ -165,6 +165,12 @@ const convertTagsToFluxFunctionString = function convertTagsToFluxFunctionString
   }
 
   if (tag.aggregateFunctionType === 'group') {
+    // if group is selected, but there are no values, don't return anything.
+    // same behavior as when 'filter' is selected but no values are chosen
+    if (!tag.values.length) {
+      return ''
+    }
```
hoorayimhelping, line 172: what do you think about adding a test for this case? there's already a test file written for this file.

src/timeMachine/utils/queryBuilder.ts
```diff
@@ -165,6 +165,12 @@ const convertTagsToFluxFunctionString = function convertTagsToFluxFunctionString
   }
 
   if (tag.aggregateFunctionType === 'group') {
+    // if group is selected, but there are no values, don't return anything.
```
Palakp41, line 168: Can we remove comments from here? We usually don't commit code with comments.

Palakp41 (approved):
Lgtm!

src/timeMachine/utils/queryBuilder.ts
jrenee42: added

Issue #376, opened 2020-12-01
Annotations MVP: redux / api / utils


kristinarobinson:
Background
We want a nice, clean interface with the backend and a consistent directory structure.

Here's an example of what the final directory structure might look like. If there's nothing to put in a directory, don't make an empty one.

src/annotations/
├── actions
├── api
├── components
├── containers
├── reducers
├── thunks
└── utils
We'll be implementing the following routes:

Routes
// KEY
METHOD - PATH : BODY ; RESPONSE

GET    - /annotations/:id : ; Annotation                             // get single annotation
PUT    - /annotations/:id : UpdateAnnotation ; Annotation      // create/override an individual annotation's start/stop time, title, description
DELETE - /annotations/:id : ; HTTP_WHATEVER                       // delete an annotation

Acceptance Criteria
annotations are treated as Resources
we can make requests and see the redux store updated for the following routes:
get a single annotation
create a single annotation
update an annotation
delete an annotation

NOTE:  subject to API verification

PR #562, opened 2021-01-27
feat(annotations): add redux

hoorayimhelping:
Connects #376

Adds the first function action creator, thunk, and reducer to the annotations redux store.

- Builds on the work of Jill and Slotman
- Lays the groundwork for future redux stuff, probably
- About that laying groundwork: adds a reducer test

here's a picture of redux tools showing the reducer being called via a call to the thunk that was added in `GetResource`
Screen Shot 2021-01-26 at 4 00 12 PM

- [x] Rebased/mergeable
- [x] Tests pass


src/annotations/api/index.ts
```diff
@@ -41,9 +41,24 @@ export const writeAnnotation = async (
...
+
+  if (res.status >= 300) {
+    throw new Error(res.data?.message)
+  }
+
+  return res.data.map((retrievedAnnotation: AnnotationStream) => ({
+    stream: retrievedAnnotation.stream,
+    annotations: retrievedAnnotation.annotations,
+  }))
+}
```
hoorayimhelping, line 57: we'll want to figure out the proper API here vis-a-vis `getAnnotation` and `getAnnotations`

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/annotations/actions/thunks.ts
```diff
@@ -0,0 +1,15 @@
+import {getAnnotations} from 'src/annotations/api'
+import {Dispatch} from 'react'
+
+import {
+  setAnnotations,
+  Action as AnnotationAction,
+} from 'src/annotations/actions/creators'
+
+export const fetchAndSetAnnotations = () => async (
```
ShmuelLotman, line 9: Love the naming here. I dislike it when fetch operations have a fetch<thing> call and then a set<thing> action... gets confusing to look at before coffee.

ShmuelLotman (approved):
Approving. This looks good to me. Thanks for bringing this across the finish line!

src/annotations/reducers/index.test.ts
```diff
@@ -0,0 +1,56 @@
+import {annotationsReducer, initialState} from 'src/annotations/reducers'
+
+const mockAnnotations = [
+  {
+    stream: 'default',
+    annotations: [
+      {
+        summary: 'hi',
+        end: '2021-01-21T00:05:51Z',
```
jrenee42, line 9: note that the 'end' and 'start' are strings.

the plot is expecting numbers (milliseconds) not date strings.
unless we want to store them like that and do conversions.

besides that the redux looks good.

hoorayimhelping: created an issue to address this: #570

PR #567, opened 2021-01-28
feat: Notebooks uses Flows API

genehynson:
Closes https://github.com/influxdata/idpe/issues/9565

### Flows API integration!

Flows will save, update, and delete from both localstorage and the API. Updates happen in a pool, every few seconds. Existing flows stored in localstorage will be migrated to the API automatically.

To enable this feature, use feature flag `notebooks-api` and make sure you're on the latest commit for monitor-ci. `make restart NODE=influxdb` is required to start `flowd`.

~NOTE: there is currently a bug with the `PATCH` flow API that's causing the Notebook to not save after creation. Compute team is aware of this and is working on a fix.~ <-- This issue is resolved.

- [x] Rebased/mergeable
- [x] Tests pass


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/flows/context/api.tsx
```diff
@@ -0,0 +1,132 @@
...
+  data: {},
+}
+let stagedFlow: PatchApiV2privateFlowsOrgsFlowParams = DEFAULT_API_FLOW
+let reportDecayTimeout = null
+let reportMaxTimeout = null
+
+const REPORT_DECAY = 2000 // number of miliseconds to wait after last event before sending
+const REPORT_MAX_WAIT = 5000 // max number of miliseconds to wait between sends
+
+export const pooledUpdateAPI = (flow: PatchApiV2privateFlowsOrgsFlowParams) => {
```
drdelambre, line 28: not for this PR, but i just had to write a simple debounce for my pr too. that means we have one for the query builder, one for notebooks, one for events, and one here. should we make a ticket to make... just.. one?

src/flows/context/flow.list.tsx
```diff
@@ -48,17 +70,14 @@ export const DEFAULT_CONTEXT: FlowListContextType = {
...
   change: (_id: string) => {},
+  getAll: () => {},
   currentID: null,
 } as FlowListContextType
 
 export const FlowListContext = React.createContext<FlowListContextType>(
   DEFAULT_CONTEXT
 )
 
-// NOTE: these have no utility, i'm just confused on how we are going to be getting
```
drdelambre, line 58: thanks for finding some utility!

src/flows/context/flow.list.tsx
```diff
@@ -181,16 +219,27 @@ export const FlowListProvider: FC = ({children}) => {
       name: flow.name,
       range: flow.range,
       refresh: flow.refresh,
-      data: flow.data.serialize(),
-      meta: flow.meta.serialize(),
+      data: flow.data.serialize ? flow.data.serialize() : flow.data,
```
drdelambre, line 222: didn't get to the bottom of this eh? for the life of me, i cant see why this condition shows up, which means it's prolly a reference error somewhere

src/flows/context/api.tsx
genehynson: yes and done: https://github.com/influxdata/ui/issues/572

drdelambre (approved):
noice. we might want to have a talk about either supporting optimistic rendering on the engineering side, or pushing for more spinners on the product side. maybe even a little spinning icon in the bottom left of the screen to indicate that system resources are syncing in the background like a video game.. idk, seems like we're injecting invisible latency here but i might be reading it wrong. can you type with the internet turned off on your browser?

src/flows/context/flow.list.tsx
genehynson: yeah I tried to hunt it down. You can reproduce it by creating a new Flow with this feature flag turned on. Figured this solution will work for now and we can try and solve it down the road

genehynson:
> noice. we might want to have a talk about either supporting optimistic rendering on the engineering side, or pushing for more spinners on the product side. maybe even a little spinning icon in the bottom left of the screen to indicate that system resources are syncing in the background like a video game.. idk, seems like we're injecting invisible latency here but i might be reading it wrong. can you type with the internet turned off on your browser?

Good idea. So I turned on "fast 3g" connection in chrome debugging tools and played around with it a bit. The UI JS for the UI took ages to load but after it did, the app experience felt unchanged. All the Flows network requests completed in 500-600ms so the experience doesn't really change at all. I also performed a migration of 3 Notebooks from local to the API (highlighted in red) and that entire operation took ~600ms since all the requests are run in parallel. I spoke with Brett about if we should be concerned for the APIs to be able to handle these migration requests in parallel and he said it absolutely should not be a problem. 

Even after creating 20 Notebooks, the GET requests still took under 600ms. 

One improvement we could make is to not wait on the POST request when creating the notebook and just update the id later. That would make creating notebooks feel snappier with a slow connection. Let me play around with that.

Screen Shot 2021-01-29 at 4 34 27 PM

PR #568, opened 2021-01-28
feat: query builder

drdelambre:
Peek 2021-01-27 16-37

While we figure out if people like notebooks, and the effect of the new metric selector exploration on their forming of that view point, i figured it'd be good to hedge our bets on that interface by adding back in everyone's favorite query builder interface to notebooks.


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

genehynson (approved):

PR #569, opened 2021-01-28
feat(experiment): empty state prompts to load data

mavarius:
Experiment code for https://github.com/influxdata/cloud-growth/issues/8

An experiment where users who have not yet loaded data into the application will see a prompt on the resource page empty states to load data instead of to create that resource.

#### Original: Create resource empty state.
image

#### Variant: Load Data empty state.
image

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/src/cloud/components/experiments/DatalessEmptyState.tsx
+++ b/src/cloud/components/experiments/DatalessEmptyState.tsx
@@ -1,15 +1,18 @@
 // Libraries
-import React, {FC} from 'react'
+import React, {FC, useEffect, useState} from 'react'
 import {useHistory} from 'react-router-dom'
 import {connect, ConnectedProps} from 'react-redux'
 import {get} from 'lodash'
 
 // Components
 import {
+  Button,
   EmptyState,
+  TechnoSpinner,
   ComponentSize,
   ComponentColor,
-  Button,
+  RemoteDataState,
+  SpinnerContainer,
 } from '@influxdata/clockface'
 
 // Utils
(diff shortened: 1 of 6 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/dashboards/components/dashboard_index/DashboardsTableEmpty.tsx
```diff
@@ -32,18 +36,48 @@ const DashboardsTableEmpty: FC<ComponentProps> = ({
...
+            onSelectTemplate={summonImportFromTemplateOverlay}
+            resourceName="Dashboard"
+            canImportFromTemplate={ALLOW_IMPORT_FROM_TEMPLATE}
+          />
+        </EmptyState>
+      }
+      variants={[
+        <GetResources
+          resources={[ResourceType.Buckets]}
+          key="9H85dx92QTylCNHIL9Y5Sg-1"
```
genehynson, line 58: What is this key for?

src/cloud/components/experiments/DatalessEmptyState.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const DatalessEmptyState: FC<Props> = ({orgID, buckets, children}) => {
+  const history = useHistory()
+
+  const handleClientLibrary = (): void => {
+    history.push(`/orgs/${orgID}/load-data/sources`)
+  }
+
+  // TODO: Replace with data loading flag from api if experiment is successful
+  if (buckets.find(bucket => bucket.type === 'user')) {
```
genehynson, line 37: What about the demo data bucket "Website Monitoring" - is that one of type 'user'? I honestly don't know

genehynson: Ok I checked - "Website Monitoring" demo bucket also has `.type = 'user'` so this logic is good 👍

genehynson (approved):
lgtm! I'm not sure what that fancy "key" is but I guess it has something to do with the Google experiment component and is necessary for this

src/dashboards/components/dashboard_index/DashboardsTableEmpty.tsx
mavarius: Because it's an array of elements react complains if they don't have a key.

src/cloud/components/experiments/DatalessEmptyState.tsx
mavarius: Demo data buckets are of type `demoData` however this does remind me that we automatically create a bucket for the user when they sign up and that initial bucket is of type `user` so this is a problem.

src/dashboards/components/dashboard_index/DashboardsTableEmpty.tsx
genehynson: oh but in that case, the key should be unique for each element, right?

src/cloud/components/experiments/DatalessEmptyState.tsx
genehynson: Hmm interesting because when I looked at the API response in cloud2 for buckets, it looks like the type is set to `user`. But it's probably good to handle the other cases too. 
image

src/dashboards/components/dashboard_index/DashboardsTableEmpty.tsx
genehynson: oh I see in this case there is only 1 element in the array. Can we maybe change the `key` value to be something else? It looks suspicious being that random string. Even just a `1` should do

src/cloud/components/experiments/DatalessEmptyState.tsx
mavarius: Oh, that's interesting mine says type: "demoData". Seems this isn't a reliable check. I'll need to find a better way to approach this.
image

mavarius: I've modified this now to get buckets of type "user" and check their cardinality to see if there's data in them or not.

genehynson: So my concern now is that we're going to potentially charge the user every time this component is mounted for a query poking at each of there buckets 😢

genehynson (approved):
lgtm!

src/cloud/components/experiments/DatalessEmptyState.tsx
mavarius: This check will stop querying buckets immediately when it finds one with data so this would only be true if they had many buckets created that were all empty. I have, however, added an additional guard to skip the query and assume they have data if they have more than 4 buckets created.

PR #574, opened 2021-01-29
chore: remove most string references of defbuck and add wrapped aliases instead

joelfried:
Makes progress on 521 - going to make one PR per string alias added so as to keep the changes more manageable.

This adds two aliases for e2e tests: cy.get('@defaultBucket') and cy.get('@defaultBucketListSelector').  The former returns "defbuck" and the latter "selector-list defbuck" but is defined so that if we changed the default bucket (in commands.ts) it would flow through.  These two aliases simplify testing code down considerably.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cypress/e2e/oss/onboarding.test.ts
+++ b/cypress/e2e/oss/onboarding.test.ts
@@ -79,7 +79,9 @@ describe('Onboarding', () => {
     cy.getByTestID('input-field--password').type(Cypress.env('password'))
     cy.getByTestID('input-field--password-chk').type(Cypress.env('password'))
     cy.getByTestID('input-field--orgname').type(Cypress.env('org'))
-    cy.getByTestID('input-field--bucketname').type(Cypress.env('bucket'))
+    cy.get<string>('@defaultBucket').then((defaultBucket: string) => {
+      cy.getByTestID('input-field--bucketname').type(defaultBucket)
+    })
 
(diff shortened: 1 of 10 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

cypress/e2e/oss/onboarding.test.ts
```diff
@@ -14,7 +15,7 @@ describe('Onboarding Redirect', () => {
 describe('Onboarding', () => {
   beforeEach(() => {
     cy.flush()
-
+    cy.wrapEnvironmentVariablesForOss()
```
drdelambre, line 18: should we move this into the flush function? or no?

joelfried: The flush function is used before the cloud e2e tests as well and the variables aren't the same at this point for OSS and Cloud because on the cloud side it gets the default org dynamically after logging in.  So I think these are better kept separate as the aliases on the cloud side happen on signin and I wouldn't want something to get added to OSS that isn't overwritten by cloud causing confusion down the line.

I do see a potential for a big performance gain in the flush function though so keep your eyes peeled for that :)

drdelambre (approved):

PR #575, opened 2021-01-29
feat: Geo Map Type Customization Options 

ShmuelLotman:
Closes #9577

This PR introduces the initial implementation for the geo map type customization options. While the ticket listed more options than are available here, this is because certain types and options were more difficult to navigate than initially anticipated and thus have been moved to future tickets that will follow this initial work
- [x] Rebased/mergeable
- [x] Tests pass


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,126 @@
...
+      return null
+  }
+}
+
+const GeoOptions: FC<GeoViewProperties> = props => {
+  const dispatch = useDispatch()
+  const handleSelectGroupClick = (activeOption: string): void => {
+    dispatch(setMapType(mapTypeHash[activeOption]))
+  }
+  const handleLatLongSelect = (latLonVal: number, type: string = 'lat') => {
```
hoorayimhelping, line 41: not a huge fan of this type of function design. we have to pass in a value that represents either a latitude or a longitude depending on the second argument that sometimes gets passed in.

much clearer illustration of intent (and future use) to do something like this:

```ts
handleSelectLatitude(latitude: number) =>
  dispatch(setLatitude(latitude))

handleSelectLongitude(longitude: number) =>
  dispatch(SetLongitude(longitude))
```

then when it's called, you can just call 
```ts
  onChange={event => selectLatitude(parseFloat(event.target.value))}
```

this way, in the future, if there are some fancy business rules around how we select latitude, like say we have to apply a projection to it, we can easily do that without changing any code that consumes this function.

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,126 @@
...
+    type === 'lon'
+      ? dispatch(setLongitude(latLonVal))
+      : dispatch(setLatitude(latLonVal))
+  }
+  const handleSetPanZoom = () => {
```
hoorayimhelping, line 46:
```suggestion
  const handleTogglePanZoom = () => {
```
the function doesn't set anything, it just toggles a state

src/timeMachine/reducers/index.ts
```diff
@@ -1081,6 +1084,133 @@ export const timeMachineReducer = (
...
+    case 'SET_RADIUS': {
+      const {radius} = action.payload
+
+      switch (state.view.properties.type) {
+        case 'geo':
+          return setViewProperties(state, {
+            layers: [{...state.view.properties.layers[0], radius}],
+          })
+      }
+    }
```
hoorayimhelping, line 1213: let's pull this out into a new file. it'll be easier to set up defaults and follow it that way

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,126 @@
...
+        <Form.Element label="Zoom">
+          <RangeSlider
+            min={0}
+            max={20}
+            value={props.zoom}
+            onChange={e => handleZoomSelect(parseFloat(e.target.value))}
+            testID="geoZoom"
+          />
+        </Form.Element>
+        {displayCustomOptions(props.layers[0].type)}
```
hoorayimhelping, line 121: same bit here about assumptions that layers is an array with an object with type

src/timeMachine/components/view_options/GeoOptions.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+    expect(spyLatitudeAction).toHaveBeenCalledWith(40)
+    expect(spyLongitudeAction).toHaveBeenCalledWith(39)
+    expect(spyZoomAction).toHaveBeenCalledWith(12)
+
+    const storeProps = store.getState().timeMachines.timeMachines.de.view
+      .properties as GeoViewProperties
+    expect(storeProps.allowPanAndZoom).toEqual(false)
+    expect(storeProps.center).toEqual({lat: 40, lon: 39})
+    expect(storeProps.zoom).toEqual(12)
+  })
```
hoorayimhelping, line 149: so i'm not sure if an integration test is the best course of action to test redux actions. the problem here is that we're deeply testing implementation, but not much behavior. we're basically saying "when this click happens, this function is called" which is a bit of antipattern. we want to assert something more along the lines of, "when a user clicks this button, the system's new state looks like this"

the difference is the case that tests behavior says "this is what i expect to happen" and the case that tests implementation says "this is what i expect you to do". tests of implementation have their place, but i don't think this is it. i think testing the reducer logic in a unit test will probably be more valuable. we can talk more about this in person though

hoorayimhelping (changes requested):
This is good overall, but there are a lot of little things that aren't very conventional within our codebase, and some things that should flat-out be changed.

I want you to replace all the default exports in new files in this diff with named exports:
https://basarat.gitbook.io/typescript/main-1/defaultisbad

src/timeMachine/reducers/index.ts
```diff
@@ -1081,6 +1098,87 @@ export const timeMachineReducer = (
...
+            layers: [defaultCircleMap],
+          })
+        }
+        default:
+          return state
+      }
+    }
+    case SET_ZOOM_VALUE: {
+      const {zoom} = action
+      switch (state.view.properties.type) {
```
hoorayimhelping, line 1132: code style tip: no need to get fancy here, just use a normal early return guard statement:

```js
if (state?.view?.properties?.type === 'geo') {
  return setViewProperties(state, {zoom})
}

return state
}
```

**BUT** is this even necessary? Won't at this point, the `type` always be geo?

src/timeMachine/utils/geoMapTypeDefaults.ts
```diff
@@ -0,0 +1,51 @@
...
+  radiusField: 'magnitude',
+  radiusDimension: {label: 'Magnitude'},
+  colorDimension: {label: 'Duration'},
+  colorField: 'duration',
+  colors: [
+    {type: 'min', hex: '#ff00b3'},
+    {value: 50, hex: '#343aeb'},
+    {type: 'max', hex: '#343aeb'},
+  ],
+}
```
hoorayimhelping, line 51: i love where your head is at here with creating these structures. let's keep them close to the redux store that uses them.

a hint that they should stay there is that the only two pieces of code that use these structures are the redux store and the code that tests the redux store.

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,132 @@
...
+  setLongitude,
+  setAllowPanAndZoom,
+  setZoomValue,
+  setMapType,
+} from 'src/timeMachine/actions/geoOptionsCreators'
+import {GeoViewProperties} from 'src/types'
+import {Grid, SelectGroup, Form, RangeSlider} from '@influxdata/clockface'
+import './GeoOptions.scss'
+const mapTypeOptions = ['Point', 'Circle', 'Heat', 'Track']
+enum LatMinMax {
```
hoorayimhelping, line 15:
```suggestion
enum LattitudeSliderRange {
```
thoughts?

src/timeMachine/components/view_options/CustomHeatmapOptions.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const dispatch = useDispatch()
+
+  const handleSetHeatmapRadius = (newRadius: number) => {
+    if (geoViewProperties?.layers[0]?.radius) {
+      dispatch(setRadius(newRadius))
+    }
+  }
+
+  return (
+    <Form.Element label="Radius" testID="heatmapradiusslider">
```
hoorayimhelping, line 28: `geo-heatmap-radius-slider`

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+  const handleTogglePanZoom = () => {
+    dispatch(setAllowPanAndZoom(!props.allowPanAndZoom))
+  }
+  const handleZoomSelect = (zoomValue: number) => {
+    dispatch(setZoomValue(zoomValue))
+  }
+  return (
+    <>
+      <SelectGroup className="mapTypeOptions" testID="maptypeselect">
```
hoorayimhelping, line 67: remove this `testID` if it's not being used

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,132 @@
...
+        </Form.Element>
+        <Form.Element label="Latitude">
+          <RangeSlider
+            min={LatMinMax.Min}
+            max={LatMinMax.Max}
+            value={props.center.lat}
+            onChange={event => {
+              handleSelectLatitude(parseFloat(event.target.value))
+            }}
+            testID="geo-latitude"
```
hoorayimhelping, line 103: suggestion: a little more helpful when reading code outside here:
`geo-latitude-slider`

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,132 @@
...
+        </Form.Element>
+        <Form.Element label="Longitude">
+          <RangeSlider
+            min={LonMinMax.Min}
+            max={LonMinMax.Max}
+            value={props.center.lon}
+            onChange={event => {
+              handleSelectLongitude(parseFloat(event.target.value))
+            }}
+            testID="geo-longitude"
```
hoorayimhelping, line 114: `geo-longitude-slider`

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,132 @@
...
+        </Form.Element>
+        <Form.Element label="Zoom">
+          <RangeSlider
+            min={ZoomMinMax.Min}
+            max={ZoomMinMax.Max}
+            value={props.zoom}
+            onChange={event => {
+              handleZoomSelect(parseFloat(event.target.value))
+            }}
+            testID="geo-zoom"
```
hoorayimhelping, line 125: `geo-zoom-slider`

src/timeMachine/components/view_options/CustomHeatmapOptions.tsx
```diff
@@ -0,0 +1,42 @@
+import React, {FC} from 'react'
+import {getActiveTimeMachine} from 'src/timeMachine/selectors'
+import {useSelector, useDispatch} from 'react-redux'
+import {Form, RangeSlider} from '@influxdata/clockface'
+import {setRadius} from 'src/timeMachine/actions/geoOptionsCreators'
+import {GeoViewProperties} from 'src/client'
+
+enum HeatMapRadiusMinMax {
```
hoorayimhelping, line 8:
```suggestion
enum HeatMapRadiusRange {
```

thoughts on `range` vs `minmax` ?

src/timeMachine/components/view_options/GeoOptions.tsx
```diff
@@ -0,0 +1,132 @@
...
+const displayCustomOptions = (mapType: MapType) => {
+  switch (mapType) {
+    case MapType.Heat: {
+      return <CustomHeatMapOptions />
+    }
+    case MapType.Track:
+    case MapType.Point:
+    case MapType.Circle:
+    default:
+      return null
```
hoorayimhelping, line 44:
```suggestion
    default: {
      return null
    }
```

hoorayimhelping (commented):
some more suggestions

hoorayimhelping (approved):

PR #578, opened 2021-01-29
feat: add panel header name to self registration scheme

drdelambre:
we're currently hand editing all of the "human readable" panel header names. lets just use the panel's type definition to grab the same text as we're putting on the button they click. talking about automating this thing:

Screenshot from 2021-01-29 12-00-01


src/flows/pipes/Youtube/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export default register => {
+  register({
+    type: 'youtube',
+    family: 'test',
+    priority: -1,
+    featureFlag: 'flow-panel--youtube',
+    button: 'Youtube',
+    component: View,
+    initial: {
+      uri: 'nBHkIWAJitg',
```
genehynson, line 13: what is this random string?

genehynson: oh it's the youtube id of some influx video lol ok

genehynson (approved):

src/flows/pipes/Youtube/index.ts
drdelambre: the id that follows a youtube video's url

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #583, opened 2021-02-01
fix: #582 better token failure error messages

### LINKED ISSUE #582: when token generation fails, the error is vague
tldr: during telegraf token generation:

1) add a check for a bucket not being present
2) show more detailed error messages to the user in the popup if we have the information; else show the generic error message

<hr>

We have information on why a token generation fails, so we should show that to the user, not bury it only in the console.log.

when a token fails;  we currently only get the message:

'Failed to create a new Telegraf Token'

we already check for no buckets (' 'A token cannot be generated without a corresponding bucket; no buckets are assigned'').

we need to add a check for the bucket not being present ('there is no bucket present with name ${bucketName}')

the error should be more specific, as long as we know the reason.
so, 
if the failure is not  because of the two reasons above, the generic message will still show.

if the failure is because of either of the two reasons above, the message will now be:

`<generic message>: <specific message>`. in the error popup.

for example, in the second error condition above, the message will now be:

"Failed to create a new Telegraf Token: there is no bucket present with name $INFLUX_BUCKET"

Screen Shot 2021-02-01 at 11 13 13 AM

This is in response to bug #577

jrenee42:
Closes #582 

here is a screenshot:

Screen Shot 2021-02-01 at 11 13 13 AM

### DIFF
--- a/src/dataLoaders/actions/dataLoaders.ts
+++ b/src/dataLoaders/actions/dataLoaders.ts
@@ -402,13 +402,18 @@ export const generateTelegrafToken = (configID: string) => async (
     const orgID = getOrg(state).id
     const telegraf = getByID<Telegraf>(state, ResourceType.Telegrafs, configID)
     const bucketName = get(telegraf, 'metadata.buckets[0]', '')
+
     if (bucketName === '') {
       throw new Error(
-        'A token cannot be generated without a corresponding bucket'
+        'A token cannot be generated without a corresponding bucket; no buckets are assigned'
       )
     }
+
     const bucket = getBucketByName(state, bucketName)
 
+    if (!bucket || !bucket.id) {
+      throw new Error(`there is no bucket present with name ${bucketName}`)
+    }
     const permissions = [
       {
         action: Permission.ActionEnum.Write,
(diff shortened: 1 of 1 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -442,7 +447,16 @@ export const generateTelegrafToken = (configID: string) => async (
     dispatch(setToken(createdToken.token))
   } catch (error) {
     console.error(error)
-    dispatch(notify(TokenCreationError))
+
+    if (error.message) {
+      const customNotification = {
```
hoorayimhelping, line 452: only nitpick is that `customNotification` might be named a little bit more richly and descriptively.

i'm not sure if this structure belongs in `shared/copy/notifications.ts` or should stay here 🤔. It's probably fine here for now since it's not being used by anything else

hoorayimhelping (approved):
looks good, nice work. one nitpick about variable names

PR #586, opened 2021-02-01
fix:  legend usability

### LINKED ISSUE #560: Visualization options: Orientation threshold and opacity design and usability
https://github.com/influxdata/ui/pull/95 added a "rotation threshold"  and opacity slider for the legend in query cells. While documenting this, I noticed some issues that I've verified with @sjwang90. Happy to hop on a call to demo if anything below isn't clear. 

## Orientation 

At X columns, the legend rotates from horizontal to vertical:

image

It would be much simpler to have a vertical/horizontal orientation toggle option. The number of columns in the query isn't going to change unless they change the query, so setting a threshold is unnecessarily complicated because all they're be doing is switching the orientation for that specific cell. I don't see any use case where having a threshold set at the cell/query level would be useful (let me know if I'm missing something). 

## Opacity

The opacity slider determines how opaque the legend appears. 

image

A couple suggestions for design improvements here:

- It should show opacity as a percentage, not a proportion of 1 (aka 55 instead of .55 in the example above).
- There's no need to have numbers on either side of the slider. Similar slider UIs on instagram, etc. just show total opacity as you adjust the slider. See a good example from Capture One [here](https://support.captureone.com/hc/en-us/articles/360002615497-Opacity-slider-effect-in-the-Layers-tool).

jrenee42:
Closes #560 

-replacing threshold amount with vertical/horizontal radio buttons
-styled the opacity to show the percentage (and removed the labels on the left and right of the slider)
-cleaned up the legend configuration style.

how it looks now:  (note that I included the segment above too, to show consistency with margins):
Screen Shot 2021-02-01 at 3 30 41 PM

- [x] Rebased/mergeable
- [x] Tests pass


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/timeMachine/components/view_options/LegendOrientation.tsx
```diff
@@ -45,6 +49,32 @@ interface OwnProps {
...
+ *          the legend is displayed vertically (each line is a column); if the number of items is below the threshold, then the legend is
+ *          displayed horizontally (each line is a row).  therefore, since the default is horizontal, the numbers being set are very high
+ *          for horizontal (high enough that the database would thrash out/throttle before it got that high), or low (0 for vertical, since
+ *          to display something it has to be higher than zero).  The default number itself has changed, so some users may see their graphs flip
+ *          (before; without the feature flag enabled, the default was hard-coded to 10; so graphs with more than 10 lines of  data in the tooltip
+ *          were vertical)
+ *
+ *    2.  Legend Opacity.  The actual number is between the minimum .2 and the maximum 1.0; but we are displaying it as a matter of percentage
+ *    3.  Row Colorization.  This is the simplest.  just a true/false flag, it is true by default.
+ */
```
hoorayimhelping, line 77: i love the legend orientation explanation, bu i think this a little too dense for an in code comment. i think it's okay to ✂️ this

src/timeMachine/components/view_options/LegendOrientation.tsx
```diff
@@ -95,50 +130,94 @@ const LegendOrientation: FC<Props> = props => {
...
+        checked={legendOrientation === HORIZONTAL}
+        onChange={() => handleSetOrientation(HORIZONTAL)}
+        type={InputToggleType.Radio}
+        size={ComponentSize.ExtraSmall}
+        color={ComponentColor['Primary']}
+        appearance={Appearance['Outline']}
+        style={{marginBottom: 6}}
+      >
+        <InputLabel
+          active={legendOrientation === 'horizontal'}
```
hoorayimhelping, line 163: missed a string literal :)

hoorayimhelping (changes requested):
Good work! A couple of little nitpicks and some suggestions

src/timeMachine/components/view_options/LegendOrientation.tsx
jrenee42: i was asked specifically to put an explanation like this in the code since we are really just using the threshold, not a flag.  if it's not here, where should it go?

hoorayimhelping: if you were asked to put it here, then great work! ignore my previous comment :)

src/timeMachine/components/view_options/LegendOrientation.tsx
```diff
@@ -95,50 +130,94 @@ const LegendOrientation: FC<Props> = props => {
...
   }
 
-  const toggleStyle = {marginTop: 4}
+  const toggleStyle = {marginTop: 12}
 
   const toggleLabelStyle = {color: '#999dab'}
 
+  // without the toFixed(0) sometimes you
+  // can get numbers like 45.000009% which we want to avoid
+  const percentLegendOpacity = (legendOpacity * 100).toFixed(0)
```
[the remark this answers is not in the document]
hoorayimhelping: narrator: it wasn't.

(we both talked and decided it's not good test material)

hoorayimhelping (approved):
great work!

PR #592, opened 2021-02-02
chore: normalizing more variables

### LINKED ISSUE #521: cypress: normalize environment variables
a weird artifact of shifting around our cluster definitions and making the tests more adaptable to new environments is that we've found multiple sources of truth within the system that are becoming a time bomb. For example, we have a concept of a user's bucket, which has a name and is defined in the database and also referenced by the test. We currently have that defined in three ways:
1. Hardcoded as the string `defbuck`
2. Passed as an environment variable, `Cypress.env('bucket')`, which is used to share the definition between the tests and the provisioning node that hydrates the initial state of the database
3. Fetched on test setup and wrapped in cypress, `cy.get('@bucket')`

The first one is terrible because it requires the developer to know magic keywords whenever writing a test or provision a cluster, which is required to be kept in their head at all times.
 The second one follows a pattern that is used commonly within our infrastructure, but which also couples the two services in a way that further extension of either application requires knowledge of magic strings.
 The last one moves all definition of what the user's available buckets, their names, and how they're created, to the service that generates that information on the database. This allows the tests to have zero information about how we create / configure / initialize the database, and instead just run tests.
 
 Obviously, the last one is the best. We should use the last one. This issue is to normalize the tests to use the last one for the following magic strings
 - Org name
 - Org ID
 -  user name
 - bucket
 - any other magic you can find
 
 For an example of the pattern, see how buckets and org information are being wrapped and used in the PR: https://github.com/influxdata/ui/pull/507

joelfried:
Closes #521 

More cypress variable normalization - this removes all cypress environment variables injected in cypress.json.  Those values are now injected via aliases that are run at the start of each e2e test.  I gave a once over to all other e2e tests and couldn't find any other magic strings to abstract out in the same way.

- [x] Rebased/mergeable
- [x] Tests pass


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

genehynson (approved):

PR #594, opened 2021-02-03
feat: refresh persist data explorer

ShmuelLotman:
Closes #5664

This PR introduces the ability to persist query state across refreshes in data explorer, including when moving from data explorer to another page, refreshing and then coming back. The test in this file may need to be adjusted as I see how Cypress handles the changes after rebuilding, so if you see that failing, im on it and it should be fixed shortly. 

This is behind the influx featureflag 'persistRefresh', so make sure to use `influx.set('persistRefresh', true)` to see this work. 
- [x] Rebased/mergeable
- [ ] Tests pass


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #595, opened 2021-02-03
feat: persist data across refreshes in data explorer

ShmuelLotman:
Closes #5664

This PR introduces the ability to persist the query state across refreshes in data explorer, even when leaving the page and refreshing in between. 

Im currently investigating whether the cypress test will be able to pick up the IIFE outside of the React component, so that test may need to be adjusted, which is why im leaving the "tests passing" section blank for now, until i can be sure. Wanted to get eyes on the change itself to get peoples thoughts. 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -670,6 +670,9 @@ describe('DataExplorer', () => {
       it('can view time-series data', () => {
         cy.get<string>('@defaultBucketListSelector').then(
           (defaultBucketListSelector: string) => {
+            cy.window().then(win => {
+              win.influx.set('persistRefresh', true)
+            })
             cy.log('can switch between editor modes')
             cy.getByTestID('selector-list _monitoring').should('be.visible')
             cy.getByTestID('selector-list _monitoring').click()
@@ -769,6 +772,16 @@ describe('DataExplorer', () => {
             cy.getByTestID('raw-data-table').should('exist')
             cy.getByTestID('raw-data--toggle').click()
             cy.getByTestID('giraffe-axes').should('exist')
+
+            cy.reload()
+
+            cy.wait(2000)
+
+            cy.getByTestID('selector-list defbuck').should('be.visible')
+
+            cy.getByTestID('selector-list m').should('be.visible')
+
+            cy.getByTestID('selector-list v').should('be.visible')
           }
         )
       })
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

cypress/e2e/shared/explorer.test.ts
```diff
@@ -769,6 +772,16 @@ describe('DataExplorer', () => {
...
+
+            cy.reload()
+
+            cy.wait(2000)
+
+            cy.getByTestID('selector-list defbuck').should('be.visible')
+
+            cy.getByTestID('selector-list m').should('be.visible')
+
+            cy.getByTestID('selector-list v').should('be.visible')
```
drdelambre, line 784: this condition kind of captures it, but add the "is visible _AND_ selected" condition to each of these

src/dataExplorer/components/DataExplorer.tsx
```diff
@@ -16,16 +16,55 @@ import {queryBuilderFetcher} from 'src/timeMachine/apis/QueryBuilderFetcher'
 import {readQueryParams} from 'src/shared/utils/queryParams'
 import ErrorBoundary from 'src/shared/components/ErrorBoundary'
 
+import {getActiveTimeMachine} from 'src/timeMachine/selectors'
+
+import {getStore} from 'src/store/configureStore'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+;(() => {
+  if (isFlagEnabled('persistRefresh')) {
```
drdelambre, line 24: no action required, but if you find yourself in a place where everything is in a condition, it sometimes makes cleaner code to invert that condition and return early. ie:
```
if (!isFlagEnabled('persistRefresh')) {
  return
}

// do the rest of your magic here
```

drdelambre (approved):
add in those test conditions to make it super good!

i think the next ask around this is gonna be to extend it to the other time machines, should be a copy pasta job though, and lets make sure nothing weird goes on here first

hoorayimhelping:
I couldn't find 5664 in idpe or influxdb. what issue is this referencing?

russorat (approved):
This is a step in the right direction IMO. I'd love to get feedback from users and optimize from there.

genehynson:
@ShmuelLotman should we revive this?

PR #612, opened 2021-02-04
fix: annotations time consistency

### LINKED ISSUE #570: Annotations: start and end time consistency within the ui's model
### Background
This is a pithy summary:
> the ui wants numbers, the backend (at least for when we are writing the annotations), wants strings

Right now, the ui's model uses strings. We need to convert the backend's needs into the ui's domain needs, which is numbers.

From Greg:
>if it makes more sense to have those values be unix epoch nanoseconds, that's an option. this api is primarily for your (the ui's) consumption

### Acceptance Criteria
- Annotation dates are represented consistently with the rest of the ui's domain model
  - under this scheme, translations or transformations between the UI and the API and the UI and Giraffe should happen at boundary layers
- Naming is consistent (`end` vs `stopValue` and `start` vs `startValue`)

ChitlangeSahas:
We need to make the annotation `start` and `end` times consistent in the UI, this PR attempts to do that. start and end will now be refactored to `startTime` and `endTime` to maintain consistency. 

Closes #570 

- [x] Rebased/mergeable
- [x] Tests pass


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/types/annotation.ts
```diff
@@ -14,14 +14,14 @@ export interface AnnotationStream {
...
+  end?: number
   stickers?: any
 }
 
 export interface DeleteAnnotation {
   stream: string
-  start: string
-  end: string
+  start: number
+  end: number
```
hoorayimhelping, line 25: i think you shot yourself in the foot here 😄 if you change these back to `string` (which they should be, these are what we send across the wire) and remove the `eslint-ignore` and `ts-ignore`, you should get a clean build.

hoorayimhelping (changes requested):

src/types/annotation.ts
hoorayimhelping, line 25: you'll want to change these back, they don't have any bearing on this problem, and they were correct before

src/annotations/api/index.ts
```diff
@@ -22,7 +22,19 @@ import {formatAnnotationQueryString} from 'src/annotations/utils/formatQueryStri
 export const writeAnnotation = async (
   annotations: Annotation[]
 ): Promise<Annotation[]> => {
-  const res = await axios.post(url, annotations)
+  // we need to convert the annotation object's start and end time fields to be
+  // string types so they can be parsed in the backend.
+  const annotationsRequestConverted = annotations.map(annotation => {
```
hoorayimhelping, line 27: so we'll want to remove these ignores - we can get TypeScript to shut up in other ways that are a little simpler.

the problem here, is that `annotations` is an `Annotation[]` (an `Annotation` array) and `Annotations` have dates as numbers, but we're trying to take an array of `Annotations` and change their number dates to strings. The problem occurs because you're trying to re-assign `annotation.start`.

Overwriting the arguments that come in from a function is generally an antipattern; you should copy them instead:

```js
const annotationsRequestConverted = annotations.map(annotation => {
  return {
    ...annotation,
    start: new Date(annotation.start).toISOString(),
    end: new Date(annotation.end).toISOString(),
  }
})
```

```js
const original = { name: 'Sahas' }
const poser = { ...original }

poser.name = 'Bucky'
console.log(original, poser)
// { name: "Sahas" }  { name: "Bucky" }
```

hoorayimhelping (changes requested):
sorry if you got notified before. left an incorrect comment, should be fixed now.

src/types/annotation.ts
ChitlangeSahas: For sure!

src/annotations/api/index.ts
ChitlangeSahas: Thank you for this solid detailed explanation Bucky! I now understand what the 3 dots do! I read up on it as well and it's very interesting how it can be used to write cleaner code.

hoorayimhelping (approved):
looks good to me, solid work.

if you would, let Jill take a look at it before you merge, please

src/annotations/api/index.ts
```diff
@@ -22,7 +22,17 @@ import {formatAnnotationQueryString} from 'src/annotations/utils/formatQueryStri
...
+    return {
+      ...annotation,
+      start: new Date(annotation.start).toISOString(),
```
jrenee42, line 30: why is this being changed to strings here?

and the naming is still not consistent (start/end vs different names on the annotations... (startTime and stopTime?  don't have time to look them up here)

also:  how is readAnnotation affected?

ChitlangeSahas: That is because according to the issue we'd like to handle the conversions at the boundary layer. Any suggestions for a better place we could do it?

I can rename the variables. 

Do you mean `getAnnotation` by chance? It is not touched in this PR.

jrenee42: why are we doing the conversions at all? this is a new service, it shouldn't need conversions @hoorayimhelping

glinton: re the naming. annotations service is expecting `start` and `end`:
```go
	EndTime   *time.Time        `json:"end,omitempty"`      // EndTime is the time of the event being annotated. Defaults to now if not set.
	StartTime *time.Time        `json:"start,omitempty"`    // StartTime is the start time of the event being annotated. Defaults to EndTime if not set.
}
```

The important bits for communicating are in the json tag (`json:"start"`). That defines what is expected in the json serialized body. I'm not sure if this renaming of vars from `start`->`startValue` is going to result in a payload with `"startValue": "some time", "endValue": "another time"}`, but that would be incorrect.

re the conversions. the expected time format is [`RFC3339Nano`](https://golang.org/pkg/time/) i can't say whether converting them a string is right or wrong, as long as it serializes as that format.

jrenee42 (approved):
approving.  I see that reconciling the front and back ends is out of scope for now; at least we are now more consistent within the app at least.

PR #625, opened 2021-02-05
fix: should get id from url before checking if flow exists

### LINKED ISSUE #621: first time entering a shared notebook causes a blank screen with no error
Upon the first time enabling the `notebooks-api` flag in a multi-user org the app produces a blank screen when you try and enter a shared notebook. 

The work around is to first "Create notebook" and then try and enter the original shared notebook and the blank screen issue goes away for good.

image

genehynson:
Closes #621 

There are three issues:
1. Users were unable to go directly to a shared Notebook url that wasn't in their browser's cache.
2. The FlowProvider expects an id to exist before rendering FlowPage but FlowPage is responsible for getting the id from the url.
3. Going to a notebook directly that doesn't exist throws an error with no message

I fixed (1) by calling `getAll` whenever the `currentID` is changed to pull down the notebooks. I fixed (2) by moving the check for the existence of the down the component tree. If the `flow` prop is undefined, assume it doesn't exist. I fixed (3) by showing an error message that says the notebook couldn't be found.

image

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/flows/context/flow.list.tsx
+++ b/src/flows/context/flow.list.tsx
@@ -280,17 +280,6 @@ export const FlowListProvider: FC = ({children}) => {
     pooledUpdateAPI(apiFlow)
   }
 
-  const change = useCallback(
-    (id: string) => {
-      if (!flows || !flows.hasOwnProperty(id)) {
-        throw new Error(`${PROJECT_NAME} does note exist`)
-      }
-
-      setCurrentID(id)
-    },
-    [setCurrentID, flows]
-  )
-
   const remove = async (id: string) => {
     const _flows = {
       ...flows,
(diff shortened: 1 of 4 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/flows/context/flow.list.tsx
```diff
@@ -318,6 +309,16 @@ export const FlowListProvider: FC = ({children}) => {
...
+  const change = useCallback(
+    (id: string) => {
+      if (!Object.keys(flows).length) {
+        getAll()
+      }
+      setCurrentID(id)
+    },
+    [setCurrentID, flows]
+  )
+
```
genehynson, line 321: Moved this below `getAll` so I can reference it

drdelambre (approved):

PR #644, opened 2021-02-08
fix: saving a query shouldnt leave query state around

### LINKED ISSUE #617: Query Builder: Creating a new query leaves the old query builder state around
### Background
Creating a new query through the query builder, then going elsewhere in the app, then coming back to query builder shows the previous query builder state.

### Expected Behavior
I'd expect after saving a query using the builder that the state would reset.

### Reproduction Steps
1. Create a query through the builder
1. Save it and add it to a dashboard
1. Go somewhere else, like the dashboard you just created
1. Go back to query builder - the state is still around

ShmuelLotman:
Closes #617 

Under the changes proposed in this PR, saving a query as a dashboard cell will redirect to the dashboard with the cell after saving it. If multiple dashboards are chosen, it will navigate to the last dashboard saved to. When returning to the query builder, it will now be reset. 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -1199,17 +1199,12 @@ describe('DataExplorer', () => {
 
         cy.getByTestID('save-as-dashboard-cell--submit').click()
 
-        // wait some time for save
-        cy.wait(2000)
-        // ensure dashboard created with cell
-        // cy.get('@org').then(({id: orgID}: Organization) => {
-        //   cy.fixture('routes').then(({orgs}) => {
-        //     cy.visit(`${orgs}/${orgID}/dashboards/`)
+        cy.location('pathname').should(
+          'match',
+          /^(?=.*dashboards)(?:(?!cell).)+$/
+        )
 
-        //   })
-        // })
-        cy.getByTestID('dashboard-card--name')
-          .contains(dashboardCreateName)
+        cy.getByTestID(`cell--draggable ${cellName}`)
           .should('exist')
           .click()
         cy.getByTestID(`cell ${cellName}`).should('exist')
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -1186,7 +1186,7 @@ describe('DataExplorer', () => {
         })
       })
 
-      it('can create new dashboard as saving target', () => {
+      it.only('can create new dashboard as saving target', () => {
         // select and input new dashboard name and cell name
         cy.getByTestID('save-as-dashboard-cell--dropdown').click()
         cy.getByTestID('save-as-dashboard-cell--create-new-dash').click()
@@ -1200,8 +1200,9 @@ describe('DataExplorer', () => {
         cy.getByTestID('save-as-dashboard-cell--submit').click()
 
         // wait some time for save
-        cy.wait(200)
+        cy.wait(2000)
         // ensure dashboard created with cell
+<<<<<<< HEAD
         cy.get('@org').then(({id: orgID}: Organization) => {
           cy.fixture('routes').then(({orgs}) => {
             cy.visit(`${orgs}/${orgID}/dashboards/`)
@@ -1213,6 +1214,19 @@ describe('DataExplorer', () => {
             cy.getByTestID(`cell ${cellName}`).should('exist')
           })
         })
+=======
+        // cy.get('@org').then(({id: orgID}: Organization) => {
+        //   cy.fixture('routes').then(({orgs}) => {
+        //     cy.visit(`${orgs}/${orgID}/dashboards/`)
+
+        //   })
+        // })
+        cy.getByTestID('dashboard-card--name')
+          .contains(dashboardCreateName)
+          .should('exist')
+          .click()
+        cy.getByTestID(`cell ${cellName}`).should('exist')
+>>>>>>> fix: working with create and update, testing in progress
       })
     })
 
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -1221,8 +1221,7 @@ describe('DataExplorer', () => {
 
         //   })
         // })
-        cy.getByTestID('dashboard-card--name')
-          .contains(dashboardCreateName)
+        cy.getByTestID(`cell--draggable ${cellName}`)
           .should('exist')
           .click()
         cy.getByTestID(`cell ${cellName}`).should('exist')
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -1186,7 +1186,7 @@ describe('DataExplorer', () => {
         })
       })
 
-      it.only('can create new dashboard as saving target', () => {
+      it('can create new dashboard as saving target', () => {
         // select and input new dashboard name and cell name
         cy.getByTestID('save-as-dashboard-cell--dropdown').click()
         cy.getByTestID('save-as-dashboard-cell--create-new-dash').click()
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -1160,13 +1160,12 @@ describe('DataExplorer', () => {
 
       it('can save as cell into multiple dashboards', () => {
         // input dashboards and cell name
-        cy.getByTestID('save-as-dashboard-cell--dropdown').click()
-        cy.getByTestID('save-as-dashboard-cell--dropdown-menu').within(() => {
-          dashboardNames.forEach(d => {
-            cy.contains(d).click()
+        dashboardNames.forEach(name => {
+          cy.getByTestID('save-as-dashboard-cell--dropdown').click()
+          cy.getByTestID('save-as-dashboard-cell--dropdown-menu').within(() => {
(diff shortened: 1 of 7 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

cypress/e2e/shared/explorer.test.ts
```diff
@@ -1198,18 +1198,12 @@ describe('DataExplorer', () => {
...
-          cy.fixture('routes').then(({orgs}) => {
-            cy.visit(`${orgs}/${orgID}/dashboards/`)
-            cy.getByTestID('dashboard-card--name')
-              .contains(dashboardCreateName)
-              .should('exist')
-              .click()
-            cy.getByTestID(`cell ${cellName}`).should('exist')
-          })
-        })
+        cy.wait(1000)
```
drdelambre, line 1201: sniff for the url change (cy.location or something) instead of having an arbitrary wait time

drdelambre (approved):

cypress/e2e/shared/explorer.test.ts
```diff
@@ -1197,19 +1197,15 @@ describe('DataExplorer', () => {
...
-            cy.getByTestID('dashboard-card--name')
-              .contains(dashboardCreateName)
-              .should('exist')
-              .click()
-            cy.getByTestID(`cell ${cellName}`).should('exist')
-          })
-        })
+        cy.location('pathname').should(
+          'match',
+          /^(?=.*dashboards)(?:(?!cell).)+$/
```
hoorayimhelping, line 1202: is there any way this can be written out in a more approachable manner? this is dense even for a regex. it took me like 2 minutes to comprehend it, and even then, i'm not super confident that i understand it well. having to use regex101.com to check what this code tests isn't great. are you confident that this regex covers all your test cases?

but even after sorta understanding it, i'm not sure what is significant about this pattern. why do urls with dashboard but not cell in them matter?

src/cells/actions/thunks.ts
```diff
@@ -69,7 +69,8 @@ export const deleteCellAndView = (
 export const createCellWithView = (
   dashboardID: string,
   view: NewView,
-  clonedCell?: Cell
+  clonedCell?: Cell,
+  redirect?: (destinationId: string) => void | undefined
```
hoorayimhelping, line 73: what's the ```void | undefined``` bit about? shouldn't `void` be fine? 

https://stackoverflow.com/a/58885486/92446

src/dataExplorer/components/SaveAsCellForm.tsx
```diff
@@ -170,21 +176,42 @@ class SaveAsCellForm extends PureComponent<Props, State> {
 
     const viewWithProps: View = {...view, name: cellName}
 
+    let succeeded = false
+    const redirectHandler = (dashboardId: string): void =>
+      history.push(`/orgs/${orgID}/dashboards/${dashboardId}`)
+
     try {
-      targetDashboardIDs.forEach(dashID => {
+      targetDashboardIDs.forEach((dashID, idx) => {
```
hoorayimhelping, line 184:
```suggestion
      targetDashboardIDs.forEach((dashID, i) => {
```

one of the few times it's better to just use a single letter since it's conventional across the industry. everybody knows what `i` represents. and it usually clues you in that you're in a loop. again, special case because it's conventional across many languages and paradigms.

src/dataExplorer/components/SaveAsCellForm.tsx
```diff
@@ -170,21 +176,42 @@ class SaveAsCellForm extends PureComponent<Props, State> {
...
+        )
       })
+      succeeded = true
     } catch (error) {
       console.error(error)
     } finally {
       this.resetForm()
-      dismiss()
+      if (succeeded) {
+        this.props.setTimeMachine('de', initialStateHelper())
```
hoorayimhelping, line 211: is this a navigation or view changing like event? how's it different from `dismiss()`? what's this do from a user's perspective?

src/dataExplorer/components/SaveAsCellForm.tsx
```diff
@@ -234,9 +261,10 @@ const mdtp = {
   onGetDashboards: getDashboards,
   onCreateCellWithView: createCellWithView,
   onCreateDashboardWithView: createDashboardWithView,
+  setTimeMachine: setActiveTimeMachine,
```
hoorayimhelping, line 264: any reason to alias this? there are no naming conflicts, and i can't think of any good reasons to remove `active` and the implications of having that in the function name.

hoorayimhelping (changes requested):
mainly a question about `setTimeMachine`,

src/dataExplorer/components/SaveAsCellForm.tsx
ShmuelLotman: This is the call that resets the query builder state. Dismiss removes the modal. We aren't in the querybuilder at this point so there is no need to call dismiss

src/dataExplorer/components/SaveAsCellForm.tsx
ShmuelLotman: yes, its annoying to me to have the imported method and the method in the component have the same name.

cypress/e2e/shared/explorer.test.ts
ShmuelLotman: This was @drdelambre 's request, i am happy to change this but this is just a check that we get to the dashboards page and do not open an individual cell.

src/dataExplorer/components/SaveAsCellForm.tsx
hoorayimhelping: @ShmuelLotman so can you remove the `succeeded` flag and move this into the `try` block and move `dismiss` into the `catch` block?

you really shouldn't have a conditional statement in a `finally` block. it kind of defeats the purpose of a `finally` statement. the `try/catch` statement kind of encodes the value of the `success` flag in its structure. if everything succeeds and you try to set the time machine and that succeeds, hooray, everything is good to go. if it fails, it'll drop into the catch block and you handle the failure case there.

src/dataExplorer/components/SaveAsCellForm.tsx
hoorayimhelping: @ShmuelLotman what about this having the same name is annoying to you?

it's annoying to me to lose track of a reference to a function when i'm globally searching a codebase.

if i'm trying to take stock of all the places where the method `setActiveTimeMachine` is called, i now have to account for this renaming that doesn't exist for any practical reason. there's no naming collision in this file and so there's no obvious practical reason to do it. it doesn't make the code more [self documenting](https://github.com/influxdata/ui/blob/master/src/billing/thunks/index.ts#L23-L25) like some aliases in our codebase might be argued to do.

ShmuelLotman: you dont lose track to the reference just because the name changes there because you still have it imported and its still being used in the file, so the file will still have reference to it. Feel somewhat like shadowing (even though its not) and it doesn't feel like clean code to have a mapped function have the same name. I dont think you lose anything by renaming it, and i think you gain the clarity of knowing you are calling the mapped function instead of looking like you are referencing the imported one without mapping.

ShmuelLotman: That being said, thats just my personal opinion and i changed it in the latest push to reflect your request. I guess if you have this.props its easier to see, but this should ideally eventually make it into a functional component and there, particularly with destructuring and/or using useDispatch, this will look different anyways.

src/dataExplorer/components/SaveAsCellForm.tsx
```diff
@@ -48,9 +53,9 @@ interface State {
 interface OwnProps {
   dismiss: () => void
 }
-
+type RouterProps = RouteComponentProps<{orgID: string}>
```
drdelambre, line 56: use the `getOrg` selector to get the org id. views shouldn't really be coupled with the route

src/dataExplorer/components/SaveAsCellForm.tsx
```diff
@@ -170,21 +176,37 @@ class SaveAsCellForm extends PureComponent<Props, State> {
...
     const viewWithProps: View = {...view, name: cellName}
 
+    const redirectHandler = (dashboardId: string): void =>
+      history.push(`/orgs/${orgID}/dashboards/${dashboardId}`)
+
     try {
-      targetDashboardIDs.forEach(dashID => {
+      targetDashboardIDs.forEach((dashID, i) => {
+        const toRedirect =
+          i === targetDashboardIDs.length - 1 ? redirectHandler : undefined
```
hoorayimhelping, line 185: you typically don't want to set values to `undefined`, you want to use `null`. `undefined` means "hasn't been assigned a value." `null` means no value. you don't assign a variable an `undefined` value, you assign it `null`. 

```ts
let foo;
foo; // undefined
```

cypress/e2e/shared/explorer.test.ts
hoorayimhelping: can you explain that in a comment so someone reading this doesn't have to compute a regex in their head?

hoorayimhelping (approved):
lgtm, but now ya gotta address alex's comment about `useSelector`

PR #652, opened 2021-02-09
feat: type ahead variable drop downs

jrenee42:
Closes #20195

behind the 'typeAheadVariableDropdown' feature flag

adds 'type-ahead' filtering for variable dropdowns in dashboards.

it used to look like this:
Screen Shot 2021-02-09 at 1 52 28 PM

it now looks like:
Screen Shot 2021-02-09 at 1 53 03 PM

behavior:
 * when you start typing, if the menu isn't already open it will open
 * when it first comes up, everything is shown (even if there is a pre-selected value); once you start typing tho it filters it down.  If you want to see everything in the list, then need to delete everything in the input field
 * when you leave the component, whatever value was selected will show up in the input; it won't let random strings show up
 * when you start typing, you can use the up and down arrows to make a selection.  if you make a selection and then click away, then the menu closes and you get that selection
 * if using the up/down arrows (or not), clicking 'return' sets the selection and closes the menu
    *if they were just typing, after pressing the return key the input field reverts to the selection that was last made (no random values are allowed)
 * it should be intuitive

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/variables/components/TypeAheadVariableDropdown.tsx
+++ b/src/variables/components/TypeAheadVariableDropdown.tsx
@@ -28,10 +28,6 @@ interface OwnProps {
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = OwnProps & ReduxProps
 
-const placeHolderText = 'select a value'
-//todo:  play with focus....if the entire item loses focus, then
-//the input should show the actual selected item, not what the user typed in
-//think on this.....
 class TypeAheadVariableDropdown extends PureComponent<Props> {
   constructor(props) {
     super(props)
@@ -44,18 +40,18 @@ class TypeAheadVariableDropdown extends PureComponent<Props> {
   }
 
   // set the 'shownValues' after loading, and
-  // resets the menuOpen variable
+  // resets the setMenuStatus variable
   componentDidUpdate(prevProps, prevState) {
     const prevVals = prevProps.values
     const {values, selectedValue} = this.props
-    const {actualVal, loaded, selectHappened, menuOpen} = this.state
+    const {actualVal, loaded, selectHappened, setMenuStatus} = this.state
     const {
       actualVal: prevActualVal,
       selectHappened: prevSelectHappened,
-      menuOpen: prevMenuOpened,
+      setMenuStatus: prevSetMenuStatus,
     } = prevState
 
-    //this is for updatintg the values:
+    //this is for updating the values:
     //(only want this to run *once* when the values get loaded)
     if (!loaded && prevVals.length !== values.length) {
       this.setState({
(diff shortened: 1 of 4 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+  onClickAwayHere = () => {
+    this.setState(this.getRealValue())
+  }
+
+  getRealValue = () => {
+    const {actualVal, typedValue} = this.state
+    const {selectedValue} = this.props
+
+    if (actualVal || selectedValue) {
+      const realValue = actualVal ? actualVal : selectedValue
```
genehynson, line 163: you could shorten this to `const realValue = actualVal ?? selectedValue` if you wanted

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+    const dropdownStatus =
+      values.length === 0 ? ComponentStatus.Disabled : ComponentStatus.Default
+
+    const placeHolderText = this.getPlaceHolderText('Select a Value')
+
+    const widthStyle = this.getWidth(placeHolderText)
+
+    return (
+      <Dropdown
+        style={{width: '140px'}}
```
genehynson, line 184: could we move this into `variable-dropdown--dropdown` class?

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+        menu={onCollapse => (
+          <Dropdown.Menu
+            style={widthStyle}
+            onCollapse={onCollapse}
+            theme={DropdownMenuTheme.Amethyst}
+          >
+            {shownValues.map(val => {
+              return (
+                <Dropdown.Item
+                  key={val}
```
genehynson, line 214: do we know that all the values will be unique? If not, we could change the key to be the second `index` parameter of map instead

genehynson (approved):
lgtm! Left a couple small suggestions

src/variables/components/TypeAheadVariableDropdown.tsx
jrenee42: i'm worried about testing implications; and have already spend all week on this pr.....
this would be something for refactor-friday

src/variables/components/TypeAheadVariableDropdown.tsx
jrenee42: well, they are all strings and this is how it is now.  it's a good question; but don't want to change too much with this component change.

src/variables/components/TypeAheadVariableDropdown.tsx
jrenee42: done.

jrenee42:
circle-ci tests all passed, despite what the issue says above.

src/variables/components/TypeAheadVariableDropdown.tsx
genehynson: I mean I highly doubt changing a css class would cause the tests to fail but ok...

samhld:
Yaaaayyy -- awesome, @jrenee42 !

Issue #620, opened 2021-02-05
Annotations: Merge existing rendering code into visualization pipeline


hoorayimhelping:
### Background

This [PR](https://github.com/influxdata/ui/pull/475) to change the rendering pipeline was unexpected and touches almost all the places the annotations rendering code was placed. As a team, we've decided that we're going to move forward with those rendering pipeline changes; because of that, this code will need to be re-worked within that pipeline.

It's probably best that we do it ourselves, rather than Alex doing it in that PR, so that we can understand the new pipeline and how our code fits within it.

This issue represents that work.

See #491 - the ticket representing the initial rendering work.

### Acceptance Criteria
- Annotations are rendered within the `Graph` graph type in the new visualization pipeline

PR #657, opened 2021-02-10
feat: add annotations fetching->redux

hoorayimhelping:
Connects #620 

Fetches annotations and stores them in redux. Doesn't render them yet.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/shared/components/RefreshingView.tsx
+++ b/src/shared/components/RefreshingView.tsx
@@ -14,12 +14,13 @@ import {getActiveTimeRange} from 'src/timeMachine/selectors/index'
 
 // Types
 import {
-  TimeRange,
-  TimeZone,
   AppState,
   DashboardQuery,
   QueryViewProperties,
+  AnnotationsList,
   Theme,
+  TimeRange,
+  TimeZone,
 } from 'src/types'
 
 interface OwnProps {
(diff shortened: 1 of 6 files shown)


src/shared/components/RefreshingView.tsx
```diff
@@ -62,7 +64,7 @@ class RefreshingView extends PureComponent<Props, State> {
   }
 
   public render() {
-    const {id, ranges, properties, manualRefresh, timeZone, theme} = this.props
+    const {id, manualRefresh, properties, ranges, theme, timeZone} = this.props
```
hoorayimhelping, line 67: alphabetization :shrug:

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/types/annotation.ts
```diff
@@ -25,3 +25,5 @@ export interface DeleteAnnotation {
   end: string
   stickers?: any
 }
+
+export type StateAnnotations = {[stream: string]: Annotation[]}
```
hoorayimhelping, line 29: I'm really not in love with this name. Up for suggestions

jrenee42: AnnotationList

that is what it is.  
plus the AnnotationState / StateAnnotations in the reducers is confusing.

jrenee42 (commented):
added suggestion:  AnnotationList instead of StateAnnotation for type

src/types/annotation.ts
hoorayimhelping: it's not quite a `list` though - it's a hashmap keyed on stream. It's kind of like a list, but kind of not

jrenee42 (approved):

PR #658, opened 2021-02-10
test: e2e tests now pass in KinD environment

genehynson:
Issue: https://github.com/influxdata/idpe/issues/9547

Explanation of changes:
- Add `cy.getByTestID('tree-nav')` after every `cy.visit` in a `beforeEach`. Why: The app seems to run slower in the cluster and this causes a race condition. Basically the tests start execution before the page fully loads. Waiting on the tree-nav element forces the beforeEach to wait for the page to load before starting the tests. 
- Add `orgID` property to the `resource` object for each permission of each token we create during the token tests. The McFly user used in the kind tests isn't a "super user" like monitor-ci's "dev_user". McFly goes through the typical onboarding process and the IDPE authorization service requires that you provide the `orgID` property unless you have these super permissions.
- Set the `defaultUser` fixture every time after we sign in. This is required because we're creating a new user for every test and need to update this fixture accordingly.
- Reloads before running the "can delete label" test because there is a race condition between how fast the API responds to creating the label and how fast the app is rendering.

Also adds a `yarn start:kind` command for starting up the dev server to be compatible with k8s-idpe

IDPE PR: https://github.com/influxdata/idpe/pull/9813 🚫 
K8S-IDPE PR: https://github.com/influxdata/k8s-idpe/pull/2375
Monitor-CI PR: https://github.com/influxdata/monitor-ci/pull/164 ✅ 

Merge this 2nd

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "tsc:cypress": "tsc -p ./cypress/tsconfig.json --noEmit --pretty --skipLibCheck",
     "cy": "CYPRESS_baseUrl=http://localhost:9999 cypress open",
     "cy:dev": "source ../monitor-ci/.env && CYPRESS_baseUrl=https://twodotoh.a.influxcloud.dev.local cypress open --config testFiles='{cloud,shared}/**/*.*'",
-    "cy:dev-oss": "source ../monitor-ci/.env && CYPRESS_baseUrl=hhttps://twodotoh.a.influxcloud.dev.local cypress open --config testFiles='{oss,shared}/**/*.*'",
+    "cy:dev-oss": "source ../monitor-ci/.env && CYPRESS_baseUrl=https://twodotoh.a.influxcloud.dev.local cypress open --config testFiles='{oss,shared}/**/*.*'",
     "generate": "oats https://raw.githubusercontent.com/influxdata/influxdb/master/http/swagger.yml > ./src/client/generatedRoutes.ts && yarn unity && yarn flows",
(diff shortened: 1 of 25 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

package.json
```diff
@@ -21,6 +21,7 @@
     "start:dev": "webpack-dev-server --config ./webpack.dev.ts --progress false",
     "start:dev-cloud": "cross-env CLOUD_LOGOUT_URL=http://localhost:8080/api/v2/signout CLOUD_URL=http://localhost:4000 webpack-dev-server --config ./webpack.dev.ts",
     "start:docker": "yarn generate && yarn build:vendor && yarn run start:dev",
+    "start:kind": "yarn build:vendor && NODE_ENV=development UI_SHA=local_dev_mode BASE_PATH=/ API_BASE_PATH=/ PORT=8080 PUBLIC=https://twodotoh.a.influxcloud.dev.local/ CLOUD_URL=/auth yarn start",
```
drdelambre, line 24: how do i change twodotoh.a.influxcloud.dev.local to "cool.kids.table"?

drdelambre (approved):
what.. too good for a `cy.wait(10000)` or something?

package.json
genehynson: hmm it's hardcoded in several places in k8s-idpe so I don't think it's expected that you would change this

genehynson: perhaps an improvement for down the road once all the cool kids are using k8s-idpe ;)

PR #664, opened 2021-02-10
fix: empty raw table view and variable refreshes

drdelambre:
apparently we got a bunch of weirdos around here that enable raw data view before even running a query, or expect their table visualizations to update when they change variables.

this one goes out to them


genehynson (approved):

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #665, opened 2021-02-10
feat(managed-functions): Add nav bar item, routes, and landing pages for managed functions pages

ebb-tide:
Closes #9783

### DIFF
--- a/src/functions/containers/FunctionEditPage.tsx
+++ b/src/functions/containers/FunctionEditPage.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {PureComponent} from 'react'
+import React, {FC} from 'react'
 
 // Components
 import {Page} from '@influxdata/clockface'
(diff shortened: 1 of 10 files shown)


src/functions/containers/FunctionEditPage.tsx
```diff
@@ -0,0 +1,25 @@
+// Libraries
+import React, {PureComponent} from 'react'
+
+// Components
+import {Page} from '@influxdata/clockface'
+
+// Utils
+import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
+
+class FunctionEditPage extends PureComponent {
```
drdelambre, line 10: why not functional components for these?

hoorayimhelping: >why not functional components for these?

why functional components?

src/shared/containers/index.tsx
```diff
@@ -73,3 +73,16 @@ export const UsagePage = lazy(() => import('src/usage/UsagePage'))
 export const BillingPage = lazy(() =>
   import('src/billing/components/BillingPage')
 )
+
+export const FunctionsPage = lazy(() =>
```
drdelambre, line 77: can make this function list page? not usually that person, but we had a problem in notebooks where this pluralized naming convention led to some mishaps during a refactor, just cause it's hard to see the `s`. if no can, no can

src/functions/containers/FunctionEditPage.tsx
drdelambre: we've been using them more and more for new components as an informal convention because hooks are dope and make state management tasks easier to implement without as much routing, curious what a pure component buys us / if it was intentional for some reason

drdelambre (approved):
i like hooks, but go nuts

src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -216,5 +216,17 @@ export const generateNavItems = (orgID: string): NavItem[] => {
         },
       ],
     },
+    {
```
drdelambre, line 219: @hoorayimhelping found another giant config object! not advocating for using context on this, but maybe it'd save some time if we added routes to this object and defaulted to lazy loading them?

drdelambre: the SetOrg looks repetitive

src/functions/containers/FunctionEditPage.tsx
ebb-tide: just being old-skool I guess! functionalized!

src/shared/containers/index.tsx
ebb-tide: can can

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #666, opened 2021-02-11
chore(unity): refine endpoints to only return whats needed

ischolten:
Worked with @subirjolly and @gspletzer to update unity swagger so that the api is returning what the UI needs. 

- [ ] Rebased/mergeable
- [ ] Tests pass


src/client/swagger/unity.yml
```diff
@@ -23,174 +23,197 @@ paths:
...
+                  ,result,table,_field,_time,_value
+                  ,,0,reads_gb,2021-01-13T22:05:00Z,120
         default:
           description: Unexpected error
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/billing/contact':
+  '/settings/notifations':
```
marcdel, line 138:
```suggestion
  '/settings/notifications':
```

randycoulman (approved):
I didn't deep dive on this, but from what I can see, it looks reasonable to me as a starting point. I expect we'll learn some things as we implement it and will make some tweaks.

Approving now so we can get this in before migrating it to a new home, but please make @marcdel's suggested spelling change before merging.

src/client/swagger/unity.yml
```diff
@@ -5,14 +5,14 @@ info:
...
-  '/billing/account':
+  '/account':
     get:
       operationId: GetAccount
       tags:
         - Account
       responses:
         '200':
-          description: Account info for billing
+          description: the account
```
marcdel, line 15: Maybe I'm thinking of this too much as a traditional REST API, but it feels really strange everything has this implicit coupling to the user's session. Though I suppose the other option is `/account/:id` and then you're implicitly coupled to the session by way or authorization (checking the logged in user's account id matches the requested one) anyway.

src/client/swagger/unity.yml
```diff
@@ -459,106 +543,64 @@ components:
...
-          billingContact,
-          deletable,
-          marketplaceSubscription,
-          pricingVersion,
-          type,
-          updatedAt,
-          users,
-          zuoraAccountId,
-        ]
+      required: [id, marketplace, type]
```
marcdel, line 591: I can see this one being tricky depending on the context(s) in which it's used

src/client/swagger/unity.yml
```diff
@@ -23,174 +23,197 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/billing/date':
+  '/billing':
```
marcdel, line 26: I don't love this top level resource being _a_ billing, because that's...not a noun. I'm not a fan of slapping an `info` suffix on things either though, and I can't think of a better name for this data clump.

These fields make sense under `Account` i.e. account.balance, account.payment_method, etc. but then you've got drag all these extra fields around anywhere you want an account.

src/client/swagger/unity.yml
```diff
@@ -23,174 +23,197 @@ paths:
...
-                $ref: '#/components/schemas/Region'
+                $ref: '#/components/schemas/Marketplace'
         default:
           description: Unexpected error
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/account/history':
+  '/billing/start_date':
```
marcdel, line 62: Is this something that could be returned as part of the `/billing` payload rather than making a separate call for it?

src/client/swagger/unity.yml
```diff
@@ -23,174 +23,197 @@ paths:
...
+
         default:
           description: Unexpected error
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-    put:
-      operationId: PutBillingContact
+  '/usage/{type}':
```
marcdel, line 99: I feel weird about having these at the same level as `/usage/vectors`. Assuming `{type}`s are vectors, I would expect this to be `/usage/vectors/:type`. If that's the case I'd probably make the route `/usage_vectors` instead?

src/client/swagger/unity.yml
```diff
@@ -23,174 +23,197 @@ paths:
...
-              $ref: '#/components/schemas/BillingNotifySettings'
+        - Usage
+      parameters:
+        - in: path
+          name: type
+          required: true
+          schema:
+            type: string
+            enum: [limits, billing, storage_gb, write_mb, reads_gb, query_count]
+          description: The usage type or vector to query. Can be a usage metric, billing stats, or rate limit events
```
marcdel, line 111: Are type and vector synonyms? If so, we should prob pick one or the other for these two endpoints.

src/client/swagger/unity.yml
marcdel: Is this distinct enough from a `Billing` to warrant its own object/route? Wonder if it makes sense to roll this under that section.

src/client/swagger/unity.yml
marcdel: Don't 🔫 me but if I were starting from scratch I'd probably lean more towards a use-case based model than a resource based one. Then split out resources if/when duplication becomes apparent. That's personal preference though, and I think the scope of these pages is small enough (for now) that either way will work fine.

src/client/swagger/unity.yml
gspletzer: That's a good point, but also can't think of a better name right now; will consider other options as we implement it.

src/client/swagger/unity.yml
gspletzer: That's a good point. This isn't really for billing; it's part of the usage page. We will see if we can include it in the usage vectors payload or similar.

src/client/swagger/unity.yml
gspletzer: Will replace `type` with `vector_name`

src/client/swagger/unity.yml
gspletzer: Because there is a `put` route for it as well, it makes sense to keep it separated.

src/client/swagger/unity.yml
```diff
@@ -593,14 +635,47 @@ components:
...
+          type: string
+        balanceThreshold:
+          description: threshold for notification
+          type: string
+        notify:
+          description: whether the user will be notified about their balance
+          type: boolean
     Invite:
       properties:
         id:
```
gspletzer, line 681: This description should probably specify Quartz ID
```suggestion
        id:
          description: the quartz id of the invite
          readOnly: true
          type: integer
```

src/client/swagger/unity.yml
```diff
@@ -300,23 +336,23 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/orgs/{orgID}/invites/{inviteID}':
+  '/orgs/{orgId}/invites/{inviteId}':
     delete:
-      operationId: DeleteOrgsIDInviteID
+      operationId: DeleteOrgsIdInviteId
```
OfTheDelmer, line 341: It might be worth noting in the response that you will get a `404` from the delete on a retry (not idempotent) or if the org id doesn't match. This Quartz behavior differs from IDPE behavior where @glinton has structured deletes to be idempotent.

OfTheDelmer: @randycoulman thoughts?

randycoulman: My initial thought is that a 404 is the right response if the org ID is missing. But if there is no invite with that ID, then we should return a successful response in order to make this idempotent.

src/client/swagger/unity.yml
randycoulman: Invites only exist in Quartz and not IDPE, so it's pretty much implied that this is a Quartz ID. But not a bad idea to be explicit about it.

src/client/swagger/unity.yml
OfTheDelmer: Agreed, I'll update our issue to reflect retry should return a successful response.

src/client/swagger/unity.yml
Asalem1: Did this ever get changed to be a part of Billing rather than it's own definition?

src/client/swagger/unity.yml
```diff
@@ -23,174 +23,197 @@ paths:
...
         default:
           description: Unexpected error
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/billing/cc':
-    get:
-      operationId: GetCCPageParams
+  '/billing/contact':
```
Asalem1, line 180: Quick question: why is the billing contact a separate API query here? Wouldn't that be considered part of the [billing information](https://github.com/influxdata/ui/pull/666/files#diff-888a46184afed8040a74345087d09db51553252713b486dbd57610e728c89d02R33)?

src/client/swagger/unity.yml
```diff
@@ -774,3 +864,140 @@ components:
...
+          example: 'http,pagerduty'
+          type: string
+    RestrictedLimit:
+      description: Restrict access entirely
+      type: integer
+      enum: [-1]
+    Unlimited:
+      description: Unbounded limit
+      type: integer
+      enum: [0]
```
Asalem1, line 1003: Are the Limits necessary? The UI currently gets that data from IDPE

src/client/swagger/unity.yml
```diff
@@ -420,6 +456,54 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  '/orgs/{orgId}/limits':
```
Asalem1, line 459: I think this is already implemented here:

https://github.com/influxdata/ui/blob/master/src/cloud/components/GetAssetLimits.tsx

Where we get this https://github.com/influxdata/ui/blob/master/src/cloud/apis/limits.ts#L20

Which is defined here:

https://github.com/influxdata/ui/blob/master/src/types/cloud.ts#L1

src/client/swagger/unity.yml
randycoulman: Billing contact is a separate entity in Quartz. It is added on the checkout page, but can be viewed and edited on the billing page.

src/client/swagger/unity.yml
randycoulman: Assuming that the user has permissions to read those limits, it's probably fine to get them directly from IDPE. I thought we needed Quartz' operator token for that, but if not, then we're good there.

src/client/swagger/unity.yml
randycoulman: We'll need them for the operator UI because it provides the ability to change limits for a user. A user wouldn't have permission to change their own limits.

src/client/swagger/unity.yml
Asalem1: Gotcha. Thanks!

src/client/swagger/unity.yml
Asalem1: Ya, I don't think there should be an issue getting it from IDPE. Especially since this implementation won't focus on integrating Cancelled users

src/client/swagger/unity.yml
Asalem1: Gotcha.  I think i bundled this up with the org limits stuff in the other comment. That makes sense, not entirely certain of the dynamic but it seems unique

src/client/swagger/unity.yml
randycoulman: No, it didn't. Feel free to move it where it makes sense for you, or we can look at options when we implement the usage page API.

Asalem1: Great! Thanks Randy!

src/client/swagger/unity.yml
ischolten: So this limits section will need to exist in the quartz api for the operator pages (not for the billing ones though). I think we can lock it down to be an operator. The main reason for this is  if I am an operator, i need to be able to view and update limits for any org no matter if my idpe user has access to it or not. For this reason, we will need to get these through quartz because the quartz token does have access to view these.

src/client/swagger/unity.yml
ischolten: This makes me think that there also needs to be a get for /billing/contact for displaying the existing billing contact on the billingn page. Rather than just updating it.

ischolten: Oh just kidding, i see thats a part of just `/billing`. That seems weird to me, but ill leave it for now.

src/client/swagger/unity.yml
ischolten: Im going to leave this as is for now unless, @Asalem1  wants to rename it or move it. I dont want to keep adding things to `/billing` since that endpoint seems to be providing more of a blob of things to render on the page rather than an actual resource.

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

src/billing/components/BillingPage.tsx
```diff
@@ -52,7 +52,9 @@ function BillingPage() {
         <Page titleTag="Billing">
           <Page.Header fullWidth={false} testID="billing-page--header">
             <Page.Title title="Billing" />
-            <LimitChecker>{!isCancelled && <RateLimitAlert />}</LimitChecker>
+            <LimitChecker>
+              <RateLimitAlert />
```
Asalem1, line 56: removed isCancelled since it's no longer relevant to the UI (i.e. Cancelled users can't access the app)

src/billing/components/BillingPageContents.tsx
```diff
@@ -4,22 +4,18 @@ import React, {FC} from 'react'
...
 
 const BillingPageContents: FC = () => {
   const [{account}] = useBilling()
 
-  switch (account.type) {
-    case 'free':
-      return <BillingFree />
-    case 'pay_as_you_go':
-      return <BillingPayAsYouGo />
-    case 'cancelled':
```
Asalem1, line 20: Removed cancelled since it's no longer relevant to the UI

src/usage/UsagePage.tsx
```diff
@@ -1,87 +1,31 @@
...
+    <Page titleTag="Usage">
+      <Page.Header fullWidth={false} testID="usage-page--header">
+        <Page.Title title="Usage" />
+        <LimitChecker>
+          <RateLimitAlert />
+        </LimitChecker>
+      </Page.Header>
+      <Page.Contents scrollable={true}>
+        {/* TODO(ariel):  find out if we want this */}
+        {/* {isCancelled && <AlertStatusCancelled />} */}
```
Asalem1, line 24: Still need to see if this is relevant for the Operator page

ischolten: I believe the AlertStatusCancelled was used for the old operator ui and is only used on regular user pages like billing, users, and usage. So it should be fine to delete.

Asalem1: Cool, removed it, thanks!

PR #678, opened 2021-02-11
feat: unifying app setting context

drdelambre:
we had a bunch of places where the app settings (timezone, presentation mode, nav bar state, theme) were being pulled from different means (sometimes selectors, sometimes connectors with straight prop access) and at different levels, where we would prop drill them 3 or 4 levels deep.

this adds a redux backed context provider so that everything can access it nicely in the same way


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

genehynson (approved):

PR #679, opened 2021-02-11
chore: upgrading to clockface 2.6.1

jrenee42:
* also: changing test as needed for clockface 2.6.1;
     This necessitated:
         - improving the component as well (you can now click on the label)
	 - the test now accesses the togglebutton via the label text; which is what the user would do (click on the label)

The FillValue's label is now a child of the togglebutton; so the component hierarchy has changed to be consistent with what the togglebutton is expecting; which takes advantage of the fact that the entire togglebutton (including the children) are now clickable in 2.6.1 

the togglebutton changed just enough to break the querybuilder test, which highlighted/took advantage of  a problem with the test (and/or the togglebutton), so it is now fixed.  

having just one PR with *just* the upgrade and the changes needed for the upgrade.

### DIFF
--- a/cypress/e2e/shared/queryBuilder.test.ts
+++ b/cypress/e2e/shared/queryBuilder.test.ts
@@ -123,8 +123,9 @@ describe('The Query Builder', () => {
       cy.getByTestID('auto-window-period').click()
 
       cy.getByTestID('duration-input--error').should('not.exist')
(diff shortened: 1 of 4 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

hoorayimhelping (approved):

PR #690, opened 2021-02-12
feat(managed-functions): Add functions list page that is hydrated from context

ebb-tide:
Closes #9784 and #9785
I'm currently mocking the responses from the API, and add and remove functions don't yet work those are up next!
Please let me know if I'm using context variable and file names wrong. Keeping in mind that this is very much placeholder will be thrown out one day code! 

Screen Shot 2021-02-16 at 9 11 09 AM

### DIFF
--- a/src/functions/components/FunctionCard.tsx
+++ b/src/functions/components/FunctionCard.tsx
@@ -1,0 +1,56 @@
+import React, {FC} from 'react'
+import {useParams, useHistory} from 'react-router-dom'
+
+// Components
+import {
+  AlignItems,
+  ComponentSize,
+  FlexDirection,
+  ResourceCard,
+  FlexBox,
+} from '@influxdata/clockface'
+import FunctionContextMenu from 'src/functions/components/FunctionContextMenu'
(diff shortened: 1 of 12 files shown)


src/functions/components/FunctionCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+import FunctionContextMenu from 'src/functions/components/FunctionContextMenu'
+import LastRunFunctionStatus from 'src/functions/components/LastRunFunctionStatus'
+
+interface Props {
+  id: string
+  name: string
+}
+
+const FunctionCard: FC<Props> = ({id, name}) => {
+  const {orgID} = useParams<{orgID: string}>()
```
drdelambre, line 21: grab the org from redux using `const { org } = useSelector(getOrg)` here instead. then we have one definition of "what is the current org" for when we finally add an org selector in there

drdelambre (approved):
:fire: :cake: :fire:

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #698, opened 2021-02-16
fix: trim url and add error for spaces in between characters in url

### LINKED ISSUE #697: Template Uploader Says Valid URL But Fails Upload
When uploading a template via URL the URL input shows that the input is valid, but then fails on upload.

ShmuelLotman:
Closes #697 

This PR introduces an error message for the template url uploader when there is a space in between characters in the url and it also trims the url before submission.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/templates/containers/CommunityTemplatesIndex.tsx
+++ b/src/templates/containers/CommunityTemplatesIndex.tsx
@@ -245,7 +245,9 @@ class UnconnectedCommunityTemplatesIndex extends Component<Props, State> {
 
     try {
       event('template_click_lookup', {
-        templateName: getTemplateNameFromUrl(this.props.stagedTemplateUrl).name,
+        templateName: getTemplateNameFromUrl(
+          this.props.stagedTemplateUrl.trim()
+        ).name,
       })
       this.props.history.push(
         `/orgs/${this.props.org.id}/settings/templates/import`
@@ -262,7 +264,7 @@ class UnconnectedCommunityTemplatesIndex extends Component<Props, State> {
     const validationMessage = validateTemplateURL(event.target.value)
 
     this.setValidationMessage(validationMessage)
-    this.props.setStagedTemplateUrl(event.target.value)
+    this.props.setStagedTemplateUrl(event.target.value.trim())
   }
 
   private handleInputKeyPress = event => {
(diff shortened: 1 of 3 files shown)


src/templates/utils/index.ts
```diff
@@ -169,6 +169,10 @@ export const validateTemplateURL = (url): string => {
     return "We can't use that URL"
   }
 
+  if (cleanUrl.includes(' ')) {
+    return 'URL cannot contain spaces'
```
hoorayimhelping, line 173: the first thing is, i think this would be easier to follow if it was moved to right under where `cleanUrl` is defined. think about the flow of this function. if the template has a space in it, we don't care if it's a community or a correct file type, it's immediately wrong and unparse-able. it's a stronger indicator and so should come first.

the other thing is, the message is kind of mechanical compared to the other messages in this file, which are a little bit more conversational.

maybe something as simple as adding `the` or `your` and dropping the `cannot`

>"Your URL can't contain spaces"

hoorayimhelping (changes requested):
minorish nits but worth tweaking just for consistency's sake

src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -259,10 +259,11 @@ class UnconnectedCommunityTemplatesIndex extends Component<Props, State> {
   }
 
   private handleTemplateChange = event => {
-    const validationMessage = validateTemplateURL(event.target.value)
+    const trimmedValue = event.target.value.trim()
```
hoorayimhelping, line 262: naming nit: `value` is a vague name. `trimmedUrl` is much more self-documenting.

src/templates/utils/index.ts
```diff
@@ -146,6 +146,10 @@ export const validateTemplateURL = (url): string => {
   */
   const cleanUrl = url.trim().split('?')[0]
 
+  if (cleanUrl.includes(' ')) {
+    return `Your URL can't contain spaces`
```
hoorayimhelping, line 150: context nit: a string wrapped in \`backticks\` implies there will be string interpolation. Better to use double quotes here, which in our codebase are only used when there's a single quote in the string.

hoorayimhelping (approved):
looks good - some minor naming nitpicks

Issue #408, opened 2020-12-07
URL Redirect Layer


alexpaxton:
**Proposal:**
We (as in the design team or anyone running UX experiments) need a means of linking to specific areas of the UI from external sources such as email campaigns.

**Current behavior:**
This is impossible because every authenticated URL contains `orgID`, cloud provider, and region that we can't stick that in email links

**Desired behavior:**
**Redirect with parameter handling.**

A URL we can link to that redirects to specific locations in the UI. This link should be provider, org, and region agnostic. For example we'd want to link to the Notebooks page from an email but not have to worry about `orgID` and whatnot. Ideally this is something that can be built once and unblock non-engineers going forward

**Alternatives considered:**
Tried doing this through Marketo but we'd still need to get `orgID`, `provider` and `region` from salesforce (which didn't work out). 

**Use case:**
This drives engagement and internal understanding of users

drdelambre:
you want to link to `/dashboards` and have it go to `/orgs/{whatever}/dashboards`? got lost in the whatnot

alexpaxton:
@drdelambre yeah pretty much. This could probably be broken down into smaller pieces, for example, handling dashboards first and then adding redirects for different parts of the UI as needed

barbaranelson:
Note: if we redirect via cloud, make sure we don't drop parameters, as currently happens for cloud2.influxdata.com (which is why design team uses the usage page URL).

http://cloud2.influxdata.com/usage/?walkme=14-301843

corriganjohn:
After talking with @121watts  and Marc Delagrammatikas (ecomm team) morning --- this seems to be a relatively simple addition to what we are currently doing in the cloud2.influxdata.com redirect. We should be able to add a new routing (eg '/append_route') that takes the current routing we already build (region/provider/cloud2 url / orgid) and allows it to accept a string value where we can append the desired route and parameters.
Should this be an ecommerce issue?

barbaranelson:
If it is in quartz code, then yes, please file a separate issue in the quartz project and we can take care of it there.

drdelambre:
i totally want to get rid of the /org/{orgid} prefix for all of our urls, which should also satisfy this. all we need to add is an org switcher, and a little bit of code that removes that prefix for any user with old links saved.

idk if arbitrary string execution / routing is a good move, usually it's not

subirjolly:
Do we still need this issue?

barbaranelson:
[...]

PR #699, opened 2021-02-16
fix: single source of truth for accessing the current org

drdelambre:
We have weird support for multi org users right now, and our app routing structure locks us into tightly coupling the concept of a user being within an organizational context, and prevents us from exploring other sharing models on the UX.

This PR decouples the route from the apps understanding of the currently selected organization. from the looks of it, most of this should get removed when we support multi org switching, as many cases just load the org to push it into a history.push command, which should get depreciated as we sort through the multi org / multi region shenanigans, because it'll get removed from the url being pushed.

my current thinking to remove the org from the url is:
 1. move the app to a single source of truth about the currently selected organization, defined in redux with `getOrg`
 2. add a mechanism within the ux for a user to switch their orgs
 3. upon deeplinking to anything `/org/:orgid/_xyz_`, set the current org to `orgid`, and redirect the route to `/_xyz`
 4. clean up anything pushing org.id into the url in the first place.
 
this should seamlessly migrate users with bookmarks, and allow us to take the migration piecemeal as we find enthusiasm for the project, while better supporting our multiorg users.

all of this is required for: https://github.com/influxdata/ui/issues/408

### DIFF
--- a/src/buckets/components/BucketList.tsx
+++ b/src/buckets/components/BucketList.tsx
@@ -1,7 +1,7 @@
 // Libraries
-import React, {PureComponent} from 'react'
-import {withRouter, RouteComponentProps} from 'react-router-dom'
-import memoizeOne from 'memoize-one'
+import React, {FC, useCallback, useMemo} from 'react'
+import {useHistory} from 'react-router-dom'
+import {useSelector} from 'react-redux'
 
 // Components
 import BucketCard from 'src/buckets/components/BucketCard'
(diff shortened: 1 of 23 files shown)


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

drdelambre:
the plan still stays the same, but if we change the order to:
 1. add an org switcher
 2. add redirect that removes the orgID params

then it should get rid of most of these changes and expose the limited cases where we didn't just pump the orgID right back into a history.push (because we will just remove them with #2)

PR #706, opened 2021-02-17
feat(managed-functions): Add create and edit pages for functions

ebb-tide:
Closes #9786 #9787

I decided to put function editor state in component state for now, we can refactor to pull it out in to context when we need persistence. 

Screen Shot 2021-02-17 at 12 54 18 PM


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

src/functions/components/FunctionCard.tsx
```diff
@@ -38,16 +39,15 @@ const FunctionCard: FC<Props> = ({id, name}) => {
...
       <FlexBox
         alignItems={AlignItems.FlexStart}
         direction={FlexDirection.Column}
         margin={ComponentSize.Medium}
       >
         <ResourceCard.Name name={name} onClick={handleClick} />
         <ResourceCard.Meta>
-          <>Last triggered 3 minutes ago</>
-          <>Trigger URL: http://moo.influxdata.com/yerfunctionisawesome</>
+          <>{description && description}</>
```
drdelambre, line 49: can't you just put... description?

ebb-tide: I can.. I had something else there before.. :)

src/functions/components/FunctionCard.tsx
```diff
@@ -11,17 +11,18 @@ import {
...
 import FunctionContextMenu from 'src/functions/components/FunctionContextMenu'
-import LastRunFunctionStatus from 'src/functions/components/LastRunFunctionStatus'
 
 // utils
 import {getOrg} from 'src/organizations/selectors'
 
 interface Props {
   id: string
   name: string
+  url?: string
```
drdelambre, line 21: instead of extending this props interface, you have access to the id and the function list from the context, and it seems like you've moved the functions to a hash, so why not
```
... ({ id }) => {
const {functionList} = useContext(FunctionListContext)
const { name, url, description } = functionList[id]
...
```
so that you can decouple the needs of this child component from the interface defined in the parent component. the parent shows a list, so it's always going to have some sense of "the order of things in a list" or of the list itself, but it shouldn't need to know that the items within the list require a name or url or description

src/functions/components/FunctionContextMenu.tsx
```diff
@@ -9,18 +9,23 @@ import {ButtonShape, ComponentColor, IconFont} from '@influxdata/clockface'
...
 const FunctionContextMenu: FC<Props> = ({id, name}) => {
+  const {remove} = useContext(FunctionListContext)
+
   const {id: orgID} = useSelector(getOrg)
 
   const history = useHistory()
 
-  const handleDelete = () => {}
+  const handleDelete = () => {
+    remove(id)
```
drdelambre, line 27: man.. that const {id} = useSelector(getOrg) gave me heart palpitations when i hit this remove. not used to the renaming syntax yet. all good though.

src/functions/containers/FunctionEditPage.tsx
```diff
@@ -1,21 +1,107 @@
...
+      <FunctionHeader
+        name={name}
+        saveFunction={updateFunction}
+        cancelFunction={cancelFunction}
+      />
+      <Page.Contents fullWidth={true} scrollable={true}>
+        <div className="function-form">
+          <div className="function-form--options">
+            <FunctionForm
+              name={name}
```
drdelambre, line 72: this interface feels weird. to pump all of these in individually, so that they can share state... idk what to do about it though, so lemme get back to it

src/functions/containers/FunctionEditPage.tsx
```diff
@@ -1,21 +1,107 @@
...
+  const {id: orgID} = useSelector(getOrg)
+
+  const [name, setName] = useState('New Function')
+  const [description, setDescription] = useState('')
+  const [params, setParams] = useState('')
+  const [script, setScript] = useState('')
+
+  const {functionsList} = useContext(FunctionListContext)
+  useEffect(() => {
+    const {name, script, description} = functionsList[functionID]
```
drdelambre, line 39: this kind of feels like the `this.setState(this.props)` pattern that causes us a bunch of problems (cause we seem to forget edge cases of mutations usually), but we also don't have a draft pattern yet...

src/functions/containers/FunctionEditPage.tsx
```diff
@@ -1,21 +1,107 @@
...
+        </div>
+      </Page.Contents>
     </Page>
   )
 }
 
-export default FunctionEditPage
+const FunctionEditPageWithProvider: FC = () => {
+  return (
+    <FunctionListProvider>
```
drdelambre, line 101: in some places for forms / overlays that involve prop drilling, we give it a new context provider just for that interface

src/functions/containers/FunctionEditPage.tsx
```diff
@@ -1,21 +1,107 @@
...
+  const [script, setScript] = useState('')
+
+  const {functionsList} = useContext(FunctionListContext)
+  useEffect(() => {
+    const {name, script, description} = functionsList[functionID]
+    // TODO what if functionID no exist??
+    setName(name)
+    setScript(script)
+    setDescription(description)
+  }, [functionsList, functionID, setName, setScript, setDescription])
```
drdelambre, line 44: i'm circling here, looking for words. but this isn't what you want right? if you have the page open in two tabs, or a server push updates the list of functions because someone is working on the same page state, you dont want to _reload_ the function into the form's state right? but use this more like a `componentDidMount`?

src/functions/components/FunctionCard.tsx
ebb-tide: getting name description and url from context now

src/functions/components/FunctionContextMenu.tsx
ebb-tide: I'll call it functionID :)

src/functions/context/api.tsx
```diff
@@ -29,16 +31,43 @@ export const deleteAPI = async (id: string) => {
...
-    // if (res.status != 200) {
-    //   throw new Error(res.data.message)
-    // } TODO
-    return {
-      functions: [
-        {name: 'functionb', id: '1', orgID: '0', script: 'lalal'},
-        {name: 'functiona', id: '2', orgID: '0', script: 'lalal'},
-      ],
+    const res = await getApiV2Functions({query: {orgID}})
+    if (res.status != 200) {
```
drdelambre, line 35: lol we seperated `notebooks` and `notebooks-api` flags so that we could independantly test the ux and the api layer as we were developing them. if notebooks-api was off, it would generate ids on saves and noop on updates and removals, while updates and adds still worked on local storage. when the flag was on, we'd still write to local storage, but also send off requests to the backend

drdelambre: i do like this rollover mechanism in lieu of that though, just for dev and demo purposes

drdelambre (approved):
thanks for changing the name to draft function and introducing the new paradigm pattern.

i'm normally not that person who cares about names, but it makes it so much easier for me to grasp the nuance of what's going on in here. And great work consolidating it all into the provider, much cleaner

src/functions/components/FunctionRunsList.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {ComponentSize, EmptyState, IndexList} from '@influxdata/clockface'
+import {FunctionListContext} from 'src/functions/context/function.list'
+import FunctionRunsRow from './FunctionRunsRow'
+
+const FunctionRunsList: FC = () => {
+  const {id: functionID} = useParams<{id: string}>()
+  const {runsList, getRuns} = useContext(FunctionListContext)
+
+  useEffect(() => {
+    getRuns(functionID)
```
drdelambre, line 15: should this happen in the provider when you set the current draft state?

ebb-tide: :( the draft state function and function to get runs for are different unfortunately. the routing branches off at functions list and you can go to edit a function or you can view its runs.

PR #710, opened 2021-02-17
feat: adding a downsample panel

### LINKED ISSUE #661: Notebooks: Downsampling Panel
we have an aggregate window interface that we are removing from the metric selector and query builder, but that doesn't erase the need to downsample in this way. lets make it it's own panel so that people can be explicit about the intent to down sample their data

drdelambre:
Peek 2021-02-17 11-33

Closes #661

context is in the issue, but we're trying to remove the implicit aggregate window from the ux. this, doesn't negate the need to aggregate window being used to downsample, just remove it from places it shouldn't be. this moves the windowing into an explicit action by the user.

so i added it here

totally needs design work before opening up to users, but all of the plumbing is there.

use `influx.set('flow-panel--downsample', true)` to turn it on


drdelambre:
please ignore the line count. all greatness is contained in the first commit.

seems like we're enforcing linting on sass files now? but.. it was never run on the existing codebase? so.. everything will explode if we try to revert in the future. needed to get prettier to pass anyways, so i ran the command.

hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

drdelambre:
now looks like this:
Screenshot from 2021-02-17 16-47-00

src/flows/pipes/Downsample/view.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+import {PipeProp} from 'src/types/flows'
+
+const Downsample: FC<PipeProp> = ({Context}) => {
+  const {data, update} = useContext(PipeContext)
+
+  const selectFn = fn => {
+    const fns = [...data.functions]
+    const fnsFound = fns.map(f => f.name).indexOf(fn)
+    if (fnsFound !== -1) {
```
genehynson, line 29: could this be `fns.find(f => f.name == fn)` and then `if (fnsFound) {` ?

drdelambre: we need the index for the splice immediately after

genehynson: _le sigh_ ... this is why I shouldn't try and multi-task

src/flows/pipes/Downsample/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        const fnSpec = fn && FUNCTIONS.find(spec => spec.name === fn.name)
+
+        if (!fnSpec) {
+          return ''
+        }
+
+        const period = config.aggregateWindow.period
+        const flux = fnSpec.flux(
+          period === 'auto' || !period ? 'v.windowPeriod' : period,
```
genehynson, line 45: wait so this expression either returns `true`, `v.windowPeriod`, or whatever the value of `period` is? Or am I mis-reading it?

genehynson: me rn 
https://media.giphy.com/media/4JVTF9zR9BicshFAb7/giphy.gif

drdelambre: lol idk i copy pasta'd it from the other working examples in our code base. read it as: if period is `auto` or falsy, use v.windowPeriod to inherit it from the application state, or else hardcode the period set by the user

src/flows/pipes/Downsample/view.tsx
```diff
@@ -0,0 +1,114 @@
...
+            style={{width: '250px'}}
+            options={FUNCTIONS.map(f => f.name)}
+            selectedOptions={data.functions.map(fn => fn.name)}
+            onSelect={selectFn}
+            buttonColor={ComponentColor.Secondary}
+            buttonIcon={IconFont.BarChart}
+          />
+        </div>
+        <div className="downsample-panel--right">
+          <h5>Window Period</h5>
```
mnorriswalker, line 75: Input label should say "Every Window Period"? Trying to connect the concept of "Apply Aggregate" to "Every Window Period"

src/flows/pipes/Downsample/index.ts
genehynson: lol ok. I'm sure it's fine. JS is so weird sometimes

genehynson (approved):

PR #716, opened 2021-02-17
feat: only render and annotate when annotate mode is active

### LINKED ISSUE #711: Annotations: only add annotations with double click when annotation mode is active
### Background
Currently, any double click will create an annotation on the graph if the feature flag is toggled, regardless of whether annotation mode is toggled. Make it so that double click only gets overwritten in annotation mode.

### Acceptance Criteria
- Annotations are only added on double click when the app is in annotation mode (i.e. when a user clicks Annotate next to the Variables button)

hoorayimhelping:
Closes #711

- Only allows annotating when the annotation toggle is active.
- Viewing annotations is only possible when the annotation toggle is active

in

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/visualization/types/Graph/view.tsx
+++ b/src/visualization/types/Graph/view.tsx
@@ -195,6 +195,8 @@ const XYPlot: FC<Props> = ({properties, result, timeRange, annotations}) => {
     ],
   }
 
+  // TODO: address this tech debt
+  // see https://github.com/influxdata/ui/issues/725
   if (isFlagEnabled('annotations') && annotationsModeIsActive) {
     const createAnnotation = userModifiedAnnotation => {
       const {message, startTime} = userModifiedAnnotation
--- a/src/visualization/types/Graph/view.tsx
+++ b/src/visualization/types/Graph/view.tsx
@@ -1,6 +1,6 @@
 // Libraries
 import React, {FC, useMemo, useContext} from 'react'
-import {useDispatch} from 'react-redux'
+import {useDispatch, useSelector} from 'react-redux'
 import {
   AnnotationLayerConfig,
   Config,
@@ -13,6 +13,22 @@ import {
 
 // Components
 import EmptyGraphMessage from 'src/shared/components/EmptyGraphMessage'
+
+// Context
+import {AppSettingContext} from 'src/shared/contexts/app'
+
+// Redux
+import {writeThenFetchAndSetAnnotations} from 'src/annotations/actions/thunks'
+import {showOverlay, dismissOverlay} from 'src/overlays/actions/overlays'
+
+// Constants
+import {VIS_THEME, VIS_THEME_LIGHT} from 'src/shared/constants'
+import {DEFAULT_LINE_COLORS} from 'src/shared/constants/graphColorPalettes'
+import {INVALID_DATA_COPY} from 'src/visualization/constants'
+
+// Types
+import {AppState, XYViewProperties} from 'src/types'
+import {VisualizationProps} from 'src/visualization'
 
 // Utils
 import {useAxisTicksGenerator} from 'src/visualization/utils/useAxisTicksGenerator'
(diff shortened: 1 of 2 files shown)


src/visualization/types/Graph/view.tsx
```diff
@@ -14,6 +14,22 @@ import {
...
+
+// Constants
+import {VIS_THEME, VIS_THEME_LIGHT} from 'src/shared/constants'
+import {DEFAULT_LINE_COLORS} from 'src/shared/constants/graphColorPalettes'
+import {INVALID_DATA_COPY} from 'src/visualization/constants'
+
+// Types
+import {AppState, XYViewProperties} from 'src/types'
+import {VisualizationProps} from 'src/visualization'
+
```
hoorayimhelping, line 32: dumb diffing tool. i moved the `// Utils` section under this section and it thinks i moved this section up.

src/visualization/types/Graph/view.tsx
```diff
@@ -223,23 +195,55 @@ const XYPlot: FC<Props> = ({properties, result, timeRange, annotations}) => {
...
+      )
+    }
+
+    config.interactionHandlers = {
+      doubleClick: doubleClickHandler,
+    }
+
     // everything is under the 'default' category for now:
-    const actualAnnotations: any[] = annotations.default ?? null
+    const selectedAnnotations: any[] = annotations?.default ?? []
```
hoorayimhelping, line 232: combined the check above (`if (isFlagEnabled('annotations') && annotations)`) into this `annotations?.`

src/visualization/types/Graph/view.tsx
```diff
@@ -223,23 +195,55 @@ const XYPlot: FC<Props> = ({properties, result, timeRange, annotations}) => {
...
 
-    if (actualAnnotations && actualAnnotations.length) {
+    if (selectedAnnotations.length) {
```
hoorayimhelping, line 234: since it defaults to an empty array, we can just check the length here, rather than existence *and* length.

src/visualization/types/Graph/view.tsx
```diff
@@ -64,6 +66,9 @@ const XYPlot: FC<Props> = ({properties, result, timeRange, annotations}) => {
   )
 
   const dispatch = useDispatch()
+  const annotationsModeIsActive = useSelector(
+    (state: AppState) => state.userSettings.showAnnotationsControls
```
drdelambre, line 70: know of any more of these one off user settings? i've been collecting them: https://github.com/influxdata/ui/blob/master/src/shared/components/TimeZoneDropdown.tsx#L12

src/visualization/types/Graph/view.tsx
jrenee42: the clause in the if statement (starting on line 198) is really long.  it is almost 60 lines; more than fits on the screen (I had to paste it into an editor to check it to understand the code).  please put it in its own method for clarity.  

just pass the config object to the method, and the method can adjust the config object; and it will be much clearer.

jrenee42 (commented):
looks good; just needs a little refactoring (if clause contents too long!) for clarity.

src/visualization/types/Graph/view.tsx
hoorayimhelping: @jrenee42 i thought a bit about this - I don't want to pull the whole thing out into a function because then it's just moving the big block of code and making the reader go read it somewhere else.

i could put only the important parts inside the flag:

```js
if (flagIsEnabled('annotations') {
  config.interactionHandlers = {
    doubleClick: doubleClickHandler,
  }

  config.layers.push(annotationLayer)
}
```
and pull everything else out of the feature flag check, but then since it's a function component that renders a graph on a hot rendering path, it's going to be constantly re-creating methods it won't use.

i agree that this isn't ideal and that the huge length of this block is a smell that this should be refactored. i'm not completely clear on what needs to be refactored yet and i'm loath to do it too early and do the wrong thing. i think what i'd like to do in a separate step is pull out the functions that aren't required to be in the rendering path and move them higher up the chain. but like i said, i'm not 100% clear on the best path forward and i'd rather wait to do this until the best way forward is a little more clear.

what do you think about me writing a ticket to address this and then adding a comment linking to that ticket? so that we can address it after we get a chance to mess with the annotations a bit more?

jrenee42: i think the above is a good template;  make a method to make the doubleclick handler,
and a method to make the annotationlayer.  our code should be readable/clear when it is first published, if we put refactoring into technical debt it may never happen.

jrenee42 (approved):
approving, giving that more issues are coming to streamline the code

hercules-influx:
Metric | Score | Level
best-practices | 80 | :yellow_circle:
seo | 80 | :yellow_circle:
pwa | null | :question:

PR #731, opened 2021-02-19
feat: moving out aggregate window

### LINKED ISSUE #650: Notebooks: I can query my data without aggregation
As a InfluxDB Cloud user, I can view the unaggregated results of a query so that I can understand the data being returned.

### Scenario: Remove Aggregate Functions
Given: I have a notebook open with a metrics selector cell
When: I select a measurement, tag, and/or field
And: I click preview
Then: There is no aggregate function applied

### Scenario: Too large
When: The query returns too much data
Then: Then a message is shown that the data is truncated and the truncated data is shown

### Technical Notes
  - The aggregation logic is going to move to the visualization cells in the notebook
  - The query built behind the scenes should not have an aggregate attached to it
  - We should investigate ways to limit the data returned, possibly using limit or sample. 100,000 of results will not be useful to the UI. We should target < 100 per series and no more than 20 series if possible.

### Design Notes
- Mockup
  - https://www.figma.com/file/8S8IH8mimYIwanZ6keIQ6U/Notebooks-Prototype?node-id=1631%3A175

### LINKED ISSUE #651: I can aggregate results for visualizations without changing the query
As a InfluxDB Cloud user, I can add one or more visualization cells to a notebook so that I can understand the shape of the data without changing the data used for custom script cells or writing to a bucket:

In other words:
- I can return data for only the size of my screen
- I don't over fetch results that exceed available screen pixels

### Scenario: Aggregate for visualization
Given: I have a valid input
When: I add a new visualization cell 
Then: An aggregate function is automatically applied
And: the function is displayed to the user - https://influxdata.github.io/clockface/?path=/story/components-dropdowns-composed--multiselectdropdown (Note: `DropUp`)
But not the window period

When: I add a new script cell after a passthrough cell (like visualization)
Then: The script I write is executing against the data from before any passthrough cells

### Scenario: Add multiple aggregates
Given: I have a visualization cell with valid inputs -  https://influxdata.github.io/clockface/?path=/story/components-dropdowns-composed--multiselectdropdown (Note: `DropUp`)
When: I select multiple aggregates 
Then: I can see the multiple aggregates visualized

### Technical Notes
  - The query might look something like this:
```
data = from |> range |> filter |> yield
viz_1 = data |> aggWindow |> yield
script_1 = data |> <custom flux code from user>
```
### Design Notes
- Mockup
   - https://www.figma.com/file/8S8IH8mimYIwanZ6keIQ6U/Notebooks-Prototype?node-id=1635%3A10265

drdelambre:
Closes #650 
Closes #651

both of these kind of involved the same work of decoupling the panel from the source query on the query provider side of the house.

none of this stuff makes sense to me anymore, which might mean it's time to refactor the model a bit to better fit current product thinking, i just haven't had the capacity to think about it yet.

dropped in the same aggregate window dropdown as the downsample panel, but that hasn't merged yet so idk where to go with that. should look pretty familiar


hercules-influx:
Metric | Score | Level
seo | 80 | :yellow_circle:
pwa | null | :question:

genehynson (approved):

drdelambre:
closing in favor of #813, which just does the merges in a different file than the diffs could show

PR #741, opened 2021-02-23
feat(managed-functions): Update swagger definitions for functions

ebb-tide:
changes:

- I had prefixed the routes as `api/v2private/functions` so switched that to `api/v2/functions`
- I've changed the response codes to match what is in the rest of our API. 
- added a `get` method to the `/invoke` path. 

REMAINING WORK:
Functions have no concept of `links`. Should they have links? We don't use them anywhere, but maybe others do. 
Functions also do not have labels. Should functions have labels?
The names of the following endpoints is confusing:
- `functions/trigger` 
- `functions/{functionID}/invoke`
- `functions/{functionID}/runs`
The problem is that the word "run" can be used as a verb as in `/functions/{functionID}/run` which would cause an execution of the function, or as a noun  as in`functions/{functionID}/runs` which returns the "runs" of a function. We are somewhat married to the noun usage of "runs" because tasks have runs. Which makes me think we should not use the verb usage. Neither invoke nor trigger are great words. Is `functions/call` better?


src/client/swagger/managedFunctions.yml
```diff
@@ -193,31 +219,72 @@ paths:
...
           schema:
             type: string
           required: true
+        - in: query
+          name: params
+          schema:
+            type: object
+            additionalProperties: true
+          style: form
+          explode: true
```
ebb-tide, line 240: As per the swagger 3.0 spec, this should evaluate to describe that query can be any key=value pair according to [this post on stack overflow](https://stackoverflow.com/questions/38187187/use-object-type-query-param-in-swagger-documentation/51277233#51277233) 
ie `/api/v2/functions/{functionID}/invoke?key1=value1&key2=value2` should be valid. 
but currently oats ignores `additionalProperties:true` and interprets this as:
ie `/api/v2/functions/{functionID}/invoke?params={key1=value1&key2=value2}` (or something like that)
oats needs to be updated to interpret this properly.

src/client/swagger/managedFunctions.yml
```diff
@@ -298,27 +365,52 @@ components:
...
+          enum: [json]
+        headers:
+          type: object
+          properties:
+            contenttype:
+              type: string
+              enum: [application/json]
+            xtestresponse:
+              type: string
+              enum: [unspecified]
```
ebb-tide, line 400: these properties are supposed to be `content-type` and `x-test-response` with dashes!! but oats breaks on these property names... so I am entering them as they are for now until oats can be updated.

src/client/swagger/managedFunctions.yml
ebb-tide: https://github.com/influxdata/idpe/issues/9908

src/client/swagger/managedFunctions.yml
ebb-tide: https://github.com/influxdata/idpe/issues/9908

ebb-tide: I actually decided to remove these, since I'd rather have no definition rather than a definitely wrong one.

src/client/swagger/managedFunctions.yml
```diff
@@ -167,57 +167,94 @@ paths:
...
               schema:
-                $ref: '#/components/schemas/FunctionRun'
+                $ref: '#/components/schemas/FunctionHTTPResponseData'
+        default:
+          description: Unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  '/api/v2/functions/{functionID}/runs':
```
wojciechka, line 231: Did we remove POST to `/api/v2/functions/{functionID}/runs` to trigger a new run of a stored function?

I suppose we don't need it for the UI - as the UI always calls the `trigger` URL, but I am wondering if this is intended.

src/client/swagger/managedFunctions.yml
```diff
@@ -60,21 +60,21 @@ paths:
...
-        description: Function to run
+        description: Function to be triggered
         required: true
         content:
           application/json:
             schema:
               $ref: '#/components/schemas/FunctionTriggerRequest'
       responses:
-        '200':
+        '201':
```
wojciechka, line 77: Just a nitpick here.

I am not sure this should be a 201 as we are not creating anything. It would make sense that `

If we may want to make it 201 just to be consistent with other APIs, I am not strong on using 200 here, but 201 does not seem natural as nothing would get created in our CRUD APIs (as this is not a stored function, no logs or runs of that trigger would get stored anywhere).

It would make sense that POST to `/api/v2/functions/{functionID}/truns` would return 201 since that API creates a new run and this is also what tasks API returns when you manually run a task. But I think we've removed this API.

https://tools.ietf.org/html/rfc7231#section-6.3.2

src/client/swagger/managedFunctions.yml
```diff
@@ -167,57 +167,94 @@ paths:
...
-          description: The ID of the function to get runs for.
+        - in: query
+          name: params
+          schema:
+            type: object
+            additionalProperties: true
+          style: form
+          explode: true
       responses:
         '200':
```
wojciechka, line 190: I think this should be `default` instead remove the default below. We also don't know if the user won't return a non-error response with specific code - like 20x responses (like 201, 204) or 30x redirects.

src/client/swagger/managedFunctions.yml
```diff
@@ -167,57 +167,94 @@ paths:
...
-          description: A list of function runs
+          description: Function invocation response
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/FunctionRunRecords'
+                $ref: '#/components/schemas/FunctionHTTPResponseData'
```
wojciechka, line 195: Perhaps this should be `FunctionHTTPResponse` - as we're not sure if the user will return any data.

i.e. it may be totally fine to return 201 without data for a function that writes data to InfluxDB based on input provided to the function. Similarly, for 30x with redirects, it's totally fine to only return the Location header and no payload.

src/client/swagger/managedFunctions.yml
```diff
@@ -167,57 +167,94 @@ paths:
...
             application/json:
               schema:
-                $ref: '#/components/schemas/FunctionRun'
+                $ref: '#/components/schemas/FunctionHTTPResponseData'
+        default:
+          description: Unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
```
wojciechka, line 230: Same comments as above.

wojciechka (commented):
LGTM in general, aside from a few nitpicks

aanthony1243:
Hi @ebb-tide  we discussed this in slack last week.  IIRC we decide to move the `ground truth` version of this file to a different location (probably for now...IDPE?).  It's OK if we then copy that file over to here IFF they are building features that depend on it.

src/client/swagger/managedFunctions.yml
```diff
@@ -167,57 +167,94 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/api/v2private/functions/{functionID}/runs':
+  '/api/v2/functions/{functionID}/invoke':
     get:
```
yquansah, line 171: Should this be a `GET` request if the function is being invoked here?

yquansah: Also I am wondering if the API to trigger a function run should be `POST /api/v2/functions/{functionID}/runs`?Because a run would be the resource that is created in the backend.

ebb-tide:
Hi @aanthony1243, I completely agree that that these swagger definitions being here is awkward, but I'm actually already building a UI for managed-functions, so the ui repo already depends on this file. Let me know if you'd like me to also move it to idpe, but I don't think I can make the ui repo depend on idpe. 

> Hi @ebb-tide we discussed this in slack last week. IIRC we decide to move the `ground truth` version of this file to a different location (probably for now...IDPE?). It's OK if we then copy that file over to here IFF they are building features that depend on it.

src/client/swagger/managedFunctions.yml
ebb-tide: yes, I replaced POST to `/api/v2/functions/{functionID}/runs` with GET or POST to `/api/v2/functions/{functionID}/invoke` for now.

src/client/swagger/managedFunctions.yml
ebb-tide: That does make sense, thanks!

src/client/swagger/managedFunctions.yml
ebb-tide: Thanks for commenting @yquansah, you can invoke a function with a GET request and pass parameters as query parameters or with a POST, and provide the parameters in the request body. That's why I couldn't have it be just a POST to `/runs`. since `GET` to `/runs` is not available.

src/client/swagger/managedFunctions.yml
ebb-tide: fixed.

src/client/swagger/managedFunctions.yml
ebb-tide: fixed this

wojciechka (approved):

aanthony1243:
I think that the right approach will be for the official file to be hosted on IDPE, with the understanding that we will develop a way to publish it to the UI in the near future (probably a Q2 project).   In the meantime, we should keep the latest version of the swagger in IDPE and manually copy it to UI the way you have it here.   It should just be understood that this version should not be modified without synchronizing the changes to IDPE.  

> Hi @aanthony1243, I completely agree that that these swagger definitions being here is awkward, but I have actually already started building a UI for managed-functions, so the ui repo already depends on this file! We can keep a copy in idpe if that helps, but I don't think I can make the ui repo depend on idpe, let me know what you'd like me to do.
> 
> > Hi @ebb-tide we discussed this in slack last week. IIRC we decide to move the `ground truth` version of this file to a different location (probably for now...IDPE?). It's OK if we then copy that file over to here IFF they are building features that depend on it.

src/client/swagger/managedFunctions.yml
```diff
@@ -5,7 +5,7 @@ info:
 servers:
   - url: ''
 paths:
-  '/api/v2private/functions':
+  '/api/v2/functions':
```
yquansah, line 8: I think this route should be paginated. So the user can provide the limit and offset but those parameters should be optional

ebb-tide: good call thanks, I'll add it to /runs as well.

PR #742, opened 2021-02-23
test: make cloud tests pass in IDPE CI

genehynson:
Some tweaks to cloud e2e tests so they will pass in IDPE-CI

https://app.circleci.com/pipelines/github/influxdata/idpe?branch=genehynson%2Fui-e2e-cloud-tests

K8s-IDPE PR: https://github.com/influxdata/k8s-idpe/pull/2433

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cypress/e2e/cloud/buckets.test.ts
+++ b/cypress/e2e/cloud/buckets.test.ts
@@ -16,85 +16,105 @@ describe('Buckets', () => {
 
   // TODO: Skipping this until we can sort out the differences between OSS and Cloud
   it('can sort by name and retention', () => {
-    const demoDataBucket = 'Website Monitoring Bucket'
-    const buckets = [demoDataBucket, 'defbuck', '_tasks', '_monitoring']
-    const retentionDesc = ['_monitoring', demoDataBucket, '_tasks', 'defbuck']
-    const retentionAsc = ['defbuck', '_tasks', '_monitoring', demoDataBucket]
-
-    // if demo data bucket doesn't exist, create a bucket with the same name
-    cy.getByTestID('resource-list').then($body => {
-      if (
-        $body.find(`[data-testid="bucket-card ${demoDataBucket}"]`).length === 0
-      ) {
-        cy.getByTestID(`bucket-card ${demoDataBucket}`).should('not.exist')
-        cy.getByTestID('Create Bucket').click()
-        cy.getByTestID('overlay--container').within(() => {
-          cy.getByInputName('name').type(demoDataBucket)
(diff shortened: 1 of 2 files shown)


cypress/e2e/cloud/buckets.test.ts
```diff
@@ -16,24 +16,30 @@ describe('Buckets', () => {
...
-      'defbuck',
-    ]
-    const retentionAsc = [
-      'defbuck',
-      '_tasks',
-      '_monitoring',
-      'Website Monitoring Bucket',
-    ]
+    const demoDataBucket = 'Website Monitoring Bucket'
+    const buckets = [demoDataBucket, 'defbuck', '_tasks', '_monitoring']
```
joelfried, line 20: Do you think it's worth using the global alias for the defaultbucket rather than the hard coded defbuck string?

joelfried (approved):

subirjolly (approved):

drdelambre (changes requested):
please dont just remove all testing that exists for sample data

drdelambre:
longer version. this is the only test we have for an entire feature set. it comes up in bugs a lot when we write code that we _think_ is querying against and org, but because of this feature, we have to query against the org referenced by the bucket defined in the query text

idk if "test is broken, delete the test" is the right methodology here

drdelambre (approved):
misread, please ignore

cypress/e2e/cloud/geoOptions.test.ts
```diff
@@ -9,6 +9,11 @@ describe('DataExplorer - Geo Map Type Customization Options', () => {
         cy.fixture('routes').then(({orgs, explorer}) => {
           cy.visit(`${orgs}/${id}${explorer}`)
           cy.getByTestID('tree-nav')
+          cy.window().then(w => {
+            // I hate to add this, but the influx object isn't ready yet
```
drdelambre, line 13: this is crazy. it's loaded like.. first. you'd think the visit would be good enough, but i guess if it's only used behind the first setorg suspense and yadda yadda...

genehynson: yeah it's super annoying. I tried waiting for other elements on the page too but nothing would do the trick. I figured these flags will be removed soon enough and then we can delete this code.

cypress/e2e/cloud/buckets.test.ts
genehynson: Ooo yes

PR #746, opened 2021-02-24
feat: standalone query context

### LINKED ISSUE #543: Notebooks: "id length" error message more detail
Receiving `id must have length of 16 bytes` when running flux script. Not sure what it applies to (don't see id anywhere?) or what I need to do to fix
image

drdelambre:
The query context that exists in flows was way too coupled with the generateMap pipeline. Took some time this morning to decouple those concepts. This also consolidates all AST node manipulation for this query path to a single file in order to keep that craziness contained. There might be some more work to do there, but less than before.

Hopefully we can move the caching layer stuff into this context as well, to help detangle that bit from time machine and allow all interfaces to use it effectively.

This also brought up the fact that any variables will be set to the selection you made at the last dashboard you visited. There's no current plans on supporting that interface on a notebook yet, just a fun fact that might bite us later.

closes #543

### DIFF
--- a/src/flows/context/query.tsx
+++ b/src/flows/context/query.tsx
@@ -100,6 +100,7 @@ export const QueryProvider: FC = ({children}) => {
   const bucketsLoadingState = useSelector((state: AppState) =>
     getStatus(state, ResourceType.Buckets)
   )
+  const org = useSelector(getOrg)
 
   const dispatch = useDispatch()
 
@@ -136,8 +137,7 @@ export const QueryProvider: FC = ({children}) => {
         node?.arguments[0]?.properties[0]?.key?.name === 'bucket'
     ).map(node => node?.arguments[0]?.properties[0]?.value.value)
     const orgID =
-      buckets.find(buck => queryBuckets.includes(buck.name))?.orgID ||
-      useSelector(getOrg).id
+      buckets.find(buck => queryBuckets.includes(buck.name))?.orgID || org.id
 
     // This is all for trying to tease out a window period
     if (usedVars.hasOwnProperty('windowPeriod')) {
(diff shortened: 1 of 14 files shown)


src/flows/context/query.tsx
```diff
@@ -112,83 +110,133 @@ export const QueryProvider: FC = ({children}) => {
...
+              name: 'windowPeriod',
+            },
+            value: {
+              type: 'DurationLiteral',
+              values: [{magnitude: foundDuration.windowPeriod, unit: 'ms'}],
+            },
+          })
         } else {
-          append()
+          ;(((optionAST.body[0] as OptionStatement) // eslint-disable-line no-extra-semi
```
genehynson, line 214: what's the point of the `;` here?

src/flows/context/query.tsx
```diff
@@ -112,83 +110,133 @@ export const QueryProvider: FC = ({children}) => {
...
+    const ast = {
+      package: '',
+      type: 'Package',
+      files: [baseAST, optionAST],
+    }
 
-        const append = () => {
-          if (stages.length) {
-            stages[stages.length - 1].instances.push(pipeID)
+    // Here we grab the org from the contents of the query
```
genehynson, line 130: could some of this be pulled out into a couple functions to make this more readable? Like the stuff between lines 130 and 236.

src/flows/context/query.tsx
drdelambre: prettier adds it to.. escape execution? idk, the robots were fighting over it. one added it, one hated it and tried to remove it, i just told them to settle it down

src/flows/context/query.tsx
drdelambre: sure

drdelambre: hope that did it

src/flows/context/query.tsx
genehynson: huh how bizarre. Well if the robots say it's correct, then I'm okay with it!

genehynson (approved):
looks great!

PR #748, opened 2021-02-24
fix: show table column options

drdelambre:
table columns weren't showing up in the table vis options. now they are


genehynson (approved):
what a novel!

PR #751, opened 2021-02-26
fix: variable dependents function properly

ShmuelLotman:
Closes #18828

The variable hydration process included improper filtering of redux state and the "selected" value was being set to the default or first available option when selecting parent variables, making the dependent variable selection break. This PR introduces a potential fix. 

- [x] Rebased/mergeable
- [x] Tests pass


drdelambre (changes requested):
where's that test?

drdelambre (approved):
thanks for the test!

PR #753, opened 2021-03-01
feat: update Mosaic visualization

### LINKED ISSUE #510: Mosaic: y-axis column selections
**Proposal:**
Mosaic's y-axis has the same role as `fill` does for other graph types. Therefore, multiple columns should be allowed. Having multiple columns creates problems with spacing in the graph itself because the column names can be many and take up too much space.

Similar to how the tooltip can hide columns, Mosaic's y-axis should be able to hide columns. "Hide" is defined here as being _not_ shown as part of the label for each row in the y-axis. The hidden column should still be part of the group of columns that defines the y-axis, if the user wants to.

In other words, the y-axis has both a "group by" set of columns, and a "labeled" set of columns, where "labeled" must be in the domain of "group by". That is to say, "labeled" columns can only be chosen from the "group by" columns. By default, all "group by" columns are also in the "labeled" columns.

**Current behavior:**
Mosaic currently has an issue where different data sets can cause rendering errors because the y-axis is poorly definied. See https://github.com/influxdata/idpe/issues/8443. This issue is meant to replace https://github.com/influxdata/idpe/issues/8443 because the bug is caused by this missing feature. 

**Use case:**
Mosaic needs to correctly render all data sets rather than just a narrow set of very simple examples. With this feature, Mosaic can be complete.

**REQUIREMENTS:**
- [ ] y-axis can be defined as a group of multiple columns in View Options
- [ ] user's selection of the group of columns for the y-axis can be saved and persisted in cells of dashboards
- [ ] y-axis labels are defined by the columns in the above requirement "y-axis can be defined"
- [ ] y-axis labels can be shorter by hiding any columns by excluding them in the label
- [ ] y-axis labels must use least one column
- [ ] user's selection of the labels for the y-axis can be saved and persisted in cells of dashboards

TCL735:
Closes #510 

- [x] Rebased/mergeable
- [x] Tests pass

Mosaic updated with options to select the data columns, labels, and separator for the y-axis.
Screen Shot 2021-02-28 at 9 08 17 PM

### DIFF
--- a/src/shared/utils/vis.ts
+++ b/src/shared/utils/vis.ts
@@ -254,12 +254,27 @@ export const defaultYSeriesColumns = (
     )
   }
   if (ySeriesColumns.length === 0) {
-    ySeriesColumns.push(
-      validColumnKeys.find(columnKey => columnKey.startsWith('_') === false)
+    const defaultKey = validColumnKeys.find(
+      columnKey => columnKey.startsWith('_') === false
     )
+    if (defaultKey) {
+      ySeriesColumns.push(defaultKey)
+    }
   }
 
   return ySeriesColumns
+}
+
+export const defaultYLabelColumns = (
+  preferredYSeriesColumns: Array<string>,
+  validYSeriesColumns: Array<string>
+): Array<string> => {
+  return Array.isArray(preferredYSeriesColumns) &&
+    Array.isArray(validYSeriesColumns)
+    ? preferredYSeriesColumns.filter(columnKey =>
+        validYSeriesColumns.includes(columnKey)
+      )
+    : []
 }
 
 export const isInDomain = (value: number, domain: number[]) =>
--- a/src/shared/utils/vis.ts
+++ b/src/shared/utils/vis.ts
@@ -250,12 +250,7 @@ export const mosaicYColumn = (
     return preferredColumnKey
   }
 
-  const invalidMosaicYColumns = new Set([
-    '_value',
-    'status',
-    '_field',
-    '_measurement',
-  ])
+  const invalidMosaicYColumns = new Set(['_value', 'status', '_field'])
   const preferredValidColumnKeys = validColumnKeys.filter(
     name => !invalidMosaicYColumns.has(name)
   )
--- a/src/shared/utils/vis.ts
+++ b/src/shared/utils/vis.ts
@@ -241,27 +241,25 @@ export const defaultYColumn = (
   return null
 }
 
-export const mosaicYColumn = (
+export const defaultYSeriesColumns = (
   table: Table,
-  preferredColumnKey?: string
-): string | null => {
-  const validColumnKeys = getStringColumns(table)
-  if (validColumnKeys.includes(preferredColumnKey)) {
-    return preferredColumnKey
+  preferredYSeriesColumns: Array<string>
+): Array<string> => {
+  const validColumnKeys = [...getStringColumns(table)]
+  let ySeriesColumns = []
+
+  if (Array.isArray(preferredYSeriesColumns)) {
+    ySeriesColumns = preferredYSeriesColumns.filter(columnKey =>
+      validColumnKeys.includes(columnKey)
+    )
+  }
+  if (ySeriesColumns.length === 0) {
+    ySeriesColumns.push(
+      validColumnKeys.find(columnKey => columnKey.startsWith('_') === false)
+    )
   }
 
-  const invalidMosaicYColumns = new Set(['_value', 'status', '_field'])
-  const preferredValidColumnKeys = validColumnKeys.filter(
-    name => !invalidMosaicYColumns.has(name)
-  )
-  if (preferredValidColumnKeys.length) {
-    return preferredValidColumnKeys[0]
-  }
-
-  if (validColumnKeys.length) {
-    return validColumnKeys[0]
-  }
-  return null
+  return ySeriesColumns
 }
 
 export const isInDomain = (value: number, domain: number[]) =>
--- a/src/timeMachine/components/FunctionModeSelector.tsx
+++ b/src/timeMachine/components/FunctionModeSelector.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FunctionComponent, useState} from 'react'
+import React, {FunctionComponent, useEffect, useState} from 'react'
 import {connect, ConnectedProps} from 'react-redux'
 
 // Components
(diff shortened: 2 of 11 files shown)


src/shared/utils/vis.ts
```diff
@@ -241,32 +241,40 @@ export const defaultYColumn = (
...
-): string | null => {
-  const validColumnKeys = getStringColumns(table)
-  if (validColumnKeys.includes(preferredColumnKey)) {
-    return preferredColumnKey
+  preferredYSeriesColumns: Array<string>
+): Array<string> => {
+  const validColumnKeys = [...getStringColumns(table)]
+  let ySeriesColumns = []
+
+  if (Array.isArray(preferredYSeriesColumns)) {
```
hoorayimhelping, line 251: if `preferredYSeriesColumns` is required, is this check necessary?

src/timeMachine/components/FunctionModeSelector.tsx
```diff
@@ -34,20 +41,35 @@ type ReduxProps = ConnectedProps<typeof connector>
...
   onMultiSelectBuilderFunction,
+  onSetFunctions,
   onSetIsAutoFunction,
-  isInCheckOverlay,
+  onSingleSelectBuilderFunction,
+  savedFunctions,
 }) => {
-  const autoFunctions = AUTO_FUNCTIONS.map(f => f.name)
+  const autoFunctions = getDefaultAutoFunctions(activeGraphType).map(
+    f => f.name
```
hoorayimhelping, line 53: i know you you didn't write this, but you would mind maybe just updating `f` to something easier to search for, like `autoFunction`, please?

src/timeMachine/components/FunctionModeSelector.tsx
```diff
@@ -62,9 +84,7 @@ const FunctionSelector: FunctionComponent<Props> = ({
       handleSetAutoFunction(false)
       return
     }
-    const newFunctions = selectedFunctions.filter(f =>
-      autoFunctions.includes(f)
-    )
+    const newFunctions = savedFunctions.filter(f => autoFunctions.includes(f))
```
hoorayimhelping, line 87: same here about renaming `f` to `savedFunction` or something like that please?

src/visualization/types/Mosaic/options.tsx
```diff
@@ -37,18 +43,12 @@ const MosaicOptions: FC<Props> = ({properties, results, update}) => {
...
-      return false
-    }
-
-    const columnType = results.table.getColumnType(key)
-
-    return columnType === 'time' || columnType === 'number'
-  })
+  // Mosaic graphs are currently limited to always using _time on the x-axis
+  //   future enhancements will depend on the visualization library
+  const xDataColumn = results.table.columnKeys.filter(key => key === '_time')
```
hoorayimhelping, line 48: is this deep chain of components for sure to exist or do we need to perform some null checks?

src/visualization/types/Mosaic/options.tsx
```diff
@@ -37,18 +43,12 @@ const MosaicOptions: FC<Props> = ({properties, results, update}) => {
...
 
   if (
-    properties.fillColumns &&
+    Array.isArray(properties.fillColumns) &&
```
hoorayimhelping, line 51: 👍 nice!

src/visualization/types/Mosaic/options.tsx
```diff
@@ -62,13 +62,70 @@ const MosaicOptions: FC<Props> = ({properties, results, update}) => {
...
+    currentYLabelColumns: Array<string>,
+    option: string
+  ): Array<string> => {
+    const columnExists = Array.isArray(currentYLabelColumns)
+      ? currentYLabelColumns.find(col => col === option)
+      : false
+    let updatedYLabelColumns = currentYLabelColumns || []
+
+    if (columnExists) {
+      updatedYLabelColumns = currentYLabelColumns.filter(fc => fc !== option)
```
hoorayimhelping, line 91: what is `fc`?

src/visualization/types/Mosaic/options.tsx
```diff
@@ -97,32 +154,25 @@ const MosaicOptions: FC<Props> = ({properties, results, update}) => {
...
-              options={numericColumns}
+              options={xDataColumn}
               selectedOption={xColumn || 'Build a query before selecting...'}
               onSelect={xColumn => {
                 update({xColumn})
               }}
               testID="dropdown-x"
               buttonStatus={
-                numericColumns.length == 0
+                xDataColumn.length == 0
```
hoorayimhelping, line 164: `.length` will always return a number, so we can do a strict check here `===` since `==` at the component level often implies that number and string values are going to be interchanged

src/visualization/types/Mosaic/view.tsx
```diff
@@ -64,14 +59,13 @@ const MosaicPlot: FunctionComponent<Props> = ({
...
-    result.table.getColumn(yColumn, 'string')
+    result.table.getColumn(ySeriesColumns[0], 'string')
   )
 
   const isValidView =
     xColumn &&
     columnKeys.includes(xColumn) &&
-    yColumn &&
-    columnKeys.includes(yColumn) &&
+    ySeriesColumns.every(col => columnKeys.includes(col)) &&
```
hoorayimhelping, line 68:
```suggestion
    ySeriesColumns.every(yCol => columnKeys.includes(yCol)) &&
```

just provides a little extra context when we're dealing with y column and x column

hoorayimhelping (approved):
looks good, just a couple of comments / questions

src/shared/utils/vis.ts
TCL735: Yes, because even TypeScript allows null

src/timeMachine/components/FunctionModeSelector.tsx
TCL735: Done.

src/timeMachine/components/FunctionModeSelector.tsx
TCL735: Done.

src/visualization/types/Mosaic/options.tsx
TCL735: It does exist. But doesn't hurt to add the optional chaining syntax. The codebase is not consistent in this regard. Some graph types use the optional chaining syntax while others do not.

I will use it here and we can address the other graph types separately.

src/visualization/types/Mosaic/options.tsx
TCL735: i think originally copied/pasted from similar code for `fillColumns`.

Will change this to `currentYLabelColumn`

src/visualization/types/Mosaic/options.tsx
TCL735: Done.

src/visualization/types/Mosaic/view.tsx
TCL735: Good point. I changed it to the singluar. It's more verbose as `ySeriesColumn` but also makes clear that Mosaic uses a special kind of yColumn and not the yColumn used in other visualizations.

PR #755, opened 2021-03-01
fix: remove table threshold crash

drdelambre:
got some copy pasta in the keys. crazy that this worked at all.


genehynson (approved):

hoorayimhelping (approved):
but [this reference](https://github.com/influxdata/ui/blob/master/src/visualization/components/internal/ThresholdsSettings.tsx#L4) should remain there, right?

PR #762, opened 2021-03-02
fix: allow notebooks to work with multiple tabs

### LINKED ISSUE #626: notebooks don't support multiple tabs
If you open two tabs of the UI and navigate to different notebooks in each of them you'll notice a very strange flashing effect. This is happening because each of the tabs are fighting over local storage and keep overwriting the currently saved notebook id.

https://user-images.githubusercontent.com/6411855/107101992-e4d43080-67cd-11eb-80f3-4a73df549bd5.mov

genehynson:
Closes #626 

Allow notebooks to work across multiple tabs. This includes having the _same_ notebook open in multiple tabs or having _different_ notebooks open in multiple tabs.

+ removes storage of currentID in localstorage so multiple tabs can independently load a unique notebook.
+ removes docVersion from FluxMonocoEditor to avoid unnecessary state changes and causing typed characters to be lost.
+ replaces `use-persisted-state` with [use-local-storage-state](https://github.com/astoilkov/use-local-storage-state) to fix crazy issue where if you updated the state too quickly with two tabs open the state would go into a recursive death. Also use-persisted-state is [not maintained anymore](https://github.com/donavon/use-persisted-state/issues/36). 
+ adds `useEvent` hook for safely adding a window event listener to components. Uses this to monitor for local storage events that signal that "queryAll" was executed in another tab. If this event is triggered, the other tabs will also execute queryAll.

Same Notebook:

https://user-images.githubusercontent.com/6411855/110043905-17097d00-7cfd-11eb-9a41-9e0af2d96761.mov

Different Notebooks:

https://user-images.githubusercontent.com/6411855/109703652-90ad3980-7b4a-11eb-9238-23461dc2c7eb.mov

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/flows/context/flow.list.tsx
+++ b/src/flows/context/flow.list.tsx
@@ -1,4 +1,4 @@
-import React, {FC, useCallback, useEffect} from 'react'
+import React, {FC, useCallback, useEffect, useState} from 'react'
 import {useDispatch} from 'react-redux'
 import {useParams} from 'react-router-dom'
 import createPersistedState from 'use-persisted-state'
@@ -37,7 +37,6 @@ import {
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 const useFlowListState = createPersistedState('flows')
-const useFlowCurrentState = createPersistedState('current-flow')
 
 export interface FlowListContextType extends FlowList {
   add: (flow?: Flow) => Promise<string>
@@ -135,7 +134,7 @@ export function hydrate(data) {
 
 export const FlowListProvider: FC = ({children}) => {
   const [flows, setFlows] = useFlowListState(DEFAULT_CONTEXT.flows)
-  const [currentID, setCurrentID] = useFlowCurrentState(null)
+  const [currentID, setCurrentID] = useState(DEFAULT_CONTEXT.currentID)
   const {orgID} = useParams<{orgID: string}>()
   const dispatch = useDispatch()
   useEffect(() => {
--- a/src/flows/context/flow.list.tsx
+++ b/src/flows/context/flow.list.tsx
@@ -1,7 +1,7 @@
 import React, {FC, useCallback, useEffect, useState} from 'react'
 import {useDispatch} from 'react-redux'
 import {useParams} from 'react-router-dom'
-import createPersistedState from 'use-persisted-state'
+import useLocalStorageState from 'use-local-storage-state'
 import {v4 as UUID} from 'uuid'
 import {
   FlowList,
@@ -35,8 +35,6 @@ import {
   notebookDeleteFail,
 } from 'src/shared/copy/notifications'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
-
-const useFlowListState = createPersistedState('flows')
 
 export interface FlowListContextType extends FlowList {
   add: (flow?: Flow) => Promise<string>
@@ -133,7 +131,7 @@ export function hydrate(data) {
 }
 
 export const FlowListProvider: FC = ({children}) => {
-  const [flows, setFlows] = useFlowListState(DEFAULT_CONTEXT.flows)
+  const [flows, setFlows] = useLocalStorageState('flows', DEFAULT_CONTEXT.flows)
   const [currentID, setCurrentID] = useState(DEFAULT_CONTEXT.currentID)
   const {orgID} = useParams<{orgID: string}>()
   const dispatch = useDispatch()
--- a/src/flows/context/flow.list.tsx
+++ b/src/flows/context/flow.list.tsx
@@ -253,9 +253,10 @@ export const FlowListProvider: FC = ({children}) => {
       name: flow.name,
       range: flow.range,
       refresh: flow.refresh,
-      data: flow.data.serialize ? flow.data.serialize() : flow.data,
-      meta: flow.meta.serialize ? flow.meta.serialize() : flow.meta,
+      data: flow.data,
+      meta: flow.meta,
       readOnly: flow.readOnly,
+      results: null,
     }
 
     setFlows({
--- a/src/flows/pipes/Markdown/MarkdownPanel.tsx
+++ b/src/flows/pipes/Markdown/MarkdownPanel.tsx
@@ -6,7 +6,6 @@ import {
   TechnoSpinner,
 } from '@influxdata/clockface'
 import classnames from 'classnames'
-import {debounce} from 'lodash'
 
 // Types
 import {MarkdownMode} from './'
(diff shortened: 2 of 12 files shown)


src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -46,7 +46,6 @@ const FluxEditorMonaco: FC<Props> = ({
 }) => {
   const lspServer = useRef<LSPServer>(null)
   const [editorInst, seteditorInst] = useState<EditorType | null>(null)
-  const [docVersion, setdocVersion] = useState(2)
```
genehynson, line 49: I removed this because it doesn't seem to be used anywhere other than the increment inside the `onChange`. The increment inside the `onChange` was causing some weird behavior where only the last character typed was recorded once I introduced the `debouce`.

drdelambre: i think this is needed by the LSP for some reason. reach out to @scbrickley to make sure i'm not just blowing bubbles

src/flows/pipes/Markdown/MarkdownPanel.tsx
```diff
@@ -6,6 +6,7 @@ import {
   TechnoSpinner,
 } from '@influxdata/clockface'
 import classnames from 'classnames'
+import {debounce} from 'lodash'
```
drdelambre, line 9: ewwwwwwwwww lodash what the heck mang

src/flows/pipes/Markdown/MarkdownPanel.tsx
```diff
@@ -63,9 +64,7 @@ const MarkdownPanel: FC<PipeProp> = ({Context}) => {
     <MarkdownModeToggle mode={data.mode} onToggleMode={handleToggleMode} />
   )
 
-  const handleChange = (text: string): void => {
-    update({text})
-  }
+  const handleChange = debounce((text: string): void => update({text}), 500)
```
drdelambre, line 67: why does this help? shouldn't we have a clear functional path from the input event, through the local storage event cycle, and back up through the rendering pipeline to display the value through props?

I'm worried that this is brushing any thing to the contrary under the rug, and is only going to make it harder to debug when these updates are being funneled through the network interface from a bunch of users

src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -39,18 +40,20 @@ const Query: FC<PipeProp> = ({Context}) => {
...
+  const updateText = text => {
+    const _queries = queries.slice()
+    _queries[activeQuery] = {
+      ...queries[activeQuery],
+      text,
+    }
+
+    update({queries: _queries})
+  }
+  const updateTextDebounce = useCallback(debounce(updateText, 500), [
```
drdelambre, line 52: same concern here. Last time we had issues with updating the flux editor between tabs (either input didn't work when a tab was open, or it started forgetting charaters), it was because of a sub component using stale state when updating, which it would do on rerenders, all red flags that needed to be routed out

drdelambre (changes requested):
not gonna argue that this may have mitigated the issue for the immediate need, but my past experience in this code is gonna make me throw some heavy air quotes on "fixed" if that debounced is really required.

lemme know why i'm off base, and good call on taking currentID out of local storage

src/flows/pipes/Markdown/MarkdownPanel.tsx
genehynson: Debounce saves from unnecessary re-renders! 

Without debounce:

https://user-images.githubusercontent.com/6411855/109856944-f105af00-7c0e-11eb-93ff-14894a047948.mov

With debounce:

https://user-images.githubusercontent.com/6411855/109856928-ec40fb00-7c0e-11eb-90de-7a9bfd44d318.mov

genehynson: I will look further into why we have this weird behavior with multiple tabs without debounce, but I think regardless debounce is a worthy performance improvement that we should use

src/flows/pipes/Markdown/MarkdownPanel.tsx
genehynson: I mean...ok, I can implement it myself but we do already have the package installed so I figured what's the harm 🤷

src/shared/components/FluxMonacoEditor.tsx
genehynson: Slack with Sean: https://influxdata.slack.com/archives/C0AD7FFS9/p1614795598154600

We're keeping track of the document version in the monoco server so we don't need to additionally track it in the flux editor component. 
https://github.com/influxdata/ui/blob/master/src/external/monaco.flux.server.ts#L278

src/flows/pipes/Markdown/MarkdownPanel.tsx
drdelambre: lemme know what you find. still leaning more towards that "hides rendering issues" than "fixes rendering issues", but totally open to new info

src/flows/pipes/RawFluxEditor/view.tsx
genehynson: For reference, here's the issue without debounce. It looks like the text is persisted only once (i.e. the "updating" logs stop when I stop tying) but the re-rendering goes on forever

https://user-images.githubusercontent.com/6411855/109862412-404ede00-7c15-11eb-8ab5-2aaef5f185db.mov

src/flows/pipes/Markdown/MarkdownPanel.tsx
genehynson: debounce removed! It's unneeded to fix this issue

src/flows/pipes/Markdown/MarkdownPanel.tsx
genehynson: for those reading this in a not-so-distant-future: the issue was with `use-persisted-state`. Switching libraries fixed the issue and removed the need for debounce.

src/flows/pipes/RawFluxEditor/view.tsx
genehynson: Same as above - this issue is fixed by switching libraries.

src/flows/context/flow.list.tsx
```diff
@@ -256,9 +253,10 @@ export const FlowListProvider: FC = ({children}) => {
       name: flow.name,
       range: flow.range,
       refresh: flow.refresh,
-      data: flow.data.serialize ? flow.data.serialize() : flow.data,
-      meta: flow.meta.serialize ? flow.meta.serialize() : flow.meta,
+      data: flow.data,
+      meta: flow.meta,
```
genehynson, line 257: FYI the new library seems to have solved this issue since it is strongly typed https://github.com/influxdata/ui/pull/567#discussion_r566996414

drdelambre (approved):
totally forgot about that query thing. excited to see _every, single, button_ working on the new library

PR #766, opened 2021-03-03
fix: general bug fixing

### LINKED ISSUE #610: Right click menu for the Flux editor has disappeared in Cloud
**Expected behavior:**
When right clicking on the flux editor, this menu should pop up where I clicked:

rightclick

**Actual behavior:**
The default right click menu for my browser shows up instead

**Environment info:**

Only appears to be a problem in cloud. OSS is unaffected.

### LINKED ISSUE #692: Graph + Single Stat has weird options and inconsistent header labels
**Graph + Single Stat:** X-axis tick configuration does isn't consistent with the rest. Isn't under it's own heading but under options
image

~~**Scatterplot:** has X/Y Tick Prefix/Suffix and no other chart type has it. Is this supposed to be "X Axis Prefix" not Tick?~~
Will remove this "X Prefix" overall - shouldn't exist since X-Axis in these charts is always time

### LINKED ISSUE #713: Remove "X Tick Prefix/Suffix" from Scatterplot & Heatmap
Remove "X Tick Prefix" and "X Tick Suffix" from charts.

Only Scatterplot & Heatmap Charts have "X Tick Prefix" and "X Tick Suffix". We actually don't need these considering the X-axis will always be time (for these charts). They currently don't have any functionality as well. 

image

### LINKED ISSUE #752: "all buckets" token doesn't excludes system buckets "_monitoring" and "_tasks"
Using OSS 2.0.4.
When creating a token, if you scope it to all buckets, the token excludes the system buckets "_monitoring" and "_tasks". 
If you want to scope a token to specific buckets, you can't select "_monitoring" and "_tasks" because they don't exist in the menu. 

Screen Shot 2021-02-26 at 1 25 11 PM

Thank you. :)

drdelambre:
just some quick ones

Closes #752 
Closes #713 
Closes #692
Closes #610

### DIFF
--- a/src/shared/components/FluxMonacoEditor.tsx
+++ b/src/shared/components/FluxMonacoEditor.tsx
@@ -17,7 +17,7 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 // Types
 import {OnChangeScript} from 'src/types/flux'
 import {EditorType} from 'src/types'
-import {editor as monacoEditor} from 'monaco-editor/esm/vs/editor/editor.api.js'
+import {editor as monacoEditor} from 'monaco-editor'
 
 import './FluxMonacoEditor.scss'
 import {Diagnostic} from 'monaco-languageclient/lib/services'
--- a/src/authorizations/components/BucketsTokenOverlay.tsx
+++ b/src/authorizations/components/BucketsTokenOverlay.tsx
@@ -80,6 +80,10 @@ class BucketsTokenOverlay extends PureComponent<Props, State> {
       activeTabWrite,
     } = this.state
 
+    const {
+        buckets
+    } = this.props
+
     return (
       <Overlay.Container maxWidth={700}>
         <Overlay.Header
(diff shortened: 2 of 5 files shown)


src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -17,7 +17,7 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 // Types
 import {OnChangeScript} from 'src/types/flux'
 import {EditorType} from 'src/types'
-import {editor as monacoEditor} from 'monaco-editor/esm/vs/editor/editor.api.js'
+import {editor as monacoEditor} from 'monaco-editor'
```
genehynson, line 20: This was the fix for the right click menu??? wat

genehynson (approved):
love the rapid fire bug fixes!

src/shared/components/FluxMonacoEditor.tsx
drdelambre: right?!

joelfried (approved):

src/shared/components/FluxMonacoEditor.tsx
ShmuelLotman: Holy crud i spent some time looking into this and never saw this. TIL.

PR #772, opened 2021-03-04
fix: respect 24 hour clock formats and allow more choices

### LINKED ISSUE #768: What time is it?
when we have a graph that spans over a large period of time, and the dashboard / time machine is set to displaying in local time, the meridian isn't included when generating axis ticks in locales that favor 12hr time periods (like the US). so i dont know if this graph is showing 4am or 4pm, and that's kind of important info to know

Screen Shot 2021-03-03 at 11 51 05 AM

TCL735:
Closes #768 

This fix is the first step in updating time formats. The next step is to implement https://github.com/influxdata/ui/issues/770

- 'HH' triggers 24 hour clock format
- Add more choices to allow users to keep 12 hour formats
- Add meridiem to 12 hour formats

Kapture 2021-03-03 at 16 11 29

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/visualization/utils/getFormatter.ts
+++ b/src/visualization/utils/getFormatter.ts
@@ -60,10 +60,14 @@ export const getFormatter = (
   }
 
   if (columnType === 'time') {
-    return timeFormatter({
+    const formatOptions = {
       timeZone: timeZone === 'Local' ? undefined : timeZone,
       format: resolveTimeFormat(timeFormat),
-    })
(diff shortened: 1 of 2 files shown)


drdelambre (approved):
i think russ might have an opinion about the ordering and the like, but the code looks fine

russorat (approved):
LGTM

PR #773, opened 2021-03-04
feat: annotation write mode toggle

### LINKED ISSUE #725: Annotations Tech Debt: Clean up flagged check in XYPlot
### Background
There is some tech debt around how annotations are rendered in the XYPlot view layer. We'll want to shore these up and come up with a good pattern for future use in future graph types.

### LINKED ISSUE #730: Annotations: Add Annotations button
### Background
I removed the Add annotation button [here](https://github.com/influxdata/ui/commit/82390639ad3be8d8a4340993890a340e3f543318#diff-2247896cc9fd441f0c090919563197e9050097e94918552cc86f47d98b143ff5L81) because it wasn't really hooked up to a great experience. We want this button to toggle creating annotations. If this button isn't on, clicking on the graph won't create annotations. If it is on, clicking (single-clicking) on the graph will start the annotation creation flow (i.e. it'll launch an overlay)

### Acceptance Criteria
- The `Annotate` button is added back to the `AnnotationsControlBar` and is functional
-  clicking (single click) the `Annotate` button toggles annotate mode, which lets the user annotate the graph by double clicking
  - clicking (single click) on the graph when not in annotate should do the default behavior (which is currently nothing)

- this is for the dashboard. The new button should be in the well that opens when you click on the annotations button

jrenee42:
Closes #730
Closes #725 

adds a toggle/checkbox button to the annotations bar, to enable 1-click annotations (double click still works too; since a double click has a single click in it)

annotations can only be written when the toggle is enabled.

### DIFF
--- a/src/annotations/components/controlBar/AnnotationsControlBar.tsx
+++ b/src/annotations/components/controlBar/AnnotationsControlBar.tsx
@@ -1,6 +1,6 @@
 // Libraries
 import React, {useState, FC} from 'react'
-import {useSelector} from 'react-redux'
+import {useDispatch, useSelector} from 'react-redux'
 import {useHistory, useParams} from 'react-router-dom'
 
 // Components
@@ -18,11 +18,15 @@ import {
 } from '@influxdata/clockface'
 import ErrorBoundary from 'src/shared/components/ErrorBoundary'
 import Toolbar from 'src/shared/components/toolbar/Toolbar'
+import {toggleSingleClickAnnotations} from 'src/annotations/actions/creators'
 import {AnnotationPills} from 'src/annotations/components/controlBar/AnnotationPills'
 import {AnnotationsSearchBar} from 'src/annotations/components/controlBar/AnnotationsSearchBar'
 
 // Selectors
-import {getAnnotationControlsVisibility} from 'src/annotations/selectors'
+import {
+  getAnnotationControlsVisibility,
+  isSingleClickAnnotationsEnabled,
+} from 'src/annotations/selectors'
 
 // Constants
 import {ORGS, SETTINGS, ANNOTATIONS} from 'src/shared/constants/routes'
@@ -32,14 +36,19 @@ export const AnnotationsControlBar: FC = () => {
   const isVisible = useSelector(getAnnotationControlsVisibility)
   const {orgID} = useParams<{orgID: string; dashboardID: string}>()
 
-  const [annotationWriteMode, setAnnotationWriteMode] = useState(false)
+  const [annotationWriteMode, setAnnotationWriteMode] = useState(
+    useSelector(isSingleClickAnnotationsEnabled)
+  )
 
   if (!isVisible) {
     return null
   }
+  const dispatch = useDispatch()
 
   const onModeChange = () => {
     //todo:  call prop method to change the state
+    dispatch(toggleSingleClickAnnotations())
+
     setAnnotationWriteMode(!annotationWriteMode)
   }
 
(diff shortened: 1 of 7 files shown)


src/annotations/components/controlBar/AnnotationsControlBar.tsx
```diff
@@ -1,35 +1,48 @@
...
   const {orgID} = useParams<{orgID: string; dashboardID: string}>()
 
-  if (!isVisible) {
-    return null
+  const inWriteMode = useSelector(isSingleClickAnnotationsEnabled)
+  const [annotationWriteMode, setAnnotationWriteMode] = useState(inWriteMode)
+
+  const dispatch = useDispatch()
+
+  const onModeChange = () => {
```
hoorayimhelping, line 43: naming nit: this is a function that makes a change to state. the fact that it happens to be used as a handler shouldn't affect the name of it. i think it's more powerful to name the function for what it does, rather than how it might be used.

something like `changeMode` vs `onModeChange`. 

then in the component, you'd have something like

`onChange={changeMode}`

src/dashboards/components/DashboardPage.tsx
```diff
@@ -145,9 +150,12 @@ const mstp = (state: AppState) => {
     state.currentDashboard.id
   )
 
+  const showAnnotationBar = state.userSettings.showAnnotationsControls || false
```
hoorayimhelping, line 153: ```state.userSettings.showAnnotationsControls || false```

i like where your head's at here, but i think this is one of those situations that calls for failing fast, because if this doesn't have a default value, it means our state isn't set up correctly. in redux, you give default values to every part of your state when you create it. that way, in component land, we can safely assume that this thing will have a default value. i appreciate you checking it, that's good practice, but that check should already be happening as part of the `userSettings` reducer `initialState` method.

src/visualization/types/Graph/view.tsx
```diff
@@ -195,41 +236,12 @@ const XYPlot: FC<Props> = ({properties, result, timeRange, annotations}) => {
...
-          dismissOverlay
-        )
-      )
-    }
-
-    config.interactionHandlers = {
-      doubleClick: doubleClickHandler,
+    if (inAnnotationWriteMode) {
+      config.interactionHandlers = {
+        singleClick: makeSingleClickHandler(),
```
hoorayimhelping, line 242: factory function, fair enough :+1:

hoorayimhelping (changes requested):
great work! love the variable naming. couple of small things more related to how you're supposed to things with redux. overall, solid work

src/annotations/components/controlBar/AnnotationsControlBar.tsx
```diff
@@ -1,35 +1,48 @@
...
 
 export const AnnotationsControlBar: FC = () => {
   const history = useHistory()
-  const isVisible = useSelector(getAnnotationControlsVisibility)
   const {orgID} = useParams<{orgID: string; dashboardID: string}>()
 
-  if (!isVisible) {
-    return null
+  const inWriteMode = useSelector(isSingleClickAnnotationsEnabled)
+  const [annotationWriteMode, setAnnotationWriteMode] = useState(inWriteMode)
```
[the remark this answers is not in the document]
jrenee42: i thought of that; but then there might be a delay....i will test it.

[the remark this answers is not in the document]
jrenee42: no delay.  so just using redux now.

hoorayimhelping (approved):

PR #788, opened 2021-03-05
fix: show an error message when queries with broken variables fail to hydrate

hoorayimhelping:
Closes idpe #9772

Dashboard cells with queries with broken variables in them will spin forever. This shows the error message the broken query throws.

Screen Shot 2021-03-04 at 3 01 08 PM

### Reproduction steps:
1. Create a variable with a broken query
   1. settings -> variables -> create
   1. put garbage into the query
1. Go to explore, write a query that uses that a broken variable.
    Assuming you named the variable `broken_variable`:
    ```
      from(bucket: "defbuck")
        |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
        |> filter(fn: (r) => r["_measurement"] == "system")
        |> filter(fn: (r) => r["_measurement"] == v.broken_variable)
        |> aggregateWindow(every: v.windowPeriod, fn: mean, createEmpty: false)
        |> yield(name: "mean")
    ```
1. Save the query and reload the page. It'll spin forever.
2. Apply this branch. Reload the page. There's an error message.

---

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/shared/components/TimeSeries.tsx
+++ b/src/shared/components/TimeSeries.tsx
@@ -226,8 +226,23 @@ class TimeSeries extends Component<Props, State> {
           cancel()
         }
       })
-      const usedVars = variables.filter(v => v.arguments.type !== 'system')
-      const waitList = usedVars.filter(v => v.status !== RemoteDataState.Done)
+      const usedVars = variables.filter(
+        variable => variable.arguments.type !== 'system'
+      )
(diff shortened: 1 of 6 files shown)


hoorayimhelping:
there are more conditions that need to be dealt with before this is ready to go.

PR #791, opened 2021-03-08
feat: default streams are now visible in Annotation bar on page load

### LINKED ISSUE #729: Annotations: in the AnnotationsControlBar, show the active streams being displayed
### Background
Clicking the `Annotations` button (in the top, next to Variables) toggles the `AnnotationsControlBar`. It's currently blank. It should instead display the currently active streams. Only the default stream should be displayed by default.

https://www.figma.com/file/AQk0tUthLFepRhFUibvzMI/Annotations?node-id=134%3A890

#### Current behavior:
Screen Shot 2021-02-18 at 4 47 25 PM

#### Default Behavior should display active streams:
Screen Shot 2021-02-18 at 4 47 40 PM

### Acceptance Criteria
- The active streams are displayed in the annotations control bar
- Only the 'default' stream is shown on page load, toggling other streams will be up to the user

ChitlangeSahas:
Closes #729 

This PR enables the AnnotationPills to display the default stream on page load.

- [x] Rebased/mergeable
- [x] Tests pass


src/annotations/constants/mocks.ts
```diff
@@ -25,6 +25,29 @@ export interface AnnotationStream {
...
+      timeStopCol: '_timeStop',
+    },
+    query: {
+      tags: {
+        _field: 'deployments',
+      },
+      measurement: 'annotations',
+      bucketName: 'defBuck',
+    },
+  },
```
hoorayimhelping, line 50: might as well remove the other annotation mocks since they're not used and they're irrelevant and confusing

hoorayimhelping (changes requested):
Looks good so far, but when you close out the annotation pill, it should stop showing that stream's annotations. So `default` should show on page load. If I want to hide those annotations, I click the `x` in the pill and they should all disappear.

src/visualization/types/Graph/view.tsx
```diff
@@ -243,7 +248,12 @@ const XYPlot: FC<Props> = ({properties, result, timeRange, annotations}) => {
       }
     }
     // everything is under the 'default' category for now:
-    const selectedAnnotations: any[] = annotations?.default ?? []
+    let selectedAnnotations: any[] = []
+
+    visibleAnnotationStreams.map(stream => {
+      selectedAnnotations = selectedAnnotations.concat(annotations[stream.id])
+    })
```
hoorayimhelping, line 255: `map` will return a new array with each item described by the function, you shouldn't perform mutative operations in them (like modifying a variable that is external to the function).

[`filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) will return an array of items that pass the test described by the function, or an empty array. I think that's what you want here. Something like:

```js
const selectedAnnotations: any[] = visibleAnnotationStreams.filter(stream => {
  if (annotations[stream.id]) {
    return annotations[stream.id]
  }
})

hoorayimhelping (changes requested):

PR #800, opened 2021-03-09
fix: escape the text from tag selectors before using as RegEx

### LINKED ISSUE #798: Data Explorer: Tag filters don't work with the forward slash character
to repro:

1. open DE, select bucket
2. in the _measurements box, type ` some/thing` in the box
3. notice that the request fails because it is an invalid regex

`message: "compilation failed: error at @7:36-7:46: regex parse error: (?i:some error: unclosed group"`

image

TCL735:
Closes #798 

- [x] Rebased/mergeable
- [x] Tests pass

Kapture 2021-03-08 at 18 00 05


drdelambre (approved):
yeah.. idk what else to do with this based on the way the query is formatted

PR #807, opened 2021-03-10
feat: new raw table!

### LINKED ISSUE #642: Data Table: I can view results simplified to my metric selector filters
As a InfluxDB Cloud user, I can s view the header annotation of the results so that I can understand the shape the data being returned.

Note: We heard in user research it was hard to see the actual metric I selected because of unnecessary fields in the Flux results that weren't sorted to my filter.

Given: I have a raw data view open and results shown
When: I see results in a formatted table:

**Headers**
The table header contains a title and two meta data items:
- column `group` annotation
- column `datatype` 

Remove:
`#default` annotation is not shown

**Row Order**
1. Measurement
2. Field
3. Value
4. Tag
5. _Time
6. `_Table` 

Remove 
- Remove _`start` and `_stop` columns

### Technical Notes
  - Double clicking on the text in any cell should highlight the contents of that cell

#### Design Notes
- Clockface Component
  - https://influxdata.github.io/clockface/?path=/story/components-table-family--table
- Striped is enabled

- Mockup
  - https://www.figma.com/file/8S8IH8mimYIwanZ6keIQ6U/Notebooks-Prototype?node-id=1635%3A10213

drdelambre:
Closes #642

### DIFF
--- a/src/flows/components/panel/Results.tsx
+++ b/src/flows/components/panel/Results.tsx
@@ -1,7 +1,7 @@
 // Libraries
 import React, {FC, useEffect, useState, useContext, useMemo} from 'react'
-import {AutoSizer} from 'react-virtualized'
 import {Table, ComponentSize, DapperScrollbars} from '@influxdata/clockface'
+import {FluxResult} from 'src/types/flows'
 
 // Components
 import Resizer from 'src/flows/shared/Resizer'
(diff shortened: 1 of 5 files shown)


src/flows/components/panel/Results.tsx
```diff
@@ -19,6 +18,165 @@ import {event} from 'src/cloud/utils/reporting'
...
+}
+
+const ResultTable: FC<TableProps> = ({
+  startRow,
+  results,
+  height,
+  pageSize,
+  setPageSize,
+}) => {
+  const kids = useMemo(() => {
```
genehynson, line 39: lol kids xD

src/flows/components/panel/Results.tsx
```diff
@@ -19,6 +18,165 @@ import {event} from 'src/cloud/utils/reporting'
...
+    if (!height) {
+      return
+    }
+
+    let runningHeight = 60 // this is to account for the space around the table
+    let rowIdx = startRow
+    let currentTable
+
+    while (startRow <= results.parsed.table.length) {
+      if (results.parsed.table.columns.table.data[rowIdx] !== currentTable) {
```
genehynson, line 49: whew! This is a very confident string of object introspection lol. How confident are you that results.parsed.table.columns.table.data will always exist?

drdelambre: these are always famous last words, but this component only renders results.parsed, and if it's gotten that far, i'm pretty sure that table.colums.table will exist until we swap out the parser

genehynson (approved):
man that was some heavy reading. nice work. lgtm other than that one concern I had which you seem confident about :)

PR #817, opened 2021-03-10
feat: delete annotations to redux and the API

### LINKED ISSUE #374: Annotations MVP: As a user I can delete an annotation
GIVEN I am looking at a dashboard cell 
AND there are annotations visible in my current time range

WHEN I hover over an annotation
THEN I see a tooltip containing time, summary, (optionally) message, a trash icon and an edit button

WHEN I mouse into the tooltip and click the trash icon
THEN I am prompted to confirm the action

WHEN I confirm the delete
THEN the Annotation is deleted
AND I see a notification fire (5s duration, success, "[AnnotationName] successfully deleted!"

UI is mostly built out for this story, just needs to be wired up and actually delete
AnnotationCard.tsx

zoesteinkamp:
Closes #374 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/src/annotations/actions/thunks.ts
+++ b/src/annotations/actions/thunks.ts
@@ -1,10 +1,11 @@
 import {
   getAnnotations,
-  writeAnnotation,
   getAnnotationStreamsDetails,
+  deleteAnnotation,
+  writeAnnotation,
 } from 'src/annotations/api'
 import {Dispatch} from 'react'
-
+import {deleteAnnotation as deleteAnnotationAction} from 'src/annotations/actions/creators'
 import {
   setAnnotations,
   setAnnotationStreams,
@@ -34,5 +35,12 @@ export const writeThenFetchAndSetAnnotations = (
 ) => async (dispatch: Dispatch<AnnotationAction>): Promise<void> => {
   await writeAnnotation(annotations)
 
+  fetchAndSetAnnotations()(dispatch)
+}
+export const deleteAnnotations = annotation => async (
+  dispatch: Dispatch<AnnotationAction>
+) => {
+  await deleteAnnotation(annotation)
+  dispatch(deleteAnnotationAction(annotation))
   fetchAndSetAnnotations()(dispatch)
 }
--- a/src/annotations/actions/creators.ts
+++ b/src/annotations/actions/creators.ts
@@ -1,4 +1,4 @@
-import {AnnotationResponse, AnnotationStream} from 'src/types'
+import {Annotation, AnnotationResponse, AnnotationStream} from 'src/types'
 
 export const ENABLE_ANNOTATION_STREAM = 'ENABLE_ANNOTATION_STREAM'
 export const DISABLE_ANNOTATION_STREAM = 'DISABLE_ANNOTATION_STREAM'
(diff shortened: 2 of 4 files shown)


src/annotations/actions/thunks.ts
```diff
@@ -23,3 +35,15 @@ export const writeThenFetchAndSetAnnotations = (
 
   fetchAndSetAnnotations()(dispatch)
 }
+export const deleteAnnotations = annotation => async (
+  dispatch: Dispatch<AnnotationAction | PublishNotificationAction>
+) => {
+  try {
+    await deleteAnnotation(annotation)
+    dispatch(deleteAnnotationAction(annotation))
+    dispatch(notify(deleteAnnotationSuccess()))
```
hoorayimhelping, line 44: let's move this down into the presentational component that calls this function

src/annotations/actions/thunks.ts
```diff
@@ -23,3 +35,15 @@ export const writeThenFetchAndSetAnnotations = (
...
+    fetchAndSetAnnotations()(dispatch)
+  } catch (error) {
```
hoorayimhelping, line 46: let's cut this whole `catch` block out and put it in the presentation component.

so the body of the function would be something like:

```ts
export const deleteAnnotations = annotation => async (
  dispatch: Dispatch<AnnotationAction | PublishNotificationAction>
) => {
  await deleteAnnotation(annotation)
  dispatch(deleteAnnotationAction(annotation))
  fetchAndSetAnnotations()(dispatch)
}

File: src/annotations/components/controlBar/AnnotationPill.tsx [hunk 3]
```diff
@@ -21,6 +22,10 @@ export const AnnotationPill: FC<Props> = ({id, name, description, color}) => {
   const handleDisable = (): void => {
     dispatch(disableAnnotationStream(id))
   }
+  const handleDeleteClick = (): void => {
+    console.log('hi im delete')
```
hoorayimhelping, line 26: hello delete! you should probably be removed

src/annotations/components/controlBar/AnnotationPill.tsx
```diff
@@ -21,6 +22,10 @@ export const AnnotationPill: FC<Props> = ({id, name, description, color}) => {
...
+    deleteAnnotations(id)
```
hoorayimhelping, line 27: so since I suggested we remove the `catch` block from the thunk above, let's add it in here:

```ts
try {
  deleteAnnotations(id)
  dispatch(notify(deleteAnnotationSuccess()))
} catch (error) {
  dispatch(notify(deleteAnnotationFailed(error.message)))  
}
```

the reason I think this makes more sense is because this creates a separation of concerns between redux and the display layer. If there's a problem in redux, it'll just throw the error and let the view layer that's using it handle it. The view layer can then get the error and handle it in a nice and helpful way.

hoorayimhelping (changes requested):
great work! a tweak about presentation layer stuff in redux

hercules-influx:
Metric | Score | Level
performance | 38 | :red_circle:
seo | 90 | :green_circle:
pwa | null | :question:

hoorayimhelping (approved):

PR #818, opened 2021-03-10
fix: deadman checks render table for non-numeric fields

genehynson:
Closes https://github.com/influxdata/idpe/issues/9590

Deadman checks should render the raw data view for non-numeric fields rather than showing an error message.

https://user-images.githubusercontent.com/6411855/110879006-13dc3700-8291-11eb-8743-0153f77927e7.mov

Test changes pass IDPE CI ✅ 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/timeMachine/components/Vis.tsx
+++ b/src/timeMachine/components/Vis.tsx
@@ -24,6 +24,10 @@ import {
   getYSeriesColumns,
 } from 'src/timeMachine/selectors'
 import {getTimeRangeWithTimezone} from 'src/dashboards/selectors'
+import {
+  setIsDisabledViewRawData,
+  setIsViewingRawData,
+} from 'src/timeMachine/actions'
 
 // Types
 import {RemoteDataState, AppState, ViewProperties} from 'src/types'
@@ -40,6 +44,8 @@ const TimeMachineVis: FC<Props> = ({
   timeRange,
   isInitialFetch,
   isViewingRawData,
+  setViewRawData,
+  setDisableRawData,
   files,
   viewProperties,
   giraffeResult,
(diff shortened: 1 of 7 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

joelfried (approved):

src/timeMachine/components/Vis.tsx
```diff
@@ -71,6 +77,26 @@ const TimeMachineVis: FC<Props> = ({
     'time-machine--view__empty': noQueries,
   })
 
+  // Handles deadman check edge case to allow non-numeric values
+  if (
```
drdelambre, line 81: i feel like this further accentuates a weird relationship between these components (vis and whatever middle bar), but idk what to do to smooth it over. Is this a sign of DRY breaking?

drdelambre (approved):
outside of it throwing red flags about tightly coupled components, seems fine

subirjolly (approved):

src/timeMachine/components/Vis.tsx
genehynson: I'm sure true DRY believers would tell us that we could decompose TimeMachine more to provide more abstraction layers that would ultimately remove some of these `if this then that` statements. But good lord I find myself questioning if adding that complexity is worth the investment. 

I'm a fan of KISS - keep it simple, stupid. 😆 A little code repetition, while it does potentially increase maintenance costs (e.g. fix a bug in two places), might be worth it if you're going to reduce the overall complexity of the code since less complexity means fewer bugs. 

So yes, I don't like this solution but de-coupling time machine sounds like a massive undertaking that we should strategize on....hopefully sooner rather than later 😅 

</end rant>

PR #820, opened 2021-03-11
feat: annotationsControlBarButton only controls the control bar

### LINKED ISSUE #819: Annotations: `AnnotationsToggleButton` should only toggle the annotations control bar
### Background
Huzzah! With the work we've recently done to hook up streams to redux and to make the adding of annotations more intuitive (see: #773, #791, #787) the the behavior of the `AnnotationsToggleButton` should be tweaked.

Currently it hides and shows annotations as well as hides and shows the AnnotationControlBar. It should hide or show the AnnotationControlBar.

### Acceptance Criteria
- `AnnotationsToggleButton` controls only the display of the `AnnotationsControlBar`
- It is named something more appropriate since it only controls the display of the control bar

ChitlangeSahas:
Closes #819

`AnnotationsToggleButton` is now refactored to be named `AnnotationsControlBarToggleButton` and it is only responsible for toggling the visibility of the control bar. 

--

Comment for future reference:
Should we add a toggle to hide / show annotations inside the control bar? It would be nice to be able to disable/enable the visibility of the annotations without having to 'x' out the streams in the control bar.

Also, not sure if this was intended but the annotations toggle does nothing in the data-explorer page. The control bar is always visible regardless.

--

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/src/annotations/components/AnnotationsControlBarToggleButton.tsx
+++ b/src/annotations/components/AnnotationsControlBarToggleButton.tsx
@@ -11,7 +11,7 @@ import {toggleShowAnnotationsControls} from 'src/userSettings/actions'
 // Components
 import {Button, ComponentColor, IconFont} from '@influxdata/clockface'
 
-export const AnnotationsToggleButton: FC = () => {
+export const AnnotationsControlBarToggleButton: FC = () => {
   const dispatch = useDispatch()
   const isVisible = useSelector(getAnnotationControlsVisibility)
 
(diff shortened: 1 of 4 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

hoorayimhelping (approved):
>Should we add a toggle to hide / show annotations inside the control bar? It would be nice to be able to disable/enable the visibility of the annotations without having to 'x' out the streams in the control bar.

Good eye on this one. We chatted over this in a meeting - right now there's only `default` being shown so it's not a YUGE deal, but we might add some kind affordance for this later.

hoorayimhelping:
>Also, not sure if this was intended but the annotations toggle does nothing in the data-explorer page. The control bar is always visible regardless.

I'm going to write up a ticket to remove annotations from Data Explorer I think

PR #832, opened 2021-03-12
fix: Giraffe gauge

### LINKED ISSUE #629: Replace Gauge graph in UI with Gauge from Giraffe
the goal of this is to replace the gauge implementation in the UI with the implementation in Giraffe lib.

This would allow us to improve the graph without making the change in two places.

zoesteinkamp:
Closes #629

We are changing the gauge graph to reference the giraffe gauge graph
- [ ] Rebased/mergeable
- [ ] Tests pass


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

zoesteinkamp:
now we just need a feature flag!

src/visualization/types/Gauge/view.tsx
```diff
@@ -29,32 +29,52 @@ const GaugeChart: FC<Props> = ({result, properties}) => {
...
+          suffix: suffix,
+          tickPrefix: tickPrefix,
+          tickSuffix: tickSuffix,
+          decimalPlaces: decimalPlaces,
+          gaugeColors: colors,
+        },
+      ],
+    }
+    return <Plot config={config} />
+  } else {
```
hoorayimhelping, line 52: since this returns, you don't even need the `else` here

hoorayimhelping (approved):

hoorayimhelping (commented):
This LGTM, but I think we should wait for Timmy

src/visualization/types/Gauge/view.tsx
TCL735: i agree with this. imo, dropping the else and putting the return outside looks better.

TCL735 (approved):

ChitlangeSahas (approved):

PR #838, opened 2021-03-15
fix: custom aggregate deselection prohibition toast

### LINKED ISSUE #25: Query Builder: Attempt to deselect function should let the user know that the action is prohibited
# Background
After delivering auto-aggregation, we have received reports from users that not being able to deselect an auto selected aggregate results in a confusing user experience. Specifically, they mention that it feels like a bug. We would like to express to users that this is indeed the expected behavior.

cant-deselect

## Acceptance Criteria

WHEN I am using the `custom` section of `Query Builder`
AND I attempt to deselect the final aggregate function in the list
THEN I should receive a notification letting me know that this behavior is not possible

ShmuelLotman:
Closes #25 

This PR introduces a toast when a user attempts to deselect the only remaining custom aggregate function, as this is a prohibited action and users have mentioned that it feels like a bug without some sort of message clarifying its intended. 

This also addresses something i noticed when working with toasts - the notification is prevented from being duplicated if the type is the same, but not if the message is the same, and it was allowing the notifications for the deselection prohibition to be duplicated. I added a check to ensure we aren't showing a duplicate message, but i'm curious if there was a reason i've failed to see as to why the check was specifically only on type and nothing else. If the reviewer can clarify, i'd appreciate it! 

- [x] Rebased/mergeable
- [x] Tests pass

Screen Shot 2021-03-15 at 9 53 09 AM

### DIFF
--- a/src/shared/copy/notifications.ts
+++ b/src/shared/copy/notifications.ts
@@ -56,6 +56,12 @@ const defaultDeletionNotification: NotificationExcludingMessage = {
 //  Misc Notifications
 //  ----------------------------------------------------------------------------
 
+export const prohibitedDeselect = (message?: string): Notification => ({
+  ...defaultErrorNotification,
+  message:
+    message ?? 'You must have at least one custom aggregate function selected',
+})
+
 export const newVersion = (version: string): Notification => ({
   ...defaultSuccessNotification,
   style: NotificationStyle.Info,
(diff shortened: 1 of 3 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

joelfried (approved):

Issue #728, opened 2021-02-18
Annotations: Measure annotations usage


hoorayimhelping:
### Background
Measuring things is probably a good idea. It's never really been proven scientifically, but phenomenologically, it seems pretty sound. We should measure how annotations are being used.

To keep things consistent, we want to organize the metrics into a bit of a taxonomy. The hierarchy should go something like:

```
page -> feature -> ui component -> state
```

For example, on a dashboard page with the annotations ui rendered.

When a user clicks the toggle to toggle annotating with one click:
`dashboard.annotations.render_annotation_toggle.clicked`

Or when a user clicks the graph and saves an annotation using the overlay that pops up
`dashboard.annotations.add_annotation.added`
or if they cancel:
`dashboard.annotations.add_annotation.canceled`

### Acceptance Criteria
- Measurements to include:
  - [ ] Annotations button is clicked, toggling the control bar (pass in the value that the state is transitioning to as a tag)
  - #### Annotations Interaction
  - [x] A graph is clicked to start the add annotation process
  - [x] An annotation is created (Annotation form is submitted)
  - [x] An annotation overlay is canceled without being created
  - [ ] An annotation is deleted
  - #### Annotation Control Bar
  - [x] Annotate button is clicked to toggle or untoggle annotate mode
  - ~[ ] A stream is searched for~
  - ~[ ] A stream is added to the list of visible streams (clicked from the stream search box)~
  - ~[ ] A stream is removed by clicking the `x` on an annotation pill~

PR #856, opened 2021-03-16
feat: add metrics call to annotations change write mode

hoorayimhelping:
Connects #728

Adds a quick example call to `event` for annotations

### DIFF
--- a/src/annotations/components/controlBar/AnnotationsControlBar.tsx
+++ b/src/annotations/components/controlBar/AnnotationsControlBar.tsx
@@ -25,8 +25,8 @@ import {AnnotationsSearchBar} from 'src/annotations/components/controlBar/Annota
 // Selectors
 import {isSingleClickAnnotationsEnabled} from 'src/annotations/selectors'
 
-// Constants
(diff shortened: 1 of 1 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

src/annotations/components/controlBar/AnnotationsControlBar.tsx
```diff
@@ -37,11 +37,14 @@ export const AnnotationsControlBar: FC = () => {
   const dispatch = useDispatch()
 
   const changeWriteMode = () => {
+    event('dashboard.annotations.change_write_mode.toggle', {
```
hoorayimhelping, line 40: this is the key that will show up in query builder

here's how it'll look in a script:
Screen Shot 2021-03-15 at 5 36 47 PM

hoorayimhelping:
>are we removing all the route constants now?

@jrenee42 I don't think we have to go nuts with it. It's not a huge deal, it's just that I've found codebases where urls are spelled out literally and easily searchable are much easier to work with, so I'm fixing these as I see them.

PR #735, opened 2021-02-19
fix: revert to clockface 2.4.1

hoorayimhelping:
Closes #734 

Clockface [2.6.0](https://github.com/influxdata/clockface/blob/master/CHANGELOG.md#260-2021-02-10) introduced a change to how checkboxes are handled. This caused a bug in Cloud2 where certain checkboxes couldn't be checked. The tests that normally catch these were turned off, and the process of turning them on produced this bug.

- [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [x] Rebased/mergeable
- [x] Tests pass

PR #862, opened 2021-03-16
feat: re-applying the typeahead dropdown

jrenee42:
it was removed when the clockface 2.4.1 upgrade had bugs so had to be reverted

   was: Revert "fix: revert to clockface 2.4.1 (#735)"

This reverts part of the commit 4f75c2c42a854974d3ead4dd87fb6fe0ffd2b941.
   (the type-ahead dropdown, that is)

### DIFF
--- a/src/variables/components/TypeAheadVariableDropdown.tsx
+++ b/src/variables/components/TypeAheadVariableDropdown.tsx
@@ -1,0 +1,308 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import {connect, ConnectedProps} from 'react-redux'
+
+// Components
+import {
+  Dropdown,
+  DropdownMenuTheme,
+  ComponentStatus,
+  Input,
+  MenuStatus,
+} from '@influxdata/clockface'
+
+// Actions
+import {selectValue} from 'src/variables/actions/thunks'
+
+// Utils
+import {getVariable, normalizeValues} from 'src/variables/selectors'
+
+// Types
+import {AppState, RemoteDataState} from 'src/types'
+
+interface OwnProps {
+  variableID: string
+  testID?: string
+  onSelect?: () => void
+}
+
+type ReduxProps = ConnectedProps<typeof connector>
+type Props = OwnProps & ReduxProps
+
+interface MyState {
+  typedValue: string
+  actualVal: string
+  selectIndex: number
+  shownValues: string[]
+  selectHappened: boolean
+  menuOpen: MenuStatus
+  loaded: boolean
+}
+
+class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
+  constructor(props) {
+    super(props)
+    this.state = {
+      typedValue: '',
+      actualVal: '',
+      selectIndex: -1,
+      shownValues: props.values,
+      selectHappened: false,
+      menuOpen: null,
+      loaded: false,
+    }
+  }
+
+  // set the 'shownValues' after loading, and
+  // resets the menuOpen variable
+  componentDidUpdate(prevProps, prevState) {
+    const prevVals = prevProps.values
+    const {values, selectedValue} = this.props
+    const {actualVal, loaded, selectHappened, menuOpen} = this.state
+    const {
+      actualVal: prevActualVal,
+      selectHappened: prevSelectHappened,
+      menuOpen: prevMenuOpen,
+    } = prevState
+
+    // this is for updating the values:
+    // (only want this to run *once* when the values get loaded)
+    if (!loaded && prevVals.length !== values.length) {
+      this.setState({
+        shownValues: values,
+        typedValue: selectedValue,
+        loaded: true,
+      })
+    }
+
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
performance | 41 | :red_circle:
seo | 90 | :green_circle:
pwa | null | :question:

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+        style={{width: '140px'}}
+        className="variable-dropdown--dropdown"
+        testID={this.props.testID || `variable-dropdown--${name}`}
+        onClickAway={this.onClickAwayHere}
+        menuOpen={menuOpen}
+        button={(active, onClick) => (
+          <Dropdown.Button
+            active={active}
+            onClick={onClick}
+            testID="variable-dropdown--button"
```
hoorayimhelping, line 193: this `testID` needs to be here for test compatibility, I'm assuming?

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+            status={dropdownStatus}
+          >
+            <Input
+              style={widthStyle}
+              placeholder={placeHolderText}
+              onChange={e => this.filterVals(e.target.value)}
```
hoorayimhelping, line 199: it's typical convention to just pass in the handler, and let it handle the event being passed in. so by making both handlers take in an event, it implicitly communicates it's an event handler. so `filterVals` argument should change from `needle` to `event`. and this call would look like:

```onChange={this.filterVals}```

File: src/variables/components/TypeAheadVariableDropdown.tsx [hunk 3]
```diff
@@ -0,0 +1,308 @@
...
+              value={typedValue}
+              onKeyDown={e => this.maybeSelectNextItem(e)}
```
hoorayimhelping, line 201: react will do this for you if you do this:

```
onKeyDown={this.maybeSelectNextItem}
```

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+        )}
+      />
+    )
+  }
+
+  private getWidth(placeHolderText) {
+    const {values} = this.props
+    const allVals = [placeHolderText, ...values]
+    const longestItemWidth = Math.floor(
+      allVals.reduce(function(a, b) {
```
hoorayimhelping, line 236: nice use of `reduce` !

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+                </Dropdown.Item>
+              )
+            })}
+          </Dropdown.Menu>
+        )}
+      />
+    )
+  }
+
+  private getWidth(placeHolderText) {
```
hoorayimhelping, line 232: let's earmark this function for David Kim's attention

hoorayimhelping (commented):
This is really solid work Jill, and pretty easy to see what all is going on. Love the keyboard shortcuts!

Going to just comment on this one, rather than request changes. I don't want to block this while I'm on PTO for day.

joelfried (approved):

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+    // (only want this to run *once* when the values get loaded)
+    if (!loaded && prevVals.length !== values.length) {
+      this.setState({
+        shownValues: values,
+        typedValue: selectedValue,
+        loaded: true,
+      })
+    }
+
+    // unset the menuOpen; it should be set to closed (or open) only once; then undone
```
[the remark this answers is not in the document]
jrenee42: from Dropdown.tsx (in clockface):
 /**
   * optional way for the owner of this dropdown to set the menu to be open or closed:
   * 'open':  the menu will be open
   * 'closed': the menu will be closed
   * any other string will be ignored.
   *
   * and the menu  (open/closed) only changes programmatically when this value changes.
   * so the owner is responsible for resetting this value
   * (if the string is set to 'open', and then the user closes it, and the code sets it to open again,
   * unless the code sets it to something else in between (like null or 'close'),
   * then nothing will happen- the menu will not open)
   * */
   MenuOpen?: MenuStatus

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -0,0 +1,308 @@
...
+      // down arrow
+      newIndex = selectIndex + 1
+    } else if (e.keyCode === 38) {
+      // up arrow
+      newIndex = selectIndex - 1
+    }
+
+    const numItems = shownValues.length
+
+    if (numItems && newIndex >= 0 && newIndex < numItems) {
```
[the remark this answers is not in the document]
jrenee42: yes, naming things to clarify is always good

PR #868, opened 2021-03-17
fix: validate variables and flag when renaming is required

### LINKED ISSUE #611: Error highlighting (red squiggly lines) doesn't go away in tools, even when there is no error present.
**Steps to reproduce:**

1. Open up the data explorer in the tools cluster.
2. Type some flux with a simple syntax error to get some red squiggly lines under your code.
3. Resolve the syntax error. The red lines should persist. When you hover your mouse over the indicated text, no error message will appear
4. Refresh the page. Now typing *any* flux code should result in red squiggly lines that do not go away.

**Expected behavior:**
Error highlighting appears when there's an error. When you hover your mouse over the red lines, you get an error message. When you resolve the error, the highlighting goes away.

**Actual behavior:**
Error highlighting does not go away, and when you hover your mouse over the lines, no error message appears.

**Environment info:**
Can reproduce in tools

Cannot reproduce in us-central-1-gcp

It may or may not be in other cloud environments

OSS is unaffected

**Additional details:**

Behavior was first reported in InfluxData's internal slack on Monday 2/1/2021. It was originally thought to be an LSP issue because it coincided with a new LSP release. While that may still be the case, two pieces of evidence indicate that this may be on the UI end of things:

1. In the past, when we've seen this kind of behavior with the LSP (highlighting a flux error and then never updating once the error has been resolved) it's been because the LSP has crashed and is no longer sending or receiving messages. In this case, the LSP is still running, i.e. autocomplete, syntax highlighting, etc. still work.
2. The timing coincides with some other bug reports regarding recent weirdness in the flux editor, which may or may not be related:

- https://github.com/influxdata/ui/issues/610
- https://github.com/influxdata/ui/issues/609

If, after digging into this, we can confirm that the problem is not with the UI but with the LSP, please open an issue in https://github.com/influxdata/flux-lsp

TCL735:
Closes #611 

Help the user create proper names for variables so that Flux does not complain with squiggly lines in the Script Editor. We do this in two ways:
- validate the variable names when manually creating or renaming
- flag the variables on the Variables tab (in Settings) that have bad names to let the user know to rename them

Screen Shot 2021-03-23 at 4 02 59 PM

### DIFF
--- a/src/variables/components/VariableList.tsx
+++ b/src/variables/components/VariableList.tsx
@@ -1,18 +1,23 @@
 // Libraries
-import React, {PureComponent} from 'react'
-import memoizeOne from 'memoize-one'
+import React, {FC, useMemo} from 'react'
 
 // Components
 import {ResourceList} from '@influxdata/clockface'
 import VariableCard from 'src/variables/components/VariableCard'
 
 // Types
-import {OverlayState, Variable} from 'src/types'
+import {Variable} from 'src/types'
 import {SortTypes} from 'src/shared/utils/sort'
 import {Sort} from '@influxdata/clockface'
 
 // Selectors
 import {getSortedResources} from 'src/shared/utils/sort'
+import {validateVariableName} from '../utils/validation'
+
+interface ValidatedVariable {
+  description: string
+  variable: Variable
+}
 
 interface Props {
   variables: Variable[]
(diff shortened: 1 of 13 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

Asalem1 (approved):
LGTM

src/variables/components/VariableList.tsx
```diff
@@ -24,73 +23,38 @@ interface Props {
...
-      sortDirection,
-      sortType
-    )
-
-    return sortedVariables.map((variable, index) => (
-      <VariableCard
-        key={variable.id || `variable-${index}`}
-        variable={variable}
-        onDeleteVariable={onDeleteVariable}
-        onEditVariable={this.handleStartEdit}
```
Asalem1, line 84: Great catching on this.

PR #886, opened 2021-03-17
chore: run UI tests via k8s-idpe in UI CI

### LINKED ISSUE #858: tests: run cypress e2e tests via k8s-idpe in UI CI
Currently the cypress e2e tests are run via monitor-ci. This is fine but now there exists a discrepancy between IDPE CI and the UI CI in how these tests are run. To avoid issues and people having to manually execute these tests, we should run the e2e tests via k8s-idpe to match IDPE. 

This will likely involve exploring how IDPE CircleCI config works and copying the "cd-app-e2e-test" config into the UI CircleCI config file. We'll have to clone the k8s-idpe repo, create the cluster, and execute the `./scripts/run-test-jobs.bash` script. 

Note: we will only be able to run the `cloud-e2e` tests via k8s-idpe until we produce an OSS UI image. For now, monitor-ci should continue to run the oss-e2e tests. 

@lukebond @ebb-tide if I'm missing anything from this, please let me know

genehynson:
Closes #858

Adds a new job `cloud-e2e-k8s-idpe` to our pipeline that runs our cloud cypress tests using k8s-idpe.

_Note: The plan is to leave `cloud-e2e-k8s-idpe` as optional for a few days and continue to run `cloud-e2e` in parallel in case any unexpected k8s-idpe failures arise. While in this state, `cloud-e2e-k8s-idpe` will not be required to pass for PRs to be mergable._


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

.circleci/config.yml
```diff
@@ -436,6 +415,68 @@ jobs:
           path: /tmp/results.html
       - store_artifacts:
           path: /tmp/results.json
+
+  cloud-e2e-k8s-idpe:
+    machine:
+      # use Ubuntu 16 image as jsonnet requires libraries newer than in Ubuntu 14
+      image: ubuntu-1604:202007-01
```
drdelambre, line 422: can we run.. 20?

genehynson: idk do we want to use a different image than IDPE CI?

genehynson: `ubuntu-2004:202101-01` and `ubuntu-2004:202010-01` are available and I could give them a try

genehynson: Oh I see we're using `ubuntu-2004:202008-01` in our other jobs, I'll try that one

.circleci/config.yml
```diff
@@ -436,6 +415,68 @@ jobs:
...
+            mkdir -p /tmp/cypress_artifacts
+      - run:
+          name: Copy over the testing variables
+          command: cp ./monitor-ci/env.testing.cloud ./monitor-ci/.env
+      - run: docker login -u=$QUAY_USER -p=$QUAY_PASS quay.io
+      - run:
+          name: Build ui-e2e image
+          command: |
+            cd ./monitor-ci && make build NODE=cypress
+      # TODO: enable this when we remove cloud-e2e job
```
drdelambre, line 452: we dont want to push these images unless the tests pass, which doesn't seem to be what this is saying. it's fine to be wasteful of the ${CIRCLE_SHA1} tags, but definitely get the latest tag out of here, or else any changes to the tests... even if they're broken... get sent directly to the next idpe PR

drdelambre: like master should have already been tested.. but this effectively says "dont test the merge" right? and we should either be more explicit about that, or move this

.circleci/config.yml
drdelambre: thanks. 16 is real old!

.circleci/config.yml
genehynson: huh yeah I see what you're saying but it's interesting because this is what cloud-e2e is doing currently, right? https://github.com/influxdata/ui/blob/master/.circleci/config.yml#L189

genehynson: I can move both of them to be executed after the tests run since that makes more sense

.circleci/config.yml
```diff
@@ -436,6 +415,68 @@ jobs:
...
+      #     steps:
+      #       # The cypress image used within the ui-e2e image is a non-standard slim image and its definition and versions are
+      #       # maintained here: https://github.com/influxdata/cypress-slim
+      #       - run: docker tag local/cypress:latest quay.io/influxdb/ui-e2e:${CIRCLE_SHA1}
+      #       - run: docker tag quay.io/influxdb/ui-e2e:${CIRCLE_SHA1} quay.io/influxdb/ui-e2e:latest
+      #       - run: docker push quay.io/influxdb/ui-e2e:${CIRCLE_SHA1}
+      #       - run: docker push quay.io/influxdb/ui-e2e:latest
+      - run:
+          name: Run E2e test
+          command: cd ./idpe/ && export QUAY_CD_USER=$QUAY_USER && export QUAY_CD_PASSWORD=$QUAY_PASS && ./scripts/ci/cd-app-e2e-test.bash | tee /tmp/artifacts/e2e-test.log || true
```
genehynson, line 454: the intention of `|| true` here is to suppress all `exit 1` codes until we want to disable `cloud-e2e` and depend on `cloud-e2e-k8s-idpe`

drdelambre: i dont understand this. we can make it fail for a week if we dont gate on it and tell people to ignore it... but it shouldn't fail anyways right?

genehynson: if it fails, it'll block their PRs. So I have to have the job "pass" while the step failed. But right, it shouldn't fail (unless the tests fail for a good reason)

genehynson: didn't realize we can control what jobs are "required" via github - removing this `|| true` nonsense

.circleci/config.yml
```diff
@@ -203,6 +182,16 @@ jobs:
       - store_artifacts:
           path: monitor-ci/test-artifacts/screenshots
           destination: test_artifacts/screenshots
+      - when:
```
drdelambre, line 185: i'm not sure if putting it as the last step of a job ensures that it it doesn't happen, or if we have to make it a different job, but we can test that easily enough after we merge

drdelambre (approved):

PR #903, opened 2021-03-18
fix: various bug fixes in notebooks

### LINKED ISSUE #890: Notebooks: Updating filter doesn't take effect
If you select a bucket and filter, and run preview, it works fine. If you then change the filters, the new filters will show in the UI but the query continues to execute with the original filters. You need to refresh the page for the new filters to take effect.

Rick saw something similar when he selected a different bucket. The query continued to run against the original bucket.

genehynson:
Closes #890

- Fixes issue where results table doesn't show results from latest query due to memoization
- Fixes issue where visualization panel shows "an error has occurred" after executing some queries due to recent annotations changes to time machine
- Fixes issue where multiple list items in metric selector had the same key
- Adds e2e test for metric selector
- Adds testIDs to some components to make testing easier

- [x] [E2E Tests pass in k8s-idpe](https://docs.influxdata.io/development/guides/local-development/#three-ways-to-run-the-ui-e2e-tests-against-the-kind-cluster) (applicable only if you edited Cypress tests)

### DIFF
--- a/src/flows/components/panel/Results.tsx
+++ b/src/flows/components/panel/Results.tsx
@@ -149,7 +149,10 @@ const ResultTable: FC<TableProps> = ({
         .fill(null)
         .map((_, idx) => {
           const cells = cols.map(c => (
-            <Table.Cell key={`t${tIdx}:h${c.name}:r${idx}`}>
+            <Table.Cell
+              key={`t${tIdx}:h${c.name}:r${idx}`}
+              testID={`table-cell ${c.data[idx]}`}
+            >
               {c.data[idx]}
             </Table.Cell>
           ))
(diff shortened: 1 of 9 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:

src/flows/components/panel/Results.tsx
```diff
@@ -171,7 +174,7 @@ const ResultTable: FC<TableProps> = ({
         </Table>
       )
     })
-  }, [height, startRow])
+  }, [height, startRow, results])
```
drdelambre, line 177: [ eye twitching intensifies ]

drdelambre (approved):
🔥
THANKS FOR THE TESTS

Issue #879, opened 2021-03-17
Cannot add threshold to single stat


genehynson:
**Steps to reproduce:**

I'm not actually sure how I got in this situation. I was editing the "customization" settings for single stat after cloning the dashboard cell and now I cannot add thresholds to this cell anymore.

https://user-images.githubusercontent.com/6411855/111526165-9fcfe200-871b-11eb-9061-417d6a60de2d.mov

**Expected behavior:**
I should be able to add thresholds.

**Actual behavior:**
I click the button and nothing happens. Refreshing does _not_ fix the issue.

**Environment info:**

Tools

**Logs:**

![image](https://user-images.githubusercontent.com/6411855/111526065-8333aa00-871b-11eb-974b-3db083c6f584.png)

PR #916, opened 2021-03-19
fix: use threshold settings from shared/components

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/879

I hypothesize that we were using the outdated `ThresholdSettings` code. Using different import from `'src/shared/components/ThresholdsSettings'` seems to fix it. 

- [ ] [E2E Tests pass in k8s-idpe](https://docs.influxdata.io/development/guides/local-development/#three-ways-to-run-the-ui-e2e-tests-against-the-kind-cluster) (applicable only if you edited Cypress tests)


hercules-influx:
Metric | Score | Level
performance | 28 | :red_circle:
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=916)

drdelambre:
lemme see where else we're using that file

genehynson:
> Metric	Score	Level
> performance	28	🔴
> accessibility	96	🟢
> best-practices	93	🟢
> seo	90	🟢
> pwa	null	❓
> [Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=916)

I'm concerned that the performance scored dropped to 28?? It was hovering above 40...

ChitlangeSahas:
I see that now, I wonder why that would be the case!

drdelambre (changes requested):
gonna ask that you flip that movement. even though the component is in src/shared, it seems like it's not shared with anything?

so it'd make more sense to just `git mv` the shared version into src/visualization

seems like a trivial distinction, but by keeping the code collocated, you group concepts in another dimension to reenforce encapsulation, hopefully curtailing arbitrary reuse and preventing the spaghetti from popping back out again

*all claims dictated, not read, and with no guarantee they were made with a complete view of ramifications

genehynson:
> I see that now, I wonder why that would be the case!

looks like it was a flake - rerun brings it back to 41

drdelambre (approved):

Issue #930, opened 2021-03-22
Update the way Upgrade Now button is rendered


Asalem1:
The UI currently relies on a user being PAYG to not render the Upgrade Now button, but we need to also consider a `free` account in a `beta` region when not rendering that button. That is to say, that if a user has signed up in a beta region with a free account, they should **not** see the Upgrade Now button. This will be blocked on the new /me API being developed by Quartz

gunnaraasen:
The specific cluster URL is europe-west1-1.gcp.cloud2.influxdata.com.

Asalem1:
Waiting on the `/me` endpoint to be created

Asalem1:
This logic should be behind a feature flag since it isn't hasn't been tested live and we don't want to muck up any existing user experiences

PR #947, opened 2021-03-23
feat: add exception for beta region deployment on upgrade now

### LINKED ISSUE #944: Hide upgrade button for beta region (GCP Belgium)
**Proposal:**
Disable the "Upgrade now" button for beta regions. The UI detects whether the region matches a hardcoded a beta region identifier and hides the checkout button if it matches.

**Current behavior:**
Quartz currently handles disabling the button. However, the UI can potentially send users directly to the checkout page, at which point an error will be returned on beta regions.

**Desired behavior:**
Beta regions do not yet support paid usage so we do not want to show the upgrade button to users in those regions.

**Alternatives considered:**
The only way we could handle this on the quartz side is to show an error page after users click on upgrade button. https://github.com/influxdata/quartz/issues/2871

Asalem1:
Closes #944 

This PR is a brief workaround until we can address the core issues for #930. As such, it's a hack to allow a beta region to be deployed today without needing to develop proper infrastructure for the core issue.

### DIFF
--- a/src/shared/components/CloudUpgradeNavBanner.tsx
+++ b/src/shared/components/CloudUpgradeNavBanner.tsx
@@ -17,7 +17,11 @@ import {
 import CloudOnly from 'src/shared/components/cloud/CloudOnly'
 
 // Constants
-import {CLOUD_URL, CLOUD_CHECKOUT_PATH} from 'src/shared/constants'
+import {
+  BETA_REGIONS,
+  CLOUD_URL,
+  CLOUD_CHECKOUT_PATH,
+} from 'src/shared/constants'
 import {
   HIDE_UPGRADE_CTA_KEY,
   PAID_ORG_HIDE_UPGRADE_SETTING,
@@ -39,9 +43,10 @@ const CloudUpgradeNavBanner: FC<StateProps> = ({inView}) => {
   // The follow up to this issue will address the hack here:
   // https://github.com/influxdata/ui/issues/930
 
-  const isBetaRegion = window.location.hostname.includes(
-    'europe-west1-1.gcp.cloud2.influxdata.com'
+  const isBetaRegion = BETA_REGIONS.some((pathName: string) =>
+    window.location.hostname.includes(pathName)
   )
+
   return (
     <>
       {inView && !isBetaRegion && (
--- a/src/shared/components/CloudUpgradeNavBanner.tsx
+++ b/src/shared/components/CloudUpgradeNavBanner.tsx
@@ -31,9 +31,20 @@ interface StateProps {
 }
 
 const CloudUpgradeNavBanner: FC<StateProps> = ({inView}) => {
+  // TODO(ariel): we need to build out an exception for beta regions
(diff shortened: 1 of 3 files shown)


drdelambre:
this hides the button itself, but not all of the popups and overlays around it

like.. what happens to the flex box in RateLimitAlertContent?

drdelambre:
and we're introducing a new source of truth around the definition of what defines a beta region... why is this dictated on a string in the frontend and not on the backend where we have access to vault or whatever determines that?

randycoulman:
@drdelambre Re: the source of truth issue. You are absolutely right. Quartz is the source of truth for this info (via a config stored in the Tools Vault instance) and the UI will ask Quartz, but we're not yet at a point where the UI can talk to Quartz, so we're doing this quick-n-dirty hack as a stop-gap to allow the deployment of a new beta region without a broken experience for users. See the comment @Asalem1 added for more details and links to issues.

randycoulman (approved):
This looks OK to me, but I'd look at @drdelambre's point about hiding other UI elements; I don't have enough context to judge whether that's a problem or not. Probably relates to how the `inView` prop is determined.

drdelambre:
@randycoulman is there a prioritized issue for the follow up work for this on your side so that we can schedule it's removal into our sprints?

randycoulman:
@drdelambre It's part of the overall UI unification epic, but there probably isn't a specific story for it yet.  We have a cleanup story for later (https://github.com/influxdata/quartz/issues/3899), but that's more of a "stop syncing state with IDPE once we no longer need to" thing.  We'll add a specific story if it's not covered by other stories.

randycoulman:
New story: https://github.com/influxdata/quartz/issues/4369

drdelambre:
@randycoulman thanks for the extra context on this one and thanks for the issue, it helped me frame how much effort we should be putting into this part

src/shared/components/CloudUpgradeNavBanner.tsx
```diff
@@ -31,9 +31,20 @@ interface StateProps {
...
+  // TODO(ariel): we need to build out an exception for beta regions
+  // This current hack is being placed to allow a Beta region to be deployed
+  // without allowing users to get navigated to a Quartz 404. This hack is being implemented
+  // to address the following issue:
+  // https://github.com/influxdata/ui/issues/944
+  // The follow up to this issue will address the hack here:
+  // https://github.com/influxdata/ui/issues/930
+
+  const isBetaRegion = window.location.hostname.includes(
+    'europe-west1-1.gcp.cloud2.influxdata.com'
```
drdelambre, line 43: can we make this as an array in case we gotta lean into this while the apis are being updated, and put it into a const somewhere now that it's shared?

drdelambre:
re empty flex boxes: i misread how the inView is set on the component. no need to restructure the thing if that's how it was working before. seems odd, but that'd be a different ticket that no one is asking for currently, so.. :+1:

drdelambre (approved):

PR #953, opened 2021-03-23
fix: telegraf no buckets overlay now works

### LINKED ISSUE #948: UI: telegraf config doesn't work properly if there are no buckets
### Background
When there are no buckets available, the configure telegraf plugin flow doesn't work and offers no way to resolve it:

### Steps to reproduce
1. Remove all buckets
1. Go to Data -> Telegraf
1. Click Create Configuration
1. Observe the weird overlay

ShmuelLotman:
Closes #948 

Under this PR, if a user has no buckets the overlay will now prompt the user to create a bucket. When they do, they will then be able to see the regular overlay as was intended. 
- [x] [E2E Tests pass in K8S-IDPE](https://docs.influxdata.io/development/guides/local-development/#three-ways-to-run-the-ui-e2e-tests-against-the-kind-cluster) (applicable only if you edited Cypress tests)
Screen Shot 2021-03-23 at 4 33 33 PM

### DIFF
--- a/src/dataLoaders/components/collectorsWizard/select/StreamingSelector.tsx
+++ b/src/dataLoaders/components/collectorsWizard/select/StreamingSelector.tsx
@@ -14,7 +14,7 @@ import {
   SquareGrid,
 } from '@influxdata/clockface'
 import {ErrorHandling} from 'src/shared/decorators/errors'
-
+import CreateBucketButton from 'src/buckets/components/CreateBucketButton'
 // Constants
 import {
   PLUGIN_BUNDLE_OPTIONS,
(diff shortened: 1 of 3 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=953)

src/dataLoaders/components/collectorsWizard/select/StreamingSelector.tsx
```diff
@@ -71,52 +71,64 @@ class StreamingSelector extends PureComponent<Props, State> {
...
+              })}
+            </SquareGrid>
+          </>
+        ) : (
+          <>
+            <Grid.Row style={{width: 'auto'}}>
+              <CreateBucketButton />
+            </Grid.Row>
+            <h3 data-testid="create-bucket-prompt">
+              Create a Bucket To Show Telegraf Plugins
```
hoorayimhelping, line 128: 🤔 What if we changed this to "You can't install a telegraf plugin until you create a bucket"

I'm not sure this current verbiage explains why the user would want to show the telegraf plugins and it feels a little gatekeeping - like "you have to make a bucket before we show you this page."

I'd prefer to explain a bit of the process to the user as we solve their problem - basically some variation of "without a bucket, you can't write to anywhere, so you need a bucket" what do you think? also @daviddkkim and @russorat

hoorayimhelping (approved):
works for me. I'd like to change the verbiage, but I don't think we should block on this for now, as long as Russ and/or David take a look

src/dataLoaders/components/collectorsWizard/select/StreamingSelector.tsx
ShmuelLotman: Yeah this was just the first thing that popped into my head, fifgured i would throw it in and discuss it when it got pushed up, thanks for bringing this up here

PR #957, opened 2021-03-23
chore(lint): no nested ternaries

### LINKED ISSUE #956: UI linting: No nested ternary operators
Turn on the lint rule `no-nested-ternary` https://eslint.org/docs/rules/no-nested-ternary

hoorayimhelping:
Closes #956

Turns on lint checking for nested ternary operators

- [ ] [E2E Tests pass in K8S-IDPE](https://docs.influxdata.io/development/guides/local-development/#three-ways-to-run-the-ui-e2e-tests-against-the-kind-cluster) (applicable only if you edited Cypress tests)

### DIFF
--- a/cypress/e2e/shared/telegrafs.test.ts
+++ b/cypress/e2e/shared/telegrafs.test.ts
@@ -76,7 +76,16 @@ describe('Collectors', () => {
           const buckets = bucketz.slice(0).sort((a, b) => {
             const _a = a.toLowerCase()
             const _b = b.toLowerCase()
-            return _a > _b ? 1 : _a < _b ? -1 : 0
+
+            if (_a > _b) {
+              return 1
+            }
+
+            if (_a < _b) {
+              return -1
+            }
+
+            return 0
           })
 
           cy.get('code').should($el => {
--- a/src/telegrafs/components/TelegrafOutputOverlay.tsx
+++ b/src/telegrafs/components/TelegrafOutputOverlay.tsx
@@ -74,7 +74,16 @@ class TelegrafOutputOverlay extends PureComponent<Props> {
       .sort((a, b) => {
         const _a = a.name.toLowerCase()
         const _b = b.name.toLowerCase()
-        return _a > _b ? 1 : _a < _b ? -1 : 0
+
+        if (_a > _b) {
+          return 1
+        }
+
+        if (_a < _b) {
+          return -1
+        }
+
+        return 0
       })
   }
 
--- a/src/timeMachine/components/WindowPeriod.tsx
+++ b/src/timeMachine/components/WindowPeriod.tsx
@@ -49,11 +49,13 @@ const WindowPeriod: FunctionComponent<Props> = ({
   let durationDisplay = period
 
   if (!period || isAutoWindowPeriod) {
-    durationDisplay = autoWindowPeriod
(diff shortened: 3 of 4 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=957)

Asalem1:
@hoorayimhelping can we merge this in?

cypress/e2e/shared/telegrafs.test.ts
```diff
@@ -76,7 +76,16 @@ describe('Collectors', () => {
...
+
+            if (_a > _b) {
+              return 1
+            }
+
+            if (_a < _b) {
+              return -1
+            }
+
+            return 0
```
jrenee42, line 88: while this is a correct undoing of the ternary operator, why not use the inbuilt string localeCompare?  (it's optimized so should be faster)

src/telegrafs/components/TelegrafOutputOverlay.tsx
```diff
@@ -74,7 +74,16 @@ class TelegrafOutputOverlay extends PureComponent<Props> {
...
+
+        if (_a > _b) {
+          return 1
+        }
+
+        if (_a < _b) {
+          return -1
+        }
+
+        return 0
```
jrenee42, line 86: see above comment...and since this is in 2 places, we should have a utils  method to case-insensitive compare strings in one place.

src/timeMachine/components/WindowPeriod.tsx
```diff
@@ -49,11 +49,13 @@ const WindowPeriod: FunctionComponent<Props> = ({
...
-        : `${AGG_WINDOW_AUTO} (${autoWindowPeriod})`
-      : AGG_WINDOW_AUTO
+    durationDisplay = AGG_WINDOW_AUTO
+    if (autoWindowPeriod) {
+      durationDisplay = `${AGG_WINDOW_AUTO} (${autoWindowPeriod})`
+      if (isInCheckOverlay) {
+        durationDisplay = `${AGG_WINDOW_AUTO} (${everyWindowPeriod})`
+      }
+    }
   }
```
jrenee42, line 59: i will look at this one again in the morning...... (TODO)

jrenee42 (changes requested):

cypress/e2e/shared/telegrafs.test.ts
hoorayimhelping: >why not use the inbuilt string localeCompare?

@jrenee42 I don't want to mix the meaning of this change. I want this change to modify the structure of the comparison without changing the meaning. I think that `localeCompare` is probably the right way to go here, but changing somewhat complicated nested ternaries is a risky enough change that I'm a little nervous changing how the comparison is done

Palakp41 (approved):
LGTM

src/timeMachine/components/WindowPeriod.tsx
jrenee42: just rechecked this.  it's good

jrenee42 (approved):

PR #961, opened 2021-03-24
feat: add yml definition for the /me endpoint

### LINKED ISSUE #960: /me endpoint for Quartz API
Develop an `/me` API endpoint swagger definition for the Quartz API that will return relevant information such as:

- account type: free/payg/cancelled
- is it a marketplace account? If so, which marketplace?
- is the org in a beta region?
- is the current user an operator?
- maybe the user’s full name if you need that anywhere
- orgID
- accountID

Asalem1:
Closes #960 

This PR is an attempt to write out an API definition for the `/me` endpoint that will be developed by the Quartz API

### DIFF
--- a/src/client/swagger/unity.yml
+++ b/src/client/swagger/unity.yml
@@ -947,6 +947,8 @@ components:
           type: string
         orgId:
           type: string
+        accountId:
+          type: string
         isBeta:
           type: boolean
           description: whether the region associated with the account is a beta region
@@ -965,15 +967,16 @@ components:
             - pay_as_you_go
       required:
         [
-          type,
-          marketplace,
-          isOperator,
-          isBeta,
-          orgId,
+          id,
+          accountId,
           email,
           firstName,
           lastName,
-          id,
+          isOperator,
+          isBeta,
+          orgId,
+          marketplace,
+          type,
         ]
     Users:
       type: object
--- a/src/client/swagger/unity.yml
+++ b/src/client/swagger/unity.yml
@@ -23,7 +23,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/me/{id}':
+  '/me':
     get:
       operationId: GetMe
       tags:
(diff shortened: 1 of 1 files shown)


src/client/swagger/unity.yml
```diff
@@ -23,6 +23,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  '/me/{id}':
```
randycoulman: Do we need an `id` here? Can we assume that this will be for the currently-authenticated user using the userID/orgID in the session?

gunnaraasen, line 26: Could the endpoint be `/me` with no ID? I think we would already know the login user ID from the cookie on the request.

Asalem1: Definitely, if that's the case then for sure

src/client/swagger/unity.yml
```diff
@@ -905,6 +934,47 @@ components:
...
+          description: the idpe id of the user
+          type: string
+        firstName:
+          type: string
+        lastName:
+          type: string
+        email:
+          type: string
+        orgId:
+          type: string
```
randycoulman, line 949: I'm torn about whether we need `id` and `orgId` in this payload since we have them in the session already. But it's probably easier in the UI to have them here. In that case, should we also include `accountId` so we have a full set?

randycoulman (approved):
A couple of suggestions below, but LGTM!

src/client/swagger/unity.yml
Asalem1: Nah, I cut it out after Gunnar's suggestion about using the cookie as an identifier

src/client/swagger/unity.yml
Asalem1: We don't necessarily need the orgID here since we can obtain that through other means, I just wasn't sure about where / when the API would be used so I'm trying to keep it as dynamic as possible. Good call on the accountId

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=961)

ischolten (approved):
LGTM

PR #969, opened 2021-03-24
fix: format times as times for humans

### LINKED ISSUE #887: i can't understand the time displayed in the simple table
The timestamp format for the current simple table is in unixtime, which is difficult for humans to understand:
image

Let's format this timestamp using the same format returned from the flux query, in UTC timezone.

`2021-03-23T18:00:00Z`

drdelambre:
Closes #887 

you mean you guys can't read unix epoch?

looks like this after fix:
Screenshot from 2021-03-24 13-16-10


Asalem1 (approved):
hoomans*

PR #971, opened 2021-03-24
feat: marketplace UI integrated for the billing page

Asalem1:
Closes https://github.com/influxdata/idpe/issues/9970

This PR adds a billing page for PAYG Marketplace users based on the Quartz specifications

Screen Shot 2021-03-24 at 2 28 57 PM

### DIFF
--- a/src/billing/api/index.ts
+++ b/src/billing/api/index.ts
@@ -32,7 +32,11 @@ const makeResponse = (status, data, ...args) => {
 export const getAccount = (): ReturnType<typeof apiGetAccount> => {
   const account: Account = {
     id: 'account_1',
-    marketplace: 'gcp',
+    marketplace: {
+      name: 'Amazon Web Services',
+      shortName: 'aws',
+      url: 'smile.amazon.com',
+    },
     type: 'pay_as_you_go',
     status: RemoteDataState.Done,
   }
--- a/src/billing/api/index.ts
+++ b/src/billing/api/index.ts
@@ -32,8 +32,8 @@ const makeResponse = (status, data, ...args) => {
 export const getAccount = (): ReturnType<typeof apiGetAccount> => {
   const account: Account = {
     id: 'account_1',
-    marketplace: 'AWS',
-    type: 'free',
+    marketplace: 'gcp',
+    type: 'pay_as_you_go',
     status: RemoteDataState.Done,
   }
   return makeResponse(200, account, 'getBillingAccount')
--- a/src/billing/api/index.ts
+++ b/src/billing/api/index.ts
@@ -32,12 +32,8 @@ const makeResponse = (status, data, ...args) => {
 export const getAccount = (): ReturnType<typeof apiGetAccount> => {
   const account: Account = {
     id: 'account_1',
-    marketplace: {
-      name: 'Amazon Web Services',
-      shortName: 'aws',
-      url: 'smile.amazon.com',
-    },
-    type: 'pay_as_you_go',
+    marketplace: null,
+    type: 'free',
     status: RemoteDataState.Done,
   }
   return makeResponse(200, account, 'getBillingAccount')
--- a/src/billing/components/BillingPageContents.tsx
+++ b/src/billing/components/BillingPageContents.tsx
@@ -4,12 +4,17 @@ import React, {FC} from 'react'
 // Components
 import BillingFree from 'src/billing/components/Free/Free'
 import BillingPayAsYouGo from 'src/billing/components/PayAsYouGo/PayAsYouGo'
+import MarketplaceBilling from 'src/billing/components/marketplace/MarketplaceBilling'
 
 // Utils
 import {useBilling} from 'src/billing/components/BillingPage'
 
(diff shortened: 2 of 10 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=971)

src/billing/api/index.ts
```diff
@@ -32,8 +32,8 @@ const makeResponse = (status, data, ...args) => {
 export const getAccount = (): ReturnType<typeof apiGetAccount> => {
   const account: Account = {
     id: 'account_1',
-    marketplace: 'AWS',
-    type: 'free',
+    marketplace: 'gcp',
```
ischolten, line 35: Btw valid marketplaces are `aws`, `gcm`, and `azure`.

src/billing/components/BillingPageContents.tsx
```diff
@@ -4,13 +4,18 @@ import React, {FC} from 'react'
...
 import BillingPayAsYouGo from 'src/billing/components/PayAsYouGo/PayAsYouGo'
+import MarketplaceBilling from 'src/billing/components/marketplace/MarketplaceBilling'
 
 // Utils
 import {useBilling} from 'src/billing/components/BillingPage'
 
 const BillingPageContents: FC = () => {
   const [{account}] = useBilling()
 
+  if (account.type === 'pay_as_you_go' && account.marketplace !== 'Zuora') {
```
ischolten, line 15: this will likely need to be `&& !!account.marketplace` because zuora is not a marketplace

randycoulman, line 15: I think this is OK, but in Quartz, we only look at whether the account has a marketplace subscription or not. I think the equivalent here would be `account.marketplace !== 'Zuora'` without the `pay_as_you_go` check.

src/types/marketplace.ts
```diff
@@ -0,0 +1 @@
+export type Marketplace = 'gcm' | 'aws' | 'azure'
```
randycoulman, line 1: Above, we're looking for `account.marketplace !== 'Zuora'`, but here `Zuora` isn't an option.  And in the spec we looked at for `/me`, we said that `marketplace` can be `null`.

Let's figure out which direction we want to go and be consistent with it. Otherwise, we're gonna get a `null` marketplace and try to compare that with `Zuora` and bad things will happen.

src/billing/components/marketplace/MarketplaceBilling.tsx
```diff
@@ -0,0 +1,36 @@
...
+  Panel,
+  ComponentSize,
+  FlexBox,
+  FlexDirection,
+  AlignItems,
+  JustifyContent,
+} from '@influxdata/clockface'
+
+import MarketplaceLink from 'src/billing/components/marketplace/MarketplaceLink'
+import PurplePanel from 'src/billing/components/marketplace/PurplePanel'
```
randycoulman, line 13: #nit I know you copied this straight from Quartz, but is there a better name than `PurplePanel` for this? Is it even purple any more? :-)

src/billing/components/marketplace/MarketplaceLink.tsx
```diff
@@ -0,0 +1,70 @@
...
+const marketplaceLinkText = marketplace => {
+  switch (marketplace) {
+    case 'aws':
+      return 'AWS Subscription Portal'
+    case 'gcp':
+      return 'Google Cloud Solution'
+    case 'azure':
+      return 'Microsoft Azure Portal'
+  }
+}
```
randycoulman, line 46: #optional Again, I know this is straight from Quartz, but would it be worth combining these two functions into a single case statement with a `linkText` key in the resulting object?

randycoulman (approved):
LGTM! A couple of optional things that are probably out of scope.

I think we need to address whether a non-marketplace account will be listed as `null` or `Zuora`, though, and make sure our API spec and all of the code is consistent in what it uses.  That can be done in a follow-up if it doesn't make sense to do here.

src/billing/api/index.ts
Asalem1: 👍🏽 I'm gonna add a type to the yml to track those changes

src/billing/components/BillingPageContents.tsx
Asalem1: Good to know, thanks!

Asalem1: nvm, answered my own question by searching the `yml`

src/billing/components/marketplace/MarketplaceLink.tsx
Asalem1: Sounds good!

randycoulman (approved):
Changes LGTM as long as design is OK with the wording changes (getting rid of the `Check My ...` stuff).

Issue #900, opened 2021-03-18
Annotations: Delete Annotation


hoorayimhelping:
### Background
As a user, I want to be able to delete an annotation that is wrong or no longer relevant

See #896 #899

![6. Edit Annotation Overlay.png](https://images.zenhubusercontent.com/5c33b08d24dda4712ef4ae10/9a166c60-555f-4a95-9e65-cd6259723a0c)

### Acceptance Criteria
- The delete Annotation functionality is hooked up to the Delete button on the above mockup

Issue #896, opened 2021-03-18
Annotations: On Annotation Click


hoorayimhelping:
### Background:
We need the ability to interact with the annotation. Clicking on it should launch a modal:

See #899 #900
![6. Edit Annotation Overlay.png](https://images.zenhubusercontent.com/5c33b08d24dda4712ef4ae10/769da93d-a983-4952-8eb4-10f7223d0706)

### Acceptance Criteria
- ~Clicking on a spot in the graph that is already occupied by an annotation will launch a dialog similar to the one in the screenshot above~ 
- Annotations will have a marker / target for easy clicking

PR #976, opened 2021-03-24
feat: working edit and delete annotations

### LINKED ISSUE #899: Annotations: Edit Annotation
### Background
As a user, I want to be able to modify the Summary and time of an annotation so I can correct mistakes

See #896, #900

Here is the original mockup. Obviously we only have two fields to edit, so the mockup should reflect that.

### Acceptance Criteria
**NOTE: These criteria do not need to be met in a single PR. I'd suggest doing the backend and redux work first, PR it, and merge it in, then do a separate bit of work to connect the UI.**

- Annotations are able to edited via the API
- Any necessary redux changes needed to enable editing are done
- Only the fields we allow changing (i.e. Start Date and Summary) are present

ShmuelLotman:
Closes #899, #900 

This PR introduces the edit and delete abilities to annotations, working off the edit annotation modal built in #896. Heads up: this will require some edits for typings, cleanup for some naming that may not be optimal and obviously tests, but the functionality should be there now, as well as toasts. Lchaim!


src/annotations/components/AddAnnotationOverlay.tsx
```diff
@@ -16,11 +16,8 @@ export const AddAnnotationOverlay: FC = () => {
   const {createAnnotation, startTime} = useSelector(getOverlayParams)
 
   const handleSubmit = (modifiedAnnotation): void => {
-    const formIsValid = true
```
hoorayimhelping, line 19: Let's not remove this here or in the other place. Instead, the fix is not to remove this, but to hook up `formIsValid` to a real validation function: https://github.com/influxdata/ui/issues/980

src/annotations/actions/thunks.ts
```diff
@@ -32,15 +42,44 @@ export const fetchAndSetAnnotations = () => async (
...
+) => async (
+  dispatch: Dispatch<AnnotationAction | NotificationAction>
+): Promise<void> => {
+  try {
+    await writeAnnotation(annotations)
 
-  fetchAndSetAnnotations()(dispatch)
+    fetchAndSetAnnotations()(dispatch)
+  } catch (err) {
+    dispatch(notify(createAnnotationFailed(err)))
```
hoorayimhelping, line 53: i think this makes more sense one layer up, in the UI / component layer. what tends to happen is that you start out with a generic error handler that applies to all calls to this `writeThenFetchAndSetAnnotations`, and then as you run into errors in different places this function is called, you start having to tweak this 'generic' function until you have cases that are specific to each use of the function. then you conclude the reason it's so janky is because the responsibility actually lies in the UI layer and that each call of it probably represents a different use case in the UI. So let's just shortcut all that pain and remove this here, and don't catch the error and let it bubble up. Then, where this is called becomes something along the lines of (I haven't tested this so it may not be 100% correct):

```ts
const makeSingleClickHandler = () => {
  const createAnnotation = userModifiedAnnotation => {
    const {message, startTime} = userModifiedAnnotation
    try {
      dispatch(
        writeThenFetchAndSetAnnotations([
          {
            summary: message,
            startTime: new Date(startTime).getTime(),
            endTime: new Date(startTime).getTime(),
          },
        ])
      )
      event('xyplot.annotations.create_annotation.create')
    } catch (err) {
      dispatch(notify(createAnnotationFailed(err)))
      event('xyplot.annotations.create_annotation.failure')
    }
  }
```

this way, we can dispatch an error message and handle an annotation creation failure metric call in the same place, which to me seems to be a strong indicator this is the proper layer for this logic

src/annotations/reducers/index.ts
```diff
@@ -69,6 +71,30 @@ export const annotationsReducer = (
         ),
       }
     }
+    case DELETE_ANNOTATION: {
+      return {
+        ...state,
+        annotations: {
+          default: state.annotations['default'].filter(
+            a => a.id !== action.annotation.id
```
hoorayimhelping, line 79:
```suggestion
            annotation => annotation.id !== action.annotation.id
```

src/annotations/components/EditAnnotationForm.tsx
```diff
@@ -63,8 +67,13 @@ export const EditAnnotationForm: FC<EditAnnotationProps> = ({
     }))
   }
 
-  const annotationStreams = useSelector(getAnnotationStreams)
+  const handleDeleteAnnotation = () => {
+    dispatch(deleteAnnotations(editAnnotationState))
```
hoorayimhelping, line 71: can you add two metric calls here (one for success, and one for failure, when you move the failure calls up to this layer), and then connect this PR to https://github.com/influxdata/ui/issues/728 and check the one check box related to deleting annotations please?

hoorayimhelping (changes requested):
really solid work! this looks very clean and well organized. have a couple suggestions, one main one is for where i think the notification logic belongs and the related metric calls to it, and for adding back in the fake validation functions

hercules-influx:
Metric | Score | Level
performance | 35 | :red_circle:
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=976)

src/annotations/actions/thunks.ts
```diff
@@ -2,17 +2,19 @@ import {
...
+  updateAnnotation,
   writeAnnotation,
 } from 'src/annotations/api'
 import {Dispatch} from 'react'
-import {deleteAnnotation as deleteAnnotationAction} from 'src/annotations/actions/creators'
 import {
   setAnnotations,
   setAnnotationStreams,
   Action as AnnotationAction,
+  editAnnotation as patchAnnotation,
```
hoorayimhelping, line 13: the lack of consistency between these two aliases is kind of confusing, since one is called `patchAnnotation` which is kind of an alias for edit, and one is called `deleteAnnotationAction`. it's not at all clear that they're the same thing later in the code.

just make this `editAnnotationAction` for consistency's sake

hoorayimhelping (approved):
lgtm, just update that one variable name

PR #977, opened 2021-03-25
fix: map getLatLon function returns correct error

### LINKED ISSUE #963: Map error when s2_cell_id missing
Expected Behavior:
Given I am user viewing data on Data Explorer, when I provide data with Lat Lon as fields, I expect that data to be used to load map and map points.

Current Behavior:
When I provide data with Lat Lon as fields, I get an error when loading map that I am missing those fields. The code is depended on a column `s2_cell_id` to get the coordinates and we need to remove that dependency.

Palakp41:
Closes #963 

Maps in Data Explorer breaks if the user provides lat lon data without `s2_cell_id`. This change is so that our check if coordinates exist is no longer dependent on `s2_cell_id`.

### DIFF
--- a/src/shared/utils/vis.ts
+++ b/src/shared/utils/vis.ts
@@ -1,5 +1,4 @@
 // Libraries
-import {S2} from 's2-geometry'
 import {Table, LineInterpolation, FromFluxResult} from '@influxdata/giraffe'
 
 // Types
@@ -335,7 +334,7 @@ export const getGeoCoordinates = (table: Table) => {
   const lonCoordinate = getColumnValue(table, 'lon')
 
   return {
-    lat: parseInt(latCoordinate.toString()),
-    lon: parseInt(lonCoordinate.toString()),
+    lat: parseInt(latCoordinate.toString(), 10),
+    lon: parseInt(lonCoordinate.toString(), 10),
   }
 }
--- a/src/shared/utils/vis.ts
+++ b/src/shared/utils/vis.ts
@@ -315,54 +315,27 @@ export const getMainColumnName = (
   return ''
 }
 
-const getS2CellID = (table: Table, index: number): string => {
-  const column = table.getColumn('s2_cell_id')
-  if (!column) {
+const getColumnValue = (table: Table, field: string) => {
+  const fieldColumn = table.getColumn('_field')
+
+  if (!fieldColumn) {
     throw new Error(
-      'Cannot retrieve s2_cell_id column - table does not conform to required structure of Table type'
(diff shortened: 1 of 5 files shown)


src/shared/utils/vis.ts
```diff
@@ -315,54 +315,27 @@ export const getMainColumnName = (
...
-  const geoCoordinateValue = S2.idToLatLng(fixed.toString())
+export const getGeoCoordinates = (table: Table) => {
+  const latCoordinate = getColumnValue(table, 'lat')
+  const lonCoordinate = getColumnValue(table, 'lon')
 
   return {
-    lat: geoCoordinateValue.lat,
-    lon: geoCoordinateValue.lng,
+    lat: parseInt(latCoordinate.toString()),
+    lon: parseInt(lonCoordinate.toString()),
```
hoorayimhelping, line 339:
```suggestion
    lat: parseInt(latCoordinate.toString(), 10),
    lon: parseInt(lonCoordinate.toString(), 10),
```

you generally always want to include a radix when calling `parseInt` because different browsers might implement it differently without one.

https://stackoverflow.com/questions/6611824/why-do-we-need-to-use-radix-parameter-when-calling-parseint

src/shared/utils/vis.ts
```diff
@@ -315,54 +315,27 @@ export const getMainColumnName = (
...
+const getColumnValue = (table: Table, field: string) => {
+  const fieldColumn = table.getColumn('_field')
+
+  if (!fieldColumn) {
     throw new Error(
-      'Cannot retrieve s2_cell_id column - table does not conform to required structure of Table type'
+      'Cannot retrieve _field column - table does not conform to required structure of Table type'
     )
   }
+  const index = fieldColumn.findIndex(val => val === field)
```
hoorayimhelping, line 326: nit: `val` is kind of vague as a name. not a huge deal since it's kind of clear what's happening but just an area for small improvement

hoorayimhelping (approved):
lgtm, once that radix is added

src/shared/utils/vis.ts
Palakp41: good idea, I updated it!

hoorayimhelping (approved):

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=977)

hoorayimhelping (approved):

PR #979, opened 2021-03-25
chore: extract zuora form out of checkout

### LINKED ISSUE #870: Integrate the Zuora Client to the UI & add tests
Currently the Zuora Client is commented out in most of the UI until the API is setup and we can obtain the proper credentials from the API. 

Once the API is ready, we need to add the Zuora Client to the UI and add some tests to ensure that everything works properly

subirjolly:
Closes #870

### DIFF
--- a/cypress/e2e/cloud/checkout.test.ts
+++ b/cypress/e2e/cloud/checkout.test.ts
@@ -57,9 +57,12 @@ describe('Checkout Page', () => {
     cy.getByTestID('checkout-upgrade--button').click()
 
     // Check all errors are visible
+    cy.getByTestID('balanceThreshold--input').scrollIntoView()
+    cy.getByTestID('balanceThreshold--form-element-error').should('be.visible')
     cy.getByTestID('balanceThreshold--form-element-error').contains(
       genericError
     )
+    cy.getByTestID('notifyEmail--form-element-error').should('be.visible')
     cy.getByTestID('notifyEmail--form-element-error').contains(genericError)
 
     // Check balance threshold specific error should exist
@@ -103,7 +106,9 @@ describe('Checkout Page', () => {
 
     cy.getByTestID('checkout-upgrade--button').click()
 
+    cy.getByTestID('city--form-element-error').should('be.visible')
     cy.getByTestID('city--form-element-error').contains(error)
+    cy.getByTestID('postalCode--form-element-error').should('be.visible')
     cy.getByTestID('postalCode--form-element-error').contains(error)
 
     cy.getByTestID('city--input').type('Blacksburg')
@@ -148,6 +153,7 @@ describe('Checkout Page', () => {
       cy.getByTestID('checkout-upgrade--button').click()
 
       // Check required fields show error
+      cy.getByTestID('city--form-element-error').should('be.visible')
       cy.getByTestID('city--form-element-error').contains(error)
 
       cy.getByTestID('city--input')
--- a/cypress/e2e/cloud/checkout.test.ts
+++ b/cypress/e2e/cloud/checkout.test.ts
@@ -57,18 +57,15 @@ describe('Checkout Page', () => {
     cy.getByTestID('checkout-upgrade--button').click()
 
     // Check all errors are visible
-    cy.getByTestID('balanceThreshold--form-element-error').should('be.visible')
     cy.getByTestID('balanceThreshold--form-element-error').contains(
       genericError
     )
-    cy.getByTestID('notifyEmail--form-element-error').should('be.visible')
     cy.getByTestID('notifyEmail--form-element-error').contains(genericError)
 
     // Check balance threshold specific error should exist
     cy.getByTestID('balanceThreshold--input')
       .clear()
       .type('0')
-    cy.getByTestID('balanceThreshold--form-element-error').should('be.visible')
     cy.getByTestID('balanceThreshold--form-element-error').contains(numberError)
 
     cy.getByTestID('notifyEmail--input')
--- a/src/billing/components/PaymentInfo/PaymentPanelBody.tsx
+++ b/src/billing/components/PaymentInfo/PaymentPanelBody.tsx
@@ -43,7 +43,6 @@ const PaymentPanelBody: FC<Props> = ({isEditing, onCancel}) => {
   if (isEditing) {
     return (
       <Panel.Body size={ComponentSize.Large}>
-        {/* FIXME: Add zuoraParams functionality */}
         <PaymentForm
           zuoraParams={{} as CreditCardParams}
           onSubmit={onSubmit}
--- a/src/billing/components/PaymentInfo/PaymentPanelBody.tsx
+++ b/src/billing/components/PaymentInfo/PaymentPanelBody.tsx
@@ -14,7 +14,7 @@ import {
   getBillingCreditCardParams,
   putBillingPaymentMethodId,
 } from 'src/billing/api'
-import { getErrorMessage } from 'src/utils/api'
+import {getErrorMessage} from 'src/utils/api'
 
 interface Props {
   isEditing: boolean
--- a/src/billing/components/PaymentInfo/PaymentPanelBody.tsx
+++ b/src/billing/components/PaymentInfo/PaymentPanelBody.tsx
@@ -14,6 +14,7 @@ import {
   getBillingCreditCardParams,
   putBillingPaymentMethodId,
 } from 'src/billing/api'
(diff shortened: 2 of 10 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=979)

src/billing/components/PaymentInfo/PaymentPanelBody.tsx
```diff
@@ -1,51 +1,75 @@
...
 const PaymentPanelBody: FC<Props> = ({isEditing, onCancel}) => {
   const [errorMessage, setErrorMessage] = useState('')
+  const [creditCardParams, setCreditCardParams] = useState(
+    EMPTY_CREDIT_CARD_PARAMS
+  )
+
+  const onSubmit = async (paymentMethodId: string): Promise<void> => {
+    const response = await putBillingPaymentMethodId(paymentMethodId)
+    if (response.status !== 200) {
+      const error = response.data
```
Asalem1, line 44: we might want to use the `getErrorMessage` utility function for this

src/shared/components/CreditCardForm/index.test.tsx
```diff
@@ -98,7 +98,7 @@ describe('CheckoutV2.ZuoraPaymentForm', () => {
   //     const latestSuccessMock = jest.fn()
 
   //     rerender(
-  //       <ZuoraPaymentForm
+  //       <CreditCardForm
```
Asalem1, line 101: do we still need this test?

src/billing/components/PaymentInfo/PaymentPanelBody.tsx
Asalem1: https://github.com/influxdata/ui/blob/master/src/utils/api.ts#L8

cypress/e2e/cloud/checkout.test.ts
```diff
@@ -57,18 +57,15 @@ describe('Checkout Page', () => {
...
       genericError
     )
-    cy.getByTestID('notifyEmail--form-element-error').should('be.visible')
     cy.getByTestID('notifyEmail--form-element-error').contains(genericError)
 
     // Check balance threshold specific error should exist
     cy.getByTestID('balanceThreshold--input')
       .clear()
       .type('0')
-    cy.getByTestID('balanceThreshold--form-element-error').should('be.visible')
```
Asalem1, line 71: Just so I'm clear, these are being removed because the following statement makes it implicit in that the error codes should not be visible right? If that's the case, maybe we should test that the error state is not visible on initialization (i.e. before clicking on the submit button and seeing all the errors, maybe we should first test to see that there is no error state to begin with)

src/billing/components/PaymentInfo/PaymentPanelBody.tsx
```diff
@@ -1,51 +1,72 @@
...
-        setErrorMessage(error)
-      }
+  const onSubmit = async (paymentMethodId: string): Promise<void> => {
+    const response = await putBillingPaymentMethodId(paymentMethodId)
+    if (response.status !== 200) {
+      setErrorMessage(getErrorMessage(response))
     } else {
-      setErrorMessage(error)
+      onCancel()
+      setErrorMessage('')
```
Asalem1, line 48: We should probably add some sort of notification saying that there was an issue if updating the payment method fails

src/shared/components/CreditCardForm/index.tsx
```diff
@@ -1,20 +1,32 @@
...
-  // FIXME: Add onFocus functionality
-  const {zuoraParams, handleSubmit, inputs} = useContext(CheckoutContext)
+export interface Props {
+  zuoraParams: CreditCardParams
+  onSubmit?: (paymentMethodId) => void
+  zuoraClient?: ZuoraClient
+}
 
-  const [client, setClient] = useState(window.Z)
+// FIXME: Add onFocus functionality
```
Asalem1, line 17: Let's get this in as the next PR since the cart abandonment work should be setup for it

src/shared/components/CreditCardForm/index.tsx
```diff
@@ -48,16 +66,12 @@ const ZuoraPaymentForm: FC = () => {
         setClient(window.Z)
       }
     }
-
-    return () => {
-      client && document.body.removeChild(document.getElementById(scriptId))
-    }
```
Asalem1, line 54: Why are we removing this?

src/shared/components/CreditCardForm/index.tsx
```diff
@@ -1,20 +1,32 @@
...
+export interface Props {
+  zuoraParams: CreditCardParams
+  onSubmit?: (paymentMethodId) => void
+  zuoraClient?: ZuoraClient
+}
 
-  const [client, setClient] = useState(window.Z)
+// FIXME: Add onFocus functionality
+const CreditCardForm: FC<Props> = ({zuoraParams, onSubmit, zuoraClient}) => {
+  const _isMounted = useRef(true)
```
Asalem1, line 19: 🤔 I'm a little fuzzy on `useRef` so my apologies for asking dumb questions - does `current` get set here on initialization? Also, what's the purpose of setting `isMounted` with a useRef here? I don't see it being set on a DOM node

cypress/e2e/cloud/checkout.test.ts
subirjolly: For some reason, `.should('be.visible')` has started failing. That's why I removed it.

src/billing/components/PaymentInfo/PaymentPanelBody.tsx
subirjolly: This is basically clearing any messages previously displayed. The error is being displayed on line 45.

src/shared/components/CreditCardForm/index.tsx
subirjolly: sounds good!

cypress/e2e/cloud/checkout.test.ts
Asalem1: 🤔  have you tried testing it locally to see why it's failing? Are there links to any artifacts from those failures? It could be that integrating Zuora has some how messed with the way the page loads and we should consider that in our tests

src/shared/components/CreditCardForm/index.tsx
subirjolly: This is causing the functionality to fail if you click on `Change Payment` multiple times. It's unable to find the script in the body but it apparently is getting added and the functionality exists on dom.

src/billing/components/PaymentInfo/PaymentPanelBody.tsx
Asalem1: Aw, i see what you're saying

cypress/e2e/cloud/checkout.test.ts
subirjolly: I have tested it locally and the span is there but cypress's failing to see it. I can dig into it.

Asalem1: That would be 🔥  since it's strange that the Zuora client would cause a previously known good test to fail. It could be an issue of Zuora taking time to load, in which case the entire page takes a while and causes some issues around there? Not sure

src/shared/components/CreditCardForm/index.tsx
subirjolly: Yea, so the tests revealed that the request were made even after the form has been unmounted. This was causing memory leak as well as forcing tests to fail. The `current` for `_isMounted` gets set as `true` and when the component gets unmounted, it's set to `false`.

Asalem1: Gotcha, I'm just trying to think of whether there's a clearer way of doing this that doesn't make use of `useRef` since we might find ourselves in a situation where we would need to list to changes made on the DOM. We could probably achieve the same thing by simply using `useState` and initializing a value in `useEffect`, etc... 

It should do the same thing, but I'm wary of `useRef` because of the way it propagates events. According to the [docs](https://reactjs.org/docs/hooks-reference.html#useref):

`Keep in mind that useRef doesn’t notify you when its content changes. `

Asalem1 (approved):
🤞🏽 the `useRef` works like it should

PR #991, opened 2021-03-26
fix: remove old templates code in task

### LINKED ISSUE #869: Create task from a template doesn't recognize if you have any templates
In tasks, if you select the option to "Create task from a template", it always says "Looks like you don't have any templates yet. Why don't you import one?". This message appears even if I have templates, and even if the templates contain tasks.

Also, when I import a template, it's not clear what I am supposed to do next.  I selected a task in the template, and ended up in the task editor for that task.  I don't know how to get back to the screen where I was creating the task.

zoesteinkamp:
Closes #869

We are removing the old template uploader in tasks. We no longer use this code and it does not make sense in the context of community templates


hercules-influx:
Metric | Score | Level
performance | 38 | :red_circle:
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=991)

hoorayimhelping (approved):

PR #1010, opened 2021-03-29
feat: UI-871 Operator Page UI

### LINKED ISSUE #871: Convert the Operator UI from the Quartz to UI
This is intentionally vague since I'm fuzzy on the details of where this will be, but the Operator UI needs to be transferred from Quartz to the UI

Asalem1:
Closes #871 

This PR integrates the UI aspect of the Operator with some assumptions about the API. Having said that, there are some key issues that need to be resolved in order to fully integrate the operator page into the UI since there are currently some cyclical dependencies between Accounts and Organizations that should be resolved / addressed in order to move forward. 

operator-ui


src/authorizations/components/TokenList.tsx
```diff
@@ -1,6 +1,5 @@
 // Libraries
 import React, {PureComponent} from 'react'
-import _ from 'lodash'
 import memoizeOne from 'memoize-one'
```
Asalem1, line 3: Unused import that I removed

src/buckets/components/BucketCard.tsx
```diff
@@ -1,7 +1,6 @@
 // Libraries
 import React, {FC} from 'react'
 import {withRouter, RouteComponentProps} from 'react-router-dom'
-import _ from 'lodash'
```
Asalem1, line 4: unused import that was removed

src/buckets/components/RenameBucketOverlay.tsx
```diff
@@ -2,8 +2,6 @@
 import React, {PureComponent} from 'react'
 import {withRouter, RouteComponentProps} from 'react-router-dom'
 
-import _ from 'lodash'
-
```
Asalem1, line 6: unused import that was removed

src/client/swagger/unity.yml
```diff
@@ -659,7 +845,29 @@ components:
             - free
             - cancelled
             - pay_as_you_go
-      required: [id, marketplace, type]
+        # TODO(ariel): this seems cyclical since accounts and orgs are subsets of one another
```
Asalem1, line 848: Would love some thoughts on this to see what makes the most sense for the API

src/client/swagger/unity.yml
```diff
@@ -1012,6 +1241,46 @@ components:
...
+        region:
+          type: string
+          description: region of the org
+        provider:
+          type: string
+          description: provider of the org
+        date:
+          type: string
+          description: date org was created
+        # TODO(ariel): this seems cyclical since accounts and orgs are subsets of one another
```
Asalem1, line 1276: Same as above

Asalem1 (commented):
I should also mention that this PR touches some extra files dealing with lodash to remove unused imports and make it clear what's being used when lodash is integrated. It's an unrelated change that I got caught in on this PR

src/operator/LimitsInput.tsx
```diff
@@ -0,0 +1,30 @@
...
+      type === InputType.Number ? parseFloat(e.target.value) : e.target.value
+
+    const newLimits = {...limits, [name]: newValue}
+    onChangeLimits(newLimits)
+  }
+  return (
+    <Input
+      type={type}
+      name={name}
+      value={limits?.[name] ?? ''}
```
randycoulman, line 24: This code uses lodash's `get` and `set` here because `name` is generally a dotted name like `rate.readKBs` or similar. I don't believe that this code (and line 17 above) will do the same thing in that case.

src/operator/OrgOverlay.tsx
```diff
@@ -0,0 +1,242 @@
...
+                spinnerComponent={<TechnoSpinner diameterPixels={100} />}
+                testID="limits-spinner-container"
+              >
+                <Grid.Row>
+                  <h4>Limits</h4>
+                  <Grid.Column widthMD={3}>
+                    <Form.Label label="Read (KBs)" testID="read-kbs" />
+                    <LimitsInput
+                      type={InputType.Number}
+                      name="rate.readKBs"
```
randycoulman, line 109: As mentioned above, I don't believe that these `name` props will work without using lodash's `get` and `set` in `LimitsInput`.

src/operator/OrgOverlay.tsx
```diff
@@ -0,0 +1,242 @@
...
+          title={orgID}
+          style={{color: '#FFFFFF'}}
+          onDismiss={() => history.goBack()}
+        />
+        <SpinnerContainer
+          loading={orgStatus}
+          spinnerComponent={<TechnoSpinner diameterPixels={100} />}
+        >
+          <Overlay.Body>
+            <Grid>
```
randycoulman, line 75: The Quartz version of this component has an `Alert` being shown in case of a request error. How is that handled here?

src/operator/ResourcesSearchbar.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const changeSearchTerm = (event: ChangeEvent<HTMLInputElement>) => {
+    setSearchTerm(event.target.value)
+  }
+
+  return (
+    <Input
+      icon={IconFont.Search}
+      placeholder={`Filter ${activeTab} by ${
+        activeTab === 'accounts' ? 'email' : 'id'
+      } ...`}
```
randycoulman, line 18: This is going to have to be updated when we add a `Users` page later. The Quartz version of this injects a placeholder from the caller; can we keep that design here to avoid having this conditional grow in complexity over time?

I think that restoring the top-level routes will force separate top-level `Accounts` and `Organizations` pages which would then be a reasonable place for this prop to be set.

src/operator/ResourcesTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+  )
+
+  // TODO(ariel): remove type forcing here, this should be resolved with
+  // API integration
+  const resources: Account[] | Organization[] =
+    activeTab === 'accounts' ? accounts : organizations
+  const headers =
+    activeTab === 'accounts' ? accountHeaderInfo : organizationColumnHeaders
+  const infos =
+    activeTab === 'accounts' ? accountColumnInfo : organizationColumnInfo
```
randycoulman, line 41: These are more conditionals that will have to change when we add the `Users` page back in.

src/operator/account/AssociatedUsersTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <EmptyState.Text>
+            No users. If this account is no longer being used, please delete it.
+          </EmptyState.Text>
+        </EmptyState>
+      )}
+    </FlexBox>
+  )
+}
+
+export default AssociatedTableUsers
```
randycoulman, line 60: I like pulling out these two separate components instead of the `AccountViewTable` in Quartz that has some weird conditional logic in it. To be fair, I think those conditionals were only added recently.

Not sure what to do with the rest of the duplication, though.

src/operator/api/index.ts
```diff
@@ -0,0 +1,408 @@
...
+      lastName: 'Salem',
+      country: 'USA',
+      street1: '123 Main St',
+      city: 'New York',
+      subdivision: 'NY',
+      postalCode: 30000,
+    },
+    users: [],
+    type: 'pay_as_you_go',
+  }
```
randycoulman, line 270: This looks like a much larger payload than I expected from `/me`, but I guess that change to the swagger file was in a previous PR? Do we need all of this here?

src/operator/constants.tsx
```diff
@@ -0,0 +1,289 @@
...
+    defaultValue: '',
+  },
+  {
+    path: 'date',
+    name: 'date',
+    defaultValue: '',
+  },
+]
+
+export const userColumnInfo: CellInfo[] = [
```
randycoulman, line 119: This one looks like leftover cruft from the partially-implemented Users page we had in here at one point. Optional whether we keep it or delete it until we bring back the Users page.

src/operator/constants.tsx
```diff
@@ -0,0 +1,289 @@
...
+        : `${value[0]}, ${value[1]} ${value[2]}`
+    },
+  },
+]
+
+export const OperatorRoutes = {
+  accounts: '/operator/accounts',
+  organizations: '/operator/organizations',
+  users: '/operator/users',
+}
```
randycoulman, line 289: Are these used anywhere? As mentioned above, I think they should be because I think we want these as separate routes.

src/operator/context/account.tsx
```diff
@@ -0,0 +1,138 @@
...
+        setAccountStatus(RemoteDataState.Error)
+        throw new Error(resp.data.message)
+      }
+
+      setAccount(resp.data)
+      setAccountStatus(RemoteDataState.Done)
+    } catch (error) {
+      setAccountStatus(RemoteDataState.Error)
+      console.error({error})
+      dispatch(notify(getAccountError(accountID)))
```
randycoulman, line 77: Interesting! So on a non-200 response, we throw an Error, but here we eat the error and dispatch a notification instead?  I'm sure this is a standard pattern in the UI codebase, but I'm curious how those two things are handled differently. If it's a long explanation, don't worry about it. This is mainly my curiosity here.

src/operator/context/operator.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const [orgsStatus, setOrgsStatus] = useState(RemoteDataState.NotStarted)
+  const [operatorStatus, setOperatorStatus] = useState(
+    RemoteDataState.NotStarted
+  )
+  const [searchTerm, setSearchTerm] = useState('')
+  const dispatch = useDispatch()
+
+  const [organizations, setOrganizations] = useState([])
+
+  // TODO(ariel): might need to debounce
```
randycoulman, line 64: Yeah, I think you will. I think we fire off a new request on every character typed in the search bar, which is excessive if someone is typing reasonably fast.

src/shared/copy/notifications.ts
```diff
@@ -104,6 +104,62 @@ export const submitError = (): Notification => ({
     'There was an error submitting the upgrade request, please try again.',
 })
 
+// Operator Notifications
+export const getOrgsError = (): Notification => ({
+  ...defaultErrorNotification,
+  duration: FIVE_SECONDS,
+  message:
+    'There was an error getting the all the organization, please try again.',
```
randycoulman, line 112:
```suggestion
    'There was an error getting the all the organizations, please try again.',
```

src/shared/copy/notifications.ts
```diff
@@ -104,6 +104,62 @@ export const submitError = (): Notification => ({
...
+export const updateLimitsError = (id: string): Notification => ({
+  ...defaultErrorNotification,
+  duration: FIVE_SECONDS,
+  message: `Could not update limits for this organization ${id}`,
+})
+
+export const updateLimitsSuccess = (id: string): Notification => ({
+  ...defaultSuccessNotification,
+  duration: FIVE_SECONDS,
+  message: `Successfully updated limits for the organization ${id}`,
```
randycoulman, line 136: This one says `the organization` while the two above say `this organization`. They should be consistent. I think I like `the organization` better, but totally up to you.

src/types/operator.ts
```diff
@@ -0,0 +1,44 @@
...
+export {
+  Account,
+  Me,
+  BillingContact,
+  Organizations,
+  OrgLimits,
+} from 'src/client/unityRoutes'
+
+export interface Organization extends GenOrg {}
+export interface Organization extends GenOrg {}
```
randycoulman: dup?

randycoulman (changes requested):
Wow, that was a huge PR to review!  Overall, I like the changes you've made, but I think there are a few things that need to be addressed:

- Extract the lodash refactoring to its own PR to make it easier to revert either of the two parts independently.
- Restore the top-level `/operator/accounts`, `/operator/organizations` (and eventually `/operator/users`) routes rather than hiding that inside a context. As I recall, having those as separate routes was an explicit feature request.
- Double-check the way `LimitsInput` is working. I think the removal of lodash `get` and `set` has broken our usage of the input because the `name` provided to the input is a dotted name.
- Restore the `debounce` around the account and org list queries; we fire off a query on each character typed in the search bar.

Every other comment is optional and up to you.

src/client/swagger/unity.yml
```diff
@@ -625,6 +805,12 @@ components:
           type: string
           enum: [aws, gcm, azure]
           example: gcm
+        subscriberId:
+          type: string
+          description: TODO - not sure what this is referring to
+        status:
+          type: string
+          description: TODO - not sure what this is referring to
```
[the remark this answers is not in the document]
Asalem1: @randycoulman which of the marketplace subscription properties should we consider as required?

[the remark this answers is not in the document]
randycoulman: Not sure; I think we only get those if there is actually a marketplace subscription and not for Zuora. Maybe @ischolten knows better than I do?

src/index.tsx
```diff
@@ -76,9 +76,24 @@ class Root extends PureComponent {
...
               )}
+              {/* Operator Routes */}
+              {/* These are lumped under individual conditions since Switch statements expects a route as a child and can't handle React Fragments */}
+              {CLOUD && isFlagEnabled('unity-operator') && (
+                <Route
+                  exact
+                  path="/operator/accounts/:accountID"
+                  component={AccountPage}
+                />
+              )}
```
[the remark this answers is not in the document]
Asalem1: Yup, that's a really good point

src/operator/OperatorPage.tsx
```diff
@@ -0,0 +1,24 @@
...
+import {CLOUD} from 'src/shared/constants'
+
+const OperatorPage: FC = () => (
+  <OperatorProvider>
+    <>
+      <Operator />
+      {CLOUD && isFlagEnabled('unity-operator') && (
+        <Route path="/operator/organizations/:orgID" component={OrgOverlay} />
+      )}
+    </>
```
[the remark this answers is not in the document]
randycoulman: I'm not sure of all the reasons that we were asked to do this; might be on the AC for the stories back when this was first built, but part of it is to allow support to bookmark the various pages, for example.

This analogous to some of the other places in the UI where you can go to a specific tab by URL route.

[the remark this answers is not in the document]
Asalem1: 🤔 k, I'll go ahead and refactor this to use links to specific pages rather than have a route agnostic one dictated by tab selections

src/operator/OrgOverlay.tsx
Asalem1: This is being handled by a global alert notification that we're dispatching in the context provider

src/operator/context/operator.tsx
Asalem1: totally, I wanted to stay away from lodash but I'll add it all back in for now

src/operator/context/account.tsx
Asalem1: Nah, not a long explanation. I'm not sure how the Quartz UI was handling errors that are thrown, but for the UI the page will error and give the user an option to submit an issue to the UI repo. Since that isn't always particularly helpful (especially for an internal tool), i figured a notification of an error would give the user a better way of understanding what exactly went wrong and a potentially actionable step without effectively crashing the browser

src/operator/ResourcesSearchbar.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {OperatorRoutes} from 'src/operator/constants'
+
+const ResourcesSearchbar: FC = () => {
+  const {pathname, searchTerm, setSearchTerm} = useContext(OperatorContext)
+
+  const changeSearchTerm = (event: ChangeEvent<HTMLInputElement>) => {
+    setSearchTerm(event.target.value)
+  }
+
+  const isOrgsTab = pathname.includes(OperatorRoutes.organizations)
```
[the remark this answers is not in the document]
Asalem1: it is a boolean

src/operator/OrgOverlay.tsx
```diff
@@ -0,0 +1,242 @@
...
+    organization,
+    orgStatus,
+    setLimits,
+    updateLimitStatus,
+  } = useContext(OverlayContext)
+
+  const {orgID} = useParams()
+  const history = useHistory()
+
+  const setInputs = limits => {
```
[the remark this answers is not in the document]
Asalem1: Good call. Didn't see the last two

src/client/swagger/unity.yml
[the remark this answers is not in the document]
ischolten: For the operator ui, the marketplaceSubscription was optional on an account. If there is a marketplaceSubscription, it will have subscriberId, status, and marketplace. These fields are just used as information to display on the operator ui so we know if its a marketplace account and additional information for investigation if need be.

Also, it looks like this `Marketplace` is actually a representation of a `MarketplaceSubscription`.

randycoulman (approved):
Changes LGTM!

src/index.tsx
```diff
@@ -76,9 +76,16 @@ class Root extends PureComponent {
...
           <Suspense fallback={<PageSpinner />}>
             <Switch>
+              {/* TODO(ariel): we need to restrict access to the checkout and operator pages based on specific critera:
+                https://github.com/influxdata/ui/issues/848
+               */}
               {CLOUD && isFlagEnabled('unity-checkout') && (
                 <Route path="/checkout" component={CheckoutPage} />
               )}
+              {/* Operator Routes */}
+              {CLOUD && isFlagEnabled('unity-operator') && (
```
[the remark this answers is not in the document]
Asalem1: They are not, but I'll make sure to add em in

src/operator/LimitsInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const newValue =
+      type === InputType.Number ? parseFloat(e.target.value) : e.target.value
+    const newLimits = set(name, newValue, limits)
+    onChangeLimits(newLimits)
+  }
+  return (
+    <Input
+      type={type}
+      name={name}
+      value={get(limits, name, '')}
```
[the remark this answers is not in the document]
Asalem1: I'm not familiar with the difference. FWIW the Quartz UI is using the generic `lodash` for the `get` and the `lodash/fp` for the set. Don't wanna mess with it too much if it's working :)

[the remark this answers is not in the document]
randycoulman: Oh, weird, OK. I thought it was using `/fp` for both. `lodash/fp` is written so that the "thing" being operated on (the limits, in this case) is the last argument in the list. It allows you to do fancy things like currying and other functional programming things.

Issue #1018, opened 2021-03-29
Local Storage fails when keys don't exist


hoorayimhelping:
### Background
When a new key is added to the redux store, it appears that the corresponding localState key doesn't get added properly. This is causing issues on tools with annotations enabled.

When Adam ran into this, I was able to inspect his local storage, and his `userSettings` didn't have the proper key `showAnnotationsControls`. It didn't exist. Because of this, react crashes when trying to access it.

![image.png](https://images.zenhubusercontent.com/5c33b08d24dda4712ef4ae10/0d1a9874-7d92-4325-8517-17b52bce9187)

ChitlangeSahas:
[WIP] Current progress:

I'm able to reproduce this by:

1. clearing my localStorage in the browser console: `localStorage.clear()`
2. going to `src/userSettings/reducers/index.ts` and commenting out code that adds the new `showAnnotationsControl` key to the local storage.

kristinarobinson:
@ShmuelLotman -- was this fixed?

hoorayimhelping:
closed with https://github.com/influxdata/ui/pull/1020

PR #1020, opened 2021-03-29
fix: prevent DashboardPage from crashing due to the lack of updated redux store keys

ChitlangeSahas:
I'm able to reproduce the error in #1018 by doing the following:

1. clearing my localStorage in the browser console: `localStorage.clear()`
2. going to `src/userSettings/reducers/index.ts` and commenting out code that adds the new `showAnnotationsControl` key to the local storage.
3. reload the dashboard, it won't load. The screen displays a similar error, the console has a React Invariant Violation error as well.

Why it's happening?

I hypothesize: We have a variable that controls the visibility of the AnnotationContolBar, the variable's state is stored in the redux. When we shipped Annotation to our new users, the new key wasn't present in their localStorage, hence the value of the `AnnotationsControlBar`'s visibility variable was `undefined`, not `false`. Using a default `false` when key isn't found in the local storage fixes this crash.

After speaking to Bucky, we concluded that _this is a bandaid solution_

Ideally we would want to value the redux store as the source of truth not the `localStorage`, like we are doing now. A good solution would be to check the local storage, use it as a cache but invalidate it when there are mismatches between keys, indicating that the `localStorage` isn't usable anymore and that it needs to update.

### DIFF
--- a/src/dashboards/components/DashboardPage.tsx
+++ b/src/dashboards/components/DashboardPage.tsx
@@ -150,7 +150,7 @@ const mstp = (state: AppState) => {
     state.currentDashboard.id
   )
 
-  const showAnnotationBar = state.userSettings.showAnnotationsControls || false
+  const showAnnotationBar = state.userSettings.showAnnotationsControls ?? false
 
   return {
     startVisitMs: state.perf.dashboard.byID[dashboard.id]?.startVisitMs,
--- a/src/dashboards/components/DashboardPage.tsx
+++ b/src/dashboards/components/DashboardPage.tsx
@@ -150,7 +150,7 @@ const mstp = (state: AppState) => {
     state.currentDashboard.id
   )
 
-  const showAnnotationBar = state.userSettings.showAnnotationsControls
+  const showAnnotationBar = state.userSettings.showAnnotationsControls || false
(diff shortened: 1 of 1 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1020)

src/dashboards/components/DashboardPage.tsx
```diff
@@ -150,7 +150,7 @@ const mstp = (state: AppState) => {
     state.currentDashboard.id
   )
 
-  const showAnnotationBar = state.userSettings.showAnnotationsControls
+  const showAnnotationBar = state.userSettings.showAnnotationsControls || false
```
hoorayimhelping, line 153: i think we want to use `??` here instead of `||`:

https://dev.to/hereisnaman/logical-or-vs-nullish-coalescing-operator-in-javascript-3851

hoorayimhelping (approved):

PR #1048, opened 2021-03-30
fix: type ahead drop down fixes

jrenee42:
Closes #21020 in influxdb
https://app.zenhub.com/workspaces/applicationui-5e3b05772922e316b3a210a4/issues/influxdata/influxdb/21020

addressing bucky's comments
the arrow buttons do NOT select the next one, they just hover like a mouse would; selection happens on return
if you type in a matching valid selection and press return, that selects the value as well
dependent variables now re-load correctly

added unit tests
added e2e tests

still testing original variable dropdown too

### DIFF
--- a/cypress/e2e/shared/dashboardsView.test.ts
+++ b/cypress/e2e/shared/dashboardsView.test.ts
@@ -283,7 +283,7 @@ describe('Dashboard', () => {
       })
     })
 
-    it('can manage variable state with a lot of pointing and clicking', () => {
+    it.only('can manage variable state with a lot of pointing and clicking', () => {
       const bucketOne = 'b1'
       const bucketThree = 'b3'
       const bucketVarName = 'bucketsCSV'
@@ -329,19 +329,19 @@ describe('Dashboard', () => {
 
               // TESTING CSV VARIABLE
               // selected value in dashboard is 1st value
-              cy.getByTestID(`variable-dropdown--${bucketVarName}`).should(
-                'contain',
-                bucketOne
-              )
+              cy.getByTestID(
+                `variable-dropdown-input-typeAhead--${bucketVarName}`
+              ).should('have.value', bucketOne)
+
               cy.window()
                 .pipe(getSelectedVariable(dashboard.id, 0))
                 .should('equal', bucketOne)
 
               // testing variable controls
-              cy.getByTestID(`variable-dropdown--${bucketVarName}`).should(
-                'contain',
-                bucketOne
-              )
+              cy.getByTestID(
+                `variable-dropdown-input-typeAhead--${bucketVarName}`
+              ).should('have.value', bucketOne)
+
               cy.getByTestID('variables--button').click()
               cy.getByTestID(`variable-dropdown--${bucketVarName}`).should(
                 'not.exist'
@@ -361,10 +361,10 @@ describe('Dashboard', () => {
               cy.get(`#${bucketThree}`).click()
 
               // selected value in dashboard is 3rd value
-              cy.getByTestID(`variable-dropdown--${bucketVarName}`).should(
-                'contain',
-                bucketThree
-              )
+              cy.getByTestID(
+                `variable-dropdown-input-typeAhead--${bucketVarName}`
+              ).should('have.value', bucketThree)
+
               cy.window()
                 .pipe(getSelectedVariable(dashboard.id, 0))
                 .should('equal', bucketThree)
@@ -421,22 +421,22 @@ describe('Dashboard', () => {
               cy.window()
                 .pipe(getSelectedVariable(dashboard.id, 0))
                 .should('equal', bucketOne)
+              cy.pause()
 
               // selected value in dashboard is 1st value
-              cy.getByTestID(`variable-dropdown--${bucketVarName}`).should(
-                'contain',
-                bucketOne
-              )
+              cy.getByTestID(
+                `variable-dropdown-input-typeAhead--${bucketVarName}`
+              ).should('have.value', bucketOne)
+
               cy.window()
                 .pipe(getSelectedVariable(dashboard.id, 0))
                 .should('equal', bucketOne)
 
               // TESTING MAP VARIABLE
               // selected value in dashboard is 1st value
-              cy.getByTestID(`variable-dropdown--${mapTypeVarName}`).should(
-                'contain',
-                'k1'
-              )
+              cy.getByTestID(
+                `variable-dropdown-input-typeAhead--${mapTypeVarName}`
+              ).should('have.value', 'k1')
               cy.window()
                 .pipe(getSelectedVariable(dashboard.id, 2))
                 .should('equal', 'v1')
@@ -448,10 +448,9 @@ describe('Dashboard', () => {
               cy.get(`#k2`).click()
 
               // selected value in dashboard is 2nd value
-              cy.getByTestID(`variable-dropdown--${mapTypeVarName}`).should(
-                'contain',
-                'k2'
-              )
+              cy.getByTestID(
+                `variable-dropdown-input-typeAhead--${mapTypeVarName}`
+              ).should('have.value', 'k2')
               cy.window()
                 .pipe(getSelectedVariable(dashboard.id, 2))
                 .should('equal', 'v2')
@@ -487,10 +486,9 @@ describe('Dashboard', () => {
                 .should('equal', 'v1')
 
               // selected value in dashboard is 1st value
-              cy.getByTestID(`variable-dropdown--${mapTypeVarName}`).should(
-                'contain',
-                'k1'
-              )
+              cy.getByTestID(
+                `variable-dropdown-input-typeAhead--${mapTypeVarName}`
+              ).should('have.value', 'k1')
               cy.window()
                 .pipe(getSelectedVariable(dashboard.id, 2))
                 .should('equal', 'v1')
--- a/cypress/e2e/shared/dashboardsView.test.ts
+++ b/cypress/e2e/shared/dashboardsView.test.ts
@@ -732,7 +732,7 @@ csv.from(csv: data) |> filter(fn: (r) => r.bucket == v.bucketsCSV)`
                 .clear()
                 .type('b3789')
 
-              //now click away
+              // now click away
               cy.getByTestID('variable-dropdown--button')
                 .eq(1)
                 .click()
@@ -748,7 +748,7 @@ csv.from(csv: data) |> filter(fn: (r) => r.bucket == v.bucketsCSV)`
                 .clear()
                 .type('b3')
 
-              //now click away
+              // now click away
               cy.getByTestID('variable-dropdown--button')
                 .eq(1)
                 .click()
--- a/cypress/e2e/shared/dashboardsView.test.ts
+++ b/cypress/e2e/shared/dashboardsView.test.ts
@@ -1,6 +1,8 @@
 import {Organization, AppState, Dashboard} from '../../../src/types'
 import {lines} from '../../support/commands'
 
+const typeAheadFlag = 'typeAheadVariableDropdown'
+
 describe('Dashboard', () => {
   beforeEach(() => {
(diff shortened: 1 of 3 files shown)


cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -287,224 +289,371 @@ describe('Dashboard', () => {
...
               })
-              // save cell
-              cy.getByTestID('save-cell--button').click()
+            })
+          })
+        })
+      })
+    })
+
+    it('dependent variables reload properly', () => {
```
hoorayimhelping, line 605: jest (and any testing framework that is an ancestor of rspect, which jest is through mocha and jasmine) is designed so that description blocks and test/it blocks flow like english sentences. this is done by considering the `it` function to be a noun in the sentence of the string that describes the test. so this description reads like:

>it dependent variables reload properly

when it'll flow better with something like:

```ts
it('reloads dependent variables properly')
```

small nit, but helps with readability and maybe clarifying what the behavior we're testing is

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -287,224 +289,371 @@ describe('Dashboard', () => {
...
+                // next one should be active (first should NOT be active)
+                cy.get(`#anotherBucket`).should('not.have.class', 'active')
+                cy.get(`#randomBucket`).should('have.class', 'active')
 
-              // selected value in cell context is 2nd value
-              cy.window()
-                .pipe(getSelectedVariable(dashboard.id, 2))
-                .should('equal', 'v2')
+                // now; press return/enter to set it:
+                cy.getByTestID(
```
hoorayimhelping, line 432: is the spacing off here?

src/variables/components/TypeAheadVariableDropdown.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+            order: ['03cbdc8a53a63000'],
+          },
+        },
+      },
+    },
+  }
+}
+
+describe('Dashboards.Components.VariablesControlBar.TypeAheadVariableDropdown', () => {
+  describe('if map type', () => {
```
hoorayimhelping, line 86: nit:
```suggestion
  describe('the map type', () => {
```

src/variables/components/TypeAheadVariableDropdown.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    // filter again by text:
+    checkDropdown('e', evalues)
+    checkDropdown('al', alvalues)
+
+    // something that won't match anything
+    // (see: https://testing-library.com/docs/guide-disappearance/#asserting-elements-are-not-present)
+    fireEvent.change(filterInput, {target: {value: 'def23'}})
+    const items = screen.queryByTestId('variable-dropdown--item')
+    expect(items).toBeNull() // it doesn't exist
+  })
```
hoorayimhelping, line 136: nice tests!

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -51,107 +56,184 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
...
+    this.state = defaultState
   }
 
   // set the 'shownValues' after loading, and
   // resets the menuOpen variable
   componentDidUpdate(prevProps, prevState) {
     const prevVals = prevProps.values
-    const {values, selectedValue} = this.props
+    const {values, selectedValue, status} = this.props
+    const {status: prevStatus} = prevProps
```
hoorayimhelping, line 75: just to combine everything, can you also set `prevVals` in this object (if it's possible). it should be something like:

```ts
const {status: prevStatus, values: prevVals} = prevProps
```

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -51,107 +56,184 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
...
     const {actualVal, loaded, selectHappened, menuOpen} = this.state
     const {
       actualVal: prevActualVal,
       selectHappened: prevSelectHappened,
       menuOpen: prevMenuOpen,
     } = prevState
 
+    let newState = {}
+
+    const justLoaded = () =>
```
hoorayimhelping, line 85: does this need to be a function? can it just be a value that is set?

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -51,107 +56,184 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
...
+     * unset the menuOpen; it should be set to closed (or open) only once; then undone
+     * this is needed because: (from clockface Dropdown.tsx documentation):
+     * if the string is set to 'open', and then the user closes it, and the code sets it to open again,
+     * unless the code sets it to something else in between (like null or 'close'),
+     * then nothing will happen- the menu will not open)
+     */
+
     if (menuOpen !== prevMenuOpen && menuOpen !== null) {
-      this.setState({menuOpen: null})
+      newState['menuOpen'] = null
```
hoorayimhelping, line 119: not a huge deal, but the only time we access arrays with bracket notation, is when it's not a valid key name. in other words 
```
newState.menuOpen = null
```
 is preferred here

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -51,107 +56,184 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
...
     }
 
     // need to have this, as the 'onClickAwayHere' gets triggered *before*
     // the selected value is set to the actualValue (it keeps re-using the original
     // property)
```
hoorayimhelping, line 124: is this comment referring to anything that you can see? i don't know what it's referring to. let's kill it, or move it to where it's relevant

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -51,107 +56,184 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
...
 
+    const needle = event?.target?.value
+
+    // if there is no value, set the shownValues to everything
+    // and set the typedValue to nothing (zero it out)
+    // reset the selectIndex too
     if (!needle) {
-      this.setState({shownValues: values, typedValue: needle})
+      this.setState({shownValues: values, typedValue: needle, selectIndex: -1})
     } else {
```
hoorayimhelping, line 163: i know we already shipped this code, but stylistically, i think this `if` statement should return early. the `else` statement is actually the function body that does what the function is named, but it's easy to miss because it's part of an `else` which has let weight than an `if` statement. if the `if` was transformed into a guard statement that returned early after it set state, the filtering could just be on equal indentation as the `if` statement

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -51,107 +56,184 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
...
+        // they used the arrows; just pressed return
+        this.handleSelect(shownValues[selectIndex], true)
+      } else {
+        // the person could have been typing and pressed return, need to
+        // make sure the value in the input field is real/legal:
+
+        // but:  if the value they typed is LEGAL (in the list/dropdown values), set it;
+        // else: reset to the previous real/legal value:
+        if (values.includes(typedValue)) {
+          // is is a real legal value
```
hoorayimhelping, line 217: >is is

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -160,7 +242,7 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
     const {selectedValue} = this.props
 
     if (actualVal || selectedValue) {
-      const realValue = actualVal ?? selectedValue
+      const realValue = selectedValue ?? actualVal
```
hoorayimhelping, line 245: out of curiosity, is this change in a test? i don't remember all the e2e details

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -208,18 +304,23 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
             onCollapse={onCollapse}
             theme={DropdownMenuTheme.Amethyst}
           >
-            {shownValues.map(val => {
+            {shownValues.map(function(value, index) {
```
hoorayimhelping, line 307:
```suggestion
            {shownValues.map((value, index) => {
```

should autobind for you and fix everything up, right?

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -208,18 +304,23 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
...
+              let classN = 'variable-dropdown--item'
+              // highlight when arrowing; like a hover
+              if (index === selectIndex) {
+                classN += ' active'
+              }
```
hoorayimhelping, line 312: we use `classnames` as a dependency, it's a great tool for this kind of stuff: https://jedwatson.github.io/classnames/

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -278,11 +378,27 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
...
       return 'Loading...'
     }
+    if (this.noValuesPresent(false)) {
+      return 'No Values'
+    }
+
     return defaultText
   }
+
+  noValuesPresent = useAllValues => {
```
hoorayimhelping, line 392: boolean arguments are usually a smell that this should just be two functions https://martinfowler.com/bliki/FlagArgument.html

what do you think?

hoorayimhelping (changes requested):
great work, awesome tests. i was going to approve it, but the autobinding thing gave me pause. let's make sure that gets figured out

cypress/e2e/shared/dashboardsView.test.ts
jrenee42: pretty-quick did this ...will adjust but if pretty-quick changes it back; well than it's a prettier issue

src/variables/components/TypeAheadVariableDropdown.test.tsx
jrenee42: i'll change it but it isn't my code (just formatted it with pretty quick)

jrenee42: not my code (was just formatted it) but will change it to 'variable map type'

src/variables/components/TypeAheadVariableDropdown.tsx
jrenee42: this change is to make sure that the selectedValue trumps the previously selected value;
it involves race conditions and such; 

it is implicitly tested, since the e2e tests select one thing, then select another thing, and click away....

but adding more tests to be sure.

src/variables/components/TypeAheadVariableDropdown.tsx
jrenee42: I separated them but kept a hidden internal method to avoid code duplication

src/variables/components/TypeAheadVariableDropdown.tsx
jrenee42: cool! using it

hercules-influx:
Metric | Score | Level
performance | 35 | :red_circle:
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1048)

hoorayimhelping (approved):
LGTM! The feature is solid! I imagine there will be some follow on edge-case bug fixes or ux fixes, but I think this will make quite a few folks happy

PR #1052, opened 2021-03-31
fix: flag the GET request to annotations

### LINKED ISSUE #1051: Annotations Bug: Annotations is incorrectly making an unflagged GET request
With the `annotations` flag off, annotations requests are still being made.

hoorayimhelping:
Closes #1051

Don't make GET requests to annotations if the flag is off

### DIFF
--- a/src/dashboards/components/DashboardPage.tsx
+++ b/src/dashboards/components/DashboardPage.tsx
@@ -62,7 +62,10 @@ class DashboardPage extends Component<Props> {
     resetQueryCache()
 
     this.emitRenderCycleEvent()
-    this.props.fetchAndSetAnnotations()
+
+    if (isFlagEnabled('annotations')) {
+      this.props.fetchAndSetAnnotations()
+    }
   }
 
   public componentWillUnmount() {


ChitlangeSahas (approved):

jrenee42:
the code itself is good.

I'm concerned about null-safe/flag safe annotation fetching;


this code in particular:

`
export const getVisibleAnnotationStreams = (state: AppState): string[] => {
  const visibleStreams = state.annotations.visibleStreamsByID

  return sortBy(visibleStreams, stream => stream)
}`

from annotations/selectors/index.ts

need to change it to:

`
export const getVisibleAnnotationStreams = (state: AppState): string[] => {
  const visibleStreams = state.annotations?.visibleStreamsByID

  const result = visibleStreams? sortBy(visibleStreams, stream => stream) : []
}`

besides that the accessing of it is all under the flag.

hoorayimhelping:
@jrenee42 I looked into this, and there are [defaults set in the annotations reducer](https://github.com/influxdata/ui/blob/master/src/annotations/reducers/index.ts#L29) which make this a safe function. There are proper initial values so that these functions work. But your comment did make me look into fetching streams with the flag off

hoorayimhelping:
@jrenee42 I spoke too soon. I did some further testing and the annotation stream request doesn't get made unless annotations are toggled on. The request happens in the annotation control bar, which isn't rendered if the annotation flag is off.

So the conclusion is: great suggestions, I checked up on those with the flag off and we're okay, and everything is safe. We just need to flag the call to GET annotations.

jrenee42 (approved):

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1052)

PR #1059, opened 2021-03-31
feat: connect missing pieces for UI unification

Asalem1:
This PR cleans up and integrates some missing dependencies for the UI based on the API contracts

### DIFF
--- a/src/billing/api/index.ts
+++ b/src/billing/api/index.ts
@@ -2,20 +2,23 @@ import {Inputs} from 'src/checkout/context/checkout'
 import {
   getAccount as apiGetAccount,
   getBilling,
+  getOrgsLimits as apiGetOrgLimits,
   postCheckout,
   getSettingsNotifications,
   getPaymentForm,
   getBillingInvoices,
   putBillingPaymentMethod,
+  putSettingsNotifications,
 } from 'src/client/unityRoutes'
 
-import {RemoteDataState, CheckoutRequest} from 'src/types'
+import {RemoteDataState} from 'src/types'
 import {
   Account,
   Invoice,
   CreditCardParams,
   BillingInfo,
   BillingNotifySettings,
+  OrgLimits,
 } from 'src/types/billing'
 
 const makeResponse = (status, data, ...args) => {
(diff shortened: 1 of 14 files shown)


src/billing/api/index.ts
```diff
@@ -84,6 +87,40 @@ export const getBillingCreditCardParams = (): ReturnType<typeof getPaymentForm>
...
+    rate: {
+      readKBs: 100,
+      writeKBs: 100,
+      cardinality: 1,
+    },
+    bucket: {
+      maxRetentionDuration: 3,
+      maxBuckets: 2,
+    },
+    task: {
```
subirjolly, line 102: It's so weird that we have an object just to contain One property. Why can't it be just `maxTasks` or maybe a key which has something like this:
```
maximums: {
    tasks: 10,
    dashboards: 7,
    checks: 4,
    ....
}
```

src/billing/thunks/index.ts
```diff
@@ -11,20 +11,27 @@ import {
...
 } from 'src/billing/api'
 
 // Types
 import {RemoteDataState} from 'src/types'
-import {Invoice} from 'src/types/billing'
+import {BillingNotifySettings, Invoice} from 'src/types/billing'
 
+// TODO(ariel): add error handling here
+// notify() will not work since Dispatch here is based on the passed in dispatch from the local reducer
+// and not from the higher level dispatch from the app.
```
Asalem1, line 34: Elaborating a bit on this here. I did an audit of the error states and it seems like the two places we aren't actively accounting for errors in the unification work are on the Users page and the Billing page. Both use the reducer pattern over the context pattern making it a little more difficult for us to simply trigger errors from the thunks. We want to refactor these to a context-based pattern in order to prevent issues like these and keep a centralized style rather than a disparate one in different parts of unification

src/billing/api/index.ts
Asalem1: Agreed. I think it's a construct of the API contracts that current exist in Quartz and IDPE.

subirjolly (approved):

PR #1071, opened 2021-04-01
feat: variables reorder drag and drop saves through logout

### LINKED ISSUE #667: Save ordering of variables in a dashboard
**Proposal:**
In the Data Explorer, when any Dashboard is viewed, the display order of the variables and their value inputs should be saved.

**Current behavior:**
In the Data Explorer, when any Dashboard is viewed, the list of variables for that Dashboard can rearranged.  However, when a user logs out, any reordering of the displayed list is not retained.

**Desired behavior:**
In the Data Explorer, when any Dashboard is viewed, the list of variables for that Dashboard can rearranged.  Save the display order with each Dashboard, so that after log out, the reordering of the displayed list is retained.

**Use case:**
InfluxData Support team has requested this.  They work with dashboards that contain many variables, and would like to have more frequently changed variables to be grouped together.
Another use case, order of variables displayed may matter.  For example, if one has a variable distance and another variable units ie km|mi, it would make sense for units to always come right after distance.
Another use case is where one variable dynamically populates another variable. In that situation, the dynamically generate variable should always come right after the variable that it’s based off of. Like a clusters variable and a hosts variable in the selected cluster.

ShmuelLotman:
Closes #667

This PR connects with the backend change that introduces a `sort_order` property on variables. This allows us to save the order when they are moved in the control bar and display them in the correct order once they are saved, or otherwise show an error notification and snap them back to match the API order.

### DIFF
--- a/src/variables/selectors/index.tsx
+++ b/src/variables/selectors/index.tsx
@@ -109,7 +109,11 @@ export const getVariablesForDashboard = (state: AppState): Variable[] => {
     Object.values(state.resources.views.byID).filter(
       variable => variable.dashboardID === state.currentDashboard.id
     )
-  ).sort((a, b) => a.sort_order - b.sort_order)
+  ).sort((a, b) => {
+    if (a.sort_order && b.sort_order) {
+      return a.sort_order - b.sort_order
+    }
+  })
 
   return variablesUsedByDashboard
 }
--- a/src/variables/selectors/index.tsx
+++ b/src/variables/selectors/index.tsx
@@ -109,7 +109,7 @@ export const getVariablesForDashboard = (state: AppState): Variable[] => {
     Object.values(state.resources.views.byID).filter(
       variable => variable.dashboardID === state.currentDashboard.id
     )
-  )
+  ).sort((a, b) => a.sort_order - b.sort_order)
 
   return variablesUsedByDashboard
 }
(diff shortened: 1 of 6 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1071)

src/variables/selectors/index.tsx
```diff
@@ -109,7 +109,7 @@ export const getVariablesForDashboard = (state: AppState): Variable[] => {
     Object.values(state.resources.views.byID).filter(
       variable => variable.dashboardID === state.currentDashboard.id
     )
-  )
+  ).sort((a, b) => a.sort_order - b.sort_order)
```
Asalem1, line 112: what happens with existing variables that don't have a sort order?

Asalem1 (commented):
What happens with existing variables that don

Asalem1 (approved):

src/variables/selectors/index.tsx
ShmuelLotman: All variables will now have a sort_order that defaults to 0

PR #1072, opened 2021-04-01
feat: complete redirect cycle to login and avoid the signin page

### LINKED ISSUE #1057: Feature Flagged avoidance of the `/signin` endpoint
There's an entire thread on the unification channel discussing this in greater detail, but the TLDR here is that since we're going to be sharing sessions across the app, a failure of `/me` should result in a redirect to the `/login` page, and  **not** make a request to the api `/signin` route to help avoid an endless login loop (among other issues listed in the thread). This should be feature flagged using the same feature flag that's currently used for session sharing so as to roll the feature out together, but should also not be triggered in the tools environment since this approach will cause issues in tools

Asalem1:
Closes #1057 

This PR integrates the redirect to login functionality when a cloud user fails the `/me` call. With session sharing allowing for access to both platforms, we want to avoid making a silent request to Auth0 (which is currently done by the `/signin` endpoint). By redirecting the user to the `/login` page, we ensure that the `/signin` endpoint will not be hit and that the user will end up in a state where they can then login to the app

### DIFF
--- a/src/Signin.tsx
+++ b/src/Signin.tsx
@@ -16,13 +16,18 @@ import {
   removeFromLocalStorage,
   setToLocalStorage,
 } from 'src/localStorage'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Actions
 import {notify as notifyAction} from 'src/shared/actions/notifications'
 
 // Constants
 import {sessionTimedOut} from 'src/shared/copy/notifications'
-import {CLOUD, CLOUD_SIGNIN_PATHNAME} from 'src/shared/constants'
+import {
+  CLOUD,
+  CLOUD_LOGIN_PATHNAME,
+  CLOUD_SIGNIN_PATHNAME,
+} from 'src/shared/constants'
 
 // Types
 import {RemoteDataState} from 'src/types'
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1072)

randycoulman (approved):
LGTM!

subirjolly (approved):

src/Signin.tsx
```diff
@@ -98,6 +103,15 @@ export class Signin extends PureComponent<Props, State> {
 
       clearInterval(this.intervalID)
 
+      if (CLOUD && isFlagEnabled('authSessionCookieOn')) {
+        const url = new URL(
+          `${window.location.origin}${CLOUD_LOGIN_PATHNAME}?redirectTo=${window.location.href}`
+        )
+        setToLocalStorage('redirectTo', window.location.href)
+        window.location.href = url.href
+        throw error
```
genehynson, line 112: sorry if this is a dumb question - why are we throwing an error here?

src/Signin.tsx
```diff
@@ -98,6 +103,15 @@ export class Signin extends PureComponent<Props, State> {
 
       clearInterval(this.intervalID)
 
+      if (CLOUD && isFlagEnabled('authSessionCookieOn')) {
```
genehynson, line 106: this flag is off by default, right?

Asalem1: yes sir

src/Signin.tsx
Asalem1: Not a dumb question - based on the previous convention of having a stopping operation if the user isn't redirected. Instead of plopping a `return` here (which would do the same thing), we would want to throw an error in the hopes that we can gather information on what happened that a simple redirect would go wrong

genehynson: got it - but in normal scenarios, the user is redirected before we get to this line so no error is thrown?

Asalem1: yup, that's the hope 🤞🏽

genehynson (approved):

PR #1079, opened 2021-04-01
feat: date time range picker prevent invalid input 

ChitlangeSahas:
Closes https://github.com/influxdata/influxdb/issues/15632

This PR improves the UX of the date-time picker:

1. Doesn't allow submissions unless the start and end dates and times are valid [start dateTime has to be before end dateTime] 
2. Updates the input field with the selected date from the picker
3. Disables the submissions when the input field is invalid.
4. Only enables submissions when the dates are changed, so user can safely cancel the operation without feeling like they accidentally changed the date-window.

This is how it looks:

https://user-images.githubusercontent.com/18511823/113359002-b705f680-92fb-11eb-916d-c03085cd5ad5.mov


src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -108,11 +124,27 @@ class DateRangePicker extends PureComponent<Props, State> {
...
-    this.setState({upper})
+    this.setState(prev => ({
+      ...prev,
+      upper,
+      validDateRange: this.isTimeRangeValid(this.state.lower, upper),
+    }))
+  }
+
+  private isTimeRangeValid = (lower: string, upper: string): boolean => {
+    return moment(lower).isBefore(upper)
```
Asalem1, line 143: Really great work on the PR. We can totally avoid using moment here by simply doing something like:

‘’’
new Date(lower).getTime() < new Date(upper).getTime()
‘’’

this effectively converts the date string to a millisecond and compares the two

Asalem1 (commented):
I wanna take another look at this when I’m by my computer but we should have some skipped tests that we can utilize to add some coverage for this that’s linked in the comments of the original issue

src/shared/components/dateRangePicker/DateRangePicker.tsx
ChitlangeSahas: Very true!! I'll add that change 👍

src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -35,7 +41,9 @@ class DateRangePicker extends PureComponent<Props, State> {
       timeRange: {lower, upper},
     } = props
 
-    this.state = {lower, upper}
+    // The button can be disabled to start with, only activate it if the values are changed.
```
Asalem1, line 44: AFAIK the lower & upper are defaulted to a true scenario. It might be worth it for us to calculate the validity of the date range on instantiation rather than assuming it's invalid

src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -70,6 +80,11 @@ class DateRangePicker extends PureComponent<Props, State> {
             onClick={this.handleSetTimeRange}
             text="Apply Time Range"
             testID="daterange--apply-btn"
+            status={
+              this.state.validDateRange
```
Asalem1, line 84: we can destructure the `validDateRange` here:

https://github.com/influxdata/ui/pull/1079/files#diff-01e353f1c997af4eed6a066de1df2aa36349b7b10b77ea445251adfd797d0b0aR51

src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -108,11 +123,27 @@ class DateRangePicker extends PureComponent<Props, State> {
   }
 
   private handleSelectLower = (lower: string): void => {
-    this.setState({lower})
+    this.setState(prev => ({
+      ...prev,
```
Asalem1, line 127: why are we concerned with setting the previous stuff?

src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -108,11 +123,27 @@ class DateRangePicker extends PureComponent<Props, State> {
...
+      lower,
+      validDateRange: this.isTimeRangeValid(lower, this.state.upper),
+    }))
   }
 
   private handleSelectUpper = (upper: string): void => {
-    this.setState({upper})
+    this.setState(prev => ({
+      ...prev,
```
Asalem1, line 135: same question as above

Asalem1 (changes requested):

src/shared/components/dateRangePicker/DateRangePicker.tsx
ChitlangeSahas: My bad, I had some bugs that I thought were because I was doing setState wrong, turned out the issue was something else. I just forgot to delete it.

src/shared/components/dateRangePicker/DatePicker.tsx
```diff
@@ -155,9 +158,24 @@ export default class DatePicker extends PureComponent<Props, State> {
...
+  private overrideInputState = (): void => {
+    const {dateTime} = this.props
+    const {inputFormat} = this.state
+
+    let value
+    if (inputFormat) {
+      value = moment(dateTime).format(inputFormat)
+    } else {
+      value = moment(dateTime).format('YYYY-MM-DD HH:mm:ss.SSS')
+    }
```
Asalem1, line 170:
```suggestion
    let value = moment(dateTime).format('YYYY-MM-DD HH:mm:ss.SSS')
    if (inputFormat) {
      value = moment(dateTime).format(inputFormat)
    }
```

src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -108,11 +123,25 @@ class DateRangePicker extends PureComponent<Props, State> {
...
+      validDateRange: this.isTimeRangeValid(upper, this.state.upper),
+    })
+  }
+
+  private isTimeRangeValid = (lower: string, upper: string): boolean => {
+    return new Date(lower).getTime() < new Date(upper).getTime()
+  }
+
+  private handleInvalidInput = () => {
+    this.setState(prev => ({...prev, validDateRange: false}))
```
Asalem1, line 144: no need to spread `prev` here

ChitlangeSahas: Sorry my bad, yes, fixing it 👍

src/shared/components/dateRangePicker/DateRangePicker.tsx
ChitlangeSahas: Like we spoke in our slack conversation, David Kim who's our UX expert was saying its preferable to keep the Button disable until the user makes a change.

src/shared/components/dateRangePicker/DateRangePicker.tsx
ChitlangeSahas: ✅

Asalem1 (approved):
Offline. This should be good to go once the changes and tests are all updated to reflect the new enhancement

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1079)

PR #1088, opened 2021-04-02
fix: prevent annotations crashing XY plot when annotations aren't passed

### LINKED ISSUE #1085: Annotations: Annotations feature flag crashing notebooks visualizations
Trying to use notebooks with the annotations feature flag enabled breaks the visualization cell.

image

ChitlangeSahas:
Closes #1085

In the view.tsx, `annotations` weren't being checked if they were empty, this caused it to crash when the XY plot was in a notebook context and no annotations were passed in. 

It didn't make sense to me why it didn't happen in the data-explorer, we use XY plot there as well. Did a little digging and found that this is because annotations were shown in data explorer prior to us constraining it to a dashboard-cell. This meant that an empty array was being passed here: [Vis.tsx/L59](https://github.com/influxdata/ui/blob/c03ce296402acd8419d361c0e0835c469cf4ffeb/src/timeMachine/components/Vis.tsx#L59) and not an `undefined` as in the context of the notebooks. 

This raises another point, should we remove annotation code from the data-explorer visualization code? 

Here is the screenshot of it working with the annotation feature flag enabled, and using the notebooks: 

Screen Shot 2021-04-02 at 7 54 33 AM

### DIFF
--- a/src/visualization/types/Graph/view.tsx
+++ b/src/visualization/types/Graph/view.tsx
@@ -262,7 +262,7 @@ const XYPlot: FC<Props> = ({
 
   if (isFlagEnabled('annotations')) {
     // annotations and the cellID might or might be provided to this graph view
-    const cellAnnotations = annotations ? annotations[cellID] ?? [] : []
+    const cellAnnotations = annotations?.cellID ?? []
     const annotationsToRender: any[] = cellAnnotations.map(annotation => {
       return {
         ...annotation,
--- a/src/visualization/types/Graph/view.tsx
+++ b/src/visualization/types/Graph/view.tsx
@@ -261,8 +261,8 @@ const XYPlot: FC<Props> = ({
   }
 
   if (isFlagEnabled('annotations')) {
-    const cellAnnotations = annotations[cellID] ?? []
-
+    // annotations and the cellID might or might be provided to this graph view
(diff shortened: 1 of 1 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1088)

drdelambre (approved):

src/visualization/types/Graph/view.tsx
```diff
@@ -261,8 +261,8 @@ const XYPlot: FC<Props> = ({
   }
 
   if (isFlagEnabled('annotations')) {
-    const cellAnnotations = annotations[cellID] ?? []
-
+    // annotations and the cellID might or might be provided to this graph view
+    const cellAnnotations = annotations ? annotations[cellID] ?? [] : []
```
hoorayimhelping, line 265:
```suggestion
    const cellAnnotations = annotations?.cellID ?? []
```

is the more compact and modern form, I think. I haven't tested it yet though.

ChitlangeSahas: I'll test this!

hoorayimhelping (approved):

PR #1096, opened 2021-04-02
chore: Clean up unity.yml to remove circular dependencies

Asalem1:
This PR addresses a couple of nagging UI issues that are lingering around the `/me` endpoint, namely that:

- the checkout & operator page should be authenticated routes
- the checkout & operator page should only be accessible with specific credentials
- the yml file had some potentially cyclical dependencies between orgs and accounts that were simplified by flattening account props onto the org

This PR does not address:

- Adding conditional rendering logic for the `upgrade now` button with meta data that should be available with the `/me` endpoint

### DIFF
--- a/src/client/swagger/unity.yml
+++ b/src/client/swagger/unity.yml
@@ -840,7 +840,7 @@ components:
           type: string
           description: account id
         quartzId:
-          type: string
+          type: number
           description: account id in quartz
         onboardingState:
           type: string
@@ -1133,7 +1133,7 @@ components:
         quartzId:
           description: the quartz id of the user
           readOnly: true
-          type: string
+          type: number
         onboardingState:
           description: onboarding state of the user
           type: string
@@ -1232,7 +1232,7 @@ components:
           type: string
           description: ID of the org
         quartzId:
-          type: string
+          type: number
           description: quartz ID of the org
         name:
           type: string
--- a/src/client/swagger/unity.yml
+++ b/src/client/swagger/unity.yml
@@ -844,7 +844,6 @@ components:
           description: account id in quartz
         onboardingState:
           type: string
-          # TODO(ariel): get clarity on this
           description: stage of onboarding the account is in
         marketplace:
           $ref: '#/components/schemas/Marketplace'
@@ -1166,28 +1165,21 @@ components:
     Me:
       properties:
         id:
+          type: string
           description: the idpe id of the user
-          type: string
-        # TODO(ariel): find out if these are account or user specific
-        firstName:
-          type: string
-        # TODO(ariel): find out if these are account or user specific
-        lastName:
-          type: string
         email:
-          # TODO(ariel): find out about this one
-          description: Not sure if this is the billing account email or the account email, but this might be unnecessary since that should be included in the Account or nested in the Account
           type: string
+          description: the email associated with the user
         account:
           $ref: '#/components/schemas/Account'
           description: User Account for the account
-        isBeta:
+        isRegionBeta:
           type: boolean
           description: whether the region associated with the account is a beta region
         isOperator:
           type: boolean
           description: whether the user is an operator
-      required: [id, account, email, firstName, lastName, isOperator, isBeta]
+      required: [id, account, email, isOperator, isRegionBeta]
     Users:
       type: object
       properties:
@@ -1254,39 +1246,35 @@ components:
         date:
           type: string
           description: date org was created
-        accountType:
+        relatedAccount:
+          $ref: '#/components/schemas/RelatedAccount'
+          description: subset of related account information
+      required: [id, quartzId, name, region, provider, date]
+    Organizations:
+      type: array
+      items:
+        $ref: '#/components/schemas/Organization'
+    RelatedAccount:
+      description: Subset of Account data related to the organization
+      type: object
+      properties:
+        id:
+          type: string
+          description: account id
+        email:
+          type: string
+          description: email associated with the account
+        type:
           type: string
           description: type of the account
           enum:
             - free
             - cancelled
             - pay_as_you_go
-        accountBalance:
+        balance:
           type: number
-          description: balance of the account
-        accountEmail:
-          type: string
-          description: billing contact email
-        accountId:
-          type: string
-          description: id of the account associated with the org
-      required:
-        [
-          id,
-          quartzId,
-          name,
-          accountType,
-          accountBalance,
-          accountEmail,
-          accountId,
-          region,
-          provider,
-          date,
-        ]
-    Organizations:
-      type: array
-      items:
-        $ref: '#/components/schemas/Organization'
+          description: remaining balance on the account, nil if none
+      required: [id, email, type, balance]
     RateLimits:
       description: Usage rate limits
       type: object
--- a/src/client/swagger/unity.yml
+++ b/src/client/swagger/unity.yml
@@ -1162,55 +1162,26 @@ components:
         id:
           description: the idpe id of the user
           type: string
+        # TODO(ariel): find out if these are account or user specific
         firstName:
           type: string
+        # TODO(ariel): find out if these are account or user specific
         lastName:
           type: string
         email:
+          # TODO(ariel): find out about this one
+          description: Not sure if this is the billing account email or the account email, but this might be unnecessary since that should be included in the Account or nested in the Account
           type: string
-        orgId:
-          type: string
-        accountId:
-          type: string
+        account:
+          $ref: '#/components/schemas/Account'
+          description: User Account for the account
         isBeta:
           type: boolean
           description: whether the region associated with the account is a beta region
         isOperator:
           type: boolean
           description: whether the user is an operator
-        marketplace:
-          $ref: '#/components/schemas/Marketplace'
-          description: which marketplace, nil if none
-        balance:
-          type: number
-          description: remaining balance on the account, nil if none
-        billingContact:
-          $ref: '#/components/schemas/BillingContact'
-          description: billing contact for the account
-        users:
-          type: array
-          items:
-            $ref: '#/components/schemas/User'
-        type:
-          type: string
-          description: type of the account
-          enum:
-            - free
-            - cancelled
-            - pay_as_you_go
-      required:
-        [
-          id,
-          accountId,
-          email,
-          firstName,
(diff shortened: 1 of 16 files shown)


src/client/swagger/unity.yml
```diff
@@ -839,6 +839,13 @@ components:
         id:
           type: string
           description: account id
+        quartzId:
+          type: string
+          description: account id in quartz
+        onboardingState:
+          type: string
+          # TODO(ariel): get clarity on this
+          description: stage of onboarding the account is in
```
Asalem1, line 848: These two properties seem to only be relevant for operators visiting the `AccountView` page. That is to say, these fields are only needed when an operator is viewing an account-specific page

src/client/swagger/unity.yml
```diff
@@ -1162,55 +1168,26 @@ components:
...
           type: string
+        # TODO(ariel): find out if these are account or user specific
         firstName:
           type: string
+        # TODO(ariel): find out if these are account or user specific
         lastName:
           type: string
         email:
+          # TODO(ariel): find out about this one
+          description: Not sure if this is the billing account email or the account email, but this might be unnecessary since that should be included in the Account or nested in the Account
```
Asalem1, line 1179: not sure what the `firstName`, `lastName`, and `email` are supposed to signify here? Where's the data supposed to be coming from? The BillingContact? User? Account?

src/client/swagger/unity.yml
```diff
@@ -1162,55 +1168,26 @@ components:
...
           type: string
-        orgId:
-          type: string
-        accountId:
-          type: string
+        account:
+          $ref: '#/components/schemas/Account'
+          description: User Account for the account
```
Asalem1, line 1183: Adding the account here simplifies the need for various meta data that exists as a property of the account.

src/client/swagger/unity.yml
```diff
@@ -1162,55 +1168,26 @@ components:
...
         firstName:
           type: string
+        # TODO(ariel): find out if these are account or user specific
         lastName:
           type: string
         email:
+          # TODO(ariel): find out about this one
+          description: Not sure if this is the billing account email or the account email, but this might be unnecessary since that should be included in the Account or nested in the Account
           type: string
-        orgId:
```
Asalem1, line 1171: Removed orgId because that should already be accessible by the time the user accesses this endpoint. Meaning, they will be authenticated and the data can be obtained by IDPE

src/client/swagger/unity.yml
```diff
@@ -1262,9 +1239,9 @@ components:
         id:
           type: string
           description: ID of the org
-        idpeID:
+        quartzId:
```
Asalem1, line 1242: Not sure if `id` should be made into `idpeID` to be explicit

src/client/swagger/unity.yml
```diff
@@ -1277,10 +1254,35 @@ components:
...
+            - pay_as_you_go
+        accountBalance:
+          type: number
+          description: balance of the account
+        accountEmail:
+          type: string
+          description: billing contact email
+        accountId:
+          type: string
+          description: id of the account associated with the org
```
Asalem1, line 1272: besides the `accountType`, the rest of the account-specific data here is listed for the operator page's purposes. The data was flattened in order to remove the circular dependency that existed between accounts & orgs.

src/client/swagger/unity.yml
```diff
@@ -1277,10 +1254,35 @@ components:
...
+      required:
+        [
+          id,
+          quartzId,
+          name,
+          accountType,
+          accountBalance,
+          accountEmail,
+          accountId,
```
Asalem1, line 1281: Since most of the `account` data is only used by the operator, i wasn't sure if we wanted to make it required or not

src/index.tsx
```diff
@@ -75,19 +70,7 @@ class Root extends PureComponent {
...
-              {/* TODO(ariel): we need to restrict access to the checkout and operator pages based on specific critera:
-                https://github.com/influxdata/ui/issues/848
-               */}
-              {CLOUD && isFlagEnabled('unity-checkout') && (
-                <Route path="/checkout" component={CheckoutPage} />
-              )}
-              {/* Operator Routes */}
-              {CLOUD && isFlagEnabled('unity-operator') && (
-                <Route path="/operator" component={OperatorPage} />
-              )}
```
Asalem1, line 88: Moved these further down the pipeline since they should only be accessible when a user is authenticated

src/operator/api/index.ts
```diff
@@ -214,64 +163,20 @@ export const getOrgs = (
...
+    accountType: 'pay_as_you_go',
+    accountBalance: 10,
+    accountId: 'pay123',
+    accountEmail: 'paid@account.com',
   }
 
   return makeResponse(200, organization)
 }
 
-export const getMe = (): ReturnType<typeof apiGetMe> => {
```
Asalem1, line 245: Moved this to the `me` reducer since we're likely to use some of the meta data provided in that request when conditionally rendering the upgrade now button

src/shared/containers/GetOrganizations.tsx
```diff
@@ -14,37 +15,78 @@ import {RemoteDataState, AppState} from 'src/types'
...
+
+const canAccessCheckout = (me: Me): boolean => {
+  if (me?.account?.type === 'pay_as_you_go') {
+    return false
+  }
+  if (me?.account?.type === 'free' && !!me?.isBeta) {
+    return false
+  }
+  return true
+}
```
Asalem1, line 38: Pretty sure these are the rules, but is there anything else i'm missing in terms of restrictions for allowing access to the checkout page?

randycoulman (commented):
Looks really good. I have a few suggestions below. I'm specifically concerned about flatting the account fields up into org given that we also have an OEM API that uses the Organization entity.

src/client/swagger/unity.yml
Asalem1: Cool, I'm pretty sure we only care about the email so I'll go ahead and remove the unused properties

src/client/swagger/unity.yml
Asalem1: 🤔 I'm not sure I follow, what're "claims" in the session cookie? 

re: reducing the number of API calls to IDPE to obtain the orgID - when a user is authenticated, one of the first API requests we make is for the orgID (among a few other requests). In this case, we won't be adding on another request, simply taking advantage of the existing flow

src/client/swagger/unity.yml
Asalem1: Got it, so in that case, does it make sense for us to maintain `id` as a reference to the IDPE id with a separate property to the `quartzId`?

src/client/swagger/unity.yml
Asalem1: What's the OEM API? Totally makes sense to have a RelatedAccount type that can be nested inside, it didn't even occur to me that it would be an option :)

src/client/swagger/unity.yml
randycoulman: The session cookie is a JWT, and the contents of a JWT are called "claims".  As long as the API request returns the orgID that's in the session cookie, that should be fine.

src/client/swagger/unity.yml
randycoulman: Yes, but only when it makes sense to return a quartzId, I think.  In general, the operator UI works in terms of Quartz IDs because it can work with not-fully-onboarded users/accounts that don't yet have an IDPE id.

src/client/swagger/unity.yml
randycoulman: We have a public API used by OEMs like PTC and Schneider Electric. It allows them to add/remove orgs from their account (they have a single account, but create an org for each of their end customers) and do stuff with org limits.

Asalem1: Awww, gotcha.

src/me/actions/creators/index.ts
```diff
@@ -1,11 +1,31 @@
...
+    type: SET_QUARTZ_ME,
+    quartzMe: me,
+    status,
+  } as const)
+
+export const setQuartzMeStatus = (status: RemoteDataState) =>
+  ({
+    type: SET_QUARTZ_ME_STATUS,
+    status,
+  } as const)
```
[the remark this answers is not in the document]
Asalem1: I hear ya. I think the the latter is a consequence of having such a widespread network of actions that have caused accidental collisions in the past between dispatched actions with the same name. Since there isn't a centralized location of all actions, we have had situations where actions have mirrored the events that they are supposed to enact (e.g. UPDATE_TASK_NAME or whatever) but found out that there were unintended consequences due to an already existing action by the same name (that was a made up example of a real thing that happened early on in my time here where I named an action precisely based on what it was supposed to do and ended up encountering some weird redux issues where another action by the same name was triggered and the reducer state got out of wack). It's definitely worth considering a different model than the current codebase conventions to help with maintainability and for intended purposes rather than forcing it like we are here.

src/me/actions/thunks/index.ts
```diff
@@ -37,3 +40,19 @@ export const getMe = () => async dispatch => {
     console.error(error)
   }
 }
+
+export const getQuartzMe = () => async dispatch => {
+  try {
+    dispatch(setQuartzMeStatus(RemoteDataState.Loading))
+    const resp = await getMeQuartz()
```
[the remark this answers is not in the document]
Asalem1: I hear ya, I'll set an alias to make it clearer

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1096)

src/operator/api/index.ts
```diff
@@ -116,95 +115,51 @@ export const getOrgs = (
   const organizations: Organization[] = [
     {
       id: '123',
-      idpeID: '123',
+      quartzId: '123',
```
randycoulman, line 118: #nit Quartz IDs will be ints, not strings.

randycoulman (approved):
One little thing to look at, but otherwise LGTM!

Issue #1083, opened 2021-04-02
Make notebooks better


drdelambre:
general round of polish is required

 - if you're deeplinked to a notebook in presentation mode, it makes no sense. it just says "No Data". that is a useless interface. make the notebook run when you land on the page
 - notebook lists dont work on tools. there is no scrolling, no filtering, no way to get to the notebook you just made. add filtering and scrolling!
 - annotations breaks notebooks visualizations. seems like we just fixed this? but now it's broken again. people who are testing annotations will think that notebooks are broken, so we should go fix annotations
 - fix aggregate window. I tried to run a |> group () |> sum() query into the single stat visualization and everything broke. when i tried to remove the aggregate window, it wouldn't let me
 - theres a dropped edge case around loading states somewhere that causes the loading spinner to stay around forever on the SimpleTable results. Barbara can reliably express it when toggling between a 7 day and 30 day time range on this notebook: https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/notebooks/074fe35728422000

drdelambre:
the annotations aspect is covered in #1088

drdelambre:
just got the weird loading state issue left on this one

drdelambre:
the loading state seems to have faded away while cleaning up other issues. so i'm closing this one

PR #1098, opened 2021-04-02
feat: add filtering and sort to the flow list

drdelambre:
this helps long lists of flows work.. better.
migrates the filter and empty states from the buckets interface while adding a sort in there as well

OH

and now that page scrolls

part of #1083


src/flows/components/FlowsIndex.tsx
```diff
@@ -1,37 +1,96 @@
...
+  const {flows} = useContext(FlowListContext)
+  const [search, setSearch] = useState('')
+  const [sortOptions, setSortOptions] = useState({
+    sortKey: 'name' as keyof Flow,
+    sortType: SortTypes.String,
+    sortDirection: Sort.Ascending,
+  })
+
+  const filteredFlows = Object.keys(flows).filter(f =>
+    flows[f].name.toLowerCase().includes(search.toLowerCase().trim())
```
Asalem1, line 32: nit: `toLocaleLowercase()` I've seen that used elsewhere in the codebase. Should we standardize before we get a random `quick hits` issue?

Asalem1 (approved):
LGTM. But what're flows?

PR #1105, opened 2021-04-03
fix: disable writing annotation in dataExplorer

### LINKED ISSUE #1091: Annotations in Cells: Disable annotation writing in Data Explorer
### Background
During the testing for #1088, we realized that it doesn't make sense to have annotations on Data Explorer since they now depend on cell ids and there are no cell ids in Data Explorer.

### Acceptance Criteria
- Annotations are not writeable in Data Explorer
  - This might be as simple as doing a check if cellID exists when we go to write the annotation. If it doesn't don't even fire the click handler.
  - Editing a cell in the VEO should behave as normally, this only applies to writing a query.

ChitlangeSahas:
Closes #1091

We want to disable on-click-to-add-annotation behavior in the data explorer because it doesn't make sense anymore. We also  want to show the annotations in the VEO, the popover that lets you edit queries. 

This PR does that. Screen recording:

https://user-images.githubusercontent.com/18511823/113462506-13d5df80-93d6-11eb-9ab6-84982ff2c04d.mov

### DIFF
--- a/src/timeMachine/selectors/index.ts
+++ b/src/timeMachine/selectors/index.ts
@@ -53,7 +53,7 @@ export const getActiveTimeMachine = (state: AppState) => {
 }
 
 export const getActiveCellID = (state: AppState) => {
-  return getActiveTimeMachine(state).view.cellID
+  return getActiveTimeMachine(state)?.view.cellID
 }
 
 export const getActiveGraphType = (state: AppState): string => {
--- a/src/timeMachine/selectors/index.ts
+++ b/src/timeMachine/selectors/index.ts
@@ -52,6 +52,10 @@ export const getActiveTimeMachine = (state: AppState) => {
   return timeMachine
 }
 
+export const getActiveCellID = (state: AppState) => {
+  return get(getActiveTimeMachine(state), 'view.cellID')
+}
+
 export const getActiveGraphType = (state: AppState): string => {
   return get(getActiveTimeMachine(state), 'view.properties.type', 'xy')
 }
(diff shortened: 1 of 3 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1105)

src/visualization/types/Graph/view.tsx
```diff
@@ -270,7 +270,8 @@ const XYPlot: FC<Props> = ({
       }
     })
 
-    if (inAnnotationWriteMode) {
+    // only pass interaction handler to giraffe if annotations write mode is enabled and cellID is present
```
hoorayimhelping, line 273: i think the comment would be more useful if it explained why cell id needs to be present. it's pretty clear why `inAnnotationWriteMode` needs to be `true`, and it's clear what it means when it's true. it's not as clear what it means for the cell id to not exist

src/timeMachine/selectors/index.ts
```diff
@@ -52,6 +52,10 @@ export const getActiveTimeMachine = (state: AppState) => {
   return timeMachine
 }
 
+export const getActiveCellID = (state: AppState) => {
+  return get(getActiveTimeMachine(state), 'view.cellID')
```
hoorayimhelping, line 56: i understand that this is conventional for this file, but we're trying to move away from using lodash functions - native functions make it almost obsolete. so just grab the active time machine and access the view.cellID directly.

hoorayimhelping (changes requested):
the comment is usually not something i'd block on, but we're trying to move away from lodash, so let's ditch that

src/timeMachine/selectors/index.ts
ChitlangeSahas: Got it! Changing now.

hoorayimhelping (approved):
great work! gogogo!

PR #1063, opened 2021-03-31
feat: added clone naming for tasks

Asalem1:
Closes #883 

This PR adds clone naming conventions to tasks.

task-clone

PR #1110, opened 2021-04-05
feat: fixed name cloning to target the task name

### LINKED ISSUE #883: Tasks: Cloned task should have a different name than the original
There is an option to clone a task in the UI, but the name of the new task is the same as the old task, making it hard to distinguish.

other resources, when cloned, give the clone a "Clone 1" at the end. We should also do that with tasks.

Asalem1:
Closes #883 

This time it should close the issue for real. Updated the regex to properly assert replacing the task name declaration rather than the first string like it was in the previous PR (#1063)

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -227,13 +227,15 @@ export const createTask = (
   orgID?: string,
   name: string = '🦄ask'
 ): Cypress.Chainable<Cypress.Response> => {
-  const flux = `option task = {
-    name: "${name}",
-    every: 24h,
-    offset: 20m
-  }
-  from(bucket: "defbuck")
-        |> range(start: -2m)`
+  const flux = `import "csv"
+
+option task = {
+  name: "${name}",
+  every: 24h,
+  offset: 20m
+}
+from(bucket: "defbuck")
+      |> range(start: -2m)`
 
   return cy.request({
     method: 'POST',
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -270,12 +270,12 @@ export const cloneTask = (task: Task) => async (
     ast.body = ast.body.map(statement => {
       if (
         statement.type === 'OptionStatement' &&
-        statement.assignment.init &&
-        Array.isArray(statement.assignment.init.properties)
+        statement.assignment.init.type === 'ObjectExpression'
       ) {
         statement.assignment.init.properties = statement.assignment.init?.properties.map(
           property => {
             if (
+              property.key.type === 'Identifier' &&
               property.key.name === 'name' &&
               property?.value?.location?.source
             ) {
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -50,7 +50,8 @@ import {taskToTemplate} from 'src/shared/utils/resourceToTemplate'
 import {isLimitError} from 'src/cloud/utils/limits'
 import {checkTaskLimits} from 'src/cloud/actions/limits'
 import {getOrg} from 'src/organizations/selectors'
-import {getStatus} from 'src/resources/selectors'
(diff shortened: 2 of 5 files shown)


cypress/support/commands.ts
```diff
@@ -227,13 +227,15 @@ export const createTask = (
...
   name: string = '🦄ask'
 ): Cypress.Chainable<Cypress.Response> => {
-  const flux = `option task = {
-    name: "${name}",
-    every: 24h,
-    offset: 20m
-  }
-  from(bucket: "defbuck")
-        |> range(start: -2m)`
+  const flux = `import "csv"
```
Asalem1, line 230: Added a generic import here to complicate the task a bit and ensure that another string is declared prior to the task variable

src/tasks/actions/thunks.ts
```diff
@@ -244,15 +245,36 @@ export const deleteTask = (taskID: string) => async (
...
-    const newTask = await api.postTask({data: resp.data})
+    const taskName = resp.data.name
+    const tasks = getAll<Task>(state, ResourceType.Tasks)
+    const allTaskNames = tasks.map(d => d.name)
+    const clonedName = incrementCloneName(allTaskNames, taskName)
+    const flux = resp.data.flux
+
+    // Searches for the beginning option task declaration, ignoring whitespace prior to the task name
+    const revisedFlux = flux.replace(
+      /option task\s*=\s*{\s*name:\s*".*?"/,
```
Asalem1, line 268: updated the regex to assert a more direct match for a task option rather

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :question:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1110)

drdelambre:
i'm not following why we need to use regex for this.
we have all of the meta data required to define that option, and it's autogenerated from that metadata in the first place (as in.. it's not code that a user would particularly have interest in personalizing), so why dont we parse the whole task script into an AST, blow away that option's tree, and then rebuild the option definition and push it back into the AST?

that'd offload the parsing complexity / rules onto the flux parsing wasm and allow the flux team to control that complexity

I'm getting flashbacks from variable parsing issues that look similar to this

src/tasks/actions/thunks.ts
```diff
@@ -244,15 +247,55 @@ export const deleteTask = (taskID: string) => async (
...
+    const {flux} = resp.data
+
+    const ast = parse(flux)
+
+    ast.body = ast.body.map(statement => {
+      if (
+        statement.type === 'OptionStatement' &&
+        statement.assignment.init &&
+        Array.isArray(statement.assignment.init.properties)
+      ) {
```
Asalem1, line 275: didn't abstract these because loops + functions !== "i'm melting"

src/tasks/actions/thunks.ts
```diff
@@ -244,15 +247,55 @@ export const deleteTask = (taskID: string) => async (
...
+    const tasks = getAll<Task>(state, ResourceType.Tasks)
+    const allTaskNames = tasks.map(d => d.name)
+    const clonedName = incrementCloneName(allTaskNames, taskName)
+    const {flux} = resp.data
+
+    const ast = parse(flux)
+
+    ast.body = ast.body.map(statement => {
+      if (
+        statement.type === 'OptionStatement' &&
```
drdelambre, line 272: i think you also have to look for the option to be named `task` somewhere in here, or else it'll overwrite the variable `name` if:
```
option v = {
    name: 'yolo'
}

option task = {
    name: 'task to clone'
}
````

Asalem1: I think that's illegal in some states but better put up some guards

drdelambre (approved):

PR #1133, opened 2021-04-06
feat: display time as formatted value in single stat graph

ChitlangeSahas:
Closes #https://github.com/influxdata/influxdb/issues/18837

Our current single stat graph and single stat with line would show a time value dataType as epoch time but the numerical value. This PR enables it to display a much more user-friendly string representation of time in the globally configured timeZone.

**The code has been updated in the UI to display time in the timeFormat, now how do we fix it in the giraffe, is the current question.** 

Screen shots:

Screen Shot 2021-04-06 at 1 51 01 PM

respects the timezone selector: 

Screen Shot 2021-04-06 at 1 53 19 PM

for singleStatWithLine:

Screen Shot 2021-04-06 at 1 58 09 PM

### DIFF
--- a/src/visualization/types/SingleStat/view.tsx
+++ b/src/visualization/types/SingleStat/view.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FC} from 'react'
+import React, {FC, useContext} from 'react'
 
 // Utils
 import LatestValueTransform from 'src/visualization/components/LatestValueTransform'
@@ -18,6 +18,9 @@ import {
 
 import './style.scss'
 import {isFlagEnabled} from '../../../shared/utils/featureFlag'
+import {DEFAULT_TIME_FORMAT} from '../../../shared/constants'
+import {getFormatter} from '../../utils/getFormatter'
+import {AppSettingContext} from '../../../shared/contexts/app'
 
 interface Props extends VisualizationProps {
   properties: SingleStatViewProperties
@@ -25,6 +28,7 @@ interface Props extends VisualizationProps {
 
 const SingleStat: FC<Props> = ({properties, result}) => {
   const {prefix, suffix, colors, decimalPlaces} = properties
+  const {timeZone} = useContext(AppSettingContext)
 
   if (isFlagEnabled('useGiraffeGraphs')) {
     const latestValues = getLatestValues(result.table)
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
performance | 46.75 | :red_circle:
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1133)

src/visualization/types/SingleStatWithLine/view.tsx
```diff
@@ -266,11 +270,20 @@ const SingleStatWithLine: FC<Props> = ({properties, result, timeRange}) => {
...
-              decimalPlaces: properties.decimalPlaces,
-              prefix: properties.prefix,
-              suffix: properties.suffix,
+            const timeFormatter = getFormatter('time', {
+              timeZone: timeZone === 'Local' ? undefined : timeZone,
+              timeFormat: DEFAULT_TIME_FORMAT,
             })
+
+            const formattedValue =
+              result.table.getColumnType('_value') == 'time'
```
hoorayimhelping, line 279: it's a small thing, but it's better to use `===` in most cases. `==` should be reserved for when you want to communicate that casting may happen. 

```js
3 === '3' // false
'3' === '3' // true
3 == '3' // true
```

so in most cases, we want to use `===` and the only time we don't is when we explicitly want to allow casting, like when converting between values that might be represented as a string or a number. For example, in HTML forms numbers get wrapped in strings. If we compare a value in redux that's a number to the value in a form that is a string, we'll want to use `==`

src/visualization/types/SingleStat/view.tsx
```diff
@@ -74,11 +78,20 @@ const SingleStat: FC<Props> = ({properties, result}) => {
...
-            decimalPlaces,
-            prefix,
-            suffix,
+          const timeFormatter = getFormatter('time', {
+            timeZone: timeZone === 'Local' ? undefined : timeZone,
+            timeFormat: DEFAULT_TIME_FORMAT,
           })
+
+          const formattedValue =
+            result.table.getColumnType('_value') == 'time'
```
hoorayimhelping, line 87: same thing here about `===`

hoorayimhelping (approved):
lgtm. suggestion about changing `==` to `===` make that change and then test that this works and as long as it's good you can merge it without a re-review

PR #1146, opened 2021-04-08
chore: tweak text

### LINKED ISSUE #1089: Remove the phrasing "InfluxDB 2.0" from the UI
For cloud, we should use InfluxDB Cloud and for OSS, we should just use InfluxDB.

https://github.com/influxdata/ui/blob/master/src/index.html#L9
https://github.com/influxdata/ui/blob/master/assets/index.html#L10

Alert and Check Status pages: #1087 

https://github.com/influxdata/ui/blob/master/src/onboarding/components/CompletionStep.tsx#L119
https://github.com/influxdata/ui/blob/master/src/onboarding/components/CompletionStep.tsx#L63
https://github.com/influxdata/ui/blob/master/src/onboarding/constants/index.ts#L1

joelfried:
Closes #1089 

Remove "Cloud" from OSS only pages.


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1146)

src/onboarding/components/CompletionStep.tsx
```diff
@@ -60,7 +63,8 @@ class CompletionStep extends PureComponent<Props> {
             <div className="wizard-step--scroll-content">
               <h3 className="wizard-step--title">You are ready to go!</h3>
               <h5 className="wizard-step--sub-title">
-                Your InfluxDB Cloud has 1 organization, 1 user, and 1 bucket.
+                Your {this.environment} has 1 organization, 1 user, and 1
```
Asalem1, line 66:
```suggestion
                Your InfluxDB {CLOUD && 'Cloud '}has 1 organization, 1 user, and 1
```

src/onboarding/components/CompletionStep.tsx
```diff
@@ -50,6 +51,8 @@ class CompletionStep extends PureComponent<Props> {
     window.removeEventListener('keydown', this.handleKeydown)
   }
 
+  private environment = CLOUD ? 'InfluxDB Cloud' : 'InfluxDB'
```
Asalem1, line 54: We can consolidate this in the example below

src/onboarding/components/CompletionStep.tsx
Asalem1, line 66: I'm a little confused by this since it's been a while since i've onboarded, but does a CLOUD user ever hit this?

src/onboarding/components/InitStep.tsx
```diff
@@ -15,7 +15,7 @@ class InitStep extends PureComponent<OnboardingStepProps> {
       <div className="wizard--bookend-step">
         <div className="splash-logo primary" />
         <h3 className="wizard-step--title" data-testid="init-step--head-main">
-          Welcome to InfluxDB Cloud
+          Welcome to InfluxDB
```
Asalem1, line 18: wouldn't we want the same conditional logic as above?

src/onboarding/components/CompletionStep.tsx
joelfried: I'm not entirely certain but since we have an environment variable and it's a simple if statement it seemed easiest to just handle the potential edge case.  If we know that this never gets hit by Cloud I'm happy to hard code the text here.

joelfried: Just to have this here in case anybody else finds it later: these pages are OSS only.

src/onboarding/components/CompletionStep.tsx
```diff
@@ -116,8 +116,8 @@ class CompletionStep extends PureComponent<Props> {
                         />
                         <dt>I've got this...</dt>
                         <dd>
-                          Jump into InfluxDB Cloud and set up data collection
-                          when you’re ready.
+                          Jump into InfluxDB and set up data collection when
+                          you’re ready.
```
Asalem1, line 120: why are these on two separate lines? Also, the previous version had `when` on the second line?

joelfried: Prettier changed this to move "when" up because there were too few characters on line 119 after "Cloud" was removed.  It wasn't a conscious decision on my part.  You can see that the change was introduced in my commit "run prettier" - all I did on that commit was run `yarn prettier:fix`.

Asalem1 (approved):

Issue #831, opened 2021-03-12
(feat) Add profile option to notebooks


barbaranelson:
**Proposal:**
Flux supports profiling, to help users understand/troubleshoot the performance of their query, as described here: https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/profiler/

The challenge is that most flux users do not know about this profiler, so they aren't getting the benefit.  This feature would add an option into notebooks to expose the profiler, and make it easier for users to enable it.

**Current behavior:**
If you know about it, you can add the profiler option to your flux script. If you don't know about it, this is not an easy feature to discover.

**Desired behavior:**
Some way to easily enable the flux profiler and see the results, without needing to know what special options to include in your flux query.

**Use case:**
This helps customers understand the performance of their flux query, so they can improve the performance by adjusting their query.

**Note:**
Reach out to Nathaniel when you start working on this because there is also an API option to enable the profiler, so it can be done more transparently than the docs above describe.

russorat:
Here's an example of enabling this from the API:

```
curl --location --request POST 'https://us-west-2-1.aws.cloud2.influxdata.com/api/v2/query?orgID=$INFLUX_ORG' \
--header 'Authorization: Token $INFLUX_TOKEN' \
--header 'Content-Type: application/json' \
--data-raw '{
    "extern": {
        "type": "File",
        "imports": [
            {
                "type": "ImportDeclaration",
                "as": {
                    "type": "Identifier",
                    "name": "profiler"
                },
                "path": {
                    "type": "StringLiteral",
                    "value": "profiler"
                }
            }
        ],
        "body": [
            {
                "assignment": {
                    "member": {
                        "object": {
                            "name": "profiler",
                            "type": "Identifier"
                        },
                        "property": {
                            "name": "enabledProfilers",
                            "type": "Identifier"
                        },
                        "type": "MemberExpression"
                    },
                    "init": {
                        "type": "ArrayExpression",
                        "elements": [
                            {
                                "type": "StringLiteral",
                                "value": "query"
                            },
                            {
                                "type": "StringLiteral",
                                "value": "operator"
                            }
                        ]
                    },
                    "type": "MemberAssignment"
                },
                "type": "OptionStatement"
            }
        ]
    },
    "query": "buckets()",
    "type": "flux",
    "dialect": {
        "annotations": [
            "group",
            "datatype",
            "default"
        ]
    }
}'
```

russorat:
[...]

PR #1149, opened 2021-04-08
feat: add profiler toggle to data explorer

russorat:
#831 but for data explorer

This adds a profiler toggle option to the data explorer and cell editor overlay. Attempts to put it behind a `profile-query` feature flag, but i probably did that wrong.

BUG: for some reason, this disables the raw data toggle when editing a check, and i can't figure out why.

Kapture 2021-04-06 at 11 31 50

### DIFF
--- a/src/external/monaco.flux.server.ts
+++ b/src/external/monaco.flux.server.ts
@@ -20,7 +20,7 @@ import {
 import {registerCompletion} from 'src/external/monaco.flux.lsp'
 import {AppState, LocalStorage} from 'src/types'
 import {getAllVariables, asAssignment} from 'src/variables/selectors'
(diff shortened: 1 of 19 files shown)


drdelambre (changes requested):
the fact that most of these are hardcoded to false, and that the only thing that can turn it back to true is a reducer that is tightly coupled to a heavily nested redux store used only for representing the view state of the timemachine interface (which isn't the only thing that queries in the system), leads me to believe that hijacking the `buildVarsOption` wasn't the correct architectural choice here.

since it's a hardcoded AST partial, why not just unshift it onto the extern.body property created by buildVarsOption in src/timeMachine/actions/queries.ts?

PR #1160, opened 2021-04-09
chore: make dashboards filter term persist across pages

### LINKED ISSUE #844: Remember the filter for the list pages
the experience today is that you go to the tasks page (or dashboards page), use the filter to find the tasks you want, then click on one of them. When you click the back button in the browser, you are back to the page with no filters in place.

Users can have 100s of tasks, so i'd love to be able to save that filter and get back to it.

I propose we find a way to keep the filter in place when a user clicks the back button specifically for the tasks page.

subirjolly:
Closes #844 

**Before**
before

**After**
after

### DIFF
--- a/cypress/e2e/shared/dashboardsIndex.test.ts
+++ b/cypress/e2e/shared/dashboardsIndex.test.ts
@@ -481,6 +481,36 @@ describe('Dashboards', () => {
         cy.getByTestID('dashboard-card').should('have.length', 1)
         cy.contains(dashboardName)
       })
+
+      it('can persist search term', () => {
+        cy.getByTestID('dashboard-card').should('have.length', 2)
+
+        cy.getByTestID('search-widget').type(dashSearchName)
+
+        cy.getByTestID('dashboard-card').should('have.length', 1)
+
+        cy.getByTestID('dashboard-card')
+          .first()
+          .getByTestID('dashboard-card--name')
+          .should('contain', dashboardName)
+
+        // Navigate Away and come back using browser's back button
+        cy.getByTestID('dashboard-card')
+          .first()
+          .getByTestID('dashboard-card--name')
+          .click()
+        cy.wait(1000)
+        cy.go('back')
+
+        cy.getByTestID('search-widget').should('have.value', dashSearchName)
+
+        // Navigate Away and come back by clicking on Boards icon
+        cy.getByTestID('nav-item-tasks')
+          .click()
+          .wait(1000)
+        cy.getByTestID('nav-item-dashboards').click()
+        cy.getByTestID('search-widget').should('have.value', dashSearchName)
+      })
     })
 
     it('can list and search dashboards on home page', () => {
--- a/cypress/e2e/shared/dashboardsIndex.test.ts
+++ b/cypress/e2e/shared/dashboardsIndex.test.ts
@@ -499,15 +499,13 @@ describe('Dashboards', () => {
           .first()
           .getByTestID('dashboard-card--name')
           .click()
-        cy.wait(1000)
+        cy.getByTestID('page-title').contains(dashboardName)
         cy.go('back')
-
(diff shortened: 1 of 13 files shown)


src/dashboards/components/dashboard_index/DashboardsIndex.tsx
```diff
@@ -104,7 +95,7 @@ class DashboardIndex extends PureComponent<Props, State> {
               <GetAssetLimits>
                 <DashboardsIndexContents
                   searchTerm={searchTerm}
-                  onFilterChange={this.handleFilterDashboards}
+                  onFilterChange={this.props.setSearchTerm}
```
Asalem1, line 98: if the `onSearch` and `onFilterChange` are being mapped to props, there's no reason for us to prop drill. We could just map it out directly to the child

src/dashboards/components/dashboard_index/DashboardsIndex.tsx
```diff
@@ -169,16 +156,19 @@ const mstp = (state: AppState) => {
     cloud: {limits},
   } = state
   const sortOptions = state.resources.dashboards['sortOptions']
+  const searchTerm = state.resources.dashboards['searchTerm']
```
Asalem1, line 159: nit: We can probably consolidate these to a `getAll`, then just get the individual properties:

https://github.com/influxdata/ui/blob/master/src/resources/selectors/index.ts#L14

So something like:

```
const {sortOptions, searchTerm} = getAll<Dashboard>(
    state,
    ResourceType.Dashoards
  )
```

Asalem1 (changes requested):
Besides the comments, my concern with this approach is that we are now dispatching an event whenever we search through dashboards. This might cause us to rerender the DOM a bunch, which might cause a performance hit. Is there any chance we can get a performance chart to see the difference between this implementation and the original? I know the original isn't what we're striving for, but it might be possible for us to use that as a barometer for success here when measuring how performant this part of the app currently is vs how it fairs with the refactor

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -481,6 +481,36 @@ describe('Dashboards', () => {
...
+          .first()
+          .getByTestID('dashboard-card--name')
+          .should('contain', dashboardName)
+
+        // Navigate Away and come back using browser's back button
+        cy.getByTestID('dashboard-card')
+          .first()
+          .getByTestID('dashboard-card--name')
+          .click()
+        cy.wait(1000)
```
joelfried, line 502: Can we handle the navigation without these hard coded wait here and below?  (And looking [here](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting) I'm not sure if both of these are necessary.  Perhaps you need to add a cy.get after the click on 501 before you go back though? Not sure on that.)

joelfried (approved):

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:
best-practices | 91 | :green_circle:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1160)

src/dashboards/reducers/dashboards.ts
```diff
@@ -150,6 +152,14 @@ export const dashboardsReducer = (
 
         return
       }
+
+      case SET_DASHBOARDS_SEARCH_TERM: {
+        const {searchTerm} = action
+
+        draftState['searchTerm'] = searchTerm
```
Asalem1, line 159: nit: for stylistic consistencies let's use dot notation

Asalem1 (approved):
Looks great. Happy to get this in first thing Monday  morning

PR #1184, opened 2021-04-12
feat: log the user out of auth0 and redirect to IDPE after

### LINKED ISSUE #1182: Logging out should also logout of Auth0 directly from the UI
This one is pretty straightforward, but since we no longer want the UI to be redirecting to Quartz after logging out, we can simply call the auth0-js client library's logout method, then pass in the existing logout as the redirectTo link to ensure that the session in Auth0 is ended, and the IDPE session is destroyed.

Asalem1:
Closes #1182 

This PR is a feature flagged implementation for the UI's logout functionality that will drastically simplify the process involved in logging a user out. However, this is not fully feature ready as we still need to whitelist the returnTo URL in the tenant's app-wide [Allowed Logout URLs as per the docs](https://auth0.github.io/auth0.js/global.html#logout)

### DIFF
--- a/src/Logout.tsx
+++ b/src/Logout.tsx
@@ -2,12 +2,19 @@
 import {FC, useEffect} from 'react'
 import {useDispatch} from 'react-redux'
 import {RouteComponentProps} from 'react-router-dom'
+import auth0js from 'auth0-js'
 
 // APIs
 import {postSignout} from 'src/client'
 
 // Constants
-import {CLOUD, CLOUD_URL, CLOUD_LOGOUT_PATH} from 'src/shared/constants'
+import {
+  CLOUD,
+  CLOUD_URL,
+  CLOUD_LOGOUT_PATH,
+  CLOUD_SIGNOUT_PATHNAME,
+} from 'src/shared/constants'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Components
 import {reset} from 'src/shared/actions/flags'
@@ -19,6 +26,18 @@ const Logout: FC<Props> = ({history}) => {
 
   useEffect(() => {
     const handleSignOut = async () => {
+      if (CLOUD && isFlagEnabled('authSessionCookieOn')) {
+        const signoutUrl = new URL(
+          `${window.location.origin}${CLOUD_SIGNOUT_PATHNAME}`
+        )
+        const auth0 = new auth0js.WebAuth({
+          returnTo: signoutUrl.href,
+        })
+        // Log the user out of Auth0, then destroy their API session
+        auth0.logout()
+        return
+      }
+
       if (CLOUD) {
         window.location.href = `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
         return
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
accessibility | 72 | :yellow_circle:
seo | 67 | :yellow_circle:
pwa | null | :hankey:
best-practices | 69 | :yellow_circle:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1184)

randycoulman (changes requested):
My understanding was that the `/api/v2/signout` endpoint in IDPE already does the Auth0 logout. If so, there should be no need to do that here.

Asalem1:
> My understanding was that the `/api/v2/signout` endpoint in IDPE already does the Auth0 logout. If so, there should be no need to do that here.

I might need to pair with @GeorgeMac about that since I'm not sure if the comment was left there as a note to self by Goller, since the UI's initial implementation should be making a request to Quartz, which makes a request to Auth0, which makes a request to IDPE, right?

randycoulman:
I'm not sure which comment you're referring to here.

The original UI implementation logged out of Quartz, which then hit the IDPE endpoint, which then (was supposed to) log out of Auth0. Quartz DOES NOT log out of Auth0 once the user is fully onboarded.

Asalem1:
> I'm not sure which comment you're referring to here.
> 
> The original UI implementation logged out of Quartz, which then hit the IDPE endpoint, which then (was supposed to) log out of Auth0. Quartz DOES NOT log out of Auth0 once the user is fully onboarded.

Aw, very good to know. Ok, I think I got my wires crossed then. K, so it seems like all we might have to do is simply call the IDPE endpoint and call it a day, without doing anything else 🤞🏽

src/Logout.tsx
```diff
@@ -19,6 +25,14 @@ const Logout: FC<Props> = ({history}) => {
...
+        const url = new URL(
+          `${window.location.origin}${CLOUD_SIGNOUT_PATHNAME}`
+        )
+        window.location.href = url.href
+        return
+      }
+
       if (CLOUD) {
         window.location.href = `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
         return
```
randycoulman, line 38:
```suggestion
      if (CLOUD) {
       const url = isFlagEnabled('authSessionCookieOn')
          ? `${window.location.origin}${CLOUD_SIGNOUT_PATHNAME}`
          : `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
        
        window.location.href = url
        return
      }
```

or something like that.  #optional

randycoulman (approved):
I think this looks right now, but let's confirm with @GeorgeMac that the signout endpoint does what we think it does in terms of dropping the cookie and hitting Auth0.

GeorgeMac:
This looks great to me! Makes more sense given where we're going with it.
Navigate to `/api/v2/signout` instead of going to Quartz? (Just to clarify I am reading it correctly).

Asalem1:
> This looks great to me! Makes more sense given where we're going with it.
> Navigate to `/api/v2/signout` instead of going to Quartz? (Just to clarify I am reading it correctly).

That's correct. We just need to verify that the `api/v2/signout` endpoint is logging the user out of Auth0 & destroying the session

Issue #1054, opened 2021-03-31
Move docker image spec into this repo


drdelambre:
We've had some issues with updating the production image of this service over the past couple months, all that could have been prevented by more automated testing.

right now our docker definitions live in monitor-ci. this repo isn't set up with the whole "you test me, i'll test you" relationship that exists between idpe and ui, or soon ui and oss, and instead has relied on the developer to test thoroughly before merging. As these systems have evolved in complexity, it has become harder to track their interdependence. so.. effectively no testing on the definitions of how tests are run and the images that are built for ui images and used across all testing stacks.

we _could_ ensure that monitor-ci ALSO runs all these tests on every change... but i suggest we try something else

and have the production ui image + cypress e2e image definitions live in this repo (ui), so that all changes to those files are tested with the service they define.

PR #1204, opened 2021-04-14
feat: containerized the UI as an independent self-contained service

Asalem1:
Related to #1054 

This PR creates a UI-specific docker container that's isolated to the UI. This works with KIND local development


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1204)

genehynson:
should we include updates to the CI pipeline in this PR? Or will that be a future one?

Asalem1:
> should we include updates to the CI pipeline in this PR? Or will that be a future one?

Updates to the CI pipeline meaning updates to monitor-ci?

genehynson:
> > should we include updates to the CI pipeline in this PR? Or will that be a future one?
> 
> Updates to the CI pipeline meaning updates to monitor-ci?

Right now the CI pipeline uses monitor-ci to build the image. I believe with this change the pipeline should now use the UI repo to build the image, right?

Asalem1:
> > > should we include updates to the CI pipeline in this PR? Or will that be a future one?
> > 
> > 
> > Updates to the CI pipeline meaning updates to monitor-ci?
> 
> Right now the CI pipeline uses monitor-ci to build the image. I believe with this change the pipeline should now use the UI repo to build the image, right?

Totally. That’s the next part of this once we get it in

Makefile
```diff
@@ -0,0 +1,53 @@
+include ./.env
+export $(shell sed 's/=.*//' ./.env)
+
+UI_GIT_SHA = $(shell git log -1 --pretty=format:"%H")
+
+start:
+	# creates the image
+	# creates the container
+	# runs that container
+	docker build -t local/chronograf:latest -f docker/Dockerfile.chronograf .
```
drdelambre, line 10: should remove this line and change the start target to
```
start: build
     #just run it now and it'll call "make build" before it starts
```

Makefile
```diff
@@ -0,0 +1,53 @@
...
+build:
+	# for forcing a rebuild of a node
+	docker build -t local/chronograf:latest -f docker/Dockerfile.chronograf.prod .
+
+####### EVERYTHING UNDER THIS LINE IS APPLICATION SPECIFIC
+.PHONY: test
+
+# running `make pretty` will run prettier on your frontend code
+pretty:
+	docker exec slowmograf npx prettier --config .prettierrc.json --write '{src,cypress}/**/*.{ts,tsx,scss}'
```
drdelambre, line 39: can we change this to docker run in case the service isn't already running? and add all of the volume mounts on here as well (i know.. compose is pretty dope)

Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+# running `make test` will run the unit tests in watch mode
+test:
+	docker exec slowmograf yarn test --watchAll
```
drdelambre, line 43: same request for this target (or remove it)

Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+# running `make tsc` will run the typescript validator in watch mode
+tsc:
+	docker exec slowmograf yarn tsc --watch
```
drdelambre, line 47: same request for this target (or remove it)

docker/nginx.conf
```diff
@@ -0,0 +1,217 @@
...
+      stub_status;
+      allow 127.0.0.1;
+      deny all;
+    }
+
+    location / {
+      try_files $uri /index.html;
+    }
+
+    # NB: ALL CSP headers are in default because otherwise for each type of source you need to fully specify allowed domains
```
drdelambre, line 150: should we remove these for this image in case community members want to use it? seems pretty specific to what we are doing. we can use an `include` directive like we do for ingress rules... so it'd look like "include(./csp.conf)" or something, and we copy that file into the environment before building in monitor-ci

env.example
```diff
@@ -0,0 +1,25 @@
+# this is the externally available public path
+# that connects the client dev server to the correct
+# host, in case you have to override it
+INGRESS_HOST=kubernetes.docker.internal
```
drdelambre, line 4: i dont think we need this?

Makefile
Asalem1: isn't the purpose of `make build` to build the production image while `make start` is used for development (i.e. Dockerfile.choronograf)?

Makefile
Asalem1: sounds good

docker/nginx.conf
Asalem1: Can we sync on this just so I'm a little clearer on the implementation?

Makefile
drdelambre: ooo i missed that appendix

env.example
```diff
@@ -0,0 +1,24 @@
...
+# that connects the client dev server to the correct
+# host, in case you have to override it
+BASE_PATH=/
+API_BASE_PATH=/
+STATIC_DIRECTORY=/static
+PUBLIC=https://twodotoh.a.influxcloud.dev.local/
+UI_SHA=local_dev_mode
+CLOUD_URL=/auth
+
+# NOTE: self signed certs aren't currently supported with
```
drdelambre, line 11: there's no ssl in this image, so these need to be cleaned up

env.example
```diff
@@ -0,0 +1,24 @@
...
+# the gateway oidc provider plugin
+# PORT_HTTP=8080
+# PORT_HTTPS=8443
+PORT_HTTPS=8080
+PORT=8080
+
+INFLUXDB_ROOT=~/ui
```
drdelambre, line 18: all of these are monitor-ci specific, not related to the running / building of this service

docker/nginx.conf
```diff
@@ -128,15 +129,15 @@ http {
   }
 
   server {
-    listen [::]:8080;
-    listen 8080;
+    listen [::]:9999;
```
drdelambre, line 132: this will break the image that gets built, as the production deployment is looking at 8080. making this configurable with an env var is probably the fastest method of resolving this, but you might want to look at the kind config and see if it's already configurable over there

docker/nginx.conf
```diff
@@ -128,15 +129,15 @@ http {
...
+    listen 9999;
 
     server_name _;
 
     # Path for static files
     root /usr/share/nginx/html;
 
-    location /status {
+    location /nginx_status {
```
drdelambre, line 140: why'd you change this?

docker/includes/csp.conf
```diff
@@ -0,0 +1,67 @@
+
```
drdelambre, line 1: this file isn't supposed to be in this repo

docker/nginx.conf
Asalem1: I'm super confused. This file was copy pasta'd in. IIRC, the original wasn't being used for anything so I think the previous port of 8080 was a vestigial port. I don't think you have a point about the env var resolving this though

docker/nginx.conf
Asalem1: all of the changes here were copied over from the monitor-ci repo. If you'll recall, all of the old nginx.conf files were deleted. This shows up as a diff in github because the file name is the same (i imagine)

env.example
Asalem1: cool 🔥

env.example
Asalem1: Ya, I think I was confused by what was needed and what wasn't. I'll clean these out

docker/includes/csp.conf
Asalem1: Right, this needs to be removed. Still testing out to see whether the build will run in KIND with it.

docker/nginx.conf
```diff
@@ -128,15 +129,15 @@ http {
   }
 
   server {
-    listen [::]:8080;
-    listen 8080;
+    listen [::]:${PORT_HTTPS};
+    listen ${PORT_HTTPS};
```
drdelambre, line 133: you didn't test this

env.example
```diff
@@ -0,0 +1,15 @@
...
+# that connects the client dev server to the correct
+# host, in case you have to override it
+BASE_PATH=/
+API_BASE_PATH=/
+STATIC_DIRECTORY=/static
+PUBLIC=https://twodotoh.a.influxcloud.dev.local/
+UI_SHA=local_dev_mode
+CLOUD_URL=/auth
+
+# the gateway oidc provider plugin
```
drdelambre, line 11: can get rid of all these comments, and change all references of `PORT_HTTPS` to `PORT`? It made sense in monitor-ci when it references the ingress port (which might be different from the service port), but not so much here when it's just the service

env.example
```diff
@@ -0,0 +1,15 @@
+# this is the externally available public path
+# that connects the client dev server to the correct
+# host, in case you have to override it
+BASE_PATH=/
+API_BASE_PATH=/
```
drdelambre, line 5: can add comments for the rest of these too?

drdelambre: just copy pasta from the dockerfile should be fine

drdelambre (approved):
:fire:

PR #1219, opened 2021-04-15
chore: clean up the unity.yml

Asalem1:
This PR updates the `unity.yml` to remove a reference to the `/account` endpoint and adds a `billingProvider` property to the `Me` property to help clean up some excessive API requests


src/billing/components/BillingPageContents.tsx
```diff
@@ -1,22 +1,23 @@
...
+import {getQuartzMe} from 'src/me/selectors'
 
 const BillingPageContents: FC = () => {
-  const [{account}] = useBilling()
+  const quartzMe = useSelector(getQuartzMe)
 
-  if (!!account.marketplace) {
+  if (!!quartzMe.billingProvider) {
     return <MarketplaceBilling />
   }
```
ischolten, line 18: a payg account will always have a billing provider bc it represents who we send billing info to. This should probably be something more like below to check if its a marketplace.

```javascript
if (!!quartzMe.accountType == 'pay_as_you_go' && quartzMe.billingProvider !== 'zuora') {
  return <MarketplaceBilling />
}
```

src/billing/components/marketplace/MarketplaceLink.tsx
```diff
@@ -1,18 +1,20 @@
...
+// Types
+import {Me} from 'src/client/unityRoutes'
 
 // Utils
-import {useBilling} from 'src/billing/components/BillingPage'
+import {getQuartzMe} from 'src/me/selectors'
 
 const buttonInfo = (
-  marketplace: Marketplace['shortName']
+  marketplace: Me['billingProvider']
```
ischolten, line 17: looks like button info has all these links hard coded. We have an endpoint for getting this information. `/api/v2/marketplace`; can that be used here instead of these hard codings in case they change? The idea is that we wouldnt have to make the same change in both quartz and the ui.

ischolten (changes requested):
Looks great! Main change is to make sure we're not always rendering the marketplace version of the billing page if the account is payg.

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1219)

src/billing/components/marketplace/MarketplaceLink.tsx
Asalem1: Definitely. Pretty sure this is based on the Quartz UI's implementation. If that endpoint is the `marketplace` one, then we can change this up to make a request to that endpoint for that data. Did we want to make the `marketplace` endpoint take a "shortName" so that we can make a request based on a region?

ischolten: Yea this is how quartz implements it. But we dont want to now have 2 sources of truth for this information when it gets moved over to the ui bc marketplace link is still going to exist in quartz. We use it in onboarding pages. 

The /marketplace endpoint should need to take anything. It will use lookup the marketplace for the account for the user in the session.

src/billing/components/BillingPageContents.tsx
```diff
@@ -1,22 +1,23 @@
...
 // Utils
-import {useBilling} from 'src/billing/components/BillingPage'
+import {getQuartzMe} from 'src/me/selectors'
 
 const BillingPageContents: FC = () => {
-  const [{account}] = useBilling()
+  const quartzMe = useSelector(getQuartzMe)
 
-  if (!!account.marketplace) {
+  if (!!quartzMe.billingProvider) {
```
ischolten, line 16:
```suggestion
  if (quartzMe.accountType === 'pay_as_you_go' && quartzMe.billingProvider !== 'zuora') {
```

src/billing/components/marketplace/MarketplaceLink.tsx
Asalem1: Aw, perfect so we can just use the marketplace API here. I'll update it

src/billing/components/marketplace/MarketplaceLink.tsx
```diff
@@ -6,43 +6,32 @@ import {
...
-  marketplace: Marketplace['shortName']
-): {link: string; text: string} => {
+const buttonInfo = (marketplace: Marketplace['shortName']): string => {
   switch (marketplace) {
     case 'aws':
-      return {
-        link: 'https://aws.amazon.com/marketplace/library/',
-        text: 'AWS Subscription Portal',
-      }
+      return 'AWS Subscription Portal'
```
ischolten, line 18: nit: marketplace will have a `name` field on it that should be respectively `AWS`, `Google Cloud` and `Microsoft Azure`. Not sure if its worth using those here, but just a thought.

ischolten (approved):
Looks awesome!

PR #1226, opened 2021-04-16
feat: re-enable auto refresh with new settings and add pause cell

### LINKED ISSUE #1220: Dashboard Auto Refresh: Re-Enable Auto Refresh In Dashboards - Using Modal
## Context

After removing the auto refresh feature from dashboards due to the complaints coming from users about leaving dashboards and coming back to discover charges they didn't expect, its time to re-enable this feature and place safeguards around it such that users are more directly engaged with and top of the process around setting/disabling auto refresh. 

### Acceptance Criteria 
1. The dashboard auto refresh is no longer a one step process, but a two step process with a modal for the settings 
              ** Given:** I am a user that wants to enable auto refresh in my dashboards
              ** When: ** I go to enable auto refresh
              ** Then: ** On click, a modal opens with the settings we will present to the user
2. When the user is in the modal, they can select settings that control different safeguards, including an inactivity timeout, the duration of 
        auto refresh, and the interval at which the refreshes will occur
            ** Given: ** I am a user enabling auto refresh 
            ** When: ** I have the modal open that controls the settings for auto refresh in dashboards
            ** Then: ** I have: 
                                       - A dropdown for refresh frequency interval 
                                       - A dropdown for duration (which should be uni-directional, allowing selection from "now" and onward)
                                       - A dropdown for inactivity timeout, which when enabled expands to enable selecting frequency, available as 
                                            hours or days
                                       - A cancel button, allowing me to close the modal 
                                       - An "Enable" button, which confirms these settings and sets them in redux
3. When the refresh settings are such that auto-refresh is active, pressing the auto refresh button will disable the process manually
            ** Given: ** I am a user who has auto refresh active
            ** When: ** I am looking at the dashboard view 
            ** Then: ** I see a button that is now red, and says "Stop Auto Refresh," which when clicked will stop the auto refresh process and 
                                reset the settings

### LINKED ISSUE #1222: Dashboard Auto Refresh: Wire Up Auto Refresh Modal to Redux
** Context ** 
Following the work in  #1220 to create the presentational interface for the auto refresh modal and it's various options, as well as the trigger mechanism for the modal / stopping the auto refresh process, we need to wire up the modal to redux, allowing us to persist state for the intended behavior 

** Acceptance Criteria **
  1. The settings in the auto refresh modal are wired up to either context for abstracted maintenance of state, or redux entirely
          **Given** I am a user updating my settings for enabling auto refresh
          **When:** I update settings in the auto refresh modal
          **Then:** These settings are managed in redux 
  2. The auto refresh state in redux tracks the auto refresh process
          **Given** I am a user who has enabled auto refresh
          **When:** I am interacting with the dashboard
          **Then:** My auto refresh settings are maintained in redux and not lost when the modal is closed

### LINKED ISSUE #1223: Dashboard Auto Refresh: Enable Inactivity Timeout
### Context

One of the primary concerns around enabling auto refresh for dashboards is the potential for dashboards to be left open and query totals to quickly get out of control. One of the ways we can alleviate this concern is by providing an option to have an inactivity timeout, which will allow the user to set a time after which, if the user has not made any actions determined as "activity" in our site, the dashboard auto refresh settings will automatically reset to their default state of off. This should allow the user greater control over their costs and prevent similar issues as previously encountered. 

### Current List of "Activity"
- Mouse move
- Key Press
- Page Load (e.g hitting f5, reload)
- Change browser tab ( **investigate this behavior** - what happens when the tab w/ dashboard is now in the background?)

### Acceptance Criteria 
 In order to determine what constitutes "activity," we will provide an initial set of assumptions, and iterate if necessary / feedback is provided that encourages changes to our understanding.
Thus, the following criteria develop: 
     1. **Given:** I am a user that has enabled auto refresh on my dashboards
     2. **When:** I leave my keyboard/mouse unattended for the time set in the form in the modal for auto refresh settings
     3. **Then:** My auto refresh settings will reset to default of off, preventing unnecessary queries

 **Given** I am a user who has left the dashboards page
**When** I perform any of the actions listed in "activities" above
  **Then** The event handlers tracking them are not fired

### LINKED ISSUE #1224: Dashboard Auto Refresh: Enable Pause Cell Feature
### Context

Within the auto refresh process, it might be desirable to have the ability to pause a cell, thereby removing it from the auto refresh cycle. This becomes particularly pertinent when different queries within a dashboard track varying metrics that require differing levels of specificity / attention. If something is relevant information, but only needs to be seen once every hour or so for freshness, it does not need to be part of the same refresh cycle, but would still be desirable as part of the dashboard. 

### Additional Details 
The cell refresh process currently involves checking the query cache and invalidating the query there before querying the backend: if paused, we don't want to invalidate the query in the cache, but skip over it. This will also apply to dashboard wide refreshes, so pausing should prevent a blanket clearing of the cache. 

### Acceptance Criteria
In order for this to be unobtrusive but useful, it will need to be a part of the cell configuration menu, available in the top right corner of the cell. 
Furthermore, it will need to be behind a feature flag, so that it can be dealt with separately and offered to users in combination with, or separately from, the refresh feature we are introducing, and so that we can gauge how users appreciate (or dont) the availability of this feature. 
**Given** I am a user _with the feature flag for pausing cells enabled_,
**When** I click pause cell in the configuration menu of a cell
 **Then** The cell does not refresh, and only does so when the user clicks "resume" 

### Future Considerations
Currently, because this epic is constrained to front end implementations only, hard refreshes will mean losing all of these settings. This is something we need to keep in mind, because it may become a pain point for users.

ShmuelLotman:
Closes #1220
Closes #1222 
Closes #1223 

### NOTE
. Per discussion with, and confirmation from, @kristinarobinson, this PR will contain all of the necessary changes for the implementation of the auto refresh feature, and a [separate PR](https://github.com/influxdata/ui/pull/1240) has been opened to handle the Pause Cell feature. 
### /Note

This PR introduces a number of specific implementations that address the acceptance criteria listed in the four tickets above. Although no tests are provided within this ticket specifically, a concurrent ticket will be worked on to address adding comprehensive test coverage to these features, and cleanup work will be done to ensure any hidden gotchas are addressed as review proceeds. 

**Small Scale Implementations** 
        A. The time/date range picker now has an option to go in only one direction, by passing in a "singleDirection" prop
        B. The auto refresh dropdown now also has the ability to show hide the automatic refresh, not only the manual 
        C. Minor cleanup and improvements to the cell refresh functionality, allowing it to be tied together more clearly with the global refresher

**Main Work**
        A. Auto Refresh Modal with Trigger Button that also stops auto refresh when on and clicked
        B. Enabling Auto Refresh
        C. Inactivity Timeout
        D. Duration for Auto Refresh
        E. Interval for Auto Refresh
        F. No queries run when dashboard not being viewed
        G. Auto Refresh is dashboard specific 
        H. Auto Refresh watches timeout when dashboard is in background / not visible
        I. Option to select infinite timeout

The gifs below demonstrate the functionality added within this PR. This is all behind a "new-auto-refresh" feature flag for the auto refresh functionality. 

Dashboard Specific Auto Refresh: https://imgur.com/RhrT88n
Duration Functionality: https://imgur.com/IH61WIU
Demo 8 second inactivity timeout: https://imgur.com/kFsCg9P
No Wasted Queries If Not Within Dashboard: https://imgur.com/RbSeT7D


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1226)

src/dashboards/components/AutoRefresh.scss
```diff
@@ -0,0 +1,46 @@
+.refresh-form-container {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 10px 50px;
```
drdelambre, line 5: clockface defines a bunch of standard paddings you should use
I bring it up because it seems like you're using a 5px grid system when our design is built around an 8px grid
using the clockface variables puts it squarely in the "not my problem" bucket

src/dashboards/components/AutoRefreshOverlay.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const status =
+      milliseconds === 0 ? AutoRefreshStatus.Paused : AutoRefreshStatus.Active
+    setRefreshContext({
+      type: 'SET_REFRESH_MILLISECONDS',
+      refreshMilliseconds: {status, interval: milliseconds},
+    })
+  }
+
+  // Because we are not using the manual refresh in the dropdown, this function is a no-op
+  const noop = useCallback((): void => {}, [])
```
drdelambre, line 45: this is all sorts of ugly. make onManualRefresh an optional param on the AutoRefreshDropdown. you can get rid of the showManualRefresh prop as well. just dont show it if you haven't passed in the optional parameter

src/dashboards/components/AutoRefreshOverlay.tsx
```diff
@@ -0,0 +1,141 @@
...
+  return (
+    <Overlay.Container maxWidth={500} testID="auto-refresh-overlay">
+      <Overlay.Header title="Configure Auto Refresh" onDismiss={onClose} />
+      <Grid>
+        <Grid.Column className="refresh-form-column">
+          <div className="refresh-form-container">
+            <span>Select Refresh Frequency: </span>
+            <AutoRefreshDropdown
+              onChoose={handleChooseAutoRefresh}
```
drdelambre, line 54: i can't believe this isn't `onSelect'. out of the scope of this pr.. just super odd

src/dashboards/components/AutoRefreshOverlay.tsx
```diff
@@ -0,0 +1,141 @@
...
+import {CustomTimeRange, AutoRefreshStatus} from 'src/types'
+
+// Context
+import AutoRefreshContextProvider, {
+  AutoRefreshContext,
+} from 'src/dashboards/components/RefreshContext'
+
+import './AutoRefresh.scss'
+
+const INACTIVITY_ARRAY = [...Array(25).keys()].map(num => num.toString())
```
drdelambre, line 26: put a comment on this for what it does. cleverness is rarely rewarded as a code base lives on

src/dashboards/components/AutoRefreshOverlay.tsx
```diff
@@ -0,0 +1,141 @@
...
+                onSelect={(timeout: string) =>
+                  setRefreshContext({
+                    type: 'SET_INACTIVITY_TIMEOUT',
+                    inactivityTimeout: timeout,
+                  })
+                }
+                buttonColor={ComponentColor.Default}
+                testID="inactivity-timeout-dropdown"
+              />
+              {state.inactivityTimeout !== 'None' && (
```
drdelambre, line 94: this isn't a design paradigm we have introduced yet. i'd err to figuring out how to use the `10d 1h 30m` time format that we use everywhere else in the app in the name of consistency, but get the sign off of daniel and russ that it's ok to introduce time input in a different way, and that this is the way

src/dashboards/components/DashboardContainer.tsx
```diff
@@ -26,12 +26,61 @@ const {Active} = AutoRefreshStatus
...
+    if (autoRefresh?.duration?.upper <= new Date().toISOString()) {
+      GlobalAutoRefresher.stopPolling()
+      dispatch(resetDashboardAutoRefresh(dashboard))
+    }
+  }, [dashboard, dispatch, autoRefresh?.duration?.upper])
+
   useEffect(() => {
     if (autoRefresh.status === Active) {
-      GlobalAutoRefresher.poll(autoRefresh.interval)
+      GlobalAutoRefresher.poll(autoRefresh, stopFunc)
```
drdelambre, line 83: i was on a big rant about how you should be able to inspect state in redux about this refresher, then to my dismay saw that i wrote it, so this is fine

src/dashboards/components/RefreshContext.tsx
```diff
@@ -0,0 +1,137 @@
...
+  override = {}
+): AutoRefreshState => {
+  return {
+    inactivityTimeout: 'None',
+    inactivityTimeoutCategory: 'Hours',
+    duration: {
+      lower: new Date().toISOString(),
+      upper: new Date().toISOString(),
+      type: 'custom',
+    } as CustomTimeRange,
```
drdelambre, line 49: i've been bitten by using this instead of selectabletime range, or whatever the parent class is of this, when it came to saving and integrating into other parts of the app. bite the bullet and figure out the typescript error now, promise it's worth it

src/dashboards/components/RefreshContext.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const [state, dispatch] = useReducer(
+    autoRefreshReducer,
+    createAutoRefreshInitialState()
+  )
+
+  const {currentDashboardId} = useSelector((appState: AppState) => ({
+    currentDashboardId: appState.currentDashboard.id,
+  }))
+
+  const reduxDispatch = useDispatch()
```
drdelambre, line 91: why isn't this named `dispatch` like every other instance of this pattern in the codebase?

src/shared/apis/queryCache.ts
```diff
@@ -124,19 +125,29 @@ class QueryCache {
...
   }
 
   resetCache = (): void => {
-    this.cache = {}
+    for (const key of Object.keys(this.cache)) {
+      if (!this.cache[key].paused) {
+        delete this.cache[key]
+      }
+    }
+    // this.cache = {}
```
drdelambre, line 146: what is this

src/shared/apis/queryCache.ts
```diff
@@ -183,6 +194,33 @@ export const resetQueryCacheByQuery = (query: string): void => {
   queryCache.resetCacheByID(queryID)
 }
 
+export const getFromQueryCacheByQuery = (
```
drdelambre, line 197: i'm too lazy to find where you copypasta'd this from. why the pasta, and not just update the original definition of the function?

src/shared/components/cells/CellContext.tsx
```diff
@@ -42,11 +55,19 @@ const CellContext: FC<Props> = ({
...
   onCloneCell,
   onDeleteCell,
   onCSVDownload,
   onRefresh,
 }) => {
+  const viewWithQueries = view as View<
+    Exclude<ViewProperties, MarkdownViewProperties>
+  >
+  const foundQueries =
+    Array.isArray(viewWithQueries?.properties?.queries) &&
```
drdelambre, line 68: reference `view` instead of `viewWithQueries` here

src/shared/components/cells/CellContext.tsx
```diff
@@ -196,11 +234,14 @@ const CellContext: FC<Props> = ({
   )
 }
 
+const mstp = (state: AppState) => ({
+  variables: getAllVariables(state),
```
drdelambre, line 238: there's a code smell here that you have to pull variables in just to pass them back down to the cache like that

src/views/actions/thunks.ts
```diff
@@ -127,7 +127,7 @@ export const getViewAndResultsForVEO = (
       })
     )
 
-    const queries = view.properties.queries.filter(({text}) => !!text.trim())
+    const queries = view.properties.queries.filter(({text}) => !!text)
```
drdelambre, line 130: PUT IT BACK

drdelambre (changes requested):
- [ ] clockface variables
 - [ ] cleanup weird prop interface additions (showManualRefresh)
 - [ ] learn how to name things (reduxDispatch)
 - [ ] be consistent with interval input (because a product is built on consistency)
 - [ ] rethink how you touched the cache
 - [ ] GET BETTER DEFAULTS

src/dashboards/components/AutoRefresh.scss
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.refresh-form-container-child {
+  flex: 1;
+  &.active {
+    display: grid;
+    grid-template-columns: 1fr 1fr;
```
Asalem1, line 12: TIL

src/dashboards/components/AutoRefresh.scss
```diff
@@ -0,0 +1,46 @@
...
+
+.refresh-timeout-span {
+  flex: 1;
+}
+
+.refresh-form-buttons {
+  display: grid;
+  grid-template-columns: auto auto;
+  grid-gap: 10px;
+  padding: 30px 10px 15px 10px;
```
Asalem1, line 31: What Boatie said about variablized cf margins & padding

src/dashboards/components/AutoRefresh.scss
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.refresh-form-cancel-button {
+  width: 60%;
```
Asalem1, line 35: i'd shy away from relative button sizes in this respect and use `@media` to set specific widths based on the screen size. The main reason being that it can be pretty unpredictable. I don't currently have context on where this exists, but I don't imagine that the button would need to be more than 200px wide (i think that's the `submit` button size that we set)

src/dashboards/components/AutoRefresh.scss
```diff
@@ -0,0 +1,46 @@
...
+  justify-self: end;
+}
+
+.refresh-form-activate-button {
+  width: 60%;
```
Asalem1, line 40: same as above. Gonna finish this review in a bit

src/dashboards/components/AutoRefreshOverlay.tsx
```diff
@@ -0,0 +1,141 @@
...
+          </div>
+          <div className="refresh-form-container">
+            <span className="refresh-form-container-child">
+              Inactivity Timeout:{' '}
+            </span>
+            <div
+              className={
+                state.inactivityTimeout === 'None'
+                  ? 'refresh-form-container-child inactive'
+                  : 'refresh-form-container-child active'
```
Asalem1, line 79: Since these both rely on the same base classname we can probably rely upon a simplified ternary for the active and inactive class

src/dashboards/components/DashboardContainer.tsx
```diff
@@ -26,12 +26,71 @@ const {Active} = AutoRefreshStatus
...
+      registerStopListeners()
+    }
+    return () => {
+      registerStopListeners()
+      document.removeEventListener('visibilitychange', visChangeHandler)
+    }
+  }, [autoRefresh?.status, autoRefresh.inactivityTimeout])
+
+  const stopFunc = useCallback(() => {
+    if (autoRefresh?.duration?.upper <= new Date().toISOString()) {
```
Asalem1, line 85: I could be wrong but I think this will always be false since the comparison is between a string? Might need a ‘getTime’ tacked to it?

src/dashboards/components/DashboardContainer.tsx
```diff
@@ -63,7 +122,7 @@ const DashboardContainer: FC<Props> = ({autoRefresh, dashboard}) => {
 
 const mstp = (state: AppState) => {
   const dashboard = state.currentDashboard.id
-  const autoRefresh = state.autoRefresh[dashboard] || AUTOREFRESH_DEFAULT
+  const autoRefresh: any = state.autoRefresh[dashboard] || AUTOREFRESH_DEFAULT
```
Asalem1, line 125: Setting the type here to any isn’t all that beneficial. Why do it?

src/dashboards/components/DashboardHeader.tsx
```diff
@@ -180,15 +191,39 @@ const DashboardHeader: FC<Props> = ({
...
-            onSetTimeRange={handleChooseTimeRange}
-            timeRange={timeRange}
-          />
+          {isFlagEnabled('new-auto-refresh') && (
+            <Button
+              text={
+                autoRefresh?.status &&
+                autoRefresh.status === AutoRefreshStatus.Active
+                  ? 'Stop Auto Refresh'
+                  : 'Enable Auto Refresh'
```
Asalem1, line 210: We can probably just make the start and stop conditional here

src/dashboards/components/DashboardHeader.tsx
```diff
@@ -180,15 +191,39 @@ const DashboardHeader: FC<Props> = ({
...
+              }
+              color={
+                autoRefresh?.status &&
+                autoRefresh.status === AutoRefreshStatus.Active
+                  ? ComponentColor.Danger
+                  : ComponentColor.Primary
+              }
+              onClick={
+                autoRefresh?.status &&
+                autoRefresh.status === AutoRefreshStatus.Active
```
Asalem1, line 220: We should probably set an ‘isActive’ const somewhere instead  of doing this 3 times

src/dashboards/components/DashboardHeader.tsx
```diff
@@ -206,11 +241,14 @@ const mstp = (state: AppState) => {
   const timeRange = getTimeRange(state)
   const org = getOrg(state)
 
+  const autoRefresh = state.autoRefresh[state.currentDashboard.id]
```
Asalem1, line 244: Given the dependent logic of this statement, I’d love to see some optional chaining so the app doesn’t fall apart if one of those assertions is false

src/dashboards/components/DashboardPage.tsx
```diff
@@ -168,4 +163,5 @@ const mdtp = {
 
 const connector = connect(mstp, mdtp)
 
+interface OwnProps {}
```
Asalem1, line 166: Why define it here if it’s empty. Just remove it from the line below?

src/dashboards/components/RefreshContext.tsx
```diff
@@ -0,0 +1,137 @@
...
+  duration: CustomTimeRange
+  inactivityTimeout: string
+  inactivityTimeoutCategory: string
+  refreshMilliseconds: AutoRefresh
+}
+
+const calculateTimeout = (timeout, timeoutUnit) => {
+  const timeoutNumber = parseInt(timeout)
+  const startTime = moment(new Date())
+  const copyStart = startTime.unix()
```
Asalem1, line 33:
```suggestion
  const startTime = moment().unix()
  const copyStart = startTime
```

src/dashboards/components/DashboardContainer.tsx
Asalem1: I was wrong. I didn't realize that that `upper` was being set to ISOString as well. Having said that I think we should probably normalize these times to numbers for the sake of safety if the implementation is iterated upon at a later date and the same thoroughness isn't maintained so that we don't end up in a situation where someone uses a `Date.toString()` and messes everything up.

src/dashboards/components/DashboardContainer.tsx
Asalem1: An explicit type would help resolve some of the confusion I had around the `upper` time comparison in the comment above

src/dashboards/components/RefreshContext.tsx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+const AutoRefreshContextProvider: FC = ({children}) => {
+  const [state, dispatch] = useReducer(
+    autoRefreshReducer,
+    createAutoRefreshInitialState()
+  )
+
+  const {currentDashboardId} = useSelector((appState: AppState) => ({
+    currentDashboardId: appState.currentDashboard.id,
```
Asalem1, line 88: note to self: how is this not a selector yet?

src/dashboards/components/RefreshContext.tsx
Asalem1: https://github.com/influxdata/ui/pull/1226/files#diff-c97616c7c842d06e243a0e48f4b5eed1b608337fce52a4d9bc8bd84d068b3965R82

src/shared/apis/queryCache.ts
Asalem1: it's from one of the other query caches :(

src/shared/components/cells/CellContext.tsx
Asalem1: Ya, I think the query cache makes that optional somewhere in one of the caches. We can probably do the same thing here

Asalem1 (changes requested):
Good morning and happy monday! A couple of things from different glances this morning, but my big takeaway is:

- Great job on implementing a new feature
- Way to go on reintegrating an old feature with a fix for the underlying issues
- Man I wish the pause feature was in a separate PR to keep the PR more straightforward
- My mind hurts trying to understand why the global reducer & state reducer both need information regarding the dashboard's autorefresh status

src/dashboards/components/AutoRefreshOverlay.tsx
ShmuelLotman: Great point. This is one of those "can't believe i didnt think of that" moments, missed this bc i was thinking of the ends and not the means here. Thanks for smacking me over the head w this!

src/dashboards/components/RefreshContext.tsx
ShmuelLotman: No error here, just a piece of code that should have been removed once i made the interface

src/shared/apis/queryCache.ts
ShmuelLotman: something that should and will no longer exist, thanks for the catch

src/shared/apis/queryCache.ts
ShmuelLotman: Updated. I should have pulled this out into its own function. Did that in the last commit

src/shared/components/cells/CellContext.tsx
ShmuelLotman: this is view, just properly typed to avoid the markdown stuff, which has no queries property

src/views/actions/thunks.ts
ShmuelLotman: if i put it back, the query builder no longer caches!

src/dashboards/components/AutoRefresh.scss
ShmuelLotman: Freaking love CSS Grid and Flex... Gd bless whoever made them

src/dashboards/components/RefreshContext.tsx
ShmuelLotman: note to you: it is now :)

src/dashboards/components/DashboardContainer.tsx
ShmuelLotman: Yup! Addressed, we are going to use getTime() now, so its much more "accurate." This was a hacky solution, thank you for pointing that out!

src/dashboards/components/RefreshContext.tsx
ShmuelLotman: I need this to stay the way it is bc of the next line

src/views/actions/thunks.ts
ShmuelLotman: I think the real question is _why_ this is the fix tho. LMK if you can take a look w me at some point

src/shared/components/cells/CellContext.tsx
ShmuelLotman: Gonna keep this for now, since we need access to all the variables and @Asalem1 approves

src/dashboards/components/AutoRefreshOverlay.tsx
daviddkkim: Hey @drdelambre, thanks for flagging this! Appreciate you hitting the brakes here and design input.

 For this usecase, I think the use of the two drop down (one for number, and one for unit of time) makes sense. While it is a new paradigm to our application, this is a familiar paradigm in modern(and not modern) applications. 
 
 cc @mavarius

src/views/actions/thunks.ts
hoorayimhelping:
```js
!!' ' // true
!!' '.trim() // false
```

hoorayimhelping (changes requested):

src/dashboards/components/AutoRefreshOverlay.tsx
```diff
@@ -0,0 +1,178 @@
...
+import './AutoRefresh.scss'
+
+// This array creates an array of [0...24] for the hours selection
+const INACTIVITY_ARRAY = [...Array(25).keys()].map(num => num.toString())
+// This line replaces the 0 (the first value) with 'None' for the dropdown
+INACTIVITY_ARRAY[0] = 'None'
+
+export const AutoRefreshForm: FC = () => {
+  const {onClose} = useContext(OverlayContext)
+  const {state, dispatch: setRefreshContext, activateAutoRefresh} = useContext(
```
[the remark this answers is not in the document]
ShmuelLotman: Context here provides a convenient and consolidated abstraction for the state management operations within the overlay component, but the autoRefresh state we need to manage across the dashboards is not quite the same and so its unnecessary to clog up the redux state with these operations. This follows other patterns across the codebase.

src/dashboards/components/DashboardContainer.tsx
```diff
@@ -21,17 +21,92 @@ import {AUTOREFRESH_DEFAULT} from 'src/shared/constants'
...
+// Notifications
+import {dashboardAutoRefreshTimeoutSuccess} from 'src/shared/copy/notifications'
+import {notify} from 'src/shared/actions/notifications'
+
 const {Active} = AutoRefreshStatus
 
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = ReduxProps
 
+let timer
```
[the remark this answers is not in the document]
ShmuelLotman: Because using state introduces infinite rerenders. declaring the timer variable from outside the component avoids this entirely. cc @drdelambre

src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -58,20 +57,24 @@ class DateRangePicker extends PureComponent<Props, State> {
...
-              maxDate={upper}
-            />
-            <DatePicker
-              dateTime={upper}
-              onSelectDate={this.handleSelectUpper}
-              onInvalidInput={this.handleInvalidInput}
-              label="Stop"
-              minDate={lower}
-            />
+            {singleDirection !== 'upper' && (
```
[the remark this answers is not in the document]
ShmuelLotman: great point

src/dashboards/components/RefreshContext.tsx
```diff
@@ -0,0 +1,146 @@
...
+    interval: number
+  }
+  infiniteDuration: boolean
+}
+
+const jumpAheadTime = (moveAheadNum: number) => {
+  return (
+    new Date().toISOString().slice(0, 10) +
+    ' ' +
+    new Date(Date.now() + moveAheadNum).toTimeString().slice(0, 8)
```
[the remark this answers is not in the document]
ShmuelLotman: Thanks for pointing this out. I had had my difficulties bc of the limitations of our time dropdown... so this was a workaround that i was now able to resolve after messing with moment for more than a moment...

src/dashboards/components/AutoRefreshOverlay.tsx
```diff
@@ -0,0 +1,178 @@
...
+import {CustomTimeRange, AutoRefreshStatus} from 'src/types'
+
+// Context
+import AutoRefreshContextProvider, {
+  AutoRefreshContext,
+} from 'src/dashboards/components/RefreshContext'
+
+import './AutoRefresh.scss'
+
+// This array creates an array of [0...24] for the hours selection
```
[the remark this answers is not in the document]
daviddkkim: I think it will be okay if we give the dropdown a height max and hide the rest w/scroll bar. 

Another option is to provide options that are multiple of 2 or 3 or some number rather than all 1~24. There's probably diminishing returns on that granular of a control on inactivity timeout.  

I think I like the idea of giving the dropdown a height max and hiding the rest w/scrollbar. What do you guys think?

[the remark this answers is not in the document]
ShmuelLotman: @daviddkkim the current implementation has a height and a scroll bar to hide the rest, so ill be keeping this the way it is since it sounds like you are on board with that

src/shared/apis/queryCache.ts
```diff
@@ -124,19 +125,28 @@ class QueryCache {
...
+    if (!this.cache[id] || this.cache[id]?.paused) {
       return
     }
     delete this.cache[id]
   }
 
   resetCache = (): void => {
-    this.cache = {}
+    for (const key of Object.keys(this.cache)) {
+      if (!this.cache[key].paused) {
```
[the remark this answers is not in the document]
Asalem1: This is outdated. Paused should be removed from this PR

src/shared/components/cells/Cell.tsx
```diff
@@ -38,7 +39,18 @@ class CellComponent extends Component<Props, State> {
     submitToken: 0,
   }
 
+  private handleRefreshProcess = (): void => {
+    const {view} = this.props
+    if (view.properties.type === 'markdown') {
+      return
+    }
+    view.properties.queries.forEach(query => {
```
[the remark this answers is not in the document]
Asalem1: It should always exist if the view isn’t a markdown.

src/dashboards/components/DashboardContainer.tsx
```diff
@@ -21,17 +21,92 @@ import {AUTOREFRESH_DEFAULT} from 'src/shared/constants'
...
+    ) {
+      registerListeners()
+    } else {
+      registerStopListeners()
+    }
+    return () => {
+      registerStopListeners()
+      document.removeEventListener('visibilitychange', visChangeHandler)
+    }
+  }, [autoRefresh?.status, autoRefresh.inactivityTimeout])
```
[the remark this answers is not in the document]
drdelambre: i dont follow the mounting thing. given the "did change" parameters passed through the useEffect's second argument, i'm reading this as "everytime the autorefresh's status or timeout changes, add or remove event listeners based on the value"

src/dashboards/components/DashboardContainer.tsx
```diff
@@ -21,17 +21,98 @@ import {AUTOREFRESH_DEFAULT} from 'src/shared/constants'
...
+      autoRefresh.status === AutoRefreshStatus.Active &&
+      autoRefresh.inactivityTimeout > 0
+    ) {
+      registerListeners()
+    } else {
+      registerStopListeners()
+    }
+    return () => {
+      registerStopListeners()
+      document.removeEventListener('visibilitychange', visChangeHandler)
```
drdelambre, line 94: if you're going to be referencing the event listener like this, you should handle it's registration inside the useEffect as well. how it's set up now, it's going to register a new event listener on every render, then deregister it _sometimes_, which will lead to a slow memory leak at best, and calling a function a billion times at worst

src/dashboards/components/DashboardContainer.tsx
```diff
@@ -21,17 +21,98 @@ import {AUTOREFRESH_DEFAULT} from 'src/shared/constants'
...
+
+  useEffect(() => {
+    if (
+      autoRefresh?.status &&
+      autoRefresh.status === AutoRefreshStatus.Active &&
+      autoRefresh.inactivityTimeout > 0
+    ) {
+      registerListeners()
+    } else {
+      registerStopListeners()
```
drdelambre, line 90: idk if you need this as the first useEffect that registers the listeners is going to deregister the listeners as soon as it sees a change

src/dashboards/components/DashboardContainer.tsx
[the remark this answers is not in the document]
ShmuelLotman: This was addressed. Resolving.

daviddkkim (commented):
This is good with me on the design side of things. Great work Shmuel !

src/dashboards/components/DashboardContainer.tsx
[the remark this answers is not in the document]
hoorayimhelping: >Because using state introduces infinite rerenders.

this seems like a smell. what is the underlying issue that is causing re-renders?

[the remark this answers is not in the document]
ShmuelLotman: Hard to ascertain, discussed w/ @drdelambre and it seems like its because the state update loses reference to the right settimeout, something which does not happen if the variable for timer is either in the ref or outside of the component. Per @drdelambre 's suggestion, the ref solution was implemented

src/dashboards/components/DashboardContainer.tsx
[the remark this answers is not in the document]
hoorayimhelping: >i dont follow the mounting thing.
when it mounts, you attach event handlers. when it unmounts you detach them. it's a common allocation and deallcation pattern.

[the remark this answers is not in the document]
ShmuelLotman: Just introduced a simplification that limits this to one central useEffect. This should allow this component to remain a functional one while also simplifying the readability.

src/shared/apis/queryCache.ts
```diff
@@ -139,6 +139,10 @@ class QueryCache {
     this.cache = {}
   }
 
+  getById = (queryID: string) => {
```
Asalem1, line 142: We should remove this since it’s not valid and I think isn’t being used

src/dashboards/selectors/index.ts
```diff
@@ -89,3 +89,6 @@ export const getAllDashboards = (state: AppState): Dashboard[] =>
 
 export const hasNoDashboards = (state: AppState): boolean =>
   state.resources.dashboards.allIDs.length === 0
+
+export const getCurrentDashboardId = (state: AppState): string =>
```
Asalem1, line 93: Is this being used?

src/shared/components/cells/CellContext.tsx
```diff
@@ -80,21 +83,6 @@ const CellContext: FC<Props> = ({
     event('editCell button Click')
   }
 
-  const refreshCell = (): void => {
```
Asalem1, line 83: I think when I originally went through this I was confused by the pause functionality here. I think we still want this in, but we’d want to do a couple things:

1. clean this up by basically using the same function that’s in the Cell.tsx file
2. Make sure you’re passing in the variables as part of the process to reset the query cache

src/shared/components/cells/Cell.tsx
```diff
@@ -38,7 +39,18 @@ class CellComponent extends Component<Props, State> {
     submitToken: 0,
   }
 
+  private handleRefreshProcess = (): void => {
+    const {view} = this.props
+    if (view.properties?.type === 'markdown') {
+      return
+    }
+    view.properties?.queries?.forEach(query => {
+      resetQueryCacheByQuery(query.text)
```
Asalem1, line 48: We need to pass in the variables here.

src/shared/components/dropdown_auto_refresh/AutoRefreshDropdown.tsx
```diff
@@ -47,7 +50,7 @@ export default class AutoRefreshDropdown extends Component<Props> {
   }
 
   public render() {
-    if (CLOUD) {
+    if (!this.props.showAutoRefresh) {
```
Asalem1, line 53: Do we want to keep this exclusively for cloud users?

Asalem1 (approved):
A couple of things regarding unused functions. A bigger sticking point is not passing the variables in order to reset the queryCache by id. I’m happy to chat about that more if you’d like. Besides that it looks good to me

src/dashboards/selectors/index.ts
ShmuelLotman: yes it is, in RefreshContext

src/shared/apis/queryCache.ts
ShmuelLotman: done.

src/shared/components/cells/Cell.tsx
ShmuelLotman: done

hoorayimhelping (approved):
Good work

like I mentioned in a PM: I'm a little nervous about using a closure to work around an infinite re-render because that sounds like an indication of a problem elsewhere, but I don't have the context you guys do. So I'll just say good luck.

src/shared/components/dropdown_auto_refresh/AutoRefreshDropdown.tsx
ShmuelLotman: This feature is already on in OSS

PR #1238, opened 2021-04-19
chore: begin consuming swagger from central location

glinton:
The [openapi repo](https://github.com/influxdata/openapi) is now public and should be the source of truth when consuming API definitions.

This PR introduces new yarn tasks to utilize improved swagger definitions, while being updated to only point to new locations for current swaggers.


package.json
```diff
@@ -50,10 +50,17 @@
     "cy": "CYPRESS_baseUrl=http://localhost:9999 cypress open",
     "cy:dev": "source ../monitor-ci/.env && CYPRESS_baseUrl=https://$INGRESS_HOST:$PORT_HTTPS cypress open --config testFiles='{cloud,shared}/**/*.*'",
     "cy:dev-oss": "source ../monitor-ci/.env && CYPRESS_baseUrl=https://$INGRESS_HOST:$PORT_HTTPS cypress open --config testFiles='{oss,shared}/**/*.*'",
-    "generate": "oats https://raw.githubusercontent.com/influxdata/influxdb/master/http/swagger.yml > ./src/client/generatedRoutes.ts && yarn unity && yarn flows && yarn managedFunctions",
-    "unity": "oats ./src/client/swagger/unity.yml > ./src/client/unityRoutes.ts",
-    "flows": "oats ./src/client/swagger/flows.yml > ./src/client/flowsRoutes.ts",
-    "managedFunctions": "oats ./src/client/swagger/managedFunctions.yml > ./src/client/managedFunctionsRoutes.ts"
+    "generate": "yarn currentApi && yarn unity && yarn flowd && yarn managedFunctions",
```
Asalem1, line 53: will we want to update this to include running all the oats commands?

glinton: maybe; likely sometime.

One of the things doing so would allow is to address the drift that has occurred between the oss and cloud apis. there is a `common` definition that continues to define the bits of the api that is 100% equal between the two platforms (not sure how those differences are handled in the ui currently, or if those drifted routes are even an issue)

For now, i don't want to add any disruption and only am generating ts clients from what should be identical swagger definitions to what is currently being consumed

hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1238)

Asalem1 (commented):
It looks good to me. I missed the L&L but I assume that the entire process is moving `yml` definitions into OpenAPI, then resolving the duplication / inconsistencies there? I've roped in a couple of folks from the e-commerce team since this will affect some of the ongoing unification work that we're doing. As definers of the `unity` API, what work will / can we do to make changes to the `yml` file if need be? Really regretting not being able to make it to the L&L, how does this affect local development? i.e. if i make a change to the yml in OpenAPI, do those changes get trigger for all environments? Sorry, I'm unfamiliar with the platform so I'm not sure if it's partitioned into a `yml` file per environment like config cat

glinton:
> I missed the L&L

Bummer! Luckily, I was told it was recorded. Yes, the first pass is moving all swagger definitions into the openapi repo for a central point of consumption as well as a better platform for defining and maintaining our various apis. common bits are useful for api uniformity, but may not map 1:1 to new javascript services (as many of the shared components map to shared golang types).

>As definers of the unity API, what work will / can we do to make changes to the yml file if need be?

Open a pr against the openapi repo, since the unity definition is "private" (prefixed with `/api/v2private`) the definition isn't claimed to be stable, so you can either run the yarn tasks pointing to the pr/branch on that repo, or once the pr's been cleared by a stakeholder, merge it to master and run that way.

>If i make a change to the yml in OpenAPI, do those changes get trigger for all environments?

Currently there are no external triggers, but where these yarn tasks point to the master branch of that repo, ui will always be up-to-date (whether a good or bad thing, we'll find out). I did try to add an oats test for the oss swagger (since it is a consumer whose compatibility we don't want to break), but it is a simple test that just makes sure oats can process the swagger without error, nothing to do with the ui's types. I'd love better testing around that eventually (whether triggering the ui image build to test, or whatever).

>I'm not sure if it's partitioned into a yml file per environment like config cat

I'm not certain how config cat does it, but the contracts in the openapi repo are typically smaller pieces of the api in a single consumable file representative of either a shared service, possibly at a different route prefix, or a unique service (in the case of flowd, mapsd, or annotationd). A note about the different route prefix, we could have condensed the swagger definitions to have a server directive of `/api` and the paths would be at `/v2/<whatever>` or `/v2private/<whatever>`, but we felt it was more useful to keep, and embrace, the granularity; allowing consumers to avoid waste (generating clients with loads of unused logic/definitions).

Asalem1:
**update** offline with Greg about this one and it seems like the dependency on OpenAPI is one way from the UI. That is to say that any changes made to OpenAPI will automatically be propagated onto all environments and may cause breaking changes if certain contracts are broken or if the yml changes are malformed. A long term solution towards getting these changes in would be to have OpenAPI run PRs against the UI image to test against the existing e2e tests in order to determine whether a change to the yml definition might negatively impact the app

glinton:
By treating the swagger as a dependency, we can mitigate surprises while still being flexible to developers. A ui developer can modify the swagger and open a pr to openapi with their desired changes and update the sha in the 'generate' directive. This will ensure the version of the swagger they want will be used, both before and after their pr is merged.

Contributors to our openapi repo can (but likely will not be required to) open a pr here updating the sha to point to their changes (in a dependabot-like manner).

Asalem1:
> Thought about this a bit more. By treating the swagger as a dependency, we can mitigate surprises while still being flexible to developers. A ui developer can modify the swagger and open a pr to openapi with their desired changes and update the sha in the 'generate' directive. This will ensure the version of the swagger they want will be used, both before and after their pr is merged.
> 
> Contributors to our openapi repo can (but likely will not be required to) open a pr here updating the sha to point to their changes (in a dependabot-like manner).

Totally, but we're still left with the issue of a breaking change to the UI via the  OpenAPI yml can still be merged into the master OpenAPI branch without throwing any alarms prior to the merge

glinton:
> breaking change to the UI via the OpenAPI yml can still be merged into the master OpenAPI branch without throwing any alarms prior to the merge

Right, but that's the same risk of any dependency. Consumers should verify before updating to the latest version. As much as I'd love to run `webpack --config ./webpack.prod.ts` (within the ui repo, from openapi ci) prior to an openapi merge, the problems that introduce outweigh the benefits imo. What are your suggestions to prevent ui consuming bad swagger?

Asalem1:
> > breaking change to the UI via the OpenAPI yml can still be merged into the master OpenAPI branch without throwing any alarms prior to the merge
> 
> Right, but that's the same risk of any dependency. Consumers should verify before updating to the latest version. As much as I'd love to run `webpack --config ./webpack.prod.ts` (within the ui repo, from openapi ci) prior to an openapi merge, the problems that introduce outweigh the benefits imo. What are your suggestions to prevent ui consuming bad swagger?

Not necessarily. We have our UI tests running in IDPE to ensure that kind of stability, and we have k8s-IDPE tests running in the UI to ensure cross-platform stability. In theory, developers should be running against a specific branch to ensure that those processes are stable, but we've seen too many incidents arise from careless engineering. By adding in a check against the UI, we're able to ensure that the PRs made to the process are stable and won't shut down production

glinton:
Maybe we are talking about different parts. currently, in your `package.json`, you are generating a client directly from a swagger in influxdb master. If a bad swagger was to merge to master right now, ui would be hosed.

https://github.com/influxdata/ui/blob/088ba4a322253cfed76296696055c3d05181216a/package.json#L53

There is no testing around that right now. UI just blindly trusts that the swagger in influxdb will be good. See https://github.com/influxdata/influxdb/pull/21283 for proof there are no tests around it. If that got merged, ui would be up a creek. Hence the solution in this pr, to pull a known stable version of the swagger so the ui can be certain merging a bad definition will not screw it up.

https://github.com/influxdata/ui/blob/bc16679f2f1c6545024616ba217adcaec640a501/package.json#L53

As a ui developer, if I change a swagger and push it up to the openapi repo, i can then update the `package.json` to use the commit i want. On push to the ui, the ui tests run while consuming the commit of the swagger that I specified so that version of the swagger will be tested, blocking my ui pr from merging if it fails.

The current swagger definition in influxdb isn't radically changed (that I know of) without ya'll being notified, and likewise, nobody will go off the rails and redefine the swagger definitions in openapi that would break ui without working closely with the ui team. If it happens by accident, ui doesn't need to consume that bad swagger commit but work with the author of the bad swagger commit to get something that works.

Asalem1:
> Maybe we are talking about different parts. currently, in your `package.json`, you are generating a client directly from a swagger in influxdb master. If a bad swagger was to merge to master right now, ui would be hosed.
> 
> https://github.com/influxdata/ui/blob/088ba4a322253cfed76296696055c3d05181216a/package.json#L53
> 
> There is no testing around that right now. UI just blindly trusts that the swagger in influxdb will be good. See [influxdata/influxdb#21283](https://github.com/influxdata/influxdb/pull/21283) for proof there are no tests around it. If that got merged, ui would be up a creek. Hence the solution in this pr, to pull a known stable version of the swagger so the ui can be certain merging a bad definition will not screw it up.
> 
> https://github.com/influxdata/ui/blob/bc16679f2f1c6545024616ba217adcaec640a501/package.json#L53
> 
> As a ui developer, if I change a swagger and push it up to the openapi repo, i can then update the `package.json` to use the commit i want. On push to the ui, the ui tests run while consuming the commit of the swagger that I specified so that version of the swagger will be tested, blocking my ui pr from merging if it fails.
> 
> The current swagger definition in influxdb isn't radically changed (that I know of) without ya'll being notified, and likewise, nobody will go off the rails and redefine the swagger definitions in openapi that would break ui without working closely with the ui team. If it happens by accident, ui doesn't need to consume that bad swagger commit but work with the author of the bad swagger commit to get something that works.

Gotcha. So can we version lock the UI's openAPI yml file and have PRs made to openAPI update the locked version with a PR in the UI automagically?

glinton:
> have PRs made to openAPI update the locked version with a PR in the UI automagically?

Absolutely, we can look into doing this in future work to the openapi repo

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](true1238)

genehynson:
> Absolutely, we can look into doing this in future work to the openapi repo

Great discussion! I think the current solution of locking the SHA is enough to merge this PR in. 

I do think we should open an issue for adding automated testing of changes made to the OpenAPI repo. We actually can do this pretty easily with our new `monitor-ci` pipeline! The OpenAPI repo would just make an API request to Circle to start the `monitor-ci` pipeline and then poll its result. I have a script setup already to do this. A similar use case we're working on right now is adding UI testing to the influxdata/influxdb repo so that changes to OSS are tested against the UI.

genehynson (approved):
lgtm

Issue #1230, opened 2021-04-16
Annotations: Correctly handle API errors


hoorayimhelping:
### Background
We currently display a success notification even if a request fails

### Reproduction Steps
1. Create an annotation
1. Edit the annotation to have more than 255 characters
   1. open the network tab
   1. save the edit
   1. observe the success message
   1. observe 400 failure network message

### Success Criteria
- API requests that fail have descriptive error messages

PR #1242, opened 2021-04-19
fix: edit annotation wait for response from API

ChitlangeSahas:
Addresses https://github.com/influxdata/ui/issues/1230

On an invalid response from the API, we were not waiting for the API call to finish, this caused a premature success notification even when the request failed in the API. This PR adds an `await` so that the API call is waited for to complete.

On editing the annotation message to more than 255 characters and pressing submit, we now see an error message and the form is not submitted. 

_Perhaps, the new form is a better solution to address the API issue, i.e not allowing more than 255 characters in the first place._

Screen Shot 2021-04-19 at 2 29 59 PM

### DIFF
--- a/src/annotations/components/EditAnnotationOverlay.tsx
+++ b/src/annotations/components/EditAnnotationOverlay.tsx
@@ -33,7 +33,7 @@ export const EditAnnotationOverlay: FC = () => {
   const dispatch = useDispatch()
   const {clickedAnnotation} = useSelector(getOverlayParams)
 
-  const handleSubmit = (editedAnnotation: EditAnnotation): void => {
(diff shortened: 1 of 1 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1242)

genehynson:
This work was completed in: https://github.com/influxdata/ui/pull/1293

PR #1257, opened 2021-04-20
fix: validate variable names on creation properly

jrenee42:
for idpe#9771

this validates them properly on the front end.
adding unit tests too

### DIFF
--- a/src/variables/utils/validation.test.ts
+++ b/src/variables/utils/validation.test.ts
@@ -1,0 +1,71 @@
+import {validateVariableName} from 'src/variables/utils/validation'
+
+// reserved variable names
+import {
+  TIME_RANGE_START,
+  TIME_RANGE_STOP,
+  WINDOW_PERIOD,
+} from 'src/variables/constants'
+
+// Mocks
+import {createVariable} from 'src/variables/mocks'
+
+describe('Validation', () => {
+  const variable3 = createVariable('apple', 'f(x: v.b)')
+  const variableB = createVariable('banana', 'f(x: v.c)')
+  const variableF = createVariable('fig', 'f(x: v.g)')
+
+  const varList = [variable3, variableB, variableF]
+
+  const getError = name => {
+    return validateVariableName(varList, name).error
+  }
+
+  describe('variable name validation', () => {
+    const fullErrorText =
+      'Variable name must start with a letter or underscore, and ' +
+      'contain only numbers, letters, and underscores.'
+
+    it('rejects names with a special character in it', () => {
+      const special1 = 'foo&'
+      expect(getError(special1)).toEqual(fullErrorText)
+    })
+    it('rejects names with a space in it', () => {
+      const special1 = 'foo hello there'
+      expect(getError(special1)).toEqual(fullErrorText)
+    })
+    it('allows letters, then letters, underscore and numbers', () => {
+      expect(getError('foo_78')).toBe(null)
+      expect(getError('Foo_788')).toBe(null)
+      expect(getError('Foo')).toBe(null)
+      expect(getError('foo')).toBe(null)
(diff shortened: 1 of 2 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1257)

src/variables/utils/validation.ts
```diff
@@ -17,8 +17,24 @@ export const validateVariableName = (
...
-    return {error: 'Variable name cannot be empty'}
+  const spaceRegex = /^\s*$/
+
+  // must start with a letter or underscore, can only contain letters, numbers, and underscores only
+  const validCharacters = /^[A-Za-z_]+\w*$/
+
+  const emptyErrorText = 'Variable name cannot be empty'
+  const fullErrorText =
+    'Variable name must start with a letter or underscore, and ' +
+    'contain only numbers, letters, and underscores.'
```
TCL735, line 28: It looks like you are using these in the unit tests. Why not just export these strings, then? In case there was any need to tweak them, it would be easier just to change the exported value rather than hunt down and update all matching literals of these errors.

src/variables/utils/validation.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  const varList = [variable3, variableB, variableF]
+
+  const getError = name => {
+    return validateVariableName(varList, name).error
+  }
+
+  describe('variable name validation', () => {
+    const fullErrorText =
+      'Variable name must start with a letter or underscore, and ' +
+      'contain only numbers, letters, and underscores.'
```
TCL735, line 27: This should be an imported value. See other comment on exporting.

src/variables/utils/validation.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    it('disallows starting with a number', () => {
+      expect(getError('7bananas')).toEqual(fullErrorText)
+    })
+    it('disallows duplicate names', () => {
+      expect(getError('banana')).toEqual('Variable name must be unique')
+      expect(getError('apple')).toEqual('Variable name must be unique')
+      expect(getError('fig')).toEqual('Variable name must be unique')
+    })
+    it('rejects empty names', () => {
+      const emptyErrorText = 'Variable name cannot be empty'
```
TCL735, line 53: This should be an imported value. See other comment on exporting.

TCL735:
This is great. Combines separate errors and actually narrows the possible valid names for variables over what is currently allowed.

Unit tests were excellent too. Other than the specific comments, is it also possible to add some tests for when the variable already exists? In other words, unit tests for when the `validateVariableName` takes 3 arguments. No big deal if it's not included. That can be added in the future.

TCL735 (approved):

src/variables/utils/validation.ts
```diff
@@ -7,6 +7,16 @@ import {
 
 const reservedVarNames = [TIME_RANGE_START, TIME_RANGE_STOP, WINDOW_PERIOD]
 
+export const fullErrorText =
+  'Variable name must start with a letter or underscore, and ' +
+  'contain only numbers, letters, and underscores.'
+
+export const emptyErrorText = 'Variable name cannot be empty'
+export const uniqError = 'Variable name must be unique'
```
TCL735, line 15: Nitpick here. Constants should be in all caps. Something like:
`FULL_ERROR_TEXT`
`EMPTY_ERROR_TEXT`
`UNIQUE_ERROR_TEXT`

TCL735 (changes requested):

TCL735 (approved):

PR #1277, opened 2021-04-22
feat: integrated public flags without negating the getFlags condition

Asalem1:
This PR aims to integrate the pre-authenticated feature flagged routes for the UI. Yesterday, a similar PR caused our feature flags to fall apart. The reason for this was because the thunk I had developed was using the existing `getFlags` as a basis for comparison. This ended up causing an incident whereby the initial setting of the public flags was setting the status of the flags to loading, thereby bypassing the original authenticated `getFlags` request, which is only requested when the `status` of the `flags` are `RemoteDataState.NotStarted`. Since the `getFlags` were never called, the authenticated feature flags were never being set. I'd be happy to go over this in more detail and would be happy to pair with anyone who knows / can think of a way to test this out before merging it in

### DIFF
--- a/src/shared/actions/flags.ts
+++ b/src/shared/actions/flags.ts
@@ -4,12 +4,14 @@ import {FlagMap} from 'src/shared/reducers/flags'
 import {getAPIBasepath} from 'src/utils/basepath'
 import {RemoteDataState} from 'src/types'
 export const SET_FEATURE_FLAGS = 'SET_FEATURE_FLAGS'
+export const SET_PUBLIC_FEATURE_FLAGS = 'SET_PUBLIC_FEATURE_FLAGS'
 export const RESET_FEATURE_FLAGS = 'RESET_FEATURE_FLAGS'
 export const CLEAR_FEATURE_FLAG_OVERRIDES = 'CLEAR_FEATURE_FLAG_OVERRIDES'
 export const SET_FEATURE_FLAG_OVERRIDE = 'SET_FEATURE_FLAG_OVERRIDE'
 
 export type Actions =
   | ReturnType<typeof setFlags>
+  | ReturnType<typeof setPublicFlags>
   | ReturnType<typeof reset>
   | ReturnType<typeof clearOverrides>
   | ReturnType<typeof setOverride>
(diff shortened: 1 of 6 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1277)

src/shared/actions/flags.ts
```diff
@@ -1,14 +1,17 @@
 import {Dispatch} from 'redux'
 import {getFlags as getFlagsRequest} from 'src/client'
 import {FlagMap} from 'src/shared/reducers/flags'
```
hoorayimhelping, line 3: i think this should be moved to a common file and included from there. as it is now, this creates a circular dependency between actions/flags and reducers/flags since they both import from each other.

src/shared/actions/flags.ts
```diff
@@ -62,3 +73,19 @@ export const getFlags = () => async (
     dispatch(setFlags(RemoteDataState.Error, null))
   }
 }
+
+export const getPublicFlags = () => async (dispatch: Dispatch<Actions>) => {
```
hoorayimhelping, line 77: this isn't an action creator, since it has side effects. it (and `getFlags`) should be moved to a thunks file. we had problems in the pace with race conditions on load because of circular dependencies. what happens is the reducer includes the action creator. normally this isn't a problem as most action creators don't have a lot of dependencies on outside things. thunks do, however, and these dependencies getting loaded is what causes the problems, because the dependency graph kind of explodes out.

i think dealing with these circular dependencies will make race conditions less likely and will make the system behave more consistently locally and out on the web.

src/shared/actions/flags.ts
```diff
@@ -62,3 +73,19 @@ export const getFlags = () => async (
...
+    const flags = await response.json()
+
+    if (flags?.status && flags?.status !== 200) {
+      throw new Error(flags.message)
+    }
+
+    dispatch(setPublicFlags(flags))
+  } catch (error) {
+    console.error(error)
+  }
```
hoorayimhelping, line 90: if you're not going to do anything with the error, let it fail up to the error boundary. As it is now, this swallows errors and makes us unaware of them unless we happen to be the client that generates one and sees it in the console.

https://digitaldrummerj.me/dont-swallow-the-exceptions/

hoorayimhelping (commented):
I noticed some circular dependencies which might lead to confusing and hard to debug race conditions. Some time last summer, I went through and found a bunch of circular dependencies and found that they were creating conditions that sounded similar to what you saw with this in tools - weird states, hard to test code, etc.

drdelambre (approved):
:shrug:

Issue #1228, opened 2021-04-16
Maps: I would like to use Maps in my Notebooks


russorat:
I tried the same query in Data Explorer and it worked fine. when i tried it in notebooks, it looks like this:

![image](https://user-images.githubusercontent.com/4805997/115059624-a4fb8a80-9e9b-11eb-9174-3c85a79dee04.png)

PR #1279, opened 2021-04-22
fix: maps in notebooks

### LINKED ISSUE #866: Error States when Map is not rendered
Currently, we have two error states when map is not rendered:
1. When Map service is down or returns an error we display:
    > Map type is having issues connecting to the server. Please try again later.'
2. When data provided is not supported by Maps, for ex: not given lat or lon, we display:
      >  Map type is not supported with the data provided: Missing latitude/longitude values

@kristinarobinson Is this wording right to display in these cases? if not, who do we reach out to get proper decision.

Palakp41:
Closes #866 #1228

When viewing maps for visualization on Notebooks/Flows, the maps weren't rendering because the properties we were sending to giraffe are the defaults (with empty layers) and not the ones we use in Data explorer.

### DIFF
--- a/src/visualization/types/Map/view.tsx
+++ b/src/visualization/types/Map/view.tsx
@@ -1,7 +1,7 @@
 // Libraries
 import React, {FC, useEffect, useState} from 'react'
 import {Plot} from '@influxdata/giraffe'
-import {RemoteDataState} from '@influxdata/clockface'
+import {RemoteDataState, InfluxColors} from '@influxdata/clockface'
 
 // Types
 import {GeoViewProperties} from 'src/types'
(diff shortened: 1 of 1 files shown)


src/visualization/types/Map/view.tsx
```diff
@@ -109,6 +109,28 @@ const GeoPlot: FC<Props> = ({result, properties}) => {
     bingKey: '',
   }
 
+  let layersOpts = layers
```
hoorayimhelping, line 112:
```suggestion
  let layersOpts = [...layers]
```

this will create a copy for you. the way it is now will create a reference, and changing `layerOpts` will change `layers`.

not a big deal thought since `layers` doesn't get used here, but good practice.

hoorayimhelping (approved):
This looks good as a fix, but do you have any idea of how to fix this up longer term? I imagine this is a problem with other visualization types as well

Palakp41:
> This looks good as a fix, but do you have any idea of how to fix this up longer term? I imagine this is a problem with other visualization types as well

I don't see any other visualization types having issues. I think we will probably need to remove this in the next epic when we use customize options for layers and color codes.

src/visualization/types/Map/view.tsx
Palakp41: yeah makes sense. good idea.

Palakp41:
alchemy please merge

Palakp41:
alchemy please merge

PR #1280, opened 2021-04-22
feat: drop in a download button

### LINKED ISSUE #640: Data Table: I can download the raw query results as a csv file
As a InfluxDB Cloud user, I can download the raw flux results so that I can view the data in a local csv editor like Excel.

### Scenario: Success
Given: I have a raw data view open and results shown
When: I click the `download CSV` button
Then: I see a loading spinner indicating that the query is being executed instead of the download csv button

When: the query is executing for more than a few seconds
Then: a cancel button is displayed to the user

When: the cancel button is clicked
Then: the query is cancelled and the file stops downloading

When: Results start being returned from the query
Then: A file download from the browser is initiated with the query results
Then: The file name is `influxdb_results_<timestamp of when the file was created>.csv`

When: The query times out from the request
Then: A message is displayed to the user indicating that partial results were returned and the resulting csv file is not complete

### Technical Notes
  - This download CSV will download the full query results, not just the limited results we see in the browser
  - This should not be subject to the 27mb limit in cloud. It will still be subject to any api or queryd timeouts.

### Design Notes
- Mockup
  - https://www.figma.com/file/8S8IH8mimYIwanZ6keIQ6U/Notebooks-Prototype?node-id=1635%3A10213

### LINKED ISSUE #1118: notebooks: Don't show success message after cancel
I had a long-running query, so I hit 'cancel'. It cancelled successfully, but then I got the usual message "Notebook preview successful". This message shouldn't display if I have cancelled the query.

### LINKED ISSUE #1155: Notebooks: Navigating away from a notebook should cancel all running queries
If i am waiting for a query to return, and i navigate away from the notebook i am on, then the pending query should be cancelled.

### LINKED ISSUE #1156: Notebooks: Hitting cancel should cancel all running queries
When i open a notebook, it runs the query in the notebook. if i hit the preview button, it starts a second query. if i try to cancel, it doesn't do anything.  The expectation is clicking the cancel button should cancel all pending queries in the notebook.

Kapture 2021-04-09 at 09 08 59

### LINKED ISSUE #1157: Notebooks: The preview button should be updated when a query is running
When you open a notebook, it runs the query behind the scenes, but doesn't give you any indication that the query is running. We should update the status of the preview button and give the users the option to cancel that query if they like.

Kapture 2021-04-09 at 09 08 59

drdelambre:
I ADDED A BUTTON!

it bypasses the visualization limits and runs a long query, right into your download folder. cleaned up some api logic while i was in there as well

Screenshot from 2021-04-22 12-09-38

closes #1155
closes #1156
closes #1157
closes #1118  
closes #640


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1280)

src/flows/components/Flow.tsx
```diff
@@ -1,15 +0,0 @@
...
-import PipeList from 'src/flows/components/PipeList'
-import {PopupDrawer, PopupProvider} from 'src/flows/context/popup'
-
-// NOTE: requires a FlowProvider and ResultsProvider
-const Flow = () => (
-  <PopupProvider>
-    <PipeList />
-    <PopupDrawer />
-  </PopupProvider>
-)
```
Asalem1, line 13: any excuse to kill a feature am i right?

src/flows/components/header/Submit.tsx
```diff
@@ -30,6 +30,7 @@ const fakeNotify = notify
 
 export const Submit: FC = () => {
   const {flow} = useContext(FlowContext)
+  const {cancel} = useContext(QueryContext)
```
Asalem1, line 33: be still my beating heart

src/flows/components/panel/Results.tsx
```diff
@@ -31,6 +36,19 @@ const Results: FC = () => {
     emptyText = 'No Data Returned'
   }
 
+  const downloadTitle = queryText
+    ? 'Download results as an annotated CSV file'
+    : 'Build a query to download your results'
+  const download = () => {
+    basic(queryText).promise.then(response => {
```
Asalem1, line 43: i feel hoodwinked. This seems like the perfect opportunity to run the query using the `queryCache` and simply retrieve the query results from there?

Asalem1 (commented):
Just a though

src/flows/components/Flow.tsx
drdelambre: omg there are a million little files that just link two components

src/flows/components/panel/Results.tsx
drdelambre: i'm working on hooking up the query cache work over on the QueryProvider side, still figuring out where variables live because they dont currently exist in notebooks (spoiler: we shouldn't have to pass the whole variable graph just to run a query)

but more to the point here: the reason the query is rerun is because we dont run the full query when trying to grab the data that gets displayed in the simple table interface. Because users normally want a quick view of the shape of the data being returned by their query, that query always has a `|> limit(n: 100)` attached to it. If the user decides that they want to download the results of that query, the idea is that they are then communicating that they want ALL the things, not just the shape, so we want to run it without the limit. Adding the cache in here wouldn't help save any resources.

Asalem1: Aw, k. Didn't realize we were limiting the results that get displayed. Guess that's an entirely different issue to overcome but 🤷🏽

src/flows/components/header/Submit.tsx
drdelambre: it works too!

Asalem1 (approved):
Cheers to not breaking things AND adding a button

Issue #1266, opened 2021-04-21
Annotations: Attempting to edit the time results in invalid date


russorat:
I tried to edit the time in an annotation, and anytime i made a change, it moved my cursor to the end of the time. Then i tried to delete a space, and the entire date was replaced with the text "invalid date"

The expectation is that the data should be editable and the validation should happen when the user tries to save the annotation or when the focus moves to another input.

![Kapture 2021-04-20 at 20 56 49](https://user-images.githubusercontent.com/4805997/115494689-ff3d7800-a21a-11eb-8e3e-23e68259a02c.gif)

ChitlangeSahas:
This was closed in #1282

PR #1282, opened 2021-04-22
fix: improved annotation start time input behavior

ChitlangeSahas:
To address https://github.com/influxdata/ui/issues/1266, I propose that we make the input field behave similar to the one we have in the date picker. It adds consistency across the app. This PR introduces that.

The behavior is as follows:

The input field makes sure that the time entered is in the format: `YYYY-MM-DD HH:mm:ss.SSS`. If it isn't, it shows an error with a message that shows what the format needs to be. 

Notes for the reviewer: 

* I am using the `moment` library to achieve formatting, we're already using it for the date picker component. If that is something we want to move away from, I can explore ways to achieve it using the default `Date` library.

* I can also add the form validation behavior, i.e. the save button is only enabled when the format in the input field is valid, if this is a way we want to move forward. 

Video:

https://user-images.githubusercontent.com/18511823/115783064-42a9fa80-a371-11eb-8b72-d91d3a4c2b83.mov

### DIFF
--- a/src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
+++ b/src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
@@ -18,7 +18,6 @@ interface Props {
 }
 
 export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
-  // initial input value is the startTime passed
   const [startTimeValue, setStartTimeValue] = useState<string>(
     moment(props.startTime).format('YYYY-MM-DD HH:mm:ss.SSS')
   )
--- a/src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
+++ b/src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
@@ -18,7 +18,6 @@ interface Props {
 }
 
 export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
-
   // initial input value is the startTime passed
   const [inputValue, setInputValue] = useState<string>(
     moment(props.startTime).format('YYYY-MM-DD HH:mm:ss.SSS')
--- a/src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
+++ b/src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
@@ -45,7 +45,7 @@ export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
     }
   }
 
-  const isInputValueValid = (inputValue: string): boolean => {
+  const isValidInputValue = (inputValue: string): boolean => {
     if (inputValue === null) {
       return true
     }
@@ -54,7 +54,7 @@ export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
   }
 
   const getInputValidationMessage = (): string => {
-    if (!isInputValueValid(startTimeValue)) {
+    if (!isValidInputValue(startTimeValue)) {
       return 'Format must be YYYY-MM-DD [HH:mm:ss.SSS]'
     }
 
--- a/src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
+++ b/src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
@@ -23,8 +23,8 @@ export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
     moment(props.startTime).format('YYYY-MM-DD HH:mm:ss.SSS')
   )
 
-  const isValidTimeFormat = (d: string): boolean => {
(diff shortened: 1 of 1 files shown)


hercules-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1282)

src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
```diff
@@ -30,6 +46,24 @@ export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
     }
   }
 
+  const isInputValueInvalid = (): boolean => {
```
hoorayimhelping, line 49: when designing methods, it's typically better name them and design them in the affirmative, not the negative. that way, if you need to negate the value, you don't have a double negative in the middle of your code.

so instead of

```ts
isInputValueInvalid()
!isInputValueInvalid()
```
you'd have
```ts
isInputValueValid()
!isInputValueValid()
```

secondly: this is relying on `inputValue` existing and being in scope. it's generally better method design to pass in dependencies and return some computation. if we ever need to move this method or reuse it, it becomes much more portable:

```ts
isInputValueValid(inputValue: string): boolean => {

// then later when it's called

if (!isInputValueValid(inputValue)) {
  return ...
}

File: src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx [hunk 2]
```diff
@@ -30,6 +46,24 @@ export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
...
+    if (inputValue === null) {
+      return false
+    }
+
+    return !isValidTimeFormat(inputValue)
+  }
+
+  const inputErrorMessage = (): string | undefined => {
```
hoorayimhelping, line 57: let's rename this to `getInputValidationMessage` or something like that. Error message implies there _is_ an error when there may not be. It also shouldn't return undefined, it'll only return a string according to this method implementation.

the same comment about passing in an argument applies here:

```ts
const getInputValidationMessage = (inputValue: string): string => {
  if (!isInputValueValid(inputValue)) {
    return ...
  }
  return ''
}

```ts

File: src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx [hunk 3]
```diff
@@ -17,10 +18,25 @@ interface Props {
 }
 
 export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
-  const validationMessage = props.startTime ? '' : 'This field is required'
+  // initial input value is the startTime passed
+  const [inputValue, setInputValue] = useState<string>(
```
hoorayimhelping, line 22: `inputValue` is a decent name, but not superb. It describe how it's used fine, which makes the code easier to follow, but if this was named more after what it is, it'll be even easier to follow along and remember the purpose of this code. What I mean is, this is the start time input for annotations, is there a name you can think of more descriptive than `inputValue` that might represent what the `inputValue` is here?

hoorayimhelping (changes requested):
Behavior looks solid, have some comments on code design / method naming and design.

src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
ChitlangeSahas: Yes absolutely! I see your point and this makes sense. I'm thinking `startTimeValue`

src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
ChitlangeSahas: Done ✅

src/annotations/components/annotationForm/AnnotationStartTimeInput.tsx
```diff
@@ -30,6 +45,24 @@ export const AnnotationStartTimeInput: FC<Props> = (props: Props) => {
     }
   }
 
+  const isInputValueValid = (inputValue: string): boolean => {
```
hoorayimhelping, line 48: small nit, `isValidTimeFormat` vs `isInputValueValid` are kind of in different voices - consistency makes it easier to follow along and ultimately understand. Small thing, but I think these tend to add up.

hoorayimhelping (approved):
good work! small nit, but probably don't need to hold it up. Run it by David if you haven't already just to give him a heads up

daviddkkim:
This looks good! Good work @ChitlangeSahas

PR #1285, opened 2021-04-23
fix(unity) Define errors for checkout and payment form

gspletzer:
- Updated happy path response for `POST /checkout` to `201` 
- Defined errors for `POST /checkout`
- Updated path for `GET /payment_form/{form}` to include `form` parameter
- Defined errors for `GET /payment_form/{form}`

### DIFF
--- a/src/client/swagger/unity.yml
+++ b/src/client/swagger/unity.yml
@@ -730,14 +730,8 @@ paths:
             schema:
               $ref: '#/components/schemas/CheckoutRequest'
       responses:
-        '204':
-          description: Account upgraded
-        default:
-          description: Unexpected error
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/Error'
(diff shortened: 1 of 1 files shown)


Asalem1 (approved):
It looks good. I’d love to just get a verification that running ‘yarn unity’ in the root directory generates the ‘unityRoutes’ file (meaning it should have function definitions and not be a blank file) before merging this but it looks great. Thanks!

gspletzer:
> It looks good. I’d love to just get a verification that running ‘yarn unity’ in the root directory generates the ‘unityRoutes’ file (meaning it should have function definitions and not be a blank file) before merging this but it looks great. Thanks!

Yarn unity has been run on my end. Is there something else I need to do so you can see it? I'm also including a screenshot of part of the file.

Screen Shot 2021-04-26 at 10 36 37 AM

PR #1302, opened 2021-04-27
chore: integrate logout functionality for logout in dropdown

Asalem1:
This PR updates the logout path for the logout button in the avatar dropdown menu to route the user to the signout path in IDPE, rather than hit the `/logout` endpoint. I think this should resolve the issue that Randy was having with logging out

### DIFF
--- a/src/pageLayout/components/UserWidget.tsx
+++ b/src/pageLayout/components/UserWidget.tsx
@@ -14,10 +14,12 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Constants
 import {
+  CLOUD,
   CLOUD_URL,
   CLOUD_USAGE_PATH,
   CLOUD_BILLING_PATH,
   CLOUD_USERS_PATH,
+  CLOUD_SIGNOUT_PATHNAME,
 } from 'src/shared/constants'
 
 // Types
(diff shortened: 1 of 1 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1302)

randycoulman (commented):
I looked at the code last night, and it looked like every logout instance in the app was routing to `/logout`, which would then render the `Logout` component which should have been logging out the correct way based on the feature flag.

Given that, I'm not sure this one change is enough, because there are other places in the app routing to `/logout` as well.  I'm almost certainly missing something here, though.

Asalem1:
> I looked at the code last night, and it looked like every logout instance in the app was routing to `/logout`, which would then render the `Logout` component which should have been logging out the correct way based on the feature flag.
> 
> Given that, I'm not sure this one change is enough, because there are other places in the app routing to `/logout` as well. I'm almost certainly missing something here, though.

Right? That's what I was thinking when I had originally implemented this functionality, which is why I thought that the API was doing the redirecting to Quartz. It is a tricky one to try and figure out for sure, since the `/logout` path should be making a request to the `signout` with the feature flag. TBH I think this that the UI's Logout is a little too messy, so this was a shot in the dark. Let's sync today about trying out the shared session and logging out from the home page to see if that functions differently

Issue #1259, opened 2021-04-20
Flaky Test: can create an annotation that is scoped to a dashboard cell


genehynson:
Test `can create an annotation that is scoped to a dashboard cell` is flaky.

Source: https://github.com/influxdata/ui/blob/master/cypress/e2e/cloud/annotations.test.ts#L243

Examples:
- I'm too lazy to find the link

I believe this is happening because of the long chaining that's happening here: https://github.com/influxdata/ui/blob/master/cypress/e2e/cloud/annotations.test.ts#L248

PR #1312, opened 2021-04-28
test: fix flaky annotations tests

### LINKED ISSUE #1258: Flaky Test: can edit an annotation by clicking on the annotation line
The test `can edit an annotation by clicking on the annotation line` is flaky.

Source: https://github.com/influxdata/ui/blob/master/cypress/e2e/cloud/annotations.test.ts#L171

Examples:
- https://app.circleci.com/pipelines/github/influxdata/monitor-ci/268/workflows/f303d31b-7478-4648-8d2d-d658027edaa4/jobs/1997/artifacts

It looks like the "annotation control bar" is already open and the test closes it but expects to be opening it.

Talked to Sahas and this is because the visibility of the annotation control bar is stored in local storage and cypress isn't clearing local storage correctly.

genehynson:
Closes #1258 , #1259

Fixes two flaky annotations tests

### DIFF
--- a/cypress/e2e/cloud/annotations.test.ts
+++ b/cypress/e2e/cloud/annotations.test.ts
@@ -257,13 +257,10 @@ describe('The Annotations UI functionality', () => {
     cy.getByTestID('button')
       .click()
       .then(() => {
-        cy.getByTestID('selector-list schmucket')
-          .click()
-          .getByTestID(`selector-list m`)
-          .click()
-          .getByTestID('selector-list v')
-          .click()
-          .getByTestID(`selector-list tv1`)
+        cy.getByTestID('selector-list schmucket').click()
+        cy.getByTestID(`selector-list m`).click()
+        cy.getByTestID('selector-list v').click()
+        cy.getByTestID(`selector-list tv1`)
           .click()
           .then(() => {
             cy.getByTestID('time-machine-submit-button').click()
--- a/cypress/e2e/cloud/annotations.test.ts
+++ b/cypress/e2e/cloud/annotations.test.ts
@@ -3,7 +3,6 @@ import {lines} from '../../support/commands'
 describe('The Annotations UI functionality', () => {
   beforeEach(() => {
     cy.flush()
-    cy.clearLocalStorage()
     cy.signin().then(() =>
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):
thanks for tracking this down Gene!

cypress/e2e/cloud/annotations.test.ts
```diff
@@ -257,13 +263,10 @@ describe('The Annotations UI functionality', () => {
...
-          .click()
-          .getByTestID(`selector-list m`)
-          .click()
-          .getByTestID('selector-list v')
-          .click()
-          .getByTestID(`selector-list tv1`)
+        cy.getByTestID('selector-list schmucket').click()
+        cy.getByTestID(`selector-list m`).click()
+        cy.getByTestID('selector-list v').click()
+        cy.getByTestID(`selector-list tv1`)
```
hoorayimhelping, line 269: @genehynson can you explain to @ChitlangeSahas and me why this fixes the flakiness?

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1312)

cypress/e2e/cloud/annotations.test.ts
genehynson: Yes! So cypress generally isn't a fan of lost of chaining. The issue is that React updates the dom a lot and if the dom is updated in the middle of one of these long chains, cypress will lose its grasp on the element and it will error out saying "element was detached from the dom". My understanding is that basically whenever you chain commands, cypress interprets that as "use your reference to the dom from earlier _and_ do xyz". When you split them apart cypress always grabs the elements from the latest dom and things are less likely to be detached.

genehynson: btw to test these changes I ran each of these tests 20 times in a row. Before the change they were flaking out every 1 in 5 runs or so. With these changes all 20 runs passed

PR #1313, opened 2021-04-29
feat(functions): Route requests to api/v2/poc-functions until the api/v2/functions EP is available

ebb-tide:
The `/functions/...` endpoints will be used for named functions(flux) only, until the managed functions (python) functionality is in place. This routes the UI endpoints to `api/v2/poc-functions` instead for now so that we can continue demoing managed functions until we decide to devote resources to building out managed functions

### DIFF
--- a/src/client/swagger/managedFunctions.yml
+++ b/src/client/swagger/managedFunctions.yml
@@ -5,7 +5,7 @@ info:
 servers:
   - url: ''
 paths:
-  '/api/v2/functions':
+  '/api/v2/poc-functions':
     get:
       operationId: GetFunctions
       tags:
@@ -72,7 +72,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/api/v2/functions/trigger':
+  '/api/v2/poc-functions/trigger':
     post:
       operationId: PostFunctionsTrigger
       tags:
@@ -98,7 +98,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/api/v2/functions/{functionID}':
+  '/api/v2/poc-functions/{functionID}':
     get:
       operationId: GetFunctionsID
       tags:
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1313)

wojciechka (approved):

genehynson (approved):

glinton:
@ebb-tide, can you reflect these changes in the openapi repo [in this file](https://github.com/influxdata/openapi/blob/master/src/svc-managed-functions.yml). Also curious why the `/api/v2` prefix was chosen instead of the `/api/v2private` (which also holds apis that are evolving).

ebb-tide:
@glinton I think I was creating this swagger at a time with a vacuum in API leadership. Please look it over critically and we can change whatever you think should change. Do we have a habit of then removing an endpoint from `api/v2private` to `api/v2` ? - I wouldn't want it to remain there forever. 
But- I think we should not add these changes to the swagger at https://github.com/influxdata/openapi/blob/master/src/svc-managed-functions.yml - this is just to keep the Proof Of Concept demo working for one org in one staging cluster, so that we can demo to customers. And actually we're changing it here so that it doesn't interfere with work being done in `/functions` now or in future.

glinton:
> this is just to keep the Proof Of Concept demo working for one org in one staging cluster, so that we can demo to customers

Interesting. Is it required to support both paths at the moment then? that `managed-functions.yml` in the openapi repo is meant to replace the definition here (centralizing all the swaggers), so it seems like it fits to just put it there. Where there's still iterations going on for that api, `/api/v2private` seems most fitting for it until it's decided it's "mostly stable", at which point it could be migrated to `/api/v2`. 

> Do we have a habit of then removing an endpoint from api/v2private to api/v2

I wouldn't say we have a habit of moving things out of `v2private`, but that's been the plan, once somthing is out of development and in a stable relationship between ui (the main consumer) and the service.

PR #1321, opened 2021-04-29
feat: resets flags state after the URL is set

Asalem1:
This PR closes the issue where users logging out using the shared session cookie were undergoing the original logout loop. This issue presented itself because the feature flag and local storage were being reset prior to hitting the condition of checking the feature flag. As such, we are now checking to see if the flag is enabled prior to resetting it

### DIFF
--- a/src/Logout.tsx
+++ b/src/Logout.tsx
@@ -24,19 +24,27 @@ const Logout: FC<Props> = ({history}) => {
   const dispatch = useDispatch()
 
   useEffect(() => {
-    const handleSignOut = async () => {
-      if (CLOUD && isFlagEnabled('authSessionCookieOn')) {
-        const url = new URL(
-          `${window.location.origin}${CLOUD_SIGNOUT_PATHNAME}`
-        )
-        window.location.href = url.href
-        return
-      }
+    const handleReset = () => {
+      dispatch(reset())
+      dispatch({type: 'USER_LOGGED_OUT'})
+    }
 
+    const handleSignOut = async () => {
       if (CLOUD) {
-        window.location.href = `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
+        const url = isFlagEnabled('authSessionCookieOn')
+          ? new URL(`${window.location.origin}${CLOUD_SIGNOUT_PATHNAME}`).href
+          : `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
+
+        if (isFlagEnabled('authSessionCookieOn')) {
+          handleReset()
+        }
+
+        window.location.href = url
         return
       } else {
+        if (isFlagEnabled('authSessionCookieOn')) {
+          handleReset()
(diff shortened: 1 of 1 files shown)


randycoulman (approved):
Seems OK to me, I think.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1321)

src/Logout.tsx
```diff
@@ -24,19 +24,27 @@ const Logout: FC<Props> = ({history}) => {
...
-      if (CLOUD && isFlagEnabled('authSessionCookieOn')) {
-        const url = new URL(
-          `${window.location.origin}${CLOUD_SIGNOUT_PATHNAME}`
-        )
-        window.location.href = url.href
-        return
-      }
+    const handleReset = () => {
+      dispatch(reset())
+      dispatch({type: 'USER_LOGGED_OUT'})
```
genehynson, line 29: do we know that this dispatch event will complete before we redirect the user to the new page?

Asalem1: it should* I don't know if there's a way to guarantee that, but I also don't know if we can say the same for how it was structured beforehand. This is only a temporary workaround for a niche problem that will be reverted once the feature flag is lifted

src/Logout.tsx
```diff
@@ -46,8 +54,9 @@ const Logout: FC<Props> = ({history}) => {
         history.push(`/signin`)
       }
     }
-    dispatch(reset())
-    dispatch({type: 'USER_LOGGED_OUT'})
+    if (!isFlagEnabled('authSessionCookieOn')) {
+      handleReset()
```
joelfried, line 58: Why are we checking before running `handleReset()`?  It looks like it runs here if the `authSessionCookieOn` is false but runs if that is true on either 39 (if cloud is true) or 46 (if cloud is false).  So doesn't `handleReset()` just always get called?

Asalem1: it does always get called. 

> This PR closes the issue where users logging out using the shared session cookie were undergoing the original logout loop. This issue presented itself because the feature flag and local storage were being reset prior to hitting the condition of checking the feature flag. As such, we are now checking to see if the flag is enabled prior to resetting it

joelfried: So is it important what other things run first?  I'm mostly just confused why we aren't just calling `handleReset()` at this point.

Asalem1: Because calling reset before calling the signout was resetting our feature flags and we weren't hitting the feature flagged condition inside the logout function

joelfried (approved):

genehynson (approved):

PR #1334, opened 2021-05-03
feat: add the ability to retry a task

### LINKED ISSUE #685: Tasks: Add a button to retry a failed task run
It would be really convenient in the UI to have the ability to click a button and retry a failed task run. I'm thinking right next to the "View Logs" button. The link to retry a task is already in the `links` section of the response.

image

zoesteinkamp:
Closes #685 

This will add the feature of retrying a task

### DIFF
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -558,6 +558,26 @@ export const createTaskFromTemplate = (template: TaskTemplate) => async (
   }
 }
 
+export const retryTask = (
+  taskID: string,
+  runID: string
+) => async (dispatch: Dispatch<Action | NotificationAction>) => {
+  try {
+    const resp = await api.postTasksRunsRetry({
+      taskID,
+      runID,
+    })
+
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
+    }
+
+    dispatch(notify(copy.taskRetrySuccess()))
+  } catch (error) {
+    console.error(error)
+  }
+}
+
 export const runDuration = (finishedAt: Date, startedAt: Date): string => {
   let timeTag = 'seconds'
 
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -571,9 +571,9 @@ export const retryTask = (taskID: string, runID: string) => async (
       throw new Error(resp.data.message)
     }
 
-    dispatch(notify(copy.taskRetrySuccess()))
+    dispatch(notify(copy.taskRetrySuccess(runID)))
   } catch (error) {
-    console.error(error)
+    dispatch(notify(copy.taskRetryFailed(error)))
   }
 }
 
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -558,10 +558,9 @@ export const createTaskFromTemplate = (template: TaskTemplate) => async (
   }
 }
 
-export const retryTask = (
-  taskID: string,
-  runID: string
-) => async (dispatch: Dispatch<Action | NotificationAction>) => {
+export const retryTask = (taskID: string, runID: string) => async (
+  dispatch: Dispatch<Action | NotificationAction>
+) => {
   try {
     const resp = await api.postTasksRunsRetry({
       taskID,
--- a/src/tasks/components/TaskRunsRow.tsx
+++ b/src/tasks/components/TaskRunsRow.tsx
@@ -1,7 +1,7 @@
 // Libraries
 import React, {PureComponent} from 'react'
 import {connect, ConnectedProps} from 'react-redux'
-import {RouteComponentProps} from 'react-router-dom'
+import {withRouter, RouteComponentProps} from 'react-router-dom'
 import moment from 'moment'
 
 // Components
@@ -131,4 +131,4 @@ const mstp = (state: AppState) => {
 const mdtp = {getLogs: getLogs, retryTask: retryTask}
 
 const connector = connect(mstp, mdtp)
-export default connector(TaskRunsRow)
+export default connector(withRouter(TaskRunsRow))
--- a/src/tasks/components/TaskRunsRow.tsx
+++ b/src/tasks/components/TaskRunsRow.tsx
@@ -39,32 +39,31 @@ class TaskRunsRow extends PureComponent<Props, State> {
   public render() {
     const {run} = this.props
     return (
-        <IndexList.Row>
-          <IndexList.Cell>{run.status}</IndexList.Cell>
-          <IndexList.Cell>
-            {this.dateTimeString(run.scheduledFor)}
-          </IndexList.Cell>
-          <IndexList.Cell>{this.dateTimeString(run.startedAt)}</IndexList.Cell>
-          <IndexList.Cell>{run.duration}</IndexList.Cell>
-          <IndexList.Cell>
-            <Button
-              key={`logs-${run.id}`}
-              size={ComponentSize.ExtraSmall}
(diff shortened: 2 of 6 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1334)

src/tasks/components/TaskRunsRow.tsx
```diff
@@ -85,6 +90,16 @@ class TaskRunsRow extends PureComponent<Props, State> {
     this.setState({isImportOverlayVisible: !this.state.isImportOverlayVisible})
   }
 
+  private handleRetry = () => {
+    const {retryTask, taskID, run} = this.props
+    try {
+      retryTask(taskID, run.id)
```
hoorayimhelping, line 96: this is an async function. If it isn't `await`ed, the same issue you and Sahas ran into with annotation errors will happen here.

you'll need to make `handleRetry` an async function

src/tasks/actions/thunks.ts
```diff
@@ -558,6 +558,25 @@ export const createTaskFromTemplate = (template: TaskTemplate) => async (
...
+      runID,
+    })
+
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
+    }
+
+    dispatch(notify(copy.taskRetrySuccess()))
+  } catch (error) {
+    console.error(error)
```
hoorayimhelping, line 576: since it seems like notifications are sent from this thunk, let's add an error notification here, rather than doing a `console.error` if there's a helpful error message returned, we'll want to show the user it.

src/tasks/actions/thunks.ts
```diff
@@ -558,6 +558,25 @@ export const createTaskFromTemplate = (template: TaskTemplate) => async (
...
+    const resp = await api.postTasksRunsRetry({
+      taskID,
+      runID,
+    })
+
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
+    }
+
+    dispatch(notify(copy.taskRetrySuccess()))
```
hoorayimhelping, line 574: let's add the task run id to the message. Something like `Task run ${runID} was successful`

src/tasks/components/TaskRunsRow.tsx
```diff
@@ -85,6 +90,16 @@ class TaskRunsRow extends PureComponent<Props, State> {
...
+      window.location.reload()
+    } catch (error) {
+      console.error(error)
+    }
```
hoorayimhelping, line 100: if we handle the error with an error notification in the thunk, this `catch` is no longer necessary

```suggestion
    }
```

hoorayimhelping (changes requested):
Nice work! A couple of small tweaks, and reminder to be careful about making `async` functions asynchronous

src/tasks/components/TaskRunsRow.tsx
```diff
@@ -85,6 +90,16 @@ class TaskRunsRow extends PureComponent<Props, State> {
     this.setState({isImportOverlayVisible: !this.state.isImportOverlayVisible})
   }
 
+  private handleRetry = () => {
+    const {retryTask, taskID, run} = this.props
+    try {
+      retryTask(taskID, run.id)
+      window.location.reload()
```
Asalem1, line 97: Re rendering these page seems a bit drastic. Is there a way to simply get the runs afresh by dispatching a thunk rather than using the reload here? I ask because a page reload will make for a clunkier UX

src/tasks/components/TaskRunsRow.tsx
```diff
@@ -85,6 +90,12 @@ class TaskRunsRow extends PureComponent<Props, State> {
     this.setState({isImportOverlayVisible: !this.state.isImportOverlayVisible})
   }
 
+  private handleRetry = async () => {
+    const {retryTask, taskID, run} = this.props
+    await retryTask(taskID, run.id)
+    window.location.reload()
```
Asalem1, line 96: we should just dispatch to get all the runs rather than reloading the page

Asalem1 (dismissed):

src/tasks/components/TaskRunsRow.tsx
zoesteinkamp: yeah i saw the other comment, i just need to add that functionality

hoorayimhelping: rather than creating a new way of doing things for only this method (and thus making the user experience slightly different), let's just follow convention with tasks and push to history. If we need to change that convention (and I suspect we will) we can change all of those at the same time as a logical change, rather than having yet another disjointed experience.

Asalem1: I might be wrong (not at a computer), but isn’t there a thunk to get the runs for a task that we can use to make for a smoother UX?

hoorayimhelping: there might be. The point I'm trying to make is, making that change is separate from making this change, which is about retries. There's already enough being changed in this PR in a part of the app we haven't spent a ton of time in without introducing a further fundamental change.

I think it'll be a smoother experience to refresh the data rather than reloading the whole page(I don't think it will be earth-shatteringly better, it'll be a marginal improvement at best), but I don't think it's worth the benefit to confuse the purpose of this PR. We can follow up with a UX change if we need to

Asalem1: This sounds like the definition of tech debt to me

hoorayimhelping: >My question is, why not just dispatch an event?

Like I said in the comments above, coming up with a different UX flow for this one part of tasks is beyond the scope of this. And because keeping things consistent, even though it **might be** less than ideal, (don't forget that this is just your opinion on what is ideal, this hasn't been discussed with our product manager or designer) is preferable to having code inconsistencies in a part of the app we don't spend a lot of time maintaining and that will most likely be changing in the future.

All the pages under tasks currently do this history-pushing behavior. Why aren't you worried about them? This feature was (supposed to be) a small, quick hit. If we need to iterate on tasks, we should plan to do that separately. If you think this is important, then feel free to prioritize working on [this ticket](https://github.com/influxdata/ui/issues/1388) rather than requesting scope creep on a PR that has been approved.

Asalem1: I think I might be misunderstanding you. I don't believe there's another case of the UI using `history.push` to effectively **reload a page**, which is what the original implementation was for this feature. It looks like it has since been updated to use `history.push` to navigate to the task edit page. If that's in fact what the product and design goal of this is (rerun the task and then be redirected to the edit page), then that's not a problem. My problem is that we were previously trying to reload a page in order to refresh the runs for a task. If we are talking about opinions, I'm of the opinion that we shouldn't have our users bearing the cost of engineering by sticking to the conventions of our app when adding in new functionality

src/tasks/components/TaskRunsRow.tsx
```diff
@@ -85,6 +91,21 @@ class TaskRunsRow extends PureComponent<Props, State> {
...
+      taskID,
+      run,
+      history,
+      match: {
+        params: {orgID},
+      },
+    } = this.props
+
+    await retryTask(taskID, run.id)
+    history.push(`/orgs/${orgID}/tasks/${taskID}/edit`)
```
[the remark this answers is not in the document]
russorat: @zoesteinkamp @hoorayimhelping can we redirect them to the /runs page?

PR #1358, opened 2021-05-05
fix: wait for query loaded to reload view

ShmuelLotman:
Closes #

This PR addresses the issue of queries that take a long time to load making an awkward UX. It does so by waiting to reload the view until after the query is loaded, so that the queries that take longer wont refresh again as long as the query is loading and thus keep the proper view until we have something new to show.


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1358)

Asalem1:
Is there a gif we can reference to see this in action? Does this interfere with the current functionality in any way? When is the `submitToken` set?

drdelambre (approved):
sorry i forgot about this one. might wanna switch the loading state to !== RemoteDataState.Loading so it doesn't get stuck in NotStarted land, but it's prolly fine

PR #1360, opened 2021-05-05
feat: cloning a notebook

### LINKED ISSUE #1236: Notebooks: Ability to clone a notebook from the list page
We would like the ability to clone a notebook so that we can start a new notebook from the state of an existing notebook on the list page

Acceptance Criteria

* As a notebook user, I can clone an exiting notebook into a new, editable notebook from the list page

### Design Notes

Design: https://www.figma.com/proto/vDXu8PqumHB2JhiXDLIhOn/Duplicate-Notebook?node-id=1%3A2&scaling=min-zoom&page-id=0%3A1

Asalem1:
Closes #1236

This PR allows users to clone a notebook.

clone-notebook


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1360)

src/flows/components/FlowContextMenu.tsx
```diff
@@ -16,27 +16,47 @@ interface Props {
...
-    </Context.Menu>
+    <Context>
+      <Context.Menu
+        icon={IconFont.Duplicate}
+        color={ComponentColor.Secondary}
+        shape={ButtonShape.Default}
+        text="Clone"
+        testID={`context-clone-menu ${name}`}
+      >
+        <Context.Item
```
drdelambre, line 40: does this create a dropdown menu with a single item in it that says "clone" after clicking the "clone" button?

if so, we should not have that, and instead just clone it

src/flows/components/FlowContextMenu.tsx
```diff
@@ -16,27 +16,47 @@ interface Props {
...
+  const {remove, clone} = useContext(FlowListContext)
 
   const handleDelete = () => {
     event('delete_notebook')
     remove(id)
   }
 
+  const handleClone = () => {
+    event('clone_notebook')
+    clone(id)
```
drdelambre, line 28: would we want to redirect to the cloned notebook?

src/flows/components/FlowContextMenu.tsx
Asalem1: It does. I tried doing that prior to this but it doesn't look like the current context.menu or context support a button as a child, so i was getting some TS issues throwing a button in. I can try refactoring this to be the same pattern as the other parts of the app where there's an icon and clicking it will toggle that little context item text to select the action to perform

src/flows/components/FlowContextMenu.tsx
Asalem1: Can do. Shouldn't that be in the actual action and not this clone handler?

src/flows/components/FlowContextMenu.tsx
drdelambre: i think you can ignore all the context stuff and through a button in there with an icon and a componentsize to do the trick

src/flows/components/FlowContextMenu.tsx
drdelambre: tough call. i think we'd always want to go to the clone given the few examples i've heard people talk about, but having redirects at the action layer has bitten us in the past on these sorts of things

i'd keep it on the view layer here in case we want to decouple routing from asset generation

Asalem1: Fiar enough

drdelambre (approved):
it's beautiful

PR #1362, opened 2021-05-05
fix: Don't clobber monitor-ci pipeline artifacts with the same name

### LINKED ISSUE #1317: artifacts from monitor-ci pipeline get overridden if they have the same filename
The script to run the monitor-ci pipeline overwrites artifact videos if they have the same filename. 

Example:
- `/shared/dashboardView.mp4` and `/cloud/dashboardView.mp4` = only one of these files will exist in the artifacts.

This line is the problem: https://github.com/influxdata/ui/blob/master/scripts/run-monitor-ci-tests.bash#L168

We need to include the directory in the `filename` somehow.

joelfried:
Closes #1317 

Updates the bash script that kicks off the monitor-ci pipeline so that when test files have the same name in the shared folder we don't clobber the other failures.

I temporarily forced test failures on this branch to show this approach worked.  The failed run within the monitor-ci pipeline can be viewed [here](https://app.circleci.com/pipelines/github/influxdata/monitor-ci/750/workflows/e512a27e-8afa-4a52-836e-520718099e69/jobs/6437/artifacts)

Previously these two tests would have only had one distinct artifact - you can see by the test failures below (screenshots taken from my watching of the video artifacts) that they were distinct failures.

image
image


scripts/run-monitor-ci-tests.bash
```diff
@@ -165,10 +165,18 @@ do
...
+							if [[ *"shared"* == "${path}" ]] ; then
+							  curl -L -s --request GET \
+                  --output "monitor-ci/test-artifacts/results/shared/${name}/${filename}" \
+                  --url "${url}" \
+                  --header "Circle-Token: ${API_KEY}"
+              else
+                curl -L -s --request GET \
+                  --output "monitor-ci/test-artifacts/results/${name}/${filename}" \
+                  --url "${url}" \
+                  --header "Circle-Token: ${API_KEY}"
```
genehynson, line 178: can we reuse the curl command just use a variable for the value for the `--output` flag? Then the if/else block would just be setting the `output` variable. Might simplify this a bit!

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1362)

scripts/run-monitor-ci-tests.bash
joelfried: Good idea - done!

genehynson (approved):

PR #1363, opened 2021-05-05
feat: improve ui image build time, size, and security

glinton:
Currently the `Dockerfile.chronograf.docker` file uses an unnecessarily large base image, installs unneeded packages, and insecurely runs a downloaded script. This pr uses a slimmer base image and saves ~475MB per base image, speeds up required dependency installs by ~38 seconds, and doesn't require insecurely running downloaded scripts.

```
$ time docker build -t test/ui -f ./originalDockerfilePreReq .
...
Successfully tagged test/ui:latest

real    0m44.174s
user    0m0.138s
sys     0m0.082s

$ time docker build -t test/ui2 -f ./newDockerfilePreReq .
...
Successfully tagged test/ui2:latest

real    0m6.009s
user    0m0.588s
sys     0m0.685s

$ docker images --format "table {{.Repository}}\t{{.Size}}"
REPOSITORY                       SIZE
test/ui2                         129MB
test/ui                          604MB
```

### DIFF
--- a/docker/Dockerfile.chronograf.prod
+++ b/docker/Dockerfile.chronograf.prod
@@ -1,14 +1,5 @@
-FROM ubuntu:bionic AS base
-
-RUN apt-get update -y && \
-    apt-get install -y \
-        nginx \
-        build-essential \
-        curl \
-        git
-
-FROM base AS repo
-WORKDIR /repo
+# from node latest alpine 3.12
+FROM node:alpine3.12 AS repo
 
 # env vars to configure the system
 
@@ -38,17 +29,20 @@ ARG INJECT_HEADER
 # Injecting strings into the html never went wrong. we use this for google tag manager
 ARG INJECT_BODY
 
-ENV PATH="/node/bin:${PATH}"
-RUN mkdir /node && \
-    curl -sL https://git.io/n-install | N_PREFIX=/node bash -s -- -q
-
 COPY . /repo/ui
 WORKDIR /repo/ui
 
+RUN apk \
+    --update \
+    --no-cache \
+    --virtual build-dependencies \
+    add \
+    git
+
 # these are all run together as docker's caching mechanism
 # makes big steps like yarn install expensive
-RUN npx yarn install --production=false && \
-    npx yarn generate && \
+RUN yarn install --production=false && \
+    yarn generate && \
     $(npm bin)/webpack --config ./webpack.${WEBPACK_FILE}.ts --bail && \
     rm -rf ./node_modules
 


docker/Dockerfile.chronograf.prod
```diff
@@ -1,14 +1,5 @@
-FROM ubuntu:bionic AS base
```
Asalem1, line 1: i'd love @drdelambre 's take on why we were using `ubuntu:bionic` as the base rather than `node:alpine`.

docker/Dockerfile.chronograf.prod
```diff
@@ -38,17 +29,20 @@ ARG INJECT_HEADER
...
+    --no-cache \
+    --virtual build-dependencies \
+    add \
+    git
+
 # these are all run together as docker's caching mechanism
 # makes big steps like yarn install expensive
-RUN npx yarn install --production=false && \
-    npx yarn generate && \
+RUN yarn install --production=false && \
```
Asalem1, line 44: wouldn't we need to install yarn prior to this command?

docker/Dockerfile.chronograf.prod
```diff
@@ -38,17 +29,20 @@ ARG INJECT_HEADER
...
-ENV PATH="/node/bin:${PATH}"
-RUN mkdir /node && \
-    curl -sL https://git.io/n-install | N_PREFIX=/node bash -s -- -q
-
 COPY . /repo/ui
 WORKDIR /repo/ui
 
+RUN apk \
+    --update \
+    --no-cache \
```
Asalem1, line 37: is there a reason we wouldn't want to cache @drdelambre ?

docker/Dockerfile.chronograf.prod
glinton: the alpine node image has yarn pre-installed :tada:

docker/Dockerfile.chronograf.prod
glinton: smaller image size mainly, but that doesn't matter since the final image is built from `nginx:stable-alpine` (see line 57/51 below)

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1363)

docker/Dockerfile.chronograf.prod
drdelambre: because we are using alpine

Asalem1: huh?

drdelambre: yeah. that's why i dont really get this PR.

We build off of bionic (should prolly update to focal if anything), as some form of ubunutu is usually cached where ever you're building (locally or in ci), because it has better developer ergonomics when working on an image (to either extend functionality or debug why things aren't being built. By removing the last stage of the build, you would have a working shell, most stack overflow results for "linux + thing i want" are directly copy pasta instead of learning the weird way of installing build tools on alpine, and most of the command line tools you'd want when poking around to find that error are available.

After the image's assets are built, it loads those minified generated assets into a minified nginx alpine image that only includes the bare minimum required to serve those files, then throws all of the dev environment away so that we can have an image that is small, fast to push around the cloud and download for local dev / testing / whatever.

glinton: changing the base image is more a side effect of this pr. the main thing is to speed up the building of the assets. smaller image means faster pulls. more specific image means less time installing tools (most of which were un-needed), and it's more secure: no more `curl -sL https://git.io/n-install | N_PREFIX=/node bash -s -- -q` :see_no_evil: :scream:

glinton:
The largest gains can be seen in CI, where this dockerfile gets called multiple times. A single `build the ui docker image` task from a typical pr and this one shows a time reduction of ~3 minutes. :tada: for less time in CI.

Before:
image

After:
image

docker/Dockerfile.chronograf.prod
glinton: Maybe you could use a different dockerfile for local debugging, but have a safe and fast one for CI, where those factors tend to outweigh convenience.

docker/Dockerfile.chronograf.prod
```diff
@@ -38,17 +29,20 @@ ARG INJECT_HEADER
 # Injecting strings into the html never went wrong. we use this for google tag manager
 ARG INJECT_BODY
 
-ENV PATH="/node/bin:${PATH}"
-RUN mkdir /node && \
-    curl -sL https://git.io/n-install | N_PREFIX=/node bash -s -- -q
```
jdstrand, line 43: :hugs:

subirjolly (approved):

Issue #1207, opened 2021-04-14
Epic: Auto Refresh Dashboard / Pause Cells 


ShmuelLotman:
### User Story
   As a user, I am now able to interact with my dashboards in a more dynamic way, display them on monitors with greater ease, maintain a more firm grasp of my query totals and overall costs, and acquire more granular control over individual cells within my dashboard. 

### Context

Approximately a year or so ago, we provided users with the ability to set dashboards to refresh at preset intervals automatically. While this was a convenient way to interact with dashboards, the frequent complaint was that users would leave their pages open overnight or otherwise unattended and thus accrue significant charges on their account. As a response to this, we disabled the feature entirely. 

During subsequent exploration of dashboard feature sets, the opportunity to reintroduce this feature presented itself as a natural extension to the work on refreshing individual cells, as a way to further expand the available features dashboard view provided to users. 

Furthermore, due to the existing query cache structure, the ability to pause cells, thereby removing them from the cycle of automatic refreshes, (as well as resume), is something that can be integrated rather seamlessly into the work scoped out within this epic. 

### Acceptance Criteria 
 In order to re-enable this feature, it is critical we address the primary concern of abandoned dashboards and runaway query totals. This can be tackled from two perspectives:
 1.  A two-step process to enable auto refresh, with clearly labeled steps that, by definition, mean a user has the clear intention to enable this feature
 2.  The ability to add an "inactivity timeout," such that a user who leaves a dashboard for a particular amount of time can be assured that their query totals will not continue to increase past an acceptable point in their absence. 

Additionally, the two-step process should provide a user with the ability to set a duration, which will serve as a hard stop on their auto refresh process, as well as the ability to set the refresh frequency. 

Finally, as mentioned above, users should be able to pause cells, thereby removing them from the auto refresh polling system, and resume them, which will put them back into the same flow. 

## Summary
 The following is a list of features required for an MVP of this feature set: 
  1. ** Two step process to enable auto refresh **
  2. ** Ability to select auto refresh interval **
  3. ** Ability to select deadline (duration of refresh) after which auto refresh stops automatically **
  4. ** Ability to select inactivity timeout, which automatically stops a user's auto refresh process when hit**
  5. ** Ability to pause cells & resume them, which removes and replaces them in the refresh process, respectively **

### Extended Outlook
[...]

PR #1364, opened 2021-05-05
feat: pause cell view only implementation

### LINKED ISSUE #1224: Dashboard Auto Refresh: Enable Pause Cell Feature

ShmuelLotman:
Closes #1224 

This PR implements Pause Cell functionality in a way that only applies to the view layer and avoids targeting the query cache or getting involved with cache invalidation in any way. In this way, we can hopefully still introduce this feature, without necessitating a discussion that, while it should still be had for the sake of maintainability and future efforts, is beyond the scope of the current epic.


Asalem1:
Did we ever get product clarity on the issues holding off the previous PR?

https://github.com/influxdata/ui/issues/1207#issuecomment-825067115

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1364)

ShmuelLotman:
@Asalem1 the view only implementation should address the concerns discussed in that comment you linked. Because this is a view only implementation and a part of the proof of concept level effort, it should allow us to see how users respond without worrying about interaction with, or invalidation of, the query cache and the potential side effects. This current implementation means that the feature operates within the same parameters as the broader auto refresh feature.

Asalem1:
> @Asalem1 the view only implementation should address the concerns discussed in that comment you linked. Because this is a view only implementation and a part of the proof of concept level effort, it should allow us to see how users respond without worrying about interaction with, or invalidation of, the query cache and the potential side effects. This current implementation means that the feature operates within the same parameters as the broader auto refresh feature.

no doubt, but it doesn't necessarily address product concerns about behavior when refreshing a screen / navigating to & from a dashboard, or configuring a dashboard cell

drdelambre:
weren't most of the product concerns stemmed from the fact that the last implementation was so tightly coupled with the query cache?

ShmuelLotman:
@drdelambre that was my thinking exactly. This functions exactly in line with the auto refresh functionality. Any concerns that should surround this issue would be the same ones that surround the auto refresh, and so far there hasn't been any of that in feedback... I'm thinking we let this fly and see how people respond.

Asalem1:
> weren't most of the product concerns stemmed from the fact that the last implementation was so tightly coupled with the query cache?

I think that was the engineering concerns of this. The product concerns stemmed from teasing out these uses cases as far as expectations.

ShmuelLotman:
@Asalem1 I believe the product direction we were taking with this was more of a feedback based approach, where we would see how these features were appreciated/ not and decide whether to keep them or not based on that.

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -1392,4 +1392,198 @@ csv.from(csv: data) |> filter(fn: (r) => r.bucket == v.bucketsCSV)`
...
+    cy.get('@org').then(({id: orgID, name}: Organization) => {
+      cy.createDashboard(orgID).then(({body}) => {
+        cy.fixture('routes').then(({orgs}) => {
+          cy.visit(`${orgs}/${orgID}/dashboards/${body.id}`)
+          cy.getByTestID('tree-nav')
+        })
+      })
+      cy.window().then(win => {
+        win.influx.set('pauseCell', true)
+        win.influx.set('newAutoRefresh', true)
```
Asalem1, line 1406: Might need to add a wait for one second here. There’s flakiness around feature flag setting in cypress tests

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -1392,4 +1392,198 @@ csv.from(csv: data) |> filter(fn: (r) => r.bucket == v.bucketsCSV)`
...
+      cy.getByTestID('page-title').click()
+      cy.getByTestID('renamable-page-title--input')
+        .clear()
+        .type('blah')
+      cy.getByTestID('save-cell--button').click()
+    })
+
+    cy.getByTestID('button').click()
+    cy.getByTestID('switch-to-script-editor').should('be.visible')
+    cy.getByTestID('switch-to-script-editor').click()
```
Asalem1, line 1452: I think we want to use cy.focus here

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -1392,4 +1392,198 @@ csv.from(csv: data) |> filter(fn: (r) => r.bucket == v.bucketsCSV)`
...
+    cy.getByTestID('toolbar-tab').click()
+
+    cy.getByTestID('overlay').within(() => {
+      cy.getByTestID('flux-editor')
+        .should('be.visible')
+        .click()
+        .focused()
+        .type(query2)
```
Asalem1, line 1460: I think we can just use the focus

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -1392,4 +1392,198 @@ csv.from(csv: data) |> filter(fn: (r) => r.bucket == v.bucketsCSV)`
...
+    cy.getByTestID('enable-auto-refresh-button').click()
+    cy.getByTestID('auto-refresh-overlay').should('be.visible')
+    cy.getByTestID('auto-refresh-overlay').within(() => {
+      cy.getByTestID('autorefresh-dropdown--button').click()
+      cy.getByTestID('auto-refresh-5s').click()
+      cy.getByTestID('refresh-form-activate-button').click()
+      cy.wait('@secondCellQuery')
+      cy.wait('@firstCellQuery')
+
+      cy.on('fail', err => {
```
Asalem1, line 1486: Remind me why we’re testing this? What’s it doing would be helpful as a comment

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -1392,4 +1392,198 @@ csv.from(csv: data) |> filter(fn: (r) => r.bucket == v.bucketsCSV)`
...
+    cy.get('@org').then(({id: orgID, name}: Organization) => {
+      cy.createDashboard(orgID).then(({body}) => {
+        cy.fixture('routes').then(({orgs}) => {
+          cy.visit(`${orgs}/${orgID}/dashboards/${body.id}`)
+          cy.getByTestID('tree-nav')
+        })
+      })
+      cy.window().then(win => {
+        win.influx.set('pauseCell', true)
+        win.influx.set('newAutoRefresh', true)
```
Asalem1, line 1505: Same as veggie about the wait

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -1392,4 +1392,198 @@ csv.from(csv: data) |> filter(fn: (r) => r.bucket == v.bucketsCSV)`
...
+    cy.getByTestID('auto-refresh-overlay').within(() => {
+      cy.getByTestID('autorefresh-dropdown--button').click()
+      cy.getByTestID('auto-refresh-5s').click()
+      cy.getByTestID('refresh-form-activate-button').click()
+      cy.wait('@secondCellQuery')
+    })
+    cy.getByTestID('cell blah').within(() => {
+      cy.getByTestID('cell-context--toggle').click()
+    })
+    cy.getByTestID('cell-context--pause').click()
```
Asalem1, line 1587: How is this test testing the pause functionality?

Asalem1 (approved):
Looks good, tests need a tweaking

cypress/e2e/shared/dashboardsView.test.ts
ShmuelLotman: Because the `done` call in the first intercept if clause will only fire if the first query is fired, which will only happen once it is unpaused, so this is testing that after we unpause, it goes back inside the autorefresh loop

Asalem1: Gotcha. That makes sense. Is there any way we could get a comment on the test to explain what's happening because I know I'm gonna see this again and be like "what?"

ShmuelLotman: Done. Pushed up a comment to explain it

PR #1366, opened 2021-05-05
feat: copy flux to clipboard as client code

### LINKED ISSUE #576: Notebooks: panel -> clipboard interface
so get this....

for any panel in a notebook, we can get the flux used to generate it's results.
also
we have the ability to copy code snippets to a user's clipboard, by clicking on a button

so why can't I have an experience within notebooks that is similar to the following feature available in chromium browsers?
Peek 2021-01-29 10-12

add a button to the panel header somewhere that lets you select a target and generates a code snippet for you in your clipboard.

initial targets i can think of off the top of my head:
 - curl
 - fetch
 - python (check out the initialize and execute sections in the data sources interface for python)
 
 weird things to watch out for:
  - we might need to allow the user to specify a token before generating the code
  - a query's orgid is determined by the query text (look at the query context for more info)
 - we gotta use absolute addresses to capture the domain that the user is currently on (us-west vs us-east needs to be baked into the generated url)

### Design Notes
Design Lead @juliajames12 

Given: A Flux script notebook panel is active
And: I can see the panel menu
When: I click Export As > Client Library
Then: I can see a sub-menu with all 11 client library options:

- Go
- Arduino
- C#
- Java
- Kotlin
- PHP
- Ruby
- Scala
- Swift
- JavaScript
- Python

When: I click Python or any of the client libraries
Then: I see a modal with code to Install Package, Initialize Client, and Execute Flux

Working File: https://www.figma.com/file/8S8IH8mimYIwanZ6keIQ6U/Notebooks-Prototype?node-id=2512%3A480

Prototype: https://www.figma.com/proto/8S8IH8mimYIwanZ6keIQ6U/Notebooks-Prototype?node-id=2512%3A480&scaling=min-zoom&page-id=1915%3A0

subirjolly:
Closes #576 

We need to be able to copy every panel in notebooks as client code. So, will be adding a dropdown button on each of the panels which, when clicked on, will display a few client options and when you click on any of those languange options, your code will be copied into the clipboard.


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1366)

src/flows/components/panel/FlowPanel.tsx
```diff
@@ -145,6 +150,12 @@ const FlowPanelHeader: FC<HeaderProps> = ({
             </FeatureFlag>
             <PanelVisibilityToggle id={id} />
             <RemovePanelButton id={id} />
+            {isFlagEnabled('CopyClientCodeToClipboard') && (
+              <CopyToClipboardProvider>
+                <CopyToClipboardButton />
+                <PanelQueryOverlay panelId={id} />
+              </CopyToClipboardProvider>
```
drdelambre, line 157: Use the popupcontext for this instead. See the exporttodashboard and exporttotask examples in src/flows.

src/flows/components/panel/PanelQueryOverlay.tsx
```diff
@@ -0,0 +1,84 @@
+// Libraries
+import React, {
+  useContext,
+  FC,
+  memo,
+  useState,
+  useEffect,
+  useCallback,
+} from 'react'
+import {withRouter, RouteComponentProps} from 'react-router-dom'
```
drdelambre, line 10: Doesn't seem to be using these?

src/flows/components/panel/PanelQueryOverlay.tsx
```diff
@@ -0,0 +1,84 @@
+// Libraries
+import React, {
+  useContext,
+  FC,
+  memo,
```
drdelambre, line 5: What caused the addition of the memo? That thing haunted my earlier code, so I'm sensitive about it, but lemme know if it was useful for something specific

src/flows/components/panel/PanelQueryOverlay.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <Overlay visible={visible}>
+      <Overlay.Container>
+        <Overlay.Header
+          title={`Export to ${CLIENT_DEFINITIONS[contentID].name} Client Library`}
+          onDismiss={() => setVisibility(!visible)}
+        />
+        <Overlay.Body>
+          <TemplateProvider variables={{token: TOKEN_PLACEHOLDER}}>
```
drdelambre, line 68: Remove this override. You're effectively creating two sources of truth here across two components when you should be defering to WriteDataContext as the sole owner of this variable's definition.

src/flows/context/query.tsx
```diff
@@ -83,6 +85,34 @@ const _walk = (node, test, acc = []) => {
...
+
+export const getBucketsFromAST = (ast: File) => {
+  return _walk(
+    ast,
+    node =>
+      node?.type === 'CallExpression' &&
+      node?.callee?.type === 'Identifier' &&
+      node?.callee?.name === 'from' &&
+      node?.arguments[0]?.properties[0]?.key?.name === 'bucket'
+  ).map(node => lodashGet(node, 'arguments.0.properties.0.value.value', ''))
```
drdelambre, line 113: We don't use lodash for this since typescript handles it with the obj?.prop?.val format

src/writeData/components/ExecuteCodeBlock.tsx
```diff
@@ -0,0 +1,85 @@
...
+  useEffect(() => {
+    if (bucket && selectedBucket !== bucket?.name) {
+      setSelectedBucket(bucket?.name)
+    }
+  }, [bucket, selectedBucket, currentQuery, bucket?.name, query])
+
+  useEffect(() => {
+    const executeCode = def.execute.toString()
+    if (!query) {
+      setExecuteCodeBlock(executeCode.replace(QUERY_PLACEHOLDER, def.query))
```
drdelambre, line 60: Use the template context's transform function and not replace. We want one definition of how a template is parsed / variables are replaced in it, not many. Let's say there's a conflict on <%= and we have to change it to <[[ or else the whole internet falls over. We want that definition to be centralized so that we don't have some case still using <%= somewhere that evaded our best grep usage

src/writeData/components/ExecuteCodeBlock.tsx
```diff
@@ -0,0 +1,85 @@
...
+      return
+    }
+
+    const ast = parse(currentQuery)
+
+    updateBucketInAST(ast, selectedBucket)
+    setExecuteCodeBlock(
+      executeCode.replace(QUERY_PLACEHOLDER, format_from_js_file(ast))
```
drdelambre, line 68: Same here

src/writeData/components/ExecuteCodeBlock.tsx
```diff
@@ -0,0 +1,85 @@
...
+  )
+
+  useEffect(() => {
+    if (bucket && selectedBucket !== bucket?.name) {
+      setSelectedBucket(bucket?.name)
+    }
+  }, [bucket, selectedBucket, currentQuery, bucket?.name, query])
+
+  useEffect(() => {
+    const executeCode = def.execute.toString()
```
drdelambre, line 58: Why are you manipulating the execute code instead of ensuring that the query variable is set in the template? Like how WriteDataContext ensures bucket and token are set. Then it takes care of all this for you

src/writeData/components/ExecuteCodeBlock.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+const QUERY_PLACEHOLDER = '<%= query %>'
+
+const ExecuteCodeBlockProvider: FC<Props> = ({query, contentID, children}) => {
+  const def = CLIENT_DEFINITIONS[contentID]
+  const [currentQuery, setQuery] = useState(query ?? def.query)
+  const [executeCodeBlock, setExecuteCodeBlock] = useState('')
+  const {bucket} = useContext(WriteDataDetailsContext)
+  const [selectedBucket, setSelectedBucket] = useState(
```
drdelambre, line 47: Executecodeblock should be using the bucket set by WriteDataContext, and not have its own version that it has to try and keep in sync with it. On query load, grab the bucket from the query, and set WriteDataContext's selected bucket to that value

src/writeData/components/WriteDataDetailsContext.tsx
```diff
@@ -42,6 +42,12 @@ const WriteDataDetailsProvider: FC = ({children}) => {
   const [bucket, setBucket] = useState(buckets[0])
   const [token, setToken] = useState(tokens[0])
 
+  useEffect(() => {
```
drdelambre, line 45: How come this was added?

src/writeData/components/WriteDataDetailsContext.tsx
```diff
@@ -65,7 +71,7 @@ const WriteDataDetailsProvider: FC = ({children}) => {
   }, [variables, org?.name])
 
   useEffect(() => {
-    if (bucket?.name === variables.bucket) {
+    if (!bucket || bucket?.name === variables.bucket) {
```
drdelambre, line 74: These aren't equivalent?

src/writeData/components/WriteDataDetailsContext.tsx
```diff
@@ -76,7 +82,7 @@ const WriteDataDetailsProvider: FC = ({children}) => {
   }, [variables, bucket?.name])
 
   useEffect(() => {
-    if (token?.token === variables.token) {
+    if (!token || token?.token === variables.token) {
```
drdelambre, line 85: Same here

src/writeData/containers/ClientLibrariesPage.tsx
```diff
@@ -37,7 +40,22 @@ interface SampleProps {
...
+  name: string
+}
+
+export const ExecuteCodeSampleBlock: FC<ExecuteCodeSampleProps> = ({name}) => {
+  const {executeCodeBlock} = useContext(ExecuteCodeBlockContext)
+
+  return (
+    <>
+      <h4>{name}</h4>
+      <CodeSnippet text={executeCodeBlock} />
```
drdelambre, line 53: This would just be def.execute like the other examples if you use that component to set the variable

src/writeData/containers/ClientLibrariesPage.tsx
```diff
@@ -89,40 +110,43 @@ const ClientLibrariesPage: FC = () => {
     <GetResources
       resources={[ResourceType.Authorizations, ResourceType.Buckets]}
     >
-      <TemplateProvider>
+      <TemplateProvider
+        variables={{token: TOKEN_PLACEHOLDER, bucket: BUCKET_PLACEHOLDER}}
```
drdelambre, line 114: Here is another example of something that defines these variables. Make it one thing, which is gonna be WriteDataContext

drdelambre (changes requested):
Pick an owner to consolidate the source of truth on these components. Keeping them all in sync while maintaining this code is going to be rough. And have executecodeblock own setting the variable, not owning the query

src/flows/components/panel/PanelQueryOverlay.tsx
subirjolly: Copy Pasta. Removed.

src/flows/components/panel/PanelQueryOverlay.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+      return stage?.text ?? ''
+    },
+    [generateMap]
+  )
+
+  useEffect(() => {
+    const currentQuery = getPanelQuery(panelId)
+    if (currentQuery !== query) {
+      setQuery(currentQuery)
```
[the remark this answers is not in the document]
subirjolly: removed the query in local state

src/writeData/components/WriteDataDetailsContext.tsx
subirjolly: merge mess, I think! Will Remove!

src/writeData/components/WriteDataDetailsContext.tsx
subirjolly: they should be but it was breaking for me if the bucket was undefined.

subirjolly: Kapture 2021-05-19 at 13 19 24

src/writeData/components/WriteDataDetailsContext.tsx
subirjolly: same as above

src/writeData/containers/ClientLibrariesPage.tsx
subirjolly: nope! let's talk :)

drdelambre (approved):

Asalem1 (commented):
A test or it didn’t happen

subirjolly:
> A test or it didn’t happen

I will add tests once the flow is completely ready. Will happen in the next story to connect panel to this.

Asalem1:
> > A test or it didn’t happen
> 
> I will add tests once the flow is completely ready. Will happen in the next story to connect panel to this.

That’s what they all say :(

Asalem1 (approved):

subirjolly:
> > > A test or it didn’t happen
> > 
> > 
> > I will add tests once the flow is completely ready. Will happen in the next story to connect panel to this.
> 
> That’s what they all say :(

Haha. You know I am not one of them!

PR #1372, opened 2021-05-06
Feat: add annotations to single stat plus line graphs

jrenee42:
Closes #21388

refactored out the annotations common code into another class; so it can be shared with another graph type.

had to put the annotation layer on top so that the editing and deleting work

additionally, the annotations only work if the 'useGiraffeGraphs' flag is on


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1372)

src/visualization/components/annotationController.tsx
```diff
@@ -0,0 +1,143 @@
+// React
```
hoorayimhelping, line 1: this file doesn't have any JSX in it, right? it should just be a ts file then.

for naming, how about just `annotations.ts` ? when I think of a controller or a manager, I think of a centralized object that delegates to sub objects or subroutines, whereas this is more of collection of similarly-concept functions

src/visualization/components/annotationController.tsx
```diff
@@ -0,0 +1,143 @@
...
+import {getErrorMessage} from 'src/utils/api'
+
+import {
+  AnnotationLayerConfig,
+  InfluxColors,
+  InteractionHandlerArguments,
+} from '@influxdata/giraffe'
+
+export const makeAnnotationClickListener = (
+  dispatch: Dispatch<any>,
```
hoorayimhelping, line 22: this method only dispatches two possible things, right? the output of `writeThenFetchAndSetAnnotations` and the output of `notify(createAnnotationFailed(getErrorMessage(err)))`

is it possible to template this dispatch call to those two return types?

something like
```ts
import {Action as NotificationAction} from 'src/shared/actions/notifications'
const SuccessAction = ReturnType<typeof setAnnotations>

Dispatch<SuccessAction | NotificationAction>
``` 

?

src/visualization/components/annotationController.tsx
```diff
@@ -0,0 +1,143 @@
...
+      )
+    )
+  }
+
+  return singleClickHandler
+}
+
+const makeAnnotationClickHandler = (
+  cellID: string,
+  dispatch: Dispatch<any>,
```
hoorayimhelping, line 69: same question here about tightening up the typing around `Dispatch`

src/visualization/types/SingleStatWithLine/view.tsx
```diff
@@ -161,161 +191,162 @@ const SingleStatWithLine: FC<Props> = ({properties, result, timeRange}) => {
...
+            </div>
+          )
+        }}
+      </LatestValueTransform>
     )
+
+    return <Plot config={config}>{statPortion}</Plot>
+  }
+  // helper functions
+  function makeConfig() {
```
hoorayimhelping, line 313: I'm not a huge fan of this - creating a method that takes no input and is only called once. I understand the intent, and how it came about (pulling out a big config object used once), but since there's no configuration available via method properties, and since it's only used in one place, all it effectively does is make execute a goto when I'm reading the code that calls `makeConfig`. 

I'd strongly prefer to leave this configuration inline and only pull it out into a method if it's called more than once. I  want to limit premature abstractions as much as possible

src/visualization/types/SingleStatWithLine/view.tsx
```diff
@@ -161,161 +191,162 @@ const SingleStatWithLine: FC<Props> = ({properties, result, timeRange}) => {
...
+      svgTextAttributes: {
+        'data-testid': 'single-stat--text',
+      },
+    }
+
+    config.layers.push(statLayer)
+
+    // adding this *after* the statLayer, it has to be the top layer
+    // for clicking to edit to function.  (if it is not the top layer it shows,
+    // but the annotations are not editable)
```
hoorayimhelping, line 245: :+1: helpful comment

src/visualization/components/annotationController.tsx
```diff
@@ -0,0 +1,143 @@
...
+          dimension: 'x',
+          pin: 'start',
+        }
+      }),
+    }
+
+    return annotationLayer
+  }
+  return null
+}
```
hoorayimhelping, line 143: do you think there's enough logic in these components to add tests to them? 

I think at the least we should update the annotations e2e tests to cover single stats.

hoorayimhelping (changes requested):
nice work! a few thoughts / suggestions

src/visualization/components/annotationController.tsx
jrenee42: i'll change it to annotationUtils.ts ; annotations is not description enough (i want to have a sense of what it does from the name)

src/visualization/components/annotationController.tsx
jrenee42: let's talk about this in person, I don't have enough context (i need to learn more about dispatch and actions)

jrenee42: conclusion:  we are using <any> because it is non-trivial and wonky.

src/annotations/components/controlBar/AnnotationsControlBar.tsx
```diff
@@ -61,7 +61,7 @@ export const AnnotationsControlBar: FC = () => {
           <TextBlock
             backgroundColor={InfluxColors.Obsidian}
             textColor={InfluxColors.Mist}
-            text="*Currently, we only support annotations on XY Plots"
+            text="*Currently, we only support annotations on Plots of type 'Graph' and 'Graph + Single Stat'"
```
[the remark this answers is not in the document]
hoorayimhelping: Ah, I understand what you meant in our Zoom call. This is a very tiny change

PR #1373, opened 2021-05-06
feat: adding query debugging to flow panels

drdelambre:
I found an error while helping someone where the markdown panel wasn't generating any queries. It was hard to realize that, so i made quick function to expose what the internal state thinks your query should be at every panel

Peek 2021-05-06 14-21


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?vars%5Blighthouse_score_branch%5D=1373)

src/flows/components/panel/FlowPanel.tsx
```diff
@@ -61,6 +64,49 @@ const FlowPanelHeader: FC<HeaderProps> = ({
...
+          source: curr.text,
+          visualization: i.modifier,
+        }
+      })
+
+      return acc
+    }, {})
+
+    /* eslint-disable no-console */
+    flow.data.allIDs.forEach(i => {
```
Asalem1, line 80: I know what this is and why this is here, but future me will be like, "the heck?" A comment on how this adds observability, etc... would be dope

src/flows/context/flow.query.tsx
```diff
@@ -242,39 +245,27 @@ export const FlowQueryProvider: FC = ({children}) => {
     Promise.all(
       map
         .reduce((acc, curr) => {
-          acc.push({
-            text: curr.text,
-            instances: curr.instances.filter(i => !i.modifier).map(i => i.id),
-          })
-
```
Asalem1, line 249: why were we doing this to begin with with the concat underneath?

src/flows/context/flow.query.tsx
```diff
@@ -242,39 +245,27 @@ export const FlowQueryProvider: FC = ({children}) => {
...
-          })
+          flow.meta.update(stage.instance, {loading: RemoteDataState.Loading})
           return query(stage.text)
             .then(response => {
-              stage.instances.forEach(pipeID => {
-                flow.meta.update(pipeID, {loading: RemoteDataState.Done})
-                forceUpdate(pipeID, response)
-              })
+              flow.meta.update(stage.instance, {loading: RemoteDataState.Done})
+              forceUpdate(stage.instance, response)
```
Asalem1, line 262: these are so much better

src/flows/context/flow.query.tsx
```diff
@@ -242,39 +245,27 @@ export const FlowQueryProvider: FC = ({children}) => {
...
             curr.instances
               .filter(i => i.modifier)
               .map(i => ({
                 text: i.modifier,
-                instances: [i.id],
+                instance: i.id,
               }))
           )
         }, [])
-        .filter(stage => !!stage.instances.length)
```
Asalem1, line 259: what was this doing and what's changed to render this step redundant?

Asalem1 (approved):
Might need a test here or there

src/flows/context/flow.query.tsx
drdelambre: in the before time, we used to join the data requirements of panels together, and send the requests as one. now each panel seems to make a slight modification to the query to scope the results it wants. simpletable adds a limit, visualization adds an aggregate, etc. So instead of doing the join (where things that dont define a query get squished together) those things just... dont get data.

src/flows/context/flow.query.tsx
drdelambre: nest on nest on nest!

src/flows/context/flow.query.tsx
drdelambre: we dont do the squishing anymore. first.. by conventions (which caused this bug), now codified

PR #1390, opened 2021-05-11
feat(notebooks): pass org id as query param when getting all notebooks

### LINKED ISSUE #1383: Notebooks: Need to be able to get notebooks for the user's active org in OSS
The new notebooks service does not include the user or org ID as parameters in the path. For the user this is fine, since the user id can be obtained from the authorization. But for OSS with its ability for a user to belong to multiple organizations, there currently isn't a way to obtain the "active" organization for a user unless it is sent by the frontend with the request, so the response from a `GET` request to `/api/v2/notebooks` may not be desirable for an OSS user belong to multiple orgs.

I think an easy way to get around this for now would be to pass the user's active org as a query parameter with the `GET` request to `/api/v2/notebooks`. This shouldn't require very much effort and would allow the OSS backend to respond appropriately to a request for the a user who belongs to multiple organizations.

Some other options might be:

- Respond with all of the notebooks for ALL of the organizations the user belongs to. This would not be consistent with requests for other resources, like dashboards.
- Make notebooks per-user user on OSS. So a request from each user would return only notebooks created by that user - this would also not be consistent with other resources, and would require an association between notebooks and users.
- Modify the OSS backend so that the user session maintains awareness of the active user, and can obtain the user's active org without the frontend request including that information. Among other things, this would require the org switching frontend component to send a request to the backend when switching orgs, and the backend to process that request and store the result somehow. In the long run, this might be a good option, but I think something like this should be done in conjunction with how IDPE ends up implementing multi-org support.

wbaker85:
Closes #1383

This will allow the multi-org-per-user feature of OSS to work with the minimal amount of effort.

We may want to revisit this down the line when InfluxDB cloud implements multi-org-per-user.

### DIFF
--- a/src/client/swagger/notebooks.yml
+++ b/src/client/swagger/notebooks.yml
@@ -57,6 +57,11 @@ paths:
   /notebooks:
     get:
       description: get all records
+      parameters:
+        - name: orgID
+          in: query
+          type: string
+          required: true
       produces:
         - application/json
       responses:
--- a/src/client/swagger/notebooks.yml
+++ b/src/client/swagger/notebooks.yml
@@ -57,11 +57,6 @@ paths:
   /notebooks:
     get:
       description: get all records
-      parameters:
-        - name: orgID
-          in: query
-          type: string
-          required: true
       produces:
         - application/json
       responses:
--- a/src/flows/context/notebooks.api.tsx
+++ b/src/flows/context/notebooks.api.tsx
@@ -76,7 +76,7 @@ export const deleteAPI = async (ids: DeleteNotebookParams) => {
 }
 
 export const getAllAPI = async (orgID: string) => {
-  const res = await getNotebooks({orgID})
+  const res = await getNotebooks({query: {orgID}})
   if (res.status != 200) {
     throw new Error(res.data.message)
   }


src/client/swagger/notebooks.yml
```diff
@@ -57,6 +57,11 @@ paths:
   /notebooks:
     get:
       description: get all records
+      parameters:
+        - name: orgID
+          in: query
+          type: string
+          required: true
```
genehynson, line 64: This file is actually generated so we'll need to make the change to notebooksd over in IDPE. I'm currently working on a massive refactor to notebooksd and I could include this in the refactor or in a PR right after the refactor is complete https://github.com/influxdata/idpe/pull/10606

wbaker85: Ahhh, I didn't realize that. Whatever's easiest for you as far as when it include it is fine with me. It's not really blocking for me on OSS right now since I can just use this branch for the frontend and get the interaction I need.

genehynson: Ok! I should have this refactor done today or tomorrow hopefully. So we should be able to get the change in by the end of the week. I'll let you know

And I don't know if I'll keep the swagger generation in notebooksd forever...but for now it makes it easy for me to evolve the api so it's useful

genehynson: added [here](https://github.com/influxdata/idpe/pull/10606/commits/23e31f26f826c282b53d12e718f4d4941e3352ad) for [this PR](https://github.com/influxdata/idpe/pull/10606)

genehynson (approved):

PR #1393, opened 2021-05-11
fix: remove weird types

drdelambre:
master failed after a merge for... this should fix it


drdelambre:
ugh.. a branch bled in there

PR #1407, opened 2021-05-12
fix(annotations): change annotation event structure so cancel metric isn't fired so often

### LINKED ISSUE #1403: Annotations: Create events are also sending cancel events
### Background
When creating a new annotation, both the `create` and the `cancel` events are being sent, leading to inaccurate counts.

### Acceptance Criteria
- create / edit / delete are mutually exclusive events from cancel

### Solution/Fix
previously the cancel event call was in onClose. all buttons have to call onClose, even the submit buttons. so submitting the form was incorrectly calling the cancel event. i moved it out of the onClose handler that is used by all annotation overlays into the actual handler function in on the form that is rendered by the overlay

hoorayimhelping:
Closes #1403

Moves cancel events lower down the stack so that they're not always called when the overlay is closed. They're now only fired when the cancel button is clicked.

### DIFF
--- a/src/annotations/components/EditAnnotationForm.tsx
+++ b/src/annotations/components/EditAnnotationForm.tsx
@@ -9,6 +9,7 @@ import {
   Grid,
   Overlay,
 } from '@influxdata/clockface'
+
 import {AnnotationMessageInput} from 'src/annotations/components/annotationForm/AnnotationMessageInput'
 import {AnnotationStartTimeInput} from 'src/annotations/components/annotationForm/AnnotationStartTimeInput'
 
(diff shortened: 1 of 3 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1407)

jrenee42:
from slack convo with @hoorayimhelping : "previously the cancel event call was in onClose. all buttons have to call onClose, even the submit buttons. so submitting the form was incorrectly calling the cancel event. i moved it out of the onClose handler that is used by all annotation overlays into the actual handler function in on the form that is rendered by the overlay"

jrenee42 (approved):

Issue #1330, opened 2021-04-30
Bump Version & Update Commit for OSS 2.0.7 Release


wbaker85:
Work-around for [influxdb#20007](https://github.com/influxdata/influxdb/issues/20007) & [influxdata-docker#484](https://github.com/influxdata/influxdata-docker/issues/484)

Before we release 2.0.7, update the version & OSS commit hash to be the correct values shown in the UI. Preferably by implementing [ui#334](https://github.com/influxdata/ui/issues/334)

PR #1418, opened 2021-05-14
feat: better version info for oss

### LINKED ISSUE #334: OSS needs a way to update the version displayed on the sign-in page without modifying UI sources
Original issue: influxdata/influxdb#20007. We have a work-around for now, so this is more a future-looking request.

The `VersionInfo` component rendered as part of the OSS sign-in page uses [hard-coded constants](https://github.com/influxdata/ui/blob/master/src/shared/components/VersionInfo.tsx#L16) to set the version / git hash that's displayed to users. These constants are [set at build time based on the environment](https://github.com/influxdata/ui/blob/master/src/shared/constants/index.ts#L62-L63).

Since OSS still contains a copy of the UI code, the git hash of the UI at build time == the git hash of the server at build time, so the UI stays accurate as long as we remember to update `version` inside of `package.json` (see influxdata/influxdb#20077). When we refactor OSS to pull artifacts from / vendor this repository, matching up the git hashes / versions will get even trickier.

IMO it'd be nice if the UI displayed two separate versions:
1. The version of the UI codebase, hard-coded at build time
2. The version of the server as reported by `/health`

wbaker85:
Closes #334 and #1330

Adds some additional version information to the OSS login page and "Getting Started" page that you see right after logging in. This information comes from the OSS backend via the `/health` endpoint. The "Frontend" hash is the one that gets set when building the UI. The hashes for Server and Frontend are clickable links that will take you to the page on github, for example to https://github.com/influxdata/influxdb/tree/4db98b4c9a.

Screen Shot 2021-05-14 at 11 17 07 AM

Screen Shot 2021-05-14 at 11 18 43 AM

### DIFF
--- a/src/shared/components/VersionInfoOSS.tsx
+++ b/src/shared/components/VersionInfoOSS.tsx
@@ -1,0 +1,50 @@
+// Libraries
+import React, {FC, useEffect} from 'react'
+import {useSelector, useDispatch} from 'react-redux'
+
+// Constants
+import {GIT_SHA} from 'src/shared/constants'
+
+// Thunks
+import {fetchVersionInfo} from 'src/shared/thunks/app'
+
+// Selectors
+import {getVersionInfo} from 'src/shared/selectors/app'
+
+const VersionInfoOSS: FC = () => {
+  const dispatch = useDispatch()
+
+  const versionInfo = useSelector(getVersionInfo)
+
+  useEffect(() => {
+    if (!versionInfo.version || !versionInfo.commit) {
+      dispatch(fetchVersionInfo())
+    }
+  })
+
+  return (
+    <>
+      InfluxDB {versionInfo.version}
+      <br />
+      Server:{' '}
+      <code>
+        <a
+          href={`https://github.com/influxdata/influxdb/tree/${versionInfo.commit}`}
+        >
+          {versionInfo.commit && versionInfo.commit.slice(0, 7)}
+        </a>
+      </code>
+      <br />
+      Frontend:{' '}
+      <code>
+        <a
+          href={`https://github.com/influxdata/ui/tree/${GIT_SHA.slice(0, 7)}`}
+        >
+          {GIT_SHA.slice(0, 7)}
+        </a>
+      </code>
+    </>
+  )
+}
+
+export default VersionInfoOSS
--- a/src/shared/thunks/app.ts
+++ b/src/shared/thunks/app.ts
@@ -8,6 +8,9 @@ export const fetchVersionInfo = () => async (
 ): Promise<VersionInfo> => {
   try {
     const url = `${getAPIBasepath()}/health`
+    // Have to use fetch here since the oats client doesn't work with the servers option in the oss swagger
+    // Ex: https://github.com/influxdata/openapi/blob/35d734671d05ba9337dd6fbd8bb5c6085482011b/contracts/oss.yml#L5887
+    // The "/health" endpoint is actually not prefixed by "api/v2", but the oats client has the prefix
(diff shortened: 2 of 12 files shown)


src/shared/components/VersionInfoOSS.tsx
```diff
@@ -0,0 +1,50 @@
...
+// Selectors
+import {getVersionInfo} from 'src/shared/selectors/app'
+
+const VersionInfoOSS: FC = () => {
+  const dispatch = useDispatch()
+
+  const versionInfo = useSelector(getVersionInfo)
+
+  useEffect(() => {
+    if (!versionInfo.version || !versionInfo.commit) {
```
wbaker85, line 20: This feels a bit clunky but it was the easiest way I could come up with to prevent the UI from hitting the `/health` endpoint again from the login page when you get to the getting started page. An alternative might be to dispatch the action in some higher-up container but I didn't want to mess with that for this simple change. Open to suggestions!

src/shared/thunks/app.ts
```diff
@@ -0,0 +1,31 @@
...
+import {getAPIBasepath} from 'src/utils/basepath'
+import {VersionInfo} from 'src/types'
+import {Action, setVersionInfo} from 'src/shared/actions/app'
+
+export const fetchVersionInfo = () => async (
+  dispatch: Dispatch<Action>
+): Promise<VersionInfo> => {
+  try {
+    const url = `${getAPIBasepath()}/health`
+    const response = await fetch(url)
```
wbaker85, line 11: Have to use `fetch` here since the oats client doesn't work with the `servers` option in the oss swagger: https://github.com/influxdata/openapi/blob/35d734671d05ba9337dd6fbd8bb5c6085482011b/contracts/oss.yml#L5887

src/utils/env.ts
```diff
@@ -4,6 +4,7 @@ module.exports = (() => {
     require('child_process')
       .execSync('git rev-parse --sq HEAD')
       .toString()
+      .replace(/^'/, '')
```
wbaker85, line 7: Without this, we were getting a leading `'` in the `GIT_SHA` when building for OSS with `yarn build`. I don't _think_ this is used anywhere else (that I could find)...but I made it a regex replacement so that it doesn't just drop the first character - it only cuts off a leading `'`.

webpack.dev.ts
```diff
@@ -29,6 +29,7 @@ module.exports = merge(common, {
       '/api/v2': 'http://localhost:8086',
       '/debug/flush': 'http://localhost:8086',
       '/oauth': 'http://localhost:8086',
+      '/health': 'http://localhost:8086',
```
wbaker85, line 32: For running local development against OSS with `yarn start`, so it can proxy the request to the backend successfully.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1418)

src/shared/thunks/app.ts
hoorayimhelping: :+1: can you add that comment to the code please?

hoorayimhelping (approved):
looks good to me! I just had one suggestion about adding a comment to the code, but it's not worth re-reviewing for that, I trust you to add the comment correctly. Nice PR, I appreciate the contextual messages :+1:

russorat (approved):
LGTM!

PR #1427, opened 2021-05-17
feat: integrating parts of the checkout page

### LINKED ISSUE #965: Integrate API for Checkout
This issue is to differentiate between the existing UI-based unification work so that we can track progress on the UI side of things. Currently the UI for the checkout page is all set and ready to go for the API. This issue is to make sure we circle back to hooking the UI up to the API once the checkout-related API work is finished

Asalem1:
Related to #965 

This PR doesn't fully close out the issue until Zuora Form functionality and API requirements can be completely tested and integrated.

The lingering issues are:

- Rendering the Zuora Form (this is a taller task to get going from the Quartz side)
- Testing out the `upgrade` and `abandon cart` functionality (this is dependent on the Zuora Form)


src/checkout/context/checkout.tsx
```diff
@@ -210,9 +215,9 @@ export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
 
     const {shouldNotify} = inputs
     if (!shouldNotify) {
+      // TODO(ariel): this seems like a hack to "make it work"
```
Asalem1, line 218: this is a note to self to investigate once the submit functionality is up

src/checkout/context/checkout.tsx
```diff
@@ -271,11 +276,13 @@ export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
...
 
-        const response = await postCheckoutInformation(paymentInformation)
+        const response = await postCheckout({data: paymentInformation})
 
-        handleSetCheckoutStatus(
+        setCheckoutStatus(
           response.status === 201 ? RemoteDataState.Done : RemoteDataState.Error
         )
+        // TODO(ariel): make sure to refetch the `/me` endpoint after a user leaves the checkout
+        // TODO(ariel): where should we redirect the user to once they've upgraded?
```
Asalem1, line 285: same as above, a note to self for when the submit functionality is up

src/checkout/context/checkout.tsx
```diff
@@ -119,33 +119,47 @@ export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
...
   }, [])
 
   useEffect(() => {
     getZuoraParams()
   }, [getZuoraParams])
 
-  const [onSuccessUrl, setOnSuccessUrl] = useState<string>('')
-
   const getBillingSettings = useCallback(
+    // TODO(ariel): this is weird, why do we need a parameter?
```
Asalem1, line 141: this is more of a design question for myself to resolve once we get the submit functionality working

src/checkout/context/checkout.tsx
```diff
@@ -119,33 +119,47 @@ export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
...
+      const response = await getPaymentForm({form: 'checkout'})
+      if (response.status !== 200) {
+        throw new Error(getErrorMessage(response))
+      }
 
-    setZuoraParams(response.data as CreditCardParams)
+      // TODO(ariel): remove type casting here when Billing is refactored and integrated
+      setZuoraParams(response.data as CreditCardParams)
+    } catch (error) {
+      // TODO(ariel): should we notify or report the error?
```
Asalem1, line 131: open question about how we should handle failures to the Zuora form? Right now a white iframe is displayed with the error message,  is that enough? Quartz does the same but can we do better?

src/shared/components/CloudUpgradeButton.tsx
```diff
@@ -49,6 +49,7 @@ const CloudUpgradeButton: FC<StateProps & OwnProps> = ({
     [`${className}`]: className,
   })
 
+  // TODO(ariel): update the link to the local checkout if the flag is on
```
genehynson, line 52: looks like you did this in `CloudUpgradeNavBanner`? Or is that a different component?

src/shared/components/CloudUpgradeNavBanner.tsx
```diff
@@ -49,23 +51,39 @@ const CloudUpgradeNavBanner: FC<StateProps> = ({inView, isRegionBeta}) => (
...
-            <a
-              className="cf-button cf-button-md cf-button-primary cf-button-stretch cloud-upgrade-banner--button upgrade-payg--button__nav"
-              href={`${CLOUD_URL}${CLOUD_CHECKOUT_PATH}`}
-              target="_self"
-            >
-              Upgrade Now
-            </a>
+            {isFlagEnabled('unityCheckout') ? (
+              <Link
+                className="cf-button cf-button-md cf-button-primary cf-button-stretch cloud-upgrade-banner--button upgrade-payg--button__nav"
```
genehynson, line 56: should we create a component for this link so the styles can be defined in one place? Looks like we're duplicating this list of styles in a few places

src/checkout/context/checkout.tsx
```diff
@@ -119,33 +119,47 @@ export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
...
+          ...fields,
+          balanceThreshold: resp.data.balanceThreshold,
+          notifyEmail: resp.data.notifyEmail,
+          shouldNotify: resp.data.isNotify,
+        })
+      } catch (error) {
+        // We're injesting the error here and leaving the default inputs since
+        // Quartz's API returns a 404 for valid requests that don't have any data
+        // Meaning, if a user hasn't filled out the notification settings, the
+        // API response is expected to return a 404 even though the query was successful
```
genehynson, line 158: should we specifically catch 404 errors and only injest those? What if quartz returns a 500 or 403 or something else?

genehynson: also fwiw it seems pretty sketch to return a 404 for a successful response. A 204 seems more appropriate imo

src/shared/components/CloudUpgradeButton.tsx
Asalem1: Different component. This one is dependent on the `LinkButton` component which takes in an `href` rather than a `to` to link to a page. I've got an issue in clockface for it here:

https://github.com/influxdata/clockface/issues/623

src/shared/components/CloudUpgradeNavBanner.tsx
Asalem1: Not sure, where else is this being used outside of the CloudUpgradeNavBanner?

src/shared/components/CloudUpgradeButton.tsx
Asalem1: JK i'm dumb. Closing that issue. Just made it a button

src/checkout/context/checkout.tsx
Asalem1: Right, I spoke to @Delmer814 and @gspletzer about it yesterday and we kind of came up with the fact that it doesn't matter whether the API returns a successful response or not, the defaults should be set based on existing metadata. I agree that it's weird to set the data in a catch statement and that we should actually handle the `404` specifically, but I think in any situation where data doesn't exist we don't want to override the existing defaults

src/shared/components/CloudUpgradeNavBanner.tsx
genehynson: Oh I was just thinking for the usage within this component. But I guess it's really only 2 places that all these styles are defined and 1 of those places will be removed once the `unityCheckout` flag is removed. So it's probably not worth changing

src/checkout/context/checkout.tsx
genehynson: Gotcha, ok that's fair!

genehynson (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1427)

PR #1455, opened 2021-05-19
feat: add rudderstack integration

joelfried:
This adds the integration with Rudderstack on top of our existing reporting API.
There's a new feature flag so that Rudderstack reporting can be used independently of the `appMetrics` flag.

Users are identified for Rudderstack as part of the auth flow and events are reported using the existing metrics reporting API.


ui-ci-bot-influx:
Metric | Score | Level
performance | null | :hankey:
accessibility | null | :hankey:
seo | null | :hankey:
pwa | null | :hankey:
best-practices | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1455)

Asalem1 (changes requested):
lots going on but it looks like this broke the app. Will take a look and give more specific feedback but in the meantime, did you run this locally using KIND and monitor-ci to see whether this works? Also, nit: after the incident we had a few weeks back can we not link to issues that exist in private repos?

src/cloud/utils/rudderstack.ts
```diff
@@ -0,0 +1,60 @@
...
+  rudderstack.load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
+}
+
+export const identify = (
+  userId?: string,
+  traits?: object,
+  options?: object,
+  callback?: Function
+): void => {
+  rudderstack.identify(userId, traits, options, callback)
```
drdelambre, line 18: why wouldn't you just call this function directly?

src/Signin.tsx
```diff
@@ -92,8 +99,20 @@ export class Signin extends PureComponent<Props, State> {
 
   private checkForLogin = async () => {
     try {
-      await client.users.me()
-
+      const user = await client.users.me()
```
drdelambre, line 102: i think this goes here:
https://github.com/influxdata/ui/blob/d44fdb618b2a857ee45995d987af01d276158cef/src/me/actions/thunks/index.ts#L22

src/Signin.tsx
```diff
@@ -31,11 +32,13 @@ import {
...
 
 // Types
-import {RemoteDataState} from 'src/types'
+import {AppState, RemoteDataState} from 'src/types'
+import {getOrg} from 'src/organizations/selectors'
 
 interface State {
   loading: RemoteDataState
   auth: boolean
+  rudderStackInitialized: boolean
```
drdelambre, line 41: instead of this state management, include the initialization in the cloud module, and not as a react component. then it only gets initialized once, when the first piece of code that requires it is included

src/cloud/apis/reporting.ts
```diff
@@ -21,20 +22,28 @@ export interface Points {
...
 export const reportPoints = (points: Points) => {
-  if (!isFlagEnabled('appMetrics')) {
-    return
+  if (isFlagEnabled('rudderstackReporting')) {
+    try {
+      points.points.forEach(point => {
+        track(point.measurement, point.fields, point.tags, () => {})
+      })
+    } catch (e) {
+      // don't want reporting errors to effect user experience
```
Asalem1, line 31: wouldn't it be helpful for us to know if the service falls over? Maybe a honeybadger alert or something?

Asalem1 (changes requested):
Overall it seems like placing this in the `Signin` isn't the place to do it. There are a lot of assumptions here that the user will be logged in. There are bigger implementation details that need to be addressed but the picture here is that as it stands, this will currently break the UI by ingesting the error returned from the `/me` request.

joelfried:
> lots going on but it looks like this broke the app. Will take a look and give more specific feedback but in the meantime, did you run this locally using KIND and monitor-ci to see whether this works? Also, nit: after the incident we had a few weeks back can we not link to issues that exist in private repos?

Yes I ran this locally using KIND and confirmed that when I hard code the constants to the correct values data does indeed flow through to the Rudderstack console.  Happy to ship you screenshots via slack if you want to see (rather than posting them here).

I linked the issue so it would get closed as the PR went in via the CI/CD pipeline.  I'll remove the link now and handle that manually when this gets in.

src/cloud/utils/rudderstack.ts
joelfried: I wanted to leave room to type the options and traits if we wanted to be more specific about the structure of the data we want to ingest.  Since this is at the POC level currently I didn't want to guess yet on how those types should look but thought doing it this way would make that easier later.  But I could take this file out and call the functions directly if you don't think that's worth it long term.

src/shared/constants/index.ts
```diff
@@ -71,6 +71,13 @@ export const API_BASE_PATH = formatConstant(process.env.API_PREFIX)
 export const HONEYBADGER_KEY = formatConstant(process.env.HONEYBADGER_KEY)
 export const HONEYBADGER_ENV = formatConstant(process.env.HONEYBADGER_ENV)
 
+export const RUDDERSTACK_DATA_PLANE_URL = formatConstant(
+  process.env.RUDDERSTACK_DATA_PLANE_URL
+)
+export const RUDDERSTACK_WRITE_KEY = formatConstant(
+  process.env.RUDDERSTACK_WRITE_KEY
```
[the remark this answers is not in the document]
joelfried: Good question! I'll investigate.

src/cloud/utils/rudderstack.ts
```diff
@@ -0,0 +1,60 @@
...
+  callback?: Function
+): void => {
+  rudderstack.page(category, name, properties, options, callback)
+}
+
+export const track = (
+  eventTitle: string,
+  properties?: object,
+  options?: object,
+  callback?: Function
```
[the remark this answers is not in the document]
joelfried: I was using the callback functions in testing and they were very useful at that point but I didn't want to leave tons of logging around.

src/cloud/apis/reporting.ts
joelfried: I followed the pattern below for app-metrics.  But I agree that sounds better and I'll look into adding this.  (Do you think it's worth adding for the appMetrics call below as well at the same time?)

src/cloud/utils/rudderstack.ts
```diff
@@ -0,0 +1,60 @@
+import {
```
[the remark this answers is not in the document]
joelfried: This is everything Rudderstack gives us to work with and I expected that at some point not long after the POC goes in and is validated we'd want to do more with these functions and by having all of this in one place together it'd make the process of building out the next set of functionality easier.  I also thought it nice to keep everything Rudderstack here so that everywhere else can just reference this file.

I can remove this file and directly call the rudderstack API calls from the other place they're used if you'd prefer that approach.  But that was what I was thinking.

src/cloud/utils/rudderstack.ts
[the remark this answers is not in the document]
Asalem1: So then we should remove it since it's just confusing

src/cloud/apis/reporting.ts
Asalem1: Truthfully I have no idea what appMetrics is for so i can't say

Asalem1:
> > lots going on but it looks like this broke the app. Will take a look and give more specific feedback but in the meantime, did you run this locally using KIND and monitor-ci to see whether this works? Also, nit: after the incident we had a few weeks back can we not link to issues that exist in private repos?
> 
> Yes I ran this locally using KIND and confirmed that when I hard code the constants to the correct values data does indeed flow through to the Rudderstack console. Happy to ship you screenshots via slack if you want to see (rather than posting them here).
> 
> I linked the issue so it would get closed as the PR went in via the CI/CD pipeline. I'll remove the link now and handle that manually when this gets in.

When you say you ran it in KIND, did you run it with the `make start` command?

src/cloud/utils/rudderstack.ts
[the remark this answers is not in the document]
Asalem1: Let’s remove it for now since we’re only using one thing. No point in optimizing for something that might not happen

src/cloud/utils/rudderstack.ts
joelfried: Shifted to calling all functions directly and if we end up wanting to do otherwise we can always change it back later.

src/Signin.tsx
joelfried: Done

src/Signin.tsx
```diff
@@ -92,8 +99,20 @@ export class Signin extends PureComponent<Props, State> {
 
   private checkForLogin = async () => {
     try {
-      await client.users.me()
-
+      const user = await client.users.me()
+      if (
+        isFlagEnabled('rudderstackReporting') &&
+        !this.state.rudderStackInitialized
```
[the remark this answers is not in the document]
joelfried: Moved to the thunk and removed the need for the state so now in the thunk the if check only depends upon the feature flag.

src/Signin.tsx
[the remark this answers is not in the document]
joelfried: Changed this back with the move to the thunk

src/cloud/utils/rudderstack.ts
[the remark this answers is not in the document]
joelfried: Rewrote the file and removed these functions.  The function that's left is called once when the file is included when the user logs in and hits the getting started page.  That way things are initialized when the user logs in and only once.

src/cloud/utils/rudderstack.ts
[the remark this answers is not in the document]
joelfried: Removed with the shift to calling functions directly.

src/cloud/apis/reporting.ts
```diff
@@ -21,20 +22,28 @@ export interface Points {
 }
 
 export const reportPoints = (points: Points) => {
-  if (!isFlagEnabled('appMetrics')) {
-    return
+  if (isFlagEnabled('rudderstackReporting')) {
+    try {
+      points.points.forEach(point => {
```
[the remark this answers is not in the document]
joelfried: Refactored.

src/Signin.tsx
```diff
@@ -92,8 +99,20 @@ export class Signin extends PureComponent<Props, State> {
 
   private checkForLogin = async () => {
     try {
-      await client.users.me()
-
+      const user = await client.users.me()
+      if (
```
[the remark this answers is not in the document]
joelfried: Shouldn't happen anymore with the new approach :)

src/cloud/utils/rudderstack.ts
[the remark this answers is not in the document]
Asalem1: I'm a little confused by this file. Why are we initializing the value in the file behind a feature flag and conditionally importing the file in the GettingStarted page? The Getting Started page isn't guaranteed to ever be hit by the user. If I get deep-linked to a particular page (i.e. a dashboard or cell), then I won't be directed to the Getting Started page unless I make an active effort to (let's face it, there's nothing really of interest on that page to get people there)

src/me/components/GettingStarted.tsx
```diff
@@ -22,9 +22,16 @@ import {AppState} from 'src/types'
...
 
+// Utils
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = RouteComponentProps & ReduxProps
 
+if (isFlagEnabled('rudderstackReporting')) {
+  require('src/cloud/utils/rudderstack')
+}
```
Asalem1, line 33: this relates to the previous comment but basically we aren't guaranteed to ever hit this page

Asalem1 (changes requested):

src/cloud/utils/rudderstack.ts
```diff
@@ -0,0 +1,15 @@
...
+import {ready, load} from 'rudder-sdk-js'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
+function initializeRudderstack() {
+  ready(() => {})
+  load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
+}
+
+if (isFlagEnabled('rudderstackReporting')) {
+  initializeRudderstack()
```
Asalem1, line 14: If the goal of this is to only call it once, we should probably plop it onto a `useEffect` in a TSX file to initialize it when the top level component loads up and that's it, especially if we want this to load regardless of whether the user is logged in or not

Asalem1 (changes requested):

src/cloud/utils/rudderstack.ts
joelfried: Is there a particular TSX file we are guaranteed to hit?  I picked this page as it's where logging in redirects you to by default as I presumed that was the normal user flow.

Asalem1: the login page redirects you there if you haven't been given a link to be redirected to. I'd say top level of the app would be the TSX file that we're guaranteed to hit `App.tsx`

joelfried: Moved!

Asalem1: Let’s sync in Monday

src/shared/constants/index.ts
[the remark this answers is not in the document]
joelfried: Added them to that dockerfile.

src/me/actions/thunks/index.ts
```diff
@@ -35,6 +36,10 @@ export const getMe = () => async dispatch => {
       user_id: user.id,
     })
 
+    if (isFlagEnabled('rudderStackReporting')) {
+      identify(user.id, {email: user.name, orgID: this.props.orgID})
```
Asalem1, line 40: where is orgID coming from?

Asalem1 (changes requested):
Looks like you're going to need to get the orgID in order for this to work, but besides that it looks good to me

src/me/actions/thunks/index.ts
Asalem1: Also, should we stick with checking this only in CLOUD?

src/cloud/apis/reporting.ts
```diff
@@ -21,20 +22,28 @@ export interface Points {
...
+        track(point.measurement, point.fields, point.tags, () => {})
```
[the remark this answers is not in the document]
joelfried: Removed.

src/cloud/apis/reporting.ts
joelfried: Added a Honeybadger report on failure here - if it gets noisy or problematic we can remove it later but that's better than not knowing.  I left the appMetrics error being swallowed.

src/me/actions/thunks/index.ts
joelfried: Updated to get orgID properly and added a check for CLOUD.

src/me/actions/thunks/index.ts
```diff
@@ -1,22 +1,30 @@
...
 import {setMe, setQuartzMe, setQuartzMeStatus} from 'src/me/actions/creators'
 
 // Reducers
 import {MeState} from 'src/me/reducers'
 
 // Types
-import {RemoteDataState} from 'src/types'
+import {RemoteDataState, GetState} from 'src/types'
+import {Dispatch} from 'react'
+import {Actions} from 'src/me/actions/creators'
```
Asalem1, line 22: nit: let's try and keep the imports organized (so `react` would be under the Libraries), `src/me/actions/creators` might be in it's own section `// Creators`

Asalem1 (approved):

src/me/actions/thunks/index.ts
joelfried: Moved

Asalem1 (approved):

PR #1468, opened 2021-05-20
build: updates to allow ui to build and run in ci

wbaker85:
Updates to the `OSS-2.0` branch to allow the UI to be built from that branch.

This required updating the `generate` command to use the swagger from the `openapi` repo instead of the `influxdb` repo (it is no longer there). I also updated the version number in the `package.json` while I was in there.

For the circle config, I copied the latest one from `master` and removed the e2e test job. Getting the e2e testing working from this branch with OSS would take a significant amount of effort, and is not worth it right now since we _are_ running the e2e tests on the OSS side, albeit with the legacy OSS tests. We should only be using this branch for one more patch release, and `OSS-2.1` will be much easier to get the e2e tests working since it will have all the latest dockerfile and pipeline updates.

In order to merge this, we will need to update the branch rules to only require the `unit`, `lint`, and semantic checks.


danxmoran (approved):
:+1: to the incremental simplification, looking forward to being even simpler in the future!

PR #1472, opened 2021-05-20
fix: re-enable feature flags for dashboard query cache

wbaker85:
Closes [influxdb#21447](https://github.com/influxdata/influxdb/issues/21447)

The dashboard autorefresh stopped working with for OSS #146, which removed the feature flags around the dashboard query cache. Since this version of the UI was not used with OSS until much more recently, the problem went undiscovered until now.

Since we don't want to bring in the new dashboard autorefresh feature with the next OSS patch release and still want to restore functionality of the dashboard autorefresh, this PR effectively reverts #146 and adds the feature flag back in, so that the query cache is not enabled for OSS.

### DIFF
--- a/src/dashboards/components/DashboardHeader.tsx
+++ b/src/dashboards/components/DashboardHeader.tsx
@@ -30,6 +30,7 @@ import {
 // Utils
 import {event} from 'src/cloud/utils/reporting'
 import {resetQueryCache} from 'src/shared/apis/queryCache'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Selectors
 import {getTimeRange} from 'src/dashboards/selectors'
@@ -107,7 +108,9 @@ const DashboardHeader: FC<Props> = ({
   }
 
   const handleChooseTimeRange = (timeRange: TimeRange) => {
-    resetQueryCache()
+    if (isFlagEnabled('queryCacheForDashboards')) {
+      resetQueryCache()
+    }
     setDashboardTimeRange(dashboard.id, timeRange)
     updateQueryParams({
       lower: timeRange.lower,
@@ -131,7 +134,9 @@ const DashboardHeader: FC<Props> = ({
 
   const resetCacheAndRefresh = (): void => {
     // We want to invalidate the existing cache when a user manually refreshes the dashboard
-    resetQueryCache()
+    if (isFlagEnabled('queryCacheForDashboards')) {
+      resetQueryCache()
+    }
     onManualRefresh()
   }
 
--- a/src/dashboards/components/DashboardPage.tsx
+++ b/src/dashboards/components/DashboardPage.tsx
@@ -59,7 +59,9 @@ const dashRoute = `/${ORGS}/${ORG_ID}/${DASHBOARDS}/${DASHBOARD_ID}`
 @ErrorHandling
 class DashboardPage extends Component<Props> {
   public componentDidMount() {
-    resetQueryCache()
+    if (isFlagEnabled('queryCacheForDashboards')) {
+      resetQueryCache()
+    }
 
     this.emitRenderCycleEvent()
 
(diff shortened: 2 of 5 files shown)


Asalem1 (changes requested):
This has been live for six months with the issue only recently being discovered. The feature flag has been deleted from config cat and IDPE, is there a PR to set those up as well? It sounds like the autorefresh work that @ShmuelLotman has done on the UI resolves some of the issues that we saw around this. Maybe it's worth pursuing that avenue of fixing through some of the existing feature flagged work (by unfeature flagging it) rather than moving this back?

wbaker85:
I definitely agree that we wouldn't want to put the flags back into the master branch of UI or deploy it with anything other than the next patch release of OSS (which will be 2.0.7). And also if we unflag the new autorefresh feature, it does seem to work as intended! But that new autorefresh feature brings in some additional functionality that we don't want to include with a patch release.

Ideally for the next OSS patch release we'd like the dashboard autorefresh to work like it used to (that is, get new data when it automatically refreshes), without bringing in anything else new. Reverting the feature flag removal was the quickest and easiest way I could see that happening for the `OSS-2.0` branch. This code is only used to bundle with OSS and we can set the OSS backend to make the flag `false`, so I hadn't planned on formally re-establishing the feature flag in IDPE or config cat.

We're planning on creating a new branch based on the latest UI in the `master` after the next patch release for `OSS-2.1`, in which we'll be able to include new features. So after that, the code in this `OSS-2.0` branch won't be used again. If there's an easy way to make the dashboard autorefresh work that doesn't involve taking the dashboard query cache out of the picture and also doesn't bring in any new features I'd be all for that, but nothing jumped out at me...although I'm sure the answer is somewhere in #1226.

Anyway, it seemed to me like reverting this part of the code to a previously working state was preferable to investing more time & energy into this branch that is nearing the end of its journey.

Not sure if that helps clarify where I'm coming from with this! Let me know if I'm way off base about anything.

Asalem1:
> I definitely agree that we wouldn't want to put the flags back into the master branch of UI or deploy it with anything other than the next patch release of OSS (which will be 2.0.7). And also if we unflag the new autorefresh feature, it does seem to work as intended! But that new autorefresh feature brings in some additional functionality that we don't want to include with a patch release.
> 
> Ideally for the next OSS patch release we'd like the dashboard autorefresh to work like it used to (that is, get new data when it automatically refreshes), without bringing in anything else new. Reverting the feature flag removal was the quickest and easiest way I could see that happening for the `OSS-2.0` branch. This code is only used to bundle with OSS and we can set the OSS backend to make the flag `false`, so I hadn't planned on formally re-establishing the feature flag in IDPE or config cat.
> 
> We're planning on creating a new branch based on the latest UI in the `master` after the next patch release for `OSS-2.1`, in which we'll be able to include new features. So after that, the code in this `OSS-2.0` branch won't be used again. If there's an easy way to make the dashboard autorefresh work that doesn't involve taking the dashboard query cache out of the picture and also doesn't bring in any new features I'd be all for that, but nothing jumped out at me...although I'm sure the answer is somewhere in #1226.
> 
> Anyway, it seemed to me like reverting this part of the code to a previously working state was preferable to investing more time & energy into this branch that is nearing the end of its journey.
> 
> Not sure if that helps clarify where I'm coming from with this! Let me know if I'm way off base about anything.

Gotcha, I think I misunderstood the purpose of this. So is the intention of this branch to be used for the release of a new UI image for OSS and will be short lived, or is the intention to merge this change into the UI?

wbaker85:
> Gotcha, I think I misunderstood the purpose of this. So is the intention of this branch to be used for the release of a new UI image for OSS and will be short lived, or is the intention to merge this change into the UI?

Yep this branch is only going to be used for the release of a new OSS UI (for a patch release) and it will be short lived. Not planning on merging this into the "actual" UI in the master branch, so it'll be completely isolated in that sense.

Asalem1 (approved):

PR #1480, opened 2021-05-20
chore: address autorefresh feedback

### LINKED ISSUE #1452: Dashboard Auto Refresh - Implement UX/UI Feedback
Addresses: https://github.com/influxdata/ui/issues/1413

After putting out the dashboard auto refresh feature in tools, we received some great feedback and commentary from some users, as well as from @barbaranelson. This, coupled with design's ideas about optional implementation, will require a few changes to the appearance of the feature. We will track those here. 

- as a user, i would like auto refresh to pause if the dashboard page is not visible on the screen (hidden tab, window covered, browser minimized).
- as a user, i would like to view the current auto refresh settings when auto refresh is enabled.
- as a user, i would like to edit the current auto refresh settings when auto refresh is enabled.
- as a user, i would like the ability to manually refresh if needed while auto refresh is enabled.
- as a user, i would like more granular options for inactivity timeout.
- as a user, i would like more options for auto-refresh rates
- as a user, i would like to view the entire end time in the modal
- as a user, i would like the refresh frequency to default to 1m when the modal is opened.

Once these are addressed, this feature should be ready to go in Production, as long as no other issues are discovered.

ShmuelLotman:
Closes #1452 

After putting out the dashboard auto refresh feature in tools, we received some great feedback and commentary from some users, as well as from @barbaranelson. This, coupled with design's ideas about optional implementation, will require a few changes to the appearance of the feature. We will track those here.

as a user, i would like auto refresh to pause if the dashboard page is not visible on the screen (hidden tab, window covered, browser minimized).
as a user, i would like to view the current auto refresh settings when auto refresh is enabled.
as a user, i would like to edit the current auto refresh settings when auto refresh is enabled.
as a user, i would like the ability to manually refresh if needed while auto refresh is enabled.
as a user, i would like more options for auto-refresh rates
as a user, i would like to view the entire end time in the modal
as a user, i would like the refresh frequency to default to 1m when the modal is opened.
Once these are addressed, this feature should be ready to go in Production, as long as no other issues are discovered.


src/client/annotationdRoutes.ts
```diff
@@ -0,0 +1,739 @@
+// This file is generated by [oats][0] and should not be edited by hand.
```
drdelambre, line 1: add this file to the .gitignore and remove it

src/dashboards/components/AutoRefresh.scss
```diff
@@ -55,9 +55,20 @@
...
 }
 
 .refresh-form-timerange-toggle {
   margin: 0px $cf-marg-b;
 }
+
+.refresh-input {
+  width: 100px;
+  input {
+    border-radius: 0px 5px 5px 0px;
```
drdelambre, line 68: use the clockface cf-radius or whatever it is

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1480)

src/dashboards/components/AutoRefresh.scss
Asalem1: ^

src/dashboards/components/AutoRefreshInput.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const dispatch = useDispatch()
+  const currentDashboardId = useSelector(getCurrentDashboardId)
+
+  const autoRefresh = useSelector(getAutoRefreshForDashboard)
+  const handleChooseAutoRefresh = (selection: string) => {
+    if (selection === 'None') {
+      dispatch(resetDashboardAutoRefresh(currentDashboardId))
+      return
+    }
+    const unit = selection[selection.length - 1]
```
Asalem1, line 47: I have no doubt this work, but it's a little confusing for me since it's making me think that `selection` is an `object` or an `array` based on the method used. 

nit: `string.charAt(selection.length - 1)`

src/dashboards/components/AutoRefreshInput.tsx
```diff
@@ -0,0 +1,100 @@
...
+    } else if (unit === 'm') {
+      multiplier = 60
+    } else if (unit === 'h') {
+      multiplier = 3600
+    }
+
+    const calculatedMilliseconds =
+      Number(selection.slice(0, selection.length - 1)) * multiplier * 1000
+
+    if (unit === 's' || unit === 'm' || unit === 'h') {
```
Asalem1, line 60: based on these conditions, it seems like we're expecting the `unit` to be `s`, `m`, or `h` otherwise `multiplier` would be undefined. My question is:

- will unit ever be anything other than `s` `m` or `h`?
- if it won't be, do we still need this check here?
- if it will be, couldn't we just check to see if `typeof multiplier === 'undefined'`?

src/dashboards/components/AutoRefreshInput.tsx
```diff
@@ -0,0 +1,100 @@
...
+        onSubmit={handleChooseAutoRefresh}
+        suggestions={SUGGESTIONS}
+        customClass="refresh-input"
+        validFunction={isValidInput}
+        dividerText="Customize"
+        dividerOnClick={() => {
+          dispatch(
+            showOverlay('toggle-auto-refresh', null, () =>
+              dispatch(dismissOverlay())
+            )
```
Asalem1, line 91: I don't have a good mental model right now of what's happening so sorry if I'm misunderstanding this, but it seems like at this point, the user has opened the overlay right? Why can't we just close it by dispatching the `dismissOverlay`?

src/shared/components/dropdown_auto_refresh/AutoRefreshDropdown.tsx
```diff
@@ -22,172 +22,142 @@ import {AutoRefresh, AutoRefreshStatus} from 'src/types'
...
 
-@ErrorHandling
-export default class AutoRefreshDropdown extends Component<Props> {
-  public static defaultProps = {
-    showAutoRefresh: true,
-    selected: {
-      interval: 0,
-      status: 'paused',
-    },
+const AutoRefreshDropdown: FC<Props> = ({
```
Asalem1, line 34: do we not want to keep error handling?

src/shared/components/dropdown_auto_refresh/AutoRefreshDropdown.tsx
```diff
@@ -22,172 +22,142 @@ import {AutoRefresh, AutoRefreshStatus} from 'src/types'
...
-                    selected={option.id === this.selectedOptionID}
-                    onClick={this.handleDropdownChange}
-                  >
-                    {option.label}
-                  </Dropdown.Item>
-                )
-              })}
-            </Dropdown.Menu>
-          )}
+      <div className={setClassName()}>
```
Asalem1, line 66: why not just inject the logic inline here, or declare a constant up above that does the same thing rather than calling a function? Derived state rather than an output of a function

src/dashboards/components/AutoRefreshInput.tsx
ShmuelLotman: if the multiplier is undefined, we dont want to submit, so its just a check to make sure we are using the right units. The only units we want to currently allow are seconds, minutes, or hours

src/dashboards/components/AutoRefreshInput.tsx
ShmuelLotman: this is just the function passed through as the "onClose" handler, and follows a pattern we use elsewhere

src/dashboards/components/AutoRefreshInput.tsx
Asalem1: gotcha. Might be clearer to say if the multiplier is undefined, return? Either way is good

src/dashboards/components/AutoRefreshInput.tsx
```diff
@@ -0,0 +1,100 @@
...
+    if (unit === 's') {
+      multiplier = 1
+    } else if (unit === 'm') {
+      multiplier = 60
+    } else if (unit === 'h') {
+      multiplier = 3600
+    }
+
+    const calculatedMilliseconds =
+      Number(selection.slice(0, selection.length - 1)) * multiplier * 1000
```
Asalem1, line 58: what if the `multiplier` is undefined

ShmuelLotman: Great catch, thanks, that should be inside of the if

ShmuelLotman: that would be fine if the only thing we needed to know is that its defined, but its not - we also need to know that it is one of the three units i listed

Asalem1 (approved):

Asalem1 (approved):
Quick question - OSS has the autorefresh functionality in the data explorer, can we validate that this:

1. Doesn't appear in Cloud on the data explorer when the feature is enabled
2. Doesn't mess with OSS's autorefresh in the data explorer?

ShmuelLotman:
@Asalem1 1. it does not, all good on that front
2. Can confirm, does not mess w/ OSS as far as i can tell

PR #1489, opened 2021-05-21
feat: notify uers that annotations are unsupported on that graph type

### LINKED ISSUE #1431: Annotations: I would like a notification when i try to add an annotation on a graph type that isn't supported
Given i am on a dashboard and i have 1 click annotation enabled
when i click on a graph type that is not supported for annotations
then i get an alert telling me this graph type isn't supported for annotations.

Notes:
- once this is added, we should remove the static text from the dashboard page about supported types
- the text of the notification should be "This graph type does not support adding annotations."
- the notice should be a warning, and stick around for a few seconds, enough time to read the text, similar to deleting an asset

hoorayimhelping:
Closes #1431

#### Note:
Doesn't work with gauge.

Screen Shot 2021-05-21 at 10 41 42 AM

https://user-images.githubusercontent.com/146112/119177478-39b85180-ba21-11eb-9eac-9f4585a5d038.mov

### DIFF
--- a/src/shared/copy/notifications.ts
+++ b/src/shared/copy/notifications.ts
@@ -1205,10 +1205,12 @@ export const createAnnotationFailed = (error: string): Notification => ({
   message: `Failed to create annotation: ${error}`,
 })
 
-export const annotationsUnsupportedOnGraph = (graphType: string = 'This graph type'): Notification => ({
+export const annotationsUnsupportedOnGraph = (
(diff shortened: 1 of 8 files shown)


hoorayimhelping:
Dashboard I used for testing: https://gist.github.com/hoorayimhelping/4b9f4eddbeb33e3721720f9b44df2e3b

TCL735 (approved):

PR #1510, opened 2021-05-25
feat: add ranged annotations to xy graph and graph + singlestat graph

### LINKED ISSUE #1417: add ranged annotations to graph and graph+ single stat
add ability to add ranged annotations 

1.  change button text from 'enable 1-click annotations'  to 'annotation write mode'
     a.  change this to 'on' by default
2.  when the toggle button is on; can add point annotations or ranged annotations
3.  when the toggle button is off, can zoom into the graph (cannot do both)
4. DON"T ADD:  just snap to times for now.   add a checkbox (or toggle?) to the annotation bar saying if the annotations snap to the nearest times available (for both ranged and point)
      a.  if off, use exact times that are clicked on the graph
      b.  if on, use the nearest time signatures
      c.  this will be on by default
5.  integrate the dragging code with the 'single click' code, since the way the dragging code is written, they need to be integrated (since we wrote our own, it interferes with the 'onClick' listener.  currently, when you zoom; after you finish the zoom the 'add annotation' popup occurs if the 'enable 1-click annotations' is enabled; which is NOT behavior we want
6.  add an 'end' field to the create/edit annotations dialog/overlay
7.  make sure this works on both graph and graph + single stat
8.  make sure editing and deleting works for ranged annotations

>>>>>>>>>>
already have most of #5 done.

jrenee42:
Closes #1417 

adds ranged annotations to graph (xy) and graph + singlestat

allows for editing and deleting of those annotations too

### DIFF
--- a/src/annotations/components/EditAnnotationForm.tsx
+++ b/src/annotations/components/EditAnnotationForm.tsx
@@ -11,7 +11,7 @@ import {
 } from '@influxdata/clockface'
 
 import {AnnotationMessageInput} from 'src/annotations/components/annotationForm/AnnotationMessageInput'
-import {AnnotationStartTimeInput} from 'src/annotations/components/annotationForm/AnnotationStartTimeInput'
+import {AnnotationTimeInput} from 'src/annotations/components/annotationForm/AnnotationTimeInput'
 
 // Constants
 import {ANNOTATION_FORM_WIDTH} from 'src/annotations/constants'
@@ -108,10 +108,11 @@ export const EditAnnotationForm: FC<Props> = (props: Props) => {
       />
       <Grid className="edit-annotation-grid">
         <Grid.Column widthSM={Columns.Twelve} widthXS={Columns.Twelve}>
-          <AnnotationStartTimeInput
+          <AnnotationTimeInput
             onChange={updateStartTime}
             onSubmit={handleKeyboardSubmit}
-            startTime={editedAnnotation.startTime}
+            time={editedAnnotation.startTime}
+            name="startTime"
           />
           <AnnotationMessageInput
             message={editedAnnotation.summary}
--- a/src/annotations/components/EditAnnotationForm.tsx
+++ b/src/annotations/components/EditAnnotationForm.tsx
@@ -101,6 +101,7 @@ export const EditAnnotationForm: FC<Props> = (props: Props) => {
     props.handleClose()
   }
 
+  // TODO:  get the correct prefix in there, multiple plot types have annotations now
   const handleDelete = () => {
     try {
       dispatch(deleteAnnotations(editedAnnotation))
@@ -121,7 +122,7 @@ export const EditAnnotationForm: FC<Props> = (props: Props) => {
         onSubmit={handleKeyboardSubmit}
         time={editedAnnotation.endTime}
         name="endTime"
-        titleText="Stop Time (UTC)"
+        titleText="Stop Time"
       />
     )
   }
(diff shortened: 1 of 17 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1510)

src/annotations/components/EditAnnotationForm.tsx
```diff
@@ -46,32 +47,45 @@ export const EditAnnotationForm: FC<Props> = (props: Props) => {
...
+    return checkAnnotationFormValidity(
+      editedAnnotation.type,
+      editedAnnotation.summary,
+      editedAnnotation.startTime,
+      editedAnnotation.endTime
+    )
   }
 
-  const updateStartTime = (newStartTime: string) => {
+  const doUpdate = (obj: any) => {
```
hoorayimhelping, line 70: something about this being called `doUpdate` feels weird. Usually having to name a method `doSomething` because `something`  is already used is a smell - maybe something is too abstract because we already have `updateAnnotation`? Or maybe it means this should be called `updateAnnotation` and the other one should be renamed to `updateAnnotationState`?

And calling an argument `obj` actually mildly offends me when a senior engineer does it. Come on, it's pretty obviously the annotation we're editing - so can you name it something like that please?

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -29,24 +29,45 @@ type AnnotationType = 'point' | 'range'
...
   onClose: () => void
 }
 
+export const checkAnnotationFormValidity = (
+  annotationType: string,
+  message: string,
+  startTime: any,
+  endTime: any
+) => {
+  const firstPart = message.length && startTime
```
hoorayimhelping, line 45:
```suggestion
  const isValidPointAnnotation = message.length && startTime
```

`firstPart` seems like a stored value, not a boolean. Even reading the line that created the value, it wasn't super clear to me that it's a boolean. You don't have to call it `isValidPointAnnotation`, but that's the concept we're trying to convey, right?

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -29,24 +29,45 @@ type AnnotationType = 'point' | 'range'
...
   onSubmit: (Annotation) => void
   onClose: () => void
 }
 
+export const checkAnnotationFormValidity = (
+  annotationType: string,
+  message: string,
+  startTime: any,
+  endTime: any
+) => {
```
hoorayimhelping, line 44: what does this function return? looks like a boolean? can you type hint that please?

hoorayimhelping (changes requested):
nice work! I have some code suggestions, a couple of take-it-or-leave-it style nits. 
I'm going to pull this down and play around with it locally

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -76,12 +115,14 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
...
-              <AnnotationStartTimeInput
+              <AnnotationTimeInput
                 onChange={updateStartTime}
                 onSubmit={handleKeyboardSubmit}
-                startTime={startTime}
+                time={startTime}
+                name="startTime"
               />
             </Grid.Row>
+            {endTimeSection}
```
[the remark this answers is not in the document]
jrenee42: i dislike that convention, it makes the final 'return' where the component is put together hard to read if it is long, but in this case it's just one item; so can inline it.

hoorayimhelping (approved):
so awesome! It's all happening!!!!
Screen Shot 2021-05-26 at 1 17 15 PM

PR #1545, opened 2021-05-28
feat(alerts): warned user of dangers of editing tasks

### LINKED ISSUE #1421: I would like a way to quickly edit the underlying task for a check or notification rule
as a user, i would like the ability to quickly customize the task behind the check.
as a user, i would like the ability to quickly customize the task behind the notification rule.

to make it easier to edit the underlying task of a check or notification rule, we should add an option to "Edit Task" to the drop down list of the Check and the Notification rule that would link the use to the edit tasks page: 

`https://us-west-2-1.aws.cloud2.influxdata.com/orgs/ORG_ID/tasks/TASK_ID/edit`

image

We can also change the icon from an eyeball to a gearbox.
We could also let them know that doing this might remove the ability to edit the task in the UI.

blegesse-git:
Closes #1421 

Changed the eyeball icon to gearbox and added "Edit Task" to the drop down list of both Check and Notification rule. 
editTask

When "Edit Task" is clicked, user is directed to edit task page. A warning toast message pops up letting user know editing code might remove the ability to edit task in the UI. 
Editwarning

### DIFF
--- a/src/shared/copy/notifications.ts
+++ b/src/shared/copy/notifications.ts
@@ -1099,13 +1099,13 @@ export const communityTemplateRenameFailed = (): Notification => ({
 export const editCheckCodeWarning = (): Notification => ({
   ...defaultErrorNotification,
   style: NotificationStyle.Info,
-  message: `Changes to Check code may prevent you from editing the Check in the visual editing experience.`,
+  message: 'Changes to Check code may prevent you from editing the Check in the visual editing experience.',
 })
 
 export const editNotificationRuleCodeWarning = (): Notification => ({
   ...defaultErrorNotification,
   style: NotificationStyle.Info,
-  message: `Changes to Notification Rule code may prevent you from editing the Notification Rule in the visual editing experience.`,
+  message: 'Changes to Notification Rule code may prevent you from editing the Notification Rule in the visual editing experience.',
 })
 
 // Notebooks
--- a/src/shared/copy/notifications.ts
+++ b/src/shared/copy/notifications.ts
@@ -18,6 +18,7 @@ import {
 import {QUICKSTART_SCRAPER_TARGET_URL} from 'src/dataLoaders/constants/pluginConfigs'
 import {QUICKSTART_DASHBOARD_NAME} from 'src/onboarding/constants/index'
 import {IconFont} from '@influxdata/clockface'
+import { info } from 'console'
 
 const bytesFormatter = binaryPrefixFormatter({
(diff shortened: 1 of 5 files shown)


src/shared/copy/notifications.ts
```diff
@@ -1096,6 +1096,18 @@ export const communityTemplateRenameFailed = (): Notification => ({
   message: `We've successfully installed your template but weren't able to name it properly. It may appear as a blank template.`,
 })
 
+export const editCheckCodeWarning = (): Notification => ({
+  ...defaultErrorNotification,
+  style: NotificationStyle.Info,
+  message: `Changes to Check code may prevent you from editing the Check in the visual editing experience.`,
```
hoorayimhelping, line 1102: small thing when writing, but ends up actually saving time when reading and maintaining code:

the backtick
```
`this is my ${variable}`
```

implies it's going to use [string templating](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) - which is the only time you'd use the backtick string character. if you're not doing string templating, just use a single / double quote instead.

src/shared/copy/notifications.ts
```diff
@@ -1096,6 +1096,18 @@ export const communityTemplateRenameFailed = (): Notification => ({
...
+})
+
+export const editNotificationRuleCodeWarning = (): Notification => ({
+  ...defaultErrorNotification,
+  style: NotificationStyle.Info,
+  message: `Changes to Notification Rule code may prevent you from editing the Notification Rule in the visual editing experience.`,
```
hoorayimhelping, line 1108: same thing here about using a single / double quote.

hoorayimhelping (changes requested):
great work! good job keeping the code consistent with the code around it.

one small nit for using backticks with no interpolation

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1545)

src/shared/copy/notifications.ts
blegesse-git: thanks bucky. I will keep this in mind going forward :)

hoorayimhelping (approved):
awesome work!

PR #1552, opened 2021-05-28
chore: updating the OpenAPI SHA to test out the getUsageStats API

Asalem1:
The primary purpose of this PR is to update the SHA for the OpanAPI reference. Unfortunately, we can't validate whether the change made to OpenAPI will fix the issue with getting the billing_stats since that depends on querying an external bucket that is inaccessible in local development (but is available in staging and production).

Wit the SHA being updated came some changes to some of the type definitions in the API for an account as it relates to the Account type. The current Account type was being used as a nested property of the Operator and contains more information than is necessary for most Account use cases and will therefore be partitioned for the Operator page.

Finally, there was some general house keeping with this PR (bumping the minimum balance threshold from 1 -> 10, adding notes around some feature flags that SHOULD NOT BE REMOVED from the UI) and removing some unused types


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1552)

genehynson (approved):

PR #1594, opened 2021-06-02
fix: add default so colorschemedropdown doesnt error if no color

### LINKED ISSUE #1445: TypeError: Cannot read property 'filter' of null
I saw this after importing the https://github.com/influxdata/community-templates/blob/master/influxdb2_oss_metrics/influxdb2_oss_metrics.yml template to the ui. Click on the edit panel for query requests and you should see the error.

Looks like sometimes `properties.color` doesn't exist: https://github.com/influxdata/ui/blob/master/src/visualization/types/Graph/options.tsx#L195

image

ShmuelLotman:
Closes #1445

Adds a default array to the colorschemedropdown value prop so that if there is not color defined, it will not error.

### DIFF
--- a/src/visualization/types/Graph/options.tsx
+++ b/src/visualization/types/Graph/options.tsx
@@ -193,7 +193,7 @@ const GraphViewOptions: FC<Props> = ({properties, results, update}) => {
           )}
           <Form.Element label="Line Colors">
             <ColorSchemeDropdown
-              value={properties.colors.filter(c => c.type === 'scale')}
+              value={properties.colors?.filter(c => c.type === 'scale') ?? []}
               onChange={colors => {
                 update({colors})
               }}


src/visualization/types/Graph/options.tsx
```diff
@@ -193,7 +193,7 @@ const GraphViewOptions: FC<Props> = ({properties, results, update}) => {
           )}
           <Form.Element label="Line Colors">
             <ColorSchemeDropdown
-              value={properties.colors.filter(c => c.type === 'scale')}
+              value={properties.colors?.filter(c => c.type === 'scale') ?? []}
```
Asalem1, line 196: should the colors have a default array?

ShmuelLotman: I'm actually not sure... it does have a default array, that is in the properties, but then when passed into the options, its null. My options as i saw them were to either: 
1. Pass the defaults here again
2. add a default array, which would enable the colors dropdown to work again,  and it seemed to have the defaults available again when i did this, so this is the route i took.

I'm trying to figure out why the default colors, which are _not_ null when they are passed into the register call in the index file, but _are_ null when destructured in the options file, are null.... if you have any thoughts, im all ears, but ill keep looking

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1594)

drdelambre (approved):

src/visualization/types/Graph/options.tsx
drdelambre: i looked at this with shmuel. if a cell is created within the frontend, we totally give it a default color array and everything is gravy. apparently the api contract (or just through stacks, idk if that's a distinction worth making) allows for cells to be created without assigning a color, like an optional param or something, so this condition shows up when making dashboards from a community template or whatever. The fallback when opening it with this fix is that the frontend goes and sets a default color for it. As those are owned by the community and not by us (so we can't change it), and the contract supports it, we should just account for it.

might be worth looking into why this didn't show up as a typescript error from openapi

Asalem1 (approved):

PR #1606, opened 2021-06-03
fix: dashboard header when removing name is back to default

ShmuelLotman:
Closes #18998

Makes the fallback option for dashboard title "Name this Dashboard" so it matches the default name

### DIFF
--- a/src/dashboards/components/DashboardPage.tsx
+++ b/src/dashboards/components/DashboardPage.tsx
@@ -114,7 +114,8 @@ class DashboardPage extends Component<Props> {
 
   private get pageTitle(): string {
     const {dashboard} = this.props
-    const title = dashboard && dashboard.name ? dashboard.name : 'Loading...'
+    const title =
+      dashboard && dashboard.name ? dashboard.name : 'Name this Dashboard'
 
     return pageTitleSuffixer([title])
   }


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1606)

src/dashboards/components/DashboardPage.tsx
```diff
@@ -114,7 +114,8 @@ class DashboardPage extends Component<Props> {
 
   private get pageTitle(): string {
     const {dashboard} = this.props
-    const title = dashboard && dashboard.name ? dashboard.name : 'Loading...'
+    const title =
+      dashboard && dashboard.name ? dashboard.name : 'Name this Dashboard'
```
Asalem1, line 118:
```suggestion
      dashboard?.name ?? 'Name this Dashboard'
```

hoorayimhelping: Just a heads up, this will change the meaning of this expression and I'm not sure we want that. Currently, if `dashboard.name` is falsy, like a `0` or `''`, it'll revert to `Name this Dashboard`. If we use the nullish coalescing, it'll only fail on `undefined` or `null`.

```js
const dashboard = {name: ''}
const title = dashboard && dashboard.name ? dashboard.name : 'Name this Dashboard'
// title: Name this Dashboard
```

```js
const dashboard = {name: ''}
const title = dashboard?.name ?? 'Name this Dashboard'
// title:""
```

hoorayimhelping (approved):
lgtm, but you might want to make sure ari is okay with this since he requested changes.

src/dashboards/components/DashboardPage.tsx
Asalem1: If a user explicitly sets the dashboard name to an empty string, shouldn't that be allowed?

PR #1636, opened 2021-06-04
fix: simple table no longer vanishes

### LINKED ISSUE #1548: Simple Table fails to refresh if you were on a table page that no longer exists.
Thank you for reporting a bug in InfluxData UI.

**Steps to reproduce:**

1. Run any query and use the raw data simple table view
2. Click though to a later page beyond 1
3. re-run the query in a way that the results are shorter (e.g. `limit(n:1)`) 

**Expected behavior:**
the table refreshes

**Actual behavior:**
the table disappears.  Anything that forces the panel to redraw (page refresh, collapsing the panel and reopening it) will make it show up.  If you are on page 1 when you re-run this bug does not occur.  

**Environment info:**
Cloud2 internal instance

genehynson:
Closes #1548

Adds `useEffect` that sets the page to `1` every time `result` changes. This fixes the unwanted behavior where the table would vanish when you switch from a dataset with more pages to one with fewer. I also renamed the variable `page` to `size`.

Test included.

https://user-images.githubusercontent.com/6411855/120874009-732ca900-c559-11eb-8e45-857936320c38.mov

### DIFF
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -1592,4 +1592,92 @@ describe('DataExplorer', () => {
       // TODO: add filter values based on dropdown selection in key / value
     })
   })
+
+  describe('simple table interactions', () => {
+    const simpleSmall = 'simple-small'
+    const simpleLarge = 'simple-large'
+    beforeEach(() => {
+      cy.window().then(win => {
+        win.influx.set('simpleTable', true)
+      })
+
+      cy.get('@org').then(({id: orgID}: Organization) => {
+        cy.createDashboard(orgID).then(({body}) => {
+          cy.getByTestID('tree-nav')
+          cy.createBucket(orgID, name, simpleLarge)
+          cy.writeData(lines(300), simpleLarge)
+          cy.createBucket(orgID, name, simpleSmall)
+          cy.writeData(lines(30), simpleSmall)
+          cy.reload()
+        })
+      })
+    })
+
+    it('should render correctly after switching from a dataset with more pages to one with fewer', () => {
+      cy.getByTestID('query-builder').should('exist')
+
+      // show raw data view of data with 100 pages
+      cy.getByTestID(`selector-list ${simpleLarge}`).should('be.visible')
+      cy.getByTestID(`selector-list ${simpleLarge}`).click()
+
+      cy.getByTestID('selector-list m').should('be.visible')
+      cy.getByTestID('selector-list m').clickAttached()
+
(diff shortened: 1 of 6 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1636)

cypress/e2e/shared/explorer.test.ts
```diff
@@ -1526,4 +1526,90 @@ describe('DataExplorer', () => {
...
+
+      cy.getByTestID('selector-list v').should('be.visible')
+      cy.getByTestID('selector-list v').clickAttached()
+
+      cy.getByTestID('selector-list tv1').clickAttached()
+
+      cy.getByTestID('time-machine-submit-button').click()
+
+      cy.getByTestID('raw-data--toggle').click()
+      cy.getByTestID('table').should('exist')
```
Asalem1, line 1566: Table isn’t quite clear on which table we’re testing. I know the test suite is designed for simple table, but someone might have set the testID for the other tables as just “table”. Might be worth being a little more specific just in case?

cypress/e2e/shared/explorer.test.ts
```diff
@@ -1526,4 +1526,90 @@ describe('DataExplorer', () => {
...
+      cy.getByTestID('selector-list v').should('be.visible')
+      cy.getByTestID('selector-list v').clickAttached()
+
+      cy.getByTestID('selector-list tv1').clickAttached()
+
+      cy.getByTestID('time-machine-submit-button').click()
+
+      // verify table still exists
+      cy.getByTestID('table').should('exist')
+      // TODO: uncomment when this bug in clockface is fixed: https://github.com/influxdata/clockface/pull/635
```
Asalem1, line 1596: should we write an issue to uncomment this with the next clockface bump?

drdelambre (approved):
so much renaming to `setPage(1)`

bump on writing an issue in clockface

cypress/e2e/shared/explorer.test.ts
genehynson: added clockface bump to this PR and uncommented this!

cypress/e2e/shared/explorer.test.ts
genehynson: changed to `simple-table`

Asalem1 (approved):

PR #1640, opened 2021-06-07
feat: clone cell to different dashboard

### LINKED ISSUE #1641: Clone/Move Cells Across Dashboards
The intent of this is to explore the ability to clone cells across dashboards, either by moving them entirely or by copying them. This will enable users to more easily manage their data and not have to manually copy queries to get a cell across dashboards

ShmuelLotman:
Closes #1641

This PR introduces the ability to clone a cell across dashboards, either by moving it entirely (which will delete it from the current dashboard) or by copying it, which will make another cell with the identical query and result, but keep both independent so that editing/deleting one will not affect the other. 

https://user-images.githubusercontent.com/29473622/121039943-21616a00-c777-11eb-8859-8225bbe946b8.mov


src/dashboards/utils/getDashboardIds.ts
```diff
@@ -0,0 +1,19 @@
...
+      orgID: orgID,
+      limit: DASHBOARD_LIMIT,
+    },
+  })
+
+  if (resp.status !== 200) {
+    throw new Error(resp.data.message)
+  }
+
+  return resp.data.dashboards
```
Asalem1, line 18: I think we'd want to normalize them

src/shared/components/cells/CellCloneOverlay.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const [otherDashboards, setOtherDashboards] = useState<Dashboard[]>([])
+  const dispatch = useDispatch()
+  const {id: orgID} = useSelector(getOrg)
+
+  const currentDashboardID = useSelector(
+    (state: AppState) => state.currentDashboard.id
+  )
+
+  useEffect(() => {
+    getDashboardIDs(orgID).then(retrievedBoards => {
```
Asalem1, line 40: I think this might be overkill? We should be able to access the list of dashboards in a selector. Even if we deep link to a specific dashboard, I think we need to load ALL the dashboards, then the cells and views for an individual dashboard? I could be wrong

src/shared/components/cells/CellCloneOverlay.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const selectedDashboard = otherDashboards.find(
+    d => d.id === destinationDashboardID
+  )?.name
+
+  return (
+    <Overlay.Container maxWidth={400}>
+      <Overlay.Header title="Move Cell" onDismiss={onClose} />
+      <Overlay.Body className="cellCloneOverlayBody">
+        <Form.Element label="" className="otherDashboardDropdown">
+          <Dropdown
```
Asalem1, line 84: nit: our dropdowns typically allow for resource creation, might be worth adding that here

Asalem1 (commented):
My biggest hang up here is the creation of a new thunk to get all the dashboards. That should exist elsewhere, but besides the point it seems a bit unnecessary for the purposes of this PR

Asalem1 (approved):
Offlined concerns. Some lingering issues would be catching the error thrown by the utility function and naming the classnames based on BEM conventions

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1640)

TCL735 (approved):

PR #1652, opened 2021-06-08
fix: update example for toRows

russorat:
Closes #

PR #1653, opened 2021-06-08
fix: made docs link to cloud or oss based on env

### LINKED ISSUE #1616: i would like to view docs for OSS when i am using InfluxDB OSS, and for Cloud when i am using InfluxDB Cloud
### Background
as of #700 and #1602 , all the urls in the app point to the cloud docs. that's fine for cloud, but because our UI is also built for OSS, the OSS users are going to the cloud docs by default. Since we have docs specifically for the OSS version, we should find a way to dynamically link the user to those docs vs the cloud ones.

### Acceptance criteria
- as an oss user, when i click on a docs link in the app, then i am redirected to the https://docs.influxdata.com/influxdb/v2.0/ docs
- as a cloud user, when i click on a docs link in the app, then i am redirected to the https://docs.influxdata.com/influxdb/cloud/ docs

wbaker85:
Closes #1616

This PR makes the docs links go to either `.../cloud/...` or `.../v2.0/...` depending on the environment. It uses a new constant `DOCS_URL_VERSION` and string interpolation on the URLs for this.

The constant is declared and exported in the `src/shared/constants/fluxFunctions.ts` file. I would have preferred to have the constant be in the `src/shared/constants/index.ts` file, but then importing it into the `fluxFunctions.ts` file was causing the cypress tests to fail to build.


src/shared/constants/fluxFunctions.ts
```diff
@@ -8012,8 +7672,7 @@ export const FLUX_FUNCTIONS: FluxToolbarFunction[] = [
       'Adds a fixed duration to time columns. The output table schema is the same as the input table.',
     example: 'timeShift(duration: 10h, columns: ["_start", "_stop", "_time"])',
     category: 'Transformations',
-    link:
-      'https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/built-in/transformations/shift/',
+    link: `https://docs.influxdata.com/influxdb/${DOCS_URL_VERSION}/reference/flux/stdlib/built-in/transformations/timeshift/`,
```
wbaker85, line 7675: This link was actually broken before - `timeshift` is the correct URL

Asalem1 (commented):
Just a product question - did we want to change CLOUD references for OSS? @russorat

cypress/e2e/shared/checks.test.ts
```diff
@@ -531,8 +531,8 @@ describe('Checks', () => {
                 cy.request(url)
                   .its('body')
                   .should(
-                    'include',
-                    'https://docs.influxdata.com/influxdb/cloud/monitor-alert/checks/create/'
+                    'match',
+                    /https:\/\/docs\.influxdata\.com\/influxdb\/(cloud||v2.0)\/monitor-alert\/checks\/create\//i
```
wbaker85, line 535: Necessary for the test to work for either cloud or oss

src/shared/constants/fluxFunctions.ts
```diff
@@ -1,5 +1,7 @@
 import {FluxToolbarFunction} from 'src/types/shared'
 
+export const DOCS_URL_VERSION = !!process.env.CLOUD_URL ? 'cloud' : 'v2.0'
```
russorat, line 3: is it possible to fetch this version from somewhere? docs will go to v2.1 and we have to remember to change this value? seems tricky.

russorat: i think you can just use `latest` instead of `v2.0` and it would be fine.

russorat:
> did we want to change CLOUD references for OSS?

If you mean change all `InfluxDB Cloud` references to `InfluxDB OSS` in the flux functions list, then no.

src/shared/constants/fluxFunctions.ts
```diff
@@ -2452,8 +2347,7 @@ export const FLUX_FUNCTIONS: FluxToolbarFunction[] = [
       'Pivots geo-temporal data into row-wise sets based on time and other correlation columns.',
     example: 'geo.toRows(correlationKey: ["_time"])',
```
russorat, line 2348:
```suggestion
    example: 'geo.toRows()',
```

incorporating #1652

src/shared/constants/fluxFunctions.ts
wbaker85: Makes sense to use `latest` - especially on the master UI branch. We might have to think about how we want this to work on release branches eventually, otherwise a 2.1 release will link to the 2.2 docs etc. if it just links to latest. Maybe that's alright though.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1653)

src/shared/constants/fluxFunctions.ts
wbaker85: This is updated now - will use the latest URL on OSS.

russorat (approved):
LGTM

Asalem1 (approved):

PR #1672, opened 2021-06-09
fix: clone cell names

ShmuelLotman:
Closes #

We need to be aware of the names in the dashboard _to which we are cloning_ cells. This will help us establish better naming conventions for the cells that are being transferred


src/shared/components/cells/CellCloneOverlay.tsx
```diff
@@ -41,6 +47,9 @@ const CellCloneOverlay: FC = () => {
     (state: AppState) => state.currentDashboard.id
   )
 
+  const allViews =
+    useSelector((state: AppState) => state.resources.views.byID) ?? {}
```
Asalem1, line 51: This might be overkill / not return the expected result when deep-linking into a dashboard. More on this in a sec

Asalem1: So in the case of tools where we have a TON of dashboards with a ton of cells and views, fetching ALL of the views just to filter that down to a specific dashboard might be a pretty extreme operation considering we are only using this for the selected dashboard's views.

One more thing to think about, if we deep-link into a dashboard, we don't run the `getDashboards` thunk, which will populate the views and cells for all the dashboards. So if you run this operation after navigating directly a dashboard, I'm not sure if you'll manage to get all the views for all the dashboards.  

I could be wrong about the last one, but I think we went over that when we talked about the `getDashboards` API request you were making to originally get the dashboards beforehand

ShmuelLotman: The last point you mentioned is handled below in the submit function - if we dont have the view for that dashboard, we will use the utils func to pull the dashboard from the backend. As to the other point, thats a good one, lol, ill adjust that now

Asalem1: Aw, yup, misread that. Good point

ShmuelLotman: FWIW, i also tested by going directly to a dashboard, such that we do not get other dashboard data in redux, and was able to clone/move successfully

Asalem1: 🥳

Asalem1 (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1672)

PR #1685, opened 2021-06-11
chore: mop up an openapi change that was made without consultation

TCL735:
Mopping up after this change: https://github.com/influxdata/openapi/pull/116

### DIFF
--- a/src/dataLoaders/reducers/telegrafEditor.ts
+++ b/src/dataLoaders/reducers/telegrafEditor.ts
@@ -104,7 +104,7 @@ hostname = ""
 ## If set to true, do no set the "host" tag in the telegraf agent.
   omit_hostname = false
 `,
-  },
+  } as TelegrafEditorBasicPlugin,
   {
     name: '__default__',
     type: 'bundle',


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1685)

Asalem1 (changes requested):
I'm a little reluctant on this one since it's weird that the openAPI change to update the SHA wouldn't have thrown a TS error to prevent this from getting in in the first place. What exactly is the issue here?

Asalem1:
Follow up on the original comment, after some investigating it seems like the UI is expecting specific types that aren't in that openAPI contract. We should either update the yaml so that the contract reflects the UI expectations, or update the UI so that it's based off of the API contract. Either way, we need to make sure that there's one source of truth here for what a telegraf plugin type can be, and go based off of that.

TCL735:
Openapi changes got reverted. No need for this anymore.

PR #1690, opened 2021-06-11
fix: delete notebooks accepts 200 or 204

genehynson:
flowd returned a 200 on successful deletion, notebooksd returns a 204 (which is the correct code to return since we do not return the entity in the response)

This change makes the UI code more flexible so that it accepts both statuses in the 200 range as success. I couldn't change the 200 to a 204 because we're still in the process of deploying notebooksd so the UI code needs to be flexible for now.

### DIFF
--- a/src/flows/context/api.tsx
+++ b/src/flows/context/api.tsx
@@ -71,7 +71,7 @@ export const createAPI = async (flow: PostFlowsOrgsFlowParams) => {
 
 export const deleteAPI = async (ids: DeleteFlowsOrgsFlowParams) => {
   const res = await deleteFlowsOrgsFlow(ids)
-  if (res.status != 200) {
+  if (res.status < 200 || res.status >= 300) {
     throw new Error(res.data.message)
   }
 }


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1690)

Asalem1 (commented):
Is this because of the different APIs? Do delete requests typically return 200 or 204?

genehynson:
> Is this because of the different APIs? Do delete requests typically return 200 or 204?

Yeah exactly, it's because of a difference in the implementation of flowd and notebooksd. The correct response should be 204 (no content) because we don't return any response data on a successful deletion.

genehynson:
Once we switch to notebooksd in production, we can enable the `notebooksApi` flag in the UI and we'll start using the openapi generated routes for notebooksd and this implementation will be used: https://github.com/influxdata/ui/blob/master/src/flows/context/notebooks.api.tsx#L73

So this is just a temporary change while we're in the process of deploying notebooksd

Asalem1 (approved):

PR #1720, opened 2021-06-15
chore: cypress set feature flags

### LINKED ISSUE #1268: tests: figure out a clean way to wait on window.influx object
In Cypress we often have to set feature flags so we can test new features. You might see this sprinkled throughout the test code in places like this: https://github.com/influxdata/ui/blob/f3660c92b544a9886f011196e1e42d2f2684fe0e/cypress/e2e/shared/dashboardsView.test.ts#L21

We set the flag with `win.influx.set('flag name', true)`

The problem is that when Cypress tries to set this before running the tests, the `influx` object in the window is not yet ready. This results in the flag _sometimes_ not being enabled correctly. And thus, a flaky test. 

Our work around has been to add a `cy.wait(1000)` before setting the flag. This seems to be enough time for the `influx` object to be ready and remove all flakiness.

`cy.wait` is ugly and should be removed. There _has_ to be a cleaner way to wait for this window object to be ready. Find it!

genehynson:
Closes #1268 

Adds `setFeatureFlags` command to cypress that uses the redux store to set the feature flag. This removes the need to `cy.wait` on setting feature flags and should help with some flakes.

In order for this to work, I had to expose the store to window via `window.store = storeSingleton`. However, I wrapped this in a conditional so it only happens when running the UI in development mode.


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1720)

.circleci/config.yml
```diff
@@ -134,7 +134,7 @@ jobs:
       - checkout
       - run:
           name: Run monitor-ci tests
-          command: API_KEY=${MONITOR_CI_API_KEY} PULL_REQUEST=${CIRCLE_PULL_REQUEST} MONITOR_CI_BRANCH="master" UI_BRANCH=${CIRCLE_BRANCH} SHA=${CIRCLE_SHA1} ./scripts/run-monitor-ci-tests.bash
+          command: API_KEY=${MONITOR_CI_API_KEY} PULL_REQUEST=${CIRCLE_PULL_REQUEST} MONITOR_CI_BRANCH="genehynson/cypress-image" UI_BRANCH=${CIRCLE_BRANCH} SHA=${CIRCLE_SHA1} ./scripts/run-monitor-ci-tests.bash
```
joelfried, line 137: Was this to test to make sure everything works in the CI pipeline?  Either way I would imagine we'd want this back to master before this gets merged.

joelfried (approved):
:fire:

.circleci/config.yml
genehynson: Correct - I have a PR to monitor-ci that needs to be merged first.

PR #1727, opened 2021-06-16
fix: Task Page/ Task Runs Page- task editing routing issue 

### LINKED ISSUE #1495: Task Runs Page UX Touch Up
### Background
The tasks runs page could be made more usable with very little effort. 

A screenshot from [the figma mockup](https://www.figma.com/file/RzQhMu3r9tyY9IYEl0JpmK/Untitled?node-id=0%3A1)

#### Main interface

*Essentially: replicate the task card on the tasks list page*

#### Breadcrumbs

*We don't have a lot of breadcrumb type navigation in the app - this might require some iteration*

### Acceptance Criteria
- The page includes a way to navigate back to the tasks list page
- The page shows the status (`active` | `inactive`) of the task
- The page has a toggle to change the status of the task
- Re-running a task should re-load the page
- UX tweaks (e.g. spacing between `View Logs` and `Retry`)

blegesse-git:
closes #1495 

To make code review/ merging easier, I've split #1655 into smaller portions.
This PR contains the remaining portion of changes implemented by PR #1655.

**Issue**: Whether user is editing task from task list page or task runs page, after clicking save, user is always routed back to task list page. 
PR **solution**:
If user edits task from task list page, after clicking save, they will be routed to task list page. 
if user edits task from task runs page, after clicking save, they will be routed to task runs page. 

https://user-images.githubusercontent.com/66275100/122278032-237ca480-ceac-11eb-9eac-6250d5231f75.mov

### DIFF
--- a/src/tasks/actions/creators.ts
+++ b/src/tasks/actions/creators.ts
@@ -148,5 +148,5 @@ export const setCurrentTasksPage = (page: string) =>
   ({
     type: SET_CURRENT_TASKS_PAGE,
     tasksPage: page
-  })
+  } as const)
 
--- a/src/tasks/actions/creators.ts
+++ b/src/tasks/actions/creators.ts
@@ -23,6 +23,7 @@ export const SET_RUNS = 'SET_RUNS'
 export const SET_LOGS = 'SET_LOGS'
 export const REMOVE_TASK = 'REMOVE_TASK'
 export const ADD_TASK = 'ADD_TASK'
+export const SET_CURRENT_TASKS_PAGE = 'SET_CURRENT_TASKS_PAGE'
 
 export type Action =
   | ReturnType<typeof setTasks>
@@ -41,6 +42,7 @@ export type Action =
   | ReturnType<typeof clearCurrentTask>
   | ReturnType<typeof removeTask>
   | ReturnType<typeof addTask>
+  | ReturnType<typeof setCurrentTasksPage>
 
 // R is the type of the value of the "result" key in normalizr's normalization
 type TasksSchema<R extends string | string[]> = NormalizedSchema<
@@ -141,3 +143,10 @@ export const setLogs = (logs: LogEvent[]) =>
     type: SET_LOGS,
     logs,
   } as const)
+
+export const setCurrentTasksPage = (page: string) =>
+  ({
+    type: SET_CURRENT_TASKS_PAGE,
+    tasksPage: page
+  })
+
--- a/src/tasks/actions/creators.ts
+++ b/src/tasks/actions/creators.ts
@@ -147,6 +147,5 @@ export const setLogs = (logs: LogEvent[]) =>
 export const setCurrentTasksPage = (page: string) =>
   ({
     type: SET_CURRENT_TASKS_PAGE,
-    tasksPage: page
+    tasksPage: page,
   } as const)
-
--- a/src/tasks/actions/creators.ts
+++ b/src/tasks/actions/creators.ts
@@ -144,7 +144,11 @@ export const setLogs = (logs: LogEvent[]) =>
     logs,
   } as const)
 
-export const setCurrentTasksPage = (page: string) =>
+export enum TaskPage {
+  TasksPage = 'TasksPage',
+  TaskRunsPage = 'TaskRunsPage',
+}
+export const setCurrentTasksPage = (page: TaskPage) =>
   ({
     type: SET_CURRENT_TASKS_PAGE,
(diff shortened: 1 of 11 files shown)


src/tasks/actions/creators.ts
```diff
@@ -141,3 +143,9 @@ export const setLogs = (logs: LogEvent[]) =>
     type: SET_LOGS,
     logs,
   } as const)
+
+export const setCurrentTasksPage = (page: string) =>
```
hoorayimhelping, line 147:
```ts
(page: string)
```
we currently only accept two types of pages right? instead of making this a type `string` which can accept any string value (e.g. `"foo"`, `""`, `"3234"`), if we make it a [TypeScript String Enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) we can list all the possible values and get warned when one isn't used:

```ts
export enum TaskPage {
  TasksPage = 'TasksPage',
  TaskRunsPage = 'TaskRunsPage'
}

export const setCurrentTasksPage = (page: TaskPage) =>
```

src/tasks/actions/thunks.ts
```diff
@@ -411,7 +431,12 @@ export const updateScript = () => async (
       throw new Error(resp.data.message)
     }
 
-    dispatch(goToTasks())
+    if (currentPage === 'TasksPage') {
+      dispatch(goToTasks())
+    } else if (currentPage === 'TaskRunsPage') {
+      dispatch(goToTaskRuns())
+    }
```
hoorayimhelping, line 438: If we use the enum described above, this would turn into:
```ts
    if (currentPage === TaskPage.TasksPage) {
      dispatch(goToTasks())
    } else if (currentPage === TaskPage.TaskRunsPage) {
      dispatch(goToTaskRuns())
    }
```

this way, we have two cases for the two enum values and we probably don't need to define a fallback function

src/tasks/actions/thunks.ts
```diff
@@ -592,3 +617,15 @@ export const runDuration = (finishedAt: Date, startedAt: Date): string => {
...
+  dispatch: Dispatch<Action>
+) => {
+  dispatch(setCurrentTasksPage('TasksPage'))
+}
+
+export const setTaskRunsPageAsCurrent = () => async (
+  dispatch: Dispatch<Action>
+) => {
+  dispatch(setCurrentTasksPage('TaskRunsPage'))
+}
```
hoorayimhelping, line 631: let's remove these functions as they're not necessary.

Thunks exist to handle [side effects](https://en.hexlet.io/courses/intro_to_programming/lessons/pure/theory_unit) when dispatching redux actions. Other thunks in our codebase have side effects like making a request, or updating a file. Since these only change the value of the redux state, these are pure actions and don't need a thunk. Dispatching an action is a UI level concern that components should handle.

In the places you call the thunk, instead just call the actions and pass in the proper value.

src/tasks/reducers/index.ts
```diff
@@ -163,5 +164,10 @@ export default (
 
         return
       }
+
+      case SET_CURRENT_TASKS_PAGE: {
```
hoorayimhelping, line 168: does there need to be a default state set for this?

src/tasks/components/TaskCard.tsx
```diff
@@ -264,6 +270,7 @@ export class TaskCard extends PureComponent<
 const mdtp = {
   onAddTaskLabel: addTaskLabel,
   onDeleteTaskLabel: deleteTaskLabel,
+  setTasksPageAsCurrent,
```
hoorayimhelping, line 273: after removing the thunk, this should simply change to `setCurrentTasksPage`. then you'd call it like

```ts
setCurrentTasksPage('TasksPage')
```

hoorayimhelping (changes requested):
good, clean code! all the requests deal with redux convention and using some of the institutional redux learning we've had over the past year or so

src/tasks/actions/creators.ts
blegesse-git: thank you for including the documentation!

src/tasks/actions/thunks.ts
blegesse-git: that makes sense. I've removed them and called the action in TaskRunsCard and TaskCard and passed the respective values

hoorayimhelping (approved):
great work!!

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:

PR #1729, opened 2021-06-16
feat: maps tooltip

### LINKED ISSUE #1715: Maps: Add ability to send tooltips in with Config
In giraffe, we have added a new property on PointMapLayer to be able to send in tooltipColumns. This issue is to update the UI to pull in the new properties.

zoesteinkamp:
Closes #1715

The ui side of tooltip columns


src/views/helpers/index.ts
```diff
@@ -403,6 +403,14 @@ const NEW_VIEW_CREATORS = {
           colorField: '_value',
           colors: DEFAULT_THRESHOLDS_GEO_COLORS,
           isClustered: false,
+          tooltipColumns: [
```
Palakp41, line 406: I don't think every dataset provided will have these columns. A lot of the times we filter out the columns and the column names like `s2_cell_id` can be named differently. What happens if this array is empty as it should be unless populated right?

Palakp41 (changes requested):
I think you need to specify the tooltip columns to be an empty array, or with just `start`, `stop`, and `time` columns, which we know always exist.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/visualization/types/Map/view.tsx
```diff
@@ -122,6 +124,7 @@ const GeoPlot: FC<Props> = ({result, properties}) => {
         colorField: '_value',
         colors: DEFAULT_THRESHOLDS_GEO_COLORS,
         isClustered: false,
+        tooltipColumns: tooltipColumns,
```
Palakp41, line 127: So the code never reaches here, hence the tooltip columns is always `[]`. This is a safe fail guard against layers. You might want to do same thing as we are doing for colors in the line 133.

PR #1730, opened 2021-06-16
feat: add auto adjust for initial static legend height

### LINKED ISSUE #1523: Static Legend: a new Dashboard cell should have default size that fits
When creating a new dashboard cell and using the static legend, the default size should fit the static legend without needing to manually resize the cell. Or at least, fit _more_ of the current static legend especially in vertical orientation. As a new feature, static legend should be readily useful without too much manual manipulation and fussing around.

Here is an example of newly created cells with static legends in vertical orientation. This is impossible to read and puts the onus on the user to resize them. Which is very frustrating.
Screen Shot 2021-05-24 at 10 06 41 AM

TCL735:
Closes #1523 

Screen Shot 2021-06-16 at 2 08 37 PM

Screen Shot 2021-06-16 at 2 09 18 PM


src/dashboards/utils/cellGetters.ts
```diff
@@ -34,7 +45,8 @@ export const isCellUntitled = (cellName: string): boolean => {
 export const getNewDashboardCell = (
   state: AppState,
   dashboard: Dashboard,
-  clonedCell?: Cell
+  clonedCell: Cell,
```
Asalem1, line 48: for anyone that's curious this function only looks like it's ever being used in one place

Asalem1, line 48: this gives me a funny feeling since clonedCell is an optional parameter that we're making required, but is still optional to the parent

TCL735: There's only 1 instance of `getNewDashboardCell` being called and it checks if clonedCell is truthy or not. This PR changes `getNewDashboardCell` so it seemed fine to adjust the parameters.

The parent has multiple calls and i didn't want to mess with the parent code in this PR.

Asalem1: for sure, I guess maybe we add a default there since the parent of `getNewDashboardCell` is `createCellWithView` which accepts an optional `clonedCell` parameter. I guess just a nit since `clonedCell` can be `undefined`?

Asalem1: https://github.com/influxdata/ui/blob/master/src/cells/actions/thunks.ts#L73

Asalem1: I don't know, my brain gets foggy at this hour so I might be super paranoid

Asalem1 (approved):

src/dashboards/utils/cellGetters.ts
TCL735: Hmm, i don't think it needs a default. i don't want to inadvertently change anything unrelated.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1730)

PR #1732, opened 2021-06-17
fix: correct default cell height for static legend and cap lower bounds

TCL735:
Part of #1523 

- Previously, i capped the upper bounds of static legend height. The lower bounds should also be capped. This is just for thoroughness, as the static legend height should come back as positive number from Giraffe's renderEffect options.
- Fix the default cell height when user creates a cell through the Data Explorer rather than "Add Cell"

### DIFF
--- a/src/visualization/utils/useStaticLegend.ts
+++ b/src/visualization/utils/useStaticLegend.ts
@@ -131,6 +131,9 @@ export const useStaticLegend = (properties): StaticLegendConfig => {
           if (heightRatio < STATIC_LEGEND_HEIGHT_RATIO_MINIMUM) {
             heightRatio = STATIC_LEGEND_HEIGHT_RATIO_MINIMUM
           }
+          if (typeof heightRatio !== 'number' || heightRatio !== heightRatio) {
+            heightRatio = STATIC_LEGEND_HEIGHT_RATIO_DEFAULT
+          }
 
           update({
             heightRatio,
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/visualization/utils/useStaticLegend.ts
```diff
@@ -122,11 +123,14 @@ export const useStaticLegend = (properties): StaticLegendConfig => {
             estimatedHeight = length * sampleMaxHeight + padding
           }
 
-          const updatedRatio = estimatedHeight / totalHeight
-          const heightRatio =
-            updatedRatio <= STATIC_LEGEND_HEIGHT_RATIO_MAXIMUM
-              ? updatedRatio
-              : STATIC_LEGEND_HEIGHT_RATIO_MAXIMUM
+          let heightRatio = estimatedHeight / totalHeight
```
hoorayimhelping, line 126: do we need a divide by zero check here, just in case `totalHeight` ever ends up being `0`?

TCL735: Yes good catch. In fact, we have to guard against all sorts of silly falsy values.

TCL735: Updated.

TCL735:
@hoorayimhelping i also used this PR to fix a minor issue i found with the default height when a user goes through the Data Explorer rather than using "Add Cell"

hoorayimhelping (approved):

PR #1736, opened 2021-06-17
feat: View list of secrets

### LINKED ISSUE #1581: Secrets: View a List of Secret Keys in an Organization
We would like the ability for users to to see a list of their secret keys with in the UI. This list should display only the secret key, using our existing list interface (https://www.figma.com/proto/1VaHno7NzQTFvzRtCnROay/Token%2FSecret?node-id=253%3A1903&scaling=min-zoom&page-id=253%3A1902)

Relevant API details:

https://docs.influxdata.com/influxdb/v2.0/api/#operation/GetOrgsIDSecrets
https://github.com/influxdata/openapi/blob/master/contracts/common.yml#L3573-L3661

**Acceptance Criteria**

* Given a user has 1 or more secrets, they can see a list of secret keys for their organization
* List view provides filter and sort combination from existing list views

joelfried:
Closes #1581 

Adds a list of existing secrets onto the Settings page behind a feature flag.  Users can copy the secret name via the Copy to Clipboard button or clicking the Secret name.

image


cypress/index.d.ts
```diff
@@ -37,6 +37,8 @@ import {
   createDashWithViewAndVar,
   createRule,
   clickAttached,
+  upsertSecret,
+  setFeatureFlags,
```
Asalem1, line 41: $

package.json
```diff
@@ -169,6 +169,7 @@
     "calculate-size": "^1.1.1",
     "chroma-js": "^1.3.6",
     "classnames": "^2.2.3",
+    "clipboardy": "^2.3.0",
```
Asalem1, line 172: If this is only being used for testing we should make it a devDependency

src/schemas/secrets.ts
```diff
@@ -0,0 +1,12 @@
+// Libraries
+import {schema} from 'normalizr'
+
+// Types
+import {ResourceType} from 'src/types'
+
+/* Tasks */
+
+// Defines the schema for the tasks resource
```
Asalem1, line 9: What're the task references for?

src/secrets/actions/thunks.ts
```diff
@@ -0,0 +1,117 @@
...
+} from 'src/secrets/actions/creators'
+
+// Utils
+import {getOrg} from 'src/organizations/selectors'
+import {getStatus} from 'src/resources/selectors'
+
+const makeEntitiesForSecrets = (response: Array<string>) => {
+  const processedData = []
+  response.forEach(element => processedData.push({id: element, key: element}))
+  return processedData
```
Asalem1, line 45: nit: 
```suggestion
  return response.map(element => {id: element, key: element})
```

src/secrets/actions/thunks.ts
```diff
@@ -0,0 +1,117 @@
...
+    },
+  })
+
+  if (resp.status === 204) {
+    dispatch(removeSecret(id))
+  } else {
+    console.error(resp.data.message)
+    dispatch(notify(deleteSecretsFailed()))
+  }
+}
```
Asalem1, line 117: nit: using the try/catch pattern we do in other thunks

src/secrets/actions/thunks.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export const deleteSecret = (id: string) => async (
+  dispatch: Dispatch<Action>,
+  getState: GetState
+) => {
+  const org = getOrg(getState())
+  const resp = await apiDeleteSecret({
+    orgID: org.id,
+    data: {
+      secrets: [id],
```
Asalem1, line 107: why is the API expecting an array of secrets to be deleted? Can we update the API definition here since we're only ever going to be deleting one at a time right?

src/secrets/components/SecretCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+  )
+}
+
+const mdtp = {
+  notify: notifyAction,
+}
+
+const connector = connect(null, mdtp)
+
+export default connector(SecretCard)
```
Asalem1, line 66: Rather than integrating all these wrappers we can just integrate the `useDispatch` hook within the component

src/secrets/components/SecretContextMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (isSuccessful) {
+      notify(copyToClipboardSuccess(copiedText, 'Secret ID'))
+    } else {
+      notify(copyToClipboardFailed(copiedText, 'Secret ID'))
+    }
+  }
+
+  return (
+    <Context>
+      <CopyToClipboard text={secret.id} onCopy={handleCopyAttempt}>
```
Asalem1, line 48: What's copied when we click the button?

src/secrets/components/SecretContextMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+  )
+}
+
+const mdtp = {
+  notify: notifyAction,
+}
+
+const connector = connect(null, mdtp)
+
+export default connector(SecretContextMenu)
```
Asalem1, line 69: same as the previous PR

src/secrets/actions/thunks.ts
```diff
@@ -0,0 +1,117 @@
...
+    )
+    dispatch(setSecrets(RemoteDataState.Done, test))
+  } else {
+    console.error(resp.data.message)
+    dispatch(setSecrets(RemoteDataState.Error))
+    dispatch(notify(getSecretsFailed()))
+  }
+}
+
+export const upsertSecret = (_secretKey: string, secretValue: string) => async (
```
Asalem1, line 75: why are we passing in the secretKey if we're never using it?

src/secrets/components/SecretsTab.tsx
```diff
@@ -0,0 +1,178 @@
...
+        <FilterSecrets
+          list={secrets}
+          searchTerm={searchTerm}
+          searchKeys={['key', 'secret[].key']}
+        >
+          {sc => (
+            <SecretsList
+              secrets={sc}
+              emptyState={SecretsEmptyState()}
+              onDeleteSecret={() => {}}
```
Asalem1, line 154: the infrastructure seems to be in place for this, why isn't it being integrated?

src/secrets/containers/SecretsIndex.tsx
```diff
@@ -12,20 +12,31 @@ import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
...
     </>
   )
 }
 
-export default SecretsIndex
+const mstp = (state: AppState) => ({org: getOrg(state)})
+
+const connector = connect(mstp, null)
+
+export default connector(SecretsIndex)
```
Asalem1, line 42: we're already getting the org in the selector above: 

https://github.com/influxdata/ui/pull/1736/files#diff-60be7299ee28a12b945e5efd3d4e70741fbb0600db9f1c998d0d2d61e3255f42R22

src/secrets/components/SecretsTab.tsx
```diff
@@ -0,0 +1,178 @@
...
+  return {secrets}
+}
+
+// const mdtp = {
+//     onDeleteSecret: deleteSecret,
+// }
+
+const connector = connect(mstp, null)
+
+export default connector(SecretsTab)
```
Asalem1, line 178: Let's try and stick the hook pattern here since that's got way less baggage

src/secrets/components/SecretsTab.tsx
```diff
@@ -0,0 +1,178 @@
...
+  sortDirection?: Sort
+  sortType?: SortTypes
+}
+
+type ReduxProps = ConnectedProps<typeof connector>
+type Props = OwnProps & ReduxProps
+
+const SecretsTab: FC<Props> = ({
+  secrets,
+  // importOverlayState = OverlayState.Closed,
```
Asalem1, line 42: is there a plan to integrate overlays for this feature?

src/secrets/actions/thunks.ts
```diff
@@ -0,0 +1,117 @@
...
+      secretsSchema
+    )
+    dispatch(setSecret(resp.data.secret.id, RemoteDataState.Done, secret))
+  } else {
+    console.error(resp.data.message)
+    dispatch(notify(upsertSecretFailed()))
+  }
+}
+
+export const deleteSecret = (id: string) => async (
```
Asalem1, line 99: this doesn't seem like it's being used

src/secrets/actions/creators.ts
```diff
@@ -0,0 +1,58 @@
...
+    schema,
+  } as const)
+
+export const removeSecret = (id: string) =>
+  ({
+    type: REMOVE_SECRET,
+    id,
+  } as const)
+
+export const setSecretOnResource = (
```
Asalem1, line 50: where is this being used?

cypress/e2e/shared/secrets.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    // Empty state exists and displays context appropriate text
+    cy.getByTestID('empty-state').should('be.visible')
+    cy.getByTestID('empty-state').contains('Secrets')
+
+    // Create single secret, make sure it's visible
+    cy.get('@org').then(({id: orgID}: Organization) => {
+      cy.upsertSecret(orgID, {toEverybody: 'rupees baby'})
+        .then(() => {
+          cy.reload()
+          cy.getByTestID('secret-card--toEverybody').should('be.visible')
```
Asalem1, line 29: can we check for the notification here?

cypress/e2e/shared/secrets.test.ts
```diff
@@ -0,0 +1,82 @@
...
+            .click({force: true})
+            .then(() => {
+              cy.task('getClipboard').should('eq', 'toEverybody')
+            })
+        })
+        // Create a second secret, make sure it's visible and test sorting
+        .upsertSecret(orgID, {CocaColaRecipe: 'lol'})
+        .then(() => {
+          cy.reload()
+          cy.getByTestID('secret-card--toEverybody').should('exist')
```
Asalem1, line 43: can we validate that the original one is still there?

Asalem1 (changes requested):
Looks great. Would love this cleaned up a little bit before getting it in. I was also curious if we plan on exposing secrets anywhere else in the app / if the list will open a new page or something like that? The reason I ask is that it seems like there's a lot of work to expose secrets in the redux state, but that might not be necessary if we plan on keeping things limited to the secrets page

src/secrets/components/SecretsTab.tsx
joelfried: The Overlay will be used for creating new secrets which is coming next.  I thought about getting rid of these commented bits for cleanliness's sake but they won't be around in this state for more than a few days so it seemed silly to take them all out to commit them and then put them all back in so I can use them again.  But if you have strong opinions about it I can remove the comments for now.

src/secrets/actions/thunks.ts
joelfried: There's a ticket to fully implement it that's on my plate for next week.

cypress/e2e/shared/secrets.test.ts
joelfried: The upsert being used on line 26 is a straight API call to push the secret in - there isn't a frontend way to add a new secret yet.  So there isn't a notification that a new secret has been added it just shows up there as it now exists in the backend.

Asalem1: Gotcha, my bad

cypress/e2e/shared/secrets.test.ts
joelfried: This one is the initial one - I checked existence on both rather than visibility.  I can add a check that both are visible as well if you'd like?

Asalem1: Sorry, yup, that's what I meant. Thanks!

src/secrets/actions/creators.ts
joelfried: Good catch this isn't and won't be.  Removed.

cypress/e2e/shared/secrets.test.ts
joelfried: Added

src/secrets/components/SecretsTab.tsx
Asalem1: nah, all good.

src/schemas/secrets.ts
joelfried: . . . leftover copypasta that should've been cleaned up.  Removed!

package.json
joelfried: Good point.  Changed.

src/secrets/actions/thunks.ts
joelfried: It's used on line 81

src/secrets/actions/thunks.ts
joelfried: The generatedRoutes.ts expected an array in `PostOrgsSecretsDeleteParams` and I wasn't sure how to go about changing it so I just rolled with what was there.  When I finish implementing deletion next week it'll always be an array with a single element as the UI has a single delete button on a given secret.  If there's a way to change that around easily from our end I don't mind doing it I just wanted to avoid involving multiple people to change the definition from a single element array to a single element.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/secrets/actions/thunks.ts
Asalem1: For sure, I was just curious if this endpoint is being used by the CLI and is built out that way or if the swagger definition and API can be updated to consolidate this behavior since it seems a bit weird to me. We've run into this in several places so I know it's not an issue for you to resolve, just curious if this was designed with the intent that an array of secret ids can be passed in or if it's only ever expecting 1 and we can update the API to address that discrepancy

src/secrets/actions/thunks.ts
Asalem1: Screen Shot 2021-06-17 at 11 51 55 AM

Not sure which line 81 you're referring to

joelfried: image
The one :point_down:

Asalem1: That's a property that you're assigning a value to. 

> image
> The one 👇

Asalem1: If you remove the `_` from secretKey it'll throw a linter error because `_secretKey` is unused.

Asalem1: Screen Shot 2021-06-17 at 12 03 12 PM

joelfried: Ah I see!  Added those brackets and will test this more heavily when the add button is implemented.  This probably saved me a bunch of confusion later - thanks!

src/secrets/components/SecretContextMenu.tsx
joelfried: The value of `text` which is the key from the secret.

Asalem1: Awww, yup didn’t see text being assigned the secret id. Cool cool cool

joelfried:
> Looks great. Would love this cleaned up a little bit before getting it in. I was also curious if we plan on exposing secrets anywhere else in the app / if the list will open a new page or something like that? The reason I ask is that it seems like there's a lot of work to expose secrets in the redux state, but that might not be necessary if we plan on keeping things limited to the secrets page

There's a ticket for linking to secrets anywhere you can edit flux so I thought it likely we'd want to be able to use these things elsewhere reasonably soon.

src/secrets/components/SecretsTab.tsx
joelfried: I started to and then realized as I was doing it that if I kept it up I'd end up doing the entire epic in one PR which is probably not the best of practices.  So I left things disconnected for now and I'll hook this up early next week.

src/secrets/containers/SecretsIndex.tsx
joelfried: Removed.

src/secrets/actions/creators.ts
```diff
@@ -0,0 +1,47 @@
+// Types
+import {RemoteDataState, SecretEntities} from 'src/types'
+import {Action as NotifyAction} from 'src/shared/actions/notifications'
+import {NormalizedSchema} from 'normalizr'
+
+export type Action =
+  | ReturnType<typeof removeSecret>
+  | ReturnType<typeof setSecret>
+  | ReturnType<typeof setSecrets>
+  | NotifyAction
```
Asalem1, line 10:
```suggestion
```

src/secrets/actions/thunks.ts
joelfried: The docs [here](https://docs.influxdata.com/influxdb/v2.0/security/secrets/manage-secrets/delete/) say (emphasis mine):
> Use the POST request method and the orgs/{orgID}/secrets/delete API endpoint to delete one **or more** secrets.

The CLI only seems to allow one delete at a time as far as I can tell (I've tried and the docs don't say "or more" there).  I agree it's a bit strange but it seems intentional so I was just going to roll with it and pass in single element arrays.

src/secrets/components/SecretsTab.tsx
joelfried: Yeah after refactoring this out a few times in this PR now I completely agree with that assessment.

src/secrets/actions/thunks.ts
Asalem1: @glinton it looks like the openAPI description here is that it's to delete a secret, but the $ref is to Secrets. Should we change the schema definition to `Secret` or update the description?

https://github.com/influxdata/openapi/blob/50ceec33e0c114d39a9399efd9a0929aa54c77b4/src/common/paths/orgs_orgID_secrets_delete.yml#L14-L20

glinton: so glad this came up. i'm not sure who added this route to the original api, but i'd love for it to change if ya'll are working on it. we can mark the one that posts to a `/delete` route as deprecated and add a `DELETE /secrets/{id}` route. The original looked to be a workaround anyway ([`"/orgs/{orgID}/secrets/delete": # had to make this because swagger wouldn't let me have a request body with a DELETE`](https://raw.githubusercontent.com/influxdata/influxdb/b66079f08470575dc1b22c80be2a6be39f6e7b4c/http/swagger.yml)).  Not knowing really what that api is used for and assuming that the keys currently passed in are actually the ids of the key, i think it would be fine to do a `delete` request to the secret url.

Regarding the question about updating the schema, I'm seeing `SecretKeys` (which as mentioned above, is ambiguous)
image

I don't remember ever checking, but i've believed that the api accepts (and requires) an array of strings.

joelfried: @glinton Implementing the delete is on my to do list for next week so this is definitely being worked on.  The spec as written has us only ever deleting one secret at a time so a `DELETE /secrets/{id}` route makes a lot of sense to me.

glinton: if you need any assistance building support for it on the api side, feel free to reach out

Asalem1 (approved):

PR #1746, opened 2021-06-17
fix: Notebook maps error

### LINKED ISSUE #1615: Maps is broken in Notebooks
Trying to run the following query:
```
import "experimental/csv"

csv.from(url: "https:/"+"/raw.githubusercontent.com/russorat/influxdb-data-examples/master/gov/usgs/earthquake/all_week-annotated.csv")
  |> filter(fn: (r) => r._field == "lat" or r._field == "lon")
```

gives me an error in notebooks. Error is on this line: https://github.com/influxdata/ui/blob/master/src/visualization/types/Map/view.tsx#L133

image

zoesteinkamp:
Closes #1615 

Okay so this fixes the color issue by auto assigning the color similar to our logic for the tooltipColumns and this has layer opts be above all code calling for layer properties. I was finding that the code was crashing because it was look for layer.properties.blah and finding nothing. So i moved all the code underneath and i set layerOpts = layers, which was missing so things were still crashing when they checked the properties.

### DIFF
--- a/src/visualization/types/Map/view.tsx
+++ b/src/visualization/types/Map/view.tsx
@@ -28,9 +28,6 @@ type GeoCoordinates = {
 const GeoPlot: FC<Props> = ({result, properties}) => {
   const {layers, zoom, allowPanAndZoom, mapStyle} = properties
   const {lat, lon} = properties.center
-  const tooltipColumns = _.isEmpty(properties.layers[0].tooltipColumns)
-    ? result.fluxGroupKeyUnion
-    : properties.layers[0].tooltipColumns
 
   const [mapServiceError, setMapServiceError] = useState<RemoteDataState>(
     RemoteDataState.NotStarted
@@ -117,6 +114,7 @@ const GeoPlot: FC<Props> = ({result, properties}) => {
     tileServerUrl: getMapboxUrl(),
     bingKey: '',
   }
+
   let layersOpts = layers
   if (!layers.length) {
     layersOpts = [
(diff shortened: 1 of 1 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/visualization/types/Map/view.tsx
```diff
@@ -126,16 +124,23 @@ const GeoPlot: FC<Props> = ({result, properties}) => {
         colorField: '_value',
         colors: DEFAULT_THRESHOLDS_GEO_COLORS,
         isClustered: false,
-        tooltipColumns: tooltipColumns,
+        tooltipColumns: [],
```
Palakp41, line 127: I think this should be updated to be:
```
tooltipColumns: [],
colors: []
```
because we want to to give them a default value but later we can check if `layerOpts[0].tooltipColumns` is empty, then we can updated to use fluxgroup key union values.

src/visualization/types/Map/view.tsx
```diff
@@ -126,16 +124,23 @@ const GeoPlot: FC<Props> = ({result, properties}) => {
...
       },
     ]
+    properties.layers = layersOpts
```
Palakp41, line 130: This seems a bit weird to me. First on Line 118, we are giving `layerOpts` value of `properties.layers` and then doing the opposite here. What if you removed this and instead of using `properties.layers` on line 133, you use `layerOpts`?

src/visualization/types/Map/view.tsx
```diff
@@ -126,16 +124,23 @@ const GeoPlot: FC<Props> = ({result, properties}) => {
...
   }
 
-  layersOpts[0].tooltipColumns = tooltipColumns
+  const tooltipColumns = _.isEmpty(properties.layers[0].tooltipColumns)
```
Palakp41, line 133: Can you update this to be: 
```
const tooltipColumns = _.isEmpty(layerOpts[0].tooltipColumns)
```

src/visualization/types/Map/view.tsx
```diff
@@ -126,16 +124,23 @@ const GeoPlot: FC<Props> = ({result, properties}) => {
...
+    ? result.fluxGroupKeyUnion
+    : properties.layers[0].tooltipColumns
 
-  if (!layers[0].colors[0].id) {
-    layersOpts[0].colors = DEFAULT_THRESHOLDS_GEO_COLORS
-  }
+  const colorChoice = _.isEmpty(properties.layers[0].colors)
```
Palakp41, line 137: Can you update this to be:
```
const colorChoice = _.isEmpty(layerOpts[0].colors)
```

Palakp41 (approved):
Looks good!

PR #1762, opened 2021-06-22
feat: send org id parameter for annotations requests to support OSS multi-org

### LINKED ISSUE #1645: Annotations: Pass OrgID for List & Create on OSS
For annotations support on OSS, we need to pass the `orgID` to the backend for the "list all" and "create" types of actions for annotations (and streams). This is because OSS allows for a user to belong to multiple organizations, and the state of which organization is "active" for the user is contained entirely in the frontend. So when a user wants to get a list of all annotations, the frontend needs to provide which organization the request is for. Similiarly, when a user creates an annotation, the organization ID needs to be persisted with the annotation information in the database, so the frontend needs to provide that ID with the request.

wbaker85:
Closes #1645

Also see https://github.com/influxdata/openapi/pull/135

This PR will pull in the API definition for the OSS annotations API, which requires an org ID parameter for certain requests where required due to the multi-org-per-user behavior of OSS. The `openapi` SHA will be updated to the commit which added this API definition.

The actual requests which use these endpoints in the UI are being updated to send the org ID parameter where needed as well. The org ID parameter is not required for cloud and sending it should not have any impact on those requests since the annotations service for cloud derives the org ID for the token. This is a similar approach as was taken for Notebooks on OSS, see: https://github.com/influxdata/ui/pull/1390

An alternative for sending the org ID parameter on requests to cloud where it is not needed would be to have some branching logic based on the `CLOUD` constant around where the request parameters are initialized to determine if the org ID should be included or not. This wouldn't be extremely complicated since there are _currently_ only a couple of places where it would be needed, but it might be more difficult to sustain. Happy to discuss this (and all aspects of this PR) further and open to considering other options!

### DIFF
--- a/src/annotations/api/index.ts
+++ b/src/annotations/api/index.ts
@@ -12,7 +12,6 @@ import {
   postAnnotation,
   putAnnotation,
   AnnotationEvent,
-  GetAnnotationsParams,
 } from 'src/client/annotationdRoutes'
 
 // Utils
@@ -23,7 +22,8 @@ export const getAnnotationStreams = (): Promise<AnnotationStream[]> => {
 }
 
 export const writeAnnotation = async (
-  annotations: Annotation[]
+  annotations: Annotation[],
+  orgID: string
 ): Promise<Annotation[]> => {
   // RFC 3339 is the standard serialization format for dates across the wire for annotations
   const annotationsRequestConverted = annotations.map(annotation => {
@@ -33,7 +33,7 @@ export const writeAnnotation = async (
       endTime: new Date(annotation.endTime).toISOString(),
     }
   })
-  const params = {data: annotationsRequestConverted}
+  const params = {data: annotationsRequestConverted, query: {orgID}}
   const res = await postAnnotation(params)
 
   if (res.status >= 300) {
@@ -55,10 +55,11 @@ export const writeAnnotation = async (
 }
 
 export const getAnnotations = async (
+  orgID: string,
   stream?: string
 ): Promise<AnnotationResponse[]> => {
-  const params: GetAnnotationsParams = {
-    query: {AnnotationListFilter: formatAnnotationQueryString({stream})},
+  const params = {
+    query: {AnnotationListFilter: formatAnnotationQueryString({stream}), orgID},
   }
   const res = await getAnnotationsApi(params)
   if (res.status >= 300) {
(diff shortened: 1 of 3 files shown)


src/annotations/api/index.ts
```diff
@@ -55,10 +55,11 @@ export const writeAnnotation = async (
 }
 
 export const getAnnotations = async (
+  orgID: string,
   stream?: string
 ): Promise<AnnotationResponse[]> => {
-  const params: GetAnnotationsParams = {
-    query: {AnnotationListFilter: formatAnnotationQueryString({stream})},
+  const params = {
```
wbaker85, line 61: Had to make this not have a specific type, otherwise it wouldn't work for the cloud annotations API which doesn't have the `orgID` parameter.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


wbaker85:
> @wbaker85 Do [this instance](https://github.com/influxdata/ui/blob/master/src/dashboards/components/DashboardPage.tsx#L63) of `fetchAndSetAnnotations` and [this instance](https://github.com/influxdata/ui/blob/master/src/visualization/utils/annotationUtils.ts#L38) of `writeThenFetchAndSetAnnotations` need to have `getStore` passed in?
> 
> edit: [this commit](https://github.com/influxdata/ui/commit/1628924c893b53dc38629e3b0a9a2ed73979c1a6#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R53) updated the open api hash. can you confirm that it's newer than the one you're referencing?

For passing in `getStore`, I don't think either of those invocations needs it. I'm basing this mostly on the fact that I didn't do it, and the code seems to work, rather than an entirely scientific explanation though! It's been a while since I took the time to understand how react/redux works under the hood, but as I understand it we can get access to the state via `getState` in the thunks because the top-level `Provider` [makes the store available with `getStore()`](https://github.com/influxdata/ui/blob/cb7cc793d8e662f74ea10f4cf5984a91542d8b88/src/index.tsx#L70) to the subcomponents and that miraculously transfers through to the thunks using `dispatch`, either via hook [here](https://github.com/influxdata/ui/blob/cb7cc793d8e662f74ea10f4cf5984a91542d8b88/src/visualization/types/Band/view.tsx#L221) (eventually makes its way to the `writeThenFetchAndSetAnnotations`), or by using the connector and `mdtp` [here](https://github.com/influxdata/ui/blob/de4407743e512ed633817e3378de9f783e3e1fec/src/dashboards/components/DashboardPage.tsx#L155) for `fetchAndSetAnnotations`.

The linked commit did update the openapi hash to a later one that this...the one right after it actually! I'll rebase and get rid of that diff.

hoorayimhelping (approved):
lgtm!

see: https://github.com/influxdata/idpe/issues/11083

Issue #1027, opened 2021-03-30
Maps:  add the ability to select the field name to use for lat/lon


Palakp41:
as a user with geo data in my instance, the data might be stored as fields with different names than the maps widget expects.

Lat Lon Stored as ~Tag Values or~ Field Values
given i have data with latitude/longitude stored as fields
when I attempt to display that data as a map
then i would like the ability to choose the fields that contain that lat/lon data

~given i have data with latitude/longitude stored as tags~
~when i attempt to display that data as a map~
~then i would like the ability to choose the tags that contain the lat/lon data~

If the Lat Lon is not provided by the user, maps should display a friendly error and direct user to use customization options.

kristinarobinson:
Removing from MVP

PR #1770, opened 2021-06-22
feat: add customization options for lat/lon or s2 column in maps

### LINKED ISSUE #1029: Maps: Ability to select tag for s2_cell_id
S2_cell_id stored as tag value ~or field values~
given i have data with the s2_cell_id calculated and stored as a tag value
when i attempt to display that data on a map
then i would like the ability to choose that s2_cell_id to use as the lat and lon values

~given i have data with the s2_cell_id calculated and stored as a field value~
~when i attempt to display that data on a map~
~then i would like the ability to choose that s2_cell_id field to use as the lat and lon values~

Palakp41:
Closes #1029 #1027 

This pr adds the ability to allow users to pick the lat/lon or s2 columns in the customization options.

Kapture 2021-06-22 at 15 44 00

### DIFF
--- a/src/shared/utils/vis.ts
+++ b/src/shared/utils/vis.ts
@@ -327,52 +327,6 @@ enum CoordinateType {
   None = 'Lat/Long not provided',
 }
 
-
-const latLonAsTags = (latLonColumns) => {
-    if (
-      latLonColumns?.lat?.key === 'tag' &&
-      latLonColumns?.lon?.key === 'tag'
-    ) {
-      console.log("latlon astagws")
-      return true
-    }
-    return false
-}
-
-const getCoordinateColumn = (table: Table, useS2CellID: boolean, s2Column: string, latLonColumns: LatLonColumns): string => {
-  try {
-    if (useS2CellID && s2Column !== "") {
-      console.log("inside s2")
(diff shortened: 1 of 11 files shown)


TCL735 (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/shared/utils/vis.ts
```diff
@@ -456,3 +461,115 @@ export const getDetectCoordinatingFields = (table: Table) => {
...
+  )
+
+  if (
+    coordinateColumn === CoordinateType.S2 ||
+    coordinateColumn === CoordinateType.Tags
+  ) {
+    return false
+  }
+  return true
+}
```
hoorayimhelping, line 575: question, what is the `flagged` part of the methods? I was trying to see where they reference feature flags but couldn't find it

hoorayimhelping (commented):
I don't have any blocking stuff, more general comments. We can follow up on them or do them now. I do think that writing some tests for this will make future maintenance easier and you should consider that.

Palakp41:
@hoorayimhelping thank you for your review! I will add these suggestions in the next pr :)

PR #1793, opened 2021-06-23
feat: sso login

subirjolly:
Closes #

We need to be able to login using SSO when a participating domain's email is trying to login. There's a fixme which is depending on #4824. Added feature flag `ssoLogin` temporarily to test in tools as this functionality can't be tested locally because we use dex for auth locally and this change requires Auth0 to redirect us to a callback route.

### DIFF
--- a/src/authorizations/apis/index.ts
+++ b/src/authorizations/apis/index.ts
@@ -40,7 +40,9 @@ export const getAuth0Config = async (
 
 export const getConnection = async (email: string): Promise<string | Error> => {
   // FIXME: When Quartz #4824 gets fixed. Change encodeURI to encodeURIComponent
-  const response = await getAuthConnection({query: {email: encodeURI(email)}})
+  const response = await getAuthConnection({
+    query: {email: encodeURIComponent(email)},
+  })
 
   if (response.status >= 500) {
     throw new Error(response.data)
--- a/src/authorizations/apis/index.ts
+++ b/src/authorizations/apis/index.ts
@@ -1,6 +1,7 @@
 import AJAX from 'src/utils/ajax'
 import {Authorization, Auth0Config} from 'src/types'
 import {getAPIBasepath} from 'src/utils/basepath'
+import {getAuthConnection} from 'src/client/unityRoutes'
 
 export const createAuthorization = async (
   authorization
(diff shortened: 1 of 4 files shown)


src/localStorage.ts
```diff
@@ -31,6 +31,7 @@ export const loadLocalStorage = (): LocalStorage => {
 
 export const setToLocalStorage = (prop: string, value: any): void => {
   try {
+    console.log({prop, value})
```
genehynson, line 34: nit: remove console logs

src/authorizations/apis/index.ts
```diff
@@ -29,9 +30,23 @@ export const getAuth0Config = async (
...
+
     return data
   } catch (error) {
     console.error(error)
     throw error
   }
 }
+
+export const getConnection = async (email: string): Promise<string> => {
+  // FIXME: When #4824 gets fixed. Change encodeURI to encodeURIComponent
```
genehynson, line 42: #4824 in this repo?

Asalem1: I imagine in quartz

src/authorizations/apis/index.ts
```diff
@@ -29,9 +30,23 @@ export const getAuth0Config = async (
...
     const data = await response.json()
+
     return data
   } catch (error) {
     console.error(error)
     throw error
   }
 }
+
+export const getConnection = async (email: string): Promise<string> => {
```
Asalem1, line 41: since we're async awaiting i don't believe we'll be returning a promise. We're also potentially returning an error

```suggestion
export const getConnection = async (email: string): string | Error => {
```

src/localStorage.ts
```diff
@@ -39,6 +40,7 @@ export const setToLocalStorage = (prop: string, value: any): void => {
 
 export const getFromLocalStorage = (prop: string): any => {
   try {
+    console.log({prop})
```
Asalem1, line 43: same as above

Asalem1 (approved):

src/localStorage.ts
subirjolly: ma bad!

src/localStorage.ts
subirjolly: ma bad!

src/authorizations/apis/index.ts
subirjolly: It does return promise but I will add error
`Promise<string | Error>`

src/authorizations/apis/index.ts
subirjolly: Updating comment

src/authorizations/apis/index.ts
Asalem1: I'm confused. Aren't you awaiting the response? That promise should be resolved if that's the case

Asalem1: Turns out i don't know anything. You're totally right Subir, my bad

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


genehynson (approved):

src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -178,11 +180,25 @@ class LoginPageContents extends PureComponent<Props> {
...
+        if (!!connection) {
+          return this.auth0.authorize({connection})
+        }
+      } catch (e) {
+        const emailError = `${getErrorMessage(
+          e
+        )}. If this issue persists, please contact support@influxdata.com`
+        return this.setState({emailError})
+      }
+    }
```
randycoulman, line 199: I know this is merged, but on the Quartz side, we ignore errors in this API call and just fall through to the normal login process. We're seeing that the Auth0 API calls fail from time-to-time and we didn't want to break login for the 99.9% of users that aren't using SSO. Up to you if you want to follow that pattern or not.

PR #1794, opened 2021-06-23
chore: split long tests

### LINKED ISSUE #1788: Split long running cypress specs into multiple files
Currently `shared/dashboardView.test.ts` and `shared/explorer.test.ts` take over 2x as long to complete as our other spec files. We should split these into multiple spec files and improve the performance of our CI.

This will require 
- splitting the tests out in a way that makes sense
- ~updating k8s-idpe's hardcoded list of tests to include the new files (I think this still exists)~ It doesn't exist anymore
- I think that's it?

image

genehynson:
Closes #1788 

Moves some of the `shared/dashboardView.test.ts` tests into a new file `shared/dashboardRefresh.test.ts`.

Moves some of the `shared/explorer.test.ts` tests into two new files: `shared/explorerVisualization.test.ts` and `shared/simpleTable.test.ts`.

This should improve the performance of our pipeline and enable us to take advantage of parallelization. 

This PR also removes the `^` from our cypress dependency, fixing us to version 7.5.0


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


drdelambre:
This goes against cypress's best practice advice last I checked, but parallization will hide that until we run out of test runner instances. Something about most time spent starting up chrome and not actually running tests

genehynson:
> This goes against cypress's best practice advice last I checked, but parallization will hide that until we run out of test runner instances. Something about most time spent starting up chrome and not actually running tests

You are probably right - I don't think Cypress would recommend creating a bunch of spec files with just a couple tests. But I messed around with the cypress dashboard and it seemed pretty obivous to me that splitting the `dashboardView` and `explorer` tests would provide some benefit considering they take over twice as long to complete as the rest of our tests do running on 10 machines. My goal was to just split the specs in half, so hopefully we can have all the specs finish in around the same amount of time. 

image

ShmuelLotman (approved):

PR #1797, opened 2021-06-24
fix: added guarding for async rendering of the usage dropdown data

Asalem1:
this PR adds some checks prior to getting the usage data to render the dropdown since the code is async but isn't dependent upon the usageVectors for rendering. This caused the usage page to break so now we're adding to checks to ensure that something exists prior to rendering it. This is the only place that should be affected since it the other places that make use of the GraphTypeSwitcher are:

- Only being generated when the data is mapped (so it won't generated without any data)
- Has the usageVector hard-coded so there is not concern about async code

### DIFF
--- a/src/usage/UsageToday.tsx
+++ b/src/usage/UsageToday.tsx
@@ -27,14 +27,25 @@ const UsageToday: FC = () => {
 
   const getUsageSparkline = () => {
     const usage = usageVectors.find(vector => selectedUsage === vector.name)
-
-    return (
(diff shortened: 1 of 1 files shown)


ShmuelLotman (approved):

subirjolly (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:

PR #1809, opened 2021-06-25
chore: use notebooks routes

genehynson:
Use `/api/v2private/notebooks` routes instead of `/api/v2private/flows` routes


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


package.json
```diff
@@ -53,16 +53,15 @@
     "generate": "export SHA=f4424fc69ad29e31110d711310df644a57fee314 && export REMOTE=https://raw.githubusercontent.com/influxdata/openapi/${SHA}/ && yarn generate-meta",
     "generate-local": "export REMOTE=../openapi/ && yarn generate-meta",
     "generate-meta": "if [ -z \"${CLOUD_URL}\" ]; then yarn generate-meta-oss; else yarn generate-meta-cloud; fi",
-    "generate-meta-oss": "yarn oss-api && yarn notebooks && yarn unity && yarn flows && yarn managedFunctions && yarn annotations-oss",
-    "generate-meta-cloud": "yarn cloud-api && yarn notebooks && yarn unity && yarn flows && yarn managedFunctions && yarn annotations",
```
drdelambre, line 57: lol what a weird diff log

drdelambre (commented):
does flowd still exist? curious about the migration pattern. like why isn't this behind a feature flag?

package.json
genehynson: lol right?? Had to read through it a couple times to make sure the change actually makes sense 😆

genehynson:
> does flowd still exist? curious about the migration pattern. like why isn't this behind a feature flag?

Flowd is dead. The /flows routes still exist behind notebooksd. I could do a feature flag...but it's so low risk I'm not sure it's worth it. In notebooksd, the /notebooks routes have an identical implementation to the /flows ones. Only the path is changing. But if you'd prefer, I can add a feature flag in

src/flows/context/api.tsx
```diff
@@ -1,28 +1,27 @@
...
-let stagedFlow: PatchFlowsOrgsFlowParams = DEFAULT_API_FLOW
+let stagedFlow: PatchNotebookParams = DEFAULT_API_FLOW
 let reportDecayTimeout = null
 let reportMaxTimeout = null
 
 const REPORT_DECAY = 500 // number of miliseconds to wait after last event before sending
 const REPORT_MAX_WAIT = 5000 // max number of miliseconds to wait between sends
 
-export const pooledUpdateAPI = (flow: PatchFlowsOrgsFlowParams) => {
+export const pooledUpdateAPI = (flow: PatchNotebookParams) => {
```
subirjolly, line 24: NIT: If we are getting away from flow(s) then shouldn't we rename these vars and CONSTANTS to notebook(s)?

genehynson: This is absolutely a fair point. But it's one of those things that if I start going down this train, there is soooooo much to refactor and rename 😭 I think it's an exercise we should do, but I think I'd like to do that in a separate PR.

drdelambre: that was the same thinking that changed them all from notebooks to flows in the first place and broke things along the way. ROI seems... low. maybe i'm misunderstanding

drdelambre (approved):
if the implementations are exactly the same, then no need for a flag. just wanted to make sure there wasn't the chance of _also_ having a schema change or something hidden behind the change

genehynson:
> if the implementations are exactly the same, then no need for a flag. just wanted to make sure there wasn't the chance of _also_ having a schema change or something hidden behind the change

In case anyone is curious:
- `/flows` routes: https://github.com/influxdata/idpe/blob/master/sc-frontend/notebooksd/src/routes/flows.ts
- `/notebooks` routes: https://github.com/influxdata/idpe/blob/master/sc-frontend/notebooksd/src/routes/notebooks.ts
- both use the `stream` controller: https://github.com/influxdata/idpe/blob/master/sc-frontend/notebooksd/src/controller/stream.ts

And fwiw I've had this change behind a flag for awhile now (`notebooksApi`) and have been testing it in prod - works fine :)

PR #1811, opened 2021-06-25
chore: update openapi sha

glinton:
Applies sha from influxdata/openapi#154

Tests the changes from above mentioned pr


genehynson:
we're way ahead of this sha so I'm going to assume it's alright to close this

PR #1819, opened 2021-06-25
feat: changed moment to new date formatter in TokenRow.tsx

### LINKED ISSUE #1768: Replace Moment: replace a single instance of moment(date).format() with our new formatter
### Background
Depends on #1767

This is the implementation portion of creating our new formatting library

- [ ] [TaskRunsRow](https://github.com/influxdata/ui/blob/master/src/tasks/components/TaskRunsRow.tsx#L84)
- [ ] [RunLogRow](https://github.com/influxdata/ui/blob/master/src/tasks/components/RunLogRow.tsx#L44)
- [ ] [TokenRow](https://github.com/influxdata/ui/blob/master/src/authorizations/components/TokenRow.tsx#L62)

### Acceptance Criteria
- an instance of using `moment` to format at date in the codebase (pick the easiest one) is converted to use our new formatter

mtian725:
Closes #1768

Replaced the moment date formatter with the new date formatter in src/authorizations/components/TokenRow.tsx.

#### Before:
Old

#### After:
New

Minute marker is different between images. However, this is most likely due to an issue with how moment is being used or a bug within moment. The image below is from my personal account on the deployed product. It would be extremely unlikely for for to have made all these tokens at minute 6.
image

### DIFF
--- a/src/authorizations/components/TokenRow.tsx
+++ b/src/authorizations/components/TokenRow.tsx
@@ -24,10 +24,8 @@ import {Context} from 'src/clockface'
 
 // Types
 import {Authorization} from 'src/types'
-import {
-  DEFAULT_TOKEN_DESCRIPTION,
-  DEFAULT_TIME_FORMAT
-} from 'src/dashboards/constants'
+import {DEFAULT_TOKEN_DESCRIPTION} from 'src/dashboards/constants'
+import {FORMAT_OPTIONS} from 'src/visualization/utils/timeFormat'
 
 interface OwnProps {
   auth: Authorization
@@ -44,8 +42,9 @@ class TokenRow extends PureComponent<Props> {
     const {auth} = this.props
     const labelText = this.isTokenEnabled ? 'Active' : 'Inactive'
 
+    console.log(FORMAT_OPTIONS[0].text)
     const date = new Date(auth.createdAt)
-    const formatter = createDateTimeFormatter(DEFAULT_TIME_FORMAT)
+    const formatter = createDateTimeFormatter(FORMAT_OPTIONS[0].text)
     const updated = formatter.format(date)
 
     return (
--- a/src/authorizations/components/TokenRow.tsx
+++ b/src/authorizations/components/TokenRow.tsx
@@ -1,7 +1,7 @@
 // Libraries
 import React, {PureComponent} from 'react'
 import {connect, ConnectedProps} from 'react-redux'
-import moment from 'moment'
+import {createDateTimeFormatter} from 'src/utils/datetime/formatters'
 
 // Actions
 import {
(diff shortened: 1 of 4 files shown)


src/authorizations/components/TokenRow.tsx
```diff
@@ -25,8 +25,8 @@ import {Context} from 'src/clockface'
 // Types
 import {Authorization} from 'src/types'
 import {
-  DEFAULT_TOKEN_DESCRIPTION,
-  UPDATED_AT_TIME_FORMAT,
+  DEFAULT_TOKEN_DESCRIPTION, 
+  UPDATED_AT_TIME_FORMAT
```
TCL735, line 29: Was this an accidental change or is there something not set correctly in your code editor?

TCL735 (changes requested):

src/authorizations/components/TokenRow.tsx
```diff
@@ -43,6 +43,11 @@ class TokenRow extends PureComponent<Props> {
     const {description} = this.props.auth
     const {auth} = this.props
     const labelText = this.isTokenEnabled ? 'Active' : 'Inactive'
+
+    const date = new Date(auth.createdAt)
+    const formatter = createDateTimeFormatter(UPDATED_AT_TIME_FORMAT)
```
hoorayimhelping, line 48: let's move this out of the render method. since this uses a format that won't change based on user input, we can call this once, before the component is created, to avoid recreating essentially a static formatter on every render loop.

so something like:

```ts
const formatter = createDateTimeFormatter(UPDATED_AT_TIME_FORMAT)
class TokenRow extends PureComponent<Props> {
```

src/authorizations/components/TokenRow.tsx
```diff
@@ -59,7 +64,7 @@ class TokenRow extends PureComponent<Props> {
           {[
             <React.Fragment key={auth.id}>
               Created at:{' '}
-              {moment(auth.createdAt).format(UPDATED_AT_TIME_FORMAT)}
+              {updated}
```
hoorayimhelping, line 67: `updated` is kind of a weird name for this date, especially since it represents `created_at` which is often contrasted with `updated_at`

personally, i think it's fine to just do
```ts
{formatter.format(date)}
```

src/authorizations/components/TokenRow.tsx
hoorayimhelping: yeah, our anal retentive linter is going to complain about this. we'll need to set up your text editor to do this for you.

in the mean time, you can run ```yarn prettier:fix``` locally to fix these

hoorayimhelping (changes requested):

TCL735 (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:
best-practices | 98 | :green_circle:


hoorayimhelping (approved):

PR #1821, opened 2021-06-28
fix: double login issue

Asalem1:
Closes the API issue for 10830 double login

This PR _potentially_ resolves the issue of a user having a stale config state when trying to login to the app. It seems like what was happening was that a user was still referencing their original auth0 config when making a request to the API, but the session for that client config was only valid for ten minutes, so the user gets redirected back to the login page and goes through the entire process again. 

The reason why this only **potentially** resolves the issue is that we can't currently test this out locally since it requires being able to login directly from the `/login` page, which  _**still**_ isn't something that's currently possible :face_palm:.

What this PR aims to achieve is twofold:

- Set an interval to refresh the session state once it's expired
- Clear out the existing auth0 config whenever the component is unmounted


GeorgeMac (approved):
This makes sense to me from what I understand of the problem.

genehynson (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:

PR #1835, opened 2021-06-28
chore:  annotation form validation

### LINKED ISSUE #1751: annotations: overlay for creating/editing
* don't close overlays if there was a validation error
* validate that the 'stop' time is after the 'start' time; show the validation error in the overlay without closing

jrenee42:
Closes #1751 

- adds validation so that end time is after start time
- dates cannot be in the future
- if the date is empty or in the wrong format, then cannot 'close' and save/create an annotation either
- if a range annotation, dates must be different

### DIFF
--- a/src/annotations/components/annotationForm/AnnotationForm.tsx
+++ b/src/annotations/components/annotationForm/AnnotationForm.tsx
@@ -57,10 +57,17 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
   const [summary, setSummary] = useState(props.summary)
   const [annotationType, setAnnotationType] = useState(props.type)
 
+  // since we start with default times via the mouse, they are valid at first
+  // (until the user edits it)
+  const [endTimeValid, setEndTimeValid] = useState(true)
+  const [startTimeValid, setStartTimeValid] = useState(true)
+
   const dispatch = useDispatch()
 
   const isValidAnnotationForm = (): boolean => {
-    const isValidPointAnnotation = Boolean(summary?.length && startTime)
+    const isValidPointAnnotation = Boolean(
+      summary?.length && startTime && startTimeValid && isStartTimeValid()
+    )
 
     // not checking if start <= end right now
     // initially, the times are numbers, and then if the user manually edits them then
@@ -68,7 +75,9 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
     // plus, the backend checks if the startTime is before or equals the endTime
     // so, letting the backend do that check for now.
     if (annotationType === 'range') {
-      return Boolean(isValidPointAnnotation && endTime && isEndTimeValid())
+      return Boolean(
+        isValidPointAnnotation && endTime && endTimeValid && isEndTimeValid()
+      )
     }
     return isValidPointAnnotation
   }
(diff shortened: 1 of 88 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/annotations/components/annotationForm/AnnotationTimeInput.tsx
```diff
@@ -43,7 +45,28 @@ export const AnnotationTimeInput: FC<Props> = (props: Props) => {
...
+    if (timeValue === null) {
+      return '' // it's not filled in yet...don't show an error (parent takes care of required fields)
+    }
+
+    if (!isValidTimeFormat(timeValue)) {
+      return `Format must be ${timeFormat}`
+    }
+
+    // it is valid
+    return null
```
hoorayimhelping, line 69: shouldn't this return `''` and not `null`?

cypress/e2e/cloud/annotations.test.ts
```diff
@@ -552,6 +558,197 @@ describe('The Annotations UI functionality', () => {
       cy.getByTestID('giraffe-annotation-tooltip').contains('im a hippopotamus')
     })
 
+    it('does not allow the form to be submitted if there is a validation error', () => {
```
hoorayimhelping, line 561: this is all client side validation, right? do we need an e2e test for this behavior? it seems more suited to a unit or integration test. you don't need to block this PR on moving this to an integration test, but I think we should move in that direction and try to avoid writing e2e tests if we can write other tests that give us confidence

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -129,11 +145,80 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
...
+     * if point annotation, do nothing
+     *
+     * if range:
+     * making sure they are not the same, and also that 'end' is after 'start'
+     */
+
+    const start = moment(startTime)
+    const end = moment(endTime)
+
+    if (end.isSame(start)) {
```
hoorayimhelping, line 188: another instance to add to the above ticket to handle moment date math https://github.com/influxdata/ui/issues/1844

hoorayimhelping (commented):
good work, thanks! a couple of questions, comments.

src/annotations/components/annotationForm/AnnotationForm.tsx
jrenee42: added

src/annotations/components/annotationForm/AnnotationTimeInput.tsx
jrenee42: added

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -129,11 +145,80 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
...
   }
 
+  // TODO:  distinguish between editing and creating for this cancel event
   const handleCancel = () => {
     event('dashboards.annotations.create_annotation.cancel')
     props.onClose()
   }
 
+  const isTimeInFuture = timeToCheck => {
+    return moment(timeToCheck).isAfter(moment())
```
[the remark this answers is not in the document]
jrenee42: added

src/annotations/components/annotationForm/AnnotationTimeInput.tsx
jrenee42: it doesn't really matter; but I changed it.  also realized that the parent doesn't actually check for 'required' when the date is null, so fixed that and adjusted the tests too.

hoorayimhelping (approved):

jrenee42:
redoing this PR with unit tests and with a fresh checkout of master

Issue #1521, opened 2021-05-20
Cannot create a cron task from the UI (javascript error)


christopheblin:
__Steps to reproduce:__

1. Create a bucket with the name `import` in it.
2. Attempt to create a task as described in the following:

Whenever I try to create a cron task from a working query, it fails with the error

    Failed to create new task: Cannot read property 'map' of null

This behavior happens from the "explore > save as" as seen in the following video but it also happens if I try to create a new task directly from the "tasks" menu

The cron is `0 * * * *` (which mean hourly if I am correct) and the offset is `2m` (to account for late data)

https://user-images.githubusercontent.com/11044877/118929226-87ab5780-b944-11eb-938f-bc577b040c55.mov

I try to follow (more or less) the tutorial at https://www.influxdata.com/blog/downsampling-influxdb-v2-0/

__Expected behavior:__
Have a clear error message if it is an error on my side

__Actual behavior:__
Cannot create a task

christopheblin:
Any news on that bug ?

I cannot create a task, it seems pretty severe to me but it is only classified sev-4 ...

Is there a workaround ?

kristinarobinson:
This is working for us in production, can you send us the query that you are using?  The video is a little blurry, so we cannot see exactly what you are sending.  Perhaps there is something in the query you are sending that is causing an issue.

christopheblin:
any request fails

let's say I'd like to keep a single measurement 

```
from(bucket: "rt")
|> range(start: -1h)
|> filter(fn: (r) => r._measurement == "gtfs-rt-import")
|> to(bucket: "rt-downsampled")
```

there is a JS error 

```
TypeError: Cannot read property 'map' of null
    at b (insertPreambleInScript.ts:32)
    at async thunks.ts:429
```

Full stacktrace

```

  | (anonymous) | @ | thunks.ts:441
-- | -- | -- | --
  | async function (async) |   |  
  | (anonymous) | @ | thunks.ts:429
  | (anonymous) | @ | index.js:8
  | (anonymous) | @ | redux.js:477
  | handleSave | @ | TaskPage.tsx:128
  | u | @ | react-dom.production.min.js:14
  | p | @ | react-dom.production.min.js:15
  | (anonymous) | @ | react-dom.production.min.js:15
  | x | @ | react-dom.production.min.js:17
  | M | @ | react-dom.production.min.js:18
  | E | @ | react-dom.production.min.js:18
  | I | @ | react-dom.production.min.js:21
[...]

PR #1839, opened 2021-06-28
fix: handle creating tasks that work with buckets with import in its name

hoorayimhelping:
Closes https://github.com/influxdata/ui/issues/1521

### Steps to reproduce:
1. Create a bucket with `import` in the name (`devbucket-import` is what I used)
1. Try to create a task that references the bucket with import in the name
    ```flux
    from(bucket: "devbucket-import")
        |> range(start: v.timeRangeStart)
        |> filter(fn: (r) => r._measurement == "system")
        |> filter(fn: (r) => r._field == "load1" or r._field == "load5" or r._field == "load15")
        |> to(bucket: "devbucket-downsampled", org: "dev")
    ```
1. Apply this fix and then retry the above

### DIFF
--- a/src/shared/utils/insertPreambleInScript.ts
+++ b/src/shared/utils/insertPreambleInScript.ts
@@ -26,7 +26,7 @@ export const insertPreambleInScript = async (
 
   const ast = resp.data.ast as Package
 
-  const imports: ImportDeclaration[] = get(ast, 'files.0.imports', [])
+  const imports: ImportDeclaration[] = ast?.files[0]?.imports ?? []
   const body: Statement[] = get(ast, 'files.0.body', [])
 
   const importsText = imports.map(d => d.location.source).join('\n')


src/shared/utils/insertPreambleInScript.ts
```diff
@@ -26,7 +26,7 @@ export const insertPreambleInScript = async (
 
   const ast = resp.data.ast as Package
 
-  const imports: ImportDeclaration[] = get(ast, 'files.0.imports', [])
+  const imports: ImportDeclaration[] = ast?.files[0]?.imports ?? []
```
hoorayimhelping, line 29: `ast.files.0.imports` comes back as `null`. `_.get` returns the default value only if the value is `undefined`. Since it was null, this was failing and `imports` was being set to `null` before being mapped over on line 32. This fix is equivalent to:

```ts
get(ast, 'files.0.imports') || []
```

genehynson (approved):

ui-ci-bot-influx:
Metric | Score | Level
performance | 66 | :yellow_circle:
seo | 90 | :green_circle:
pwa | null | :hankey:

PR #1852, opened 2021-06-29
feat: notifications on login

ShmuelLotman:
Closes #10955

This utilizes the new usage API to fetch the number of write limit hits a user has had in the past hour (time subject to change) and if it surpasses a given limit (currently 100) it will notify the user on login. 

This also makes minor changes to the existing notification structure for write limits


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/cloud/components/RateLimitAlert.tsx
```diff
@@ -45,51 +62,74 @@ const RateLimitAlert: FC<Props> = ({
...
+              color={ComponentColor.Primary}
+              size={ComponentSize.Small}
+              onClick={appearOverlay}
+              text="Request Write Limit Increase"
+            />
+          )
+        )
+      }
+    }
+  }, [showUpgrade, status])
```
Asalem1, line 106: should we listen to changes in the resources?

src/cloud/components/RateLimitAlert.tsx
```diff
@@ -29,14 +31,29 @@ import {CLOUD} from 'src/shared/constants'
...
+
+import {notify} from 'src/shared/actions/notifications'
+import {writeLimitReached} from 'src/shared/copy/notifications'
+
+// Selectors
+import {shouldShowUpgradeButton} from 'src/me/selectors'
+import {dismissOverlay, showOverlay} from 'src/overlays/actions/overlays'
+import {UpgradeContent} from 'src/cloud/components/RateLimitAlertContent'
+
+import './RateLimitAlert.scss'
```
Asalem1, line 43: not sure if this would need to be included in chronograf.scss?

src/cloud/components/onboarding/useGetUserStatus.ts
```diff
@@ -83,29 +83,67 @@ export const getUserStatus = (table: Table): USER_PILOT_USER_STATUS[] => {
...
+  orgID: string,
+  queryLength?: string
+): Promise<Table> => {
+  let csvToParse = ''
+  if (getOrgsUsage) {
+    if (isFlagEnabled('newUsageAPI')) {
+      const usage = await getOrgsUsage({
+        orgID,
+        query: {
+          start: queryLength ?? '-30d',
```
Asalem1, line 96: dang, that's a long default:

1. How intense is the payload that's returned from a 30d query
2. Should it just be `30d`? I ask because that's the way quartz's usage API is constructed, so just checking to be sure

src/cloud/components/onboarding/useGetUserStatus.ts
```diff
@@ -83,29 +83,67 @@ export const getUserStatus = (table: Table): USER_PILOT_USER_STATUS[] => {
...
+        },
+      })
+      if (usage?.status === 200) {
+        console.warn(`Usage: ${usage}`)
+        csvToParse = usage.data?.trim().replace(/\r\n/g, '\n')
```
Asalem1, line 101: hello old friend

src/cloud/components/onboarding/useGetUserStatus.ts
```diff
@@ -83,29 +83,67 @@ export const getUserStatus = (table: Table): USER_PILOT_USER_STATUS[] => {
...
+    return table
+  }
+}
+
+export const getUserWriteLimitHits = async (
+  orgID: string,
+  queryLength?: string
+): Promise<number> => {
+  try {
+    const table = await queryUsage(orgID, queryLength ?? '-1h')
```
Asalem1, line 118: if this the only place `queryUsage` is being called? If so it seems like the default queryLength can just take place in the query if the `queryLength` is undefined right? You've got `-1h` and `-30d`, just a thought

src/me/containers/MePage.tsx
```diff
@@ -21,19 +23,74 @@ import AlertsActivity from 'src/me/components/AlertsActivity'
...
+const QUERY_WRITE_LIMIT_HITS = 100
+
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = ReduxProps
 
+let hasCalled = false // We only want to show the write limit notification once
+
 @ErrorHandling
 export class MePage extends PureComponent<Props> {
+  async componentDidMount() {
```
Asalem1, line 52: 1. Should this be flagged?
2. re: the comment about this only wanting to be called once - what if the user hits their limit after this mounts? Also, why are we doing the same thing here as we are in the WriteLimits component?

src/shared/components/notifications/Notifications.tsx
```diff
@@ -43,11 +44,12 @@ const Notifications: FC = () => {
               id={id}
               icon={icon}
               duration={duration}
-              size={ComponentSize.ExtraSmall}
+              size={ComponentSize.Small}
               gradient={gradient}
               onTimeout={handleDismiss}
               onDismiss={handleDismiss}
               testID={`notification-${style}`}
+              style={{maxWidth: '600px', alignItems: 'center'}}
```
Asalem1, line 52: Not sure if we want to change notifications all over the app do we?

Asalem1 (commented):
Overall solid, but I'm not sure if things are in the right place if that makes sense? It seems like we want to trigger a notification on the app when a specific threshold is met, but we're setting that notification to be triggered from two different places with two different mechanisms for listening for those? I'd love @drdelambre's thoughts

Asalem1:
Also, should the notification be with a `if CLOUD` or something?

src/cloud/components/RateLimitAlert.tsx
ShmuelLotman: Listening to changes on the resources is unnecessary, it updates anyways by virtue of the referential inequality (i think), ive tested locally and it seems to update correctly

ShmuelLotman:
@Asalem1 the reason there are two different triggers is because these are two different sorts of notifications - one is the write limit for when its hit once, at the time its hit, and one is for x amount of times in the past hour, which were separate requirements and have separate conditions for rendering and separate messaging.

src/shared/components/notifications/Notifications.tsx
ShmuelLotman: This was requested by @mavarius

src/me/containers/MePage.tsx
ShmuelLotman: See comment below, this is a different implementation with a different use case

src/cloud/components/onboarding/useGetUserStatus.ts
ShmuelLotman: 2. it shouldnt
1. We are going to test that shortly, this endpoint needs to be in PROD to be able to work

src/shared/components/notifications/Notifications.tsx
Asalem1: K, @mavarius we cool to update all the notifications? If that's the case, set the styling for the `Notifications` in clockface so that's the default? If not, then should we make this a specific UI-default with a className setting these properties?

Asalem1:
Sounds good re: the two different triggers. Is this just for cloud?

src/cloud/components/onboarding/useGetUserStatus.ts
Asalem1: classic

src/cloud/components/RateLimitAlert.tsx
ShmuelLotman: leave my modules alone! ;)

src/cloud/components/onboarding/useGetUserStatus.ts
ShmuelLotman: Yeah, this makes this very difficult to implement because im essentially going blind. The plan is to work with @mavarius to flip the feature flag on for just him in PROD, then see how it works and slowly expand it out from there.

src/cloud/components/onboarding/useGetUserStatus.ts
ShmuelLotman: at _this_ point yes, but i wanted this to be a bit more extendable so ive passed a param with a default

ShmuelLotman:
This is just for cloud, and that is behind the flag in useGetUserStatus.ts

Asalem1 (approved):
Just want to know if we should update clockface to set the default notification size there or in a classname in the app. Other then that looks good to me

src/cloud/components/onboarding/EngagementLink.tsx
```diff
@@ -18,16 +18,20 @@ const EngagementLink: FC = () => {
...
+    }
+  }, [])
+
+  useEffect(() => {
+    if (userpilot) {
       sendToUserPilot()
       userpilot.reload()
     }
-  }, [pathname, org, me, usageDataStates])
+  }, [pathname, org, me])
```
Asalem1, line 34: for some reason my brain is telling me useCallback goes with useEffect....

src/cloud/components/onboarding/useGetUserStatus.ts
```diff
@@ -1,20 +1,16 @@
...
-let getOrgsUsage = null
+let getUsage = null
 
 if (CLOUD) {
-  getOrgsUsage = require('src/client').getOrgsUsage
+  getUsage = require('src/client/unityRoutes').getUsage
 } else {
-  getOrgsUsage = null
+  getUsage = null
 }
```
Asalem1, line 14: I don't think we need the else right?

src/cloud/components/onboarding/useGetUserStatus.ts
```diff
@@ -80,48 +62,109 @@ export const getUserStatus = (table: Table): USER_PILOT_USER_STATUS[] => {
...
+    event('cloud.onboarding.set_user_status_failure', {
+      context: JSON.stringify(err),
+    })
+  }
 
   return {usageDataStates}
 }
 
-export default useGetUserStatus
+export default handleGetUserStatus
```
Asalem1, line 170: so much less hack, love it! Quick thought though, some of those will return A LOT of data with the `30d`. Do we need to query that long?

Asalem1 (approved):
Couple things about useEffect + useCallback, not sure if it's needed but 🤷🏽 
Also, do we have to query 30d worth of data for this or can we shorten that, the flux query might take a bit to resolve so not sure what the UX will be like. With that in mind, do we want to promise all the queries or do we only care if one of them returns a specific value? If that's the case then it might be worth a different promise-based approach where we resolve them individually

ShmuelLotman:
To answer your last question first, we need all of them to accurately ascertain user status, so we need to wait for them all. 

Regarding whether we need 30 days, thats something that I can ask @bpalani when he gets back, but as of now that was the requested timeframe.

src/cloud/components/onboarding/useGetUserStatus.ts
ShmuelLotman: see response below

src/cloud/components/onboarding/useGetUserStatus.ts
ShmuelLotman: true

src/cloud/components/onboarding/EngagementLink.tsx
ShmuelLotman: dont think its needed in this case - but i did add a memo() around the component bc it is rendered as a sibling of large parts of the main app and does not need to be rerendered every single time.

Issue #1796, opened 2021-06-24
Reintegrate e2e tests for unification work and add more tests based on different provisioning of user


Asalem1:
This issue is a follow-up to the work that's been done to deploy a mock quartz API for testing purposes so that we can reintegrate tests our e2e tests within our pipeline. 

With the new API we can also provision users differently so we can add more use cases for users that are free vs payg, operators, in a beta region, and have usage data vs have no usage data

PR #1861, opened 2021-06-30
feat: add e2e tests to unification work via quartz mock

Asalem1:
This PR is related to #1796 . 

Now that we have an image of the quartz-mock API being served in quay we can reintegrate our e2e tests to ensure that they'll pass in our CI/CD pipeline.

This PR aims to reintegrate tests for the users and usage page. The users page is a basic POC for CRUD operations against the quartz-mock API, while the usage page tests is a bit more of a blueprint for what the billing page might look like so I'd love thoughts on the implementation.

At a high level, I've broken the usage page into two different tests to test two different user states:

1. A free user without data
2. A PAYG user with data

Theses states are toggled by adding a new cy.quartzProvision call that can set specific flags on the API without needing to do much else except to expect different results from the API.

I had originally tried to nest the `quartzProvision` call on an individual `it` level, but that seemed to beget flakiness so I took it to mean that it was the incorrect implementation. I'm not sure if it's worth just splitting this up into two different test files (i.e. UsageData.test.ts & UsageNoData.test.ts)


src/App.tsx
```diff
@@ -49,7 +49,7 @@ const App: FC = () => {
       load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
     }
     setAutoFreeze(false)
-  }, [RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL])
+  }, [])
```
Asalem1, line 52: these were unnecessary in the useEffect since they should never change (linter was throwing a warning)

src/operator/api/index.ts
```diff
@@ -178,8 +178,8 @@ export const getOrgById = (_id: string): ReturnType<typeof getOperatorOrg> => {
     account: {
       type: 'pay_as_you_go',
       balance: 10,
-      id: 'pay123',
-      email: 'paid@account.com',
+      email: 'payg@influxdata.com',
+      id: '123',
```
Asalem1, line 182: changes in this file reflect contract definition changes with the SHA bump

src/shared/containers/GetFlags.tsx
```diff
@@ -17,11 +17,12 @@ import {getFlags} from 'src/shared/thunks/flags'
...
-type ReduxProps = ConnectedProps<typeof connector>
-type Props = ReduxProps
-
-const GetFlags: FC<Props> = ({status, flags}) => {
+const GetFlags: FC = () => {
   const dispatch = useDispatch()
+  const flags = useSelector(activeFlags)
+  const status = useSelector(
+    (state: AppState) => state.flags.status || RemoteDataState.NotStarted
+  )
```
Asalem1, line 25: So in the process of trying to figure out why the heck everything was refreshing, I started to poke around here to see if it was the `mstp` (it wasn't), but decided to keep this here since it cleans up the previous implementation which was a mix of two implementations (connect & redux hooks)

src/shared/containers/GetFlags.tsx
```diff
@@ -36,7 +37,7 @@ const GetFlags: FC<Props> = ({status, flags}) => {
         return prev
       }, {})
     )
-  }, [flags])
+  }, [Object.keys(flags).length]) // eslint-disable-line react-hooks/exhaustive-deps
```
Asalem1, line 40: doing comparisons on `flags` was triggering this to run a bunch because JS object comparisons

src/shared/containers/GetMe.tsx
```diff
@@ -39,7 +39,7 @@ class GetMe extends PureComponent<Props, State> {
     return (
       <SpinnerContainer loading={loading} spinnerComponent={<TechnoSpinner />}>
         <Switch>
-          <Route render={props => <GetFlags {...props} />} />
+          <Route component={GetFlags} />
```
Asalem1, line 42: related to the previous comment about the refactor

src/shared/containers/SetOrg.tsx
```diff
@@ -70,29 +70,21 @@ import {updateReportingContext} from 'src/cloud/utils/reporting'
...
-  history,
-}) => {
+const SetOrg: FC = () => {
   const [loading, setLoading] = useState(RemoteDataState.Loading)
   const dispatch = useDispatch()
+  const orgs = useSelector((state: AppState) =>
+    getAll<Organization>(state, ResourceType.Orgs)
+  )
+  const history = useHistory()
+  const {orgID} = useParams<{orgID: string}>()
```
Asalem1, line 86: related to the previous comment about the refactor

src/usage/context/usage.tsx
```diff
@@ -178,27 +207,32 @@ export const UsageProvider: FC<Props> = React.memo(({children}) => {
...
         billingStats,
+        billingStatsStatus,
         handleSetSelectedUsage,
         handleSetTimeRange,
         rateLimits,
+        rateLimitsStatus,
         selectedUsage,
         timeRange,
         usageStats,
+        usageStatsStatus,
```
Asalem1, line 235: added loading states to make things nicer

cypress/e2e/cloud/billing.test.ts
```diff
@@ -1,6 +1,5 @@
 import {Organization} from '../../../src/types'
 
-// Skipping this until we get the CI/CD pipeline worked out for the `/quartz/me` endpoint
 describe.skip('Billing Page', () => {
```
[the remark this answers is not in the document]
Asalem1: ya, I didn't realize I tossed the skip. I threw that back in there since i didn't wanna get the billing page tests on this PR

src/shared/containers/GetFlags.tsx
genehynson: would it be better to `useMemo` here?

Asalem1: 🤷🏽  what would the benefits of one over the other over both be?

src/shared/copy/notifications.ts
```diff
@@ -1345,7 +1345,7 @@ export const inviteFailed = (): Notification => ({
 
 export const invitationResentSuccessful = (): Notification => ({
   ...defaultSuccessNotification,
-  message: `Invitation Sent`,
+  message: `Invitation resent`,
```
genehynson, line 1348: I think it should be `re-sent`. `resent` has a different meaning 😅

Asalem1: ya, hilarious I saw that and thought, something doesn't look right. Will do

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/shared/containers/GetFlags.tsx
genehynson: I think `useMemo` would help if the same number of flags were present in the array, but the contents of the array had changed. In that case `useMemo` would trigger but checking the length would not. However, I'm not sure if that edge case is applicable here

drdelambre: possible to remove this? we want this to update the reporting context every time they change the VALUE of the flags too

Asalem1: Totally hear that. Does useEffect do checks against objects for strict equality?

genehynson: I was actually just talking with @ShmuelLotman about this yesterday. The deps array for hooks does not do any deep object comparison. So if you're trying to watch for changes against a non-primitive type, you'll need to do something like `useMemo` to prevent an infinite re-render. And that's because the array's memory reference will change all the time even though the values do not.

drdelambre: totally understand deep equality checks, but the reference to this _should_ be changing every time the values are updated (and not touched if they're not changed) through some sort of `flags = {...oldflags, ...newflags}` operation. if something is making that reference change all the time, we should alter the point of ingestion to... not... as the relationship is usually few:many between ingestion and consumption points in the app (few things should be setting this value.. many things should be reading this value)

Asalem1: kk, just ended up moving it back to `flags` and we can fix the events that retrigger this call some other time

drdelambre (approved):

genehynson (approved):

Issue #1714, opened 2021-06-15
SimpleTable: TypeError: Cannot read property '0' of undefined


russorat:
![image](https://user-images.githubusercontent.com/4805997/122088751-e684b500-cdba-11eb-9031-8bd7e13cd369.png)

```
from(bucket: "telegraf")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "cpu")
  |> filter(fn: (r) => r["_field"] == "usage_user")
  //|> aggregateWindow(every: v.windowPeriod, fn: mean, createEmpty: false)
  |> yield(name: "mean")
```

Data i was trying to render:
[2021-06-15-09-21_influxdb_data.csv.zip](https://github.com/influxdata/ui/files/6657162/2021-06-15-09-21_influxdb_data.csv.zip)

Asalem1:
@russorat I'm having a hard time reproducing this locally with the query listed in the screenshot and the query you wrote on the issue with the data you uploaded. Can you still reproduce this?

russorat:
are you using the `influx.set("simpleTable", true)` feature flag? 

i've been able to repro it with the following query:

```
import "experimental/csv"

csv.from(url: "https://gist.githubusercontent.com/russorat/861fc2aafe0ddc70d8a89b5a983cd8b4/raw/04f5376a87e571ce0bc9730a9cada7495a871141/results.csv")
  //|> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "cpu")
  |> filter(fn: (r) => r["_field"] == "usage_user")
  //|> aggregateWindow(every: v.windowPeriod, fn: mean, createEmpty: false)
  |> yield(name: "mean")
```

Asalem1:
I

genehynson:
I was able to reproduce this in tools yesterday. Going to take another look today to see if I can isolate it and fix it

PR #1868, opened 2021-06-30
fix: simple table 0 of undefined and pagination bugs

### LINKED ISSUE #1813: SimpleTable: Cannot read property '0' of undefined
to repro:
1. ensure you have the simpleTable flag enabled
2. Open data explorer, select raw data view, and open script editor
3. paste the following script
```
import "experimental"
import "experimental/geo"
import "experimental/csv"

csv.from(url: "https://raw.githubusercontent.com/influxdata/influxdb2-sample-data/master/usgs-earthquake-data/all_week-annotated.csv")
  |> filter(fn: (r) => r._measurement == "earthquake" and r._field == "depth")
  |> group(columns: ["_field"])
  |> limit(n: 10)
  |> keep(columns: ["_field", "_value"])
  |> group()
```

Run the query, see the results.

change the `limit(n: 10)` to `limit(n:20)` and run the script again. notice the error.

image

genehynson:
Closes #1813 and #1714

The original issues are resolved simply by changing the offset calculation to prevent the offset from becoming -1. However, this bug highlighted another issue with simpletable where the records from the `n-1` page were duplicated on the `nth` page. The solution to that was to set the `totalPages` variable whenever the result set changes or the height of the table changes. We cannot calculate the `totalPages` using `size` and `offset` because those numbers change as each page is rendered and not all pages show the same number of records.

I added unit and cypress tests

### DIFF
--- a/src/visualization/context/pagination.tsx
+++ b/src/visualization/context/pagination.tsx
@@ -1,27 +1,35 @@
 import React, {FC, createContext, useState, useCallback} from 'react'
-
+import {
+  calcOffset,
+  calcNextPageOffset,
+  calcPrevPageOffset,
+} from '../utils/paginationUtils'
 interface PaginationContextType {
   offset: number // the start index
   size: number // the size of the page
   total: number // the total number of entries
+  totalPages: number // the total number of pages
 
   next: () => void
   previous: () => void
 
   setSize: (size: number) => void
   setPage: (page: number) => void
+  setTotalPages: (totalPages: number) => void
 }
 
 const DEFAULT_CONTEXT: PaginationContextType = {
   offset: 0,
   size: 0,
   total: 0,
+  totalPages: 0,
 
   next: () => {},
   previous: () => {},
 
   setSize: (_size: number) => {},
   setPage: (_page: number) => {},
+  setTotalPages: (_totalPages: number) => {},
 }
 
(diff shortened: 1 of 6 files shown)


Asalem1 (approved):

src/visualization/context/pagination.tsx
```diff
@@ -53,7 +53,9 @@ export const PaginationProvider: FC<PaginationProviderProps> = ({
 
   const setPage = useCallback(
     (page: number) => {
-      setOffset(Math.min(Math.max(0, (page - 1) * size), total - size))
+      setOffset(
+        Math.min(Math.max(0, (page - 1) * size), Math.max(0, total - size))
```
subirjolly, line 57: Quick unit test!
example:
page: 1
size: 2
total: 11

Now see if offset is correct.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/visualization/context/pagination.tsx
genehynson: good catch. This ended up being a pretty involved fix but take a look at what I ended up doing and let me know your thoughts.

subirjolly: will approve as soon as you push up your fix again.

subirjolly (approved):

PR #1890, opened 2021-07-02
fix(notebooks): run preview only when presentation mode is on

genehynson:
Closes https://github.com/influxdata/idpe/issues/10960

If presentation mode is on, run preview when entering the notebook and display the visualizations.

If presentation mode is off, do not run preview when entering the notebook.

Adds cypress tests for these two scenarios.

https://user-images.githubusercontent.com/6411855/124320574-fdeccd80-db30-11eb-9a0d-31be02d98f97.mov

### DIFF
--- a/src/flows/context/flow.list.tsx
+++ b/src/flows/context/flow.list.tsx
@@ -82,6 +82,7 @@ export function serialize(flow) {
       if (meta) {
         return {
           ...flow.data.byID[id],
+          id,
           title: meta.title,
           visible: meta.visible,
         }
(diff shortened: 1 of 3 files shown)


src/flows/context/flow.list.tsx
```diff
@@ -82,6 +82,7 @@ export function serialize(flow) {
       if (meta) {
         return {
           ...flow.data.byID[id],
+          id,
```
genehynson, line 85: the reason the visualizations weren't displaying after the initial query when the notebook is loaded is because we weren't persisting the pipe's ID. So on initial load, the pipe's ID is regenerated and the results from the query get lost because the old IDs no longer exist. Persisting the IDs solves this problem.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


joelfried (approved):
LGTM

PR #1918, opened 2021-07-07
feat: notebooks column panel

### LINKED ISSUE #662: Notebooks: Column mapping panel
Add a panel that can do these things to any of the columns:
- drop a column from the result -https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/built-in/transformations/drop/#columns
- rename a column - https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/built-in/transformations/rename/
- cast the type of a column into something else
This list shouldn't be considered exhaustive, we just want something that'll allow us to do explicit base level transformations against tabular data, go nuts

ShmuelLotman:
Closes #662 

Introduces a panel that allows for the handling of column visibility and naming, as well as future operations such as mapping and others, should we determine them to be necessary / useful. 

https://user-images.githubusercontent.com/29473622/124782296-bcca3400-df09-11eb-9879-3920ab553912.mov


src/flows/pipes/Columns/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      updatedTableKeys: {},
+    },
+    generateFlux: (pipe, create, append) => {
+      append(`__CURRENT_RESULT__ |> limit(n: 100)`)
+      if (!Object.values(pipe.updatedTableKeys).length) {
+        return
+      }
+
+      const toggleHide = Object.entries(
+        pipe.updatedTableKeys as {[_: string]: TableColumnKey}
```
Asalem1, line 19: why are we typecasting this?

src/flows/pipes/Columns/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+            a.push(`"${k}"`)
+          }
+          return a
+        } else {
+          return a
+        }
+      }, [])
+
+      const rename = Object.entries(
+        pipe.updatedTableKeys as {[_: string]: TableColumnKey}
```
Asalem1, line 34: same as above?

src/flows/pipes/Columns/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+        if (!v.visible) {
+          if (v.name !== k) {
+            a.push(`"${v.name}"`)
+          } else {
+            a.push(`"${k}"`)
+          }
+          return a
+        } else {
+          return a
+        }
```
Asalem1, line 30: nit:

flip it and reverse it so we don't have nested if statements:

```
if (v.visible) {
  return a
}
if (v.name !== k) {...}

```

src/flows/pipes/Columns/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+        create(`
+                __PREVIOUS_RESULT__
+                |> rename(columns: {${rename.join(', ')}})
+                `)
+      }
+
+      if (toggleHide.length) {
+        create(`__PREVIOUS_RESULT__
+               |> drop(columns: [${toggleHide.join(', ')}])`)
+      }
```
Asalem1, line 54: I think I'm a little confused as to what we're doing here. Is the goal of this to rename a column, then drop the column?

src/flows/pipes/Columns/styles.scss
```diff
@@ -0,0 +1,6 @@
+.columns-panel--grid {
+  width: 100%;
+  display: grid;
+  grid-template-columns: repeat(3, minmax(200px, 1fr));
+  grid-gap: 15px;
```
Asalem1, line 5: why a grid and not flex?

src/flows/pipes/Columns/view.tsx
```diff
@@ -0,0 +1,126 @@
...
+  }, [results?.parsed?.table])
+
+  useEffect(() => {
+    // const updatedKeys = Object.entries(tableColumnKeys).reduce((a, [k, v]) => {
+    //   if (v.name === k) {
+    //     return a
+    //   } else {
+    //     return {...a, [k]: {...v}}
+    //   }
+    // }, {})
```
Asalem1, line 48: do we need this?

src/flows/pipes/Columns/view.tsx
```diff
@@ -0,0 +1,126 @@
...
+                {[
+                  <React.Fragment key={v.name + Math.random()}>
+                    <FlexBox
+                      direction={FlexDirection.Row}
+                      alignItems={AlignItems.Center}
+                      margin={ComponentSize.Medium}
+                      stretchToFitWidth={true}
+                    >
+                      <SlideToggle
+                        onChange={() => handleUpdateVisible(k, !v.visible)}
```
Asalem1, line 111: if this is a toggler where the visibility is always the opposite of what the current state is, shouldn't we just make that a concern of the function and not pass in a parameter?

src/flows/pipes/Columns/view.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const handleUpdateVisible = (oldName: string, newVisibleStatus: boolean) => {
+    if (tableColumnKeys[oldName].visible === newVisibleStatus) {
+      return
+    }
+
+    setTableColumnKeys(prev => ({
+      ...prev,
+      [oldName]: {
+        ...tableColumnKeys[oldName],
+        visible: newVisibleStatus,
```
Asalem1, line 77: connected to the issue below, why not just use the previous state to set the visibility to the opposite of what it was?

```
visible: !prev[oldName].visible
```

or something like that

Asalem1 (approved):
Great job. Is there a gif?

Asalem1:
Large scale question - did we want this to be a panel in and of itself or should we make it a sidebar? It seems a bit ... weird to have a panel dedicated to the visualization of the following panel right?

juliajames12:
I agree with @Asalem1. Now that we have the side panel, I think it would be more intuitive to access the column mapping there. Kind of like we have the visualization settings of the side panel.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/flows/pipes/Columns/index.tsx
ShmuelLotman: No, its to be _able_ to do both. Thats why they are in separate conditionals. One for toggling visibility and one for changing name.

src/flows/pipes/Columns/styles.scss
ShmuelLotman: Bc Grid freaking rules

src/flows/pipes/Columns/index.tsx
ShmuelLotman: bc we need to have a type for the pipe.updatedTableKeys

src/flows/pipes/Columns/view.tsx
ShmuelLotman: It's not just about the toggling visibility, i also need the old name value that is passed in here to reference the correct column

drdelambre:
@Asalem1 i think you're missing that this panel isn't for modifying the visualization of a query's results, but for modifying the query itself. this allows the user to drop columns from one data source, rename any others, and write them off to another data source in their cleaned state. As this is an atomic transformation of the source query in the notebook, and not a frontend only configuration of a single visualization, it should definitely be defined as a panel.

drdelambre (changes requested):
you gotta make this look more inline with the app's aesthetic, this thing looks totally random and there is way too much whitespace contributing to... lemme know if you need help with that

src/flows/pipes/Columns/index.tsx
drdelambre: so this will only let you either rename a column or drop them i think. i'd expect this to work better:
```
if (rename.length || toggleHide.length) {
    let query = '__PREVIOUS_RESULT__'
    if (rename.length) {
        query += ` |> rename(columns: {${rename.join(', ')}})`
    }
    if (toggleHide.length) {
        query += ` |> drop(columns: [${toggleHide.join(', ')}])`
    }
    create(query)
}
```

drdelambre: and does the toggleHide part have to be `toggleHide.map(h => '"${h}"').join(', ')`?

src/flows/pipes/Columns/view.tsx
```diff
@@ -0,0 +1,117 @@
...
+      },
+    }))
+  }
+
+  if (!Object.values(tableColumnKeys)) {
+    return <h1>No Columns To Show. Preview To See Results.</h1>
```
drdelambre, line 74: dope

src/flows/pipes/Columns/view.tsx
drdelambre: i think ari is talking about letting the toggle chose the state in case the internal state gets out of sync with the view state between the two unrelated components. i totes agree. ie, should be:
```onChange={(state) => handleUpdateVisible(k, state)}```

File: src/flows/pipes/Columns/index.tsx (hunk 4 above)
Comment (drdelambre) [reply src/flows/pipes/Columns/index.tsx:54]: disregard the last part.. seems like you did that in the reduce

**Review by drdelambre — APPROVED** [review]
thanks for updating the design. double check that calling create twice works as expected before merging, because idk if that's guaranteed by the notebook model or a happy coincidence

File: src/flows/pipes/Columns/index.tsx (hunk 4 above)
Comment (ShmuelLotman) [reply src/flows/pipes/Columns/index.tsx:54]: Tested this and it allows you to do both

File: src/flows/pipes/Columns/index.tsx [hunk 10]
```diff
@@ -0,0 +1,57 @@
+import View, {TableColumnKey} from './view'
+
+export default register => {
+  register({
+    type: 'columnEditor',
+    family: 'passThrough',
```
drdelambre, line 6: oh i think this might be a 'transform' type

PR #1942, opened 2021-07-07
chore: update optimize-css-assets-webpack-plugin

genehynson:
This resolves the dependabot alert for is-svg. Upgrading `optimize-css-assets-webpack-plugin` removes the dependency on `is-svg`.

Leaving as a draft until I do some testing


src/dashboards/components/NoteEditorPreview.test.tsx
```diff
@@ -1,48 +0,0 @@
...
-        const {container} = render(
-          <NoteEditorPreview note={note} scrollTop={0} onScroll={() => {}} />
-        )
-        expect(
-          container.querySelector('img[src*="https://i.imgur.com/k3wIaNU.gif"]')
-        )
-      })
-    })
-  })
-})
```
Asalem1, line 48: For context, this is being removed because it's covered by an e2e test in OSS and Cloud

genehynson:
So it looks like updating the `optimize-css-assets-webpack-plugin` package breaks the `react-markdown` package. All the unit tests involving react-markdown error with `TypeError: Cannot read property 'Tag' of undefined`. And we can't upgrade `react-markdown` easily because they introduced some breaking changes that would require a lot of refactoring on our side. So I'm inclined to skip this for now 🙁

genehynson:
Created an issue to investigate updating `react-markdown` which would unblock this PR: https://github.com/influxdata/ui/issues/1947. I'm going to close this PR for now.

PR #1946, opened 2021-07-08
feat: redirect away from the login page when a user has a valid session

Asalem1:
Closes #https://github.com/influxdata/influxdb/issues/18362

This one's been on my mind. We shouldn't allow the user to access the login page once a valid session has been established since that can be a confusing state to be in, both for the user and for the browser in terms of establishing a session. This PR basically utilizes the existing session management for pre-auth users and makes a request to check whether a user has an established session prior to displaying the login page. If the user already has an established session, we just return them to whichever route they came from

### DIFF
--- a/src/onboarding/containers/LoginPage.tsx
+++ b/src/onboarding/containers/LoginPage.tsx
@@ -18,6 +18,8 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import ErrorBoundary from 'src/shared/components/ErrorBoundary'
 import LoginPageContents from 'src/onboarding/containers/LoginPageContents'
 
+const EMPTY_HISTORY_STACK_LENGTH = 2
+
 export const LoginPage: FC = () => {
   const [hasValidSession, setHasValidSession] = useState(false)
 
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/onboarding/containers/LoginPage.tsx
```diff
@@ -9,30 +9,59 @@ import {
...
-          element={HeadingElement.H1}
-          type={Typeface.Rubik}
-          weight={FontWeight.Regular}
-          className="cf-funnel-page--title"
+export const LoginPage: FC = () => {
+  const [hasValidSession, setHasValidSession] = useState(false)
+
+  const getSessionValidity = useCallback(async () => {
+    try {
+      await client.users.me()
```
genehynson, line 26: so this throws an exception if a non-200 status code is returned?

Asalem1: yup. The API will return an error if the session doesn't exist. It's the same logic behind the current signin flow:

https://github.com/influxdata/ui/blob/master/src/Signin.tsx#L97

genehynson (approved):

Issue #1844, opened 2021-06-29
Replace Moment: Handle the moment functions


zoesteinkamp:
There are a few moment functions we need to change including toDate, duration, humanize, add, subtract, toIosString, isValid, fromNow, valueOf.

- [ ] [ActionTimeInput](https://github.com/influxdata/ui/blob/3b83a4185d28bb66a3010b6ffbfb5926c0eda5e9/src/annotations/components/annotationForm/AnnotationTimeInput.tsx#L57)(toDate, toISOString) 
- [x] [BucketOverlayForm](https://github.com/influxdata/ui/blob/master/src/buckets/components/BucketOverlayForm.tsx#L170) (duration, humanize )
- [x] [RefreshContext](https://github.com/influxdata/ui/blob/6380a4a90f5660b84d0c995afd0c52532b9bd4a9/src/dashboards/components/RefreshContext.tsx#L35)(add)
- [x] [index.ts](https://github.com/influxdata/ui/blob/3b83a4185d28bb66a3010b6ffbfb5926c0eda5e9/src/dashboards/selectors/index.ts#L70) (subtract)
- [x] [predicatesThunk](https://github.com/influxdata/ui/blob/3b83a4185d28bb66a3010b6ffbfb5926c0eda5e9/src/shared/actions/predicatesThunks.ts#L67) (toISOString)
- [x] [DatePicker.tsx](https://github.com/influxdata/ui/blob/3b83a4185d28bb66a3010b6ffbfb5926c0eda5e9/src/shared/components/dateRangePicker/DatePicker.tsx#L40)(isValid)
- [x] [DeleteDataForm.tsx](https://github.com/influxdata/ui/blob/3b83a4185d28bb66a3010b6ffbfb5926c0eda5e9/src/shared/components/DeleteDataForm/DeleteDataForm.tsx#L108)(toIosString)
- [x] [duration.ts](https://github.com/influxdata/ui/blob/3b83a4185d28bb66a3010b6ffbfb5926c0eda5e9/src/shared/utils/duration.ts#L147)(subtract, toISOString)
- [x] [relativeTimestampFormatter.ts](https://github.com/influxdata/ui/blob/3b83a4185d28bb66a3010b6ffbfb5926c0eda5e9/src/shared/utils/relativeTimestampFormatter.ts#L7) (fromNow)
[...]

PR #1954, opened 2021-07-08
refactor: remove moment.add() from RefreshContext

ChitlangeSahas:
Completes part of #1844 

This PR deletes the `moment` library usage and redoes the logic using the vanilla `Date` library. 

**Notes for the reviewer:**
* Helper function `incrementDate` was written
* This function made the most sense because we have a dropdown in the overlay form where we ask the user to select option ranging from minutes to Days. The helper function is designed to address those three.

_I did something with rebasing and ended up with 90 commits, since it was an old branch, so had to force push a bunch of things_


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


src/shared/utils/dateTimeUtils.ts
```diff
@@ -16,3 +16,25 @@ export const timesNeedConverting = newTimeRange => {
   // they will be formatted the same
   return !isUtcTime(newTimeRange.lower)
 }
+
+// takes in Date object and adds a duration to it.
+// To add 1 hour, value = 1, unit = 'h'
+// similarly, to add a minute. value = 1, unit = 'm'
+export function incrementDate(input: Date, value: number, unit?: string): Date {
```
hoorayimhelping, line 23: When reading this method name, `incrementDate` sounds like i'm doing something like
```ts
const date = new Date()
date++
```

what do you think about something like

```ts
addDurationToDate(date: Date, duration: number, unit?: string): Date {
  const clonedDate = new Date(input)
  ...
```
? Is that the intent: we want to add a duration to a date (I personally prefer `duration` since it's less ambiguous than `value`). Your intent with `result = new Date(input)` is to clone the date so you don't change the argument passed in, right? I'd say use a variable name to make that clear.

src/dashboards/components/RefreshContext.tsx
```diff
@@ -14,9 +13,14 @@ import {
...
 
+// Utils
+import {createDateTimeFormatter} from 'src/utils/datetime/formatters'
+import {incrementDate} from 'src/shared/utils/dateTimeUtils'
+
 export const AutoRefreshContext = createContext(null)
 
-const DEFAULT_TIME_AHEAD = '01:00'
+// default time increment is 1 hour
+const DEFAULT_TIME_AHEAD = 1
```
hoorayimhelping, line 23:
```suggestion
const oneHourInTheFuture = 1
```

or 
```ts
const ONE_HOUR_IN_THE_FUTURE = 1
```
I think the idea that this is a default is implied and somewhat apparent in this variable. I think it's more documenting to name it after what it's value represents, rather than what it does. Letting the reader know this is one hour in the future, rather than a default value, which is kind of apparent from using it, makes it so the reader doesn't have to scroll up to find out what the default value is. If you can encode a comment inside a variable name and remove the need for a comment, that's probably a well-named variable.

I think it's fine to not name this like a constant in all caps. when I think of a constant, I think of something that's constant, like the date texas became a state, or the lat/long of los angeles. This is more relative to the system we're using, but it's not a big deal.

src/dashboards/components/RefreshContext.tsx
```diff
@@ -31,21 +35,24 @@ export interface AutoRefreshState {
...
+  const startTime = new Date()
+  const copyStart = startTime.getTime()
+  const endTime = incrementDate(
+    startTime,
+    timeoutNumber,
     timeoutUnit[0].toLowerCase()
   )
-  const cutoff = endTime.unix() - copyStart
-  return cutoff * 1000
+  const cutoff = endTime.getTime() - copyStart
```
hoorayimhelping, line 53: :+1: nice. This way doesn't lose millisecond precision like getting the time in seconds than multiplying by 1000 does.

hoorayimhelping (changes requested):
great work! just a couple of small little nits

src/shared/utils/dateTimeUtils.ts
ChitlangeSahas: Yes, that method name and the signature makes more sense! We are adding a time duration to our date object.

src/dashboards/components/RefreshContext.tsx
```diff
@@ -31,21 +34,24 @@ export interface AutoRefreshState {
 }
 
 const jumpAheadTime = () => {
-  return moment()
-    .add(moment.duration(DEFAULT_TIME_AHEAD))
-    .format('YYYY-MM-DD HH:mm:ss')
+  const newTime = addDurationToDate(new Date(), oneHourInTheFuture, 'h')
```
hoorayimhelping, line 37: 👌👌👌👌

hoorayimhelping (approved):

PR #1958, opened 2021-07-09
fix: Fix options in notebooks

### LINKED ISSUE #1951: Maps: fix broken customization options in Notebooks

zoesteinkamp:
Closes #1951 

Alright this fixes the options in notebooks.

### DIFF
--- a/src/visualization/types/Map/GeoOptions.tsx
+++ b/src/visualization/types/Map/GeoOptions.tsx
@@ -1,6 +1,4 @@
 import React, {CSSProperties, FC} from 'react'
-import {isEmpty} from 'lodash'
-import {DEFAULT_THRESHOLDS_GEO_COLORS} from 'src/shared/constants/thresholds'
 
 import {
   Grid,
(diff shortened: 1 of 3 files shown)


src/visualization/types/Map/GeoOptions.tsx
```diff
@@ -36,6 +39,27 @@ export enum MapType {
 
 export const GeoOptions: FC<Props> = ({properties, update, results}) => {
   const {useS2CellID} = properties
+
+  if (!properties.layers.length) {
```
TCL735, line 43: `!properties?.layers?.length` is probably better unless you can guarantee `properties.layers` always exists here

src/visualization/types/Map/GeoOptions.tsx
```diff
@@ -1,4 +1,7 @@
 import React, {CSSProperties, FC} from 'react'
+import _ from 'lodash'
```
genehynson, line 2: please don't import all of lodash - just import the functions you need 🙏

genehynson:
do we have any tests for rendering maps visualization in notebooks?

src/visualization/types/Map/GeoOptions.tsx
zoesteinkamp: got it!

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1958)

Palakp41 (changes requested):
I think instead of updating properties and adding colors in GeoOptions, seems a bit weird. What if instead of checking if layers is empty here, what if we update layers to not be an empty array here: https://github.com/influxdata/ui/blob/master/src/visualization/types/Map/properties.ts#L42?

We can use the default values like this: 
```
    {
      type: 'pointMap',
      colorDimension: {label: 'Value'},
      colorField: '_value',
      isClustered: false,
      tooltipColumns: [],
    }
  ```

Palakp41 (approved):
Looks good to me!

PR #1959, opened 2021-07-09
chore: don't re-render functions on every keystroke

### LINKED ISSUE #1754: typing in notebooks flux editor is slow
Typing in the notebooks flux editor is slow. This causes weird behavior when you have multiple tabs open with different notebooks.

I've noticed that if I close the functions panel, the editor is faster. But others have reported that does not solve the slowness for them.

I think some component tree is being re-rendered unnecessarily with every key stroke. 

Note: we do not want to add debounce to the editor to work around this problem. We should try and reduce the number of re-renders occurring first

Slack convo: https://influxdata.slack.com/archives/C01E1H8D04Q/p1623789945202400

genehynson:
Closes #1754 

Functions were being rendered multiple times on every keystroke and was causing a slow typing experience. I wrapped it in `useMemo` and now it only renders when the Function's search box is used. 

### Notebooks:

Before: 1 keystroke = 2 renders = 40ms render time
After: 1 keystroke = 0 renders = 16ms render time

Screen Shot 2021-07-09 at 1 47 17 PM
Screen Shot 2021-07-09 at 1 44 25 PM

### Data Explorer:

Before: 1 keystroke = 10 renders = 47ms render time
After: 1 keystroke = 0 renders = 5ms render time

image
image

### Example typing speed improvement

Before:

https://user-images.githubusercontent.com/6411855/125136684-28560200-e0c0-11eb-8b90-ac102ba1d21c.mov

After:

https://user-images.githubusercontent.com/6411855/125136787-53d8ec80-e0c0-11eb-85c3-86c86d4d7dc5.mov

### DIFF
--- a/src/flows/pipes/RawFluxEditor/functions.tsx
+++ b/src/flows/pipes/RawFluxEditor/functions.tsx
@@ -48,57 +48,59 @@ const Functions: FC<Props> = ({onSelect}) => {
     [search]
   )
 
-  let fnComponent
+  return useMemo(() => {
+    let fnComponent
 
-  if (!Object.keys(filteredFunctions).length) {
-    fnComponent = (
-      <EmptyState size={ComponentSize.ExtraSmall}>
-        <EmptyState.Text>No functions match your search</EmptyState.Text>
-      </EmptyState>
-    )
-  } else {
-    fnComponent = Object.entries(filteredFunctions).map(([category, fns]) => (
-      <dl className="flux-toolbar--category" key={category}>
-        <dt className="flux-toolbar--heading">{category}</dt>
-        {fns.map(fn => (
-          <Fn
-            onClickFunction={onSelect}
-            key={`${fn.name}_${fn.desc}`}
-            func={fn}
-            testID={fn.name}
-          />
-        ))}
-      </dl>
-    ))
-  }
+    if (!Object.keys(filteredFunctions).length) {
+      fnComponent = (
+        <EmptyState size={ComponentSize.ExtraSmall}>
+          <EmptyState.Text>No functions match your search</EmptyState.Text>
+        </EmptyState>
+      )
+    } else {
+      fnComponent = Object.entries(filteredFunctions).map(([category, fns]) => (
+        <dl className="flux-toolbar--category" key={category}>
+          <dt className="flux-toolbar--heading">{category}</dt>
+          {fns.map(fn => (
+            <Fn
+              onClickFunction={onSelect}
+              key={`${fn.name}_${fn.desc}`}
+              func={fn}
+              testID={fn.name}
+            />
+          ))}
+        </dl>
+      ))
+    }
 
-  const body = isFlagEnabled('flow-sidebar') ? (
-    <div className="flux-toolbar--list" data-testid="flux-toolbar--list">
-      {fnComponent}
-    </div>
-  ) : (
-    <DapperScrollbars className="flux-toolbar--scroll-area">
+    const body = isFlagEnabled('flow-sidebar') ? (
       <div className="flux-toolbar--list" data-testid="flux-toolbar--list">
         {fnComponent}
       </div>
-    </DapperScrollbars>
-  )
+    ) : (
+      <DapperScrollbars className="flux-toolbar--scroll-area">
+        <div className="flux-toolbar--list" data-testid="flux-toolbar--list">
+          {fnComponent}
+        </div>
+      </DapperScrollbars>
+    )
 
-  return (
-    <div className="flux-toolbar">
-      <div className="flux-toolbar--search">
-        <Input
-          type={InputType.Text}
-          icon={IconFont.Search}
-          placeholder="Filter Functions..."
-          onChange={updateSearch}
-          value={search}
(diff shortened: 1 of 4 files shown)


src/timeMachine/components/fluxFunctionsToolbar/FluxFunctionsToolbar.tsx
```diff
@@ -56,36 +52,22 @@ class FluxFunctionsToolbar extends PureComponent<Props, State> {
...
-
-  private handleClickFunction = (func: FluxToolbarFunction) => {
-    this.props.onInsertFluxFunction(func)
-  }
-}
-
-const mstp = (state: AppState) => {
-  const activeQueryText = getActiveQuery(state).text
-
-  return {activeQueryText}
```
genehynson, line 82: I didn't see any reason for this component to need the `activeQueryText` so I removed that dependency

src/flows/pipes/RawFluxEditor/functions.tsx
```diff
@@ -48,57 +48,59 @@ const Functions: FC<Props> = ({onSelect}) => {
     [search]
   )
 
-  let fnComponent
+  return useMemo(() => {
```
genehynson, line 51: the diff looks crazy, but all I did was wrap a large section of this component in a `useMemo` which caused the indentation to change and then prettier adjusted some things too

src/timeMachine/components/fluxFunctionsToolbar/FluxFunctionsToolbar.tsx
```diff
@@ -28,22 +25,21 @@ interface OwnProps {
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = OwnProps & ReduxProps
 
-interface State {
-  searchTerm: string
-}
+const FluxFunctionsToolbar: FC<Props> = (props: Props) => {
```
genehynson, line 28: converted this component to a FC to take advantage of useMemo

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1959)

drdelambre (approved):
Looks good. That ish is expensive

src/timeMachine/components/fluxFunctionsToolbar/FluxFunctionsToolbar.tsx
genehynson: ok I lied, it's needed

src/flows/pipes/RawFluxEditor/functions.tsx
hoorayimhelping: protip: add `?w=1` to the url to ignore whitespace

https://github.com/influxdata/ui/pull/1959/files?w=1

genehynson: DUDE

Asalem1 (approved):

src/timeMachine/components/fluxFunctionsToolbar/FluxFunctionsToolbar.tsx
genehynson: lied again, it's not needed. The issue was that we weren't re-rendering when props changed

genehynson:
Turns out we set the `editorInstance` after the editor is mounted which creates a chicken and the egg problem here because the functions are rendered without the editorInstance and because they are not re-rendered, the "Inject" button doesn't work initially. So I memoized the parent components where `editorInstance` is stored. This is necessary because simply adding the `onSelect` function to the memoize dependency array causes the re-render on every keystroke. So the parent component should only re-render when specific deps are changed, such as the `editorInstance`, the `activeQueryTabIndex`, or if the user wants to hide the function panel

src/timeMachine/components/fluxFunctionsToolbar/FluxFunctionsToolbar.tsx
```diff
@@ -56,36 +55,22 @@ class FluxFunctionsToolbar extends PureComponent<Props, State> {
...
 
 const mdtp = {
   onSetActiveQueryText: setActiveQueryText,
 }
 
-const connector = connect(mstp, mdtp)
+const connector = connect(null, mdtp)
 
-export default connector(ErrorHandling(FluxFunctionsToolbar))
+export default connector(FluxFunctionsToolbar)
```
Asalem1, line 76: nit: this was refactored to a functional component, why not use hooks here for dispatching the query?

Asalem1 (approved):
Still looks good to me

src/timeMachine/components/fluxFunctionsToolbar/FluxFunctionsToolbar.tsx
genehynson: turns out there actually is no need for this function to be connected to redux at all so I just removed it

PR #1985, opened 2021-07-12
feat: adding robust validation to the AnnotationForm, with unit tests

### LINKED ISSUE #1751: annotations: overlay for creating/editing

jrenee42:
Closes #1751 

- adds validation so that end time is after start time
- dates cannot be in the future
- if the date is empty or in the wrong format, then cannot 'close' and save/create an annotation either
- if a range annotation, dates must be different


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1985)

src/annotations/components/annotationForm/AnnotationForm.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+  START_TIME_IN_FUTURE_MESSAGE,
+  TIMES_ARE_SAME_MESSAGE,
+  WRONG_ORDER_MESSAGE,
+} from 'src/annotations/components/annotationForm/AnnotationForm'
+import {fireEvent} from '@testing-library/react'
+import {createDateTimeFormatter} from 'src/utils/datetime/formatters'
+import {REQUIRED_ERROR} from 'src/annotations/components/annotationForm/AnnotationTimeInput'
+
+const setup = (annoType: 'point' | 'range', startTime, endTime) => {
+  const doNothing = () => {}
```
hoorayimhelping, line 15: i think conventionally in tests, this should be `jest.fn()`

src/annotations/components/annotationForm/AnnotationForm.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+const setupPoint = time => {
+  return setup('point', time, null)
+}
+
+const setupRange = (start, end) => {
+  return setup('range', start, end)
+}
+
+describe('Annotation Form Component tests', () => {
```
hoorayimhelping, line 38:
```suggestion
describe('Annotation Form Component', () => {
```
nit, but we're describing the form, not the tests :)

src/annotations/components/annotationForm/AnnotationForm.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const fiveMinutesAgo = new Date().getTime() - 1000 * 60 * 5
+  const oneHourAgo = new Date().getTime() - 1000 * 60 * 60
+  const oneHourInFuture = fiveMinutesAgo + 60 * 60 * 1000
```
hoorayimhelping, line 41: dealing with actual time in tests is a kind of dangerous thing, especially around new years and daylight savings switch-overs.

i think it might be safer to just pick an arbitrary date in the past and use that as `now` and make all other dates relative to that. it could be as complicated as mocking out `Date.now` and calling that (which is identical to `new Date().getTime()`) or as simple as saying

```ts
const now = 1626135535000
const fiveMinutesAgo = now - 1000 * 60 * 5
```

etc

src/annotations/components/annotationForm/AnnotationForm.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  const formatter = createDateTimeFormatter('YYYY-MM-DD hh:mm:ss a')
```
hoorayimhelping, line 43: 👍

src/annotations/components/annotationForm/AnnotationForm.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+      'This field is required'
+    )
+    expect(getByTestId('annotation-submit-button')).toBeDisabled()
+
+    fireEvent.change(getByTestId('edit-annotation-message'), {
+      target: {value: 'abcde'},
+    })
+
+    expect(queryByTestId('form--element-error')).toBeNull()
+    expect(getByTestId('annotation-submit-button')).not.toBeDisabled()
```
hoorayimhelping, line 58: so i think as a general rule, unit / integration tests should only test one action. this is testing two actions - the failure of the validation, and the success of it being fixed. what do you think of having a describe section for the validation failures and then having one for it fixing it?

i was mainly reading this thread: https://softwareengineering.stackexchange.com/a/171502/10616

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -49,25 +50,37 @@ interface Props {
...
-    // not checking if start <= end right now
-    // initially, the times are numbers, and then if the user manually edits them then
-    // they are strings, so the simple compare is non-trivial.
-    // plus, the backend checks if the startTime is before or equals the endTime
-    // so, letting the backend do that check for now.
     if (annotationType === 'range') {
-      return Boolean(isValidPointAnnotation && endTime)
+      return Boolean(
+        isValidPointAnnotation &&
+          endTime &&
```
hoorayimhelping, line 80: do you need to wrap this whole expression in a `Boolean`? or just the `endTime` since it seems to be the only non boolean here?

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -49,25 +50,37 @@ interface Props {
...
+  // (until the user edits it)
+  const [endTimeFormatValid, setEndTimeFormatValid] = useState(true)
+  const [startTimeFormatValid, setStartTimeFormatValid] = useState(true)
+
   const dispatch = useDispatch()
 
   const isValidAnnotationForm = (): boolean => {
-    const isValidPointAnnotation = Boolean(summary?.length && startTime)
+    const isValidPointAnnotation = Boolean(
+      summary?.length && startTime && startTimeFormatValid && isStartTimeValid()
```
hoorayimhelping, line 74: same question here about wrapping this in a `Boolean`

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -137,6 +152,74 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
...
+
+    const start = moment(startTime)
+    const end = moment(endTime)
+
+    if (end.isSame(start)) {
+      return {
+        isValid: false,
+        message: TIMES_ARE_SAME_MESSAGE,
+      }
+    }
```
hoorayimhelping, line 194: product question: should we give the user a choice to make a point annotation here? @russorat @daviddkkim

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -137,6 +152,74 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
...
+
+  /**
+   * If there is a problem with the end time (with respect to the start time)
+   * return an error message here, along with the isValid flag
+   *
+   * If no message, then it is valid (just return a true isValid flag)
+   *
+   * since this is only vaild for range annotations, if it is a point annotations just return true
+   * */
+  const validateEndTime = () => {
```
hoorayimhelping, line 174: can you add a return type to this?

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -137,6 +152,74 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
     props.onClose()
   }
 
+  const isTimeInFuture = timeToCheck => {
```
hoorayimhelping, line 155: what is `timeToCheck`? is it a `Date`? a timestamp?

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -137,6 +152,74 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
...
+    }
+
+    if (isTimeInFuture(endTime)) {
+      return {isValid: false, message: END_TIME_IN_FUTURE_MESSAGE}
+    }
+
+    return {isValid: true, message: null}
+  }
+
+  const getEndTimeValidationMessage = () => {
```
hoorayimhelping, line 207: let's get some return types on these helper functions in this section

hoorayimhelping (changes requested):
nice work! have a couple of questions and requests

src/annotations/components/annotationForm/AnnotationForm.tsx
daviddkkim: Would allowing them to make a point annotation when start and end time is the same cause any entanglement in the way annotation is implemented? 

If it does, I don't know if it's worth the refactor because we are giving them a clear message on what is causing the error and the user can correct it.

src/annotations/components/annotationForm/AnnotationForm.test.tsx
jrenee42: the issue is that one of the things I am testing is that future dates are illegal/invalid.  new years/ dst/est switchovers are rare enough it isn't a problem for this one test.

src/annotations/components/annotationForm/AnnotationForm.test.tsx
jrenee42: that would be a duplication of tests, since need to verify that it is invalid , and then that fixing it makes it valid.

(test a, then second test: a and b.... where the second test supersedes the first test)

src/annotations/components/annotationForm/AnnotationForm.tsx
jrenee42: it wouldn't cause a problem (leaving the times the same, they *are* the same)
but the ui tells the difference between point and range because point has the same times.  
so it would make the ui, coding and usability much more complicated.
(nontrivial refactor)

src/annotations/components/annotationForm/AnnotationForm.test.tsx
jrenee42: when future times aren't being tested this will be fine,
but when testing the future; it won't work.

src/annotations/components/annotationForm/AnnotationForm.tsx
jrenee42: will do

src/annotations/components/annotationForm/AnnotationForm.tsx
jrenee42: number or string.  added comments.

src/annotations/components/annotationForm/AnnotationForm.tsx
jrenee42: after I set the return types on the helper functions ,this isn't needed any more.  so that is fixed now.

src/annotations/components/annotationForm/AnnotationForm.tsx
jrenee42: see above; fixed now

src/annotations/components/annotationForm/AnnotationForm.tsx
russorat: this is fine as is.

src/annotations/components/annotationForm/AnnotationForm.test.tsx
hoorayimhelping: I still prefer that to having multiple actions packed into a test, but that's an ideal to strive for.

If you prefer to keep it this way, can you update the test descriptions in these to say that they also fix the validation issues?

src/annotations/components/annotationForm/AnnotationForm.test.tsx
hoorayimhelping: >when future times aren't being tested this will be fine,
>but when testing the future; it won't work.

how does the code determine if the date is in the future? does it calculate the future based on the time now, e.g. `Date.now()` / `new Time().getTime()` ? or does it calculate the future based on the time it's comparing e.g. `isInTheFuture = endDate > startDate`?

if it's the former, we'll have to mock out what time we return from `Date.now()` / `new Date()`

>new years/ dst/est switchovers are rare enough it isn't a problem for this one test.

I'm going to push back here and say it's not acceptable to have a situation where a test will knowingly break 2-3 times a year. Not sure if that's what we're looking at, but if that's the case, we can do better.

jrenee42: for future testing, it does a compare to see if the time to check is After moment()
which gives it the current time.

i will come up with a scheme for injecting/mocking this after lunch.

src/annotations/components/annotationForm/AnnotationForm.test.tsx
jrenee42: i'll update the descriptions.

jrenee42: ...updated the description at the top, since each test messes up the validation, checks that the button is disabled, then fixes it and confirms that the validation message is gone and that the button is enabled.

src/annotations/components/annotationForm/AnnotationForm.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+  TIMES_ARE_SAME_MESSAGE,
+  WRONG_ORDER_MESSAGE,
+} from 'src/annotations/components/annotationForm/AnnotationForm'
+import {fireEvent} from '@testing-library/react'
+import {createDateTimeFormatter} from 'src/utils/datetime/formatters'
+import {REQUIRED_ERROR} from 'src/annotations/components/annotationForm/AnnotationTimeInput'
+
+// setting 'now' as the first of march, 2021 at 11 am.
+// the clocks changed on march 14, btw
+const now = new Date('2021-03-01 11:00 am').getTime()
```
hoorayimhelping, line 16: 👍

hoorayimhelping (approved):
awesome work!

Issue #1935, opened 2021-07-07
Query/Query Result Sync Issue


subirjolly:
Before
![Before.gif](https://images.zenhubusercontent.com/5fa9843597598f467dccb2b9/a0812268-5101-4b9f-81be-ceb7ec79aa88)

After
![After.gif](https://images.zenhubusercontent.com/5fa9843597598f467dccb2b9/c3a8af4f-9eaf-4293-942a-22ece1e4222d)

The Issue:
In both the gifs, you can see that the cell isn't loading after you click on `Submit` and then the `Green Check`. The underlying issue, I think, is that there's no global query context and we have to keep/sync the states of the query in multiple places/components. The solution is to move towards having a global query object/context which keeps track on all the queries and we don't have to rely on different mechanisms to store query and their states.

@drdelambre @Asalem1 Thoughts?

subirjolly:
https://app.zenhub.com/workspaces/applicationsui-unity-606b1041215d5f0017898847/issues/influxdata/ui/1935

drdelambre:
yeah, this is by design to reduce running the same query (just to return the same results). the need for centralizing the query status is that the "query is currently running" is predominately handled by the view layer. i remember work was done in here to explicitly share this state between the cell and the veo, but iirc it only works one way (i dont think a pending cell's request can be viewed as pending in the VEO, nor can you cancel it once in there). centralizing these affordances (i have a query, what is it's loading state, how can i cancel it, etc), should remove the issue of duplicating these mechanisms / sources of truth for the query's state, and removing it from the current redux implementation should decouple the mechanisms from their current dependence on internal time machine state

subirjolly:
This note is for myself. I need to do the query caching by using global query context when we press cmd+enter instead of clicking Submit.

To summarize the work so far and my findings. I have created a global query context which can cache queries and their results for a default of 30 seconds and will clean up the queries after 10 minutes on any interaction with the context. There need to be a couple of phases for this rollout.

**Phase 1:** Create GlobalQueryContext and use it only when Running the queries. So, if the query result hasn't been cached in the context then the query will be run and if the same query gets run from any part of the UI then you will be getting the cached result instantaneously. In order to do that, we will have to get the data from this caching layer and insert into the respective places where the data currently is expected to reside. This change needs to happen across the UI.

[...]

PR #1991, opened 2021-07-13
fix: fix cancel query issue in #1935

subirjolly:
c3a8af4f-9eaf-4293-942a-22ece1e4222d

The issue was that when you click on `Submit` and before the query finishes, you click on `Save`, the query still gets cancelled but due to the discrepancy in syncing of the 2 states the loading spinner keeps on going.

I dug deeper into the issue and found that we aren't really rejecting when we need to reject a promise. Instead, we are calling a reject function which we created. So, I am adding a fix which we may or may not end up keeping. The fix is to actually reject after running our custom reject function. This actually acknowleges that the promise is finished and is not in lingering state. Makes sure a resolve or a reject has happened. There's a side effect(maybe?) which is kind of unexpected but below is the gif showing that. Basically, the cell we were trying to edit(save) will error out but when you reload it will show the data correctly.

Kapture 2021-07-13 at 14 54 28

@Asalem1 @drdelambre  Please let me know what you think of this solution.

### DIFF
--- a/src/shared/apis/singleQuery.ts
+++ b/src/shared/apis/singleQuery.ts
@@ -49,6 +49,7 @@ export function RunQueryPromiseMutex<T>() {
             })
             .catch((error: Error) => {
               ret.reject(error)
+              reject(error)
             })
         }),
         cancel: ret.cancel,


russorat:
i can't comment on the technical parts, but from the standpoint of fixing the spinners getting out of sync, i like this.

in terms of a cell showing an error immediately after editing, that doesn't feel great, even if it is an edge case. can we make an issue to investigate possible solutions to that for later?

subirjolly:
> i can't comment on the technical parts, but from the standpoint of fixing the spinners getting out of sync, i like this.
> 
> in terms of a cell showing an error immediately after editing, that doesn't feel great, even if it is an edge case. can we make an issue to investigate possible solutions to that for later?

I have created an issue to fix the underlying architecture for Queries. I am going to create a Global Query Context which is going to be responsible for handing all interactions with queries and their responses, system wide.

Asalem1:
I think the solution might be to just not cancel all running queries from the submit query button when it unmounts. I would track down the PR that added that and see if the point was to cancel a query when we leave the data explorer, and just try to cancel the query when leaving that page or if it’s something else

drdelambre:
the code change is fine in here.

i want to echo the last part of ari's comment in that canceling the query on clicking the green button is a junk user experience, and i think fixing that would address russ's point, but all of that seems like a different PR than this; centralizing loading state to accomplish the product intent being different than catching an edge condition you found along the way

Asalem1 (approved):

PR #1993, opened 2021-07-13
chore: revert to usage api

ShmuelLotman:
Closes #

After much deliberation, it was determined that we are in fact going to be using the experimental Usage api for the userpilot statuses, so this is to revert the integration we have now to usage api instead of usage api quartz.


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1993)

genehynson (approved):

src/cloud/components/onboarding/EngagementLink.tsx
```diff
@@ -19,12 +19,12 @@ const EngagementLink: FC = () => {
...
   useEffect(() => {
-    if (isFlagEnabled('newUsageAPI')) {
-      handleGetUserStatus().then(({usageDataStates}) =>
+    if (isFlagEnabled('newUsageAPI') && org?.id) {
+      handleGetUserStatus(org.id).then(({usageDataStates}) =>
         setDataStates(usageDataStates)
       )
     }
-  }, [])
+  }, [org?.id])
```
mavarius, line 27: Will using the orgID here cause it to only update the usage state once per login or on full page reload? We may need to add another flag to recheck the usage state more frequently since it will change during the course of regular use. For instance, we could recheck on path change or something like that.

src/cloud/components/onboarding/useGetUserStatus.ts
```diff
@@ -3,50 +3,61 @@ import {event} from 'src/cloud/utils/reporting'
...
-  getUsage = require('src/client/unityRoutes').getUsage
+  getUsage = require('src/client').getOrgsUsage
 }
 
 export enum USER_PILOT_USER_STATUS {
   NEW_USER = 'NEW_USER',
   DELETED_ALL_DATA = 'DELETED_ALL_DATA',
   NON_WRITING_USER = 'NON_WRITING_USER',
   ACTIVE_USER = 'ACTIVE_USER',
-  ACTIVE_WRITING_USER = 'ACTIVE_WRITING_USER',
```
mavarius, line 19: Thanks for removing this one.

mavarius (approved):
Looks good, just a question on how frequently we're updating the usage state value.

Issue #1544, opened 2021-05-28
I would like the ability to load a non-annotated CSV file into InfluxDB Cloud from the browser


russorat:
This should be broken out into smaller chunks of work. Just using this to organize everything.

### Target Customer / Background data / Context / Market Size
We have heard concerns that users are having trouble loading CSV data through the UI. this is because the UI only handles a specific annotated csv format provided by the query endpoint. This is causing users to get stuck at the very beginning of the journey with what is a very simple csv file.

See also: 
 - https://github.com/influxdata/ui/issues/920
 - https://github.com/influxdata/ui/issues/170

### Goals
- I would like the ability to upload non-annotated csv files through the UI. The goal is NOT to support all CSV files, but enough so that users can get started quickly.
- The Influx CLI tool already has an extremely flexible CSV parser. Ideally, the UI would leverage that technology directly or a similar pattern. The Flux engine also has a way to parse CSV data (see #920). That could be another way.
- Telegraf also has a way to parse CSV data into Line Protocol. There could be some learnings there as well: https://github.com/influxdata/telegraf/tree/master/plugins/parsers/csv

### Acceptance Crit
- As a user, I would like the ability to dryrun the upload process to verify my data parses correctly before loading it (see `dryrun` option in the CLI tool)
- As a user, I can choose the separator to use for my CSV file (comma, tab, pipe, etc) or provide my own
- As a user, I can choose which columns are tags, and which columns are fields
- As a user, I can customize the name of each column
- As a user, I can provide a constant value to use as the measurement name
- As a user, I can choose to ignore one or more columns
- As a user, I can choose a column to use as the measurement name
- As a user, I can choose the column to use as the date timestamp
- As a user, I can customize the date time format used to interpret the timestamp in my data
- As a user, I can choose to insert the data using the current time
- As a user, I can customize the data types of the field columns
- As a user, I can provide a default value for a column if none exists
- As a user, I am presented with an error message explaining why my upload failed
- As a user, I can return to my file upload to make changes if there is an error
- As a user, I can choose to ignore upload errors and continue
- As a user, I can choose any number of header lines to skip at the beginning of the file

### Release Plan
- This should be broken up into smaller parts

### Notes
- Here is information about loading CSV data through the command line: https://docs.influxdata.com/influxdb/cloud/write-data/developer-tools/csv/
- Being able to preview the 
[...]

PR #1996, opened 2021-07-14
fix: resolves csv uploader parsing issue by using native methods

### LINKED ISSUE #920: Idea: Stop parsing CSV in the UI
We have mechanisms to upload csv data into the UI. We have custom parsers for that code.

This issue it to replace our javascript based parsing with calls to the query api using the `csv.from` and `to` syntax for flux: https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/csv/from/

the experience could be similar to dropping an image into a github issue. User could drop an LP file into our UI, and we upload the file and give it a unique url. then we run `csv.from(url: "that new url")`

We could do that on any code editor in the UI.

Asalem1:
Closes #920
Closes https://github.com/influxdata/influxdb/issues/21636

This PR makes a change to the implementation for the CSV Uploader to use the native csv.from rather than parsing the CSV and writing is a line protocol in 6 concurrent streams. After doing some testing it seems like the speed with which the 6 concurrent streams were being written is only marginally faster than writing using the csv.from implementation. In doing so, this resolves issues that have come up due to parsing. It also allows users to upload csv that isn't properly formatted according to line protocol.

This PR also fixes an issue with the abort controller reference where users that cancelled a query would have to reload the page in order to get a new abort controller for the query. The issue manifested itself after a query was aborted and a user tried to upload a different file (the original abort controller being in a cancelled state wouldn't allow for subsequent uploads). 

Finally, this PR adds an actionable button when users have finished uploading their CSV so that they can just click to navigate directly to their bucket. Ideally this would link to a notebook we'd create for them, but I think that's a ways a way at this point.

csv-uploader


Asalem1:
@russorat can you verify that this may close #1544 or did you have something else in mind?

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=1996)

russorat:
i ❤️ this!

src/writeData/components/fileUploads/CsvMethod.tsx
```diff
@@ -46,11 +41,19 @@ const CsvMethod: FC = () => {
...
   }
 
+  let body = <CsvUploaderBody bucket={bucket.name} />
+
+  if (uploadState !== RemoteDataState.NotStarted) {
+    body = <StatusIndicator />
+  }
+
+  const handleSeeUploadedData = () => {
+    history.push(`/orgs/${orgId}/data-explorer?bucket=${bucket.name}`)
```
drdelambre, line 51: IM COMING FOR YOU LINK

Asalem1: totally

drdelambre (approved):
RIP

cypress/e2e/shared/buckets.test.ts
```diff
@@ -108,100 +108,6 @@ describe('Buckets', () => {
...
-    })
-
-    // this is currently not producing success, its actually failing, im going to write a separate issue for this
-    it('closes the overlay upon a successful delete with predicate submission', () => {
-      cy.getByTestID('delete-checkbox').check({force: true})
-      cy.getByTestID('confirm-delete-btn').click()
-      cy.getByTestID('overlay--container').should('not.exist')
-      cy.getByTestID('notification-success').should('have.length', 1)
-    })
-  })
```
Asalem1, line 204: deleted those since it ain't gonna happen

kyleawayan:
Amazing! Thank you!

PR #2013, opened 2021-07-14
fix: make DatePicker and TimeRangeDropdown respect the user timezone setting

### LINKED ISSUE #1827: Bug: Dashboard "Custom Time Range" picker does not show correct times.
# Summary

The custom time range picker does not respect the the local/utc timezone toggle button. The picker always loads with a timestamp that appears to be UTC e.g. `2021-06-28T16:04:52.486Z`. If `Local` time is selected, it should show times in the local timezone (e.g. +7). If `UTC`, then utc. 

Worse, when `UTC` is selected for the dashboard, the picker is actually showing _local_ times with simply a `Z`. See the attached gif. The dashboard is set to utc, when I pick times in the modal e.g. `18:00`, the displayed time says `2021-06-24T01:00:00.185Z` instead of the correct `2021-06-23T18:00:00.185Z`, which you can see once I close the picker modal and look at the displayed time range.

time picker

**Steps to reproduce:**
1. Create a dashboard with at least one cell with a valid, data returning query.
2. Change the dashboard time from `Local` to `UTC`.
3. Pick a custom time range and note how selecting a UTC date and time does not match the displayed time in the modal, but the correct time range is displayed after clicking `Apply Time Range`

**Expected behavior:**
Dates and times displayed should show the correct time _in the selected timezone_.

**Actual behavior:**
See above: datetimes are incorrect.

**Environment info:**
influx cloud2

ChitlangeSahas:
Closes #1827

This PR addresses two components and makes them respect the App's global timezone settings:

* The `TimeRangeDropdown`: This component needed the timezone information so it could display the time in the correct zone, thus respecting the user preference.
* The `DatePicker`: that was based on `react-datepicker` needed some tweaks, the underlying library has a known [issue](https://github.com/Hacker0x01/react-datepicker/issues/498) of forcing the `Local` timezone. 

A brief of the solution:

When the User selects `UTC` in the dropdown the `DatePicker` converts that UTC into `Local time`. We need to display the UTC time in the `DatePicker`, so I added the `UTC time offset` to the `Local` time, and hence it displays the desired UTC time.  

The date picker works fine with `Local` time zone settings, no changes were necessary for that. 

Video: 

https://user-images.githubusercontent.com/18511823/125701148-59daf02d-3a9d-4838-9e2d-4e3482dfd9fd.mov


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2013)

src/shared/utils/duration.ts
```diff
@@ -166,8 +173,9 @@ export const getTimeRangeLabel = (
...
   }
   if (timeRange.type === 'custom') {
-    const lower = moment(timeRange.lower).format(TIME_RANGE_FORMAT)
-    const upper = moment(timeRange.upper).format(TIME_RANGE_FORMAT)
+    const formatter = createDateTimeFormatter(TIME_RANGE_FORMAT, timeZone)
+    const lower = formatter.format(new Date(timeRange.lower))
+    const upper = formatter.format(new Date(timeRange.upper))
     if (singleDirection && singleDirection === TimeRangeDirection.Upper) {
       return upper
     } else if (
```
TCL735, line 180: i know you didn't write this, but while we are making changes to this file... can we also add one more change?

`singleDirection && singleDirection === TimeRangeDirection.Lower` is too wordy. The first part is not necessary. If `singleDirection` is `undefined` then it won't equal to our comparator.

ChitlangeSahas: Yeah absolutely! I can make that change 👍

TCL735: There's one more below it that i couldn't quote.

ChitlangeSahas: I missed it the first time too! changed ✔️

TCL735 (approved):

src/shared/utils/duration.ts
```diff
@@ -166,14 +171,12 @@ export const getTimeRangeLabel = (
     return timeRange.lower
   }
   if (timeRange.type === 'custom') {
-    const lower = moment(timeRange.lower).format(TIME_RANGE_FORMAT)
-    const upper = moment(timeRange.upper).format(TIME_RANGE_FORMAT)
-    if (singleDirection && singleDirection === TimeRangeDirection.Upper) {
+    const formatter = createDateTimeFormatter(TIME_RANGE_FORMAT, timeZone)
+    const lower = formatter.format(new Date(timeRange.lower))
+    const upper = formatter.format(new Date(timeRange.upper))
```
hoorayimhelping, line 176: nice!

hoorayimhelping (changes requested):
good work! one small suggestion to use the `getTimeZone` reducer

src/shared/components/dateRangePicker/DatePicker.tsx
```diff
@@ -172,7 +187,15 @@ export default class DatePicker extends PureComponent<Props, State> {
...
 
   private handleSelectDate = (date: Date): void => {
-    const {onSelectDate} = this.props
+    const {onSelectDate, timeZone} = this.props
+
+    if (timeZone === 'UTC') {
+      // (sahas): the react-datepicker forces the timezone to be the Local timezone.
+      // so when our app in in UTC mode, to make the datepicker respect that timezone,
+      // we have to manually manipulate the Local time and add the offset so that it displays the correct UTC time in the picker
+      date.setMinutes(date.getMinutes() - date.getTimezoneOffset())
```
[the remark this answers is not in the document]
ChitlangeSahas: Very true I should have thought of that! I'll make a function now!

src/shared/components/TimeRangeDropdown.tsx
```diff
@@ -180,4 +183,8 @@ class TimeRangeDropdown extends PureComponent<Props, State> {
   }
 }
 
-export default TimeRangeDropdown
+const mapStateToProps = state => ({
+  timeZone: state.app.persisted.timeZone || 'Local',
```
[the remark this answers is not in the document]
ChitlangeSahas: ✅

src/utils/datetime/formatters.ts
```diff
@@ -177,6 +177,35 @@ export const createDateTimeFormatter = (
...
+          })
+
+        return `${dateParts.year}-${dateParts.month}-${dateParts.day}`
+      }
+
+      return {
+        format: formatDate,
+      }
+    }
+
```
ChitlangeSahas, line 208: We needed a formatter case and here is why:

I was getting a bunch of e2e test failures and couldn't understand why, for a long time. I thought it was just flakiness, but when I looked at the artifacts, I noticed that the new Formatter was not handling the `YYYY-MM-DD` case.

The cypress would `type` : `2019-09-02T08:00:00Z` but the text field would read `2019-09-02 12:00:00 AM T08:00:00Z`. I traced that down and turns out we were not handling the `YYYY-MM-DD` case so that upon a valid case of `YYYY-MM-DD`, the formatter didn't error out.

ChitlangeSahas:
This PR became way too old, and so #2155 will take it from here.

PR #2024, opened 2021-07-15
feat: global search

### LINKED ISSUE #1848: Global Search
AS a user in the InfluxDB UI
WHEN I want to find a resource
THEN I am able to search for the `name`, `label`, `description` from anywhere in the application
AND find a link to that resource

ShmuelLotman:
Closes #1848

This implements a global search feature - one that allows users to search documentation throughout any page on our application by hitting the "ctrl + h" key combination to bring up the interface. This currently utilizes the Algolia api and was implemented in communication w/ Scott Anderson of the Docs team so as to coordinate efforts. 

Support has been working on a Coveo implementation and, as talks progress, we may end up going with the Coveo implementation in the future, but that remains a question and, even if we _do_ use Coveo, we won't be ready to implement it on the UI side for a while, so there's no reason to wait off on the feature. 

https://user-images.githubusercontent.com/29473622/125817874-b69516a6-ca68-4066-b0d4-cb5b079f7642.mov


russorat (commented):
love it!

hoorayimhelping:
What are Coveo and Algolia?

UX note: `ctrl` is usually used on unix/linux and windows, while `cmd` is typically used for the same function on macOS. Is that something we want to account for? We don't really have any conventions around keyboard shortcuts in our UI, and we're starting to add more. @russorat I think this is something we should think about / discuss from a high level

genehynson (changes requested):
remove the yarn-error.log!

src/shared/search/GlobalSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const facetFilters = useMemo(
+    () => ({
+      facetFilters: [['project: influxdb', 'flux:true'], 'version: cloud'],
+    }),
+    []
+  )
+  return showState ? (
+    <ClickOutside onClickOutside={() => setShowState(false)}>
+      <DocSearchModal
+        apiKey="ba4435a9d456ac0d954cc276206eac06"
```
genehynson, line 70: is this something we want to leave in our public repo?

russorat:
> What are Coveo and Algolia?
> 
> UX note: `ctrl` is usually used on unix/linux and windows, while `cmd` is typically used for the same function on macOS. Is that something we want to account for? We don't really have any conventions around keyboard shortcuts in our UI, and we're starting to add more. @russorat I think this is something we should think about / discuss from a high level

agree our keyboard shortcuts need some standardization, but we can tackle that in a separate issue along with discoverability concerns. i wouldn't hold this PR up to solve those though. We can track usage with the included event and build towards increasing it. incremental progress here makes sense.

src/shared/search/GlobalSearch.tsx
sanderson: This is a read-only API key that is ok to show publicly.

sanderson:
> What are Coveo and Algolia?

Algolia is the search indexing service currently used to provide search results for the in-docs search. Coveo is a content indexing service the Support team is using to provide search functionality to the support portal. It provides results from mulitple InfluxData content sources (docs, blogs, community, etc.). Coveo ties directly into Salesforce and the tools the Support team is using. We're looking at what it would take to use Coveo for searching the docs, but are still in the the exploratory phase.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2024)

.gitignore
```diff
@@ -21,6 +21,7 @@ release
 cypress.env.json
 junit.xml
 npm-error.log
+yarn-error.log
```
genehynson, line 24: thanks!

mavarius:
> What are Coveo and Algolia?
> 
> UX note: `ctrl` is usually used on unix/linux and windows, while `cmd` is typically used for the same function on macOS. Is that something we want to account for? We don't really have any conventions around keyboard shortcuts in our UI, and we're starting to add more. @russorat I think this is something we should think about / discuss from a high level

Keyboard shortcuts won't be the only way to access this. The homepage changes coming shortly will have a dedicated docs panel with a search bar and the nav changes coming to Clockface 3.0 will have an ever-present search button. This first iteration though gets the ball rolling.

src/shared/search/GlobalSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+import {DocSearchModal} from '@docsearch/react'
+import {v4 as uuid} from 'uuid'
+import {ClickOutside} from 'src/shared/components/ClickOutside'
+import '@docsearch/react/style'
+
+import './GlobalSearch.scss'
+
+import {event} from 'src/cloud/utils/reporting'
+
+const Hit: FC<any> = memo(({hit, children}) => {
```
genehynson, line 12: should this be `useMemo`?

src/shared/search/GlobalSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+    event(
+      'clicked search result',
+      {
+        resultURL: clickedElem.href,
+        fromPage: location.pathname,
+        query: searchQuery?.value,
+      },
+      {}
+    )
+  }, [linkID, location.pathname])
```
genehynson, line 29: should we be concerned about having a `uuid()` in the deps array? Feel like that's something that might change considering line 13?

src/shared/search/GlobalSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+    } else if (e.key === 'Escape') {
+      setShowState(false)
+    }
+  }
+  useEffect(() => {
+    document.addEventListener('keydown', toggleShowSearch)
+    return () => document.removeEventListener('keydown', toggleShowSearch)
+  }, [])
+
+  const facetFilters = useMemo(
```
genehynson, line 61: does this need to be memoized? Seems to be hardcoded

src/shared/search/GlobalSearch.tsx
drdelambre: usually not. appId as well. @genehynson can you help @ShmuelLotman figure out how to route this through the build pipeline so that we can have this defined as an environment variable with a default set for development and one baked into the production builds, even if they're the same key.
the case that no key is set would also need to be handled in here as well

package.json
```diff
@@ -176,6 +177,7 @@
     "connected-react-router": "^6.8.0",
     "d3-format": "^1.3.2",
     "d3-scale": "^2.1.0",
+    "docsearch.js": "^2.6.3",
```
drdelambre, line 180: is this still needed?

src/shared/search/GlobalSearch.tsx
drdelambre: even if it's a read-only api key, we'd still want to be able to configure it on the build level

src/shared/search/GlobalSearch.tsx
ShmuelLotman: no, useMemo is to memoize a value, memo is to memoize a React component

src/shared/search/GlobalSearch.tsx
ShmuelLotman: the reason its in here is to track that change

src/shared/search/GlobalSearch.tsx
ShmuelLotman: according to the Algolia docs, all parameters passed into the component should be memoized

package.json
ShmuelLotman: no, good catch, thanks

src/shared/search/GlobalSearch.tsx
genehynson: sorry maybe I'm not quite following - but wouldn't `linkID` change every time this is rendered? So it will basically always re-render and will never stay memoized?

drdelambre (approved):

genehynson (approved):
love. it.

PR #2045, opened 2021-07-16
fix: add try/catch so that no matter what things don't break

joelfried:
This was the function that errored earlier today and caused a prod incident.  This will make it so that even if it errors the rest of the site doesn't go down.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -47,11 +47,16 @@ const App: FC = () => {
 
   useEffect(() => {
     if (CLOUD && isFlagEnabled('rudderstackReporting')) {
-        try {
-            load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
-        } catch (e) {
-            console.error("Error loading Rudderstack with wk: ", RUDDERSTACK_WRITE_KEY, " at: ", RUDDERSTACK_DATA_PLANE_URL)
-        }
+      try {
+        load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
+      } catch (e) {
+        console.error(
+          'Error loading Rudderstack with wk: ',
+          RUDDERSTACK_WRITE_KEY,
+          ' at: ',
+          RUDDERSTACK_DATA_PLANE_URL
+        )
+      }
     }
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -47,7 +47,11 @@ const App: FC = () => {
 
   useEffect(() => {
     if (CLOUD && isFlagEnabled('rudderstackReporting')) {
-      load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
+        try {
+            load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
+        } catch (e) {
```
Asalem1, line 52:
```suggestion
        } catch {
```

That should pass the linter

Asalem1 (approved):

src/App.tsx
joelfried: Thanks!

src/App.tsx
```diff
@@ -47,7 +47,16 @@ const App: FC = () => {
...
-      load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
+      try {
+        load(RUDDERSTACK_WRITE_KEY, RUDDERSTACK_DATA_PLANE_URL)
+      } catch {
+        console.error(
+          'Error loading Rudderstack with wk: ',
+          RUDDERSTACK_WRITE_KEY,
+          ' at: ',
+          RUDDERSTACK_DATA_PLANE_URL
+        )
```
genehynson, line 58: I think we should report this error to Honeybadger so we can see it happening on our user's browsers

genehynson: check out the `reportErrorThroughHoneyBadger` function

joelfried: Good call, added

genehynson (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2045)

PR #2054, opened 2021-07-19
refactor: remove moment from timeMachine/selectors/index.ts

### LINKED ISSUE #2053: Moment Refactor: remove moment from timeMachine/selectors
We need to remove `moment` from in `timemachine/selectors/index.ts`

ChitlangeSahas:
Closes #2053

part of : #1844

This PR removes `moment` from `src/timeMachine/selectors/index.ts`


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2054)

src/timeMachine/selectors/index.ts
```diff
@@ -293,14 +295,27 @@ export const getStartTime = (timeRange: TimeRange) => {
   }
 }
 
+// this function takes in a string that represents a date, and swaps the timezone to local.
+// if the date passed was 2021-09-02 12:00:00 UTC the function will return 2021-09-02 12:00:00 <Local TZ>
+const setTimeToLocal = (date?: string): Date => {
```
hoorayimhelping, line 300: i'm confused about the intent of this.

`valueOf` and `getTime` return the number of milliseconds and seconds in UTC. are you trying to return the number of milliseconds in local time?

ChitlangeSahas: So, when we pass a string like `2019-09-02` into moment, it interprets the timezone as Local.
Whereas the Date package interprets it as UTC time zone. 

This makes a slight discrepancy in the `valueOf` return value. So to fix it, I change the timezone of the passed string to local. Just like the `[setTimeToUTC](https://github.com/influxdata/ui/blob/e198ecfb1341dbbeb2b8a2d9082d718919704810/src/dashboards/selectors/index.ts#L66)`

I found that out after reading this: https://stackoverflow.com/a/31402874

hoorayimhelping: >I'm trying to get the UTC timestamp in milliseconds but for local time

@ChitlangeSahas is the value we have passed in **always** going to be a string like `YYYY-MM-DD` ?

src/timeMachine/selectors/index.ts
```diff
@@ -293,14 +295,27 @@ export const getStartTime = (timeRange: TimeRange) => {
...
+  const localTime = date ? new Date(date) : new Date()
+  localTime.setMinutes(localTime.getMinutes() + getTimezoneOffset())
+
+  return localTime
+}
+
```
hoorayimhelping, line 306: I hate to do this so late in the game, but I think this method should be moved into `dateTimeUtils`

ChitlangeSahas: Oh yes totally, i'll move it!

hoorayimhelping (approved):

PR #2083, opened 2021-07-20
fix: task run retry no longer opening editor

### LINKED ISSUE #2067: Task runs page: retry button should refresh the task list, not open the editor
### Background
**Steps to reproduce:**

1.  Open any task run history page
2. click `retry` for a failed run

**Expected behavior:**
the task retries successfully and the page refreshes to indicate the change in status. 

**Actual behavior:**
the task retries successfully and then redirects the user to the task editor page

**Environment info:**

- System info: internal cloud instance.  

### Acceptance Criteria
- When the `retry` button on the tasks runs page is clicked, the page refreshes, rather than going to the Edit Task page.
  - This should ideally be done using redux so the page automatically re-renders. We previously tried it with a hard refresh

mtian725:
Closes #2067

Use of `getRuns(taskID)` is following approach from the function `handleRunTask()` in `TaskRunsCard.tsx`

## Before 
https://user-images.githubusercontent.com/39343323/126373776-cfa4bfe6-27ca-4914-b8d3-799c9580e532.mov

## After
https://user-images.githubusercontent.com/39343323/126374199-8977d8a5-c1e4-4889-a970-1a02064e69ad.mov

### DIFF
--- a/src/tasks/components/TaskRunsRow.tsx
+++ b/src/tasks/components/TaskRunsRow.tsx
@@ -9,7 +9,7 @@ import RunLogsOverlay from 'src/tasks/components/RunLogsList'
 import {FormattedDateTime} from 'src/utils/datetime/FormattedDateTime'
 
 // Actions
-import {getLogs, retryTask} from 'src/tasks/actions/thunks'
+import {getLogs, retryTask, getRuns} from 'src/tasks/actions/thunks'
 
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):
looks good to me!

are you having issues with sorting on the tasks runs page?

PR #2101, opened 2021-07-21
fix: #1830 toggling time fix

### LINKED ISSUE #1830: Bug: Dashboard UTC/Local toggle should not change the time range.
# Summary

When using a Custom Time Range, switching between `Local` and `UTC` timezones changes the absolute time range displayed. It should only change the timezone; the start and stop times should not change. If you have a relative time like last `1hr` and change the timezone button, the displayed times simply update to reflect the new timezone. For custom time ranges, the whole cell reloads. It should do the same as the relative time ranges.

custom time range tz

**Steps to reproduce:**

1. Create a dashboard with at least one cell returning data from a valid query.
2. Observe the dashboard is set to `Local` (default) or `UTC` timezone.
3. Select a `Custom Time Range` that returns data (see also https://github.com/influxdata/ui/issues/1827)
4. Observe the times displayed in the cell for your time range.
5. Select the other timezone (Local -> UTC or UTC -> Local)
6. Observe that the cell reloads as the absolute time range for start and stop has been modified.

**Expected behavior:**
Changing the displayed timezone should not change the absolute start and stop times. The "units" should just be updated.

**Actual behavior:**
See above: Cells reload with new start and stop times when changing the displayed timezone.

**Environment info:**
influx cloud2

jrenee42:
Closes #1830 
checks if the time string actually needs converting before converting it.

keeping conversion in if necessary for future timezone/flux work.

### DIFF
--- a/src/dashboards/selectors/index.ts
+++ b/src/dashboards/selectors/index.ts
@@ -28,12 +28,39 @@ export const getTimeRange = (state: AppState): TimeRange => {
   return state.ranges[contextID] || DEFAULT_TIME_RANGE
 }
 
+/**
+ *  is the string a valid UTC time without any time zone information?
+ *  ie:  "2021-07-17T14:00:00.000Z" is valid
+ *  if it has any timezone offset, then it is not a valid UTC time string
+ *
+ * @param str
+ */
+function isUtcTime(str) {
+  const regex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/
+  return regex.test(str)
+}
+
+/** is the time already in utc format?  if so it does not need converting. */
+const timesNeedConverting = newTimeRange => {
+  // if the time range is *already* utc, does not need converting.  check just the first part, as
+  // they will be formatted the same
+  return !isUtcTime(newTimeRange.lower)
+}
+
+/**
+ * if the custom time is already in utc mode, don't convert it
+ * not removing the conversion clause entirely as timezone enabled flux is on the horizon
+ **/
 export const getTimeRangeWithTimezone = (state: AppState): TimeRange => {
   const timeRange = getTimeRange(state)
   const timeZone = getTimeZone(state)
 
   const newTimeRange = {...timeRange}
-  if (timeRange.type === 'custom' && timeZone === 'UTC') {
+  if (
+    timeRange.type === 'custom' &&
+    timesNeedConverting(newTimeRange) &&
+    timeZone === 'UTC'
+  ) {
     // conforms dates to account to UTC with proper offset if needed
     newTimeRange.lower = setTimeToUTC(newTimeRange.lower)
     newTimeRange.upper = setTimeToUTC(newTimeRange.upper)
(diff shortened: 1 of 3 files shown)


src/dashboards/selectors/index.ts
```diff
@@ -28,12 +28,39 @@ export const getTimeRange = (state: AppState): TimeRange => {
...
+function isUtcTime(str) {
+  const regex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/
+  return regex.test(str)
+}
+
+/** is the time already in utc format?  if so it does not need converting. */
+const timesNeedConverting = newTimeRange => {
+  // if the time range is *already* utc, does not need converting.  check just the first part, as
+  // they will be formatted the same
+  return !isUtcTime(newTimeRange.lower)
```
hoorayimhelping, line 47: let's move these into `src/shared/utils/datetime` since they're date/time utils. not sure how we should name the file they belong in, any ideas?

also how about some test cases to try to catch some edge cases?

src/dashboards/selectors/index.ts
```diff
@@ -28,12 +28,39 @@ export const getTimeRange = (state: AppState): TimeRange => {
...
 }
 
+/**
+ *  is the string a valid UTC time without any time zone information?
+ *  ie:  "2021-07-17T14:00:00.000Z" is valid
+ *  if it has any timezone offset, then it is not a valid UTC time string
+ *
+ * @param str
+ */
+function isUtcTime(str) {
```
hoorayimhelping, line 38: what do you think of tweaking `str` (which is kind of obvious) to something like `formattedTime` or something similar to indicate that it's a string in the form of `2021-07-17T00:00:00.000Z`?  i know it's in the comment, but some people don't read those, and they get erased in sourcemaps

hoorayimhelping (changes requested):
awesome find! i have a request to put them in a different and add tests

src/dashboards/selectors/index.ts
jrenee42: dateTimeUtils.ts should be fine.  

there is already unit tests for the conversion function.  

i can add to make sure the case with custom === true and needs conversion (is both true and false) are in the unit tests.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2101)

hoorayimhelping (approved):
LGTM!

one opinion: i'd prefer the git history have a list of commit messages that explain what they do, regardless of outside context. do you think you could punch up the commit message a little bit, and move the ticket number into the commit message description please?

jrenee42:
> LGTM!
> 
> one opinion: i'd prefer the git history have a list of commit messages that explain what they do, regardless of outside context. do you think you could punch up the commit message a little bit, and move the ticket number into the commit message description please?

sure, the ticket is already there but I can put more context into the messages.

PR #2114, opened 2021-07-22
fix: global search css

ShmuelLotman:
Closes #

Changes the import and testing against the prod image shows this worked, so here's hoping

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -26,7 +26,7 @@ import {disablePresentationMode} from 'src/shared/actions/app'
 // Styles
 import 'src/style/chronograf.scss'
 import '@influxdata/clockface/dist/index.css'
-import '@docsearch/react/style'
+import '@docsearch/css'
 const rootNode = getRootNode()
 
 const SESSION_KEY = 'session'


Asalem1 (approved):

hoorayimhelping:
The purpose of a commit message is to explain what the commit does without requiring someone to read a `git log` to inspect the change. When tracking down bugs that get reported, I often go through [recent commits](https://github.com/influxdata/ui/commits/master) to see if anything changed that might be a culprit. A commit message like `easy does it` forces me (or anyone else) to inspect the change because it doesn't provide enough context into what the change is. I'm all for whimsy and having fun, but not at the expense of the whole existence of the thing we're having fun with (a message that explains the change).

In the future you can try to summarize the change that happens in the commit please? Even the branch name `fix global search css` would be better because it clues a reader in that this is a CSS change, and probably won't be the cause of a logic bug.

This article explains the why really well https://chris.beams.io/posts/git-commit/

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2114)

PR #2125, opened 2021-07-23
feat: adds selector method to return the user's preferred time format for a view

### LINKED ISSUE #2122: Dashboards: getTimeFormatForCell
We need to know the user's preferred time format for this cell when displaying annotation dates.

### Acceptance Criteria
- there's a selector in `views/selectors` that gets the current time format, or the default if it's `''`

ChitlangeSahas:
Closes #2122

Notes for the reviewer:

* the issue we made (#2122) asks for a method called `getTimeFormatFromCell`, but since we were putting that method in the `views/selector`, I thought we should name it as `getTimeFormatFromView`. Not sure what the hierarchy is, is `Cell` a `View`? i noticed that the `cellID` and the view's `id` were the same. 

Also since the `timeFormat` was in `resources.views.view.properties.timeFormat` 

Added the unit tests to test the function as well.

### DIFF
--- a/src/views/selectors/index.ts
+++ b/src/views/selectors/index.ts
@@ -3,6 +3,7 @@ import {AppState, View, ResourceType, Dashboard} from 'src/types'
 
 // Selectors
 import {getByID} from 'src/resources/selectors'
+import {DEFAULT_TIME_FORMAT} from 'src/shared/constants'
 
 export const getViewsForDashboard = (
   state: AppState,
@@ -21,4 +22,22 @@ export const getViewsForDashboard = (
   )
 
   return views
+}
+
+export const getTimeFormatForView = (
+  state: AppState,
+  viewID: string
+): string => {
+  const view = getByID<View>(state, ResourceType.Views, viewID)
+
+  let timeFormat = ''
+  if ('timeFormat' in view.properties) {
+    timeFormat = view.properties.timeFormat
+  }
+
+  if (timeFormat === '') {
+    return DEFAULT_TIME_FORMAT
+  }
+
+  return timeFormat
 }
--- a/src/views/selectors/index.ts
+++ b/src/views/selectors/index.ts
@@ -30,7 +30,6 @@ export const getTimeFormatForView = (
 ): string => {
   const view = getByID<View>(state, ResourceType.Views, viewID)
 
-  let timeFormat = ''
   if ('timeFormat' in view.properties) {
     return view.properties.timeFormat
   }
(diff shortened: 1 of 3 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2125)

src/views/selectors/index.ts
```diff
@@ -22,3 +23,21 @@ export const getViewsForDashboard = (
...
+  let timeFormat = ''
+  if ('timeFormat' in view.properties) {
+    timeFormat = view.properties.timeFormat
+  }
+
+  if (timeFormat === '') {
+    return DEFAULT_TIME_FORMAT
+  }
+
+  return timeFormat
```
hoorayimhelping, line 42: we can make this a little more compact if you like:
```suggestion
  if ('timeFormat' in view.properties) {
    return view.properties.timeFormat
  }

  return DEFAULT_TIME_FORMAT
```

basically, rather than storing the value into a variable and returning it later, return as soon as we're certain we have a good value. otherwise return the default

hoorayimhelping (commented):
looks good! have a stylistic thing to make the code a little bit simpler.

make sure you test this with different date formats against annotations to make sure it allows different date formats. i think we only currently validate that the format matches the default

src/views/selectors/index.ts
ChitlangeSahas: I'll test this on annotations and then make this change 👍

hoorayimhelping:
@ChitlangeSahas I think a view is supposed to be a generic concept and a cell is an implementation of that concept. I _think_ that a view editor overlay (the thing that opens when you configure a dashboard cell) is also a view. So within our system, a view seems to be a generic rendering target.

ChitlangeSahas:
> make sure you test this with different date formats against annotations to make sure it allows different date formats. i think we only currently validate that the format matches the default

Tested with annotations!

src/views/selectors/index.ts
ChitlangeSahas: I just had to tweak it a little to accommodate when timeFormat is `''`

hoorayimhelping (approved):
nice work!

PR #2126, opened 2021-07-23
feat: ui placeholder implementation for usage panel

ShmuelLotman:
Closes #

This is a placeholder implementation for the usage panel, and will only be shown once IDPE provides us with a count endpoint for resources, as currently our implementation requires us to pull in the resources themselves and do a count, which is super inefficient. 

Screen Shot 2021-07-23 at 9 32 26 AM

### DIFF
--- a/src/me/components/Resources.tsx
+++ b/src/me/components/Resources.tsx
@@ -3,6 +3,7 @@ import React, {FC} from 'react'
 
 // Components
 import UsagePanel from 'src/me/components/UsagePanel'
+import Support from 'src/me/components/Support'
 import {
   Panel,
   FlexBox,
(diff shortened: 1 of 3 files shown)


src/me/components/Resources.tsx
```diff
@@ -29,83 +30,104 @@ interface Props {
...
-    return (
-      <FlexBox
-        direction={FlexDirection.Column}
-        alignItems={AlignItems.Stretch}
-        stretchToFitWidth={true}
-        margin={ComponentSize.Small}
-      >
-        {isFlagEnabled('docSearchWidget') ? (
-          <Panel testID="documentation--panel">
+const ResourceLists: FC<Props> = ({me}) => {
```
Asalem1, line 33: super weird that this is being prop drilled down here, especially since the parent component isn't making use of the `me` property. If we're already refactoring this, maybe we could just grab that state within the component rather than getting it from the parent?

src/me/components/Resources.tsx
```diff
@@ -29,83 +30,104 @@ interface Props {
...
-              <DocSearchWidget />
+              <GetResources resources={[ResourceType.Dashboards]}>
+                <DashboardsList />
+              </GetResources>
             </Panel.Body>
           </Panel>
+        </>
+      )}
+      <Panel>
+        {isFlagEnabled('newUsagePanel') ? (
```
Asalem1, line 90:
```suggestion
        {CLOUD && isFlagEnabled('newUsagePanel') ? (
```

Asalem1 (approved):
Great job. I really like this piece of functionality as an at-a-glance usage preview.

nit: can we not prop drill the `me` since that doesn't seem to be used by the parent? Also, let's make sure that the functionality is scoped to Cloud (since the usage endpoint is a cloud only one)

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2126)

src/me/components/Resources.tsx
ShmuelLotman: Done

src/me/components/Resources.tsx
ShmuelLotman: Good point. I checked this out too and great catch, no reason to have this hanging out here.

PR #2129, opened 2021-07-23
feat: pinned items (DO NOT MERGE) 

ShmuelLotman:
Closes #

This is not mergeable until we have the route implementation for the pinned items routes, but i want to get this up for review.


src/dashboards/components/dashboard_index/DashboardCard.tsx
```diff
@@ -83,10 +93,11 @@ class DashboardCard extends PureComponent<Props> {
     )
   }
 
-  private handleUpdateDashboard = (name: string) => {
+  private handleUpdateDashboard = async (name: string) => {
     const {id, onUpdateDashboard} = this.props
 
     onUpdateDashboard(id, {name})
+    await updatePinnedItemByParam(id, {name})
```
Asalem1, line 100: since i'm unfamiliar with pinnedItems, why is it important to await this?

src/me/components/PinnedItems.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+import {useHistory} from 'react-router-dom'
+
+const PinnedItems: FC = () => {
+  const history = useHistory()
+  const followMetadataToRoute = useCallback(
+    (data: PinnedItem) => {
+      let routeToFollow = ''
+      switch (data.type) {
+        case PinnedItemTypes.Dashboard:
```
Asalem1, line 24: is a PinnedItemType different than a ResourceType? Do we want them to be different?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,158 @@
...
+    body: JSON.stringify(item),
+  })
+  return await added.json()
+}
+
+const updatePinnedItem = async (_id: string, item: Partial<PinnedItem>) => {
+  await fetch(``, {
+    method: 'PUT',
+    headers: {
+      'Content-Type': 'application/json',
```
Asalem1, line 72: I get that these are all placeholders, can we get a TODO with an issue linked to move this over to a generated route so we don't end up with UI code for this?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,158 @@
...
+  }, [])
+
+  return (
+    <PinnedItemsContext.Provider value={{pinnedItems, deletePinnedItemsHelper}}>
+      {children}
+    </PinnedItemsContext.Provider>
+  )
+}
+
+export default memo(PinnedItemsProvider)
```
Asalem1, line 158: Can we get error handling on this or a TODO for it?

Asalem1 (commented):
Some stuff to think about. I don't think we necessarily want to make the resource-based CRUD operations async since the user doesn't necessarily actively interact with the pinned items at that point.

src/dashboards/components/dashboard_index/DashboardCard.tsx
Asalem1: Do pinned items exist on the home page or are they pinned resources, like a dashboard pinned to a list of dashboards, a cell pinned to that dashboard, a panel pinned to a notebook, a notebook pinned to a notebook list?

ShmuelLotman: its on the homepage ill post a screenshot

src/me/components/PinnedItems.tsx
ShmuelLotman: yes, because the concept of a pinned item transcends resources

src/shared/contexts/pinneditems.tsx
ShmuelLotman: yup, sure thing

src/dashboards/components/dashboard_index/DashboardCard.tsx
Asalem1: no worries. If that's the case do we need to make this awaited? Like, the user will still be on whatever resource, the promise should resolve when they get to the homepage / the home page will load up the resources when navigated to right? Seems minor, I don't know the implications of keeping this async or not but my impression of an async/await would have ramifications for the existing view and not some other part of the app that's not connected to the current state of affairs?

ShmuelLotman: thats a fair point, i was thinking of it from the perspective of whether or not the user would immediately expect it to show, and ramifications if it didn't show RIGHT away when they click on the page, but i guess thats both not an immediate concern and not something that outweighs the potential detriment to other experiences. Ill adjust this, thanks for changing my thinking on this!

PR #2135, opened 2021-07-26
fix: force timeFormat to be DEFAULT_TIME_FORMAT when it is an empty string

### LINKED ISSUE #2123: Cells that haven't set their time format will display the default format incorrectly
### Background
The default time format is incorrectly displayed on visualizations. This is due to multiple errors in code. **This will only happen during display if the user hasn't explicitly set a time format option in the graph display options.**

The first error occurs in [`getFormatter`](https://github.com/influxdata/ui/blob/master/src/visualization/utils/getFormatter.ts#L29) trying to assign a default value to the property `timeFormat`. `timeFormat` is [by default, an empty string '',](https://github.com/influxdata/idpe/blob/master/chronograf/service/dashboard.go#L750) and will never be overwritten to the default value.

The second error occurs in concert with the first. If the column type is time, `timeFormat` is [coerced](https://github.com/influxdata/ui/blob/83a98721495f55284706652d9a996081aeaf35e2/src/visualization/utils/timeFormat.ts#L31) into a property that will default to `DEFAULT_TIME_FORMAT`. The code will then [check the format and convert the date into or out of 24 hour format.](https://github.com/influxdata/ui/blob/83a98721495f55284706652d9a996081aeaf35e2/src/visualization/utils/getFormatter.ts#L67) Unfortunately, the code doesn't refer to this recently coerced value, it refers to passed in `timeFormat`, which will be `''` if the user hasn't set the property. **The result of this is that if the user hasn't set the time format, it will incorrectly be 12 hour time when it should be 24 hour time.**

### Steps to reproduce
1. Create a dashboard
1. Create a cell and add a query
   1. Observe the time, it should be in the default format, 24 hour time
1. Save the cell
1. Look at the dashboard's timestamps, back to 12 hour time
1. Edit options, set the time format to default (` YYYY-MM-DD HH:mm:ss`), Save the cell
1. The dashboard time should correctly be our default

https://user-images.githubusercontent.com/146112/126722951-1d08a84d-4b18-43e0-8c58-1e0c3666645a.mov

ChitlangeSahas:
Closes #2123

Just a small bug fix, for when the API returns an empty string, not `undefined` for the timeFormat, we need to force it to be `DEFAULT_TIME_FORMAT`

### DIFF
--- a/src/visualization/utils/getFormatter.test.ts
+++ b/src/visualization/utils/getFormatter.test.ts
@@ -1,0 +1,113 @@
+import {getFormatter} from './getFormatter'
+import {DEFAULT_TIME_FORMAT} from '../../shared/constants'
+import {Base, TimeZone} from '../../types'
+
+// July 4th, 1983, 20:00:00 UTC
+const timestamp = 426196800000
+
+describe('getFormatter', () => {
+  it("should format correctly for base 10 and 'number' columntype", function() {
+    const formatterOptions = {
+      prefix: 'prefix ',
+      suffix: ' suffix',
+      base: '10' as Base,
+      format: true,
+    }
+
+    const formatter = getFormatter('number', formatterOptions)
+
+    const formattedValue = formatter(100000)
+
+    expect(formattedValue).toBe('prefix 100k suffix')
+  })
+
+  it("should format correctly for base '' 'number' columntype", function() {
+    const formatterOptions = {
+      prefix: 'prefix ',
+      suffix: ' suffix',
+      base: '' as Base,
+    }
+
+    const formatter = getFormatter('number', formatterOptions)
+
+    const formattedValue = formatter(100000)
+
+    expect(formattedValue).toBe('prefix 100k suffix')
+  })
+
+  it("should format correctly for base 2 'number' columntype", function() {
+    const formatterOptions = {
+      prefix: 'prefix ',
+      suffix: ' suffix',
+      base: '2' as Base,
+      format: false,
+    }
+
+    const formatter = getFormatter('number', formatterOptions)
+
+    const formattedValue = formatter(100000)
+
+    expect(formattedValue).toBe('prefix 100000 suffix')
+  })
+
+  // time column type
+  it('should format correctly for columnType = time, timeFormat = DEFAULT_TIME_FORMAT and timeZone = UTC ', () => {
+    const formatterOptions = {
+      timeZone: 'UTC' as TimeZone,
+      timeFormat: DEFAULT_TIME_FORMAT,
+    }
+
+    const formatter = getFormatter('time', formatterOptions)
+
+    const date = new Date(timestamp)
+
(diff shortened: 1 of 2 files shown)


src/visualization/utils/getFormatter.ts
```diff
@@ -30,6 +30,12 @@ export const getFormatter = (
     format,
   }: GetFormatterOptions = {}
 ): null | ((x: any) => string) => {
+  // timeFormat was being passed in an empty string, which meant TypeScript will not
+  // replace it with the default, so we need to do it here with a check.
+  if (timeFormat === '') {
```
ChitlangeSahas, line 35: I realize I am modifying the passed-in function parameter, it seemed like a better idea than creating a new variable and naming it something else. But I can change the approach if this is not acceptable.

TCL735 (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2135)

src/visualization/utils/getFormatter.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    const formatter = getFormatter('number', formatterOptions)
+
+    const formattedValue = formatter(100000)
+
+    expect(formattedValue).toBe('prefix 100k suffix')
+  })
+
+  it("should format correctly for base '' 'number' columntype", function() {
+    const formatterOptions = {
```
ChitlangeSahas, line 25: I also added tests for the other part of the getFormatter, which isn't connected to this PR's fix (time related formatting), so if we want to add those tests as a separate PR I can do that!

hoorayimhelping: nice!

hoorayimhelping (approved):
1 line code fix, 143 lines of tests. nice

src/visualization/utils/getFormatter.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    const formatter = getFormatter('time', formatterOptions)
+
+    const date = new Date(timestamp)
+
+    const formattedTime = formatter(date)
+
+    expect(formattedTime).toBe('1983/07/04 20:00:00')
+  })
+
+  it('should format correctly for columnType = time , timeFormat = YYYY/MM/DD HH:mm:ss and timeZone = Local', () => {
```
hoorayimhelping, line 129: i think you'll have to skip this test for now, CI is always in UTC timezone and i haven't looked into how to set a specific timezone for a test

ChitlangeSahas: Got it!

PR #2136, opened 2021-07-26
fix: only add v vars and not func vars

### LINKED ISSUE #2084: Feature `filterExtern` with AST parsing does not process query dashboard variables correctly.
## Summary

Flux query variables that populate in dashboards are not extracted correctly with the `filterExtern` feature flag is enabled. Specifically, extra expressions that look like variables and have the same name as a variable are also added to the dashboard. 

This has several issues

1. it's confusing
2. query variables are evaluated before dashboard queries adding latency to dashboards.

The `filterExtern` flag was added here: https://github.com/influxdata/ui/pull/1633 Disabling the flag reverts to a regex that appears to work although it detects variables even if commented out.

The problematic code is in the ASTIM parsing. The code looks for tree nodes of type `Identifier` which is the type used for both variables like `v.windowPeriod` AND function arguments like in `filter(fn (r) => r.foo == "bar")`. If `foo` happens to be the name of a variable, then the `foo` variable is added to the dashboard even though it isn't used in any flux query.

https://github.com/influxdata/ui/blob/ac755f005c9a497497fc4c5fe2276e579ccae20e/src/variables/utils/astim.ts#L14-L16

I don't know if the AST labels ui variables with an identifier different than functional arguments. 

Of note: the default approach with a regex seems to work ok as long as the variable isn't in comments and the `v` injected record isn't renamed in the flux like `x=v`.
https://github.com/influxdata/ui/blob/ac755f005c9a497497fc4c5fe2276e579ccae20e/src/variables/utils/hydrateVars.ts#L81-L87

**Steps to reproduce:**

1. Create 2 variables, one `bar` and one `foo`. (any type works - pick csv/list)
2. With `filterExtern ` true, create a dashboard with a single cell and this flux (use a real bucket).
```
from(bucket: "bucket")
  |> range(start: -1m, stop: now())
  |> filter(fn: (r) => r.foo == "foo") // incorrectly parsed variable r.foo
  |> filter(fn: (r) => r.foobar == v.bar) // correctly parsed variable v.bar
```
3. Save and reload the query and notice that both variables `foo` and `bar` show up in the dashboard header area.

**Expected behavior:**
Only `v.` variables in flux queries produce variable widgets.

**Actual behavior:**
Extra variable dropdowns appear

**Environment info:**
cloud2

subirjolly:
Closes #2084

**Problem**
As described in #2084, ASTIM has been including all the variables found in the query regardless of their source object.

**Proposed Solution**
ASTIM only reads variables from the system object(`v`). There's a caveat that if someone renames `v` to something else, for example: `x=v`, then this fix is going to ignore all the `x.VARS`.

### DIFF
--- a/src/variables/utils/astim.ts
+++ b/src/variables/utils/astim.ts
@@ -12,7 +12,7 @@ export interface ASTIM {
 }
 
 const parseAllVariables = (ast: File): MemberExpression[] => {
-  return findNodes(ast, node => node?.property?.type === 'Identifier')
+  return findNodes(ast, node => node?.property?.type === 'Identifier' && node?.object?.name === 'v')
 }
 
 export const parseASTIM = (query: string): ASTIM => {
--- a/src/variables/utils/astim.ts
+++ b/src/variables/utils/astim.ts
@@ -12,7 +12,10 @@ export interface ASTIM {
 }
 
 const parseAllVariables = (ast: File): MemberExpression[] => {
-  return findNodes(ast, node => node?.property?.type === 'Identifier' && node?.object?.name === 'v')
+  return findNodes(
+    ast,
+    node => node?.property?.type === 'Identifier' && node?.object?.name === 'v'
+  )
 }
 
 export const parseASTIM = (query: string): ASTIM => {


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2136)

drdelambre (approved):

PR #2154, opened 2021-07-27
fix: throw error when maps coordiates are NaN

### LINKED ISSUE #2096: Maps: Selecting an invalid lat or lon column when configuring map leads to typeerror
to repro:
1. open data explorer
2. use the query below
3. run the query
4. switch to map
5. open map options
6. toggle off use s2_cell_id
7. choose lat for one column and some other column for lon (one that doesn't have a valid lon)
8. see the error below

```
import "influxdata/influxdb/schema"
import "experimental/csv"

csv.from(url: "https://raw.githubusercontent.com/russorat/influxdb-data-examples/master/iss/iss-historical-annotated.csv") 
  |> schema.fieldsAsCols()
```

image

Palakp41:
Closes #2096 

The reason maps was failing with `invalid Latlng object` is because we are trying to render Leaflet through giraffe even when lat/lon doesn't match the type of coordinates. So in this pr, I have added if the column chosen by the user is not type number, then throw an error instead of rendering giraffe and failing horrible..

### DIFF
--- a/src/shared/utils/vis.test.ts
+++ b/src/shared/utils/vis.test.ts
@@ -286,4 +286,18 @@ describe('getGeoCoordinate', () => {
 
     expect(() => getGeoCoordinates(table, 0)).toThrow()
   })
+  it('returns an error when Coordinate column is not anumber', () => {
+    const table = ({
+      getColumn: columnName => {
+        if (columnName === '_field') {
+          return ['lat', 'type', 'lat', 'type']
+        } else if (columnName === '_value') {
+          return [5, "random", 7, "string"]
+        }
+        return null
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2154)

hoorayimhelping (approved):

PR #2158, opened 2021-07-27
feat: add flow metadata to index page

### LINKED ISSUE #2142: Notebooks: Update Notebooks List Page to Display Created Date Time and Last Modified
We would like to update the Notebooks list page to display a Created Date Time and Last Modified (Displayed as "Last Updated") in minutes / hours / days etc. 

We can use the existing inline list item details UI to display the Created Date Time and Last Modified(Displayed as "Last Updated"):

**Created (replacing "Last completed" as "Created")**

**Last Modified (Displayed as "Last Updated")**

Created should be listed first (to the left) and then Last Modified (Displayed as "Last Updated") should come next.

## Acceptance Criteria

- As a user, when visiting the Notebooks list page I see Created and Last Modified (Displayed as "Last Updated") on the Notebooks list page

Asalem1:
Closes #2142 

This PR adds createdAt and updatedAt displays for the notebooks index page

flow-meta-data


src/flows/context/flow.list.tsx
```diff
@@ -173,6 +177,8 @@ export const FlowListProvider: FC = ({children}) => {
         readOnly: false,
         range: DEFAULT_TIME_RANGE,
         refresh: AUTOREFRESH_DEFAULT,
+        createdAt: new Date().toISOString(),
+        updatedAt: new Date().toISOString(),
```
Asalem1, line 181: not sure about applying defaults here. This should be from the API but we create and set the flow on the UI so maybe?

src/flows/context/flow.list.tsx
```diff
@@ -209,6 +215,8 @@ export const FlowListProvider: FC = ({children}) => {
         data: flow.data,
         meta: flow.meta,
         readOnly: flow.readOnly,
+        createdAt: flow.createdAt ?? new Date().toISOString(),
+        updatedAt: flow.updatedAt ?? new Date().toISOString(),
```
Asalem1, line 219: same as above, figured this could be a fail-safe in case the metadata doesn't exist?

src/flows/context/flow.list.tsx
```diff
@@ -296,7 +313,14 @@ export const FlowListProvider: FC = ({children}) => {
     const data = await getAllAPI(org.id)
     if (data && data.flows) {
       const _flows = {}
-      data.flows.forEach(f => (_flows[f.id] = hydrate(f.spec)))
+      data.flows.forEach(f => {
+        _flows[f.id] = {
+          ...hydrate(f.spec),
```
Asalem1, line 318: not sure whether we should update the spec to include the createdAt and updatedAt metadata, or if that's a different issue

src/types/flows.ts
```diff
@@ -119,6 +119,8 @@ export interface Flow {
   meta: ResourceManipulator<PipeMeta>
   results: FluxResult
   readOnly?: boolean
+  createdAt?: Date
+  updatedAt?: Date
```
Asalem1, line 123: not sure if this should be optional or if this should be required? Like, when we create a flow that data doesn't exist so it seems like it should be optional but 🤷🏽

src/flows/context/flow.list.tsx
drdelambre: yeah i think we should skip the defaults and conditionally render? like we're telling the user "this was last saved at XYZ" and if it was never saved......

Asalem1: Right totally. I guess by the time they actually access the data (after creation) that data will already be saved so it doesn't really make sense

src/types/flows.ts
drdelambre: updated should totes be updated... and thinking about how many times we have leaned on the offline first mechanism.... i feel like the created date should also be optional to communicate "this has been first saved to the system"

src/flows/components/FlowCard.tsx
```diff
@@ -5,13 +5,15 @@ import {useParams, useHistory} from 'react-router-dom'
...
+import {Flow} from 'src/types'
 
 interface Props {
   id: string
-  name: string
+  flow: Flow
 }
 
-const FlowCard: FC<Props> = ({id, name}) => {
+const FlowCard: FC<Props> = ({id, flow}) => {
```
drdelambre, line 15: i was just poking around in here... all of this prop drilling is.. a bit much. we should be either just accepting the id and using the flow.list context to grab the flow data, or not pass in the id and double down on prop drilling

Asalem1: Ya, I was a bit surprised about that the drilling but it seems like part of it is for filtering the sorted data. I can definitely get the flow based on the id though so I'll clean that up

src/flows/context/flow.list.tsx
```diff
@@ -222,7 +228,14 @@ export const FlowListProvider: FC = ({children}) => {
     }
     let id: string = `local_${UUID()}`
     try {
-      id = await createAPI(apiFlow)
+      const flow = await createAPI(apiFlow)
```
drdelambre, line 231: yeah.. this is all hamajang. i think you need to have the flow run through the hydrate function before adding it places... and delete all of the modifying of _flow in here. `_flow = hydrate(flow)` should centralize the hydration logic and remove some bug around resource hooks not updating

src/flows/context/flow.list.tsx
drdelambre: i think that's a different issue, but you should def normalize how + when + where these values are being loaded into the data... maybe open up the hydrate function to handle the response from the api?

Asalem1: Totally, I think I did open the hydrate function to do that, but in this case the spec doesn't contain an updatedAt and createdAt property. Should I just pass those properties in from the flow as something like:

```
hydrate({...f.spec, updatedAt: f.updatedAt, createdAt: f.createdAt})
```
or would the way to go be about updating the `spec`? I'm still fuzzy on where the source of truth for the spec is and if it's different to align with API expectations

drdelambre: no.. the api should have a concept of a "notebook object" type right? of which, this is taking a properties of.. in this case `spec`.. i'm saying, don't update `spec`, but update the hydrate function to consume that higher order "notebook object" whatever that is

Asalem1: Aw, gotcha. That makes sense...I think. I'll take a look at the code

Asalem1: Honestly after trying to rework this, I'm not sure if that's the solution. It just adds unnecessary complexity to the hydrate function. I think if we had a definitively typed object that was being passed in but it seems like we're using the hydrate function in a couple places and it might be outta scope to type it here

src/flows/context/flow.list.tsx
Asalem1: I totally hear what you're saying but I think this kind of goes along with the previous sentiment of trying to update hydrate to do all the things. It adds unnecessary complexity to the hydrate function that is entirely dependent on how a flow is structured. Updating this to be something like:

```
const notebook = await createAPI(apiFlow)
_flow = hydrate(notebook)
```

drops the local context that the DB has no concept about (like range, readOnly and refresh) which are properties of the spec (in this case not returned from the API). Let's sync up later

src/flows/context/flow.list.tsx
```diff
@@ -222,7 +228,15 @@ export const FlowListProvider: FC = ({children}) => {
...
+      id = flow.id
+
+      _flow = hydrate({
+        ...flow,
```
Asalem1, line 235: not sure if this is the solution to the above comment

src/flows/context/flow.list.tsx
```diff
@@ -296,7 +312,13 @@ export const FlowListProvider: FC = ({children}) => {
     const data = await getAllAPI(org.id)
     if (data && data.flows) {
       const _flows = {}
-      data.flows.forEach(f => (_flows[f.id] = hydrate(f.spec)))
+      data.flows.forEach(f => {
+        _flows[f.id] = hydrate({
+          ...f.spec,
+          createdAt: f.createdAt,
+          updatedAt: f.updatedAt,
+        })
```
Asalem1, line 320: same as the above

Asalem1 (commented):
TLDR; didn't update `hydrate` in favor of just updating the passed in params to avoid complexity within the hydrate function. If we wanna do that, it might be worth considering in a separate issue?

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2158)

src/flows/context/flow.list.tsx
```diff
@@ -254,6 +265,8 @@ export const FlowListProvider: FC = ({children}) => {
       meta: flow.meta,
       readOnly: flow.readOnly,
       results: null,
+      createdAt: flow.createdAt,
```
drdelambre, line 268: are these still needed?

Asalem1: in theory no

drdelambre:
https://github.com/influxdata/ui/pull/2158/files#diff-548571d7e56189763714fe7a4da3f42fda1df52fe103ef8b2a25862a9f8d02f6R277

why isn't this just the output of serialize?

src/flows/context/flow.list.tsx
```diff
@@ -246,32 +257,14 @@ export const FlowListProvider: FC = ({children}) => {
...
-        id,
-        orgID: org.id,
-        name: flow.name,
-        spec: serialize(data),
-      },
-    }
-    pooledUpdateAPI(apiFlow)
+    const apiFlow = serialize(flow, org.id)
+
+    pooledUpdateAPI({id, ...apiFlow})
```
Asalem1, line 267: i ended up not having serialize handle the id since a flow being serialized doesn't always have an ID and it was causing issues

drdelambre (approved):
I think I understand what is going on here. Thanks for normalizing the ingress/egress formats

PR #2183, opened 2021-07-29
fix: send correct type of value when updating domain for histograms and heat maps

### LINKED ISSUE #2182: Cannot change x axis domain in Histogram
__Steps to reproduce:__

1. Create an Histogram diagram
2. Change the values of X-Axis to custom using numbers like 1000 - 1000.0 - 1001,0 etc..
3. The diagram in the "edit" window change like I want it to change
4. Click save

Screenshot 2021-07-29 at 10 17 09

__Expected behavior:__
I was expecting the diagram to change accordingly in the dashboard view

__Actual behavior:__
The diagram do not show any data anymore and an error popup with  "Failed to update cell" appeared.

Screenshot 2021-07-29 at 10 18 19

Logs shown by the service are below.

__Environment info:__
InfluxDB 2.0.7 in a container with influxdb:latest image

__Logs:__

`ts=2021-07-29T08:18:13.244360Z lvl=error msg="api error encountered" log_id=0VcT6iHG000 handler=dashboards error="json: cannot unmarshal string into Go struct field HistogramViewProperties.xDomain of type float64"`

wbaker85:
Closes #2182

Updates the `setDomain` function for histograms and heat maps to correctly set the value as a number rather than a string.

I confirmed this works for saving a histrogram and heatmap on both OSS and remocal.

### DIFF
--- a/src/visualization/types/Histogram/options.tsx
+++ b/src/visualization/types/Histogram/options.tsx
@@ -56,13 +56,13 @@ const HistogramOptions: FC<Props> = ({properties, results, update}) => {
   }
 
   const setDomain = (axis: string, domain: [number, number]): void => {
-    let bounds: [string | null, string | null]
+    let bounds: [number | null, number | null]
 
     if (domain) {
       const [min, max] = domain
       bounds = [
-        min === null ? null : String(min),
-        max === null ? null : String(max),
+        min === null ? null : min,
+        max === null ? null : max,
       ]
     } else {
       bounds = [null, null]
--- a/src/visualization/types/Histogram/options.tsx
+++ b/src/visualization/types/Histogram/options.tsx
@@ -60,7 +60,7 @@ const HistogramOptions: FC<Props> = ({properties, results, update}) => {
 
     if (domain) {
       const [min, max] = domain
-      bounds = [min === null ? null : min, max === null ? null : max]
+      bounds = [min, max]
     } else {
       bounds = [null, null]
     }
--- a/src/visualization/types/Histogram/options.tsx
+++ b/src/visualization/types/Histogram/options.tsx
@@ -60,10 +60,7 @@ const HistogramOptions: FC<Props> = ({properties, results, update}) => {
 
     if (domain) {
       const [min, max] = domain
-      bounds = [
-        min === null ? null : min,
-        max === null ? null : max,
-      ]
+      bounds = [min === null ? null : min, max === null ? null : max]
     } else {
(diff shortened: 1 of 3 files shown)


wbaker85:
Might need to play around with this a little more to test out some edge cases

src/visualization/types/Histogram/options.tsx
```diff
@@ -56,14 +56,11 @@ const HistogramOptions: FC<Props> = ({properties, results, update}) => {
...
-    let bounds: [string | null, string | null]
+    let bounds: [number | null, number | null]
 
     if (domain) {
       const [min, max] = domain
-      bounds = [
-        min === null ? null : String(min),
-        max === null ? null : String(max),
-      ]
+      bounds = [min === null ? null : min, max === null ? null : max]
```
hoorayimhelping, line 63:
```suggestion
      bounds = [min, max]
```

now that we're not casting to a string, we don't need this anymore - it doesn't actually do anything. Looks like this was put in to prevent `new String(null)` which returns `"null"` rather than `null`

hoorayimhelping (changes requested):

src/visualization/types/Histogram/options.tsx
wbaker85: Nice...I went ahead and did that for both the Histogram and Heatmap, and also the Scatterplot which is exactly the same function. This does call to attention the rather unsatisfying fact that the same function is duplicated in 3 different places, but at least it's consistent!

wbaker85:
> Might need to play around with this a little more to test out some edge cases

Ran a variety of manual tests with IDPE and OSS backends and didn't see anything unusual. Could have min/max left blank, min > max, etc. and the cells would still save. Also I was able to save a cell with a custom domain, revert the UI code and then successfully open the dashboard with that cell and it appeared fine. The only thing that might happen in that situation from what I could tell is that you could not update the domain from an older (perhaps cached) version of the UI - would get a server error, but upon refreshing the page the cell would still work (just not be updated).

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2183)

hoorayimhelping (approved):
thanks for so thoroughly testing this and looking into it!

PR #2190, opened 2021-07-30
fix(ci): set is_failure for each failed job

genehynson:
actually set is_failure flag when we're looping over the failed job names


joelfried (approved):
:crossed_fingers: looks like this should do it.  Now let's just hope there's not too much damage mitigation to do.

PR #2195, opened 2021-07-30
feat: redesigned tokens resource card

### LINKED ISSUE #2102: Tokens: New Resource Card
In the tokens list, the resource cards currently looks like: 

We want the new resource card to look like:

Steps to implement this:
1.) Copy the `TokensList` component in the redesigned folder.
2.) Copy the `TokenRow` component in the redesigned folder.
3.) Update the `TokenRow` import in `redesigned/TokensList` to match the `redesigned/TokenRow`.
2.) Edit `redesigned/TokenRow` to implement the new resource card.
4.) Add component tests around the new resource card (example: expecting certain test ID's to be visible)

blegesse-git:
Closes #2102 

With token redesign feature flag on (influx.toggle('tokenUIRedesign')), the resource card looks like this. 

Screen Shot 2021-07-30 at 12 42 01 PM

### DIFF
--- a/src/authorizations/components/redesigned/TokenList.tsx
+++ b/src/authorizations/components/redesigned/TokenList.tsx
@@ -14,6 +14,9 @@ import {Sort} from '@influxdata/clockface'
 
 // Utils
 import {getSortedResources} from 'src/shared/utils/sort'
+import {incrementCloneName} from 'src/utils/naming'
+
+import { createAuthorization } from 'src/authorizations/apis';
 
 type SortKey = keyof Authorization
 
@@ -83,9 +86,30 @@ export default class TokenList extends PureComponent<Props, State> {
         key={auth.id}
         auth={auth}
         onClickDescription={this.handleClickDescription}
+        onClone={this.handleClone}
       />
     ))
   }
+
+  //
+  private handleClone = async (id: string) => {
+      const { auths } = this.props;
+
+      const authFound = auths.find((auth) => auth.id === id);
+
+      if (!authFound) return null;
+
+      const authsDescriptions = auths.map((auth) => auth.description);
+
+      const clonedAuth = {
+          ...authFound,
+          description: incrementCloneName(authsDescriptions, authFound.description),
+      };
+
+      console.log('Cloning results', clonedAuth);
+
+      await createAuthorization(clonedAuth);
+  };
 
   private handleDismissOverlay = () => {
     this.setState({isTokenOverlayVisible: false})
--- a/src/authorizations/components/redesigned/TokenList.tsx
+++ b/src/authorizations/components/redesigned/TokenList.tsx
@@ -1,0 +1,98 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import memoizeOne from 'memoize-one'
+
+// Components
+import {Overlay, ResourceList} from '@influxdata/clockface'
+import TokenRow from 'src/authorizations/components/redesigned/TokenRow'
+import ViewTokenOverlay from 'src/authorizations/components/ViewTokenOverlay'
+
+// Types
+import {Authorization} from 'src/types'
+import {SortTypes} from 'src/shared/utils/sort'
+import {Sort} from '@influxdata/clockface'
+
+// Utils
+import {getSortedResources} from 'src/shared/utils/sort'
+
+type SortKey = keyof Authorization
+
+interface Props {
+  auths: Authorization[]
+  emptyState: JSX.Element
+  searchTerm: string
+  sortKey: string
+  sortDirection: Sort
+  sortType: SortTypes
+  onClickColumn: (nextSort: Sort, sortKey: SortKey) => void
+}
+
+interface State {
+  isTokenOverlayVisible: boolean
+  authInView: Authorization
+}
+
+export default class TokenList extends PureComponent<Props, State> {
+  private memGetSortedResources = memoizeOne<typeof getSortedResources>(
+    getSortedResources
+  )
+
+  constructor(props) {
+    super(props)
+    this.state = {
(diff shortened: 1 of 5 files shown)


src/authorizations/components/redesigned/TokenRow.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    expect(getByTestId('delete-token')).toBeVisible()
+  })
+
+  it('executes the delete method', async () => {
+    const {getByTestId} = setup()
+
+    fireEvent.click(getByTestId('delete-token'))
+    await waitFor(() => expect(deleteAuthorization).toBeCalled())
+    expect(deleteAuthorization).toBeCalledTimes(1)
```
hoorayimhelping, line 55: right now, this test is a bit tautological. E.g. "call the delete method, was it called? yep it was called."

I think we can make it a little more valuable by asserting that the row we deleted is no longer there. I'm not sure if that's even possible without changing the store to 'simulate' a delete event, so this comment might be a little more theoretical in nature.

When we test things, we want to test the effect of the operation on the system. In this case, we want to test that when we click the delete button, we call the delete api method (:+1: on that) *and* that the deleted item no longer shows up in the data and the UI. Since this is a mocked, that last step is kind of difficult when we mock out the API layer. The general flow is
1. click delete
1. call API to delete stuff
1. API does API stuff
1. call API to get updated list of items
1. render new list without that deleted item

to do that without an actual API, we'd have to mock out steps 2 and 4 (3 is simulated). This test currently mocks out step 2. To really test this, we'd have to also mock out test 4, but that's still kind of fake.

Hope that explanation was useful and not a wall of obvious text

src/authorizations/components/redesigned/TokenList.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <TokenRow
+        key={auth.id}
+        auth={auth}
+        onClickDescription={this.handleClickDescription}
+        onClone={this.handleClone}
+      />
+    ))
+  }
+
+  private handleClone = () => {}
```
hoorayimhelping, line 91: nailed it haha.

I'm guessing the `clone` functionality is new and hasn't been implemented yet? if so, how about a comment saying that

src/authorizations/components/redesigned/TokenRow.tsx
```diff
@@ -0,0 +1,134 @@
...
+import {Context} from 'src/clockface'
+
+// Types
+import {Authorization} from 'src/types'
+import {
+  DEFAULT_TOKEN_DESCRIPTION,
+  UPDATED_AT_TIME_FORMAT,
+} from 'src/dashboards/constants'
+
+import {relativeTimestampFormatter} from 'src/shared/utils/relativeTimestampFormatter'
```
hoorayimhelping, line 34: do we not also want to use `createRelativeFormatter` from `'src/utils/datetime/formatters'` ?

src/authorizations/components/redesigned/TokenRow.tsx
```diff
@@ -0,0 +1,134 @@
...
+}
+
+const mdtp = {
+  onDelete: deleteAuthorization,
+  onUpdate: updateAuthorization,
+}
+
+const connector = connect(null, mdtp)
+
+export default connector(TokenRow)
```
hoorayimhelping, line 134:
```suggestion
export const TokenRow = connector(TokenRow_)
```
and then you'd have to rename this to `TokenRow_`

hoorayimhelping (commented):
awesome work!!

I have a few minor suggestions about formatting

src/authorizations/components/redesigned/TokenList.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  private handleDismissOverlay = () => {
+    this.setState({isTokenOverlayVisible: false})
+  }
+
+  private handleClickDescription = (authID: string): void => {
+    const authInView = this.props.auths.find(a => a.id === authID)
```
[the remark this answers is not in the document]
blegesse-git: that's a really good point, bucky. It seems harmless now but it might be a headache to understand later down the line.  It's updated to be more descriptive now :)

src/authorizations/components/redesigned/TokenList.tsx
blegesse-git: lol yes. i initially had the clone functionality here but realized there is another ticket dedicated to solve that problem.

src/authorizations/components/redesigned/TokenRow.tsx
blegesse-git: hmm..  `relativeTimestampFormatter` is actually using `createRelativeFormatter` in order to return the date in _from now_ format. here is what `relativeTimestampFormatter` looks like
Screen Shot 2021-07-30 at 5 02 32 PM

hoorayimhelping: haha, I can't even remember my own code that I wrote 👴

src/authorizations/components/redesigned/TokenRow.test.tsx
hoorayimhelping: like we talked about on a Zoom call, this case is covered by an e2e test, so we should just remove this

hoorayimhelping (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2195)

PR #2196, opened 2021-07-30
chore: make room for query scopes

drdelambre:
based on a conversation a bit ago with @Asalem1 and some experiments i ran last month, we found that by making a few differences in the way a notebook generates it's queries, we can open up the ability for notebooks to query other regions, query multiple time ranges, or support scoped variables. In order to reduce the surface area of these changes, the collapsotron feature was removed as well as moving the printMap function into the query context where it should have been in the first place.

Changes to the interface:
generateFlux(random stuff) has been split apart into scope, visual, and source. the create vs append distinction wasn't clear to devs using it, so instead of adding MORE transformation functions, they were split into separate definitions (where visual is the old append and source is the old create). Seems like we dont need the distinction between source and visual, but i figured that should be another PR.
scope: any panel can now add whatever it wants to a scope object that is then passed to successive panels when generating their queries. this allows one panel to set a variable, and have the following panels use it.
the concept of stages has been removed. there was a mechanism where panels that repeated queries were joined into "stages" so that the query could be run once, and the result could be shared amongst all instances that used it. after we split the distinction between "source queries" and "visual queries".. the need for this sort of deduping fell to the wayside, as we were no longer firing every query, every time. any other query deduping should probably be handled by a query cache from this point. now every notebook panel has it's own query definition, and it's still called a stage (because it roughly means the same thing to the rest of the application, just not to the query mechanism), but no more crazy mappings


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2196)

src/flows/components/panel/FlowPanel.tsx
```diff
@@ -206,7 +148,7 @@ const FlowPanel: FC<Props> = ({
                 <FeatureFlag name="flow-debug-queries">
                   <SquareButton
                     icon={IconFont.BookCode}
-                    onClick={printMap}
+                    onClick={() => printMap(id, true)}
```
Asalem1, line 151: why wouldn't `printMap` just be handling true cases?

src/flows/context/flow.query.tsx
```diff
@@ -6,57 +6,56 @@ import {RunModeContext, RunMode} from 'src/flows/context/runMode'
...
 
 // Types
 import {AppState, RemoteDataState, Variable} from 'src/types'
 
-interface Instance {
-  id: string
-  modifier?: string
-}
-
 export interface Stage {
```
Asalem1, line 27: "got rid of stages" huh?

src/flows/context/flow.query.tsx
```diff
@@ -144,110 +141,121 @@ export const FlowQueryProvider: FC = ({children}) => {
...
+      acc.push(meta)
+      return acc
+    }, [])
+
+    return stages
   }
 
-  const query = (text: string): Promise<FluxResult> => {
+  // TODO figure out a better way to cache these requests
+  const getPanelQueries = (id: string, withSideEffects?: boolean): Stage => {
```
Asalem1, line 197: if only there was a query cache we could use

src/types/flows.ts
```diff
@@ -135,6 +135,16 @@ export interface FlowList {
...
 }
 
+export interface VariableMap {
+  [key: string]: Variable
+}
+
+// TODO: type this better. there are required properties, that have types, but
+// we also need this to stay open for panels to register whatever they want
+export interface QueryScope {
+  [props: string]: any
```
Asalem1, line 145: nice. This might be one of those `type` vs `interface` things?

Asalem1 (approved):
Looks pretty straightforward in terms of the distinction, i'm just wondering whether there are ramifications for sidebar in how it now relates to a given panel query (not sure if those were related) and whether this changes the interaction between a pipe's query and it's visualization? I'm tired so I might not be thinking clearly here but for some reason my mental model isn't up to snuff

src/flows/context/flow.query.tsx
drdelambre: I PUT A DISCLAIMER IN THE DESCRIPTION

src/flows/components/panel/FlowPanel.tsx
drdelambre: in case you wanna test it with the actual notebook state

src/types/flows.ts
drdelambre: man i hope someone figures this out. it was super confusing and googling anything about typescript just gives you "in typescript, === tests equality AND type"

Asalem1: what happens if you create a general QueryScope interface that you then extend and add required properties to it?

drdelambre: Tried that. Same error

PR #2203, opened 2021-08-01
feat: make date time formatter hourCycle h23; minor refactor

### LINKED ISSUE #2202: Make hour cycle for the new date-time formatter to be h23
We need make the hour cycle for the new formatter to be h23. Which means instead of `01:00 - 24:59` it will go from `00:00 - 23:59`.

For reference: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle

https://en.wikipedia.org/wiki/ISO_8601#Times

>As of ISO 8601-1:2019 midnight may only be referred to as "00:00", corresponding to the beginning of a calendar day. Earlier versions of the standard allowed "24:00" corresponding to the end of a day, but this is explicitly disallowed by the 2019 revision.

ChitlangeSahas:
Closes #2202

This PR adds the functionality to make the date time formatter's hourCycle to be h23. Currently by default, its h24, which means the time formatting formats in the range : `01:00 - 24:59` 

The more familiar one, h23, formats it from `00:00 - 23:59`. This was the app's default formatting as well when using `momentjs`

Some reviewer notes: 

* I refactored the settings for the formatter, previously, the `dateTimeOptions` had the following parameters:

```
const dateTimeOptions: any = {
  hour12: true,
  day: '2-digit',
  month: '2-digit',
  year: 'numeric',
  hour: 'numeric',
  minute: 'numeric',
  second: 'numeric',
}
```

Notice, the `hour12`, set to `true` by default, this felt like it was going against our `DEFAULT_TIME_FORMAT` which says 24h is the default. So instead of turning all the `hour12` to `false` where 24 hour was needed, I turned on 24 hour by default, and explicitly added `hour12: true` where needed.

--

Screenshot example, when we use moment this is what the formatting looks like: 

Screen Shot 2021-08-01 at 7 06 42 PM

When we switch it out with the existing, in progress date-time formatter, the same time looks like this: 

Screen Shot 2021-08-01 at 7 21 54 PM

### DIFF
--- a/src/utils/datetime/formatters.test.ts
+++ b/src/utils/datetime/formatters.test.ts
@@ -71,7 +71,7 @@ describe('the DateTime formatter', () => {
       expect(formatter.format(date)).toBe(`1983-07-04 ${hourUTC24}:00:00`)
     })
 
-    it('formats DateTimes in the default time YYYY-MM-DD HH:mm:ss in UTC, for midnight timestamp', () => {
+    it('formats DateTimes in the default time YYYY-MM-DD HH:mm:ss in UTC, format midnight as 00:00 not 24:00', () => {
       const date = new Date(timestamp)
       date.setHours(date.getHours() + 4) // 20:00 + 4 hrs = Midnight
       const formatter = createDateTimeFormatter('YYYY-MM-DD HH:mm:ss', 'UTC')
--- a/src/utils/datetime/formatters.test.ts
+++ b/src/utils/datetime/formatters.test.ts
@@ -71,6 +71,13 @@ describe('the DateTime formatter', () => {
       expect(formatter.format(date)).toBe(`1983-07-04 ${hourUTC24}:00:00`)
     })
 
+    it('formats DateTimes in the default time YYYY-MM-DD HH:mm:ss in UTC, for midnight timestamp', () => {
+      const date = new Date(timestamp)
+      date.setHours(date.getHours() + 4) // 20:00 + 4 hrs = Midnight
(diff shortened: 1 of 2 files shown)


hoorayimhelping (commented):
Great work! Good job finding this and fixing it.

One request: can you add a test that says that we format midnight as 00:00 not 24:00 please?

ChitlangeSahas:
> Great work! Good job finding this and fixing it.
> 
> One request: can you add a test that says that we format midnight as 00:00 not 24:00 please?

@hoorayimhelping yes sir! Will figure out a way to do it, right away! our dummy timestamp is at 20:00 so I wasn't sure if I should make a new one, which is called `midNightTimeStamp`.  I'll think of a way 👍

hoorayimhelping:
@ChitlangeSahas it's fine to create a new timestamp for this test. You can also add 14400 seconds (4 hours in seconds) to the existing one

ChitlangeSahas:
> @ChitlangeSahas it's fine to create a new timestamp for this test. You can also add 14400 seconds (4 hours in seconds) to the existing one

@hoorayimhelping Got it!!

hoorayimhelping (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2203)

PR #2214, opened 2021-08-02
fix: add commas to number formatting in table

jrenee42:
Closes #17456

uses the formatter already in place from the single stat layer.

### DIFF
--- a/src/visualization/types/Table/TableCell.tsx
+++ b/src/visualization/types/Table/TableCell.tsx
@@ -15,6 +15,7 @@ import {
   TableViewProperties,
 } from 'src/types/dashboards'
 import {PropsMultiGrid} from 'src/visualization/types/Table/MultiGrid'
+import {formatStatValue} from '@influxdata/giraffe'
 
 export interface CellRendererProps {
   columnIndex: number
@@ -272,12 +273,9 @@ class TableCell extends PureComponent<Props> {
       return (this.fieldName || '').toString()
     }
 
-    if (
-      !isNaN(+data) &&
-      decimalPlaces.isEnforced &&
-      decimalPlaces.digits < 100
-    ) {
-      return (+data).toFixed(decimalPlaces.digits)
+    if (!isNaN(+data)) {
+      // method needs the first arg to be a number to work properly
+      return formatStatValue(+data, {decimalPlaces})
     }
 
     return (data || '').toString()
--- a/src/visualization/types/Table/view.tsx
+++ b/src/visualization/types/Table/view.tsx
@@ -17,6 +17,7 @@ import EmptyGraphMessage from 'src/shared/components/EmptyGraphMessage'
 // Utils
 import {tableFromFluxResult} from 'src/shared/parsing/flux/response'
 import {AppSettingContext} from 'src/shared/contexts/app'
+import {parseFromFluxResults} from 'src/timeMachine/utils/rawFluxDataTable'
 
 // Types
 import {TableViewProperties} from 'src/types'
@@ -27,6 +28,13 @@ import {
   DEFAULT_SORT_DIRECTION,
   KEYS_I_HATE,
 } from './constants'
+import {
+  Config,
+  DEFAULT_TABLE_COLORS,
+  HoverTimeProvider,
+  Plot,
+} from '@influxdata/giraffe'
+import {isFlagEnabled} from '../../../shared/utils/featureFlag'
 
 interface Props extends VisualizationProps {
   properties: TableViewProperties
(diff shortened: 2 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2214)

src/visualization/types/Table/TableCell.tsx
```diff
@@ -272,12 +273,9 @@ class TableCell extends PureComponent<Props> {
...
 
-    if (
-      !isNaN(+data) &&
-      decimalPlaces.isEnforced &&
-      decimalPlaces.digits < 100
-    ) {
-      return (+data).toFixed(decimalPlaces.digits)
+    if (!isNaN(+data)) {
+      // method needs the first arg to be a number to work properly
+      return formatStatValue(+data, {decimalPlaces})
```
hoorayimhelping, line 278: do you know what's with these casts to numbers? do you think it would be more idiomatic and communicative to do a `parseInt(data, 10)` (or `parseFloat(data)` if it's floating point)?

src/visualization/types/Table/view.tsx
```diff
@@ -27,6 +28,13 @@ import {
   DEFAULT_SORT_DIRECTION,
   KEYS_I_HATE,
 } from './constants'
+import {
+  Config,
+  DEFAULT_TABLE_COLORS,
+  HoverTimeProvider,
+  Plot,
+} from '@influxdata/giraffe'
+import {isFlagEnabled} from '../../../shared/utils/featureFlag'
```
hoorayimhelping, line 37:
```suggestion
import {isFlagEnabled} from 'src/shared/utils/featureFlag'
```

webpack aliases `src` for us so we don't need to use relative paths. I think Sahas also had an issue with this with WebStorm

src/visualization/types/Table/view.tsx
```diff
@@ -81,71 +89,145 @@ const TableGraphs: FC<Props> = ({properties, result}) => {
...
+          tableTheme: theme,
+        },
+      ],
+    }
+    return (
+      <HoverTimeProvider>
+        <Plot config={config} />
+      </HoverTimeProvider>
+    )
+  } else {
```
hoorayimhelping, line 164:
```suggestion
```

we're returning in the `if` statement; ditching the `else` will leave this component's default/happy path as the main pathway

hoorayimhelping (changes requested):
the whole point of this was to add commas to numbers > 1000, right? can you write a test for that case here please?

src/visualization/types/Table/TableCell.tsx
```diff
@@ -272,12 +273,9 @@ class TableCell extends PureComponent<Props> {
       return (this.fieldName || '').toString()
     }
 
-    if (
-      !isNaN(+data) &&
-      decimalPlaces.isEnforced &&
-      decimalPlaces.digits < 100
-    ) {
-      return (+data).toFixed(decimalPlaces.digits)
+    if (!isNaN(+data)) {
```
TCL735, line 276: This is clearly an edge case. Be careful with `isNaN` which is short for `window.isNaN`. Especially in combination with the casting operator here. You are coercing non-numbers into numbers and then checking if they are `NaN` or not. An empty string gets coerced into `0` which is a number. Are you sure that empty strings should be treated as `0` rather than just simply keeping them as an empty string?

After you remove the casting operator, please also look into the `isNaN` usage, because `isNaN` will do its own coercing and you will end up again with the case of empty string being forced into being `0`.

There's double coercion going on here. If empty string should not be treated as `0` please remove both usages of the coercion.

TCL735 (changes requested):

jrenee42:
> the whole point of this was to add commas to numbers > 1000, right? can you write a test for that case here please?

there are tests for that function in the unit tests in giraffe.  I added another unit test here as well; even though there already was one:  https://github.com/influxdata/giraffe/pull/649

src/visualization/types/Table/TableCell.tsx
jrenee42: this is out of scope for this fix.  after the giraffe pr gets merged in (a duplicate of this one, essentially), will apply those same changes here to fix empty strings from incorrect coercion.

jrenee42:
closing without merging; as the giraffe table is superseding this table; so not improving this table

PR #2218, opened 2021-08-02
feat: recently used items [DO NOT MERGE]

### LINKED ISSUE #1903: Dynamic Homepage - Pinned Items - Frontend Implementation
_fulfills effort for epic #1895 - Dynamic Homepage_

** Pinned items feature**

* Enabling users to "pin" items that they use frequently to the homepage would allow for reduced navigation
complexity and increase interaction time with components that are frequently in use/ drive query count.
**Effort Scope**
* The scope / purpose of this ticket will be to provide the frontend implementation for the pinned items feature. This will involve utilizing mocks provided in the link below. 

Intended Outcome / Acceptance Criteria

A component displaying cards that link to specific pinned items the user has determined belong on their homepage, as well as the ability from specific pages to pin an item to the homepage. 

Relevant Mock: https://www.figma.com/file/GT5UFaojilawBM5vgLf3na/Homepage-2021?node-id=0%3A1

ShmuelLotman:
Closes #1903 

Introduces pinned items into the UI, waiting on the backend (grace test merge, deployment of service) for this to be merged


genehynson:
@ShmuelLotman maybe click the "convert to draft" button so it doesn't accidentally get merged?

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2218)

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }, [])
+
+  return (
+    <PinnedItemsContext.Provider value={{pinnedItems, deletePinnedItemsHelper}}>
+      {children}
+    </PinnedItemsContext.Provider>
+  )
+}
+
+export default memo(PinnedItemsProvider)
```
Asalem1, line 151: why are we awaiting all the things?

src/tasks/components/TaskCard.test.tsx
```diff
@@ -26,6 +27,17 @@ const setup = (override = {}) => {
     onCreateLabel: jest.fn(),
     setCurrentTasksPage: jest.fn(),
     labels: [], // all labels
+    isPinned: false,
+    org: {id: 'BUCKSINSIX', name: 'Milwaukee Bucks'},
+    me: {
+      id: 'FORTHECULTURE',
+      name: 'Bucks',
```
Asalem1, line 34:
```suggestion
      name: 'Lakers',
```

src/tasks/components/TaskCard.test.tsx
```diff
@@ -26,6 +27,17 @@ const setup = (override = {}) => {
     onCreateLabel: jest.fn(),
     setCurrentTasksPage: jest.fn(),
     labels: [], // all labels
+    isPinned: false,
+    org: {id: 'BUCKSINSIX', name: 'Milwaukee Bucks'},
```
Asalem1, line 31:
```suggestion
    org: {id: 'Mambacita :(', name: 'Los Angeles Lakers'},
```

cypress/e2e/cloud/pinned.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    cy.signin().then(() =>
+      cy.fixture('routes').then(({orgs}) => {
+        cy.get('@org').then(({id}: any) => {
+          orgID = id
+          cy.setFeatureFlags({
+            pinnedItems: true,
+            docSearchWidget: true,
+          }).then(() => {
+            cy.wait(1000)
```
genehynson, line 17: remove this, please! The issue here is that you're setting the feature flags and then navigating to a different page. We should try to flip the order. Or use the nav bar to navigate to the new page. See this: https://influxdb.atlassian.net/wiki/spaces/PB/pages/2367356935/Cypress#Feature-Flags

cypress/e2e/cloud/pinned.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    it('unpins a card which removes it from the pinned list', () => {
+      cy.getByTestID('dashboard-card')
+        .first()
+        .trigger('mouseover')
+        .within(() => {
+          cy.getByTestID('context-pin-menu').click()
+          cy.getByTestID('context-pin-dashboard').click()
+        })
+      cy.visit('/')
```
genehynson, line 100: I would add a `cy.getByTestID('nav-tree')` here after the `cy.visit('/')` so the test waits for the page to load before continuing. Should reduce flakes.

cypress/e2e/cloud/pinned.test.ts
```diff
@@ -0,0 +1,330 @@
...
+        cy.getByTestID('dashboard-card--name-button')
+          .first()
+          .click()
+
+        cy.get('.cf-input-field')
+          .type('Bucks In Six')
+          .type('{enter}')
+      })
+
+      cy.visit('/')
```
genehynson, line 83: I would add a `cy.getByTestID('nav-tree')` here after the `cy.visit('/')` so the test waits for the page to load before continuing. Should reduce flakes.

cypress/e2e/cloud/pinned.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    it('unpins when the underlying resource is removed', () => {
+      cy.getByTestID('dashboard-card')
+        .first()
+        .trigger('mouseover')
+        .within(() => {
+          cy.getByTestID('context-delete-menu').click()
+          cy.getByTestID('context-delete-dashboard').click()
+        })
+
+      cy.visit('/')
```
genehynson, line 123: I would add a `cy.getByTestID('nav-tree')` here after the `cy.visit('/')` so the test waits for the page to load before continuing. Should reduce flakes.

cypress/e2e/cloud/pinned.test.ts
```diff
@@ -0,0 +1,330 @@
...
+        .first()
+        .trigger('mouseover')
+        .then(() => {
+          cy.getByTestID('context-pin-menu').click()
+          cy.getByTestID('context-pin-task').click()
+        })
+    })
+
+    it('can pin a task to the homepage', () => {
+      cy.visit('/')
```
genehynson, line 179: here's another one, maybe check for others

genehynson (changes requested):
Thanks for writing tests! Left a couple suggestions

src/shared/contexts/pinneditems.tsx
ShmuelLotman: Patience is a virtue

Seriously tho, no particular reason one way or another, notice where i invoke them now, they are not being called synchronously so either way, shouldn't be an issue

src/tasks/components/TaskCard.test.tsx
ShmuelLotman: how dare you

cypress/e2e/cloud/pinned.test.ts
ShmuelLotman: AH thank you. I was wondering why i needed to add this back, apologies!

src/shared/contexts/pinneditems.tsx
Asalem1: let's remove it. Feels like something should be done afterwards, but we're not

ShmuelLotman: sounds good

cypress/e2e/cloud/pinned.test.ts
genehynson: one more here!

src/dashboards/components/dashboard_index/DashboardCards.tsx
```diff
@@ -120,6 +130,9 @@ class DashboardCards extends PureComponent<OwnProps & StateProps> {
                 updatedAt={meta.updatedAt}
                 description={description}
                 onFilterChange={onFilterChange}
+                isPinned={
+                  !!pinnedItems.find(item => item?.metadata.dashboardID === id)
+                }
```
genehynson, line 135: I assume we're not concerned about every dashboard cell running this `.find` because there will only ever be 5 pinned items right?

genehynson: although it would be dope regardless to convert this to an object so that we can change this from O(n) to O(1) lookup

src/flows/components/header/index.tsx
```diff
@@ -21,21 +21,23 @@ import RenamablePageTitle from 'src/pageLayout/components/RenamablePageTitle'
...
 
   const handleRename = (name: string) => {
     update({name})
+    updatePinnedItemByParam(id, {name})
   }
 
   const printJSON = () => {
     /* eslint-disable no-console */
-    console.log(JSON.stringify(serialize(flow, id), null, 2))
+    console.log(JSON.stringify(serialize(flow, orgID), null, 2))
```
genehynson, line 40: intentional change?

src/me/components/PinnedItems.tsx
```diff
@@ -0,0 +1,104 @@
...
+              contextMenu={
+                <Context>
+                  <Context.Menu
+                    icon={IconFont.Trash}
+                    color={ComponentColor.Danger}
+                    testID="pinneditems-delete--menu"
+                  >
+                    <Context.Item
+                      label="Unpin"
+                      action={async () => await handleDeletePinnedItem(item.id)}
```
genehynson, line 77: should this be async/await? will it cause the UI to hang?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,151 @@
...
+  Notebook = 'notebook',
+  Task = 'task',
+  Alert = 'alert',
+  UploadCSV = 'uploadCsv',
+  UploadLP = 'uploadLP',
+  ClientLibrary = 'clientLibrary',
+}
+export const PinnedItemsContext = createContext(null)
+
+// The routes below are placeholders until we get the implementation of the swagger-defined routes through the pipeline
```
genehynson, line 49: we can delete these now right?

src/tasks/containers/TasksPage.test.tsx
```diff
@@ -42,6 +42,10 @@ const localHistory = createMemoryHistory({initialEntries: ['/']})
 
 withRouterProps.match.params.orgID = orgs[0].id
 
+jest.mock('src/shared/constants/index', () => ({
+  CLOUD: true,
+}))
+
```
genehynson, line 48: does this affect all the other tests in this file? Should we move this mock into your specific test that requires this variable to be set?

src/flows/components/header/index.tsx
ShmuelLotman: Yes, if you look, i had to map the id to a different name due to a property name conflict

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,151 @@
...
+let deletePinned
+let postPinned
+
+if (CLOUD) {
+  const pinnedItemFunctions = require('src/client/pinnedItemRoutes')
+  getPinned = pinnedItemFunctions.getPinned
+  putPinned = pinnedItemFunctions.putPinned
+  postPinned = pinnedItemFunctions.postPinned
+  deletePinned = pinnedItemFunctions.deletePinned
+}
```
genehynson, line 23: this looks fishy to me. We're saying to require this routes file which only exists for cloud and assign the imported functions to these variables. But the rest of this file assumes those variables are defined...which isn't the case for OSS right?

src/shared/contexts/pinneditems.tsx
ShmuelLotman: On second thought, id like to leave the functions async. i still want to have the functions be able to have the error handling that i've left on there, and ultimately this will still run async. I think no harm done, no?

src/me/components/PinnedItems.tsx
ShmuelLotman: reMoved

src/shared/contexts/pinneditems.tsx
ShmuelLotman: Deleted

src/tasks/containers/TasksPage.test.tsx
ShmuelLotman: Even better news, we dont even need it anymore after my latest changes!

cypress/e2e/cloud/pinned.test.ts
ShmuelLotman: Done!

src/shared/contexts/pinneditems.tsx
ShmuelLotman: Changed this around so that its handled anywhere where we make this import. Wont get imported and/or used unless its both CLOUD and the flag for pinned items is true. That way, there is no way this can be accessed or used without it being in CLOUD

ShmuelLotman: LMK your thoughts

src/dashboards/components/dashboard_index/DashboardCards.tsx
```diff
@@ -18,6 +18,13 @@ import {LimitStatus} from 'src/cloud/actions/limits'
 
 // Utils
 import {extractDashboardLimits} from 'src/cloud/utils/limits'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {CLOUD} from 'src/shared/constants'
+
+let getPinnedItems
+if (CLOUD) {
+  getPinnedItems = require('src/shared/contexts/pinneditems').getPinnedItems
```
genehynson, line 26: Why do we have to use require here? why can't we just import this?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,151 @@
...
+  if (pinnedRes.status !== 200) {
+    throw new Error('Unable to retrieve pinned items')
+  } else {
+    return pinnedRes.data
+  }
+}
+
+const removePinnedItem = async (id: string) => {
+  await deletePinned({id})
+}
```
genehynson, line 60: I'm confused why this fn exists? It's not exported and just calls the exported version. Can we remove it?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,151 @@
...
+    throw new Error('Unable to retrieve pinned items')
+  } else {
+    return
+  }
+}
+
+const updatePinnedItem = async (id: string, item: Partial<PinnedItem>) => {
+  await putPinned({id, data: item})
+
+  return
```
genehynson, line 73: don't think we need this return

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,151 @@
...
+const removePinnedItem = async (id: string) => {
+  await deletePinned({id})
+}
+const addPinnedItem = async (item: Partial<PinnedItem>) => {
+  const pinnedRes = await postPinned({data: item})
+  if (pinnedRes.status !== 200) {
+    throw new Error('Unable to retrieve pinned items')
+  } else {
+    return
+  }
```
genehynson, line 67: don't think we need this return

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+const updatePinnedItem = async (id: string, item: Partial<PinnedItem>) => {
```
genehynson, line 70: should this function be handling errors?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,151 @@
...
+  } catch (err) {
+    console.error(err)
+  }
+}
+
+export const deletePinnedItem = async (itemID: string) => {
+  try {
+    await removePinnedItem(itemID)
+  } catch (err) {
+    console.error(err)
```
genehynson, line 103: I'm not sure that logging api errors is secure. I think we would want to handle this differently. See how notebooks does it in src/flows/contex/api.ts. These should probably be shown to the user as notifications, right?

src/shared/contexts/pinneditems.tsx
genehynson: if this file is only imported when CLOUD is set, can't we just import the functions from `pinnedItemRoutes` normally in this file?

src/dashboards/components/dashboard_index/DashboardCards.tsx
ShmuelLotman: because of the unique syntax, we do have to use require, its a pattern we have elsewhere

src/shared/contexts/pinneditems.tsx
ShmuelLotman: yes

src/shared/contexts/pinneditems.tsx
ShmuelLotman: yes, you are correct - i should change this now

src/flows/components/FlowCards.tsx
```diff
@@ -27,6 +34,14 @@ const NoMatches = () => {
 }
 
 const FlowCards: FC<Props> = ({flows, search}) => {
+  const [pinnedItems, setPinnedItems] = useState([])
+
+  useEffect(() => {
+    if (isFlagEnabled('pinnedItems') && CLOUD) {
+      getPinnedItems().then(res => setPinnedItems(res))
```
genehynson, line 41: can we catch failures from the API here too?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
+
+export const updatePinnedItemByParam = async (id: string, updateParams: {}) => {
+  const pinnedItems = await getPinnedItems()
+  const toUpdateItem = pinnedItems?.find(item =>
+    Object.values(item.metadata).includes(id)
+  )
+  if (toUpdateItem) {
+    await putPinned({
```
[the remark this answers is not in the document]
ShmuelLotman: not needed - its handled in the consumers

genehynson (approved):
approved. now can I pin this PR please? 😆

src/flows/components/FlowCards.tsx
```diff
@@ -27,6 +34,18 @@ const NoMatches = () => {
 }
 
 const FlowCards: FC<Props> = ({flows, search}) => {
+  const [pinnedItems, setPinnedItems] = useState([])
+
+  useEffect(() => {
+    if (isFlagEnabled('pinnedItems') && CLOUD) {
+      getPinnedItems()
+        .then(res => setPinnedItems(res))
```
Asalem1, line 42: what if we just set this as a set and we could do some o(1) look up like `set.has`. Would look something like:

`setPinnedItems(new Set(...res))` then later on just checking to see if `pinnedItems.has` or something

src/flows/components/FlowContextMenu.tsx
```diff
@@ -18,18 +22,53 @@ import {PROJECT_NAME_PLURAL} from 'src/flows'
...
+      })
+      dispatch(notify(pinnedItemSuccess('notebook', 'added')))
+    } catch (err) {
+      dispatch(notify(pinnedItemFailure(err.message, 'create')))
+    }
+  }
 
   const handleDelete = () => {
     event('delete_notebook')
+    deletePinnedItemByParam(id)
```
Asalem1, line 71: should this be feature flagged and guarded about the CLOUD?

src/flows/components/header/index.tsx
```diff
@@ -21,21 +21,23 @@ import RenamablePageTitle from 'src/pageLayout/components/RenamablePageTitle'
...
 
 const FlowHeader: FC = () => {
-  const {update, flow} = useContext(FlowContext)
-  const {id} = useSelector(getOrg)
+  const {update, flow, id} = useContext(FlowContext)
+  const {id: orgID} = useSelector(getOrg)
 
   const handleRename = (name: string) => {
     update({name})
+    updatePinnedItemByParam(id, {name})
```
Asalem1, line 35: should this be flagged?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,141 @@
...
+export const addPinnedItem = async (item: Partial<PinnedItem>) => {
+  const pinnedRes = await postPinned({data: item})
+  if (pinnedRes.status !== 200) {
+    throw new Error('Unable to retrieve pinned items')
+  }
+}
+
+export const deletePinnedItemByParam = async (param: string) => {
+  try {
+    const items = await getPinnedItems()
```
Asalem1, line 67: pretty sure this will need an empty `{}`
```suggestion
    const items = await getPinnedItems({})
```

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const toDeleteItem = items.find(item =>
+      Object.values(item.metadata).includes(param)
+    )
```
Asalem1, line 70: so confused, shouldn't this be handled by the API?

src/shared/contexts/pinneditems.tsx
```diff
@@ -0,0 +1,141 @@
...
+    throw new Error('Unable to delete pinned item')
+  }
+}
+
+export const updatePinnedItemByParam = async (id: string, updateParams: {}) => {
+  const pinnedItems = await getPinnedItems()
+  const toUpdateItem = pinnedItems?.find(item =>
+    Object.values(item.metadata).includes(id)
+  )
+  if (toUpdateItem) {
```
Asalem1, line 93: same as the above?

src/tasks/components/TaskCard.tsx
```diff
@@ -103,8 +121,29 @@ export class TaskCard extends PureComponent<
...
+      })
+      this.props.sendNotification(pinnedItemSuccess('task', 'added'))
+    } catch (err) {
+      this.props.sendNotification(pinnedItemFailure(err.message, 'task'))
+    }
+  }
+
+  private handleOnDelete = () => {
+    this.props.onDelete(this.props.task)
+    deletePinnedItemByParam(this.props.task.id)
```
Asalem1, line 143: should this be flagged?

Asalem1 (commented):
A couple things, it seems like the UI is concerned with too much with regards to the API (i.e. validating that a thing exists before updating or deleting it). Also, we're updating some existing methods without flagging (for cloud and the feature) the new functionality on the view layer. Not sure if the stuff is flagged when it's called, but it seems like we should probably do that on the view layer. Then again, what do I know 🤷🏽

src/shared/contexts/pinneditems.tsx
ShmuelLotman: nope, bc its the wrapper

src/shared/contexts/pinneditems.tsx
ShmuelLotman: this is helping us determine based on a parameter whether the request is valid without having to do a separate lookup for the resources using database calls. I just chose to imlpement it on the frontend so we dont make a call to delete in the frist place if we have nothing to delete

Asalem1: Totally, I guess my concern is that we end up firing two requests if we want to delete or update, and only 1 if it's invalid ... best case scenario right? Seems like an API concern and not a UI concern

src/shared/contexts/pinneditems.tsx
ShmuelLotman: see above comment

src/shared/contexts/pinneditems.tsx
ShmuelLotman: Yeah, i could put up a ticket to move this over to the backend, good idea. Backend changes require significant....process

Asalem1 (approved):

PR #2250, opened 2021-08-04
feat: timeFormat validation

### LINKED ISSUE #1771: Replace Moment: Provide validation functionality
### Background
Depends on #1767

See #1769

Related to #1769, we want to tell if a date string is a valid format that we support.

### Acceptance Criteria
- Given a formatted date string, return whether it's a valid date format we support

ChitlangeSahas:
Closes #1771

We need a validator so we can make sure the date as string, are in valid format.


src/utils/datetime/validator.test.ts
```diff
@@ -0,0 +1,52 @@
+import {isValid} from './validator'
+
+describe('validation tests', () => {
+  it('should do the thing', function() {
+    expect(isValid('1999-02-09 23:00:00', 'YYYY-MM-DD HH:mm:ss')).toBeTruthy()
+    expect(
+      isValid('1999-02-09 12:00:00 PM -0700', 'YYYY-MM-DD hh:mm:ss a ZZ')
```
ChitlangeSahas, line 7: Here, we are testing for local timezone, but it's okay because we provide the timezone. There will be no conflict in the CI.

src/utils/datetime/validator.ts
```diff
@@ -0,0 +1,42 @@
+import {isMatch} from 'date-fns'
+import {DEFAULT_TIME_FORMAT} from 'src/shared/constants'
+
+const formatterToDateFnsMap = {
+  'YYYY-MM-DD HH:mm:ss': 'yyyy-MM-dd HH:mm:ss', // DEFAULT_TIME_FORMAT
```
ChitlangeSahas, line 5: I tried replacing the `'YYYY-MM-DD HH:mm:ss'` with `DEFAULT_TIME_FORMAT` but it would not let me access it with `formatterToDateFnsMap['YYYY-MM-DD HH:mm:ss']. Perhaps a JS thing?

src/utils/datetime/validator.ts
```diff
@@ -0,0 +1,41 @@
...
+  'MMMM D, YYYY HH:mm:ss': 'LLLL d, yyyy HH:mm:ss',
+  'MMMM D, YYYY hh:mm:ss a': 'LLLL d, yyyy hh:mm:ss bb',
+  'dddd, MMMM D, YYYY HH:mm:ss': 'EEEE, LLLL d, yyyy HH:mm:ss',
+  'dddd, MMMM D, YYYY hh:mm:ss a': 'EEEE, LLLL d, yyyy hh:mm:ss bb',
+}
+
+const getDateFnsFormatString = (format: string): string => {
+  const dateFnsFormatString = formatterToDateFnsMap[format]
+    ? formatterToDateFnsMap[format]
+    : 'yyyy-MM-dd HH:mm:ss'
```
ChitlangeSahas, line 29: This needs to be `DEFAULT_TIME_FORMAT` but in `date-fns` library's accepted string so cant use our constant.

src/utils/datetime/validator.test.ts
```diff
@@ -0,0 +1,52 @@
+import {isValid} from './validator'
+
+describe('validation tests', () => {
```
hoorayimhelping: we're describing the validator itself, not the tests

src/utils/datetime/validator.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    expect(
+      isValid('Tuesday, February 9, 1999 23:00', 'dddd, MMMM D, YYYY HH:mm:ss')
+    ).toBeFalsy()
+    expect(
+      isValid(
+        'Tuesday, February 9, 1999 11:00:00',
+        'dddd, MMMM D, YYYY hh:mm:ss a'
+      )
+    ).toBeFalsy()
+  })
```
hoorayimhelping, line 94: nice tests! i don't think you need to change anything here, but just for your own knowledge, there's a school of thought that says we should only have [one assertion per test](https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test) for a number of reasons - the main being that multiple assertions often confuse the intent of code. another is that when a test fails, if there's one assertion per test, then which test the failure occurred on is obvious and it tells us about the nature of the failure. i don't think these are reasons to change your tests when they pass, but just for your own understanding of testing methodologies, i think you should familiarize yourself with them

hoorayimhelping (approved):
nice work!

src/utils/datetime/validator.test.ts
ChitlangeSahas: Got it! I'll read that! That also makes sense, if it were to fail, it'd be hard to know what case failed exactly.

src/utils/datetime/validator.test.ts
```diff
@@ -0,0 +1,147 @@
...
+  it('should be strict on date formats', function() {
```
ChitlangeSahas, line 101: Formats should be strict, so we test that with these tests here.

ChitlangeSahas:
The lint job timing will remain a mystery.

PR #2252, opened 2021-08-04
fix: SimpleTable resizing bug that creates identical pages

### LINKED ISSUE #2170: SimpleTable pagination is limited to only certain pages if the table is resized
SimpleTable pagination is limited to only certain pages if the table is resized.

- build a query and submit
- switch to raw data view
- resize the table
- click on the last page in the pagination
- use an arrow button to paginate
- observe that the pagination recycles without going through all of the pages (many pages are no longer accessible through pagination)

https://user-images.githubusercontent.com/10736577/127400422-48aa0279-8478-43d6-8396-87c1ac4f2076.mp4

mtian725:
Closes #2170

## Context
The reason why there were duplicate pages with only a single row of data in #2170 is due to `setTotalPages(Math.ceil((result?.table?.length ?? 0) / size))`. Because `size` refers to how many rows are in the table not how many can fit, going to the last page where there is a partially filled table will cause the calculation to set the total number of pages to be much larger than what it should be. In the calculation of what rows appear on that page, if the page selected is out of the true bounds (meaning the math will result in a unexpectedly large offset), it will default it to show only the last row of data, which is what happened in #2170.

The change to setMaxSize has little to do with the issue, but it causes issues when resizing as well.

## Demo
https://user-images.githubusercontent.com/39343323/128261892-f73b1942-e2a5-4422-935a-2077e256e0ab.mov

## Warning
This fix also bring up another issue where the middle part of the pagination just disappears. However, it was concluded that it is most likely a clockface issue. This currently does not appear on production.

If the video is not render, then go here for the issue and a video in more detail. https://github.com/influxdata/clockface/issues/654

Uploading Screen Recording 2021-08-04 at 5.43.20 PM.mov…

### DIFF
--- a/src/visualization/context/pagination.tsx
+++ b/src/visualization/context/pagination.tsx
@@ -63,7 +63,7 @@ export const PaginationProvider: FC<PaginationProviderProps> = ({
 
   const previous = useCallback(() => {
     setOffset(calcPrevPageOffset(offset, size))
-  }, [offset, size, setOffset]) 
+  }, [offset, size, setOffset])
 
   const setPage = useCallback(
     (page: number) => {
--- a/src/visualization/context/pagination.tsx
+++ b/src/visualization/context/pagination.tsx
@@ -59,17 +59,17 @@ export const PaginationProvider: FC<PaginationProviderProps> = ({
     } else {
       setOffset(offset + size)
     }
-  }, [offset, size, setOffset])
+  }, [offset, size, setOffset]) // eslint-disable-line react-hooks/exhaustive-deps
 
   const previous = useCallback(() => {
     setOffset(calcPrevPageOffset(offset, size))
-  }, [offset, size, setOffset])
+  }, [offset, size, setOffset]) 
 
   const setPage = useCallback(
     (page: number) => {
       setOffset(calcOffset(page, maxSize, total))
     },
-    [offset, maxSize, setOffset]
+    [offset, maxSize, setOffset] // eslint-disable-line react-hooks/exhaustive-deps
   )
 
   return (
--- a/src/visualization/types/SimpleTable/PagedTable.tsx
+++ b/src/visualization/types/SimpleTable/PagedTable.tsx
@@ -278,21 +278,21 @@ const PagedTable: FC<Props> = ({result, properties}) => {
 
   useEffect(() => {
     setSize(size)
-  }, [size])
+  }, [size]) // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
     setMaxSize(measurePage(result, 0, height))
-  }, [height, result])
+  }, [height, result]) // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
     setPage(1)
-  }, [result])
+  }, [result]) // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
     if (maxSize) {
       setTotalPages(Math.ceil((result?.table?.length ?? 0) / maxSize))
     }
-  }, [maxSize])
+  }, [maxSize]) // eslint-disable-line react-hooks/exhaustive-deps
 
   const inner =
     !!size &&
--- a/src/visualization/types/SimpleTable/PagedTable.tsx
+++ b/src/visualization/types/SimpleTable/PagedTable.tsx
@@ -292,7 +292,7 @@ const PagedTable: FC<Props> = ({result, properties}) => {
     if (maxSize) {
       setTotalPages(Math.ceil((result?.table?.length ?? 0) / maxSize))
     }
-  }, [maxSize]) // eslint-disable-line react-hooks/exhaustive-deps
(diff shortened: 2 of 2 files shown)


src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -281,20 +281,18 @@ const PagedTable: FC<Props> = ({result, properties}) => {
...
   }, [result])
 
   useEffect(() => {
-    if (size) {
-      setTotalPages(Math.ceil((result?.table?.length ?? 0) / size))
+    if (maxSize) {
+      setTotalPages(Math.ceil((result?.table?.length ?? 0) / maxSize))
     }
-  }, [height, result])
+  }, [maxSize])
```
TCL735, line 295: Nice work! 🔥 

However, I see lint warnings for the hooks that you have updated. To be fair, the lint warnings were already there. Are we just going to ignore the lint warnings? There's a way to disable the lint warnings if we don't want them there. Add a comment add the end of each line of the dependency array that disables the lint rule for that line.

See here for examples:
Screen Shot 2021-08-04 at 3 46 15 PM

TCL735 (changes requested):

src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -278,23 +278,21 @@ const PagedTable: FC<Props> = ({result, properties}) => {
...
+  }, [result]) // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
-    if (size) {
-      setTotalPages(Math.ceil((result?.table?.length ?? 0) / size))
+    if (maxSize) {
+      setTotalPages(Math.ceil((result?.table?.length ?? 0) / maxSize))
     }
-  }, [height, result])
+  }, [maxSize]) // eslint-disable-line react-hooks/exhaustive-deps
```
mtian725, line 295: Disabling lint ruling here because the warning ask for the setters to be included in the dependencies. For `result?.table?.length`, the entire visualization re-renders when result changes, and given a `result` and it's `table`, the length will remain static. It will only change when result changes, which again, re-renders the entire visualization.

image

src/visualization/context/pagination.tsx
```diff
@@ -59,17 +59,17 @@ export const PaginationProvider: FC<PaginationProviderProps> = ({
...
     setOffset(calcPrevPageOffset(offset, size))
-  }, [offset, size, setOffset])
+  }, [offset, size, setOffset]) 
 
   const setPage = useCallback(
     (page: number) => {
       setOffset(calcOffset(page, maxSize, total))
     },
-    [offset, maxSize, setOffset]
+    [offset, maxSize, setOffset] // eslint-disable-line react-hooks/exhaustive-deps
```
mtian725, line 72: Disabling lint ruling here because the warning is asking for `total` to be included. `total` is the total number of rows in the data, and that does not change. See previous comment explaining why not having `result?.table?.length` as a dependency is fine. 

image

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2252)

src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -274,27 +274,25 @@ const PagedTable: FC<Props> = ({result, properties}) => {
   }, [result, offset, height])
   const tables = useMemo(() => {
     return subsetResult(result, offset, size, properties.showAll)
-  }, [result, offset, size])
+  }, [result, offset, size]) // eslint-disable-line react-hooks/exhaustive-deps
```
mtian725, line 277: Disabling lint ruling here because the warning ask for the `properties.showAll` to be added. Currently, `properties.showAll` doesn't change in the UI, but if it was, the process of handling it (like the other visualization components) would re-render the visualization.

image

TCL735:
🚢

TCL735 (approved):

src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -292,7 +292,7 @@ const PagedTable: FC<Props> = ({result, properties}) => {
     if (maxSize) {
       setTotalPages(Math.ceil((result?.table?.length ?? 0) / maxSize))
     }
-  }, [maxSize]) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [maxSize, result]) // eslint-disable-line react-hooks/exhaustive-deps
```
mtian725, line 295: This change fixes an issue where changing data buckets does not change the total number of pages in the pagination bar. Originally, this hook depended on the `maxSize` changing, but when you submit a new query with either the same bucket or a different one, the actual `maxSize` does not change, so the pagination bar doesn't update.

PR #2269, opened 2021-08-05
feat: remote db panel

drdelambre:
Peek 2021-08-05 11-12

this is just the view layer of a new panel that, in conjunction with #2196, and a following PR to glue those two together, will allow users to compare query results between their multiple regions or self hosted instances. There's a bunch of kinks to work out before we turn this on, like disabling / reworking resource generation for cross region queries (bucket and task creation mostly), but this creates a place for us to start iterating on the idea


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2269)

src/flows/pipes/Region/view.tsx
```diff
@@ -0,0 +1,301 @@
...
+    value: 'https://influxdb.aws.influxdata.io',
+  },
+  {
+    label: 'Local Kind Cluster',
+    group: 'Development',
+    flag: 'local-dev',
+    value: 'https://twodotoh.a.influxcloud.dev.local',
+  },
+  {label: 'Self Hosted', value: 'self-hosted'},
+]
```
Asalem1, line 70: is there any way these can not be in the UI? It just .... stinks. Also, do we want to allow querying of tools?

src/flows/pipes/Region/view.tsx
```diff
@@ -0,0 +1,301 @@
...
+const Source: FC<PipeProp> = ({Context}) => {
+  const {data, update} = useContext(PipeContext)
+  const [error, setError] = useState<boolean>(false)
+  const flags = useSelector(activeFlags)
+
+  const updater = option => {
+    if (option.value === 'self-hosted') {
+      update({
+        source: 'custom',
+        region: 'https://localhost:8086',
```
Asalem1, line 81: should this be based on anything in the dev env?

src/flows/pipes/Region/view.tsx
```diff
@@ -0,0 +1,301 @@
...
+        source: 'static',
+        region: option.value,
+      })
+    }
+  }
+
+  const [options, selected] = useMemo(() => {
+    const hashed = JSON.parse(JSON.stringify(REGIONS)).reduce(
+      (acc, curr) => {
+        if (curr.flag && !flags[curr.flag]) {
```
Asalem1, line 94: where are the flags being set on `curr`?

src/flows/pipes/Region/view.tsx
drdelambre: is there an api that provides all of our supported regions? i'd love to use that if it exists.

Asalem1: Nope, but maybe there should be?

src/flows/pipes/Region/view.tsx
drdelambre: nope. it can be cloud hosted and is run relative to the browser. its up to the user to ensure that this url is available to their task they generate

src/flows/pipes/Region/view.tsx
drdelambre: in the REGION const above

Asalem1: Aw, totally missed that. Maybe we can type that data if we're declaring it on the UI

Asalem1 (approved):

joelfried:
@drdelambre please pull or rebase master to get the latest CI changes on this branch before merging.

PR #2291, opened 2021-08-09
feat: add more events and throw in an experiment to run

drdelambre:
we think that people would want to build tasks in notebooks more than the current task creation interface, they just don't know how to. but we dont know that for sure, so lets give users who have demonstrated that they intend to create a task the option to use a notebook instead. if they dont want to, they just dont have to click on the link.

this experiment is to drive more people who want to create tasks through notebooks, and to validate that the task creation interface is better in notebooks. we can measure success by the conversion of people who generate a task through the notebook interface. more sustained tasks exported... the more confidence we have in depreciating the existing task interface. if a bunch of people click on the link and no more tasks are generated, we know we have more work to do on the notebooks workflow for task users.

the CTA is behind a feature flag so we can build up a baseline for the conversion of the existing task creation funnel before turning it on.

Screenshot from 2021-08-09 13-13-08

### DIFF
--- a/src/tasks/components/TaskHeader.tsx
+++ b/src/tasks/components/TaskHeader.tsx
@@ -1,5 +1,6 @@
 // Libraries
 import React, {PureComponent} from 'react'
+import {Link} from 'react-router-dom'
 
 // Components
 import {
@@ -9,6 +10,7 @@ import {
   Page,
 } from '@influxdata/clockface'
 import RateLimitAlert from 'src/cloud/components/RateLimitAlert'
+import {FeatureFlag} from 'src/shared/utils/featureFlag'
 
 interface Props {
   title: string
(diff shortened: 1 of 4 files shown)


src/tasks/components/TaskHeader.tsx
```diff
@@ -27,6 +29,12 @@ export default class TaskHeader extends PureComponent<Props> {
           <RateLimitAlert />
         </Page.Header>
         <Page.ControlBar fullWidth={true}>
+          <Page.ControlBarLeft className="task-header--cta">
+            <FeatureFlag name="flowsCTA">
```
Asalem1, line 33: will wrapping the feature flag here or around the ControlBarLeft make a difference?

Asalem1 (approved):

src/tasks/components/TaskHeader.tsx
drdelambre: not that i saw

PR #2300, opened 2021-08-10
feat: add 'Dispose the Client' to code samples, add 'executeFull' template for uses in 'Notebook: Copy as'

### LINKED ISSUE #2075: Notebooks: client sample code in 'copy as' is inconsistent
In the 'copy as' feature, we generate code that the user can cut/paste into their client.  Ideally the code should be functionally equivalent in all languages. Here's a summary of my recommended changes.  

General statements:
1. Some of the client code includes print statements and others do not. I would either leave them all out, or add them to the ones that don't have them.  (My vote is to leave them out, as it is cleaner code and a client API developer can add them if they want.)
2. Some of the client code includes code to iterate over the query results, but other code does not. It simply assigns the query results to an object and leaves it up to the developer to figure out how to iterate over the object.  I think it's useful example code (may not be that intuitive on how to iterate over the results) so I would add it in any language that it isn't included in.  

Client specific comments:
Arduino:
This one is really confusing, and seems like a different sample app for wi-fi settings. I would change it to just the code that creates the InfluxDBClient with the parameters populated from the query (URL, ORG, BUCKET, TOKEN). Then have it execute the query from the notebook and get the results, like the other languages.

Go:  I don't know Go, but there is a call to client.close that seems to come too early in the sequence (unless defer causes it to do the right thing).

Swift: There is a call to client.close that seems to come too early (before the query). I'm not a swift developer so maybe this is fine.  There is also a print statement referring to CPU usage which isn't valid. I'd delete it.

bednar:
Closes #2075

1. The code samples contains **Dispose the Client** section (for clients who need it only)
image
2. The query examples contains snipped for iterating results
image
3. The "Export to" function in "Notebooks" contains valid code
image

**Fixed Libraries**
- [x] Arduino
- [x] C#
- [x] GO
- [x] Java
- [x] JavaScript/Node
- [x] Kotlin
- [x] PHP
- [x] Python
- [x] R
- [x] Ruby
- [x] Scala
- [x] Swift


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2300)

bednar:
Source code are rebased.

PR #2311, opened 2021-08-10
fix: notification panel now setting the fields as columns

### LINKED ISSUE #2259: Notebook Notification Panel: Field dropdown should include pivoted columns only
Because the data flows through `fieldsAsCols()` before it goes to the check or notification functions, it means the data that populates this list also needs to be pivoted before being displayed. As is, notifications or checks will always fail because there is no `_value` column in the data flowing into the alerting fucntions.

image

Asalem1:
Closes #2259 

Pretty sure this does the trick. It might not. Not sure what this is supposed to do truthfully

### DIFF
--- a/src/flows/pipes/Notification/Threshold.tsx
+++ b/src/flows/pipes/Notification/Threshold.tsx
@@ -67,7 +67,7 @@ export const THRESHOLD_TYPES = {
 const Threshold: FC = () => {
   const {data, update, results} = useContext(PipeContext)
 
-  const numericColumns = [...new Set(results.parsed.table.columnKeys)]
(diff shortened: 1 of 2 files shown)


drdelambre (changes requested):
we want to do this on the frontend to decouple the visualization query from the product requirement

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2311)

Issue #1555, opened 2021-05-28
Telegraf Refresh: Add Bucket--From Plugin, Create Mode


kristinarobinson:
GIVEN that a user is on the `Load Data` page,
WHEN the user selects `Add Plugin to New Config` CTA from any of the available Telegraf plug-ins,
Then the user is taken to a modal where the user can optionally name the new configuration file AND
the user can select a bucket from the user's available buckets from a dropdown list

https://www.figma.com/file/hMUcxA5RdoEhdJLvClWWB7/Agent-Configuration?node-id=263%3A337

TCL735:
WIP is here: https://github.com/influxdata/ui/tree/feat_1555_plugin_create_config_mode

TCL735:
The actual configuration text for each plugin has a source of truth that is located here: https://github.com/influxdata/telegraf/blob/master/etc/telegraf.conf

Give the above, the `.md` files located at **/src/writeData/components/telegrafPlugins** are not necessarily correct.

We will need a separate issue to capture the work required to parse the source of truth and import the individual configurations as text into the text editor.

TCL735:
Work in progress. Branch is here: https://github.com/influxdata/ui/tree/feat_1555_plugin_create_config_mode

PR #2312, opened 2021-08-10
feat: move create bucket form into its own component

TCL735:
As part of #1555 

Currently, the form that creates a new bucket is implemented directly into its own overlay. Soon, there will be multiple places in the ui that can create new buckets, and each of these places may use the overlay a little differently. Therefore, let's separate the overlay from the form so that a common form can be re-used.

This is what happened:
- almost all of the code for `CreateBucketOverlay` moved to `CreateBucketForm` (new file)
- only the `onClose` method was not moved
- `CreateBucketForm` contains all of the previous logic, except `onClose` which is now a prop
- `CreateBucketOverlay` imports the new component and passes in `onClose`

### DIFF
--- a/src/buckets/components/CreateBucketForm.tsx
+++ b/src/buckets/components/CreateBucketForm.tsx
@@ -18,7 +18,7 @@ import {
   initialBucketState,
   DEFAULT_RULES,
 } from 'src/buckets/reducers/createBucket'
-import {AppState, Bucket} from 'src/types'
+import {AppState, Bucket, RetentionRule} from 'src/types'
 
 // Selectors
 import {getOrg} from 'src/organizations/selectors'
@@ -29,7 +29,7 @@ interface CreateBucketFormProps {
   onClose: () => void
 }
 
-const CreateBucketForm: FC<CreateBucketFormProps> = props => {
+export const CreateBucketForm: FC<CreateBucketFormProps> = props => {
   const {onClose} = props
   const org = useSelector(getOrg)
   const isRetentionLimitEnforced = useSelector((state: AppState): boolean => {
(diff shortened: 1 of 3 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2312)

src/buckets/components/CreateBucketForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+      onSubmit={handleSubmit}
+      onChangeInput={handleChangeInput}
+      retentionSeconds={retentionSeconds}
+      onChangeRuleType={handleChangeRuleType}
+      onChangeRetentionRule={handleChangeRetentionRule}
+    />
+  )
+}
+
+export default CreateBucketForm
```
hoorayimhelping, line 108: let's try to avoid default exports https://basarat.gitbook.io/typescript/main-1/defaultisbad

src/buckets/components/CreateBucketForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return false
+  })
+  const overlayParams = useSelector(getOverlayParams)
+  const reduxDispatch = useDispatch()
+  const [state, dispatch] = useReducer(
+    createBucketReducer,
+    initialBucketState(isRetentionLimitEnforced, org.id)
+  )
+
+  const retentionRule = state.retentionRules.find(r => r.type === 'expire')
```
hoorayimhelping, line 48:
```suggestion
  const retentionRule = state.retentionRules.find(rule => rule.type === 'expire')
```

hoorayimhelping (approved):
Nice work. Just a couple of small comments

src/buckets/components/CreateBucketForm.tsx
TCL735: Oops, yeah. I actually had this changed in the other branch. When i created a new branch to get just this change, i forgot about this change. Good call. 👍

src/buckets/components/CreateBucketForm.tsx
TCL735: Sure, will use a named export here.

Issue #2286, opened 2021-08-09
Notebooks: add deadman checks to the alerts panel


juliajames12:
We would like to lowercase the list of condition options and add the "missing for longer than" option which would create a deadman alert.

Current:
![Screen Shot 2021-08-09 at 11.54.35 AM.png](https://images.zenhubusercontent.com/6058abb75f3c06c1d25467ae/fe6c2a39-72de-4493-b7b9-6b7834a6d3a0)

New:
![Screen Shot 2021-08-09 at 11.54.45 AM.png](https://images.zenhubusercontent.com/6058abb75f3c06c1d25467ae/733240f6-5bd2-4311-8c45-117a77d3a8ec)

Deadman alert:
![Screen Shot 2021-08-27 at 2.06.31 PM.png](https://images.zenhubusercontent.com/6058abb75f3c06c1d25467ae/6ead5c9d-aac8-4183-944c-e606e0a0d2e8)

Docs for deadman alert: https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/monitor/deadman/#detect-if-a-host-hasnt-reported-in-the-last-five-minutes

### Acceptance 

- All options are lowercase
- "missing for longer than" is available to choose from
- "missing for longer than" creates a deadman alert and uses a time input

drdelambre:
i did the capitalization part separately to focus the bulk of the work here

drdelambre:
recap from meeting: it makes no sense to have a threshold and deadman check on the same alert. somehow, from the text of this dropdown, the user should understand that, and we'll cross that bridge when we hit it.

the intent is to not gate the experience on whether or not the user understands what a deadman check is, so no "choose a threshold or choose a deadman" before the dropdown shows up, no "click here to read docs about the difference", no "i hope you read a blog before showing up".... just give them some options. we can worry about any validation _after_ we have some data about if and how people are mixing these up.

Asalem1:
Blocked for me due to dev environment remocal issues. WIll mark as unblocked once those get sorted

PR #2317, opened 2021-08-11
feat: alert condition capitalization

drdelambre:
this is the easy part #2286 so i can rename the issue to focus on adding deadman checks


Asalem1 (approved):
game changer

PR #2367, opened 2021-08-13
refactor: replace moment toISOString from the DatePicker.tsx

### LINKED ISSUE #2366: Replace moment in Datepicker
A final issue to replace moment in DatePicker. Remove, `toISOString` , `format` and `isValid` etc.

ChitlangeSahas:
Closes #2366
Part of #1844

Removes moment from `DatePicker`.

### DIFF
--- a/src/shared/utils/dateTimeUtils.test.ts
+++ b/src/shared/utils/dateTimeUtils.test.ts
@@ -1,4 +1,4 @@
-import {addDurationToDate} from './dateTimeUtils'
+import {addDurationToDate, isISODate} from './dateTimeUtils'
 
 // July 4th, 1983, 20:00:00 UTC ===> 1983-07-04T20:00:00.000Z
 const timestamp = 426196800000
@@ -32,5 +32,17 @@ describe('incrementDate', () => {
     const futureDate = addDurationToDate(new Date(timestamp), -1, 'd')
 
     expect(futureDate.toISOString()).toBe('1983-07-03T20:00:00.000Z')
+  })
+})
+
+describe('isISODate', function() {
+  it('should return true for a valid ISO date format', function() {
+    expect(isISODate('1983-07-03T20:00:00.000Z')).toBeTruthy()
+  })
+  it('should return false for an invalid ISO date format', function() {
+    expect(isISODate('1983-07-03A20:00:00.000')).toBeFalsy()
+  })
+  it('should return false for an empty string', function() {
+    expect(isISODate('')).toBeFalsy()
   })
 })
--- a/src/shared/utils/dateTimeUtils.ts
+++ b/src/shared/utils/dateTimeUtils.ts
@@ -46,6 +46,7 @@ export function addDurationToDate(
 // checks whether the passed date is ISO format
 // see: https://stackoverflow.com/a/52869830
 export function isISODate(dateString: string): boolean {
+  // this regex tests the input string for the ISO format, YYYY-MM-DDTHH:mm:ss.sssZ
   if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(dateString)) {
     return false
   }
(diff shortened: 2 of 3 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2367)

src/shared/utils/dateTimeUtils.ts
```diff
@@ -42,3 +42,12 @@ export function addDurationToDate(
     }
   }
 }
+
+// checks whether the passed date is ISO format
+export function isISODate(dateString: string): boolean {
+  if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(dateString)) {
```
hoorayimhelping, line 48: a couple of questions about this regex

- if you found this somewhere on the internet, can you make a comment that links to the page that created it?

- is it necessary? what does this provide over doing

 ```ts
return date.toISOString() === dateString
```

File: src/shared/utils/dateTimeUtils.test.ts [hunk 2]
```diff
@@ -34,3 +34,15 @@ describe('incrementDate', () => {
     expect(futureDate.toISOString()).toBe('1983-07-03T20:00:00.000Z')
   })
 })
+
+describe('isISODate', function() {
```
hoorayimhelping, line 38: :+1: nice!

hoorayimhelping (commented):
nice. I had a question about the regex validation

src/shared/utils/dateTimeUtils.ts
ChitlangeSahas: Yes! I found it on the stack overflow and I'll link it 👍 

Why its necessary over date.toISOString === dateString ? : 
* the input to the function is dynamic, because its coming from the text input field, and until the input is valid, the `Date()` parsing crashes, and that makes the function not return a valid boolean. So I needed to add a layer that checks for valid ISO string.

jrenee42: can you please also put in a comment that explains (in english....with a valid example) of what the regex is trying to validate.  (i find them hard to read and easy to get wrong; so an explanation helps making sure everything is good)

ChitlangeSahas: For sure I can!

ChitlangeSahas: Update: i ended up discarding the regex idea and wrapped the `new Date()` in a try catch, so that when it crashes the function catches that and returns false. My regex solution was only partially correct because there are more than one valid ISO formats.

hoorayimhelping: cool, that is what I probably would have ended up recommending anyway. Regexes, while very powerful, are often full of non-obvious edge cases that have to be discovered over time (just read over some of the comments on SO posts that have regexes - there's almost always a history of comments saying "this doesn't work for [specific edge case]." This way, we let the JavaScript engine do all the work for us

src/shared/utils/dateTimeUtils.ts
```diff
@@ -42,3 +42,13 @@ export function addDurationToDate(
     }
   }
 }
+
+// checks whether the passed date is ISO format
+export function isISODate(dateString: string): boolean {
+  try {
+    const date = new Date(dateString)
+    return date.toISOString() === dateString
+  } catch (error) {
```
hoorayimhelping, line 51: not a big deal or worth fixing up since it's irrelevant, but I think if you're not going to make use of `error` you can just do:

```ts
try {
  const date = new Date(dateString)
  return date.toISOString() === dateString
} catch {
  return false
}
```

hoorayimhelping (approved):

src/shared/utils/dateTimeUtils.ts
ChitlangeSahas: Ahh! I tried leaving it like this: 
```
} catch() {
  return false
}
```
and it didn't work, I see, I need no parenthesis. I'll keep that in mind for the future 👍

PR #2385, opened 2021-08-16
fix: some notebooks sidebar toggling issues, default names for notebooks and export as alert task stuff

### LINKED ISSUE #2210: Notebooks: flowSidebar: Clicking the Functions button should collapse the functions list
I think this interaction should be improved. I think it can be summarized as follows:

- clicking the functions button should toggle the functions list. This seems to work for opening it up, but not for closing it.
- clicking the three dots menu should always open the menu on first click. currently, if you click the functions button, then the three dots, nothing happens. you need to click it again.
- clicking the three dots shouldn't close the functions list. Since we have a dedicated button for the functions list, this feels unnecessary. 

Kapture 2021-08-02 at 14 05 45

### LINKED ISSUE #2370: Notebooks: Updates to Export as Task naming and placement and Check Every and Offset placement
We'd like to make the following tweaks to the placement and naming:

- Rename "Export as Task" to "Export as Alert Task"
- Move the "Export as Alert Task" button to be in the bottom right corner of the alert panel
- Move Check Every and Offset fields to be below the conditions

### LINKED ISSUE #2380: Notebooks: Update to auto-populating Notebooks title
We'd like to update the auto-populated for all notebooks to use the name "Untitled Notebook". Since we've added metadata to the notebook list items, the current auto-populated name that includes the time the document was created is redundant. "Untitled ____" is also commonly used for auto-populating the name of a document. 

Current auto-populated name:

New auto-populated name:

Asalem1:
Closes #2210
Closes #2370 
Closes #2380 

notebooks-stuff


src/flows/pipes/Schedule/ExportTaskButton.tsx
```diff
@@ -33,17 +34,19 @@ const ExportTaskButton: FC<Props> = ({generate}) => {
     })
   }
 
+  const text = `Export as${alert ? ' Alert' : ''} Task`
```
Asalem1, line 37: this feels wrong

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2385)

src/flows/pipes/Schedule/ExportTaskButton.tsx
drdelambre: totally does. why not just pass in the `text` to the component? it should be more flexible than a boolean in case we decide to add more export task types

Asalem1: totally. I was thinking of trying to use the ID to identify whether it should render one string or another. The consideration there is that the title of the overlay should also change to Alert Task rather than Task right? Unless we wanna follow that pattern and have the title of the overlay controlled by a parent passing in a prop?

drdelambre: i mean you're injecting a lot of context into this button that it doesn't really need right? you're forcing it to have a concept of an alert, and a task, and whatever else comes down the pipe, when it's not doing anything with those concepts. instead, it'd be cleaner to say "you are a button and you display text" and have the parent push in whatever text needs to show up here.

drdelambre (approved):

PR #2391, opened 2021-08-17
feat: tokens accordion for buckets and telegrafs

### LINKED ISSUE #1925: Tokens: Buckets and Telegraf configs have expandable accordions with individual configuration options
Both Buckets and Telegraf config resources have expandable accordions that have individual configuration options.  
If the header level is clicked, then the individual levels are all clicked.

Palakp41:
Closes #1925 

Kapture 2021-08-17 at 13 24 10


src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
```diff
@@ -9,13 +9,19 @@ import {
...
 } from '@influxdata/clockface'
 
+import ResourceAccordion from './ResourceAccordion'
+
 interface OwnProps {
   onClose: () => void
 }
 
 export const CustomApiTokenOverlay: FC<OwnProps> = props => {
+  const resources = ['telegrafs', 'buckets']
```
Palakp41, line 24: FYI: Resources hardcoded right now but will change in a different issue, where we will use the api to get these resources.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2391)

src/authorizations/components/redesigned/ResourceAccordion.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+interface OwnProps {
+  resources: string[]
+}
+interface State {
+  permissions: any
+}
+interface StateProps {
+  telegrafPermissions: any
+  bucketPermissions: any
```
hoorayimhelping, line 34: do we have any tickets to address the `any` types here? should we?

src/authorizations/components/redesigned/ResourceAccordion.tsx
```diff
@@ -0,0 +1,154 @@
...
+        buckets: props.bucketPermissions,
+      },
+    }
+  }
+
+  public render() {
+    const {resources} = this.props
+    const {permissions} = this.state
+
+    if (resources) {
```
hoorayimhelping, line 67: stylistic nit: it's probably simpler and easier to follow if we invert this and do

```ts
if (!resources) {
  return null
}
```
and then have the `resources.map` be the default codepath

src/authorizations/components/redesigned/ResourceAccordion.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const buckets = getAll<Bucket>(state, ResourceType.Buckets)
+  const bucketPermissions = {}
+  buckets.forEach(b => {
+    bucketPermissions[b.id] = {
+      id: b.id,
+      orgID: b.orgID,
+      name: b.name,
+      permissions: {read: false, write: false},
+    }
+  })
```
hoorayimhelping, line 149: what do you think about changing `t` to ~`token`~ `telegraf` and `b` to `bucket`?

TCL735:
This is good except for the two possibly CSS-related issues that I DM'ed you about regarding the flickering and inconsistent clickable area that expands the bar.

TCL735 (approved):

TCL735:
> This is good except for the two possibly CSS-related issues that I DM'ed you about regarding the flickering and inconsistent clickable area that expands the bar.

They could also be isolated to Clockface components, which should be addressed there.

src/authorizations/components/redesigned/ResourceAccordionBody.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const {resourceName, permissions, onToggle, title} = props
+
+  const handleReadToggle = id => {
+    onToggle(resourceName, id, PermissionType.Read)
+  }
+  const handleWriteToggle = id => {
+    onToggle(resourceName, id, PermissionType.Write)
+  }
+
+  const accordionBody = telegraf => (
```
hoorayimhelping, line 35: can this function definition be pulled outside of the component so that it's only defined once, but called on every render? I don't think it needs to be re-defined every render call, right?

edit: looks like it depends on `handleReadToggle` and `handleWriteToggle` which depend on `resourceName`, so if those two functions are changed to have `id` *and* `resourceName` be arguments, I think we can pull all three function definitions out of the render function.

i don't even know how much more efficient is is, but it seems like it couldn't hurt

src/authorizations/components/redesigned/ResourceAccordionHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export const ResourceAccordionHeader: FC<OwnProps> = props => {
+  const {resourceName} = props
+  const readToggle = () => {
+    const {permissions} = props
+    let tag = true
+    if (permissions && Object.keys(permissions).length > 0) {
+      Object.keys(permissions).map(key => {
+        permissions[key].permissions.read === true
```
hoorayimhelping, line 29: if I understand this correctly - we're trying to get the value of all permissions in the accordion and if they're all `true`, set the header as `true`, right?

i think `some` is the most appropriate function here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some.

>`some` tests whether at least one element in the array passes the test implemented by the provided function. It returns true if, in the array, it finds an element for which the provided function returns true; otherwise it returns false. It doesn't modify the array. 

```ts
tag = !Object.keys(permissions).some(key => permissions[key].permissions.read === false)
```

if it finds one instance of `read` equal to false, `some` passes the check and returns `true`. we negate the call (`!Object.keys...`), so if `false` is found, we'll actually set the value of `tag` to false and vice versa if it's not found.

src/authorizations/components/redesigned/ResourceAccordionHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+          ? tag === false
+            ? (tag = false)
+            : (tag = true)
+          : (tag = false)
+      })
+    } else {
+      tag = false
+    }
+    return tag ? true : false
```
hoorayimhelping, line 38: can just do

```ts
return tag
```
since it returns the boolean value of `tag`

src/types/tokens.ts
```diff
@@ -0,0 +1,4 @@
+export enum PermissionType {
+  Read = 'read',
+  Write = 'write'
+}
```
hoorayimhelping, line 4: we'll want a newline here. https://unix.stackexchange.com/a/18789/30757

hoorayimhelping (commented):
hey nice work! have some suggestions to make the code a little bit simpler / easier to follow

src/authorizations/components/redesigned/ResourceAccordion.tsx
Palakp41: Yes, we should. I created types for these

src/authorizations/components/redesigned/ResourceAccordion.tsx
Palakp41: good one :)

src/authorizations/components/redesigned/ResourceAccordion.tsx
Palakp41: Actually, I am going to create a type for this in a different ticket. This is a bit more complicated, and kind of depends on rest of the resources to be implemented.

src/authorizations/components/redesigned/ResourceAccordionBody.tsx
Palakp41: Good question, I think `handleReadToggle` and `handleWriteToggle` are onChange handlers, which means it only provides with the `value` of the toggle when called. I will still need handlers in this component to get the value of toggle clicked, then I can execute the functions that we moved out with `id` and `resourceName`. What do you think?

hoorayimhelping: it's up to you. I think this is a micro-optimization at this point, more just a general question on how to do things

hoorayimhelping (approved):

PR #2394, opened 2021-08-18
feat: onscroll handle notebooks presets

ShmuelLotman:
Closes #

Makes the scroll all purdy for the notebooks presets and removes the flag so that this can be a first class feature

### DIFF
--- a/cypress/e2e/shared/flows.test.ts
+++ b/cypress/e2e/shared/flows.test.ts
@@ -295,7 +295,7 @@ describe('Flows', () => {
     cy.getByTestID('nav-item-flows').click()
     cy.reload()
     cy.getByTestID('tree-nav').should('be.visible')
-    cy.getByTestID('resource-editable-name').should('be.visible')
+    cy.getByTestID('resource-editable-name').should('exist')
     cy.getByTestID('resource-editable-name').click()
 
     // visualizations should not exist
@@ -365,7 +365,7 @@ describe('Flows', () => {
     cy.getByTestID('nav-item-flows').click()
     cy.reload()
     cy.getByTestID('tree-nav').should('be.visible')
-    cy.getByTestID('resource-editable-name').should('be.visible')
+    cy.getByTestID('resource-editable-name').should('exist')
     cy.getByTestID('resource-editable-name').click()
 
     // visualizations should exist
--- a/src/flows/components/FlowsIndex.tsx
+++ b/src/flows/components/FlowsIndex.tsx
@@ -22,38 +22,59 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 import 'src/flows/components/controlSearchBar.scss'
 
-const fadeOutOnScroll = element => {
-  const header = document.getElementsByClassName('cf-page-header')[0]
-  if (!element || !header) {
-    return
-  }
-
-  const distanceToTop =
-    window.pageYOffset +
-    element.getBoundingClientRect().top +
-    header.getBoundingClientRect().top
-  const elementHeight = element.offsetHeight
-  const scrollTop = document.documentElement.scrollTop
-
-  let opacity = 1
-
-  if (scrollTop > distanceToTop) {
-    opacity = 1 - ((scrollTop - distanceToTop) / elementHeight) * 1.8
-  }
-
-  if (opacity >= 0) {
-    element.style.opacity = opacity
-  }
-}
-
-function scrollHandler() {
-  const elem = document.getElementById('presetContainer')
-  if (elem) {
-    fadeOutOnScroll(elem)
-  }
-}
-
 const FlowsIndex = () => {
+  const [showButtonMode, setShowButtonMode] = useState(false)
+
+  const fadeOutOnScroll = element => {
+    const header = document.getElementsByClassName('cf-page-header')[0]
+    if (!header) {
+      return
+    }
+
+    const distanceToTop =
+      window.pageYOffset + element.getBoundingClientRect().top
+    const elementHeight = element.offsetHeight
+    const scrollTop = document.documentElement.scrollTop
+
+    console.log(
+      'distance:',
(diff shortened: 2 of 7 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2394)

cypress/e2e/shared/flows.test.ts
```diff
@@ -295,7 +295,7 @@ describe('Flows', () => {
     cy.getByTestID('nav-item-flows').click()
     cy.reload()
     cy.getByTestID('tree-nav').should('be.visible')
-    cy.getByTestID('resource-editable-name').should('be.visible')
+    cy.getByTestID('resource-editable-name').should('exist')
```
genehynson, line 298: why was this changed required?

ShmuelLotman: because in the slimmed down cypress window its not visible due to the box of the presets, so we need it to _be_ there,but being visible wasn't possible

src/flows/components/FlowsIndex.tsx
```diff
@@ -61,32 +107,62 @@ const FlowsIndex = () => {
...
     >
-      <PageHeader fullWidth={false}>
+      <PageHeader
+        fullWidth={false}
+        className={`${showButtonMode && 'withButtonHeader marginHeader'}`}
+      >
         <Page.Title title={PROJECT_NAME_PLURAL} />
+        {showButtonMode && (
+          <>
+            <PresetFlows buttonMode={showButtonMode} />
```
drdelambre, line 117: why are the concept of "buttonMode" bleeding between these components? since it's already in a boolean conditional, just conditionally render something else

src/flows/components/FlowsIndex.tsx
```diff
@@ -61,32 +107,62 @@ const FlowsIndex = () => {
...
-          {!isFlagEnabled('presetFlows') && <FlowCreateButton />}
-        </Page.ControlBarRight>
-      </Page.ControlBar>
-      <Page.Contents scrollable={true}>
-        {isFlagEnabled('presetFlows') && <PresetFlows />}
-        <FlowCards flows={flowList} search={search} />
-      </Page.Contents>
+      <DapperScrollbars onScroll={scrollHandler} id="scrollFlows">
+        <Page.Contents fullWidth={false} id="fadebox" ref={fadingBoxRef}>
+          <PresetFlows buttonMode={false} />
```
drdelambre, line 142: surprised this didn't trip up the linter. remove this prop

src/flows/components/FlowsIndex.tsx
```diff
@@ -61,32 +107,62 @@ const FlowsIndex = () => {
...
+          <Page.ControlBar fullWidth={false}>
```
drdelambre, line 143: is fullWidth explicitly true? if not, remove this one as well

src/flows/components/PresetFlows.tsx
```diff
@@ -17,17 +17,53 @@ const PresetMap = {
...
+  ) : (
     <Grid>
       <Grid.Row>
         <Grid.Column
           widthXS={Columns.Twelve}
           widthSM={Columns.Eight}
           widthMD={Columns.Ten}
         >
-          <div className="flows-index--presetContainer">
+          <div id="presetContainer" className="flows-index--presetContainer">
```
drdelambre, line 66: dont use ids in react, it causes weird shit to happen. if you need to access something directly, use a ref

src/flows/components/FlowsIndex.tsx
```diff
@@ -61,32 +107,62 @@ const FlowsIndex = () => {
       titleTag={pageTitleSuffixer([PROJECT_NAME_PLURAL])}
       testID="flows-index"
     >
-      <PageHeader fullWidth={false}>
+      <PageHeader
+        fullWidth={false}
+        className={`${showButtonMode && 'withButtonHeader marginHeader'}`}
```
drdelambre, line 112: i didn't realize you created both of these classes. merge their definitions and get rid of one

drdelambre (changes requested):
keep it in a flag until it's live

cypress/e2e/shared/flows.test.ts
genehynson: how does the next line work then? Cypress' `.click()` only works if the element is visible

ShmuelLotman: it works even if its off screen, as far as i could see and local testing bore out that its ok

genehynson: huh ok! cypress must have changed that behavior then

src/flows/components/FlowsIndex.tsx
ShmuelLotman: yeah it looks like it is unfortunately

src/flows/components/PresetFlows.tsx
ShmuelLotman: this was a floating id, thanks for pointing this out, removing it. It wasn't actually doing anything

src/flows/components/controlSearchBar.scss
```diff
@@ -0,0 +1,40 @@
+.cf-page-control-bar--fixed {
+  padding: 0px !important;
+}
+
+.flows-index--presetButtons {
+  animation: fadeIn 0.5s;
+  -webkit-animation: fadeIn 0.5s;
```
drdelambre, line 7: i dont think you need these browser specific rules as the browser wars are over and any rubble that remains should be added by our css parser.

drdelambre (approved):

PR #2411, opened 2021-08-18
test: fix cypress tsconfig

genehynson:
Found that we should be scanning all .ts files in sub-directories in our cypress ts-config

### DIFF
--- a/cypress/tsconfig.json
+++ b/cypress/tsconfig.json
@@ -15,6 +15,6 @@
   "include": [
     "../node_modules/cypress/**/*",
     "../node_modules/@types/chai/*",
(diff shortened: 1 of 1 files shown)


TCL735 (approved):

hoorayimhelping:
is this for typechecking test files to avoid things like missing  a `return` in a promise?

genehynson:
> is this for typechecking test files to avoid things like missing a `return` in a promise?

I don't think we have that specific linting rule turned on. This just fixes an issue I was experiencing in my IDE where everything that started with `cy.` was underlined in red.

PR #2421, opened 2021-08-19
fix: wait for root page to load before continuing test

genehynson:
Still trying to hunt down the cause for the timeouts during the initial UI page load in CI. The /api/v2/me requests are being aborted which means that we're navigating away from the root page prematurely. And because the request is aborted, the auth flows gets screwed up which causes the infinite spinner.

Going to try and wait for the page to load after the initial `cy.visit('/')` to see if that helps. I also cleaned up the handling of promises in the loginViaDex function

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -113,7 +113,6 @@ export const loginViaDex = (username: string, password: string) => {
                 })
                 .then(() => {
                   cy.visit('/')
-                  // cy.getCookie('session').should('exist')
                   cy.location('pathname').should('not.eq', '/signin')
                   return cy.getByTestID('tree-nav')
                 })
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -87,33 +87,38 @@ export const loginViaDex = (username: string, password: string) => {
           followRedirect: false,
           method: 'GET',
         })
-        .then(secondResp => {
-          cy.request({
-            url: Cypress.env(DEX_URL_VAR) + secondResp.headers.location,
-            method: 'POST',
-            form: true,
-            body: {
-              login: username,
-              password: password,
-            },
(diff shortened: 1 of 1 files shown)


cypress/support/commands.ts
```diff
@@ -87,33 +87,38 @@ export const loginViaDex = (username: string, password: string) => {
...
+                .request({
+                  url: thirdResp.redirectedToUrl,
+                  followRedirect: true,
+                  form: true,
+                  method: 'POST',
+                  body: {req: req, approval: 'approve'},
+                })
+                .then(() => {
+                  cy.visit('/')
+                  // cy.getCookie('session').should('exist')
```
hoorayimhelping, line 116: might as well remove this line

cypress/support/commands.ts
```diff
@@ -87,33 +87,38 @@ export const loginViaDex = (username: string, password: string) => {
...
+                  cy.location('pathname').should('not.eq', '/signin')
+                  return cy.getByTestID('tree-nav')
```
hoorayimhelping, line 118: this diff is all sorts of wacky. This line is the main change, right?

hoorayimhelping (approved):

cypress/support/commands.ts
genehynson: yeah I also cleaned up the returns - changed ` => {` to `=>`

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2421)

PR #2436, opened 2021-08-23
feat: query dependents

### LINKED ISSUE #2242: Notebooks: Idea: Add the ability to run a single panel and update panels "below" it
use case:
i was using a notebook to collect an bunch of individual stats from a measurement. Once I had one stat, i added a new source underneath it and continued working. as the list of individual stats started to grow, previewing my notebook took longer and longer, even though i wasn't interested in the pervious panels.

proposal:
add a way to preview just a single panel, and update the panels after it (or dependant on it, not sure what the right wording is). Ideally, i'd like to run the panel and update any visualizations related to that panel without executing all the queries in the notebook.

ShmuelLotman:
Closes #2242 

Adds ability to query dependents in the dropdown of the panel sidebar

### DIFF
--- a/src/flows/components/Sidebar.tsx
+++ b/src/flows/components/Sidebar.tsx
@@ -119,7 +119,7 @@ export const MenuButton: FC<ButtonProps> = ({id}) => {
 
 const Sidebar: FC = () => {
   const {flow, add} = useContext(FlowContext)
-  const {getPanelQueries} = useContext(FlowQueryContext)
+  const {getPanelQueries, queryDependents} = useContext(FlowQueryContext)
   const {id, hide, menu, showSub} = useContext(SidebarContext)
 
   const sections = ([
@@ -213,6 +213,13 @@ const Sidebar: FC = () => {
         {
           title: 'Export to Client Library',
           menu: <ClientList />,
+        },
+        {
+          title: 'Preview Subsequent',
+          action: () => {
+            event(`Querying panel ${flow.data.get(id).id} and its dependents`)
+            queryDependents(flow.data.get(id).id)
+          },
         },
       ],
     },
--- a/src/flows/context/flow.query.tsx
+++ b/src/flows/context/flow.query.tsx
@@ -38,6 +38,7 @@ export interface FlowQueryContextType {
   basic: (text: string) => any
   simplify: (text: string) => string
   queryAll: () => void
+  queryDependents: (startID: string) => void
   getPanelQueries: (id: string, withSideEffects?: boolean) => Stage
   status: RemoteDataState
   getStatus: (id: string) => RemoteDataState
@@ -50,6 +51,7 @@ export const DEFAULT_CONTEXT: FlowQueryContextType = {
   basic: (_: string) => {},
   simplify: (_: string) => '',
   queryAll: () => {},
+  queryDependents: () => {},
   getPanelQueries: (_, _a) => ({
     id: '',
     scope: {vars: {}},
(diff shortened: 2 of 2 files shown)


src/flows/components/Sidebar.tsx
```diff
@@ -214,6 +214,13 @@ const Sidebar: FC = () => {
           title: 'Export to Client Library',
           menu: <ClientList />,
         },
+        {
+          title: 'Preview Subsequent',
+          action: () => {
+            event(`Querying panel ${flow.data.get(id).id} and its dependents`)
```
Asalem1, line 220: wondering how relevant it is to get the id for eventing vs just knowing how often this triggered? Feels excessive

Asalem1, line 220: wondering how relevant it is to get the id for eventing vs just knowing how often this triggered? Feels excessive

src/flows/context/flow.query.tsx
```diff
@@ -278,6 +280,58 @@ export const FlowQueryProvider: FC = ({children}) => {
...
+    sendEvent(notebookQueryKey)
+    _queryDependents(startID)
+  }
+
+  const _queryDependents = (startID: string) => {
+    if (status === RemoteDataState.Loading) {
+      return
+    }
+
+    let map = generateMap(runMode === RunMode.Run)
```
Asalem1, line 293: just curious why we're going with the run implementation vs preview. I no longer understand the difference or what the consequences of this condition is vs the alternative

src/flows/context/flow.query.tsx
```diff
@@ -278,6 +280,58 @@ export const FlowQueryProvider: FC = ({children}) => {
...
+
+    if (!map.length) {
+      return
+    }
+
+    map = map.slice(map.indexOf(map.find(m => m.id === startID)))
```
Asalem1, line 299: this feels way more complicated than it needs to be. Maybe i'm just confused, why not just get an id reference and slice from there rather than do the find, index, slice? Something like:

```
const index = map.indexOf(m => m.id === startID)
if (index === -1) {
  ... not sure what we'd wanna do here
} else {
  map = map.slice(index)
}

File: src/flows/context/flow.query.tsx [hunk 4]
```diff
@@ -278,6 +280,58 @@ export const FlowQueryProvider: FC = ({children}) => {
...
+    event('Running Notebook QueryAll')
```
Asalem1, line 300:
```suggestion
    event('Running Notebook QueryDependents')
```

src/flows/context/flow.query.tsx
```diff
@@ -278,6 +280,58 @@ export const FlowQueryProvider: FC = ({children}) => {
...
+              setLoading(loading)
+              forceUpdate(stage.id, response)
+            })
+            .catch(e => {
+              forceUpdate(stage.id, {
+                error: e.message,
+              })
+              loading[stage.id] = RemoteDataState.Error
+              setLoading(loading)
+            })
```
Asalem1, line 320: nit on the setLoading. I feel like we might find ourselves in a weird lifecycle place with this logic. is it possible to refactor this to use a `setLoading(prev => ())` approach to prevent potential weirdness?

src/flows/context/flow.query.tsx
ShmuelLotman: That makes two of us

src/flows/context/flow.query.tsx
ShmuelLotman: ooh an opportunity for me to use findIndex :)

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2436)

src/flows/components/Sidebar.tsx
```diff
@@ -214,6 +214,13 @@ const Sidebar: FC = () => {
           title: 'Export to Client Library',
           menu: <ClientList />,
         },
+        {
+          title: 'Preview Subsequent',
```
drdelambre, line 218: why is this in here?

drdelambre: more constructively, the intent of this product change is to make iterating on an individual panel feel faster. less waiting for irrelevant panels to run, faster iteration per panel (submitting multiple times before discovering the right flux combination), lighter weight exploration. hiding that behind a dropdown and giving a weird name like "preview subsequent" seems counter productive? can we instead put it in the footer of every panel with a play icon that says "preview / run from here"?

ShmuelLotman: Sure thing. This was an initial pass at putting this somewhere, precisely so i could get this feedback. Your point makes perfect sense, ill drop it in the footer! Thanks @drdelambre

PR #2440, opened 2021-08-23
chore: reduce the size of the first overlay in Plugin page configuration creation

TCL735:
Part of #1555 

BEFORE, too wide:
Screen Shot 2021-08-19 at 1 47 54 PM

AFTER, much better:
Screen Shot 2021-08-19 at 1 48 33 PM


hoorayimhelping (approved):
easily the best diff i've ever seen

PR #2448, opened 2021-08-24
chore: change alerts activity text

### LINKED ISSUE #2422: Alerts display on home screen says checks don't exist when they do
In tools, I was looking at the list of alerts on the home screen, and noticed that some of them were clickable and some were not. When I hovered over the ones that were not links, I see the message "The check that created this no longer exists".  But, if I go to the alerts screen, I can see that check. Also, if a check is running, I don't know how it doesn't exist.  See screenshots for details.

Screen Shot 2021-08-19 at 10 17 41 AM
Screen Shot 2021-08-19 at 10 18 02 AM

subirjolly:
Closes #2422 

When you hover over a custom task which is listed in the Alerts Activity table, the current message is: `The check that created this no longer exists`. This needs to be updated so that this message makes more sense and is more relevant to the actual scenario where this activity row is actually generated by a custom task. New message: `This status was generated from a custom alert Task`

### DIFF
--- a/src/checks/components/CheckActivityTableField.tsx
+++ b/src/checks/components/CheckActivityTableField.tsx
@@ -22,7 +22,7 @@ const CheckActivityTableField: FC<Props> = ({row: {checkName, checkID}}) => {
     return (
       <div
         className="check-name-field"
-        title="The check that created this no longer exists"
+        title="This status was generated from a custom alert Task"
       >
         {checkName}
       </div>


Asalem1:
I'm confused. This doesn't seem to resolve any confusion. What happens when we run a task and delete that task? In that case the previous message of a task no longer existing is relevant. I don't think the solution - even temporarily - is to change the text to be more applicable for a case where a check doesn't exist. If it is because the system bucket is running a task, then we need a way to differentiate between that case and the case that I outlined.

PR #2449, opened 2021-08-24
feat: integrate test notifications for endpoints

### LINKED ISSUE #2287: Notebooks: I can test an alert from the alert panel
We would like for users to be able to test an alert from the alert panel to ensure the alert endpoint is set up properly, message shows up as intended, and alert conditions are accurately set.

### Acceptance Criteria

- When a user clicks on the test endpoint button a message is sent to the chosen endpoint
- When the test alert is sent, a notification at the top of the notebook shows that a test alert has been sent to the chosen endpoint

Design: @juliajames12 

Design file: https://www.figma.com/file/a7qLfWdlUDrhab9aWIxE65/Create-an-Alert-from-Notebook?node-id=684%3A4957

Asalem1:
Closes #2287

This PR adds the ability for a user to send a test notification with integrated services like making an http post request, a pagerduty alert, and a slack notification.

### DIFF
--- a/src/flows/pipes/Notification/Endpoints.tsx
+++ b/src/flows/pipes/Notification/Endpoints.tsx
@@ -2,6 +2,8 @@ import Slack from 'src/flows/pipes/Notification/Slack'
 import PagerDuty from 'src/flows/pipes/Notification/PagerDuty'
 import HTTP from 'src/flows/pipes/Notification/HTTP'
 
+const TEST_NOTIFICATION = 'This is a test notification'
+
 export const DEFAULT_ENDPOINTS = {
   slack: {
     name: 'Slack',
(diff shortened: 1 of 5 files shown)


src/flows/pipes/Notification/Endpoints.tsx
```diff
@@ -27,6 +31,16 @@ export const DEFAULT_ENDPOINTS = {
...
   )`,
+    generateTestQuery: data => `
+  slack.message(
+    url: "${data.url}",
+    channel: "${data.channel || ''}",
+    text: "${TEST_NOTIFICATION}",
+    color: "${data.color || '#34BB55'}"
+  )
+
+  array.from(rows: [{value: 0}])
```
Asalem1, line 42: need the array.from on all these or the query returns an error because ... flux

src/flows/pipes/Notification/PagerDuty.tsx
```diff
@@ -43,7 +43,7 @@ const PagerDuty: FC = () => {
       <Form.Element label="Routing Key">
         <Input
           name="key"
-          type={InputType.Text}
+          type={InputType.Password}
```
Asalem1, line 46: changed this for consistency. Not married to the idea, but we should consider allowing the user to toggle visibility here

drdelambre: there's a greater arc where we're supposed to have support for secrets in here, but that hasn't come from design yet, so this is totally fine

drdelambre (approved):
this looks great!

PR #2454, opened 2021-08-24
test: update test e2e remocal script

TCL735:
Now that remocal namespaces are not limited to starting with "twodotoh-dev" let's update the script to allow e2e tests to run locally with any namespace name.


hoorayimhelping (approved):
I love it and am all for it!

let's give a chance for more folks to weigh in before merging though in case this messes up people's workflows

genehynson (approved):
lgtm! We don't use this command in CI, so the only folks affected will be the UI devs. A message in #team-ui should suffice.

PR #2498, opened 2021-08-30
fix: query variable read-only issue fix

### LINKED ISSUE #2439: Alerts Creation Broken
Thank you for reporting a bug in InfluxData UI.

- Please ask usage questions on the Influx Community site.
  - https://community.influxdata.com/
- Please add a :+1: or comment on a similar existing bug report instead of opening a new one.
  - https://github.com/influxdata/ui/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+is%3Aclosed+sort%3Aupdated-desc+label%3Akind%2Fbug+
- Please check whether the bug can be reproduced with the latest release.
- The fastest way to fix a bug is to open a Pull Request.
  - https://github.com/influxdata/ui/pulls

**Steps to reproduce:**
`Alerts` creation broken when you enter `UI` and without using any other functionality, navigate to `Alerts` page and try to create an alert. Attached are the 2 gifs where Alerts creations is:
1. Working(in **working.gif**) when you use other functionality before attempting to create an alert.
working

2. Broken(in **broken.gif**) when you navigate to `Alerts` page directly and attemp to create an alert.
broken

**Expected behavior:**
Alerts creation **should work**.

**Actual behavior:**
Gets the following error when used in erraneous condition.
Screen Shot 2021-08-23 at 4 30 26 PM

**Environment info:**
Local, Tools

subirjolly:
Closes #2439 

`vari` is a read-only variable and that prevents us from overwriting `results` in `vari?.arguments?.values?.results`. So, the solution is to make a deep copy and use it as the variable so that we can overwrite any part of `vari` and its children.

### DIFF
--- a/src/variables/selectors/index.tsx
+++ b/src/variables/selectors/index.tsx
@@ -148,10 +148,6 @@ export const getVariable = (state: AppState, variableID: string): Variable => {
   const ctx = state?.resources?.variables?.values?.[contextID]
   let vari = state?.resources?.variables?.byID?.[variableID]
 
-  if (vari) {
-    vari = JSON.parse(JSON.stringify(vari))
-  }
-
   if (ctx && ctx.values && ctx.values.hasOwnProperty(variableID)) {
     vari = {...vari, ...ctx.values[variableID]}
   }
--- a/src/variables/selectors/index.tsx
+++ b/src/variables/selectors/index.tsx
@@ -145,13 +145,11 @@ export const sortVariablesByName = (variables: Variable[]): Variable[] =>
 
 export const getVariable = (state: AppState, variableID: string): Variable => {
   const contextID = currentContext(state)
-  const ctx = get(state, ['resources', 'variables', 'values', contextID])
-  let vari = get(state, ['resources', 'variables', 'byID', variableID])
+  const ctx = state?.resources?.variables?.values?.[contextID]
+  let vari = state?.resources?.variables?.byID?.[variableID]
 
   if (vari) {
-    vari = JSON.parse(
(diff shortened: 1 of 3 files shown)


src/variables/selectors/index.tsx
```diff
@@ -148,6 +148,12 @@ export const getVariable = (state: AppState, variableID: string): Variable => {
   const ctx = get(state, ['resources', 'variables', 'values', contextID])
   let vari = get(state, ['resources', 'variables', 'byID', variableID])
 
+  if (vari) {
+    vari = JSON.parse(
+      JSON.stringify(get(state, ['resources', 'variables', 'byID', variableID]))
```
drdelambre, line 153: Don't use get, use ?

drdelambre: Also why are you refetching this and not wrapping the original accessor?

drdelambre:
I don't understand the readonly terminology here when you triggered a default immutable state. So even if the variables aren't updated, they get seen as being updated by the shallow object compare. This would be better applied at the mutation and not at the selector

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2498)

drdelambre:
TURNS OUT THE VARIABLES DIDNT EXIST.. or something like that

drdelambre (approved):
lol an hour for adding a single resourcetype. i hate this variable subsystem

PR #2508, opened 2021-08-30
fix(notebooks): should be able to delete broken panels

### LINKED ISSUE #2469: Corrupt notebook (from column selector panel)
I created a notebook in prod with a flux query.  I then added a column selector. It seemed to work at first, but then when I went to preview it again, I got an error.  Now the notebook shows with an error (instead of the column selector) and if I try to delete the column selector panel I get the error:

TypeError: Cannot read property 'family' of undefined
    at Object.disable (Sidebar.tsx:186)
    at Sidebar.tsx:225
    at Array.filter (<anonymous>)
    at Sidebar.tsx:223
    at Array.map (<anonymous>)
    at Y (Sidebar.tsx:221)
    at Vo (react-dom.production.min.js:150)
    at ki (react-dom.production.min.js:178)
    at Fa (react-dom.production.min.js:232)
    at Na (react-dom.production.min.js:233)
ra @ react-dom.production.min.js:198
i.componentDidCatch.n.callback @ react-dom.production.min.js:211
Ki @ react-dom.production.min.js:193
Xi @ react-dom.production.min.js:193
La @ react-dom.production.min.js:216
Ba @ react-dom.production.min.js:220
(anonymous) @ react-dom.production.min.js:250
t.unstable_runWithPriority @ scheduler.production.min.js:18
xs @ react-dom.production.min.js:250
As @ react-dom.production.min.js:249
Cs @ react-dom.production.min.js:248
Ts @ react-dom.production.min.js:251
Tn @ react-dom.production.min.js:85
A.Object.isExtensible.Object.isExtensible.e.___hb.e.___hb @ honeybadger.js:466

Notebook export is here:
{
  "data": {
    "orgID": "3f730102d8daf6ed",
    "name": "Untitled Notebook",
    "spec": {
      "name": "Untitled Notebook",
      "readOnly": false,
      "range": {
        "type": "selectable-duration",
        "label": "Past 1h",
        "lower": "now() - 1h",
        "upper": null,
        "seconds": 3600,
        "duration": "1h",
        "windowPeriod": 10000
      },
      "refresh": {
        "status": "paused",
        "duration": null,
        "interval": 0,
        "infiniteDuration": false,
        "inactivityTimeout": 0
      },
      "createdAt": "2021-08-26T00:18:28.755Z",
      "updatedAt": "2021-08-26T00:23:09.280Z",
      "pipes": [
        {
          "id": "local_8a43309f-1f8a-4695-8bf4-03a37215bd2b",
          "type": "rawFluxEditor",
          "queries": [
            {
              "text": "// Uncomment the following line to continue building from the previous cell\n// __PREVIOUS_RESULT__\n\nfrom(bucket: \"bnelson's Bucket\")\n  |> range(start: -15m, stop: now())\n  |> group(columns: [\"host\", \"_measurement\"], mode:\"by\")\n\n",
              "editMode": "advanced",
              "builderConfig": {
                "tags": [],
                "buckets": [],
                "functions": []
              }
            }
          ],
          "activeQuery": 0,
          "title": "Query to Run",
          "visible": true
        },
        {
          "id": "local_41472780-80fc-4e6e-b876-664230674d04"
        },
        {
          "id": "local_f072aef3-0655-4d51-b486-09f628352572",
          "type": "visualization",
          "period": "10s",
          "properties": {
            "type": "simple-table",
            "showAll": false
          },
          "title": "Validate the Data",
          "visible": true
        }
      ]
    }
  }
}

genehynson:
Related #2469 

Not sure what caused the original issue and I was unable to reproduce it after trying all morning. But I was able to add the broken notebook into my env and from there, found the reason why the broken panel couldn't be deleted. Only happens when the `flowSidebar` flag is on. 

before:

image

after:

https://user-images.githubusercontent.com/6411855/131393217-640ff5ea-5b80-47c6-8ef6-403634ed886e.mov

### DIFF
--- a/src/flows/components/Sidebar.tsx
+++ b/src/flows/components/Sidebar.tsx
@@ -183,7 +183,7 @@ const Sidebar: FC = () => {
               return true
             }
 
-            if (!/^(inputs|transform)$/.test(PIPE_DEFINITIONS[type].family)) {
+            if (!/^(inputs|transform)$/.test(PIPE_DEFINITIONS[type]?.family)) {
               return true
             }
 


src/flows/components/Sidebar.tsx
```diff
@@ -183,7 +183,7 @@ const Sidebar: FC = () => {
               return true
             }
 
-            if (!/^(inputs|transform)$/.test(PIPE_DEFINITIONS[type].family)) {
```
Asalem1, line 186: 🤷🏽 super confused how this is it but 🤷🏽 ?

Asalem1 (approved):

drdelambre (approved):
:fire:

drdelambre:
but this doesn't close the issue

genehynson:
> but this doesn't close the issue, just bumps down the severity

valid point, I won't close the issue

PR #2519, opened 2021-08-31
chore: add coveo script behind FF

joelfried:
This way we can toggle it for environments as needed.

The dynamic injection is the expedient method of getting this feature out behind a flag for demonstration purposes. The public key has been provided by Coveo and this was an implementation suggestion that allows us to work with their team on ensuring our integration is working as necessary. This shouldn't introduce any security concerns im aware of, and will allow us to work further with Coveo on developing a future-proof solution for integration

While this does represent a diversion from our usual implementation patterns, it should be noted that this suggestion was provided directly by the Coveo team and allows us to iterate quickly on the feature within a testing context and a public key that they can immediately invalidate should anything arise that is of concern.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -62,6 +62,16 @@ const App: FC = () => {
         })
       }
     }
+    if (CLOUD && isFlagEnabled('coveo')) {
+      const script = document.createElement('script')
+      script.src =
+        'https://platform.cloud.coveo.com/rest/organizations/influxdatanonproduction1eh1cn7go/pages/95a20052-218d-4048-9081-16dbcebbdebb/inappwidget/loader'
+      script.async = true
+      document.body.appendChild(script)
+      return () => {
+        document.body.removeChild(script)
+      }
+    }
     setAutoFreeze(false)
   }, [])
 


src/App.tsx
```diff
@@ -62,6 +62,16 @@ const App: FC = () => {
         })
       }
     }
+    if (CLOUD && isFlagEnabled('coveo')) {
+      const script = document.createElement('script')
```
drdelambre, line 66: why are we dynamically injecting code like this?

drdelambre:
and not something from here, like the rest of our libraries we include
https://www.npmjs.com/search?q=%40coveo

src/App.tsx
ShmuelLotman: The dynamic injection is just to get this feature out behind a flag in order to demonstrate it to the Coveo team as a POC. Once this is done, we can work on getting a more stable / futureproof implementation in our code if necessary. I'll be in contact with Coveo after this is displayed to discuss alternative implementation options if needed.

src/App.tsx
```diff
@@ -62,6 +62,16 @@ const App: FC = () => {
...
+      script.src =
+        'https://platform.cloud.coveo.com/rest/organizations/influxdatanonproduction1eh1cn7go/pages/95a20052-218d-4048-9081-16dbcebbdebb/inappwidget/loader'
```
genehynson, line 68: is it safe to have this api-key embedded url in this open source repo?

ShmuelLotman: Yes, this is a public key that the Coveo team has sent us, they can invalidate it once we are done demonstrating this feature, and its not going to be connected to billing or anything that would endanger our users, at least not in any way im aware of.

genehynson: More worried about it endangering us - if someone decided they wanted to try out Coveo in their app and decided to use our url 😆

src/App.tsx
ShmuelLotman: While this does represent a diversion from our usual implementation patterns, it should be noted that this suggestion was provided directly by the Coveo team and allows us to iterate quickly on the feature within a testing context and a public key that they can immediately invalidate should anything arise that is of concern.

drdelambre: so we're going to remove it / relocate it once we've decided it's worth more investment?

drdelambre (approved):
great concern from gene, but i'm not too worried about the token as long as we get a new one before integrating this properly (with env vars piped into the image). can we confirm that we will be getting a new one before merging?

genehynson:
Pretty sure we'll need to update the nginx whitelist domains list to include coveo for this to work

PR #2525, opened 2021-08-31
fix: make the notebooks time range dropdown work

drdelambre:
omg this started as an attempt to make the timerange dropdown work with the flux editor in notebooks. for some reason, only when updating the flux in a flux panel, and not when updating any other panel, it would reset the time on the page to whatever the time previously was, where previously is defined as a random point in the past.

i got in there, and saw that some panels were updating with stale state, some with current state, and it took too flipping long to reason about any of it.

so i kinda removed the resource.hook thing... cause that was annoying, and I cleaned up a bunch of refreshing of state cycles (where it's just setting it's value to the same thing in a useEffect because a useCallback changed or something), because i couldn't reason about the issue with them firing off all the time

and i karate chopped a bunch of update functions.. because they were stale for no reason

then i got into a fun one where i couldn't add a panel and remove a panel in one... hook event loop iteration... because the outdated state of one was clobering the mutations of the other, because at the end of the day, immutable says "you gotta run new state = { ...old state, new props } somewhere". the way i found to resolve THAT issue was that you need to mutate both by cloning an object like we've been doing, but also mutate an object in place by reference so that the last update function to pick up on the change will get the real old state for the clone and not the OLD old state....

look at the addition of the remove and changes to add in the flow.current context for more details


daviddkkim:
This was a captivating read - good storytelling

src/flows/components/panel/FlowPanel.tsx
```diff
@@ -183,7 +206,7 @@ const FlowPanel: FC<Props> = ({
             <Handle
               dragRef={handleRef}
               onStartDrag={handleMouseDown}
-              dragging={isDragging}
+              dragging={isDragging == 2}
```
Asalem1, line 209:
```suggestion
              dragging={isDragging === 2}
```

src/flows/components/panel/FlowPanelTitle.tsx
```diff
@@ -12,16 +12,22 @@ interface Props {
...
 
   const onChange = (e: ChangeEvent<HTMLInputElement>): void => {
-    onTitleChange(e.target.value)
+    update({
+      meta: {
+        ...flow.meta,
+        byID: {
+          ...flow.meta.byID,
+          [id]: {
+            ...flow.meta.byID[id],
```
Asalem1, line 25: 🤮

src/flows/components/panel/PanelVisibilityToggle.tsx
```diff
@@ -24,8 +24,17 @@ const PanelVisibilityToggle: FC<Props> = ({id}) => {
...
+      meta: {
+        ...flow.meta,
+        byID: {
+          ...flow.meta.byID,
+          [id]: {
+            ...flow.meta.byID[id],
+            visible: !meta.visible,
+          },
+        },
+      },
```
Asalem1, line 37: I feel like top-level we can just do something like `update(prev => {...prev, allThatOtherStuff})` and that might reduce some of the lifecycle stuff?

src/flows/context/flow.current.tsx
```diff
@@ -47,20 +45,45 @@ export const FlowProvider: FC = ({children}) => {
     delete initial.title
     initial.id = id
 
-    flows[currentID].data.add(id, initial)
-    const {resource, onChange} = flows[currentID].meta.add(id, {
+    flows[currentID].data.byID[id] = initial
+    flows[currentID].meta.byID[id] = {
       title,
       visible: true,
-    }) as {resource: Resource<any>; onChange: (_: Resource<any>) => {}}
```
Asalem1, line 54: nice

ui-ci-bot-influx:
Metric | Score | Level
performance | 67 | :yellow_circle:
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2525)

Asalem1 (commented):
nothing glaringly obvious sticking around but there's a sticking about the data that's being passed into the update function. Like, it seems like an awful lot for a lower level component to need to know about that entire structure to update a property right? I know the reasons behind it, but maybe we don't need to do that, or maybe i'm just looking at it again in the wrong light? 

Also, should this affect ALL instances of `update` and `flow.data.get(id)`? Since I think there are a few out there that weren't touched

drdelambre:
i split the update function into updateMeta, updateData, and updateOther. Most of this complication is abstracted away at the pipe context, where it maps updateData to update. the others are used to update the notebook specific data

Asalem1 (approved):
approveOther

PR #2530, opened 2021-09-02
feat: completed deadman check functionality

### LINKED ISSUE #2286: Notebooks: add deadman checks to the alerts panel

Asalem1:
Closes #2286 

This feature adds deadman alerts as a possible option in the dropdown for the alerts panel. There's a lot of repetition here so I'm open to a different approach, especially one that isn't as brute-forced with setting the variables, but maybe it's not as brute forced as I think it is? Tried making do with the reduce function but things were getting a little out of control in there.

deadman-alert

### DIFF
--- a/src/flows/context/query.tsx
+++ b/src/flows/context/query.tsx
@@ -233,132 +233,90 @@ const _addWindowPeriod = (ast, optionAST): void => {
 }
 
 export const simplify = (text, vars: VariableMap = {}) => {
-  const ast = parse(text)
-  const usedVars = _getVars(ast, vars)
+  try {
+    const ast = parse(text)
+    const usedVars = _getVars(ast, vars)
 
-  // Grab all global variables and turn them into a hashmap
-  // TODO: move off this variable junk and just use strings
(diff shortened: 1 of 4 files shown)


drdelambre (approved):
looks good. thanks for gating both sides of the deadman selected condition (has to be the first one, can't add if it is the first one)

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2530)

PR #2532, opened 2021-09-02
feat: create token overlay has all the resources

### LINKED ISSUE #1928: Tokens: Get list of resources to populate resource list
List of resources available from OATs consuming cloud swagger.  Fill in clockface component in array.

Palakp41:
Closes #1928 

Kapture 2021-09-10 at 11 06 46

### DIFF
--- a/src/authorizations/actions/thunks.ts
+++ b/src/authorizations/actions/thunks.ts
@@ -168,17 +168,12 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
   }
 }
 
-export const getResources = () => async dispatch => {
-  try {
-    dispatch(setAllResources(RemoteDataState.Loading))
-    const resp = await api.getResources({headers: {}})
+export const getAllResources = () => async dispatch => {
+  const resp = await api.getResources({headers: {}})
 
-    if (resp.status !== 200) {
-      throw new Error(resp.data.message)
-    }
-    const resources = resp.data
-    dispatch(setAllResources(RemoteDataState.Done, resources))
-  } catch (e) {
-    console.error(e)
+  if (resp.status !== 200) {
+    throw new Error(resp.data.message)
   }
+  const resources = resp.data
+  dispatch(setAllResources(resources))
 }
--- a/src/authorizations/actions/thunks.ts
+++ b/src/authorizations/actions/thunks.ts
@@ -168,7 +168,7 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
   }
 }
 
-export const getResources = () => async (dispatch) => {
+export const getResources = () => async dispatch => {
   try {
     dispatch(setAllResources(RemoteDataState.Loading))
     const resp = await api.getResources({headers: {}})
@@ -178,8 +178,7 @@ export const getResources = () => async (dispatch) => {
     }
     const resources = resp.data
     dispatch(setAllResources(RemoteDataState.Done, resources))
-  }
-  catch (e) {
+  } catch (e) {
     console.error(e)
   }
 }
--- a/src/authorizations/actions/thunks.ts
+++ b/src/authorizations/actions/thunks.ts
@@ -16,6 +16,7 @@ import {
   setAuthorizations,
   removeAuthorization,
   setCurrentAuthorization,
+  setAllResources,
 } from 'src/authorizations/actions/creators'
 import {notify} from 'src/shared/actions/notifications'
 
@@ -164,5 +165,21 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
     event('token.delete.failure', {id, name})
     console.error(e)
     dispatch(notify(authorizationDeleteFailed(name)))
+  }
+}
+
+export const getResources = () => async (dispatch) => {
+  try {
+    dispatch(setAllResources(RemoteDataState.Loading))
+    const resp = await api.getResources({headers: {}})
+
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
+    }
+    const resources = resp.data
+    dispatch(setAllResources(RemoteDataState.Done, resources))
+  }
+  catch (e) {
+    console.error(e)
   }
 }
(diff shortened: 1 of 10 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2532)

src/authorizations/actions/thunks.ts
```diff
@@ -166,3 +167,18 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
...
+    const resp = await api.getResources({headers: {}})
+
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
+    }
+    const resources = resp.data
+    dispatch(setAllResources(RemoteDataState.Done, resources))
+  } catch (e) {
+    console.error(e)
+  }
```
hoorayimhelping, line 183: If we're not going to gracefully handle the error here, I think we should let it fail by removing this `catch` block and have the UI / component layer handle this with a notification

src/authorizations/components/redesigned/GenerateTokenDropdown.tsx
```diff
@@ -27,6 +29,7 @@ const GenerateTokenDropdown: FC<ReduxProps & GenerateTokenProps> = ({
   }
 
   const handleCustomApi = () => {
+    getResources()
```
hoorayimhelping, line 32: I think the error should be handled here. If `handleCustomApi` was made `async`, this call can get wrapped in a `try/catch` block and in the `catch` block, we can fire off a notification saying "there was an error, please reload" or something along those lines.

As it stands now, if the fetch fails, it will log in the console and not let the user know anything has happened.

src/authorizations/actions/thunks.ts
```diff
@@ -166,3 +167,18 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
     dispatch(notify(authorizationDeleteFailed(name)))
   }
 }
+
+export const getResources = () => async dispatch => {
```
hoorayimhelping, line 171: I'm a little confused about the intent of this - what does this do that the current `GetResources` component doesn't?

src/authorizations/components/redesigned/ResourceAccordion.tsx
```diff
@@ -72,32 +86,40 @@ class ResourceAccordion extends Component<Props, State> {
...
-          <GetResources resources={[ResourceType[resourceName]]}>
-            {!isEmpty(permissions[resource].sublevelPermissions) &&
-              this.getAccordionBody(resourceName, resource)}
-          </GetResources>
+          {resourceName === 'Telegrafs' || resourceName === 'Buckets' ? (
+            <GetResources resources={[ResourceType[resourceName]]}>
+              {!isEmpty(permissions[resource].sublevelPermissions) &&
+                this.getAccordionBody(resourceName, resource)}
+            </GetResources>
+          ) : null}
```
hoorayimhelping, line 94: not a huge deal, but i think it's more conventional in our codebase to do: 

```suggestion
          {resourceName === 'Telegrafs' || resourceName === 'Buckets' && (
            <GetResources resources={[ResourceType[resourceName]]}>
              {!isEmpty(permissions[resource].sublevelPermissions) &&
                this.getAccordionBody(resourceName, resource)}
            </GetResources>
          )}
```

src/authorizations/components/redesigned/ResourceAccordion.tsx
```diff
@@ -72,32 +86,40 @@ class ResourceAccordion extends Component<Props, State> {
...
+    if (resourceName === 'Telegrafs') {
+      return (
+        <ResourceAccordionBody
+          resourceName={resource}
+          permissions={permissions[resource].sublevelPermissions}
+          onToggle={this.handleIndividualToggle}
+          title="Individual Telegraf Configuration Names"
+        />
+      )
+    } else if (resourceName === 'Buckets') {
```
hoorayimhelping, line 111: again, not a huge deal, but stylistically you can have less syntax by removing these `else` statements since each `if` statement `returns`

hoorayimhelping (commented):
mainly a couple of questions, plus a suggestion about handling an api error

src/authorizations/actions/thunks.ts
Palakp41: @hoorayimhelping `GetResources` component takes an array of `ResourceTypes` as a prop and only fetches those resources. This new `getResources` function is fetching all resource names from the new api route that was created. Hope that makes sense!

hoorayimhelping: @Palakp41 that makes sense. What do you think about naming this `getAllResources` so it's clear we're fetching them all?

src/authorizations/actions/thunks.ts
Palakp41: Good idea

src/authorizations/components/redesigned/GenerateTokenDropdown.tsx
```diff
@@ -26,8 +31,13 @@ const GenerateTokenDropdown: FC<ReduxProps & GenerateTokenProps> = ({
     onShowOverlay('add-master-token', null, onDismissOverlay)
   }
 
-  const handleCustomApi = () => {
-    onShowOverlay('add-custom-token', null, onDismissOverlay)
+  const handleCustomApi = async () => {
+    try {
+      await getAllResources()
+      onShowOverlay('add-custom-token', null, onDismissOverlay)
+    } catch (e) {
```
hoorayimhelping, line 38: minor nit: I don't think this [`(e)` is necessary](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#the_exception_identifier) if it's not being used, but it's a minor thing. It's more just to make it slightly easier to read / understand.

hoorayimhelping (approved):
LGTM

src/authorizations/components/redesigned/GenerateTokenDropdown.tsx
Palakp41: Okay, will remove it in another pr

Issue #2160, opened 2021-07-27
Telegraf Refresh: plugin configuration text source of truth


TCL735:
The source of truth for the configuration text of each telegraf plugin is located here: https://github.com/influxdata/telegraf/blob/master/etc/telegraf.conf

As such, the above file needs to be parsed into individual configurations. The parsing should be automated, repeatable, and usable outside of the UI application (ie it should be a build script or other cli tool).

TCL735:
Branch here: https://github.com/influxdata/ui/tree/feat_2160_parse_telegraf_conf

TCL735:
Currently blocked by this: https://github.com/influxdata/ui/issues/2526

TCL735:
No longer blocked. Using a workaround.

TCL735:
As of this message, no longer blocked, as idpe has just been updated with the necessary work.

PR #2537, opened 2021-09-03
fix: update parsing and fix spacing in .conf files

TCL735:
Part of #2160 

Fixes:
- there was still a leading space in the .conf files that previously were commented out
- uncommented files inadvertently lost the leading open bracket


src/writeData/utils/parsePluginsConfigurationText.ts
```diff
@@ -260,12 +260,26 @@ const fetch = path => {
...
+    return (
+      configurationLines
+        .map(text => {
+          let sliceCounter = 0
+          if (text[0] === '#') {
+            sliceCounter += 1
+          }
+          if (text[1] === ' ') {
+            sliceCounter += 1
+          }
```
hoorayimhelping, line 276: this will work for any comment like:

```
#comment
# comment
```

yes ? will it work on

```
#       comment
##   comment
#     # comment
```
? a test might be helpful here

src/writeData/components/telegrafInputPluginsConfigurationText/swap.conf
```diff
@@ -1,3 +1,3 @@
- Read metrics about swap memory usage
-[inputs.swap]]
- # no configuration
+# Read metrics about swap memory usage
+[[inputs.swap]]
```
hoorayimhelping, line 2: looks like the newline at the end of files misses files with no configuration?

hoorayimhelping (commented):
I'm not 100% on what the intent is, so I'll leave these comments without a request or approval and see where we go from here

src/writeData/utils/parsePluginsConfigurationText.ts
TCL735: This part only happens when we know for sure the entire block starts is commented (every line starts with '#') like this
```
# # comment
# [[input.someplugin]]
#   importantArg=value
#   # some other comment about args
#  anotherArg=otherValue
#
```

So, we check if the first character is the comment. This is mostly unnecessary but just to be safe. Then we check if the next value is a space. It seems like commented blocks in `telegraf.conf` will always have a space following the leading comment. We want to trim that off. But only a single space, because we want to preserve the indentation.

Comments inside a commented block should follow the pattern. If they happen to have additional spacing before their own "#" then it really doesn't matter for purposes of the configuration file being valid. But I really don't think this would happen, as there is already additional pattern matching before we even get to this point of formatting the actual text.

src/writeData/components/telegrafInputPluginsConfigurationText/swap.conf
TCL735: Fixed

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2537)

src/writeData/utils/parsePluginsConfigurationText.ts
TCL735: I just found one edge case: `zipkin.conf`

But the exception here was for what is considered a commented block rather than weird spacing on the comments.

I pushed up a fix to account for cases like `zipkin.conf`

hoorayimhelping (approved):

zoesteinkamp (approved):
This seems like a fine style change to me

PR #2543, opened 2021-09-03
feat: Measurement schema- add read only schema section

### LINKED ISSUE #2507: add ability to view measurement schema names  and csv contents on a bucket when it is being edited
show the names of the current measurement schemas in the advanced tab so they can be viewed; along with a way to view/download the csv schema file that is the columns.

add a metric/event for each time a file is downloaded

this image; minus the 'add measurement schema button' is what the ui will look like after this issue is done.
note that if there are no measurement schemas, then nothing will be displayed.

once the 'add' function is implemented, then instead a string will display saying there are no measurement schemas on this bucket yet.

additionally, if the schema type is 'implicit'; then this section will *never* show as it is not relevant.

jrenee42:
Closes #2507 

adds read-only panel for viewing the measurement schemas, shows each schema's name and you can download each of the columns individually in a json file


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2543)

src/buckets/components/createBucketForm/BucketOverlayForm.tsx
```diff
@@ -18,8 +18,17 @@ import {
...
+import {MeasurementSchemaSection} from './MeasurementSchemaSection'
+
+let MeasurementSchemaList = null,
+  SchemaType = null
+
+if (CLOUD) {
+  SchemaType = require('src/client/generatedRoutes').MeasurementSchema
+  MeasurementSchemaList = require('src/client/generatedRoutes')
+    .MeasurementSchemaList
+}
```
hoorayimhelping, line 31: where did this pattern come from and is anyone aware if it's actually doing anything beneficial? it looks like it's there as a performance thing to not load components that aren't used (assumptions about intent are dangerous), but is that actually having any measurable effect on bundle size, app performance or feel of the app? did anyone check check up on any of this before putting this in the codebase?

it's definitely making the code more complex and as we've already seen, is a source of catastrophic errors.

src/buckets/components/createBucketForm/MeasurementSchemaSection.tsx
```diff
@@ -0,0 +1,111 @@
...
+interface PanelProps {
+  measurementSchema: typeof MeasurementSchema
+  index?: number
+}
+
+export const EditingPanel: FC<PanelProps> = ({measurementSchema, index}) => {
+  const handleDownloadSchema = () => {
+    const {name} = measurementSchema
+    const contents = JSON.stringify(measurementSchema.columns)
+    downloadTextFile(contents, name || 'schema', '.json')
```
hoorayimhelping, line 41: i think we should probably measure the usage of this new feature with some events. what do you think?

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
```diff
@@ -70,10 +79,16 @@ const UpdateBucketOverlay: FunctionComponent<Props> = ({
         handleClose()
         return
       }
+
       setBucketDraft(resp.data as OwnBucket)
 
       setSchemaType(resp.data.schemaType)
 
+      if ('explicit' === resp.data.schemaType) {
```
hoorayimhelping, line 87: oh, you don't see a lot of yoda conditionals these days. i don't think we do that anywhere and i think with our lint rules, we're safe from accidentally doing `resp.data.schemaType = 'explicit'`

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
```diff
@@ -70,10 +79,16 @@ const UpdateBucketOverlay: FunctionComponent<Props> = ({
...
+        const schema = await getSchema(bucketID)
+        setMeasurementSchemaList(schema)
```
hoorayimhelping, line 89: this is another place to put `event` calls for measuring usage of this feature

hoorayimhelping (commented):
have some questions about importing and adding `eventing`

src/buckets/components/createBucketForm/BucketOverlayForm.tsx
jrenee42: if it's not there, then the eslint when it is *not* cloud fails b/c those calls are only available in cloud.

i'm using the same exact pattern as in my earlier commits in this epic.  

i copied the pattern from elsewhere ; if there's a better pattern you know of please let me know and I can switch to it

jrenee42: see ui/src/cloud/actions/limits.ts

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
jrenee42: are you saying you don't like this style?  I prefer this style so put it in when I write new code

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
jrenee42: this is just getting the data when the user is viewing or editing a bucket.  so not actually an event.

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
jrenee42: (do you want me to reverse this, or don't bother?) thanks @hoorayimhelping

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
jrenee42: i *did* just check and there is no event fired when the user clicks the button to view the settings, so I added one there  ( event('bucket.view.settings')

src/buckets/components/createBucketForm/BucketOverlayForm.tsx
hoorayimhelping: thanks for the information. So this is less a performance pattern and more to get around fundamental issues in the architecture of the app. The first is that our app doesn't make any architectural distinction between oss and cloud. All the distinctions are made at code levels with branching `if (CLOUD)` statements. The second is that openapi _does_ make a distinction between cloud and OSS types, and we have to account for that with our lack of architectural distinction

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
hoorayimhelping: i'm generally leery of introducing one off patterns, even if they're better because of situations like this:

hoorayimhelping (approved):
LGTM. I addressed your previous comments but I didn't tag you so you may not have gotten alerted on them.

PR #2570, opened 2021-09-09
fix: pass time zone information to giraffe table graph

### LINKED ISSUE #2569: Giraffe Graphs are not respecting the TimeZone setting
When the `useGiraffeGraphs` flag is on, we can see that the table graph does not respect the timezone for the time columns. 

image

We need to pass TimeZone information to the giraffe layer from the UI.

ChitlangeSahas:
Closes #2569

Passes timezone information to the giraffe table graph.

Screen shot: 
Screen Shot 2021-09-09 at 9 33 35 AM


hoorayimhelping (approved):
outstanding!

PR #2585, opened 2021-09-10
fix: better handling for empty states in template mode

### LINKED ISSUE #2555: Notebooks: Fix error message on simple table prior to data selection
When I enter a template that includes a simple table visualization panel and no metrics selected, I get this error message: 

This message is confusing. Instead it should display "This cell will display results from the previous cell" until data has been inputed.

joelfried:
Closes #2555 

Screenshots of how notebooks look when loaded from templates without changes to the panels:

Screen Shot 2021-09-13 at 11 48 21 AM
Screen Shot 2021-09-13 at 11 23 04 AM
Screen Shot 2021-09-13 at 11 22 42 AM

### DIFF
--- a/src/flows/pipes/RawFluxEditor/index.ts
+++ b/src/flows/pipes/RawFluxEditor/index.ts
@@ -1,5 +1,4 @@
 import {parse, format_from_js_file} from '@influxdata/flux'
-import {find} from 'src/flows/context/query'
 import View from './view'
 import './style.scss'
 
@@ -33,11 +32,9 @@ export default register => {
         const ast = parse(
           data.queries[data.activeQuery].text.replace(PREVIOUS_REGEXP, query)
         )
-        find(ast, node => !!Object.keys(node.comments || {}).length).forEach(
-          node => {
-            delete node.comments
-          }
-        )
+        if (!ast.body.length) {
+          return ''
+        }
         return format_from_js_file(ast)
       } catch {
         return
--- a/src/flows/pipes/RawFluxEditor/index.ts
+++ b/src/flows/pipes/RawFluxEditor/index.ts
@@ -1,4 +1,5 @@
 import {parse, format_from_js_file} from '@influxdata/flux'
+import {find} from 'src/flows/context/query'
 import View from './view'
 import './style.scss'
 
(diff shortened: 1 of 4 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2585)

genehynson:
@drdelambre thoughts on this one? Just want to make sure this change doesn't have any other side effects that you might know of

drdelambre (changes requested):
this is all in the wrong place. have the visualization panel return nothing for it's visual function if there's no query and this should all just work

drdelambre:
make sure query exists here: https://github.com/influxdata/ui/blob/master/src/flows/pipes/Visualization/index.ts#L18
and make sure the `_build` function is returning something

src/flows/pipes/RawFluxEditor/index.ts
```diff
@@ -33,11 +32,9 @@ export default register => {
...
           data.queries[data.activeQuery].text.replace(PREVIOUS_REGEXP, query)
         )
-        find(ast, node => !!Object.keys(node.comments || {}).length).forEach(
-          node => {
-            delete node.comments
-          }
-        )
+        if (!ast.body.length) {
+          return ''
+        }
```
genehynson, line 37: a change must have been made in the wasm because now `parse` handles comments differently. `parse` removes them automatically and the property `.comments` no longer exists.

genehynson:
Thanks for the feedback @drdelambre ! I refactored the implementation - let me know what you think

src/flows/pipes/RawFluxEditor/index.ts
drdelambre: I don't think the distinction matters for this pr, but in case we have to do something with comments or query comparison later, are you saying that a valid query, with a comment and some flux, when passed through parse, then through format, will have the comments removed?

drdelambre (approved):
This will work and fixes the problem, but I don't think the reasoning is correct

src/flows/pipes/RawFluxEditor/index.ts
genehynson: after more testing, it looks like I was wrong. We still do need to parse comments. Reverted this change

PR #2591, opened 2021-09-13
chore: add convert to flux to vis dropdown

### LINKED ISSUE #2473: notebooks: Visualization panel should have a "convert to flux" option
~The "export to client library" option only makes sense for the panels that represent flux queries.  It shouldn't be an option on the visualization panel. I would only include it in the panels that have a "Convert to |> flux" option.~

We should add a "Convert to |> flux" option for viz panels.

ShmuelLotman:
Closes #2473

Adds "convert to flux" to visualization panel dropdown

### DIFF
--- a/src/flows/components/Sidebar.tsx
+++ b/src/flows/components/Sidebar.tsx
@@ -182,7 +182,10 @@ const Sidebar: FC = () => {
               return true
             }
 
-            if (!/^(inputs|transform)$/.test(PIPE_DEFINITIONS[type]?.family)) {
+            if (
+              !/^(inputs|transform)$/.test(PIPE_DEFINITIONS[type]?.family) &&
+              flow.data.byID[id].type !== 'visualization'
+            ) {
               return true
             }
 
--- a/src/flows/components/Sidebar.tsx
+++ b/src/flows/components/Sidebar.tsx
@@ -184,7 +184,7 @@ const Sidebar: FC = () => {
 
             if (
               !/^(inputs|transform)$/.test(PIPE_DEFINITIONS[type]?.family) &&
-              flow.data.byID[id].type !== 'visualization'
+              type !== 'visualization'
             ) {
               return true
             }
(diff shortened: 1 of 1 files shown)


src/flows/components/Sidebar.tsx
```diff
@@ -182,7 +182,10 @@ const Sidebar: FC = () => {
               return true
             }
 
-            if (!/^(inputs|transform)$/.test(PIPE_DEFINITIONS[type]?.family)) {
+            if (
+              !/^(inputs|transform)$/.test(PIPE_DEFINITIONS[type]?.family) &&
+              flow.data.byID[id].type !== 'visualization'
```
genehynson, line 187: `type` is already defined above, right? We can change this to `type !== 'visualization'` which improves readability

genehynson, line 187: also should this be `type === 'visualization'`? Do you have a gif of this working?

genehynson (changes requested):
I tested these changes out and I don't think they actually build the query correctly. The visualization panels have the `.visual` query that would need to be appended to the `.source` query. Currently, only the `.source` query is being shown in the resulting flux script

Example: 
image


I believe it should be 
```
buckets()
|> limit(n: 100)
```

src/flows/components/Sidebar.tsx
ShmuelLotman: As commented below, it works, i guess tho we need to put in the .visual in there for it to fully function, i missed that it wasn't there when it converted to flux

ShmuelLotman: Done

genehynson (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2591)

genehynson (approved):
works great now! Thanks!

PR #2596, opened 2021-09-13
fix: explicitly set width for pagination controls

### LINKED ISSUE #2594: Task Pagination: Paging calculation is off
### Issue
Paging through the list results in weird behavior

hoorayimhelping:
Closes #2594

When the Pagination Nav doesn't have a width, it leads to wonky pagination calculations. The fix is to set its width.

https://user-images.githubusercontent.com/146112/133125180-bc58c817-a10a-460b-9327-92fe426582dd.mov


TCL735 (approved):

daviddkkim (approved):
LGTM. 
The component itself could be more elegant but this should do for now if it works!

PR #2604, opened 2021-09-13
feat(notebooks): expressions sidebar

### LINKED ISSUE #2285: Notebooks: I can choose from a list of expressions to add to my alert message
We would like to make it easier and more intuitive for users to edit an alert message within a notebook by providing a list of expressions that they can inject. 

### Acceptance Criteria

- When a user clicks on the expressions button it opens up a side panel on the right (we can use the same side panel format as the functions side panel)
- The expressions list includes a list of expressions available for that dataset
- A user can inject an expression and it will appear where their cursor is in the alert message box
- A user can close the expressions panel either by clicking the "x" button or clicking the expressions button again

Design: @juliajames12 

Design file: https://www.figma.com/file/a7qLfWdlUDrhab9aWIxE65/Create-an-Alert-from-Notebook?node-id=664%3A9052

genehynson:
Closes #2285

Dynamically populate list of expressions the user can insert into an alert message based on the results of their query. e2e test included.

image

image

https://user-images.githubusercontent.com/6411855/133347694-f1db1a6b-8249-491b-afab-758bd295b894.mov

### DIFF
--- a/src/flows/pipes/Notification/Expression.tsx
+++ b/src/flows/pipes/Notification/Expression.tsx
@@ -19,29 +19,25 @@ const ToolbarExpression: FC<Props> = ({
   onClickFunction,
   testID,
 }) => {
-  const expressionRef = createRef<HTMLDListElement>()
   const handleClickFunction = () => {
     onClickFunction(expression)
   }
 
   return (
-    <>
-      <dd
-        ref={expressionRef}
-        data-testid={testID}
-        className="flux-toolbar--list-item flux-toolbar--function"
-      >
-        <code>{expression}</code>
-        <Button
-          testID={`flux--${testID}--inject`}
-          text="Inject"
-          onClick={handleClickFunction}
-          size={ComponentSize.ExtraSmall}
-          className="flux-toolbar--injector"
-          color={ComponentColor.Primary}
-        />
-      </dd>
-    </>
+    <dd
+      data-testid={testID}
+      className="flux-toolbar--list-item flux-toolbar--function"
+    >
+      <code>{expression}</code>
+      <Button
+        testID={`flux--${testID}--inject`}
+        text="Inject"
+        onClick={handleClickFunction}
+        size={ComponentSize.ExtraSmall}
+        className="flux-toolbar--injector"
+        color={ComponentColor.Primary}
+      />
+    </dd>
   )
 }
 
--- a/src/flows/pipes/Notification/Expression.tsx
+++ b/src/flows/pipes/Notification/Expression.tsx
@@ -1,0 +1,50 @@
+// Libraries
+import React, {FC, createRef} from 'react'
+
+// Component
+import {Button, ComponentSize, ComponentColor} from '@influxdata/clockface'
+
+interface Props {
+  expression: string
+  onClickFunction: (expression: string) => void
+  testID: string
+}
+
+const defaultProps = {
+  testID: 'alert-expression',
+}
+
+const ToolbarExpression: FC<Props> = ({
+  expression,
+  onClickFunction,
+  testID,
+}) => {
+  const expressionRef = createRef<HTMLDListElement>()
+  const handleClickFunction = () => {
+    onClickFunction(expression)
+  }
+
+  return (
+    <>
+      <dd
+        ref={expressionRef}
+        data-testid={testID}
+        className="flux-toolbar--list-item flux-toolbar--function"
+      >
+        <code>{expression}</code>
+        <Button
+          testID={`flux--${testID}--inject`}
+          text="Inject"
+          onClick={handleClickFunction}
+          size={ComponentSize.ExtraSmall}
+          className="flux-toolbar--injector"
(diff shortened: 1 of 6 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2604)

src/flows/pipes/Notification/Expressions.tsx
```diff
@@ -0,0 +1,220 @@
...
+        schema[measurements[ni]].measurements = filteredMeasurements
+      }
+
+      const filteredSystems = [
+        '_check_id',
+        '_check_name',
+        '_level',
+        '_source_measurement',
+        '_type',
+      ].filter(s => s.toLowerCase().includes(search.toLowerCase()))
```
genehynson, line 92: Russ didn't know of a way to dynamically populate these. So that's why they are hard coded here

src/flows/pipes/Notification/Expressions.tsx
```diff
@@ -0,0 +1,220 @@
...
+      const filteredTags = columns.filter(
+        c =>
+          !!out.columns[c].data[ni] &&
+          c.toLowerCase().includes(search.toLowerCase())
+      )
+      if (filteredTags.length) {
+        schema[measurements[ni]].tags = filteredTags
+      }
+
+      const otherColumns = ['_start', '_stop', '_measurement']
```
genehynson, line 106: This is the list of columns that Russ wanted to include to start. We might change the list in the future.

src/flows/pipes/Notification/Expression.tsx
```diff
@@ -0,0 +1,50 @@
...
+}) => {
+  const expressionRef = createRef<HTMLDListElement>()
+  const handleClickFunction = () => {
+    onClickFunction(expression)
+  }
+
+  return (
+    <>
+      <dd
+        ref={expressionRef}
```
drdelambre, line 30: what is this for?

genehynson: lol no idea. copy pasta. Seems to work fine without it so I'll throw it away

src/flows/pipes/Notification/Expressions.tsx
```diff
@@ -0,0 +1,195 @@
...
+    })
+    const schema = {
+      measurements: new Set<string>(),
+      fields: new Set<string>(),
+      tags: new Set<string>(),
+      columns: [],
+      system: [],
+    }
+
+    for (ni = 0; ni < len; ni++) {
```
drdelambre, line 66: this might be faster... been a bit since i perf'd sets like this
```
new Set(measurements.filter(m => m.toLowerCase().includes(search.toLowerCase())))
new Set(fields.filter(f => f.toLowerCase().includes(search.toLowerCase())))
... maybe tags? ...
```

src/flows/pipes/Notification/Expressions.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const [search, setSearch] = useState('')
+  const {flow, updateData} = useContext(FlowContext)
+  const updateSearch = useCallback(
+    e => {
+      setSearch(e.target.value)
+    },
+    [search, setSearch]
+  )
+
+  const parsedResultToSchema = (parsed: FromFluxResult): SchemaExpressions => {
```
drdelambre, line 33: seems like we can also move some of these functions outside of the component to reduce the overhead of re-declaring them every render function

drdelambre:
crazy question. do we want to use monaco instead of the text area so that we can normalize the injection pattern?

genehynson:
> crazy question. do we want to use monaco instead of the text area so that we can normalize the injection pattern?

That is a crazy question. Guess the downside would be a performance hit to the initial load of the panel...and the lingering "monaco editor is slow" stuff right?

drdelambre:
totally gotta balance that perf hit of the lazy load on the initial load, but for the record the monaco being slow is actually the LSP being slow last i checked (only affecting flux queries)

genehynson:
> totally gotta balance that perf hit of the lazy load on the initial load, but for the record the monaco being slow is actually the LSP being slow last i checked (only affecting flux queries)

ahhh gotcha. Well then I can certainly swap it out and give it a shot

src/flows/pipes/Notification/Expressions.tsx
genehynson: nice yeah I was able to refactor that function and clean it up more with this suggestion

genehynson:
alright swapped it out for the monaco editor and refactored to use the existing injection pattern @drdelambre

russorat:
this looks really nice!

drdelambre (approved):

PR #2610, opened 2021-09-14
feat: store pagination and filter state in url on change

hoorayimhelping:
Closes https://github.com/influxdata/idpe/issues/11749

Stores the current choice for page and sorting options in the url.
1. Allows a user to reload the page and see the same results they just saw
2. Allows a user to share their page with a teammate who can then see the same results

https://user-images.githubusercontent.com/146112/133320621-000137b3-1d07-4cf1-a144-d2eaa6d4b5e9.mov

### DIFF
--- a/src/tasks/pagination/TasksPage.tsx
+++ b/src/tasks/pagination/TasksPage.tsx
@@ -86,7 +86,7 @@ class TasksPage extends PureComponent<Props, State> {
     this.props.getTasks(-1) // -1 means fetch all tasks with no limit
     this.props.getLabels()
 
-    let sortType = this.state.sortType
+    let sortType: SortTypes = this.state.sortType
     const params = new URLSearchParams(window.location.search)
 
     let sortKey: TaskSortKey = 'name'
@@ -100,7 +100,7 @@ class TasksPage extends PureComponent<Props, State> {
       sortType = SortTypes.String
     }
 
-    let sortDirection = this.state.sortDirection
+    let sortDirection: Sort = this.state.sortDirection
     if (params.get('sortDirection') === Sort.Ascending) {
       sortDirection = Sort.Ascending
     } else if (params.get('sortDirection') === Sort.Descending) {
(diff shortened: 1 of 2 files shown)


src/tasks/pagination/TasksPage.tsx
```diff
@@ -177,6 +200,11 @@ class TasksPage extends PureComponent<Props, State> {
     sortDirection: Sort,
     sortType: SortTypes
   ) => {
+    const url = new URL(location.href)
+    url.searchParams.set('sortKey', sortKey)
+    url.searchParams.set('sortDirection', sortDirection)
+    history.replaceState(null, '', url.toString())
```
ChitlangeSahas, line 206: First I thought passing `null` was going to overwrite the browser history, making the back button useless. But then I read some more on the docs of `replaceState` and this makes sense. 

So we want to be able to go back to a previous page in the pagination, through browser history, is that a correct interpretation?

ChitlangeSahas (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2610)

src/tasks/pagination/TasksPage.tsx
hoorayimhelping: >So we want to be able to go back to a previous page in the pagination, through browser history, is that a correct interpretation?

@ChitlangeSahas 

Great question, love where your head's at 😄 What this lets us do is change the url without changing history. So as you paginate through results, it'll update the url with `?page=$x` but not put those in the history stack, so you don't have to hit the back button to go through your pagination history.

If you click Boards in the nav, then click Tasks, then paginate through 5 pages, then hit back, you'll be back at the dashboards page.

ChitlangeSahas: Ahhh! Total opposite of what I thought it was doing, good to know! Thank you for explaining!

PR #2614, opened 2021-09-14
feat: Edit overlay permissions

### LINKED ISSUE #1966: Tokens:  Edit Overlay should display the permissions for the token
As a user who is editing a token,
I would like the ability to see the permissions for the token.
The permissions for the token should be disabled and not selectable.

blegesse-git:
Closes #1966 

Inside the Edit Token Overlay, user can now see permissions for a token. 
The permissions are disabled and not selectable. 

Screen Shot 2021-09-14 at 4 37 10 PM
Screen Shot 2021-09-14 at 4 37 30 PM

### DIFF
--- a/src/authorizations/components/redesigned/EditResourceAccordion.tsx
+++ b/src/authorizations/components/redesigned/EditResourceAccordion.tsx
@@ -1,0 +1,77 @@
+// Libraries
+import {Component} from 'react'
+
+interface Props {
+  permissions: any
+}
+
+export class ResourceAccordion extends Component<Props, {}> {
+  public render() {
+    // const {resources} = this.props // []
+    const {permissions} = this.props
+    /* {
+      "telegrafs": {
+      id: ""
+      orgID:
+      subLevelPermissions: {
+        id: {
+          id: ""
+          permissions: {read: false, write: false}
+        }
+      }
+      },
+      annotations: {read: false, write: false}
+    }
+    */
+
+    console.log(Object.keys(permissions))
+
+    if (!permissions) {
+      return null
+    }
+    // return resources.map(resource => {
+    //   const resourceName = resource.charAt(0).toUpperCase() + resource.slice(1)
+    //   // console.log("resource", resource) resorce = buckets or telegrafs
+    //   console.log(permissions[resource])
+    //   return (
+    //     <Accordion key={resource}>
+    //       <ResourceAccordionHeader
+    //         resourceName={resourceName} // Buckets
+    //         permissions={permissions[resource]} // props.bucketPermissions
+    //         onToggleAll={this.handleToggleAll}
+    //       />
+    //       {resourceName === 'Telegrafs' ||
+    //         (resourceName === 'Buckets' && (
+    //           <GetResources resources={[ResourceType[resourceName]]}>
+    //             {!isEmpty(permissions[resource].sublevelPermissions) &&
+    //               this.getAccordionBody(resourceName, resource)}
+    //           </GetResources>
+    //         ))}
+    //     </Accordion>
+    //   )
+    // })
+  }
+
+  // getAccordionBody = (resourceName, resource) => {
+  //   const {permissions} = this.props
+  //   if (resourceName === 'Telegrafs') {
+  //     return (
+  //       <ResourceAccordionBody
+  //         resourceName={resource}
+  //         permissions={permissions[resource].sublevelPermissions}
+  //         onToggle={this.handleIndividualToggle}
+  //         title="Individual Telegraf Configuration Names"
+  //       />
+  //     )
+  //   } else if (resourceName === 'Buckets') {
+  //     return (
+  //       <ResourceAccordionBody
+  //         resourceName={resource}
+  //         permissions={permissions[resource].sublevelPermissions}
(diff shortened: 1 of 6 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2614)

src/authorizations/utils/permissions.ts
```diff
@@ -135,3 +135,72 @@ export enum BucketTab {
   AllBuckets = 'All Buckets',
   Scoped = 'Scoped',
 }
+
+export const formatPermissionsObj = permissions => {
+  const newPerms = permissions.reduce((acc, {action, resource}) => {
+    const {type, id, orgID, name} = resource
+    let p
```
hoorayimhelping, line 142:
```ts
let p
```

😆😆😆😆

giffing aside, I'm not a very big fan of single letter variable names because they greatly increase the ambiguity of the code. Now it's pretty obvious that in the context of this code review `p` represents permissions, but the problem is that we're looking at this in the context of a code review about permissions. Looking at this code in isolation during a normal operation, it may not be clear what `p` is. Going from the other direction, if we wanted to search the entire codebase for `permission` we may miss this variable, or its usage in the context of a page of search results may not be very clear.

Also, we could remove an `else` case many lines away from the variable being created, and increase the understanding of the code if we did something like

```ts
let permissions = {
  read: action === 'read',
  write: action === 'write',
}

File: src/authorizations/utils/permissions.ts [hunk 2]
```diff
@@ -135,3 +135,72 @@ export enum BucketTab {
...
+              read: action === 'read',
+              write: action === 'write',
+            },
+          }
+        }
+      } else {
+        p[action] = true
+      }
+    } else {
+      if (id && (type === 'buckets' || type === 'telegrafs')) {
```
hoorayimhelping, line 164: should this be

```ts
} else if (id && (type === 'buckets'....
```
?

src/authorizations/utils/permissions.ts
```diff
@@ -135,3 +135,72 @@ export enum BucketTab {
...
+          read: action === 'read',
+          write: action === 'write',
+        }
+      }
+    }
+
+    return {...acc, [type]: p}
+  }, {})
+
+  Object.keys(newPerms).map(resource => {
```
hoorayimhelping, line 191:
```suggestion
  Object.keys(newPerms).forEach(resource => {
```

`map` and `forEach` both loop over every item of an array and apply the specified function to the array. The difference is that `map` produces an array of the same size as the input array and expects the function to return the new item of the array. `forEach` simply iterates over the array and does some operation. `forEach` is designed to be used with side effects like you're doing here.

hoorayimhelping (changes requested):
Nice work on these complex data structure issue. Have some suggestions

src/authorizations/components/redesigned/EditResourceAccordion.tsx
```diff
@@ -0,0 +1,61 @@
+// Libraries
+import React, {Component} from 'react'
+import {isEmpty} from 'lodash'
+
+import {Accordion} from '@influxdata/clockface'
+
+import {ResourceAccordionHeader} from './ResourceAccordionHeader'
+import {ResourceAccordionBody} from './ResourceAccordionBody'
```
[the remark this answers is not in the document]
blegesse-git: that's a really good point. done and done!

src/authorizations/components/redesigned/EditResourceAccordion.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export class EditResourceAccordion extends Component<Props, {}> {
+  public render() {
+    const {permissions} = this.props
+    if (!permissions) {
+      return null
+    }
+
+    return Object.keys(permissions).map(key => {
+      const resourceName = key.charAt(0).toUpperCase() + key.slice(1)
```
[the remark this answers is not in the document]
blegesse-git: ahhh! what a cool shortcut.

src/authorizations/utils/permissions.ts
blegesse-git: you know what, i deserved that meme LOL. we even went back to refactor the code but that one flew over our head. yup! `let p` is really dumb

src/authorizations/utils/permissions.ts
blegesse-git: there are a lot of if else statement here so i get the confusion. but this one should remain `else` because this else statement is there in case a specific resource is not already in the accumulator. in the case where we are iterating over a new resource type, then we are checking if that object has an id. if so, we are structuring the new object differently. if it doesn't have an id, (line180 takes care of that) then we are creating an object with only 2 properties. 

first if statement --> if (acc already has a (bucket) property) .... do this. else { if (current object has an id AND type is bucket or telegraf) {do this ....} else { do this}}

hoorayimhelping (approved):

PR #2615, opened 2021-09-14
chore: if provsion fails, flush and retry

genehynson:
We've had some weirdness where `/flush` requests fail and then cause subsequent `/provision` requests to fail. This change retries the /flush after a failed /provision

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -573,7 +573,7 @@ export const createToken = (
   })
 }
 
-export const setupUser = (): Cypress.Chainable<Cypress.Response<any>> => {
+export const setupUser = (): Cypress.Chainable<any> => {
   const defaultUser = Cypress.env('defaultUser')
   const userParam = defaultUser ? `?user=${defaultUser}` : ''
   return cy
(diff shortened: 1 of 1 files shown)


cypress/support/commands.ts
```diff
@@ -582,23 +582,31 @@ export const setupUser = (): Cypress.Chainable<Cypress.Response<any>> => {
...
     .then(response => {
-      expect(response.status).to.eq(200)
-      Cypress.env('defaultUser', response.body.user.name)
-      return response
+      if (response.status === 200) {
+        Cypress.env('defaultUser', response.body.user.name)
+        return response
+      } else {
+        // if for some reason the user wasn't flushed, do it now
+        return cy.flush().then(_ => cy.setupUser().then(res => res))
```
drdelambre, line 590: is this how recursion works in cypress? the "its always a promise except when..." logic gets me turned around sometimes

drdelambre (approved):

cypress/support/commands.ts
genehynson: I think as long as we're returning a `Cypress.Chainable` object we should be ok

PR #2632, opened 2021-09-16
fix: remove duplicated input configuration from the backend when telegrafUiRefresh is active

### LINKED ISSUE #2625: Telegraf Refresh: some input plugins' configuration text show up twice
The new feature where Telegraf configurations may be created from a plugin's page has a redundancy problem. Some plugins have their configurations show up twice. This is due to both the backend and the frontend inserting the configurations for the same plugin.

Here is the list of plugins where this happens:
```
"cpu"
"disk"
"diskio"
"docker"
"file"
"kernel"
"kubernetes"
"logparser"
"mem"
"net_response"
"net"
"nginx"
"processes"
"procstat"
"prometheus"
"redis"
"swap"
"syslog"
"system"
"tail"
```

Steps to reproduce:
With the feature flag `telegrafUiRefresh` turned on
1. Navigate: Data > Sources
2. Scroll down and click on any input plugin
3. Click on Use this plugin
4. Click on Create a new configuration
5. Follow the overlays and complete the creation process
6. Navigate: Data > Telegraf
7. Open the configuration created from step 5
8. Scroll through the editor and notice that inputs for the plugin shows up twice

TCL735:
Closes #2625 

Certain plugins no longer have a duplicated `inputs` in the configuration.  See the list in #2625 

Here is an example with `inputs.kubernetes`

https://user-images.githubusercontent.com/10736577/133538891-78e6a22c-57ef-4a39-a106-1d90e26221f2.mp4

### DIFF
--- a/src/writeData/components/PluginCreateConfigurationFooter.tsx
+++ b/src/writeData/components/PluginCreateConfigurationFooter.tsx
@@ -1,6 +1,7 @@
 // Libraries
 import React, {FC, useEffect, useMemo} from 'react'
 import {connect, ConnectedProps} from 'react-redux'
+import {useParams} from 'react-router'
 
 // Components
 import {
@@ -21,6 +22,10 @@ import {PluginCreateConfigurationStepProps} from 'src/writeData/components/Plugi
 // Selectors
 import {getDataLoaders} from 'src/dataLoaders/selectors'
 import {getAll} from 'src/resources/selectors'
+
+type ParamsType = {
+  [param: string]: string
+}
 
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = PluginCreateConfigurationStepProps & ReduxProps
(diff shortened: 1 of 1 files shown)


src/writeData/components/PluginCreateConfigurationFooter.tsx
```diff
@@ -43,10 +48,24 @@ const PluginCreateConfigurationFooterComponent: FC<Props> = props => {
...
       const {config} = telegrafConfig
-      onUpdateTelegraf({...telegrafConfig, config: `${config}${pluginConfig}`})
+      const position =
+        typeof config === 'string'
+          ? config.indexOf(`[[inputs.${contentID}]]`)
+          : -1
+      const updatedConfig =
+        position === -1
+          ? `${config}${pluginConfig}`
+          : `${config.substring(0, position)}${pluginConfig}`
```
hoorayimhelping, line 63: i think these two expressions should be put in the same `if` statement. if I'm reading this right, if `config === 'string'`, then `updatedConfig` gets set to
```ts
`${config.substring(0, position)}${pluginConfig}`
```

otherwise it's set to

```ts
`${config.substring(0, position)}${pluginConfig}`
```

and `position` is just a transitory variable. I think having these in the same `if` statement would make the intent a little more clear.

if you don't decide to do this, i think at the very least the ternary expressions should be parallel-constructed so that the cases line up.

in the first ternary, if the expression `config === 'string'` is true, `position` gets set to the `indexOf` in the first case, or `-1` in the second case. in the second ternary, they're flipped so that if `-1` is used, it's the first case and the second case maps to getting the `indexOf`

hoorayimhelping (approved):
looks good to me, have a suggestion on making the conditional expression a little bit more readable

src/writeData/components/PluginCreateConfigurationFooter.tsx
TCL735: I see what you mean about them being confusing. Yes, the second ternary, for updatedConfig, should be constructed differently. I will use `position >= 0` as the comparator instead.

`position` is not transitory. It is needed, because even if `typeof config === 'string` it is possible for `position` to be -1, when the `config.indexOf` call cannot find a match.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2632)

PR #2633, opened 2021-09-16
feat(docsearch): update facetFilters for docsearch

sanderson:
This updates the Algolia docsearch facetFilters to match the new search tag pattern used in the docs. Search will temporarily stop working between the time we deploy the changes to the docs site and reindex our docs and when this change goes live.

### DIFF
--- a/src/shared/search/DocSearch.tsx
+++ b/src/shared/search/DocSearch.tsx
@@ -40,7 +40,7 @@ const DocSearch: FC<DocSearchProps> = ({type}) => {
 
   const facetFilters = useMemo(
     () => ({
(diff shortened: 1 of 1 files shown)


genehynson (commented):
to avoid search going down for that period of time, should we put these changes behind a feature flag so we can toggle it in configcat?

sanderson:
@genehynson Actually... I just realized it won't break. The old facet filters will still work. So we're probably good to just let this work it's way through CI. We still need to apply this change, I just realized that it shouldn't stop returning results. The facets we're currently using will still be in place. We probably don't need to a add a feature flag for this small of a change.

genehynson (approved):

PR #2645, opened 2021-09-17
chore: fix memory leak

### LINKED ISSUE #2641: Pinned Items: React memory leaks
### Background
With the pinned Items flagged turned on, a couple pages have react memory leaks. I've seen this affect Dashboards and Tasks, and it might affects Flows as well. Basically anywhere `getPinnedItems` is called where it sets component state.

Here's a screenshot from the dashboard page

and here is where the line points to:
Screen Shot 2021-09-16 at 11 27 08 AM

### Steps to reproduce
1. turn on pinned items flag (`influx.toggle('pinnedItems')`)
1. make sure you have a dashboard with cards on it
1. on the left nav, click Boards then immediately click a different navbar item. The goal here is to get the request that is launched on `componentDidMount` to fire but leave before it sets component state
1. observe the error

### Potential fix:
- When the component unmounts, set an instance property called something like `isUnmounting` to true. Before setting state, make sure `isUnmounting` is false. https://gist.github.com/hoorayimhelping/eb7274c5a26411c35ef9271c6f136e08

ShmuelLotman:
Closes #2641 

Fixes react memory leaks in taskslist and dashboardcards components

### DIFF
--- a/src/dashboards/components/dashboard_index/DashboardCards.tsx
+++ b/src/dashboards/components/dashboard_index/DashboardCards.tsx
@@ -40,6 +40,7 @@ interface OwnProps {
 
 class DashboardCards extends PureComponent<OwnProps & StateProps> {
   private _observer
+  private _isMounted = true
   private _spinner
 
   private memGetSortedResources = memoizeOne<typeof getSortedResources>(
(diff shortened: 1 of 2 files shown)


joelfried (approved):
LGTM

PR #2415, opened 2021-08-18
feat: simple table

ChitlangeSahas:
Closes #1726

Brings in Simple table from `giraffe`.  The changes are behind the `useGiraffeGraphs` feature flag.

https://user-images.githubusercontent.com/18511823/130877218-a1274ca2-e165-4f85-8174-73534f1f6d24.mov

PR #2664, opened 2021-09-21
chore: update open api sha

genehynson:
Trying to update the open api sha and had to fix a few things so tests pass

### DIFF
--- a/src/annotations/api/index.ts
+++ b/src/annotations/api/index.ts
@@ -12,6 +12,8 @@ import {
   postAnnotation,
   putAnnotation,
   AnnotationEvent,
+  Error as ResponseError,
+  AnnotationList,
 } from 'src/client/annotationdRoutes'
 
 // Utils
@@ -63,10 +65,10 @@ export const getAnnotations = async (
   }
   const res = await getAnnotationsApi(params)
   if (res.status >= 300) {
-    throw new Error(res.data?.message)
+    throw new Error((res.data as ResponseError).message)
   }
 
-  return res.data.map((retrievedAnnotation: AnnotationResponse) => ({
+  return (res.data as AnnotationList).map((retrievedAnnotation: any) => ({
     stream: retrievedAnnotation.stream,
     annotations: retrievedAnnotation.annotations,
   }))
--- a/src/timeMachine/reducers/index.ts
+++ b/src/timeMachine/reducers/index.ts
@@ -694,7 +694,10 @@ export const timeMachineReducer = (
     }
 
     case 'SET_BACKGROUND_THRESHOLD_COLORING': {
-      const viewColors = state.view.properties.colors as Color[]
+      let viewColors = []
+      if ('colors' in state.view.properties) {
+        viewColors = state.view.properties.colors as Color[]
+      }
 
       const colors = viewColors.map(color => {
         if (color.type !== 'scale') {
@@ -711,7 +714,10 @@ export const timeMachineReducer = (
     }
 
     case 'SET_TEXT_THRESHOLD_COLORING': {
-      const viewColors = state.view.properties.colors as Color[]
+      let viewColors = []
+      if ('colors' in state.view.properties) {
+        viewColors = state.view.properties.colors as Color[]
+      }
 
       const colors = viewColors.map(color => {
         if (color.type !== 'scale') {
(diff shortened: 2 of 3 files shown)


src/timeMachine/reducers/index.ts
```diff
@@ -694,7 +694,10 @@ export const timeMachineReducer = (
     }
 
     case 'SET_BACKGROUND_THRESHOLD_COLORING': {
-      const viewColors = state.view.properties.colors as Color[]
+      let viewColors = []
+      if ('colors' in state.view.properties) {
+        viewColors = state.view.properties.colors as Color[]
+      }
```
genehynson, line 700: stole this from: https://github.com/influxdata/ui/pull/2415/files#r696188499

src/annotations/api/index.ts
```diff
@@ -63,10 +65,10 @@ export const getAnnotations = async (
   }
   const res = await getAnnotationsApi(params)
   if (res.status >= 300) {
-    throw new Error(res.data?.message)
+    throw new Error((res.data as ResponseError).message)
   }
 
-  return res.data.map((retrievedAnnotation: AnnotationResponse) => ({
+  return (res.data as AnnotationList).map((retrievedAnnotation: any) => ({
```
genehynson, line 71: Not sure if changing the `AnnotationResponse` to `any` was the way to go here. The IDE says that the items in `AnnotationList` are not of type `AnnotationResponse` because `stream` and `annotations` properties are optional. @hoorayimhelping thoughts on this?

genehynson: Also the `as AnnotationList` casting is required because the IDE says the type is actually `AnnotationList | ResponseError`

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2664)

drdelambre (approved):

src/annotations/api/index.ts
glinton: yes, openapi#154 caused issues when I tested them in #1811 I understood someone was going to update the tests for the actual response coming from annotations and we could merge the openapi pr. when 1811 got closed, i went ahead and merged 154 understanding that some changes to the ui tests would need to be made when y'all pulled in the latest spec.

genehynson: @hoorayimhelping we good to move forward with this change? Any changes you'd like to propose?

hoorayimhelping: @genehynson I'd prefer we didn't type a response that used to have a type as `any`. I'm still not quite following these changes @glinton - it seems like we've actually lost resolution here instead of gained it.

glinton: Sounds like that might be related to https://github.com/influxdata/oats/issues/24, where `any` is used on a referenced type. The changes to openapi define the actual [response structure](https://github.com/influxdata/openapi/pull/154/files#diff-8c843eba1d4939da04c779f4c3de77308fa51dc7dc48176fb0be042c4d9a28a9R342-R352). Because the openapi definition [references the type](https://github.com/influxdata/openapi/pull/154/files#diff-8c843eba1d4939da04c779f4c3de77308fa51dc7dc48176fb0be042c4d9a28a9R352) for the array, oats must be marking it as `any`.

Any news on whether oats will get some attention or if y'all will be switching to a better generator?

glinton: maybe explicitly marking them as `required` in swagger would help, but that seems a bit pedantic. wouldn't be the worst thing, but it feels odd to mark something as "required" where it's a response structure rather than request :thinking:

genehynson: Tried marking them "required" in the swagger on this branch and it didn't change the generated code. https://github.com/influxdata/openapi/tree/genehynson/trying-something-annotations-2

glinton: Just noticed that as well

genehynson: @hoorayimhelping would it be ok if I merged this in to unblock other changes? I could open an issue to follow up on this specific typing problem

hoorayimhelping: Yeah, go for it and merge @genehynson. Don't let this discussion block you

PR #2665, opened 2021-09-21
feat: move api changes to use openapi

drdelambre:


src/flows/pipes/QueryBuilder/readOnly.tsx
```diff
@@ -21,7 +21,7 @@ const QueryBuilder: FC<PipeProp> = ({Context}) => {
   const {data} = useContext(PipeContext)
 
   return (
-    <Context>
+    <Context resizes>
```
genehynson, line 24: this wasn't part of my commit. Not sure why it's here, but it should be removed

genehynson:
added back to the original PR

PR #2677, opened 2021-09-21
feat: share template

drdelambre:
we should be able to clone a notebook by it's share id. this PR aims to do that. not expecting to drive traffic here until ephemeral notebooks is done and we have a clearer picture on how this might work


src/flows/templates/types/share.ts
```diff
@@ -0,0 +1,13 @@
+export default register =>
+  register({
+    type: 'share',
+    init: accessID =>
+      fetch(`/api/share/${accessID}`)
```
genehynson, line 5: this is very slick. Nicely done

src/flows/templates/types/share.ts
```diff
@@ -0,0 +1,13 @@
...
+        .then(res => res.json())
+        .then(res => {
+          delete res.id
```
genehynson, line 8: anything else we would need to delete? What about the orgID?

genehynson (approved):
took me a second to figure out how this works. For anyone else checking out this PR, go to https://[ns].a.influxcloud.dev.local/notebook/from/share/[accessID]

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2677)

genehynson:
@drdelambre maybe 3rd time's the charm? 😆

drdelambre:
Oh I sure hope so, debugging the prod assets is.... Not fun

drdelambre:
oh wow! just keep merging ™️

src/flows/templates/types/share.ts
asalem1: Don't we have a generated function we can use here?

PR #2679, opened 2021-09-21
chore: update the bug report template.

ChitlangeSahas:
Closes #

Being on bug duty twice I noticed two things:

* Quite a few bug reports were entirely just blank templates, I think it is because the current one is too noisy and the reporter thinks that the bug information is pre-populated.
* As a bug squasher, the report has too many things, unorganized, the greeting message need not be shown in the final report, there was no separation between the bug and the environment information. No hint to let the user know to add screenshots/visual information.

Screenshot

Screen Shot 2021-09-21 at 12 44 36 PM

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,7 @@ assignees: ''
 
 ---
 
-Thank you for reporting a bug in InfluxData UI.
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):

hoorayimhelping:
nice!

genehynson (approved):

PR #2693, opened 2021-09-22
fix: clicking on the markdown doesnt make it editable

### LINKED ISSUE #2674: Cant click on a link in a markdown panel
whenever i try to click on a link in a markdown panel, it switches into edit mode. even when the notebook is in presentation mode. super weird, it shouldn't do that.

also while were in there, the markdown panel shouldn't show up as editable if the notebook is in presentation mode at all, even if the panel was in edit mode when the notebook was switched into presentation mode

asalem1:
Closes #2674 

This PR updates the markdown panel so that clicking on it while the markdown pipe is in `preview` mode doesn't make it editable. It also introduces a new pattern where a flow that's in presentation will display the markdown as being in preview mode

markdown-pipe

### DIFF
--- a/src/flows/pipes/Markdown/view.tsx
+++ b/src/flows/pipes/Markdown/view.tsx
@@ -12,6 +12,7 @@ import MarkdownModeToggle from './MarkdownModeToggle'
 import {MarkdownRenderer} from 'src/shared/components/views/MarkdownRenderer'
 import {PipeContext} from 'src/flows/context/pipe'
 import {PipeProp} from 'src/types/flows'
+import {FlowContext} from 'src/flows/context/flow.current'
 
 const MarkdownMonacoEditor = lazy(() =>
   import('src/shared/components/MarkdownMonacoEditor')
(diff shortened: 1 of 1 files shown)


src/flows/pipes/Markdown/view.tsx
```diff
@@ -45,10 +43,10 @@ const MarkdownPanel: FC<PipeProp> = ({Context}) => {
     </Suspense>
   )
 
-  if (data.mode === 'preview') {
+  if (flow.readOnly || data.mode === 'preview') {
```
genehynson, line 46: good call!

genehynson (approved):

src/flows/pipes/Markdown/view.tsx
```diff
@@ -12,17 +12,15 @@ import MarkdownModeToggle from './MarkdownModeToggle'
...
+import {FlowContext} from 'src/flows/context/flow.current'
 
 const MarkdownMonacoEditor = lazy(() =>
   import('src/shared/components/MarkdownMonacoEditor')
 )
 
 const MarkdownPanel: FC<PipeProp> = ({Context}) => {
   const {data, update} = useContext(PipeContext)
-
-  const handlePreviewClick = (): void => {
```
drdelambre, line 23: I am fully ok with this functionality change

drdelambre (approved):

PR #2699, opened 2021-09-24
fix: use a separate component hierarchy to prevent unmounting during telegraf configuration creation

### LINKED ISSUE #2663: Telegraf Refresh: Create a Configuration does not add the selected input plugin
As a follow-up to https://github.com/influxdata/ui/issues/1556 and the work in https://github.com/influxdata/ui/pull/2582 there is a problem: the created configuration does not have the selected input plugin's text

**EDIT**: ensure that the feature flag is on. In the browser console: `influx.set('telegrafUiRefresh', true)`

**Steps to reproduce**:
- Navigate to Data > Telegraf
- Click on "+ Create Configuration"
- Complete the steps in the overlays to create a configuration
   - select a bucket
   - select a plugin
   - click continue
   - edit the name (optional)
   - edit the description (optional)
   - edit the configuration text (optional)
 - Click Save and Test
 - Click Finish
 - Open the Telegraf configuration that was just created by clicking its name in the list
 - Scroll through the editor that appears and check for **agent**, **outputs**, and **inputs**
 - Notice that **inputs** is missing

**Expected behavior**: A complete Telegraf configuration has **agent**, **outputs**, and **inputs**

TCL735:
Closes #2663 

Fixes the "Create Configuration" flow from Data > Telegraf. Previously, the selected input plugin would not be saved even though the configuration itself was saved. The update api call was not being made because the component responsible for making that call was unmounted before it could do so.

We fix this by
- creating a new component hierarchy that uses the new telegraf configuration creation wizard properly (does not unmount)
- making minor adjustments to the Footer to allow for the usage from a different flow
- restoring CollectorsStepSwitcher to an earlier version by removing unused code

https://user-images.githubusercontent.com/10736577/134601560-1665d2c9-94d2-438c-a7c1-a3ca66641df5.mp4

### DIFF
--- a/src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
+++ b/src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
@@ -7,16 +7,16 @@ import {withRouter, RouteComponentProps} from 'react-router-dom'
 // Components
 import {ErrorHandling} from 'src/shared/decorators/errors'
 import {Overlay} from '@influxdata/clockface'
+import PageSpinner from 'src/perf/components/PageSpinner'
 import {PluginCreateConfigurationFooter} from 'src/writeData/components/PluginCreateConfigurationFooter'
 
-const spinner = <div />
 const TelegrafUIRefreshStepSwitcher = Loadable({
   loader: () =>
     import(
       'src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshStepSwitcher'
     ),
   loading() {
-    return spinner
+    return <PageSpinner />
   },
 })
 
@@ -49,11 +49,15 @@ type Props = ReduxProps & RouteComponentProps<{orgID: string}>
 
 @ErrorHandling
 class TelegrafUIRefreshWizard extends PureComponent<Props> {
-  public state = {pluginConfig: '', isValidConfiguration: false}
+  public state = {
+    pluginConfig: '',
+    isValidConfiguration: false,
+    isVisible: true,
+  }
 
   public componentDidMount() {
     const {bucket, buckets} = this.props
-    if (!bucket && buckets && buckets.length) {
+    if (!bucket && Array.isArray(buckets) && buckets.length) {
       const {orgID, name, id} = buckets[0]
       this.props.onSetBucketInfo(orgID, name, id)
     }
@@ -69,7 +73,7 @@ class TelegrafUIRefreshWizard extends PureComponent<Props> {
     }
 
     return (
-      <Overlay visible={true}>
+      <Overlay visible={this.state.isVisible}>
         <Overlay.Container maxWidth={1200}>
           <Overlay.Header
             title="Create a Telegraf Configuration"
@@ -89,6 +93,7 @@ class TelegrafUIRefreshWizard extends PureComponent<Props> {
     const {onClearDataLoaders, onClearSteps} = this.props
     onClearDataLoaders()
     onClearSteps()
+    this.setState({isVisible: false})
     history.push(`/orgs/${org.id}/load-data/telegrafs`)
   }
 
--- a/src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
+++ b/src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
@@ -1,0 +1,178 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import Loadable from 'react-loadable'
+import {connect, ConnectedProps} from 'react-redux'
+import {withRouter, RouteComponentProps} from 'react-router-dom'
+
+// Components
+import {ErrorHandling} from 'src/shared/decorators/errors'
+import {Overlay} from '@influxdata/clockface'
+import {PluginCreateConfigurationFooter} from 'src/writeData/components/PluginCreateConfigurationFooter'
+
+const spinner = <div />
+const TelegrafUIRefreshStepSwitcher = Loadable({
+  loader: () =>
+    import(
+      'src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshStepSwitcher'
+    ),
+  loading() {
+    return spinner
+  },
+})
+
+// Actions
+import {notify as notifyAction} from 'src/shared/actions/notifications'
+import {
+  clearSteps,
+  decrementCurrentStepIndex,
+  incrementCurrentStepIndex,
+  setBucketInfo,
+  setCurrentStepIndex,
+  setSubstepIndex,
+} from 'src/dataLoaders/actions/steps'
+
+import {clearDataLoaders} from 'src/dataLoaders/actions/dataLoaders'
+
+// Types
+import {AppState, Bucket, ResourceType} from 'src/types'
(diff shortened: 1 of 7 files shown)


src/dataLoaders/components/collectorsWizard/select/SelectCollectorsStep2.tsx
```diff
@@ -73,33 +79,10 @@ export class SelectCollectorsStep extends PureComponent<Props, State> {
             onSelectBucket={this.handleSelectBucket}
           />
         </DapperScrollbars>
-        <OnboardingButtons
-          autoFocusNext={true}
-          nextButtonStatus={this.nextButtonStatus}
-          className="data-loading--button-container"
-        />
```
TCL735, line 80: Handled by the Footer in the Wizard.

src/dataLoaders/components/collectorsWizard/select/SelectCollectorsStep2.tsx
```diff
@@ -73,33 +79,10 @@ export class SelectCollectorsStep extends PureComponent<Props, State> {
...
-      return ComponentStatus.Disabled
-    }
-
-    if (!selectedBucketName) {
-      return ComponentStatus.Disabled
-    }
-
-    return ComponentStatus.Default
-  }
-
```
TCL735, line 102: Now changed to `ComponentDidUpdate`

src/writeData/components/PluginCreateConfigurationWizard.tsx
```diff
@@ -40,17 +40,17 @@ const PluginCreateConfigurationStepSwitcher = Loadable({
...
   onSetSubstepIndex: (currentStepIndex: number, subStepIndex: number) => void
-  substepIndex: number
   pluginConfig: string
-  setPluginConfig: (config: string) => void
-  isValidConfiguration: boolean
-  setIsValidConfiguration: (isValid: boolean) => void
   pluginConfigName: string
+  setIsValidConfiguration: (isValid: boolean) => void
+  setPluginConfig: (config: string) => void
+  substepIndex: number
```
TCL735, line 53: alphabetizing

src/dataLoaders/components/collectorsWizard/CollectorsStepSwitcher.tsx
```diff
@@ -3,42 +3,29 @@ import React, {PureComponent} from 'react'
...
-      configure = <PluginCreateConfigurationWizard />
-    } else {
-      selector = <SelectCollectorsStep {...stepProps} />
-      configure = <PluginConfigSwitcher />
-    }
 
     switch (stepProps.currentStepIndex) {
       case CollectorsStep.Select:
-        return selector
+        return <SelectCollectorsStep {...stepProps} />
```
hoorayimhelping, line 26: is this flagged elsewhere?

src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
```diff
@@ -0,0 +1,178 @@
...
+import Loadable from 'react-loadable'
+import {connect, ConnectedProps} from 'react-redux'
+import {withRouter, RouteComponentProps} from 'react-router-dom'
+
+// Components
+import {ErrorHandling} from 'src/shared/decorators/errors'
+import {Overlay} from '@influxdata/clockface'
+import {PluginCreateConfigurationFooter} from 'src/writeData/components/PluginCreateConfigurationFooter'
+
+const spinner = <div />
```
hoorayimhelping, line 12: should this be an actual loading spinner?

src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
```diff
@@ -0,0 +1,178 @@
...
+type ReduxProps = ConnectedProps<typeof connector>
+type Props = ReduxProps & RouteComponentProps<{orgID: string}>
+
+@ErrorHandling
+class TelegrafUIRefreshWizard extends PureComponent<Props> {
+  public state = {pluginConfig: '', isValidConfiguration: false}
+
+  public componentDidMount() {
+    const {bucket, buckets} = this.props
+    if (!bucket && buckets && buckets.length) {
```
hoorayimhelping, line 56:
```suggestion
    if (!bucket && Array.isArray(buckets) && buckets.length) {
```

_probably_ better to use an array check, rather than a falsy check, just for self documenting code purposes.

src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
```diff
@@ -0,0 +1,178 @@
...
+  public render() {
+    const {currentStepIndex} = this.props
+    let overlayBodyClassName = 'data-loading--overlay'
+
+    if (currentStepIndex === 1) {
+      overlayBodyClassName = ''
+    }
+
+    return (
+      <Overlay visible={true}>
```
hoorayimhelping, line 72: so is this overlay always going to be shown? will the component always be in the DOM?

hoorayimhelping (approved):
looks good to me, just a couple of oddball questions

src/dataLoaders/components/collectorsWizard/CollectorsStepSwitcher.tsx
TCL735: No. It was flagged for one purpose and I moved this flagging to a higher component. The other work flow does not come into this hierarchy.

src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
TCL735: Yes, technically it is always shown once mounted. The `onDismiss` handler uses `history.push` to re-route the user which causes the Overlay to unmount from the DOM.

But this is a good observation. Clearly we are only relying on history.push to unmount the Overlay. I think it would be better to also set the visibility to `false` in the `onDimiss` handler. I will push up that change.

src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
TCL735: 👍

src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
TCL735: Good question. This was carried over from another Wizard that this new Wizard is based on. The old Wizard clearly has some tech debt regarding the spinner.

I have put in a `<PageSpinner />` rather than a `<div>` for the new Wizard. Looks much better.

Thanks for the sharp observation!

ui-ci-bot-influx:
Metric | Score | Level
performance | 66 | :yellow_circle:
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2699)

PR #2712, opened 2021-09-27
chore: upgrade lsp, fix breaking changes

### LINKED ISSUE #2553: Fix / address lsp issues after lsp upgrade
There has been a rewrite of the lsp, bringing it to version ~v0.6.2~ (now 7.0). There was an attempt to upgrade the UI to the latest version but we are seeing several test failures. We would like to address these failures so that we can get the latest rewrite integrated into the UI so that we do not fall behind. 

This effort will be coordinated by @genehynson from the UI team and @rockstar on the Flux team.

Existing PR with failures: https://github.com/influxdata/ui/pull/2444

genehynson:
Closes #2553 

Upgrades us to the new flux library (v0.8.0) and fixes the corresponding breaking changes


src/external/monaco.flux.messages.ts
```diff
@@ -167,13 +167,16 @@ export const parseResponse = (response: ServerResponse): LSPResponse => {
     throw new Error(error)
   }
 
-  const split = (message || '').split('\r\n')
+  if (message) {
```
genehynson, line 170: new lsp library will return an undefined `message` so we need to handle that here

src/flows/pipes/Notification/view.tsx
```diff
@@ -9,7 +9,7 @@ import React, {
   Suspense,
 } from 'react'
 import {useDispatch} from 'react-redux'
-import {parse, format_from_js_file} from '@influxdata/flux'
+import {parse, format_from_js_file} from '@influxdata/flux-lsp-browser'
```
genehynson, line 12: these functions live with the wasm in the new library

src/external/monaco.flux.messages.ts
rockstar: For other reviewers: in the spec, some requests don't have responses. The old LSP would just spit some empty JSON over the wire, but the new LSP doesn't respond with anything if nothing is expected.

src/external/monaco.flux.messages.ts
```diff
@@ -167,13 +167,16 @@ export const parseResponse = (response: ServerResponse): LSPResponse => {
...
-    return JSON.parse(split.slice(2).join('\n'))
-  } catch (e) {
-    throw new Error('failed to parse LSP response')
+    try {
+      return JSON.parse(split.slice(2).join('\n'))
+    } catch (e) {
+      throw new Error('failed to parse LSP response', e)
+    }
   }
+  return undefined
```
rockstar, line 179: This is probably a question for the UI team specifically, but `return undefined` is equivalent to no return at all. Not sure if we want to be explicit here or not.

src/external/monaco.flux.server.ts
```diff
@@ -53,7 +53,7 @@ export interface WASMServer extends Server {
   register_tag_keys_callback: (TagKeysCallback) => void
 }
 
-import {format_from_js_file} from '@influxdata/flux'
+import {format_from_js_file} from '@influxdata/flux-lsp-browser'
```
rockstar, line 56: So glad this came together so easily.

rockstar (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2712)

src/external/monaco.flux.messages.ts
genehynson: fair point. I thought typescript would complain without the return, but looks like it doesn't care so I'll remove this

jestSetup.ts
```diff
@@ -25,6 +25,29 @@ window.flushAllPromises = async () => {
 fetchMock.enableMocks()
 jest.mock('src/shared/utils/errors')
 
+/*
+  N.B. when using react testing library render()
+  the mocked values in src/external/parser are ignored.
+  So, need to mock here as well
+*/
+jest.mock('src/external/parser', () => ({
```
genehynson, line 33: need to mock this for all jest tests that render components

genehynson: thanks for pointing this out. I refactored this implementation to look like the other ones and was able to delete the other mocking instances...except for 1 where it was testing a specific response from the parser and needed to be mocked on its own

cypress/e2e/shared/explorer.test.ts
```diff
@@ -450,16 +450,15 @@ describe('DataExplorer', () => {
           .within(() => {
             cy.get('textarea').type('foo |> bar', {force: true})
 
-            cy.get('.squiggly-error').should('be.visible')
+            // TODO the new flux lsp doesn't support squigglies.
+            // Turn this back on after [INSERT ISSUE] is fixed
```
genehynson, line 454: @rockstar reminder to both of us to put the link to the squiggily issue here before we merge this

src/external/monaco.flux.messages.ts
```diff
@@ -159,29 +158,17 @@ export const signatureHelp = (
...
-  }
-
-  const split = (message || '').split('\r\n')
-
-  try {
-    return JSON.parse(split.slice(2).join('\n'))
-  } catch (e) {
-    throw new Error('failed to parse LSP response')
+export const parseResponse = (response: string): LSPResponse => {
+  if (response) {
```
genehynson, line 162: the new library either returns a message or throws an error - no more "message.get_error()"

src/external/monaco.flux.messages.ts
```diff
@@ -159,29 +158,17 @@ export const signatureHelp = (
...
+      throw new Error('failed to parse LSP response')
+    }
   }
 }
 
 export async function sendMessage(message: LSPMessage, server) {
-  const stringifiedMessage = JSON.stringify(message),
-    size = stringifiedMessage.length
-
-  const fullMessage = `Content-Length: ${size}\r\n\r\n${stringifiedMessage}`
```
genehynson, line 183: the new library does not require sending `Content-Length`

src/external/monaco.flux.server.ts
```diff
@@ -44,142 +37,52 @@ import {
...
+  private server: Lsp
   private messageID: number = 0
   private documentVersions: {[key: string]: number} = {}
   public store: Store<AppState & LocalStorage>
+  private p2m: ProtocolToMonacoConverter
 
-  constructor(server: WASMServer, reduxStore = getStore()) {
+  constructor(server: Lsp, reduxStore = getStore()) {
     this.server = server
-    this.server.register_buckets_callback(this.getBuckets)
```
genehynson, line 128: these handlers are not available on the new library yet. They will be re-introduced as notification messages which we will handle in the `onMessage` handler

src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -34,41 +29,22 @@ export interface EditorProps {
...
     'flux-editor--monaco__autogrow': autogrow,
   })
 
-  const updateDiagnostics = (diagnostics: Diagnostic[]) => {
-    if (editorInst) {
-      const results = p2m.asDiagnostics(diagnostics)
-      monacoEditor.setModelMarkers(editorInst.getModel(), 'default', results)
-    }
-  }
-
```
genehynson, line 64: we no longer need to handle setting the model markers in the React component - the `monaco.flux.server` handles this for us when it receives a message via `onMessage` and updates the appropriate monaco instance via the URI in the message

cypress/e2e/shared/explorer.test.ts
```diff
@@ -437,14 +437,16 @@ describe('DataExplorer', () => {
 
       cy.getByTestID('time-machine--bottom').then(() => {
         cy.getByTestID('flux-editor', {timeout: 30000})
+          .wait(1000) // TODO: find another way to wait on the LSP to be initalized
```
genehynson, line 440: this is required otherwise the lsp isn't initialized in time...need to look and see if there is something else we can wait on

src/external/monaco.flux.server.ts
```diff
@@ -44,142 +37,52 @@ import {
...
-      v1.tagValues(bucket:"${bucket}", tag: "${tag}")`
-
-  event('runQuery', {context: 'monaco'})
-  const raw = await runQuery(orgID, query).promise
-  if (raw.type !== 'SUCCESS') {
-    throw new Error('failed to get tagKeys')
-  }
-
-  return raw
-}
```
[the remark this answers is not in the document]
genehynson: Yeah. I figured since we would have to re-write them anyways I would just remove them for now until they are implemented in the lsp

cypress/e2e/shared/explorer.test.ts
genehynson: found that I could do `cy.get('.squiggly-error', {timeout: 30000})` instead

src/external/monaco.flux.messages.ts
asalem1: Question, where does it land with table columns named error that aren’t necessarily errors?

asalem1 (approved):
Holy moly, way to go on this. Thanks for seeing it through. Not sure if this question is relevant but I think we used to build in some logic in webpack to deal with WASM. Not sure if that’s still relevant here or if we can remove that too?

src/external/monaco.flux.messages.ts
```diff
@@ -41,7 +40,7 @@ const createRequest = (id: number, method: string, params: object = {}) => {
 }
 
 export const initialize = (id: number) => {
-  return createRequest(id, 'initialize')
+  return createRequest(id, 'initialize', {capabilities: {}})
```
drdelambre, line 43: what else can this be?

src/external/monaco.flux.server.ts
drdelambre: why would the autocomplete functionality be replaced by notification messages?

src/external/parser.ts
```diff
@@ -1,13 +1,16 @@
...
+} from '@influxdata/flux-lsp-browser'
 /*
   NOTE: This is a work around for flux being generated (from rust) for the browser and jest tests running in
         a node environment (this is only for handling tests). If a test requires a specific AST result
         then you will need to mock that out in the test.
 */
 export const parse = (script): File => {
   if (window) {
-    return require('@influxdata/flux').parse(script)
+    return flux_parse(script)
```
drdelambre, line 13: this change is because you mocked the library?

drdelambre: if that's the case, then... shouldn't this file be retired / roll this distinction to that part of the code?

drdelambre (approved):
sad to hear we lose the typeahead, and i'm confused about the jest strategy, but that's not a super big priority for me

drdelambre:
this weirdness is what @asalem1 was talking about:
https://github.com/influxdata/ui/blob/master/webpack.common.ts#L49-L57

it's in there because one of our libraries is using the standard async loader, another isn't. the onigasm vs the fluxparser iirc, but i cant remember which was  doing which

genehynson:
> sad to hear we lose the typeahead, and i'm confused about the jest strategy, but that's not a super big priority for me

the jest stuff is a little weird for sure. The issue is that now the `parse` and `format_from_js_file` functions are in the same library as the wasm so we can't use them directly because we can't load the wasm into the jest environment. @rockstar mentioned that they will separate these functions out into another library in a future release so we should be able to clean it up then 😄

genehynson:
> Question, where does it land with table columns named error that aren’t necessarily errors?

@asalem1 the new library will actually throw an error now so it should resolve the issue of table columns named `error` that aren't actually an error.

src/external/monaco.flux.messages.ts
genehynson: here are all the options we can provide to `capabilities` https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#clientCapabilities . The `initialize` [request](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#initialize) requires that you at least provide the empty `capabilities` object - so we were not adhering to the LSP spec until now.

For now, to keep us

src/external/parser.ts
genehynson: let me double check but I think I started trying to remove this when I started work on this PR by in Sept but I don't remember what happened.

genehynson: yeah I can't get the mocks to work without this file. Once we get the next update to the lsp we shouldn't need to mock these functions anymore and can easily clean this up. So I'll defer this refactoring until then.

PR #2720, opened 2021-09-27
feat: better supported text

### LINKED ISSUE #2707: Notebooks: Present an empty Flux script when you select "Schedule a Task"
When a user opens the "Schedule a Task" option, they see some slightly confusing boilerplate code. We would like to remove that boilerplate code and present an empty Flux panel so that the user may begin authoring with a clean slate. 

## Acceptance Criteria
- When a user opens the "Schedule a Task" notebook, they see a blank Flux panel for editing.

drdelambre:
Closes #2707

### DIFF
--- a/src/flows/templates/types/task.ts
+++ b/src/flows/templates/types/task.ts
@@ -14,8 +14,7 @@ export default register =>
             activeQuery: 0,
             queries: [
               {
-                text:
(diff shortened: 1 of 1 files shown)


genehynson (approved):
lol branch name `alex_empty`.

PR #2721, opened 2021-09-27
fix: allow imports for previous_results

### LINKED ISSUE #2609: __PREVIOUS_RESULT__ substitution needs an upgrade
we still have a regexp mutating a query string floating around in the code base. CRAZY.
this is breaking whenever we have imports in the previous result
one place where this is becoming a problem is the use case of
```
panel one:
import "influxdata/influxdb/sample" 
sample.data(set: "noaa")

panel two:
sorted = __PREVIOUS_RESULT__
    |> keep(columns: ["_time"])
    |> group()
    |> sort(columns: ["_time"])

sorted
    |> first(column: "_time")
    |> yield(name: "first")
```

this seems like it would also be a problem if multiple instances were used like:
```
panel one:
import "influxdata/influxdb/sample" 
sample.data(set: "noaa")
    |> keep(columns: ["_time"])
    |> group()
    |> sort(columns: ["_time"])

panel two:
__PREVIOUS_RESULT__
    |> first(column: "_time")
    |> yield(name: "first")

__PREVIOUS_RESULT__
    |> last(column: "_time")
    |> yield(name: "last")
```

we want to make sure that we pull out the imports from the incoming query and join them with the outgoing query and use AST substitution instead of regexp every time we are dealing with __PREVIOUS_RESULT__

drdelambre:
Closes #2609

joins imports from both queries
injects body of the last query using ast (kind of)
supports multiple injections

### DIFF
--- a/src/flows/pipes/RawFluxEditor/index.ts
+++ b/src/flows/pipes/RawFluxEditor/index.ts
@@ -4,8 +4,6 @@ import View from './view'
 import ReadOnly from './readOnly'
 import './style.scss'
 
-const PREVIOUS_REGEXP = /__PREVIOUS_RESULT__/g
-
 export default register => {
   register({
     type: 'rawFluxEditor',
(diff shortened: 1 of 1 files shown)


genehynson (approved):
it's a good thing it supports multiple injections so it can get that double dose of the covid vax 😆

drdelambre:
just got the required booster to get through the CI pipeline as well!

PR #2723, opened 2021-09-28
feat: create mini file dnd element

### LINKED ISSUE #2659: create new mini-file upload widget
Uploading Measurement Schema files (https://app.zenhub.com/workspaces/applicationdumplings-60a80d872533670017a6b107/issues/influxdata/ui/2510) needs a widget to upload files.

the current existing widget has these limitations:

1) there are three of them that are almost but not quite exactly the same
2) the drop file listener attaches to the window, and not to the component, such that there can only be one drag-and-drop file uploader per page.  there needs to be multiple of these per page. (the user may be uploading multiple measurement schemas)
    a) if there is more than one, once you do a drop onto one, they *all* get the same file
    b)I tried retrofitting the listener to a component, it was not successful
3) the drop file listener also accepts all file types.  the  file input (click to upload) takes an acceptable file extension list, but the drop listener just accepts everything
4) the retrofitting is getting to a point where a new component is needed to cut down the complexity.
5) I am worried that doing too much work on the original component may break existing functionality.
6) I tried re-using the css file from the original component, but since the original is a class component, and the new one is a functional component, the variable 'glue' that populates the variables is problematic; so the plan is to write a new css file based on the original css file (it will be a subset) and then later unify them.

jrenee42:
Closes #2659

### DIFF
--- a/src/buckets/components/createBucketForm/MiniFileDnd.tsx
+++ b/src/buckets/components/createBucketForm/MiniFileDnd.tsx
@@ -103,29 +103,30 @@ export const MiniFileDnd: FC<Props> = ({
     }
   }
 
-  const handleFileClick = (e: any): void => {
-    const file: File = e.currentTarget.files[0]
+  const handleFileClick = (event: any): void => {
+    const file: File = event.currentTarget.files[0]
 
     if (!file) {
       return
     }
 
-    e.preventDefault()
-    e.stopPropagation()
+    event.preventDefault()
+    event.stopPropagation()
 
     // don't need to see if the file type is valid; the input filters it already for us
     processFile(file)
   }
 
-  const dropHandler = (e: any): void => {
-    const file = e.dataTransfer.files[0]
-    const fileType = file.type
+  const dropHandler = (event: any): void => {
+    const file = event.dataTransfer.files[0]
 
     if (!file) {
       return
     }
-    e.preventDefault()
-    e.stopPropagation()
+    event.preventDefault()
+    event.stopPropagation()
+
+    const fileType = file.type
 
     if (allowedTypes.includes(fileType)) {
       processFile(file)
--- a/src/buckets/components/createBucketForm/MiniFileDnd.tsx
+++ b/src/buckets/components/createBucketForm/MiniFileDnd.tsx
@@ -110,9 +110,6 @@ export const MiniFileDnd: FC<Props> = ({
       return
     }
 
-    event.preventDefault()
-    event.stopPropagation()
-
     // don't need to see if the file type is valid; the input filters it already for us
     processFile(file)
   }
--- a/src/buckets/components/createBucketForm/MiniFileDnd.tsx
+++ b/src/buckets/components/createBucketForm/MiniFileDnd.tsx
@@ -32,7 +32,18 @@ export const setGrammar = (fileTypes: string[]) => {
   return addOrToLastOne.join(', ')
 }
 
-/** read here:
+/**
+ *  This is a small File Drag-and-Drop Input
+ *  It just has some text, with an outline.  there is no icon.
+ *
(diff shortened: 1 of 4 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2723)

src/buckets/components/createBucketForm/MiniFileDnd.tsx
```diff
@@ -0,0 +1,190 @@
...
+    e.preventDefault()
+    e.stopPropagation()
+
+    // don't need to see if the file type is valid; the input filters it already for us
+    processFile(file)
+  }
+
+  const dropHandler = (e: any): void => {
+    const file = e.dataTransfer.files[0]
+    const fileType = file.type
```
hoorayimhelping, line 122: feels like this line

```ts
const fileType = file.type
```

should get moved under

```ts
if (!file) {
  return
}
```

src/buckets/components/createBucketForm/MiniFileDnd.tsx
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    e.preventDefault()
+    e.stopPropagation()
+
+    // don't need to see if the file type is valid; the input filters it already for us
+    processFile(file)
+  }
+
+  const dropHandler = (e: any): void => {
```
hoorayimhelping, line 120:
```suggestion
  const dropHandler = (ev: any): void => {
```

for consistency and also cause short variable names just make comprehending code a little bit more difficult. i'd prefer if these `ev` were named `event` or if you _must_ abbreviate them, `evt`

src/buckets/components/createBucketForm/MiniFileDnd.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+  const handleFileClick = (e: any): void => {
+    const file: File = e.currentTarget.files[0]
+
+    if (!file) {
+      return
+    }
+
+    e.preventDefault()
+    e.stopPropagation()
```
hoorayimhelping, line 114: out of curiosity, why are we stopping the event from propagating up?

src/buckets/components/createBucketForm/MiniFileDnd.tsx
```diff
@@ -0,0 +1,190 @@
...
+      ref={inputEl}
+      className="drag-and-drop--input"
+      accept={allowedExtensions}
+      onChange={handleFileClick}
+    />
+  )
+
+  // it gets clicked programmatically since the input itself has display='none'
+  const handleFileOpen = (): void => {
+    inputEl.current.click()
```
hoorayimhelping, line 160: do we need to check if `inputEl` has a `current` element? or is it pretty much impossible to run the `handleFileOpen` function if the component isn't mounted in the DOM?

hoorayimhelping (commented):
a couple of small comments and questions

src/buckets/components/createBucketForm/MiniFileDnd.tsx
jrenee42: correct, cannot run the function if it isn't mounted

src/buckets/components/createBucketForm/MiniFileDnd.tsx
jrenee42: i copied it from the previous dnd file component.  I tested it without those two lines, and no change, so I removed them.

hoorayimhelping (approved):

PR #2732, opened 2021-09-29
feat: add editing an existing telegraf config with a plugin

### LINKED ISSUE #1744: Telegraf Refresh: Select Configuration File--Edit Mode
GIVEN  that a user in on the `Load Data` page, 
WHEN the user selects the `Add to Existing Configuration` CTA from one of the available Telegraf Plugins, 
THEN a modal is displayed with a list of the user's available config files, and a `Continue Configuring` button is displayed.

Note 1:  The name field is optional.
Note 2:  The config file field is required and can only have 1 selection.

https://www.figma.com/file/hMUcxA5RdoEhdJLvClWWB7/Agent-Configuration?node-id=263%3A0

### LINKED ISSUE #1748: Telegraf Refresh: Validate Configuration--Edit Mode
Given a user is in the Add Plugin to Existing Config flow
When the user selects a config file,
Then check whether the selected plug-in already exists within the selected configuration file.  If it does, display a warning `This Agent Configuration already contains {name} Input Plugin.  Are you sure you want to add another one?`. Provide a CTA to continue.

https://www.figma.com/file/hMUcxA5RdoEhdJLvClWWB7/Agent-Configuration?node-id=382%3A0

### LINKED ISSUE #2696: Telegraf Refresh: e2e test for adding a plugin to an existing Telegraf configuration
Create a Cypress e2e test for Data > Sources > input plugin > Add to an existing configuration to ensure that:

- [ ] happy path of adding to an existing configuration works
- [ ] user can add a duplicate by confirming
- [ ] user can close the overlay during any step, restart from beginning, and then successfully complete happy path
- [ ] user can edit the configuration with any text

TCL735:
Closes #1744 
Closes #1748 
Closes #2696 

Create a new config with a plugin. Then add the same plugin again to the same config:

https://user-images.githubusercontent.com/10736577/135343488-788300ed-566e-4fcf-808a-ca9b293ce54c.mp4

Create a new config with a plugin. Then add a different plugin to the same config:

https://user-images.githubusercontent.com/10736577/135344225-7afd0d4b-750d-4989-aa59-266241fdeae4.mp4

### DIFF
--- a/src/writeData/components/AddPluginToConfiguration.tsx
+++ b/src/writeData/components/AddPluginToConfiguration.tsx
@@ -1,10 +1,15 @@
 // Libraries
-import React, {FC} from 'react'
+import React, {FC, useEffect} from 'react'
+import {connect, ConnectedProps} from 'react-redux'
 import {Route, Switch} from 'react-router-dom'
 
 // Components
 import {Columns, ComponentColor, Dropdown, Grid} from '@influxdata/clockface'
-import PluginCreateConfigurationWizard from 'src/writeData/components/PluginCreateConfigurationWizard'
+import PluginCreateConfiguration from 'src/writeData/components/PluginCreateConfigurationWizard'
+import PluginAddToExistingConfiguration from 'src/writeData/components/PluginAddToExistingConfiguration/Wizard'
+
+// Actions
+import {getTelegrafs} from 'src/telegrafs/actions/thunks'
 
 // Constants
 import {
@@ -13,6 +18,12 @@ import {
   ORG_ID,
   TELEGRAF_PLUGINS,
 } from 'src/shared/constants/routes'
+
+// Types
+import {AppState, ResourceType, Telegraf} from 'src/types'
+
+// Selectors
+import {getAll} from 'src/resources/selectors'
 
 // Styles
 import 'src/writeData/components/AddPluginToConfiguration.scss'
@@ -30,84 +41,121 @@ interface AddPluginToConfigurationCTAProps {
   >
 }
 
-export const AddPluginToConfigurationCTA: FC<AddPluginToConfigurationCTAProps> = ({
-  contentID,
-  history,
-  orgID,
-  thumbnail,
-  pageContent,
-}) => (
-  <>
-    <div className="write-data--details">
-      <Grid>
-        <Grid.Row>
-          <Grid.Column
-            widthXS={Columns.Twelve}
-            widthMD={Columns.Six}
-            widthLG={Columns.Three}
-          >
-            <div className="write-data--details-thumbnail">{thumbnail}</div>
-            <Dropdown
-              button={(active, onClick) => (
-                <Dropdown.Button
-                  active={active}
-                  color={ComponentColor.Primary}
-                  style={{width: '100%'}}
-                  onClick={onClick}
-                >
-                  Use this plugin
-                </Dropdown.Button>
-              )}
-              className="use-plugin--telegraf-configuration"
-              menu={onCollapse => (
-                <Dropdown.Menu onCollapse={onCollapse}>
-                  <Dropdown.Item
-                    key="Create-new-configuration-telegraf-plugin"
-                    value="Create a new configuration"
-                    onClick={() => {
-                      history.push(
-                        `/${ORGS}/${orgID}/load-data/${TELEGRAF_PLUGINS}/${contentID}/new`
-                      )
-                    }}
-                    selected={false}
-                    testID="create-new-configuration-from-plugin--dropdown-item"
-                  >
-                    <span>Create a new configuration</span>
-                  </Dropdown.Item>
-                  <Dropdown.Item
-                    key="Add-to-existing-configuration-telegraf-plugin"
-                    value="Add to an existing configuration"
-                    onClick={() => {}}
-                    selected={false}
-                  >
-                    <span>Add to an existing configuration</span>
-                  </Dropdown.Item>
-                </Dropdown.Menu>
-              )}
-              style={{width: 'auto'}}
-              testID="add-plugin-to-configuration--dropdown"
-            />
-          </Grid.Column>
-          <Grid.Column
-            widthXS={Columns.Twelve}
-            widthMD={Columns.Six}
-            widthLG={Columns.Nine}
-          >
-            <div
-              className="write-data--details-content markdown-format"
-              data-testid="load-data-details-content"
+type ReduxProps = ConnectedProps<typeof connector>
+type Props = ReduxProps & AddPluginToConfigurationCTAProps
+
+const AddPluginToConfigurationCTAComponent: FC<Props> = props => {
+  const {
+    contentID,
+    fetchTelegrafs,
+    history,
+    orgID,
+    pageContent,
+    telegrafs,
+    thumbnail,
+  } = props
+
+  useEffect(() => {
+    fetchTelegrafs()
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
+
+  return (
+    <>
+      <div className="write-data--details">
+        <Grid>
+          <Grid.Row>
+            <Grid.Column
+              widthXS={Columns.Twelve}
+              widthMD={Columns.Six}
+              widthLG={Columns.Three}
             >
-              {pageContent}
-            </div>
-          </Grid.Column>
-        </Grid.Row>
-      </Grid>
-    </div>
-    <Switch>
-      <Route
-        path={`/${ORGS}/${ORG_ID}/load-data/${TELEGRAF_PLUGINS}/${CONTENT_ID}/new`}
-        component={PluginCreateConfigurationWizard}
-      />
-    </Switch>
-  </>
+              <div className="write-data--details-thumbnail">{thumbnail}</div>
+              <Dropdown
+                button={(active, onClick) => (
+                  <Dropdown.Button
+                    active={active}
+                    color={ComponentColor.Primary}
+                    style={{width: '100%'}}
+                    onClick={onClick}
+                  >
+                    Use this plugin
+                  </Dropdown.Button>
+                )}
+                className="use-plugin--telegraf-configuration"
+                menu={onCollapse => (
+                  <Dropdown.Menu onCollapse={onCollapse}>
+                    <Dropdown.Item
+                      key="Create-new-configuration-telegraf-plugin"
+                      value="Create a new configuration"
+                      onClick={() => {
+                        history.push(
+                          `/${ORGS}/${orgID}/load-data/${TELEGRAF_PLUGINS}/${contentID}/new`
+                        )
+                      }}
+                      selected={false}
+                      testID="create-new-configuration-from-plugin--dropdown-item"
+                    >
+                      <span>Create a new configuration</span>
+                    </Dropdown.Item>
+                    {telegrafs?.length ? (
+                      <Dropdown.Item
+                        key="Add-to-existing-configuration-telegraf-plugin"
+                        value="Add to an existing configuration"
+                        onClick={() => {
+                          history.push(
+                            `/${ORGS}/${orgID}/load-data/${TELEGRAF_PLUGINS}/${contentID}/add`
+                          )
+                        }}
+                        selected={false}
+                      >
+                        <span>Add to an existing configuration</span>
+                      </Dropdown.Item>
+                    ) : null}
+                  </Dropdown.Menu>
+                )}
+                style={{width: 'auto'}}
+                testID="add-plugin-to-configuration--dropdown"
+              />
+            </Grid.Column>
+            <Grid.Column
+              widthXS={Columns.Twelve}
+              widthMD={Columns.Six}
+              widthLG={Columns.Nine}
+            >
+              <div
+                className="write-data--details-content markdown-format"
+                data-testid="load-data-details-content"
+              >
+                {pageContent}
+              </div>
+            </Grid.Column>
+          </Grid.Row>
+        </Grid>
+      </div>
+      <Switch>
+        <Route
+          path={`/${ORGS}/${ORG_ID}/load-data/${TELEGRAF_PLUGINS}/${CONTENT_ID}/new`}
+          component={PluginCreateConfiguration}
+        />
+        <Route
+          path={`/${ORGS}/${ORG_ID}/load-data/${TELEGRAF_PLUGINS}/${CONTENT_ID}/add`}
+          component={PluginAddToExistingConfiguration}
+        />
+      </Switch>
+    </>
+  )
+}
+
+const mstp = (state: AppState) => ({
+  telegrafs: getAll<Telegraf>(state, ResourceType.Telegrafs),
+})
+
+const mdtp = {
+  fetchTelegrafs: getTelegrafs,
+}
+
+const connector = connect(mstp, mdtp)
+
+export const AddPluginToConfigurationCTA = connector(
+  AddPluginToConfigurationCTAComponent
 )
--- a/src/writeData/components/AddPluginToConfiguration.tsx
+++ b/src/writeData/components/AddPluginToConfiguration.tsx
@@ -107,6 +107,7 @@ const AddPluginToConfigurationCTAComponent: FC<Props> = props => {
                           )
                         }}
                         selected={false}
+                        testID="add-to-existing-configuration-from-plugin--dropdown-item"
                       >
                         <span>Add to an existing configuration</span>
                       </Dropdown.Item>
--- a/src/writeData/components/AddPluginToConfiguration.tsx
+++ b/src/writeData/components/AddPluginToConfiguration.tsx
@@ -20,10 +20,10 @@ import {
 } from 'src/shared/constants/routes'
 
 // Types
-import {AppState, ResourceType, Telegraf} from 'src/types'
+import {AppState} from 'src/types'
 
 // Selectors
-import {getAll} from 'src/resources/selectors'
+import {getAllTelegrafs} from 'src/resources/selectors'
 
 // Styles
 import 'src/writeData/components/AddPluginToConfiguration.scss'
@@ -152,7 +152,7 @@ const AddPluginToConfigurationCTAComponent: FC<Props> = props => {
 }
 
 const mstp = (state: AppState) => ({
-  telegrafs: getAll<Telegraf>(state, ResourceType.Telegrafs),
+  telegrafs: getAllTelegrafs(state),
 })
 
 const mdtp = {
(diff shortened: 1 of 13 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2732)

src/writeData/components/AddPluginToConfiguration.tsx
```diff
@@ -30,84 +41,121 @@ interface AddPluginToConfigurationCTAProps {
...
+}
+
+const mstp = (state: AppState) => ({
+  telegrafs: getAll<Telegraf>(state, ResourceType.Telegrafs),
+})
+
+const mdtp = {
+  fetchTelegrafs: getTelegrafs,
+}
+
```
TCL735, line 156: Added this to allow the dropdown to know whether or not to display the "Add to an existing configuration" option. When there are not telegraf configurations, this would not make any sense, and so we hide this option when there are no configurations yet.

src/writeData/components/PluginCreateConfigurationOptions.tsx
```diff
@@ -55,7 +55,7 @@ const PluginCreateConfigurationOptionsComponent: FC<Props> = props => {
   }
 
   const handleSelectBucket = (bucket: Bucket) => {
-    const {orgID, id, name} = bucket
+    const {orgID, name, id} = bucket
```
TCL735, line 58: Flipped the order to match the order of the arguments in the function call a few lines below.

src/writeData/components/AddPluginToConfiguration.tsx
```diff
@@ -30,84 +41,121 @@ interface AddPluginToConfigurationCTAProps {
...
+    fetchTelegrafs,
+    history,
+    orgID,
+    pageContent,
+    telegrafs,
+    thumbnail,
+  } = props
+
+  useEffect(() => {
+    fetchTelegrafs()
```
hoorayimhelping, line 59: should this only `fetchTelegrafs` when the list of `telegrafs` changes? does this need to be fetched every time this component is rendered?

src/writeData/components/AddPluginToConfiguration.tsx
```diff
@@ -30,84 +41,121 @@ interface AddPluginToConfigurationCTAProps {
...
+              widthMD={Columns.Six}
+              widthLG={Columns.Three}
+            >
+              <div className="write-data--details-thumbnail">{thumbnail}</div>
+              <Dropdown
+                button={(active, onClick) => (
+                  <Dropdown.Button
+                    active={active}
+                    color={ComponentColor.Primary}
+                    style={{width: '100%'}}
```
hoorayimhelping, line 78: so i recently learned passing props as objects is a big cause of re-renders. since this is a completely static style, can you pull `{width: '100%'}` outside of this component and store it as a variable?

something like:

```ts
const fullWidth = {width: '100%'}
const AddPluginToConfigurationCTAComponent: FC<Props> = props => {
```

and then use it like:

```ts
                  <Dropdown.Button
                    active={active}
                    color={ComponentColor.Primary}
                    style={fullWidth}
```

that'll keep this from always re-rendering, since it won't need to [recreate the object on every render](https://www.debugbear.com/blog/react-rerenders#passing-objects-as-props)

TCL735:
I have added e2e tests. This will also close #2696

src/writeData/components/AddPluginToConfiguration.tsx
```diff
@@ -30,84 +41,122 @@ interface AddPluginToConfigurationCTAProps {
...
+                menu={onCollapse => (
+                  <Dropdown.Menu onCollapse={onCollapse}>
+                    <Dropdown.Item
+                      key="Create-new-configuration-telegraf-plugin"
+                      value="Create a new configuration"
+                      onClick={() => {
+                        history.push(
+                          `/${ORGS}/${orgID}/load-data/${TELEGRAF_PLUGINS}/${contentID}/new`
+                        )
+                      }}
```
hoorayimhelping, line 94: this is another place that will cause re-renders since this anonymous function gets re-created every render loop. I'm not sure if there's a way around it since it depends on `contentID`, but I just wanted to explain this thing.

It _might_ re-render less if the anonymous function is pulled out into a named function like:

```ts
const createNewPlugin = function createNewPlugin() {
  history.push(`/${ORGS}/${orgID}/load-data/${TELEGRAF_PLUGINS}/${contentID}/new`)
}
```
and then passing `createNewPlugin` to `onClick`. probably not worth implementing here, but something I'd like all of us to start thinking about when writing code.

src/writeData/components/AddPluginToConfiguration.tsx
```diff
@@ -30,84 +41,122 @@ interface AddPluginToConfigurationCTAProps {
...
+                        }}
+                        selected={false}
+                        testID="add-to-existing-configuration-from-plugin--dropdown-item"
+                      >
+                        <span>Add to an existing configuration</span>
+                      </Dropdown.Item>
+                    ) : null}
+                  </Dropdown.Menu>
+                )}
+                style={{width: 'auto'}}
```
hoorayimhelping, line 117: another place to replace the object being recreated with a static object

src/writeData/components/AddPluginToConfiguration.tsx
```diff
@@ -30,84 +41,122 @@ interface AddPluginToConfigurationCTAProps {
...
+                        history.push(
+                          `/${ORGS}/${orgID}/load-data/${TELEGRAF_PLUGINS}/${contentID}/new`
+                        )
+                      }}
+                      selected={false}
+                      testID="create-new-configuration-from-plugin--dropdown-item"
+                    >
+                      <span>Create a new configuration</span>
+                    </Dropdown.Item>
+                    {telegrafs?.length ? (
```
hoorayimhelping, line 100: since the `else` case of this is `: null`, this can be safely changed to:

```ts
{telegrafs?.length && (

File: src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx [hunk 8]
```diff
@@ -0,0 +1,143 @@
...
+
+  if (!selectedTelegraf) {
+    setIsValidConfiguration(false)
+  }
+
+  const selectedName = selectedTelegraf
+    ? selectedTelegraf.name
+    : 'Select a configuration'
+
+  const handleSelectTelegrafConfiguration = (telegraf: Telegraf) => {
```
hoorayimhelping, line 52: if `buckets` are passed into this function, it can be created outside of the function component so it's not re-created on every render. if you name it by doing something like

```ts
const handleSelectTelegrafConfiguration = function handleSelectTelegrafConfiguration() {
```
it should re-render even less. When react receives an anonymous function, it will always re-render it since it's different from the previous anonymous function, even if the bodies do the same thing.

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
```diff
@@ -0,0 +1,143 @@
...
+        button={(active, onClick) => (
+          <Dropdown.Button
+            active={active}
+            onClick={onClick}
+            status={ComponentStatus.Default}
+          >
+            {selectedName}
+          </Dropdown.Button>
+        )}
+        menu={onCollapse => (
```
hoorayimhelping, line 83: potential opportunity to pull out an anonymous function into a named function

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
```diff
@@ -0,0 +1,143 @@
...
+        testID="plugin-add-to-configuration--select-telegraf"
+      />
+    </>
+  )
+}
+
+const mstp = (state: AppState) => {
+  const buckets = getAll<Bucket>(state, ResourceType.Buckets)
+  const {telegrafConfigID} = getDataLoaders(state)
+  const telegrafs = getAll<Telegraf>(state, ResourceType.Telegrafs)?.sort(
```
hoorayimhelping, line 109: i think this is a really good opportunity to use a memoized selector using reselect. let's you and me talk tomorrow about creating a selector here and using reselect so that it's not re-sorting on every render call.

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
```diff
@@ -0,0 +1,143 @@
...
+    telegrafConfigID,
+    telegrafs,
+  }
+}
+
+const mdtp = {
+  onSetBucketInfo: setBucketInfo,
+  onSetTelegrafConfigDescription: setTelegrafConfigDescription,
+  onSetTelegrafConfigID: setTelegrafConfigID,
+  onSetTelegrafConfigName: setTelegrafConfigName,
```
hoorayimhelping, line 138: are you a fan of remapping the actions into this style?

I much prefer not renaming these, so something like:

```ts
const mdtp = {
  setBucketInfo,
  setTelegraf,
  setTelegrafConfigID,
  setTelegrafConfigName
}
```

i'd prefer to keep them named this way for 2 reasons:

1. `onAction` is usually the property that represents what to do in response to an event, eg `onClick`, `onSelect`, `onChange`. It's kind of awkward to have something like `onChange={onSetBucketInfo}`. In my opinion, it's much more natural to say `onChange={setBucketInfo}` - it reads like an english sentence
2. if you search the whole codebase for `setBucketInfo` (or worse, `setBucketInfo\(`) you won't be able to see `onSetBucketInfo`. You'll have to find where it's called in this function with a separate search.

src/writeData/components/PluginAddToExistingConfiguration/Customize.tsx
```diff
@@ -0,0 +1,167 @@
...
+    setPluginConfig,
+    telegrafConfig,
+    telegrafConfigDescription,
+    telegrafConfigName,
+  } = props
+
+  const handleError = error => {
+    setIsValidConfiguration(false)
+    setPluginConfig(`${error}`)
+    console.error(error)
```
hoorayimhelping, line 50: I'm trying to get us away from `console.error` in favor of a notification to the user that's actionable. Is it reasonable to have a notification here?

src/writeData/components/PluginAddToExistingConfiguration/Customize.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const handleChangeConfig = config => setPluginConfig(config)
+
+  const handleNameInput = (event: ChangeEvent<HTMLInputElement>) => {
+    onSetTelegrafConfigName(event.target.value)
+  }
+
+  const handleDescriptionInput = (event: ChangeEvent<HTMLInputElement>) => {
+    onSetTelegrafConfigDescription(event.target.value)
+  }
```
hoorayimhelping, line 93: :+1:

src/writeData/components/PluginAddToExistingConfiguration/Customize.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const telegrafs = getAll<Telegraf>(state, ResourceType.Telegrafs)
+
+  const selectedTelegraf = telegrafs?.find(
+    telegraf => telegraf.id === telegrafConfigID
+  )
+  return {
+    telegrafConfig: selectedTelegraf?.config ?? '',
+    telegrafConfigDescription,
+    telegrafConfigName,
+  }
```
hoorayimhelping, line 156: I think there might be some opportunity here to memoize these selectors as well

src/writeData/components/PluginAddToExistingConfiguration/Footer.tsx
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  const pattern = `[[inputs.${pluginConfigName}]]`
+
+  let FooterButtons = null
+
+  if (currentStepIndex === PluginConfigurationStep.Configure) {
+    FooterButtons = (
+      <FooterConfigure
+        hasDuplicatePlugin={telegrafConfig?.config?.includes(pattern) ?? false}
```
hoorayimhelping, line 161: Just be aware that `??` only fails on `null` and `undefined`. It seems like that's what we want here, but I just wanted to call that out

hoorayimhelping (commented):
A+ on alphabetizing things! ⭐ 

have a few comments - let me know what you think. Totally okay to push back if you don't agree with them

src/writeData/components/AddPluginToConfiguration.tsx
TCL735: This component is very high up on the hierarchy. It only renders once when a plugin's page is loaded. There may be re-renders but not re-fetches. The `useEffect` has an empty dependency list which means it only runs once when the component is mounted.

The beauty of this fetching is that on the plugin's page, the user can create a new telegraf configuration. But that gets added to state, which then is part of the resources checked by the sub-components. In other words, `telegrafs` remain updated after 1 fetch when the component is mounted, and there is no re-fetching even when a new telegraf is created on this page.

src/writeData/components/AddPluginToConfiguration.tsx
TCL735: Updated.

src/writeData/components/AddPluginToConfiguration.tsx
TCL735: Updated.

src/writeData/components/AddPluginToConfiguration.tsx
TCL735: Updated.

src/writeData/components/AddPluginToConfiguration.tsx
TCL735: Updated.

src/writeData/components/PluginAddToExistingConfiguration/Customize.tsx
TCL735: No reason. It gets caught. But I thought it would be easier to spot an error if someone happens to be debugging. I can remove it.

src/writeData/components/PluginAddToExistingConfiguration/Footer.tsx
TCL735: Yes, completely intended here. If any property is missing (`undefined`) or if the pattern can't be found (`undefined`), there are no duplicates (`false`). `null` can also substitute for the `undefined` if a property is deliberately nothing.

src/writeData/components/PluginAddToExistingConfiguration/Customize.tsx
TCL735: i tried. Memoizing here has no effect. This component has the input fields for name and description, and also the editor for the configuration, all of which the user can change. `mstp` is responsible for updating the props to this component when the state changes. The state changes whenever the user changes any of the input fields or the contents of the editor.

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
TCL735: We have to rename them, or else there is a mirroring conflict. `mdtp` actually curries the functions in the object and creates new functions. Hence the renaming. If we keep the same name then there would be two functions named `setBucketInfo` - one of which is imported and one which is created by `mdtp`. It is very confusing to debug if there is ever an issue and the function is named the same thing even though the passed-in-prop version of that function has different arguments.

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
TCL735: This is true. For a large number of telegrafs... there is a LOT of sorting going on.

I think I can get started on writing a memoized `getTelegrafs` selector that is already sorted.

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
TCL735: i made a long video of this and pinged you in Slack.

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
TCL735: Similar to above. I made a video of this and pinged you in Slack.

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
hoorayimhelping: >If we keep the same name then there would be two functions named setBucketInfo - one of which is imported and one which is created by `mdtp`

there are two functions named that, true. But it's only an issue if they're destructured out of the `props` object. Even then, it's not a collision issue, because the scope of the function component is more specific / deeper than the scope that the functions are imported into. (JavaScript is function scoped, so new functions create new variable scopes).

```ts
const name = 'Bucky'

const whoIsNumberOne = () => {
  const name = 'Timmy'
  return name
}

console.log(whoIsNumberOne()) // Timmy
```

I'd make the argument that keeping them within the props object (rather than destructuring) is actually the most documenting and easy to understand, especially in the context of a render method, or a function component.

```ts
const GiveMeTheTime: FC = (props) => {
  const computedValue = Date.now() * this.props.timeZone

  return (
    <div className={this.props.timeClassName}>
      {computedValue}
    </div>
  )
}
```

I think that this form is more self-documenting, because it is explicit about which values being used are from props, and which are computed (anything with `this.props` is from props, anything without it isn't).

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
TCL735: I have a solution!

First, it should be noted that getting a resource, in this case Telegrafs and Buckets, from state is relatively quick. It's just accessing a property in a giant nested object. Accessing is very fast. Therefore, we can't really do anything to make getting a resource from state any more efficiency. It's fine to keep the accessing part.

The part that should be optimized is the sorting. We do not want to re-sort an array of resources every time we access it from state. (Resources are stored in state unsorted, in the order they were added.)

We can use a `useEffect` that only runs on mounting to do the sorting. This `useEffect` will have no dependencies and will never run again on re-renders. Thus, our sorting will never be done more than once per component mount.

This particular component does not allow for adding more Telegrafs or Buckets. We can safely assume that one sorting done initially will include all of the Telegrafs and Buckets that are available to the user.

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
hoorayimhelping: tl:dw: naming these doesn't affect re-rendering in this component tree

src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx
hoorayimhelping: >Therefore, we can't really do anything to make getting a resource from state any more efficient.

Probably true, but we can do a lot to make it less efficient, and we need to be on the guard for that.

hoorayimhelping (approved):

PR #2748, opened 2021-09-30
test(cypress): export dashboard

Dfldr:
This PR adds a test that ensures that both export options work properly.

### DIFF
--- a/cypress/e2e/shared/dashboardsIndex.test.ts
+++ b/cypress/e2e/shared/dashboardsIndex.test.ts
@@ -1,4 +1,4 @@
-import { Organization } from '../../../src/types'
+import {Organization} from '../../../src/types'
 
 const newLabelName = 'click-me'
 const dashboardName = 'Bee Happy'
@@ -9,8 +9,8 @@ describe('Dashboards', () => {
   beforeEach(() =>
     cy.flush().then(() =>
       cy.signin().then(() =>
-        cy.fixture('routes').then(({ orgs }) => {
-          cy.get<Organization>('@org').then(({ id }: Organization) => {
+        cy.fixture('routes').then(({orgs}) => {
+          cy.get<Organization>('@org').then(({id}: Organization) => {
             cy.visit(`${orgs}/${id}/dashboards-list`)
             cy.getByTestID('tree-nav')
             cy.exec('rm cypress/downloads/*', {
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2748)

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -544,4 +544,48 @@ describe('Dashboards', () => {
...
+      cy.createDashboard(orgID).then(({body}) => {
+        cy.fixture('routes').then(({orgs}) => {
+          cy.visit(`${orgs}/${orgID}/dashboards/${body.id}`)
+          cy.getByTestID('tree-nav')
+          cy.getByTestID('nav-item-dashboards').click()
+          cy.getByTestID('dashboard-card').invoke('hover')
+          cy.getByTestID('context-export-menu').click()
+          cy.getByTestID('context-menu-item-export').click()
+          cy.getByTestID('button').click()
+          cy.wait(1000)
```
genehynson, line 562: lets try and find another way to wait here other than a fixed time amount. Could we use `cy.intercept` to wait on the download request to finish?

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -544,4 +544,48 @@ describe('Dashboards', () => {
       })
     })
   })
+
+  before('Clear downloads folder', () => {
+    cy.exec('rm cypress/downloads/*', {log: true, failOnNonZeroExit: false})
```
genehynson, line 549: I believe we can move this into the `beforeEach` since we want this directory to be clean on every test

genehynson (changes requested):
thanks for adding this. You'll want to run `yarn prettier:fix` to resolve the linting issues. And I left a couple other suggestions

cypress/e2e/shared/dashboardsIndex.test.ts
Dfldr: I decided to delete cy.wait from the code, since every command has a 4s timeout by Cypress' default configuration.
If the downloaded file is not read by cy.readfile in this time range, the test will fail.
Also, I was unable to use cy.intercept, because the file is not stored on a server. The download is provided by a blob.

cypress/e2e/shared/dashboardsIndex.test.ts
Dfldr: removal of the file has been moved into beforeEach

genehynson (approved):

genehynson:
@Dfldr Looks like the copy to clipboard test is failing in CI. I think it might be because we run the tests in headless mode in CI and `cy.task('getClipboard')` might not work in headless mode. I'll leave it to you to investigate further

PR #2749, opened 2021-09-30
fix: notification endpoints toggle on token creation

### LINKED ISSUE #2682: Tokens: NotificationEndpoint read/write toggle is broken
The read/write toggle on Custom Token creation for NotificationEndpoint is broken. This is probably happening because we change the resource name from `notificationEndpoint` to `NotificationEndpoint` for displaying but our key in the permissions object refers to `notifcationEndpoint`. We might need to fix this by changing the first letter to lowercase when toggling

Palakp41:
Closes #2682 

Reverting this change because lodash's `capitalize` function, capitalizes the first letter but lower cases everything else. This breaks the notification endpoints because they are being stored in our obj as `notificationEndpoints` and we are trying to access `notificaitonendpoints`. So this fix only capitalizes first letter for display of the resources on the CustomApiTokenOverlay

### DIFF
--- a/src/authorizations/components/redesigned/ResourceAccordion.tsx
+++ b/src/authorizations/components/redesigned/ResourceAccordion.tsx
@@ -1,6 +1,6 @@
 // Libraries
 import React, {Component} from 'react'
-import {isEmpty, capitalize} from 'lodash'
+import {isEmpty} from 'lodash'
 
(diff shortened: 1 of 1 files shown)


blegesse-git (approved):
LGTM!

PR #2752, opened 2021-09-30
feat(flagged): fetch all buckets from the API

### LINKED ISSUE #2628: Fetch All Buckets: Turn on fetching all buckets
### Background
To actually fetch all buckets, we'll need to make the specific API request and apply that to the query builder and notebooks experience

hoorayimhelping:
Closes #2628

flagged behind `fetchAllBuckets` feature flag

### DIFF
--- a/src/buckets/api/index.ts
+++ b/src/buckets/api/index.ts
@@ -1,0 +1,31 @@
+import {normalize} from 'normalizr'
+
+import {getBuckets} from 'src/client'
+import {fetchDemoDataBuckets} from 'src/cloud/apis/demodata'
+
+import {arrayOfBuckets} from 'src/schemas'
+import {Bucket, BucketEntities} from 'src/types'
+
+import {CLOUD} from 'src/shared/constants'
+import {BUCKET_LIMIT} from 'src/resources/constants'
+
+export const fetchAllBuckets = async (orgID: string, limit = BUCKET_LIMIT) => {
+  const resp = await getBuckets({
+    query: {orgID, limit},
+  })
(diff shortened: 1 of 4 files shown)


src/buckets/api/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import {getBuckets} from 'src/client'
+import {fetchDemoDataBuckets} from 'src/cloud/apis/demodata'
+
+import {arrayOfBuckets} from 'src/schemas'
+import {Bucket, BucketEntities} from 'src/types'
+
+import {CLOUD} from 'src/shared/constants'
+import {BUCKET_LIMIT} from 'src/resources/constants'
+
+export const fetchAllBuckets = async (orgID: string, limit = BUCKET_LIMIT) => {
```
hoorayimhelping, line 12: this method isn't a thunk, like the comment says, it doesn't belong in the thunks file

src/buckets/api/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (CLOUD) {
+    demoDataBuckets = await fetchDemoDataBuckets()
+  }
+
+  return normalize<Bucket, BucketEntities, string[]>(
+    [...resp.data.buckets, ...demoDataBuckets],
+    arrayOfBuckets
+  )
+}
```
hoorayimhelping, line 31: this method isn't a thunk, like the comment says, it doesn't belong in the thunks file

jrenee42 (approved):
looks good to me

PR #2757, opened 2021-09-30
test: update flag in telegrafs e2e test

TCL735:
Needed to make the pipeline pass again.


hoorayimhelping (approved):

TCL735:
A thought just came to me... flag removal for `telegrafUiRefresh` is going to involve some work with the e2e in this file. Either a lot of tests will be stripped out, or they need to be re-worked heavily. Possibly both.

PR #2766, opened 2021-10-01
Explicit buckets/add panels updated

### LINKED ISSUE #2510: in editing mode, add ability to add more measurement schemas
from the settings button->bucket form in the advanced section; have a way to add more measurement schemas

jrenee42:
Closes #2510 

includes validation: is the file valid, is the json valid, is the schema name valid? and notifications

does not do every single check, but does return any error messages from the backend
(minimum requirements for a valid schema, and name uniqueness within the schema names and within each measurement schema itself)

Screen Shot 2021-10-01 at 4 58 37 PM

https://user-images.githubusercontent.com/3900960/135685803-0c7c100e-fded-4286-b32e-e141215bec56.mov


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2766)

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const typeStrings = ['timestamp', 'tag', 'field']
+const dataTypeStrings = ['integer', 'float', 'boolean', 'string', 'unsigned']
+
+/**
+ * manually checking that the literal string union types are correct
+ *
+ * our types are generated, and changing it so there is an exported array that is *then* used
+ * for literal union type creation is out of scope.
+ * https://newbedev.com/checking-validity-of-string-literal-union-type-at-runtime
```
hoorayimhelping, line 11: we have [prior examples](https://github.com/influxdata/ui/blob/master/src/types/buckets.ts#L5) of extending generated types if we need to. Not sure if that's what this comment is saying...

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+        // not there:
+        if (item.type === 'field') {
+          dataTypePartIsValid = false
+        }
+      }
+      return hasName && hasType && dataTypePartIsValid
+    })
+
+    return validArray.reduce((prevVal, curVal) => prevVal && curVal, true)
+  }
```
hoorayimhelping, line 48: I'm trying to follow this functional logic and it seems like it can be simplified and made more efficient (currently it's n * n) by not using `map` and `reduce`. If this was a basic `for` loop, any time you encountered a bad value, you could just `return false` from the entire loop and short circuit needing to finish looping over the rest of the array and then looping over it again. 

```ts
if (Array.isArray(columns)) {
  for (let i = 0; i < columns.length; i++) {
    if (!('name' in item && typeof item.name === 'string') {
      return false
    }

    if (!('type' in item && typeStrings.includes(item.type))) {
      return false
    }

    if ('dataType' in item) {
      if (!dataTypeStrings.includes(item.dataType)) {
        return false
      }
    } else if (item.type === 'field') {
      return false
    }
  }
  return true  
}

return false
```

is this not equivalent in logic but simpler in terms of execution?

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
```diff
@@ -131,10 +156,58 @@ const UpdateBucketOverlay: FunctionComponent<Props> = ({
     }
   }
 
+  const isValid = () => {
+    // are there measurement schemas?
+    const haveSchemas =
+      !!newMeasurementSchemaRequests && !!newMeasurementSchemaRequests.length
```
hoorayimhelping, line 162: I think it might be more explicit and clear if instead of doing a casted boolean check,  you check if it's an array:

```ts
const haveSchemas =
      Array.isArray(newMeasurementSchemaRequests) && newMeasurementSchemaRequests.length

File: src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx [hunk 4]
```diff
@@ -131,10 +156,58 @@ const UpdateBucketOverlay: FunctionComponent<Props> = ({
...
+
+    if (!haveSchemas) {
+      // no schemas, nothing to validate, so everything is fine
+      return true
+    }
+    // if so, are they all valid?
+    const reducer = (previousValue, currentValue) =>
+      previousValue && currentValue.valid
+
+    const alltrue = newMeasurementSchemaRequests.reduce(reducer, true)
```
hoorayimhelping, line 172: the [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) method is made for this kind of calculation.

```ts
const allTrue = !newMeasurementSchemaRequests.some(schema => schema.valid === false)

File: src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx [hunk 5]
```diff
@@ -131,10 +156,58 @@ const UpdateBucketOverlay: FunctionComponent<Props> = ({
...
+
+        mSchemas.forEach(createRequest => {
+          event('bucket.schema.explicit.editing.uploadSchema')
+          onAddMeasurementSchemaToBucket(
+            bucketDraft.id,
+            bucketDraft.orgID,
+            bucketDraft.name,
+            createRequest
+          )
+        })
```
hoorayimhelping, line 206: this looks to be another `n*n` loop that doesn't need to loop twice. why not just call `forEach` on `newMeasurementSchemaRequests` and skip mapping over the whole array just to loop over it again?

hoorayimhelping (changes requested):
looks good so far, have a few questions on naming and looping

src/buckets/actions/thunks.ts
```diff
@@ -305,6 +313,37 @@ export const deleteBucketLabel = (bucketID: string, label: Label) => async (
   }
 }
 
+export const addSchemasToBucket = (
+  bucketID: string,
+  orgID: string,
+  bucketName: string,
+  schema: typeof MeasurementSchemaCreateRequest
```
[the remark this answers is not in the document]
jrenee42: this is just one schema.  the method is called for each measurementschema that is to be added
(already packaged up into requests)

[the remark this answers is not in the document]
jrenee42: you are right, i'll remove the 's' for it to be singular.

src/buckets/actions/thunks.ts
```diff
@@ -305,6 +313,37 @@ export const deleteBucketLabel = (bucketID: string, label: Label) => async (
...
+
+    dispatch(
+      notify(measurementSchemaAdditionSuccessful(bucketName, schema.name))
+    )
+  } catch (error) {
+    console.error(error)
+    const message = getErrorMessage(error)
+    dispatch(
+      notify(measurementSchemaAdditionFailed(bucketName, schema.name, message))
+    )
```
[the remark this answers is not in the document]
jrenee42: hmm....i agree that the error handling could be improved.  i copied the pattern from the updateBucket method.

to be consistent; changing them both would be necessary.

i vote to put this as a tech debt enhancement and leave it as it is (since this pattern is already handled by the code.)

src/buckets/components/createBucketForm/MeasurementSchemaSection.tsx
```diff
@@ -25,18 +40,187 @@ if (CLOUD) {
...
+  // overridden.  with the propagation, the error state only gets called once properly.
+  const handleUploadFile = (contents: string, fileName: string) => {
+    // keep swapping out the file, cancel out of the dialog, x out the adding line....etc
+
+    // do parsing here;  to check in the correct format:
+    let columns = null
+    if (contents) {
+      // parse them; if kosher; great!  if not, set errors and do not proceed
+      // don't need to wrap this in try/catch since the caller of this function is inside a try/catch
+      columns = JSON.parse(contents)
```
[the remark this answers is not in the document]
jrenee42: it shouldn't be wrapped in a try/catch; because the caller of this function is inside a try/catch (see the comment above it).
we want the caller of this function; the drag and drop component, to catch the error.  the code is SO much simpler that way.

that way, the miniFileDND component does the correct error decorations, and passes up to its parent (the addingPanel) that there is an error, since most of the errors stay within the MiniFileDnd component.

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+ * it gives us a relevant error message if the uniqueness fails.
+ *
+ * The backend also checks if the columns themselves are proper, there must be at least two columns:
+ *   1) a time field of time 'timestamp'
+ *   2) a field with type 'field'
+ *
+ *   not checking for those at this time on the frontend, letting the backend do the checking;
+ *   the code propagates up the error messages to show the user.
+ * */
+export const areColumnsKosher = columns => {
```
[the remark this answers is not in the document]
jrenee42: kosher means 'proper'; but sure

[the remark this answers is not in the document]
jrenee42: and it doesn't catch 'all' the invalidities, so renaming it to "areColumnsProper"

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
jrenee42: it's saying that in order to reuse the string arrays, the literal union types need to be generated differently.  so I am just retyping the string arrays

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
jrenee42: thanks!  I keep forgotting about the some method

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
jrenee42: unfortunately that will not work.  if you return something from within a for loop, it just keeps looping.  but using the every() method, similar to the some() method, is the way to go;  for example; this code which I just ran in my console:

```
let bar = [true, true, false, true, true, false]

let checkIt = (arr) => {
for(let i=0; i<arr.length; i++){
    if (!arr){
        console.log(`at index ${i}`)
        return false;
    }
}
return true;
}

checkIt(bar)

let checkIt2= (arr) => {
    let allGood = (x) => x;
    return arr.every(allGood)
}

console.log('checkit:', checkIt(bar))
console.log("checkit2:", checkIt2(bar))
```

the first console.log returns true, the second returns false (false is the correct answer here).
so I'll switch it to 'every'

jrenee42: keeping the variable names (hasName, hasType) for clarity

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
jrenee42: it's 2n (n loop twice), but yeah can put it in one loop

src/buckets/components/createBucketForm/BucketOverlayForm.tsx
```diff
@@ -45,36 +48,50 @@ interface Props {
...
   schemaType: 'implicit' | 'explicit'
+  newMeasurementSchemas: typeof MeasurementSchemaCreateRequest[]
 }
 
 export default class BucketOverlayForm extends PureComponent<Props> {
   constructor(props) {
     super(props)
 
     this.onChangeSchemaTypeInternal = this.onChangeSchemaTypeInternal.bind(this)
+    this.onUpdateSchemasInternal = this.onUpdateSchemasInternal.bind(this)
```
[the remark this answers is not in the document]
jrenee42: the class component and older-style code (it has get's) threw me for a loop and I fell back to my old way of binding.  fixed now :)

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
jrenee42: actually, switching to use 'every' because negative tests are to be avoided if possible.
but yeah, using optimized array methods

` const alltrue = newMeasurementSchemaRequests.every(schema => schema.valid)`

src/buckets/actions/thunks.ts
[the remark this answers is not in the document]
hoorayimhelping: >i vote to put this as a tech debt enhancement and leave it as it is (since this pattern is already handled by the code.)

fair enough. Would you mind writing up a ticket for that please?

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
hoorayimhelping: cool, I figured I misunderstood the comment

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
hoorayimhelping: >however, the 'every' function (similar to the some function) will work too. so will use that.

cool, works for me. Also I realized after I wrote this review that this is a `n + n` not `n*n` loop. Minor detail but I wanted to correct it.

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
hoorayimhelping: even better! I _knew_ there was a cleaner way to do it than negating `some`

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
hoorayimhelping: ha! I actually commented [the same thing before I saw this comment](https://github.com/influxdata/ui/pull/2766#discussion_r721748759)

src/buckets/components/createBucketForm/MiniFileDnd.tsx
```diff
@@ -158,6 +162,8 @@ export const MiniFileDnd: FC<Props> = ({
     inputEl.current.click()
   }
 
+  // maybe add a title or popover to the left....talk to david kim
```
hoorayimhelping, line 165: ✂️ this comment, but talk to David first :)

hoorayimhelping (approved):

src/buckets/components/createBucketForm/MiniFileDnd.tsx
jrenee42: he's on vacation for a week!  will talk to him when he gets back

src/buckets/actions/thunks.ts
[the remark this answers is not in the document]
jrenee42: done; https://app.zenhub.com/workspaces/applicationdumplings-60a80d872533670017a6b107/issues/influxdata/ui/2799

src/buckets/components/createBucketForm/MiniFileDnd.tsx
jrenee42: (i put a comment on the epic to talk to him; concerned that an email or a slack would get lost)

Issue #2756, opened 2021-09-30
Telegraf Refresh: Go to Telegraf pages after configuration


sjwang90:
**Background:** 
When a brand new user goes through the steps of creating a Telegraf configuration from the Data Sources pages, we want to give them guidance on how to View/Edit the configuration they just made. Since we are removing the "Previous" button, we want to give them guidance on how to continue editing their config if they want.

**Proposal:** 
When a user completes the "Create a new configuration" steps from the Sources page and clicks `Finish` (on the Test your Configuration page) the modal should close and take them to the Load Data > Telegraf page for them to view the completed Telegraf. 
- If the user exits out of the modal (clicks X) the modal should just close keeping whatever page it's still on

When a user completes the "Add to an existing configuration" steps from the Sources page and clicks `Finish` (on the Test your Configuration page) the modal should close and take them to the Load Data > Telegraf page for them to view the edited Telegraf. 
- If the user exits out of the modal (clicks X) the modal should just close keeping whatever page it's still on

Nothing needs to be done when "Creating a Configuration" from the Telegrafs page since the user will already be on that page.
 
----
Clicking "Finish"

<img width="1196" alt="Screen Shot 2021-09-30 at 2 58 00 PM" src="https://user-images.githubusercontent.com/32681364/135535797-60374de9-4532-48cb-a22a-61689d1f44eb.png">

Will bring you to view all your Telegrafs
![image](https://user-images.githubusercontent.com/32681364/135535649-eba50dbd-2746-43fa-a248-7880cbb9efc6.png)

PR #2768, opened 2021-10-01
feat: go to telegraf page 

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/2756

Go to telegraf page after creating the telegraf config.

https://user-images.githubusercontent.com/18511823/135695539-b24618dd-b1b4-4cfc-b970-eb7fe89458ce.mov


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2768)

src/dataLoaders/components/collectorsWizard/verify/VerifyCollectorsStep.tsx
```diff
@@ -63,6 +73,9 @@ export class VerifyCollectorStep extends PureComponent<Props> {
             text="Finish"
             type={ButtonType.Submit}
             testID="next"
+            onClick={() => {
+              this.props.history.push(`/orgs/${orgID}/load-data/telegrafs`)
+            }}
```
hoorayimhelping, line 78: let's pull this method out into an instance method since it doesn't rely on any data computed while rendering and it's inefficient to recreate identical anonymous functions every render.

src/dataLoaders/components/collectorsWizard/verify/VerifyCollectorsStep.tsx
```diff
@@ -84,4 +97,6 @@ const mstp = ({
   token,
 })
 
-export default connect<StateProps, {}, OwnProps>(mstp)(VerifyCollectorStep)
+const connector = connect<StateProps, {}, OwnProps>(mstp)
```
hoorayimhelping, line 100: I think this only needs `StateProps`, but I'm not sure.

does this work:
```ts
const connector = connect<StateProps>(mstp)

**Review by hoorayimhelping — CHANGES_REQUESTED** [review]
let me know how these suggestions work

File: src/dataLoaders/components/collectorsWizard/verify/VerifyCollectorsStep.tsx (hunk 1 above)
Comment (ChitlangeSahas) [reply src/dataLoaders/components/collectorsWizard/verify/VerifyCollectorsStep.tsx:78]: Ah right I keep forgetting about this! I'll do that 👍

File: src/dataLoaders/components/collectorsWizard/verify/VerifyCollectorsStep.tsx (hunk 2 above)
Comment (ChitlangeSahas) [reply src/dataLoaders/components/collectorsWizard/verify/VerifyCollectorsStep.tsx:100]: It seems to work, @TCL735 what do you think? the `OwnProps` has `CollectorStepProps` which seems to be relevant to the modal.

Comment (TCL735) [reply src/dataLoaders/components/collectorsWizard/verify/VerifyCollectorsStep.tsx:98]: If it works, then i would say this is just an optional TypeScript thing to add explicit types.

https://react-redux.js.org/using-react-redux/usage-with-typescript#typing-the-connect-higher-order-component

File: cypress/e2e/shared/telegrafPlugins.test.ts [hunk 3]
```diff
@@ -133,6 +133,13 @@ describe('Sources > Telegraf Plugins', () => {
       cy.getByTestID('next').click()
       cy.getByTestID('overlay--mask').should('not.exist')
 
+      // go back to the sources page
```
ChitlangeSahas, line 136: New flow goes to the telegraf page after we save a config so we need to be back at the sources page, for this test.

hoorayimhelping (approved):
lgtm, but wait for Timmy to look over it since he knows more about the feature than I do

TCL735 (approved):

TCL735:
I have sanity tested this locally as well. Looks good.

PR #2790, opened 2021-10-04
feat: add form to org cancellation

### LINKED ISSUE #2173: Notebooks: i don't want to have to delete a bunch of notebooks i never intended to keep
user:  I just spent a bit of time rummaging around in the cluster, and I’ve created a ton of random notebooks that I now need to delete. I didn’t intend to keep any of these notebooks long term - I was just exploring the data.

- as a notebooks user, i would like a way to indicate that i intend to keep this notebook around long term, otherwise, the notebook is removed when i navigate away.

When a user opens a new notebook (either by editing a resource, creating a new notebook, or linking to a new notebook from another part of the application), the user is made aware that the notebook will not be saved until they perform an explicit action to save the notebook. This explicit action could be a button, or it could be something like giving the notebook a non-default name. Let's explore these options and see which one flows better.

the notebook will still auto-save as it does today. the notebook can autosave before the user explicitly indicates they intend to keep this notebook around long term. Once they tell us to keep the notebook, auto-save continues as usual. If they never indicate they want to keep the notebook, it should not show up in the notebooks page.

if a user attempts to navigate away from a notebook before explicitly saving it, they should be presented with a warning that their progress will be lost and an option so save it.

fyi @barbaranelson

subirjolly:
Closes #2173 

Product Requirements

**Free Tier Proposed Options:**
**Field** - Dropdown - “What is your primary reason for leaving?” (Mandatory)
- It doesn't work for my use case
- I want to join my account to another organization
- I found an alternative product
    * Short text input field - “What is the alternative?” (not a required field)
- I want to sign up for a new account using a marketplace option
- Other reason (but don’t open up short text field like we used to)

Long Text field (not mandatory) [only display after they’ve chosen an option from the dropdown]
- How can we improve?

Also, send the gathered data along with meta information to `rudderstack`.

Kapture 2021-10-04 at 14 23 30


asalem1:
Is there a gif to reference?

subirjolly:
> Is there a gif to reference?

Added one!

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2790)

src/organizations/components/DeleteOrgOverlay.tsx
```diff
@@ -18,22 +18,80 @@ import {
...
+    setFormErrors(errors)
+
+    return !hasErrors
+  }
+
+  const sendDetailsToRudderstack = () => {
+    const payload = {
+      org: org.id,
+      tier: quartzMe?.accountType,
+      email: quartzMe?.email,
```
asalem1, line 77: getting the accountType here seems a bit redundant since the feature is only available for free customers right? If that's the case, we could also just get the email from the `user`, no need to add in another spot to get and set quartz data

src/organizations/components/DeleteOrgOverlay.tsx
```diff
@@ -18,22 +18,80 @@ import {
...
+
+    if (!suggestions) {
+      errors['suggestions'] = 'Please enter suggestions'
+      hasErrors = true
+    }
+
+    setFormErrors(errors)
+
+    return !hasErrors
+  }
```
asalem1, line 71: wondering if we should memoize this

asalem1 (commented):
Has design seen how long and skinny it looks with these additions? It might be valuable to give that form some breathing room

src/organizations/components/DeleteOrgOverlay.tsx
subirjolly: we do need this as we are going to add a similar thing for PAYG with the same payload

src/organizations/components/DeleteOrgOverlay.tsx
subirjolly: Memoizing this. This does change the behavior of the form. Instead of evaluating on click, it's now evaluating on every change. Thoughts?
Kapture 2021-10-04 at 18 30 52

subirjolly: i like this more. :)

src/organizations/components/DeleteOrgContext.tsx
```diff
@@ -0,0 +1,67 @@
...
+    setShortSuggestion(suggestion)
+  }
+
+  const changeShortSuggestionFlag = (flag: boolean) => {
+    setEnableShortSuggestion(flag)
+  }
+
+  const changeSuggestions = newSuggestions => {
+    setSuggestions(newSuggestions)
+  }
```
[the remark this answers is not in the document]
subirjolly: Simplified it instead

asalem1 (approved):
Approving this so that it's not a blocker when @juliajames12 gets us feedback on the overlay size

src/organizations/components/DeleteOrgReasonsForm.tsx
```diff
@@ -31,17 +17,25 @@ enum VariableItems {
   OTHER_REASON = 'Other reason',
 }
 
-function DeleteOrgReasonsForm({errors = {}}: ComponentProps) {
-  const [workingVariable, setWorkingVariable] = useState('')
+function DeleteOrgReasonsForm() {
+  const [workingVariable, setWorkingVariable] = useState('USE_CASE_DIFFERENT')
```
asalem1, line 21: nit: we have an enum, let's use that:

`useState<VariableItems>(VariableItems.USE_CASE_DIFFERENT)`

src/organizations/components/DeleteOrgReasonsForm.tsx
```diff
@@ -31,17 +17,25 @@ enum VariableItems {
...
     setSuggestions,
+    reason,
+    setReason,
   } = useContext(DeleteOrgContext)
 
+  useEffect(() => {
+    if (reason !== VariableItems[workingVariable]) {
+      setReason(VariableItems[workingVariable])
+    }
+  }, [reason, workingVariable])
```
asalem1, line 37: why is this in a useEffect and not in function that's invoked?

src/organizations/components/DeleteOrgReasonsForm.tsx
```diff
@@ -102,16 +96,15 @@ function DeleteOrgReasonsForm({errors = {}}: ComponentProps) {
...
         className="element improvement-suggestion--input"
       >
         <TextArea
           className="improvement-suggestions-input"
           onChange={e => setSuggestions(e.target.value)}
-          size={ComponentSize.Medium}
           testID="improvement-suggestions-input"
           value={suggestions}
           placeholder="How can we improve?"
+          style={{height: '120px'}}
```
asalem1, line 107: why not add this to the className?

src/organizations/components/DeleteOrgReasonsForm.tsx
subirjolly: > This isn't going to give me the key. Instead, it will give me the value of that enum

asalem1: Not sure what that means

src/organizations/components/DeleteOrgReasonsForm.tsx
subirjolly: Initially, this is not set. So, we need this to have a default value stored in the context. Unless we store these `VariableItems` in the context and provide a default one?

asalem1: What do you mean, you’re providing an initial value in the state above right?

src/organizations/components/DeleteOrgReasonsForm.tsx
subirjolly: Saw it used as this at some other place(s). Should I still move it to a css class?

src/organizations/components/DeleteOrgReasonsForm.tsx
asalem1: Like, the setting of state should be triggered based on a user interaction of selecting a property, not based on the byproduct of hook, unless I’m missing something

PR #2792, opened 2021-10-04
chore: misc notebooks things

### LINKED ISSUE #2787: Style fixes to resizer in Notebooks
We'd like to position the resizer component that's on several panels to be centered within the panel (I believe it used to be, but it was somehow moved). I'd also like the change the color to gray (#bec2cc) so it's less highlighted/visually distracting.

Current resizer is uncentered and white:

genehynson:
Closes #2787

- Centers the resizer on the panel
image

- When toggling the share menu, don't `queryAll`
- Add eventing for creating, deleting, and visiting shared notebooks
- Change the color of the share icon when the notebook is already shared
image

### DIFF
--- a/src/flows/components/PipeList.tsx
+++ b/src/flows/components/PipeList.tsx
@@ -1,9 +1,8 @@
 // Libraries
-import React, {FC, useContext, useEffect} from 'react'
+import React, {FC, useContext} from 'react'
 
 // Contexts
 import {FlowContext} from 'src/flows/context/flow.current'
-import {FlowQueryContext} from 'src/flows/context/flow.query'
 
 // Components
 import FlowPipe from 'src/flows/components/FlowPipe'
@@ -12,13 +11,6 @@ import InsertCellButton from 'src/flows/components/panel/InsertCellButton'
 
 const PipeList: FC = () => {
   const {flow} = useContext(FlowContext)
-  const {queryAll} = useContext(FlowQueryContext)
-
-  useEffect(() => {
-    if (flow.readOnly) {
-      queryAll()
-    }
-  }, [])
 
   if (!flow.data || !flow.data.allIDs.length) {
     return <EmptyPipeList />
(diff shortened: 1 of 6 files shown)


src/flows/components/header/TimeRangeDropdown.tsx
```diff
@@ -28,7 +29,10 @@ const TimeRangeDropdown: FC = () => {
     if (!flow.range) {
       return
     }
-    queryAll()
+    if (ref.current !== flow.range.lower) {
+      queryAll()
+      ref.current = flow.range.lower
+    }
```
genehynson, line 35: had to do this so we don't `queryAll` when this component is mounted (which happens when you toggle the share menu)

drdelambre: Refs like this, similar to a "hasRun" boolean, are a red flag that something else is wrong.

drdelambre: talked this out to myself on slack. the problem is that this component is in charge of running the query on change AND mount, which is causing issues when the component remounts as the header is toggled. instead of adding conditions to allow this component to manage that complexity, and since this useeffect has nothing to do with the render cycle, it makes sense to just remove the responsibility off of the view layer and closer to the data layer within the context.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2792)

drdelambre (approved):

src/flows/components/PipeList.tsx
```diff
@@ -12,13 +11,6 @@ import InsertCellButton from 'src/flows/components/panel/InsertCellButton'
 
 const PipeList: FC = () => {
   const {flow} = useContext(FlowContext)
-  const {queryAll} = useContext(FlowQueryContext)
-
-  useEffect(() => {
-    if (flow.readOnly) {
-      queryAll()
-    }
```
genehynson, line 20: moved this responsibility to flow.query

src/flows/components/header/TimeRangeDropdown.tsx
```diff
@@ -24,13 +22,6 @@ const TimeRangeDropdown: FC = () => {
     [updateOther]
   )
 
-  useEffect(() => {
-    if (!flow.range) {
-      return
-    }
-    queryAll()
-  }, [flow.range])
-
```
genehynson, line 33: moved this responsibility to flow.query

PR #2797, opened 2021-10-05
Remove Simple Table Feature Flag

### LINKED ISSUE #2708: Remove SimpleTable Feature Flag
Given the maturity of SimpleTable, we would like to remove the feature flag both for general feature flag cleanup and for some dependencies on the Dumplings team. 

## Acceptance Criteria

- Feature flag for SimpleTable is removed

joelfried:
Closes #2708


src/timeMachine/components/Vis.tsx
```diff
@@ -84,12 +84,12 @@ const TimeMachineVis: FC<Props> = ({
 
   // Handles deadman check edge case to allow non-numeric values
   const viewRawData =
-    isViewingRawData ||
+    isViewingRawData &&
```
joelfried, line 87: Taking this entire block out resulted in failed checks e2e tests, but leaving it as OR caused everything to show as the old raw table view.  This seemed simple enough and got tests passing locally but if there's a better way here I'm all for it.

genehynson: ugh this logic will forever haunt me. I think your change looks good tho

genehynson: ah wait. It might not actually work. I think this would only show the raw data view for numeric checks. If you run a query that returns string results, it should render the raw data view automatically. Does that still work? I'm also a little confused why we have to touch this logic at all just to remove this flag 🤔

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2797)

src/timeMachine/components/Vis.tsx
```diff
@@ -82,42 +75,6 @@ const TimeMachineVis: FC<Props> = ({
...
-                  <RawFluxDataTable
-                    parsedResults={parsedResults}
-                    width={width}
-                    height={height}
-                  />
-                )
-              }}
-            </AutoSizer>
-          </EmptyQueryView>
-        </ErrorBoundary>
```
asalem1, line 116: Wait, you're gonna need this back in here, except replace the RawFluxDataTable with SimpleTable I think

asalem1 (changes requested):

joelfried:
> we should also be deleting the RawFluxDataTable stuff here if we're defaulting to the SimpleTable

That was the state at commit https://github.com/influxdata/ui/commit/7c0fb82f9ebc355794e074912d23ca44d6aa3695 - and that pipeline failed (monitor-ci pipeline 5845 fwiw) like this:
image

I think this must be a bug in simple table and may well exist when the flag is turned off right now however the way the test was done before explicitly carved out the feature flag true state and thus we didn't know that crash happened until trying to remove the flag.

This was what I meant when I said in standup the other day that there may be a blocking bug that needs fixing before we can remove the feature flag.

asalem1:
> > we should also be deleting the RawFluxDataTable stuff here if we're defaulting to the SimpleTable
> 
> That was the state at commit [7c0fb82](https://github.com/influxdata/ui/commit/7c0fb82f9ebc355794e074912d23ca44d6aa3695) - and that pipeline failed (monitor-ci pipeline 5845 fwiw) like this: image
> 
> I think this must be a bug in simple table and may well exist when the flag is turned off right now however the way the test was done before explicitly carved out the feature flag true state and thus we didn't know that crash happened until trying to remove the flag.
> 
> This was what I meant when I said in standup the other day that there may be a blocking bug that needs fixing before we can remove the feature flag.

Right, I got that. I think it's a bug that you introduced here:

https://github.com/influxdata/ui/pull/2797#discussion_r725305098

src/timeMachine/components/Vis.tsx
joelfried: Ended up reverting this change.

src/timeMachine/components/Vis.tsx
```diff
@@ -101,17 +103,10 @@ const TimeMachineVis: FC<Props> = ({
...
-                    width={width}
-                    height={height}
-                  />
-                )
-              }}
-            </AutoSizer>
+            <SimpleTable
+              result={parsedResults}
+              properties={simpleTableProperties}
+            />
```
asalem1, line 109: You might need AutoSizer here, not sure

asalem1: Turns out we don't

asalem1 (commented):
We also have a `src/timeMachine/utils/rawFluxDataTable.ts` file with some utility functions that will no longer be used as a result of this. We should go ahead and check that file and removes the utility functions and tests that correspond to them since they'll no longer be of any relevance to us

asalem1 (approved):
🎉 check out all that red

PR #2810, opened 2021-10-05
feat: Explicit buckets- add schemas on create

### LINKED ISSUE #2505: add ability to add  measurement schemas to a bucket when it is created

jrenee42:
Closes #2505 

Can now add measurement schemas during bucket creation.  

the measurement schema section appears/disappears when you toggle the 'schematype' radiobutton.

this means that if you add data, then switch back to implicit, the section disappears (along with its data).

it does correct validation:  it only validates the data when the schema type is explicit.
and you don't have to add measurement schemas if you are creating an explicit bucket, measurement schemas can be added later.

added these two events:

bucket.schema.explicit.creation.uploadSchema: triggers for each schema being added
bucket.creation: when a bucket is created (this didn't already exist, so added it)

Screen Shot 2021-10-05 at 2 56 30 PM

Screen Shot 2021-10-05 at 2 56 39 PM

the validation, backend calls, etc are all the same as from when the bucket already exists.

the backend call needs the bucket id, so the bucket is created first; then the schema is added.  so the notifications show up for adding the schema (successfully or unsuccessfully) too.

### DIFF
--- a/src/buckets/actions/thunks.ts
+++ b/src/buckets/actions/thunks.ts
@@ -149,10 +149,7 @@ export const createBucketAndUpdate = (
     dispatch(addBucket(newBucket))
     dispatch(checkBucketLimits())
 
-    if (resp.data.schemaType === 'explicit' && schemas && schemas.length) {
-      // in case they choose explicit, add schemas, then change their mind back and
-      // change it to implicit; or if explicit and choose not to add schemas at this time.
-
+    if (schemas && schemas.length) {
       schemas.forEach(mschemaCreateRequest => {
         addMeasurementSchemaToBucketInternal(
           resp.data.id,
@@ -369,7 +366,7 @@ export const addSchemaToBucket = (
   bucketName: string,
   schema: typeof MeasurementSchemaCreateRequest
 ) => async (dispatch: Dispatch<Action>) => {
-  await addMeasurementSchemaToBucketInternal(
+  addMeasurementSchemaToBucketInternal(
     bucketID,
     schema,
     orgID,
--- a/src/buckets/actions/thunks.ts
+++ b/src/buckets/actions/thunks.ts
@@ -150,7 +150,7 @@ export const createBucketAndUpdate = (
     dispatch(addBucket(newBucket))
     dispatch(checkBucketLimits())
 
-    if (resp.data.schemaType === 'explicit') {
+    if (resp.data.schemaType === 'explicit' && schemas && schemas.length) {
       // in case they choose explicit, add schemas, then change their mind back and
       // change it to implicit
 
@@ -166,7 +166,9 @@ export const createBucketAndUpdate = (
         )
       })
     } else {
-      console.log(' implicit, not adding any schemas....')
+      console.log(
+        ' implicit (or no schemas present), not adding any schemas....'
+      )
     }
 
     // think....only call update after measurement schemas added (if there are schemas to add???)
--- a/src/buckets/actions/thunks.ts
+++ b/src/buckets/actions/thunks.ts
@@ -365,7 +365,7 @@ export const addSchemaToBucket = (
   orgID: string,
   bucketName: string,
   schema: typeof MeasurementSchemaCreateRequest
-) => async (dispatch: Dispatch<Action>) => {
+) => (dispatch: Dispatch<Action>) => {
   addMeasurementSchemaToBucketInternal(
     bucketID,
     schema,
--- a/src/buckets/actions/thunks.ts
+++ b/src/buckets/actions/thunks.ts
@@ -125,7 +125,8 @@ export const createBucket = (bucket: OwnBucket) => async (
 
 export const createBucketAndUpdate = (
   bucket: OwnBucket,
-  update: (bucket: Bucket) => void
+  update: (bucket: Bucket) => void,
+  schemas: typeof MeasurementSchemaCreateRequest[]
 ) => async (
   dispatch: Dispatch<Action | ReturnType<typeof checkBucketLimits>>,
   getState: GetState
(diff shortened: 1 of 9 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2810)

src/buckets/actions/thunks.ts
```diff
@@ -144,8 +145,25 @@ export const createBucketAndUpdate = (
       bucketSchema
     )
 
+    // just created the bucket, now add any schemas that might be there:
     dispatch(addBucket(newBucket))
     dispatch(checkBucketLimits())
+
+    if (resp.data.schemaType === 'explicit' && schemas && schemas.length) {
+      // in case they choose explicit, add schemas, then change their mind back and
+      // change it to implicit; or if explicit and choose not to add schemas at this time.
```
hoorayimhelping, line 154: I'm a little confused by this comment. what line does it apply to? Is it guarding against the situation described in the comment, or is it enabling the situation described in the comment?

src/buckets/actions/thunks.ts
```diff
@@ -344,6 +363,21 @@ export const addSchemaToBucket = (
   }
 }
 
+export const addSchemaToBucket = (
+  bucketID: string,
+  orgID: string,
+  bucketName: string,
+  schema: typeof MeasurementSchemaCreateRequest
+) => async (dispatch: Dispatch<Action>) => {
+  await addMeasurementSchemaToBucketInternal(
```
hoorayimhelping, line 372: does this need to be `await`ed if there's nothing below this code doing anything with the result?

is this equivalent to

```ts
addMeasurementSchemaToBucketInternal(
  bucketID,
  schema,
  orgID,
  bucketName,
  dispatch
).then()
```
?

src/buckets/components/createBucketForm/BucketOverlayForm.tsx
```diff
@@ -117,30 +117,36 @@ export default class BucketOverlayForm extends PureComponent<Props> {
       />
     )
 
-    const measurementSchemaSection =
-      schemaType === 'explicit' ? measurementSchemaComponent : null
+    const showMeasurementSchemaSection =
+      (isEditing && readOnlySchemaType === 'explicit') ||
+      schemaType === 'explicit'
+
+    const measurementSchemaSection = showMeasurementSchemaSection
```
hoorayimhelping, line 124: :+1: clean!

src/buckets/components/createBucketForm/CreateBucketForm.tsx
```diff
@@ -19,16 +19,21 @@ import {
   DEFAULT_RULES,
 } from 'src/buckets/reducers/createBucket'
 import {AppState, Bucket, RetentionRule} from 'src/types'
+import {event as influxEvent} from '../../../cloud/utils/reporting'
```
hoorayimhelping, line 22: is `event` already used here? any reason not to just leave this as `event`?

also, can you change this import to be absolute, not relative please?

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
```diff
@@ -82,3 +82,17 @@ export const isNameValid = name => {
...
+    Array.isArray(newMeasurementSchemaRequests) &&
+    newMeasurementSchemaRequests.length
+
+  if (!haveSchemas) {
+    // no schemas, nothing to validate, so everything is fine
+    // even if it is in explicit mode, can add schemas later
+    return true
+  }
+  // if so, are they all valid?
+  return newMeasurementSchemaRequests.every(schema => schema.valid)
```
hoorayimhelping, line 97: :+1: very comprehensible function

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
```diff
@@ -39,6 +39,7 @@ import {getBucketFailed} from 'src/shared/copy/notifications'
 import {OwnBucket} from 'src/types'
 import {CLOUD} from 'src/shared/constants'
 import {event} from '../../../cloud/utils/reporting'
+import {areNewSchemasValid} from './MeasurementSchemaUtils'
```
hoorayimhelping, line 42: can you turn these into absolute imports please?

hoorayimhelping (commented):
I have one or two requests and a couple of questions so I'm just going to leave this as a comment rather than approval or denial - let me know if the questions make sense

src/buckets/components/createBucketForm/CreateBucketForm.tsx
jrenee42: when i put it in as 'event' i got errors.  (there are form events and such on the page).  I can try again.

will find an absolute import and copy it.

src/buckets/components/createBucketForm/MeasurementSchemaUtils.ts
jrenee42: lol :)

src/buckets/components/createBucketForm/CreateBucketForm.tsx
jrenee42: fixed; the method that event(<event string>) gets called in has an argument/parameter called 'event'.  so fixed now.

src/buckets/actions/thunks.ts
jrenee42: hmm...I thought I needed the await when i did the refactor.  turns out I don't.  removing it.

PR #2816, opened 2021-10-06
test(cypress): notification history tests

karel-rehor:
Adds tests of the Notification history page accessible from a Notification Rule

* New group _notification history_
   * Shows and filters history 
   * Refreshes history
   * Switches between local and UTC time
* Support commands
   * In Utils timestamp calculator
   * In commands
      * More general replacement for createRule
      * Modified LP writing commands to make use of named buckets, e.g. can write notification records directly to `_monitoring` bucket.


cypress/e2e/shared/notificationRules.test.ts
```diff
@@ -292,4 +302,457 @@ describe('NotificationRules', () => {
...
+          rec.valField.key
+        }=${rec.valField.val.replace(/ /g, '\\ ')} ` + // end tags start fields
+        `_source_timestamp=${rec.sourceTimestamp},_status_timestamp=${
+          rec.statusTimestamp
+        },_message="${rec.message.replace(/ /g, ' ')}",dead="true"`
+
+      cy.writeLPData({lines: [lp], offset: offset, namedBucket: '_monitoring'})
+    }
+
+    beforeEach('Generate alert records', () => {
```
genehynson, line 407: I've seen flaky tests in the past from `beforeEach` functions that don't handle async functions & promises. I think we should refactor this to use cypress chaining (`cy.xyz().then(() => cy.abc().then(() => ...`) to make sure the beforeEach doesn't finish prematurely.

cypress/e2e/shared/notificationRules.test.ts
```diff
@@ -292,4 +302,457 @@ describe('NotificationRules', () => {
...
+        const stamps = timestamps
+          .toArray()
+          .map(n => new Date(n.innerText).getTime())
+        // assert sort desc by date
+        for (let i = 1; i < stamps.length; i++) {
+          expect(stamps[i - 1]).to.be.above(stamps[i])
+        }
+      })
+      // verify first row
+      cy.get('[class$=ScrollContainer] > div:nth-of-type(1)').within(() => {
```
genehynson, line 545: is there a reason we aren't using testIDs here? I find being dependent on CSS class names to be an issue because those change pretty often. We're working on a pretty major overhaul of Clockface which will probably change most CSS class names and this test will have to be completely re-written. If we use testIDs, we don't have that issue

cypress/e2e/shared/notificationRules.test.ts
```diff
@@ -292,4 +302,457 @@ describe('NotificationRules', () => {
...
+      })
+      // verify fourth row
+      cy.get('[class$=ScrollContainer] > div:nth-of-type(4)').within(() => {
+        cy.get('.level-table-field--warn').should('be.visible')
+      })
+      // check links...
+      //    ...To Check
+      cy.getByTestID('overlay').should('not.exist')
+      cy.get(
+        '[class$=ScrollContainer] > div:nth-of-type(4) .event-row--field:nth-of-type(3) > a'
```
genehynson, line 579: this is a great example where a testID should be used. It's really hard to tell what you're trying to click on here just by reading through the test code

cypress/e2e/shared/notificationRules.test.ts
```diff
@@ -292,4 +302,457 @@ describe('NotificationRules', () => {
...
+          '\\ '
+        )},` +
+        `_sent=${rec.sent},_source_measurement=${rec.source},_type=deadman,${
+          rec.valField.key
+        }=${rec.valField.val.replace(/ /g, '\\ ')} ` + // end tags start fields
+        `_source_timestamp=${rec.sourceTimestamp},_status_timestamp=${
+          rec.statusTimestamp
+        },_message="${rec.message.replace(/ /g, ' ')}",dead="true"`
+
+      cy.writeLPData({lines: [lp], offset: offset, namedBucket: '_monitoring'})
```
genehynson, line 404: change to `return cy.writeLPData(...)`. because cy.writeLPData is async, we need to return the chainable object here to handle the promise correctly

cypress/e2e/shared/notificationRules.test.ts
```diff
@@ -292,4 +302,457 @@ describe('NotificationRules', () => {
...
+                source: 'wumpus',
+                sent: 'true',
+                valField: {key: 'foo', val: 'bar'},
+                sourceTimestamp: calcNanoTimestamp('22m'),
+                statusTimestamp: calcNanoTimestamp('21m'),
+                message: 'This is a fake warning message',
+              },
+              '20m'
+            )
+            writeMonitoringRecord(
```
genehynson, line 501: this function is technically async because `cy.writeLPData` is async. So we need to chain off of it like `writeMonitoringRecord(...).then(() => ...`

cypress/e2e/shared/notificationRules.test.ts
```diff
@@ -292,4 +302,457 @@ describe('NotificationRules', () => {
...
+          const levels = rows
+            .find('.event-row--field:nth-of-type(2)')
+            .toArray()
+            .map(r => r.innerText)
+          expect(levels).to.deep.eq(['info', 'ok', 'info', 'crit', 'warn'])
+        })
+    })
+
+    // N.B. if local === utc then skip
+    if (new Date().getTimezoneOffset() !== 0) {
```
genehynson, line 718: instead of this we could modify `package.json` scripts to set the `TZ` variable like this: https://github.com/cypress-io/cypress/issues/1043#issuecomment-623293020

cypress/support/commands.ts
```diff
@@ -507,6 +513,9 @@ export const createTelegraf = (
     },
   })
 }
+/*
+// Did not find where this is used
+// Replaced with more general command below
```
genehynson, line 518: I'm fine with removing this function altogether, also couldn't find any usages

genehynson (changes requested):
left some comments. I appreciate the thoroughness of these tests! I think we should consider using testIDs instead of css class names though and need to be carful that we're handling cypress chainable promises correctly

cypress/e2e/shared/notificationRules.test.ts
karel-rehor: Agreed.  But will need to add the data-testid attributes myself.  See response to comment on line 545.

cypress/e2e/shared/notificationRules.test.ts
karel-rehor: I agree, data-testid values are preferable because theoretically less prone to change.  However, when investigating the Check Statuses page, I could find no `data-testid` attributes below `data-testid='page-contents'`.  Coming from QA, I'm hesitant to modify code in the application under test, in other words to insert `data-testid` attributes into the page.  However, if you have no problem with this, I'll go ahead and do it and then modify the test code accordingly.

cypress/e2e/shared/notificationRules.test.ts
karel-rehor: Generalized this to a `localInit()` method that returns CypressChainable.  It's at the start of the chain.  It handles a lot of local data.  It is probably too specific to warrant being generalized into a command.

cypress/support/commands.ts
karel-rehor: Done

cypress/e2e/shared/notificationRules.test.ts
genehynson: Cool, looks much better. You could move it back into a `beforeEach` to avoid having to call it in each test. As long as the code within the `beforeEach` returns a Cypress chainable object, it will wait for completion before starting the test. But this implementation is fine too.

cypress/e2e/shared/notificationRules.test.ts
genehynson: Yeah I think we should add the testID attributes if you don't mind. Most of the Clockface components come with generic testIDs for free. But if one is missing, adding it to the component here in the UI repo is the right move. Sorry to suggest so much refactoring...but I know this Clockface overhaul is coming soon and a lot of these CSS classes will change so it's important that we use testIDs for new tests

cypress/e2e/shared/notificationRules.test.ts
genehynson: Ah rats ok. We have a new ability in our pipeline to run tests 20x so we could do that to check for flakiness before merging.

genehynson:
@karel-rehor I did some work today to get the pipeline to set the TZ variable. I merged everything in so you should be good to try and run this again. I also think we should probably undo the changes to package.json so that developers can experience the tests running in their local timezone when running locally. I was able to set the TZ variable in the pipeline definition itself, outside of this configuration. Let me know if it doesn't work. Thanks!

genehynson (approved):
thanks again for this work! lgtm. I checked into our Cypress dashboard and it looks like it is not reporting these new tests as flaky (yay!) so you are good to merge without needing to run it 20x.

PR #2822, opened 2021-10-07
refactor: notification endpoints now self-register

genehynson:
This will allow us to quickly scale the number of endpoints we offer in notebooks in a maintainable way.


drdelambre (approved):
OH HELL YEAH

PR #2823, opened 2021-10-07
feat: create bucket during Telegraf create configuration mode

### LINKED ISSUE #2771: Telegraf Refresh: Create a new bucket from Telegrafs page
Add a "+Create a Bucket" option to the Create a Configuration workflow from the Telegrafs tab.

image

In our user research, users went immediately to the Telegrafs tab to create a config and didn't have a bucket created so they ended up going back to the buckets tab to create a bucket. I know we had to break out out the workflow when doing it for the sources page, if we need to do something similar just for create a bucket that's fine. 

Requirements:
- "+Create a Bucket" should be an option in the buckets selection dropdown and leads to the bucket creation overlay immediately
- In a Plugin's page, creating a new configuration should **_not_** default the user with a selected bucket upon seeing the first overlay the first time
- In a Plugin's page, user should see "Choose a bucket" and be forced to select a bucket before continuing to next step (this is similar to the Telegrafs tab)
- Creating a new bucket should select that bucket when the user is done with bucket creation

TCL735:
Closes #2771 

- "+ Create a Bucket" option added to the dropdown in Telegraf page when creating a new configuration
- user must select a bucket, there is no default selection, when adding a plugin to a new configuration
- user cannot continue to next step until a bucket is selected during configuration creation
- newly created buckets are sorted and become the user's selection during configuration

https://user-images.githubusercontent.com/10736577/136435336-d882eaf4-da40-4c50-ad96-f9fac03f7eab.mp4

### DIFF
--- a/cypress/e2e/shared/telegrafPlugins.test.ts
+++ b/cypress/e2e/shared/telegrafPlugins.test.ts
@@ -112,6 +112,12 @@ describe('Sources > Telegraf Plugins', () => {
         .click()
         .clear()
         .type(configurationName)
+      cy.getByTestID('plugin-create-configuration-options--select-bucket')
+        .should('be.visible')
+        .click()
+      cy.getByTestID('dropdown-item')
+        .eq(1) // "+ Create a Bucket" is 0 then existing buckets start at 1
+        .click()
 
       cy.getByTestID('plugin-create-configuration-continue-configuring').click()
       cy.getByTestID('plugin-create-configuration-customize').should(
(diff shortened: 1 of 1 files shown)


src/dataLoaders/components/BucketsDropdown.tsx
```diff
@@ -53,7 +53,7 @@ class BucketsDropdown extends PureComponent<Props> {
       return 'Choose a bucket'
     }
 
-    return buckets.find(bucket => bucket.id === selectedBucketID).name
+    return buckets.find(bucket => bucket.id === selectedBucketID)?.name ?? ''
```
TCL735, line 56: `Array.find` can return `undefined`, therefore it is common sense to guard against it with empty string.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2823)

PR #2839, opened 2021-10-08
tests: Explicit buckets e2e tests

ChitlangeSahas:
Closes #

### DIFF
--- a/cypress/e2e/cloud/buckets.test.ts
+++ b/cypress/e2e/cloud/buckets.test.ts
@@ -113,7 +113,7 @@ describe('Explicit Buckets', () => {
       cy.getByTestID('measurement-schema-section-parent').should('not.exist')
     })
   })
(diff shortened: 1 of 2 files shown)


cypress/e2e/cloud/buckets.test.ts
```diff
@@ -113,6 +113,110 @@ describe('Explicit Buckets', () => {
...
+      })
+    cy.getByTestID('accordion-header').click()
+
+    cy.getByTestID('measurement-schema-readOnly-panel-0')
+      .should('exist')
+      .within(() => {
+        cy.getByTestID('measurement-schema-name-0')
+          .contains('first schema file')
+          .should('exist')
+      })
```
jrenee42, line 218: looks great!  please add a section for downloading the file and checking the contents (the same code as in the first test you wrote, above)

jrenee42 (changes requested):

PR #2848, opened 2021-10-11
feat(tasks): when fetching all tasks, fetch basic version

hoorayimhelping:
Closes https://github.com/influxdata/idpe/issues/11742

Updated the tasks page to use the basic form when fetching all tasks. This makes it so that the flux query text isn't fetched. It's instead fetched when we load the individual task for the tasks run page, or to edit the task query.

Screen Shot 2021-10-11 at 12 28 07 PM

### DIFF
--- a/src/shared/containers/SetOrg.tsx
+++ b/src/shared/containers/SetOrg.tsx
@@ -12,6 +12,7 @@ import {
   TasksPagePaginated,
   TaskPage,
   TaskRunsPage,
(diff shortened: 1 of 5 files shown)


ChitlangeSahas (approved):

hoorayimhelping:
dependent on this PR (or a similar one) being merged in https://github.com/influxdata/openapi/pull/208

PR #2881, opened 2021-10-13
fix: run visualization queries for the panel you just added

drdelambre:
this was super confusing for users. why would have to run the notebook to see the alert panel options? "BECAUSE REASONS" didn't test well with our users, so lets just run the query for them.


asalem1:
> like how do you mean?

Like when you add a visualization after a script editor, or a column editor, or whatever else needs to be run in order to populate the results. Like should add a visualization panel inherently run the query?

PR #2884, opened 2021-10-14
feat: reorg notebook nav

drdelambre:
Screen Shot 2021-10-14 at 8 58 38 AM

paring down the options available to the user when making a notebook and renaming some nav items to make them more inline with the rest of the app's folksonomy.

some feedback was that it was rough for a user to add a visualization panel, then find the dropdown, then know what a simple table was... at the bottom of a list of other equally meaningless options, just to get introspection into their data. so i added a new panel called 'Table' that is just a simple table without any of the options to change it, and updated the templates to use it. this wont break existing visualizations, just dresses up the base functionality a little better for the use case.

also removed the "blank" template. people weren't using it

### DIFF
--- a/src/flows/pipes/MetricSelector/index.ts
+++ b/src/flows/pipes/MetricSelector/index.ts
@@ -11,6 +11,7 @@ export default register => {
     component: View,
     readOnlyComponent: ReadOnly,
     button: 'Metric Selector',
+    disabled: true,
     initial: {
       field: '',
       measurement: '',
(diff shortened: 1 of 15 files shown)


src/flows/pipes/MetricSelector/index.ts
```diff
@@ -11,6 +11,7 @@ export default register => {
     component: View,
     readOnlyComponent: ReadOnly,
     button: 'Metric Selector',
+    disabled: true,
```
asalem1, line 14: What's this do?

drdelambre: just hides it from the user without removing it from the code. it'll be back and i still like using it, so........

asalem1 (approved):

PR #2892, opened 2021-10-15
test: reduce the amount of lines in annotationsSetup

TCL735:
Let's reduce the resource intensiveness of the tests before trying to massively test for flakiness


TCL735:
The test failure here is legitimate.

You know what's funny? `cy.writeData()` actually has no effect on this test, other than to slow down the tests with a POST request. The data is set somewhere else and it's always a nearly vertical line.

Will look into this some more.

TCL735:
Ok, my mistake. I had two copies of the UI repository open in my code editor and was editing the wrong one. No wonder there was no effect... will try this again and report back.

Issue #2864, opened 2021-10-12
Flaky Test: creates a dashboard and downloads JSON - Dashboards


genehynson:
The test `creates a dashboard and downloads JSON - Dashboards` is flaky.

It appears to fail most often in the oss-e2e tests. 

Failure: 
```
AssertionError: Timed out retrying after 10000ms: `cy.readFile("cypress/downloads/dashboard.json")` failed because the file does not exist at the following path:
```

Examples:
- https://app.circleci.com/pipelines/github/influxdata/monitor-ci/5999/workflows/c5b53bdd-dc27-4d3b-8338-0aa518cd9432
- https://app.circleci.com/pipelines/github/influxdata/ui/8651/workflows/727c1a85-4ddd-4383-843f-08286c457f9f/jobs/55830

PR #2893, opened 2021-10-15
fix: flaky behavior of previously added test

Dfldr:
This PR fixes flaky behavior of the test added in https://github.com/influxdata/ui/pull/2748.
Removal of the downloaded file has been moved back to before from beforeEach.
I ran the test several times, both in Chrome and Firefox headless mode. All the runs were completed successfully.


cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -549,23 +545,29 @@ describe('Dashboards', () => {
...
+  it('creates a dashboard and downloads JSON', () => {
     cy.get('@org').then(({id: orgID}: Organization) => {
       cy.createDashboard(orgID).then(({body}) => {
         cy.fixture('routes').then(({orgs}) => {
           cy.visit(`${orgs}/${orgID}/dashboards/${body.id}`)
           cy.getByTestID('tree-nav')
           cy.getByTestID('nav-item-dashboards').click()
           cy.getByTestID('dashboard-card--name').click()
-          cy.getByTestID('page-title').type('dashboard') // dashboard name added to prevent failure due to downloading JSON with a different name
+          cy.getByTestID('page-title').type('dshboard') // dashboard name added to prevent failure due to downloading JSON with a different name
```
genehynson, line 563: the `dshboard` name caused issues last time for OSS, can we just name it `dashboard` please? See where I fixed it here: https://github.com/influxdata/ui/pull/2808

genehynson (changes requested):
nice - glad it was an easy fix. The CI ran the test 20x and it passed every time so I'm good with this change. Just fix the typo `dshboard` -> `dashboard`

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -552,9 +548,14 @@ describe('Dashboards', () => {
...
+  before(() =>
+    cy.exec('rm cypress/downloads/*', {
+      log: true,
+      failOnNonZeroExit: false,
+    })
+  )
 
-  // Skipping until https://github.com/influxdata/ui/issues/2864 is resolved by Bonitoo
-  it.skip('creates a dashboard and downloads JSON', () => {
+  it('creates a dashboard and downloads JSON', {browser: '!firefox'}, () => {
```
genehynson, line 558: why are we skipping this in firefox?

genehynson (changes requested):
I don't want to skip this test in firefox

cypress/e2e/shared/dashboardsIndex.test.ts
Dfldr: I tried to solve the flaky behavior of the test. I managed to fix the behavior in Chrome, but Firefox kept failing from time to time. I changed the readfile several times to try out possible fixes, but none worked. Because there is no other way to ensure that the file downloaded, I decided to skip this test in Firefox to ensure it's stability. UI itself shouldn't be the problem, since I was unable to make the problem appear manually.

Dfldr: So far as I can tell, there is no work around. Cy.readfile is the only option to check content of the cypress/downloads folder. Cy.fixture searches for files form the cypress/fixtures/ folder. 
The download process is handled internally in the browser (via blob). Because of that, checking the initiation of the download process is not an option eighter.
Also, there is not a notification which would indicate that the file downloaded successfully, like with the second export option (copy to clipboard).
In Firefox, the JSON file occasionally downloaded without it's content (NULL). Because of that, I added ".should('not.be.null')", to make the test fail due to a failure caused by the UI. Unfortunately, the problem persisted. When this error appears, cy.readfile is unable to read the mentioned file.
Similar problem appeared in Chrome, when I moved the removal of previously downloaded file to "beforeEach".
Putting it back in "before" fixed this behavior for Chrome.
I've been running the dashboardsIndex test for a few hours to provide you with some screenshots of the failure, but it hasn't crashed a single time. I will try to commit a version without excluding Firefox.

genehynson (approved):
thanks for getting this working!

Dfldr:
Closes: #2864

PR #2900, opened 2021-10-15
feat: Display error correctly when tokens fail to create

### LINKED ISSUE #1930: Token: Display error from Tokens API
Display the error (?) from the Tokens API

blegesse-w:
Closes #1930

Changes: 

- One way we are avoiding creation of broken or invalid tokens is by disabling the generate token button when user hasn’t selected any permissions. Right when user checks a permission from the accordion the status of the button gets set to active 
and in the case of user selecting a permission and then removing all selections, the `isPermSelected` and `isSubPermSelected` methods monitor the permissions object and deactivate the `generate` button. 

https://user-images.githubusercontent.com/66275100/137537497-a978790e-8246-4a15-a4db-c4b7a859a5df.mov

### DIFF
--- a/src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
+++ b/src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
@@ -226,7 +226,6 @@ const CustomApiTokenOverlay: FC<Props> = props => {
   }
 
   const generateToken = async () => {
-    
     const {onCreateAuthorization, orgID, showOverlay, orgName} = props
     const apiPermissions = formatApiPermissions(permissions, orgID, orgName)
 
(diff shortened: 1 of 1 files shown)


src/authorizations/actions/thunks.ts
```diff
@@ -121,8 +122,7 @@ export const createAuthorization = (auth: Authorization) => async (
     dispatch(notify(authorizationCreateSuccess()))
   } catch (error) {
     event('token.create.failure')
-    const message = error.data ? error.data.message : null
-    console.error(message)
+    const message = getErrorMessage(error)
```
Palakp41, line 125: Yay! So happy that we are using an existing util function :)

src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
```diff
@@ -151,9 +155,78 @@ const CustomApiTokenOverlay: FC<Props> = props => {
     ] = !permValue
 
     setPermissions(newPerm)
+    isSubPermSelected(newPerm)
   }
 
-  const generateToken = () => {
+  const isPermSelected = perm => {
```
Palakp41, line 161: Let's move this into utils folder if it's not too much work.

Palakp41 (approved):
Looks great!

PR #2906, opened 2021-10-15
feat: ephemeral notebook link

drdelambre:
Kapture 2021-10-15 at 14 39 02

`influx.set('boardWithFlows', true)`


drdelambre:
tasks were weird because you open the task run history from the list and not the task itself

PR #2914, opened 2021-10-18
feat: add zuora outage page and panel

subirjolly:
if `quartzZuoraDisabled` is enabled then we should see the Outage Panel or Page for Zuora fields.

Billing Page
image

Checkout Page
image

### DIFF
--- a/src/checkout/CheckoutPage.tsx
+++ b/src/checkout/CheckoutPage.tsx
@@ -12,8 +12,11 @@ const CheckoutV2: FC = () => {
     <CheckoutProvider>
       <>
         <SuccessOverlay />
-        {isFlagEnabled('quartzZuoraDisabled') && <ZuoraOutagePage />}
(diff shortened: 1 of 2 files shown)


src/checkout/CheckoutPage.tsx
```diff
@@ -4,13 +4,16 @@ import React, {FC} from 'react'
...
+import ZuoraOutagePage from 'src/shared/components/zuora/ZuoraOutagePage'
 
 const CheckoutV2: FC = () => {
   return (
     <CheckoutProvider>
       <>
         <SuccessOverlay />
-        <CheckoutForm />
+        {isFlagEnabled('quartzZuoraDisabled') && <ZuoraOutagePage />}
+        {!isFlagEnabled('quartzZuoraDisabled') && <CheckoutForm />}
```
genehynson, line 16: this could probably be converted into a ternary right? I think that would read better

genehynson (approved):
lgtm, just one minor nit

genehynson (approved):

Issue #2871, opened 2021-10-12
Remote CSV should be available under load-data UI


barbaranelson:
We have added a remote CSV (load CSV from URL) in Notebooks. It would be great if we could expose it in the load-data part of the UI too.

PR #2915, opened 2021-10-18
feat: easier sample data

### LINKED ISSUE #2606: Idea: Notebooks: Incorporate sample data sources into metric selector
We were trying to provide a starting dataset for users to explore with their notebook. the problem is that if i don't have buckets, or any data, and want to play with notebooks, you have to learn how to add the remote csv panel, delete the metric selector, then maybe write custom flux?

We could create an additional notebook template for starting with a sample data set, but then they don't get experience using the metric selector. 

Today, in the metric selector, you have the option to choose from a list of buckets:
image

The idea would be to rephrase this to say "Select a datasource" and group the dropdown into "Buckets" and "Sample Data Source" where they would be able to select USGS or NOAA (the same list from `sample.list()`).

drdelambre:
Kapture 2021-10-18 at 10 32 20

Closes #2606
Deals a little with the concerns from #2871

this treats sample data as a bucket in order to get the first time user up and running with as little investment possible.

### DIFF
--- a/src/flows/context/bucket.scoped.tsx
+++ b/src/flows/context/bucket.scoped.tsx
@@ -71,7 +71,37 @@ export const BucketProvider: FC<Props> = ({panel, children}) => {
               }
               return acc
             },
-            {system: [], user: []}
+            {
+              system: [],
+              user: [],
+              sample: [
+                {
+                  type: 'sample',
+                  name: 'Air Sensor Data',
+                  id: 'airSensor',
+                },
+                {
+                  type: 'sample',
+                  name: 'Coinbase bitcoin price',
+                  id: 'bitcoin',
+                },
+                {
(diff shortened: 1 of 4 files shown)


src/flows/context/bucket.scoped.tsx
```diff
@@ -71,7 +71,37 @@ export const BucketProvider: FC<Props> = ({panel, children}) => {
               }
               return acc
             },
-            {system: [], user: []}
+            {
+              system: [],
+              user: [],
+              sample: [
```
drdelambre, line 77: these exist inside of a query that we can use (`sample.list()`), but the result set doesn't include a human readable label for the dataset, so we're hard coding the list until it does

genehynson (approved):

PR #2934, opened 2021-10-19
fix: fix flaky zuora outage test

subirjolly:


TCL735:
That's seems a bit arbitrary. But ok. You can deal with it. 😄

TCL735 (approved):

PR #2943, opened 2021-10-19
docs: update readme links

genehynson:
No-op to get the pipeline rolling again


genehynson:
> TIL our public readme exposes info about private repos and internal docs. Super cool!

yeah...figured we can fix that later lol. Just trying to roll out a no-op 😅

PR #2951, opened 2021-10-20
test: update annotations tests to prevent detaching

TCL735:
From observing repeated flakiness, it is clear that Annotations tests have a common problem. Elements are detached from the DOM when a test command tries to execute. Let's guard against these.

- assert that certain elements are not detached
- execute certain commands only after the above
- use a new click target for annotations
- add a `testID` for the annotation message form to assert visibility

### DIFF
--- a/cypress/e2e/util/annotationsSetup.ts
+++ b/cypress/e2e/util/annotationsSetup.ts
@@ -1,6 +1,12 @@
 import {Organization} from '../../../src/types'
 import {points} from '../../support/commands'
 
+const ANNOTATION_TEXT = 'im a hippopotamus'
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
performance | 64 | :yellow_circle:
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2951)

ChitlangeSahas (approved):

Issue #2981, opened 2021-10-21
Alerts page not accessible from collapsed nav bar in Clockface 3 upgrade


ChitlangeSahas:
The `Alerts` page is not accessible from the collapsed menu in the new Clockface 3 upgrade.

PR #2982, opened 2021-10-21
fix: add alerts as sub menu item

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/2981

Just adds the Alerts as a sub menu item in the nav hierarchy. 

image


daviddkkim (approved):
lgtm

PR #2983, opened 2021-10-21
feat: batching state updates

drdelambre:
this moves the csv parser into a web worker to take parsing of large annotated CSVs out of the main UX thread
also caches query parsing as that wasm was generating a lot of overhead with so many mallocs
also the biggest change is that we now batch state updates for when queries are returned while another query is processing. react has a weird thing where it'll enqueue these state updates and run them separately, which causes a bunch of collisions

some memoization tasks snuck in here. they're net positive towards the perf, but only improved it by a little, where as batching halfed (or more) processing times. keeping them in here, but there's more that can be done to minimize the cost of updating the result / status states

### DIFF
--- a/src/flows/pipes/Table/index.ts
+++ b/src/flows/pipes/Table/index.ts
@@ -1,5 +1,7 @@
 import View from './view'
+import {parse} from '@influxdata/flux'
 import {parseQuery} from 'src/shared/contexts/query'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 import {SUPPORTED_VISUALIZATIONS} from 'src/visualization'
 
@@ -19,7 +21,9 @@ export default register => {
       }
 
       try {
-        const ast = parseQuery(query)
+        const ast = isFlagEnabled('fastFlows')
+          ? parseQuery(query)
+          : parse(query)
         if (!ast.body.length) {
           return ''
         }
--- a/src/flows/pipes/Visualization/index.ts
+++ b/src/flows/pipes/Visualization/index.ts
@@ -1,6 +1,8 @@
 import View from './view'
 import './style.scss'
+import {parse} from '@influxdata/flux'
 import {parseQuery} from 'src/shared/contexts/query'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 import {SUPPORTED_VISUALIZATIONS} from 'src/visualization'
 
@@ -19,7 +21,9 @@ export default register => {
       }
 
       try {
-        const ast = parseQuery(query)
+        const ast = isFlagEnabled('fastFlows')
+          ? parseQuery(query)
+          : parse(query)
         if (!ast.body.length) {
           return ''
         }


package.json
```diff
@@ -215,6 +216,7 @@
     "use-local-storage-state": "^9.0.2",
     "uuid": "^3.2.1",
     "webpack-bundle-analyzer": "^3.6.0",
+    "worker-plugin": "^5.0.1",
```
asalem1, line 219: is this just for IE support or what? Like, why not use the native `Worker` api?

src/flows/components/FlowPipe.tsx
```diff
@@ -10,23 +9,22 @@ export interface FlowPipeProps {
...
+        <Pipe
+          Context={props => {
+            const _props = {
+              ...props,
+              id,
+            }
+            return createElement(FlowPanel, _props)
+          }}
+        />
+      </PipeProvider>
```
asalem1, line 24: sanity check - why was the previous implementation being used and why the change?

src/flows/context/flow.query.tsx
```diff
@@ -374,20 +374,21 @@ export const FlowQueryProvider: FC = ({children}) => {
...
-              setStatuses({[stage.id]: RemoteDataState.Done})
               setResult(stage.id, response)
+              setStatuses({[stage.id]: RemoteDataState.Done})
             })
             .catch(e => {
-              setStatuses({[stage.id]: RemoteDataState.Error})
               setResult(stage.id, {
                 error: e.message,
               })
+              setStatuses({[stage.id]: RemoteDataState.Error})
```
asalem1, line 391: not sure about the level of effort, but should these be coupled to reduce some of that enqueueing you mentioned? Like consolidate the status and result into 1 result object with 2 properties?

src/flows/context/pipe.tsx
```diff
@@ -52,13 +54,13 @@ export const PipeProvider: FC<PipeContextProps> = ({id, children}) => {
...
+          results: result || {...DEFAULT_CONTEXT.results},
+          loading: status || RemoteDataState.NotStarted,
           readOnly: flow.readOnly,
         }}
       >
         {children}
       </PipeContext.Provider>
     )
-  }, [flow, id, results, statuses, children, updater])
+  }, [flow, id, result, status, children, updater])
```
asalem1, line 65: 👍🏽

src/flows/context/results.tsx
```diff
@@ -32,41 +33,85 @@ export const ResultsProvider: FC = ({children}) => {
   const [statuses, setStatuses] = useState<Hash<RemoteDataState>>({})
   const [results, setResults] = useState({...EMPTY_STATE})
 
+  const updates = useRef([])
```
asalem1, line 36: Never thought i'd live to see the day

src/flows/context/flow.query.tsx
```diff
@@ -374,20 +374,21 @@ export const FlowQueryProvider: FC = ({children}) => {
...
+
     Promise.all(
       map
         .filter(q => !!q.visual)
         .map(stage => {
           return query(stage.visual, stage.scope)
             .then(response => {
-              setStatuses({[stage.id]: RemoteDataState.Done})
               setResult(stage.id, response)
+              setStatuses({[stage.id]: RemoteDataState.Done})
```
drdelambre, line 385: this change gets rid of a weird flash of the old results would show before showing the new results

src/flows/context/results.tsx
```diff
@@ -32,41 +33,85 @@ export const ResultsProvider: FC = ({children}) => {
...
+    if (rChange) {
+      setResults({...results})
+    }
+
+    if (sChange) {
+      setStatuses({...statuses})
+    }
+
+    setInc(inc + 1)
+  }, [inc])
```
asalem1, line 73: this feels super wrong. I get why you're doing it, but it's inherently embedding another layer of complexity where we're now using state to manage state setting? That's at least what it feels like. I know why you're doing it, I just don't like it.

src/flows/context/results.tsx
drdelambre, line 36: this is similar to how y-redis handles decoupling two async event loops

src/flows/context/results.tsx
```diff
@@ -32,41 +33,85 @@ export const ResultsProvider: FC = ({children}) => {
...
+
   const value = {
     results,
     setResult: (id: string, result: Partial<FluxResult>) => {
-      results[id] = {
-        source: '',
-        parsed: null,
-        ...results[id],
-        ...result,
+      if (!isFlagEnabled('fastFlows')) {
```
asalem1, line 78: Omg this flag name with the `Run` notebook functionality

src/flows/pipes/MetricSelector/view.tsx
```diff
@@ -35,7 +35,7 @@ const DataSource: FC<PipeProp> = ({Context}) => {
   }
 
   return (
-    <BucketProvider panel={id}>
+    <BucketProvider>
```
asalem1, line 38: 👍🏽

src/flows/pipes/QueryBuilder/context.tsx
```diff
@@ -349,7 +355,9 @@ export const QueryBuilderProvider: FC = ({children}) => {
           })
           setCardMeta([...cardMeta])
         })
-        .catch(() => {})
+        .catch(e => {
+          console.error(e)
```
asalem1, line 359: and possible report an event?

src/flows/pipes/Table/index.ts
```diff
@@ -1,5 +1,5 @@
 import View from './view'
-import {parse} from '@influxdata/flux'
+import {parseQuery} from 'src/shared/contexts/query'
```
asalem1, line 2: why?

asalem1 (commented):
Submitting what I've got on my mind before checking out the webworker stuff

src/shared/contexts/csv.worker.ts
```diff
@@ -0,0 +1,358 @@
...
+    }
+  }
+}
+
+const ctx: Worker = self as any
+
+ctx.addEventListener('message', msg => {
+  const [idx, csv] = msg.data
+  ctx.postMessage([idx, fromFlux(csv)])
+})
```
asalem1, line 358: so what I'm gather from this file is that the webworker can't make use of existing files or functions, or dependencies right? That's the reason for the pasta? Nothing new here right?

src/shared/contexts/query.tsx
```diff
@@ -392,12 +407,126 @@ export const simplify = (text, vars: VariableMap = {}) => {
...
+  }
+})()
+
 export const QueryProvider: FC = ({children}) => {
   const buckets = useSelector((state: AppState) => getSortedBuckets(state))
   const bucketsLoadingState = useSelector((state: AppState) =>
     getStatus(state, ResourceType.Buckets)
   )
-  const [pending, setPending] = useState({} as CancelMap)
+  const pending = useRef({} as CancelMap)
```
asalem1, line 529: 🤯

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=2983)

src/flows/components/FlowPipe.tsx
drdelambre: i think it's more of an oversight of when we added the "pass the container as a prop" idea. turns out that oversight cost us 15ms over a large number of results coming in

src/flows/pipes/Table/index.ts
drdelambre: good catch. needs to be wrapped in a flag

src/shared/contexts/csv.worker.ts
drdelambre: nothing new, except we can't have anything more than a plain old javascript object being passed back and forth (had to remove simpeTable). we couldn't import a giraffe function directly because of how the exports are exposed in the npm module.

src/shared/contexts/csv.worker.ts
```diff
@@ -0,0 +1,358 @@
+import {csvParse, csvParseRows} from 'd3-dsv'
+
+interface FromFluxResult {
+  error?: Error
+  // The single parsed `Table`
+  table: Table
+
+  // The union of unique group keys from all input Flux tables
+  fluxGroupKeyUnion: string[]
+}
```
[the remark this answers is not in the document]
drdelambre: too many, and after all of this is migrated over... it'll just be the one in src/types/flows

PR #3086, opened 2021-10-29
chore: handle 503 from quartz billing

subirjolly:
When the zuora outage flag is turned on in quartz, it starts sending `503` which UI isn't able to handle and crashes on billing page. See image below:
image

___________________________

So, adding a default empty `BillingInfo` and now will be accessing nullable fields.

If the quartz outage flag isn't turned on and there's a problem with the billing route, this is what this page will look like:
image

___________________________

When we set `influx.set('quartzZuoraDisabled', true)`, we should be seeing this:
image

______________________________

Checkout Page looks normal when the flag isn't turned on and looks like this when the flag is turned on:
image

### DIFF
--- a/src/billing/components/BillingContactDisplay.tsx
+++ b/src/billing/components/BillingContactDisplay.tsx
@@ -13,29 +13,32 @@ const BillingContactDisplay: FC = () => {
     <Panel.Body size={ComponentSize.Large} testID="billing-contact">
       <Grid>
         <Grid.Row>
-          <BillingContactItem header="First Name" value={contact.firstName} />
-          <BillingContactItem header="Last Name" value={contact.lastName} />
+          <BillingContactItem header="First Name" value={contact?.firstName} />
+          <BillingContactItem header="Last Name" value={contact?.lastName} />
         </Grid.Row>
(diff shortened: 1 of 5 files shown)


asalem1 (changes requested):
We should do this for the Checkout page too

subirjolly:
> We should do this for the Checkout page too

Yes, made this PR a draft. I was doing that as well as adding screenshots for review description. Definitely a good call!

subirjolly:
> > We should do this for the Checkout page too
> 
> Yes, made this PR a draft. I was doing that as well as adding screenshots for review description. Definitely a good call!

Checkout page seems to be fine!

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3086)

asalem1 (approved):

genehynson (approved):

PR #3124, opened 2021-11-02
fix: fix stackdriver page

### LINKED ISSUE #3068: Stackdriver plugin page won't load on OSS
Note: This is not an issue with the _latest_ UI. It is an issue with the UI that was branched off of for the OSS 2.1 release. Running the latest UI with OSS actually works, so something must have changed between when OSS 2.1 branched off and now to make it work.

Trying to open the stackdriver telegraf input plugin page results in an error:
image

Clicking this results in...
image

wbaker85:
Closes #3068

This is a backport PR; it is backporting some changes from the `master` branch into `OSS-2.1` to fix a bug that has already been fixed on `master`. We don't want to branch off of the latest `master` at this point because for the 2.1 OSS release (imminent) we aren't planning to release with the clockface updates.

Primarily a clean cherry pick of https://github.com/influxdata/ui/pull/3039.  The file `src/buckets/components/context/lineProtocol.tsx` was updated based on https://github.com/influxdata/ui/pull/2919 to take care of a typescript check error.


src/buckets/components/context/lineProtocol.tsx
```diff
@@ -68,6 +68,20 @@ export const LineProtocolProvider: FC<Props> = React.memo(({children}) => {
     setWriteError('')
   }
 
+  /**
```
wbaker85, line 71: This file updated based on https://github.com/influxdata/ui/pull/2919 to clear up a typescript check error since we are using the most recent version of the `openapi` spec in `OSS-2.1`. All the other files are cherry-picked from https://github.com/influxdata/ui/pull/3039

src/writeData/components/ClientCodeCopyPage/InstallPackageHelper.tsx
```diff
@@ -13,14 +13,13 @@ import {
...
 import {MarkdownRenderer} from 'src/shared/components/views/MarkdownRenderer'
-import {Renderer} from 'react-markdown'
 
 // Styles
 import 'src/writeData/components/ClientCodeCopyPage/InstallPackageHelper.scss'
 
 interface Props {
   text: string
-  codeRenderer: Renderer<HTMLPreElement>
+  codeRenderer: any
```
hoorayimhelping, line 22: only concern here is that we're losing type information going from `Renderer` to `any`. Is there any way to avoid this?

hoorayimhelping (approved):
looks good to me, but a question on converting a type hint of `Renderer` to `any` since we're losing information.

src/writeData/components/ClientCodeCopyPage/InstallPackageHelper.tsx
wbaker85: Hrmm a valid concern for sure but I'm going to have to punt on it and defer to https://github.com/influxdata/ui/pull/3039 which is where this code is from, and matches the impl on `master`: https://github.com/influxdata/ui/blob/9edf248bb259dd4dfa49b9ebc7f9a65b84c18b02/src/writeData/components/ClientCodeCopyPage/InstallPackageHelper.tsx#L17-L20

I'd guess that there were some changes in the new version of `react-markdown` that made `Rendered` no longer compatible, but I can't say for sure. For now I'll keep this code consistent with `master`, and if it gets changed there in the future we'll pick it up when we create new OSS release branches off of `master`.

PR #3148, opened 2021-11-03
fix: export client css got borked

drdelambre:
they now look like this

Screen Shot 2021-11-03 at 2 26 16 PM


genehynson (approved):
mind the gap!

PR #3150, opened 2021-11-03
build: enhance OSS release scripts

wbaker85:
This is some automation for creating OSS releases. When a tag like OSS-2.2.0 (for example) is pushed, this workflow will create a changelog, build the UI for OSS, and create a draft release with the change in the release notes and the built artifacts attached. This should be far superior to doing those things locally in the process of creating a release of the UI for OSS.

The draft release looks like this:
image

TODO when/if this merges: Update the `TEAM_EDGE_BOT_GITHUB_TOKEN` to be a token for the `team_edge_influx` bot, and make sure the `team_edge_influx` bot has public repo access so that it can create releases.


.circleci/config.yml
```diff
@@ -170,6 +170,71 @@ jobs:
...
+          at: .
+      - run:
+          name: Install ghr
+          command: |
+            go install github.com/tcnksm/ghr@v0.14.0
+      - run:
+          name: Push draft release
+          command: |
+            ghr \
+              -t ${TEAM_EDGE_BOT_GITHUB_TOKEN} \
```
genehynson, line 228: any concerns about doing this in a public repo? someone could theoretically branch the UI repo and modify this file so that it creates a new draft release. Totally fine if it's not a concern, just checking

.circleci/config.yml
```diff
@@ -170,6 +170,71 @@ jobs:
       - store_artifacts:
           path: monitor-ci/test-artifacts/results/remocal
           destination: test_artifacts/results/remocal
+
+  oss-changelog:
+    docker:
+      - image: quay.io/influxdb/changelogger:0dbf37ba8a912460358d322dcad414b80ce074f2
```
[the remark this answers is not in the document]
wbaker85: I'm pinning it at that tag specifically in an attempt to prevent any breaking changes in the `changelogger` image from breaking this process unknowingly. I think the behavior of the image should be pretty well fixed at this point so there shouldn't be a need to frequently update this tag here.

.circleci/config.yml
wbaker85: Hmmm that's a good point. I don't think it would be a concern but I am a bit hesitant now to take the automation this far. I'll have to think about this a bit more - I won't be able to respond fully for a bit, so I'll convert this to a draft for now.

scripts/update-oss-master-release.bash
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
wbaker85, line 1: Renamed from the previous `scripts/publish-oss-release.bash` script.

wbaker85:
I decided to move the changelog generation logic into a separate script that can be used for creating named releases like `OSS-2.2.0` etc. This will be run locally instead of automatically in CI but should eliminate any potential for weirdness from branch the repo and pushing tags inadvertently triggering releases.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3150)

genehynson (approved):
lgtm

PR #3175, opened 2021-11-08
fix: generate correct name for orgs tokens

### LINKED ISSUE #3174: Tokens: generateDescription function in utils creates incorrect name for Orgs read/write token
## About the bug

**Steps to reproduce:**

1. Go to tokens page 
2. click on `Generate API Token` button
3. create a read/write token for `All Orgs` without adding description 

**Expected behavior:**
A token with `Read orgs Write orgs` description should be created 

**Actual behavior:**
A token with `Read orgs dev Write orgs dev` description is created 

**Visual Proof:**
 

## About your environment

**Environment info:**

**Config:**

**Logs:**

**Performance:**

```sh
iostat -xd 1 30 > iostat.txt
```

blegesse-git:
Closes #3174 

Added an if statement inside `generateDescription` function in utils that checks if received api permission array is an orgs permission. If true, the function will take org's unique data structure into account and generate its name accordingly.

Top token has correct description (fix) . Botton token is wrong (before fix) 
Screen Shot 2021-11-08 at 12 55 53 PM

### DIFF
--- a/src/authorizations/utils/permissions.ts
+++ b/src/authorizations/utils/permissions.ts
@@ -329,14 +329,20 @@ export const generateDescription = apiPermissions => {
     }
   } else {
     apiPermissions.forEach(perm => {
-      if (perm.resource.name) {
+      if (perm.resource.type === 'orgs') {
         generatedDescription += `${capitalize(perm.action)} ${
           perm.resource.type
-        } ${perm.resource.name} `
+        }  `
       } else {
-        generatedDescription += `${capitalize(perm.action)} ${
-          perm.resource.type
-        } `
+        if (perm.resource.name) {
+          generatedDescription += `${capitalize(perm.action)} ${
+            perm.resource.type
+          } ${perm.resource.name} `
+        } else {
+          generatedDescription += `${capitalize(perm.action)} ${
+            perm.resource.type
+          } `
+        }
       }
     })
   }


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3175)

src/authorizations/utils/permissions.ts
```diff
@@ -329,14 +329,20 @@ export const generateDescription = apiPermissions => {
...
+        }  `
       } else {
-        generatedDescription += `${capitalize(perm.action)} ${
-          perm.resource.type
-        } `
+        if (perm.resource.name) {
+          generatedDescription += `${capitalize(perm.action)} ${
+            perm.resource.type
+          } ${perm.resource.name} `
+        } else {
```
hoorayimhelping, line 341: it might be easier to read without all the extra brackets and syntax if instead of an `else` this is just set as a default case. I personally find it easier to follow, as it generally puts the default case at the top, then follows it with a special case triggered by a conditional

```ts
} else {
  generatedDescription += `${capitalize(perm.action)} ${perm.resource.type} `
  if (perm.resource.name) {
    generatedDescription += `${capitalize(perm.action)} ${
      perm.resource.type
    } ${perm.resource.name} `
  }
}
```

hoorayimhelping (approved):

src/authorizations/utils/permissions.ts
```diff
@@ -329,14 +329,20 @@ export const generateDescription = apiPermissions => {
...
+          generatedDescription += `${capitalize(perm.action)} ${
+            perm.resource.type
+          } `
```
hoorayimhelping, line 344: I originally left a comment on the this about structure, but after looking over it more clearly (like I should have done _before_ leaving the comment), I realized the comment was wrong because it didn't take into account the concatenation operator. My bad!

PR #3178, opened 2021-11-08
feat: add payg usage panel

subirjolly:
Add PAYG Credit Usage Panel

image

Specifically,
image

With enough credit
image

Less than 15 days remaining
image

### DIFF
--- a/src/me/components/Resources.tsx
+++ b/src/me/components/Resources.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FC, memo} from 'react'
+import React, {FC, memo, useMemo} from 'react'
 
 import {
   Panel,
@@ -22,8 +22,24 @@ import LogoutButton from 'src/me/components/LogoutButton'
 import DashboardsList from 'src/me/components/DashboardsList'
 import GetResources from 'src/resources/components/GetResources'
 import {ResourceType} from 'src/types'
+import UsagePanel from './UsagePanel'
+import UsageProvider from 'src/usage/context/usage'
+import {useSelector} from 'react-redux'
+import {getMe} from '../selectors'
 
 const ResourceLists: FC = () => {
+  const {quartzMe} = useSelector(getMe)
+  const creditDaysRemaining = useMemo(() => {
+    const startDate = new Date(quartzMe?.paygCreditStartDate)
+    const current = new Date()
+    const diffTime = Math.abs(current.getTime() - startDate.getTime())
+    const diffDays = Math.ceil(diffTime / (1000 * 60 * 60 * 24))
+
+    return 30 - diffDays
+  }, [quartzMe?.paygCreditStartDate])
+
+  const paygCreditEnabled = creditDaysRemaining > 0 && creditDaysRemaining <= 30
+
   return (
     <FlexBox
       direction={FlexDirection.Column}
@@ -63,6 +79,11 @@ const ResourceLists: FC = () => {
             </Panel.Body>
           </Panel>
         </>
+      )}
+      {isFlagEnabled('paygCheckoutCredit') && paygCreditEnabled && (
+        <UsageProvider>
+          <UsagePanel />
+        </UsageProvider>
       )}
       <Panel>
         <Panel.Footer>
--- a/src/me/components/Resources.tsx
+++ b/src/me/components/Resources.tsx
@@ -1,6 +1,6 @@
 // Libraries
 import React, {FC, memo, useMemo} from 'react'
-
+import {useSelector} from 'react-redux'
 import {
   Panel,
   FlexBox,
@@ -11,21 +11,22 @@ import {
   HeadingElement,
   FontWeight,
 } from '@influxdata/clockface'
-import VersionInfo from 'src/shared/components/VersionInfo'
 
 // Types
+import {ResourceType} from 'src/types'
 
-import DocSearchWidget from 'src/me/components/DocSearchWidget'
+// Utils
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import UsageProvider from 'src/usage/context/usage'
+import {getMe} from 'src/me/selectors'
 
+// Components
+import UsagePanel from 'src/me/components/UsagePanel'
+import DocSearchWidget from 'src/me/components/DocSearchWidget'
 import LogoutButton from 'src/me/components/LogoutButton'
 import DashboardsList from 'src/me/components/DashboardsList'
 import GetResources from 'src/resources/components/GetResources'
-import {ResourceType} from 'src/types'
-import UsagePanel from './UsagePanel'
-import UsageProvider from 'src/usage/context/usage'
-import {useSelector} from 'react-redux'
-import {getMe} from '../selectors'
+import VersionInfo from 'src/shared/components/VersionInfo'
 
 const ResourceLists: FC = () => {
   const {quartzMe} = useSelector(getMe)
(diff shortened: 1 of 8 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3178)

src/usage/context/usage.tsx
```diff
@@ -132,6 +158,58 @@ export const UsageProvider: FC<Props> = React.memo(({children}) => {
     handleGetBillingDate()
   }, [handleGetBillingDate])
 
+  const getComputedUsage = (vector_name: string, csvData: string) => {
+    const vectors = {
+      storage_gb: v => v * 0.02,
+      writes_mb: v => v * 0.002,
+      reads_gb: v => v * 0.09,
```
asalem1, line 165: what do these calculations mean?

src/usage/context/usage.tsx
```diff
@@ -132,6 +158,58 @@ export const UsageProvider: FC<Props> = React.memo(({children}) => {
...
+          })
+        )
+      })
+
+      Promise.all(promises).then(result => {
+        const usage: number = result
+          .reduce((a: number, b) => a + parseFloat(b), 0)
+          .toFixed(2)
+        setCreditUsage(usage)
+        setCreditUsageStatus(RemoteDataState.Done)
```
asalem1, line 203: just thinking out loud that these could probably be coupled together to reduce a rerender cycle

asalem1 (changes requested):
Biggest question is about the `30` here.

src/me/components/Resources.tsx
```diff
@@ -11,19 +11,36 @@ import {
...
 
 const ResourceLists: FC = () => {
+  const {quartzMe} = useSelector(getMe)
+  const creditDaysRemaining = useMemo(() => {
+    const startDate = new Date(quartzMe?.paygCreditStartDate)
+    const current = new Date()
+    const diffTime = Math.abs(current.getTime() - startDate.getTime())
+    const diffDays = Math.ceil(diffTime / (1000 * 60 * 60 * 24))
+
+    return 30 - diffDays
```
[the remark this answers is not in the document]
subirjolly: 30 Days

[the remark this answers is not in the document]
asalem1: 👍🏽 . K, can we make it a const somewhere and use that for these 30 day comparisons?

src/me/components/Resources.tsx
```diff
@@ -64,6 +69,9 @@ const ResourceLists: FC = () => {
           </Panel>
         </>
       )}
+      {isFlagEnabled('paygCheckoutCredit') && paygCreditEnabled && (
```
asalem1, line 72: Not sure if this is relevant or not, but we might need to make this cloud only, with a check for uiUnification since this’ll be weird in OSS and tools

src/me/components/UsagePanel.tsx
```diff
@@ -1,44 +1,80 @@
...
+// Constants
+import {PAYG_CREDIT_DAYS, PAYG_MAX_CREDIT} from 'src/shared/constants'
+
+// Types
+import {RemoteDataState} from 'src/types'
 
 const UsagePanel: FC = () => {
+  const {creditUsage, creditDaysRemaining} = useContext(UsageContext)
+
+  const getUsageBars = () => {
```
asalem1, line 27: Why not just in-line this or make it a Functional Component?

src/me/containers/MePage.tsx
```diff
@@ -130,7 +131,9 @@ export class MePage extends PureComponent<Props> {
                 </FlexBox>
               </Grid.Column>
               <Grid.Column widthSM={Columns.Four} widthMD={Columns.Three}>
-                <Resources />
+                <UsageProvider>
```
asalem1, line 134: Actually we might need to perform the uiUnification check here

src/usage/context/usage.tsx
```diff
@@ -132,6 +165,75 @@ export const UsageProvider: FC<Props> = React.memo(({children}) => {
...
+          status: RemoteDataState.Done,
+        })
+      })
+    } catch (error) {
+      setCreditUsage({
+        ...creditUsage,
+        status: RemoteDataState.Done,
+      })
+    }
+  }, [])
```
asalem1, line 232: We may need  to either pass in the credit usage or list it as a dependency

subirjolly: good catch!

src/me/components/Resources.tsx
subirjolly: don't think this is relevant as the payg credit start date won't be available and thus this should never be visible.

asalem1: It's relevant. In order to determine the `paygCreditEnabled` value we make a request to quartz for the `paygCreditStartDate` right? `new Date(undefined) === Invalid Date`. What I'm getting at is that we need to make sure that we only derive that data for Cloud users that aren't tools.

src/me/components/Resources.tsx
```diff
@@ -64,6 +69,9 @@ const ResourceLists: FC = () => {
           </Panel>
         </>
       )}
+      {isFlagEnabled('uiUnificationFlag') &&
+        isFlagEnabled('paygCheckoutCredit') &&
+        paygCreditEnabled && <UsagePanel />}
```
asalem1, line 74: something about this makes me feel weird but 🤷🏽

src/usage/context/usage.tsx
```diff
@@ -132,6 +165,76 @@ export const UsageProvider: FC<Props> = React.memo(({children}) => {
...
+    const tot = values.reduce((a, b) => a + b, 0)
+    return vectors[vector_name](tot)
+  }
+
+  const handleGetCreditUsage = useCallback(() => {
+    try {
+      setCreditUsage({
+        ...creditUsage,
+        status: RemoteDataState.Loading,
+      })
```
asalem1, line 197:
```suggestion
      setCreditUsage(prev => ({
        ...prev,
        status: RemoteDataState.Loading,
      }))
```

src/usage/context/usage.tsx
```diff
@@ -132,6 +165,76 @@ export const UsageProvider: FC<Props> = React.memo(({children}) => {
...
+        setCreditUsage({
+          amount,
+          status: RemoteDataState.Done,
+        })
+      })
+    } catch (error) {
+      setCreditUsage({
+        ...creditUsage,
+        status: RemoteDataState.Done,
+      })
```
asalem1, line 230:
```suggestion
      setCreditUsage(prev => ({
        ...prev,
        status: RemoteDataState.Done,
      }))
```

asalem1 (approved):
Sorry, just realizing that the `creditUsage` in the `useCallback` was only being used to set the initial state for loading. We can bypass that by using the `prev` value of the setState. Besides that I think we're all set?

src/me/components/Resources.tsx
subirjolly: me too!

asalem1: Like, it feels weird to me that this particular component may or may not have access to context based on the conditional wrapper above. We should probably default `paygCreditEnabled` to false, then check to see if we're in a place where `uiUnification` is on, then set the value based on that, but I'm not sure if that's kosher in terms of react hooks being called within `ifs`

PR #3181, opened 2021-11-08
test: another take at the flaky flows test

genehynson:
This changes the test to wait on the `PATCH` request after editing the notebook before navigating away from the page. Passed 20x locally with firefox

### DIFF
--- a/cypress/e2e/shared/flows.test.ts
+++ b/cypress/e2e/shared/flows.test.ts
@@ -200,6 +200,7 @@ describe('Flows', () => {
       .first()
       .click()
     cy.getByTestID('time-machine-submit-button').should('be.visible')
+    cy.intercept('PATCH', '**/notebooks/*').as('updateNotebook')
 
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):
thanks for jumping on this!

TCL735:
@genehynson isn't it better to wait on something like a toast notification? something tangible that is already part of the UI rather than artificial that only happens in the tests themselves

genehynson:
> @genehynson isn't it better to wait on something like a toast notification? something tangible that is already part of the UI rather than artificial that only happens in the tests themselves

Totally agree. However, we don't actually have any UI feature that notifies the user when the notebook has been saved. All changes are persisted to local storage as the user interacts with the UI - and then eventually are sent to the backend in a batched update request. It's more or less like Google doc's "auto save". The test runs so fast that sometimes it navigates away from the page before the update request happens in the background. So I'm just waiting on the update request to happen before we navigate away

PR #3193, opened 2021-11-09
feat: faster notebook loading

drdelambre:
Turns out that we read `generateMap` a lot more than we update it, and that actions which update it are usually discrete where as the reads tend to be all bunched together. By switching where the map gets built, i took a nearly 8s blocking task on load for my "ux task usage" notebook and brought it down to ~550ms. typing performance seems to be the same.


src/flows/context/flow.list.tsx
```diff
@@ -307,6 +307,11 @@ export const FlowListProvider: FC = ({children}) => {
   )
 
   const migrate = async () => {
+    const localFlows = Object.keys(flows).filter(id => id.includes('local'))
+    if (!localFlows.length) {
```
drdelambre, line 311: this was causing a double rerender for all notebooks because it pushes a new state into localstorage right off the bat

src/flows/context/flow.query.tsx
```diff
@@ -199,7 +207,19 @@ export const FlowQueryProvider: FC = ({children}) => {
...
+    _map.current = stages
+  }
+
+  useEffect(() => {
+    _generateMap()
+  }, [flow])
+
+  const generateMap = (_withSideEffects?: boolean): Stage[] => {
+    if (!_map.current) {
+      _generateMap()
```
asalem1, line 219: would we want to return this?

drdelambre: we do

asalem1 (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3193)

src/flows/context/flow.list.tsx
genehynson: honestly we could get rid of all this migrate garbage, right? that was there to help migrate notebooks to the API way back in the day when notebooksd was first born

drdelambre: idk. it did come in handy when we were doing some migrations right? i think it might all break (if it doesn't already) when @asalem1 starts dropping in the "and you need a websocket channel" stuff, so lets do it then

PR #3194, opened 2021-11-09
fix: fix error boundary issue where View not rerendering

### LINKED ISSUE #2935: notebooks: Band graph crashed and then couldn't recover the panel
I created a notebook and selected the Band visualization, and the panel crashed with an error: 
react-dom.production.min.js:198 TypeError: Cannot read properties of undefined (reading 'xs')
    at band.ts:57
    at Array.forEach (<anonymous>)
    at ys (band.ts:56)
    at ws (band.ts:455)
    at memoize-one.esm.js:34
    at e.getSpec (PlotEnv.ts:303)
    at uy (PlotResizer.tsx:41)
    at Wo (react-dom.production.min.js:150)
    at ki (react-dom.production.min.js:178)
    at ja (react-dom.production.min.js:232)

I then changed from the Band visualization to a Graph visualization, but I still get the internal error in the panel. I had to delete the panel and add a new panel with a Graph visualization for it to recover.  It would be good if changing the visualization would cause the panel to recover.

subirjolly:
Closes #2935

BEFORE
Before

AFTER
After

### DIFF
--- a/src/shared/components/ErrorBoundary.tsx
+++ b/src/shared/components/ErrorBoundary.tsx
@@ -38,6 +38,16 @@ class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
     })
   }
 
+  public shouldComponentUpdate() {
+    if (this.state.error) {
+      this.setState(() => {
+        return {error: null}
+      })
+    }
+
+    return true
+  }
+
   public render() {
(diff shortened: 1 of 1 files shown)


asalem1 (approved):

asalem1:
Nice find!

PR #3197, opened 2021-11-10
fix: fix rudderstack feature flag name

subirjolly:
There was a typo in RudderStack Reporting flag. The actual flag name is: `rudderstackReporting` but it was typoed as `rudderStackReporting`..


asalem1 (approved):
👏🏽

PR #3204, opened 2021-11-10
chore: turn off to bucket panel flag

### LINKED ISSUE #3157: Notebooks: Remove Export to Bucket
Export to bucket is more of an advanced functionality that causes side-effects in the UI. We do not find that it is something that we want to support currently. 

Given the above, we would like to remove the export to bucket functionality for now.

genehynson:
Closes #3157 

turns off the to bucket panel flag. Existing panels will not be removed, but new ones cannot be added.

image


drdelambre (approved):
please double check at standup that we want to do this (disable our 4th most popular panel) before merging

PR #3207, opened 2021-11-10
feat: add checkout query param

subirjolly:
We need to distinguish between a user upgrading and a user just got created and wants to become a PAYG user. This query param will help us distinguish that.

Update: The initial solution was to add a query param which was fallible and suspetible to unauthorized usage by existing users. Instead we have switched to using `GoogleOptimizeExperiment` component which allows the usage of experiments transparently in our app.


src/checkout/context/checkout.tsx
```diff
@@ -1,7 +1,8 @@
 // Libraries
-import React, {FC, useCallback, useEffect, useState} from 'react'
+import qs from 'qs'
+import React, {FC, useCallback, useEffect, useMemo, useState} from 'react'
 import {useDispatch, useSelector} from 'react-redux'
-import {useHistory} from 'react-router-dom'
+import {useHistory, useParams} from 'react-router-dom'
```
asalem1, line 5: looks like `useParams` isn't being used?

subirjolly: done!

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3207)

src/checkout/context/checkout.tsx
```diff
@@ -249,6 +252,13 @@ export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
     return errs.length
   }
 
+  const isPaygCreditActive = useMemo(() => {
+    const query = new URLSearchParams(window.location.search)
+
+    const hasSignup = query.get('signup')
```
asalem1, line 258: do we wanna validate this against anything? Right now:

`?signup=foo`, `?signup=bar`, & `?signup=influxdataisthebest` would all be considered a signup

asalem1 (commented):
Is there a gif I can reference? Looks kosher, just wanna make sure it doesn't return

src/checkout/context/checkout.tsx
subirjolly: Any value for signup should be okay. If you think we should be strict, then i can add the validation.

asalem1: It should also be noted that anyone can insert a query parameter and mock this situation out, so not sure how solid of a solution this is.

asalem1 (approved):
Offlined my concerns about the entire design of this. The UI has no way of actually knowing about a user's stage of setup. In addition, there is no way of guarding against actors trying to take advantage of this solution. Those concerns don't seem to be of concern for the development of this feature. I'd love to see some metrics associated with how often it's getting hit and by whom so we can verify that the target audience is in fact the one that's making use of this

src/checkout/constants.ts
```diff
@@ -0,0 +1 @@
+export const PAYG_CREDIT_EXPERIMENT_ID = 'OWMWJ2XjTsW6Wny5AroCCg'
```
asalem1, line 1: Just gonna bring this up since it seems like it should be mentioned that we shouldn't have this in the UI as a string, that it should be an environment variables, and that it should live in the same place as all our other env variables

asalem1 (changes requested):

src/checkout/constants.ts
subirjolly: so, should i use it using `process.env`?

asalem1 (approved):

PR #3231, opened 2021-11-15
fix: individual buckets and telegrafs names should render when Buckets and Telegrafs accordion expands 

### LINKED ISSUE #3226: Tokens: individual buckets and telegrafs inside ResourceAccordion only show up after user exists and re-open Custom Api Token overlay
## About the bug

**Steps to reproduce:**

1. Go to API Tokens page in the UI
2. Click on `Generate API Token` > click on `Buckets` or `Telegrafs`

**Expected behavior:**
Buckets drop down should display All Buckets and Individual Bucket Names 

**Actual behavior:**
Only All Buckets shows up. For Individual Bucket Names to show up inside the accordion, user has to exit and re-open the overlay again.

**Visual Proof:**
https://images.zenhubusercontent.com/279622976/97ad77d0-4597-4325-b540-44893835ad9c/screen_recording_2021_11_12_at_12_34_38_pm.mov

## About your environment

**Environment info:**

**Config:**

**Logs:**

**Performance:**

```sh
iostat -xd 1 30 > iostat.txt
```

blegesse-git:
Closes #3226 

Background: 
Inside CustomAPITokenOverlay component, inside mstp, we fetch all telegrafs and buckets from redux store and populate fetched data inside bucketPermissions and telegrafPermissions object. 

Then after the component mounts, the useEffect hook constructs a new permissions object (perms) using our state props (bucketPermissions and telegrafPermissions) and updates the permission state. 

Before Fix: 
everytime props.bucketPermissions and props.telegrafPermissions change, the useEffect function check if the state has telegraf and bucket property. if false -> it updates the state (permissions). which causes components to re-render. 
But just because our permissions object has telegraf and buckets properties, it doesn't mean they have sublevel permission values. This is what's causing the bug mention in #3226. 

https://user-images.githubusercontent.com/66275100/141816576-947501c8-df04-41d5-b928-a13573522c9c.mov

Fix: 

instead of updating the state inside the useEffect every time the component mounts, we are now only updating state when props.bucketPermissions has sublevelPermissions value. this brings down the amount of times we update state from 3 times to only once.

https://user-images.githubusercontent.com/66275100/141816768-02faee27-a0e2-4059-a643-01d87e297570.mov

### DIFF
--- a/src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
+++ b/src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
@@ -97,9 +97,9 @@ const CustomApiTokenOverlay: FC<Props> = props => {
       })
       setPermissions(perms)
     }
-    // Each time remoteDataState changes, the useEffect hook will be called. 
-    // BUT, code inside the hook won't run until remoteDataState is 'Done'. 
-    // Only then will props.bucketPermissions.sublevelPermissions will have value. 
+    // Each time remoteDataState changes, the useEffect hook will be called.
+    // BUT, code inside the hook won't run until remoteDataState is 'Done'.
+    // Only then will props.bucketPermissions.sublevelPermissions will have value.
     // Consequently, we update the permissions state.
   }, [props.remoteDataState])
 
--- a/src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
+++ b/src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
@@ -97,6 +97,10 @@ const CustomApiTokenOverlay: FC<Props> = props => {
       })
       setPermissions(perms)
     }
+    // Each time remoteDataState changes, the useEffect hook will be called. 
+    // BUT, code inside the hook won't run until remoteDataState is 'Done'. 
+    // Only then will props.bucketPermissions.sublevelPermissions will have value. 
+    // Consequently, we update the permissions state.
   }, [props.remoteDataState])
 
   const handleDismiss = () => {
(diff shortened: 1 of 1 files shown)


src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
```diff
@@ -81,23 +82,22 @@ const CustomApiTokenOverlay: FC<Props> = props => {
...
+          perms[resource] = props.telegrafPermissions
+        } else if (resource === 'buckets') {
+          perms[resource] = props.bucketPermissions
+        } else {
+          perms[resource] = {read: false, write: false}
+        }
+      })
+      setPermissions(perms)
+    }
+  }, [props.remoteDataState])
```
jrenee42, line 100: what if the permissions change?  (or that doesn't happen, or if the remote data state changes then the permissions change?)

jrenee42 (changes requested):
see clarifying question, besides that code looks good!

src/authorizations/components/redesigned/CustomApiTokenOverlay.tsx
blegesse-git: so the reason why i added props.remoteDataState inside the dependency array is because while buckets and telegrafs are being fetched, the remoteDataState will start at `NotStarted` then `Loading` and then once we get the data, it updates to `Done`. Each time remoteDataState changes, the useEffect hook will be called. BUT, none of the code inside the hook will run until remoteDataState is Done. Only then will props.bucketPermissions.sublevelPermissions will have value aka won't be empty. In which case, we can finally update the permissions state. I hope that answers your questions

jrenee42: it does!  please add that as a comment to the code; then it'll be good to go :)

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3231)

jrenee42 (approved):

PR #3237, opened 2021-11-15
feat: added metrics for user events 

### LINKED ISSUE #3132: Tokens: add metrics for user events
Capture metrics in new Tokens UI:
- [x] when editing a token is a success
- [x] when editing a token is a failure
- [x] when active/inactive happens (word differently)
- [x] when description on resource card is edited

- [x] when deleting a token is a success
- [x] when deleting a token is a failure

- [x] when cloning a token is a success
- [x] when cloning a token is a failure

- [x] when adding an all-access token is a success
- [x] when adding a custom api token is a success
- [x] when adding a token is a failure
- [x] fire each event when overlays are opened

blegesse-git:
Closes #3132 

Added metrics for user events for the following: 
- when editing a token is a success
- when editing a token is a failure
-  when active/inactive happens (word differently)
-  when description on resource card is edited
-  when deleting a token is a success
-  when deleting a token is a failure
-  when cloning a token is a success
-  when cloning a token is a failure
-  when adding an all-access token is a success
-  when adding a custom api token is a success
-  when adding a token is a failure
-  fire each event when overlays are opened

### DIFF
--- a/src/authorizations/components/redesigned/AllAccessTokenOverlay.tsx
+++ b/src/authorizations/components/redesigned/AllAccessTokenOverlay.tsx
@@ -27,6 +27,7 @@ import {OverlayContext} from 'src/overlays/components/OverlayController'
 
 // Utils
 import {allAccessPermissions} from 'src/authorizations/utils/permissions'
+import {event} from 'src/cloud/utils/reporting'
 
 // Selectors
 import {getOrg} from 'src/organizations/selectors'
@@ -54,6 +55,7 @@ const AllAccessTokenOverlay: FC<OwnProps> = props => {
     }
     dispatch(createAuthorization(token))
     handleDismiss()
+    event('token.allAccess.success', {meID, name: description})
     dispatch(showOverlay('access-token', null, () => dismissOverlay()))
   }
 
--- a/src/authorizations/components/redesigned/AllAccessTokenOverlay.tsx
+++ b/src/authorizations/components/redesigned/AllAccessTokenOverlay.tsx
@@ -55,7 +55,7 @@ const AllAccessTokenOverlay: FC<OwnProps> = props => {
     }
     dispatch(createAuthorization(token))
     handleDismiss()
-    event('token.allAccess.success', {meID, name: description})
+    event('token.allAccess.create.success', {meID, name: description})
     dispatch(showOverlay('access-token', null, () => dismissOverlay()))
   }
 
--- a/src/authorizations/components/redesigned/EditTokenOverlay.tsx
+++ b/src/authorizations/components/redesigned/EditTokenOverlay.tsx
@@ -100,21 +100,15 @@ const EditTokenOverlay: FC<Props> = props => {
     }
   }
 
-  const onSave = async () => {
+  const onSave = () => {
     const {auth, updateAuthorization} = props
 
-    try {
-      await updateAuthorization({
-        ...auth,
-        description: description,
-        status: togglestatus ? 'active' : 'inactive',
-      })
-      event('token.edit.success', {id: auth.id, name: description})
-      handleDismiss()
-    } catch {
-      event('token.edit.failure', {id: auth.id})
-    }
-    
+    updateAuthorization({
+      ...auth,
+      description: description,
+      status: togglestatus ? 'active' : 'inactive',
+    })
+    handleDismiss()
   }
 
   return (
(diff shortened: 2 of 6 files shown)


src/authorizations/components/redesigned/AllAccessTokenOverlay.tsx
```diff
@@ -54,6 +55,7 @@ const AllAccessTokenOverlay: FC<OwnProps> = props => {
     }
     dispatch(createAuthorization(token))
     handleDismiss()
+    event('token.allAccess.success', {meID, name: description})
```
hoorayimhelping, line 58:
```suggestion
    event('token.allAccess.create.success', {meID, name: description})
```

src/authorizations/components/redesigned/GenerateTokenDropdown.tsx
```diff
@@ -29,14 +32,17 @@ const GenerateTokenDropdown: FC<ReduxProps & GenerateTokenProps> = ({
 
   const handleAllAccess = () => {
     showOverlay('add-master-token', null, dismissOverlay)
+    event('all access overlay accessed')
```
hoorayimhelping, line 35:
```suggestion
    event('generate_token_dropdown.all_access_overlay.opened')
```

Alex does his own thing with a less structured style of event name (which is why you might see some event names like 'all access overlay accessed'), but Kristina and I would prefer a more organized style, that has a more categorical approach because it's easier to search for, and it provides some kind of organization.

You don't have to use the exact suggestion, but it should follow a general pattern like:

`component.action.outcome`

src/authorizations/components/redesigned/EditTokenOverlay.tsx
```diff
@@ -71,15 +72,20 @@ const EditTokenOverlay: FC<Props> = props => {
...
-      status: togglestatus ? 'active' : 'inactive',
-    })
-    handleDismiss()
+    try {
+      await updateAuthorization({
+        ...auth,
+        description: description,
+        status: togglestatus ? 'active' : 'inactive',
+      })
+      event('token.edit.success', {id: auth.id, name: description})
```
Palakp41, line 84: Could this event be handled in the thunk instead of here? That way we can get rid of the try/catch

src/authorizations/components/redesigned/TokenRow.tsx
```diff
@@ -136,28 +137,37 @@ class TokensRow extends PureComponent<Props> {
...
-      ...this.props.auth,
-      description: incrementCloneName(allTokenDescriptions, description),
-    })
-    this.props.showOverlay('access-token', null, () => dismissOverlay())
+    try {
+      await this.props.createAuthorization({
+        ...this.props.auth,
+        description: incrementCloneName(allTokenDescriptions, description),
+      })
+      event('token.clone.success', {id: this.props.auth.id, name: description})
```
Palakp41, line 151: This event being here makes sense, because clone and create use the same thunk so we want to differentiate between the two somehow.

Palakp41 (approved):
Everything looks good, made one small suggestion 👍

src/authorizations/components/redesigned/GenerateTokenDropdown.tsx
```diff
@@ -29,14 +32,17 @@ const GenerateTokenDropdown: FC<ReduxProps & GenerateTokenProps> = ({
...
   }
 
   const handleCustomApi = async () => {
     try {
       await getAllResources()
       showOverlay('add-custom-token', null, dismissOverlay)
+      event('custom API overlay accessed')
     } catch (e) {
       dispatch(notify(getResourcesTokensFailure()))
+      event('custom API overlay failed to open')
```
hoorayimhelping, line 45: same comment here for these event names

src/authorizations/components/redesigned/TokenRow.tsx
```diff
@@ -136,28 +137,37 @@ class TokensRow extends PureComponent<Props> {
...
+      this.props.showOverlay('access-token', null, () => dismissOverlay())
+    } catch {
+      event('token.clone.failure', {id: this.props.auth.id, name: description})
+    }
   }
 
   private handleClickDescription = () => {
     const {onClickDescription, auth} = this.props
+    event('edit token overlay accessed')
```
hoorayimhelping, line 160: same here for the event name

hoorayimhelping (changes requested):
great work on your first metrics PR! have a couple minor suggestions / edits

src/authorizations/components/redesigned/GenerateTokenDropdown.tsx
blegesse-git: that totally makes sense. I will update the event names accordingly

hoorayimhelping (approved):
🎉

PR #3238, opened 2021-11-15
chore: update Giraffe

TCL735:
Update Giraffe

- updates dependencies with versions suggested by dependabot for security
- adds the "repository" field in Giraffe's package.json


hoorayimhelping (approved):
looked good when I tested it locally

PR #3250, opened 2021-11-16
fix: remove maps as a vis option for OSS

### LINKED ISSUE #3247: Map vis shows up as an option for OSS
https://github.com/influxdata/ui/pull/2420 removed the feature flags for Maps and replaced them with the "cloud only" logic that one would expect to prevent them from showing up on OSS. However, they are still showing up on OSS, both in notebooks and in data explorer options:

image

image

wbaker85:
Closes #3247

Makes it so that maps don't show up as a visualization option in OSS.


genehynson (approved):

hoorayimhelping (approved):
don't you just not love to don't not hate double negatives!

subirjolly (approved):

PR #3277, opened 2021-11-17
feat: add checkout query param2

subirjolly:
We need a checkout code on top of the experiment id.

Use Case. if the user is in the experiment and they choose not to upgrade and then later decide to upgrade. they shouldn’t see the credit.

### DIFF
--- a/src/checkout/context/checkout.tsx
+++ b/src/checkout/context/checkout.tsx
@@ -87,6 +87,8 @@ export const CheckoutContext = React.createContext<CheckoutContextType>(
   DEFAULT_CONTEXT
 )
 
+const CHECKOUT_PARAM_CODE = '757vMPJiCMdTFkO4'
+
 export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
   const dispatch = useDispatch()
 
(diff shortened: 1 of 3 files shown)


src/checkout/context/checkout.tsx
```diff
@@ -87,6 +87,8 @@ export const CheckoutContext = React.createContext<CheckoutContextType>(
   DEFAULT_CONTEXT
 )
 
+const CHECKOUT_PARAM_CODE = '757vMPJiCMdTFkO4'
```
asalem1, line 90: ummm ... isn't this the `env` you set earlier?

subirjolly: No. This is just a random string Nick generated.

asalem1: k. Do we want to expose it on the UI?

asalem1: As in, is there a way to validate it in case Nick wants to generate a different one that doesn't actively involve making a PR to the UI repo?

subirjolly: This code is only useful if you have the experiment ID set already. If you don't then finding this code is completely useless. So, if someone has experiment enabled and still knows the query param and this code then they should be able to upgrade and it is intentional.

asalem1: 🤷🏽

asalem1 (approved):
🤷🏽

PR #3281, opened 2021-11-18
fix: reenables the run on add ability

### LINKED ISSUE #3232: notebook add panel regression
when adding a panel, it should run the flux for that panel. it seems like generate map isn't including the most recently added panel when it runs the queries. this makes notebooks, especially panels the require data to display like the alert panel, hard to use

drdelambre:
Closes #3232

re-enables the interface reloading on panel add / removal / cloning.

i hate this solution, because the issue feels like it's covering up a problem that is rooted more in a misrepresented responsibility, where the flow.query context needs to recalculate if it's parent context (flow.current) had manipulated a nested object, but i couldn't quite get myself to saying that the generateMap function should be owned by the flow.current context. it might very well make more sense there, but while i hem and haw over that, we can at least unbreak the interface for our users.

### DIFF
--- a/src/flows/context/flow.query.tsx
+++ b/src/flows/context/flow.query.tsx
@@ -217,14 +217,8 @@ export const FlowQueryProvider: FC = ({children}) => {
     // _map.current and a rerender that updates the panel view components within the same render cycle
     // leading to a panel on the list without a corresponding map entry
     const forceUpdate =
-      (flow?.data?.allIDs ?? [])
-        .slice(0)
-        .sort((a, b) => a.localeCompare(b))
-        .join(' ') !==
-      (_map.current ?? [])
-        .map(m => m.id)
-        .sort((a, b) => a.localeCompare(b))
-        .join(' ')
+      (flow?.data?.allIDs ?? []).join(' ') !==
+      (_map.current ?? []).map(m => m.id).join(' ')
 
     if (forceUpdate) {
       _generateMap()
--- a/src/flows/context/flow.query.tsx
+++ b/src/flows/context/flow.query.tsx
@@ -213,10 +213,20 @@ export const FlowQueryProvider: FC = ({children}) => {
   }, [flow])
 
   const generateMap = (): Stage[] => {
-    if (
-      !_map.current ||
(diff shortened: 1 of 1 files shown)


src/flows/context/flow.query.tsx
```diff
@@ -213,10 +213,20 @@ export const FlowQueryProvider: FC = ({children}) => {
...
 
   const generateMap = (): Stage[] => {
-    if (
-      !_map.current ||
-      (!_map.current.length && (flow?.data?.allIDs ?? []).length)
-    ) {
+    // this is to get around an issue where a panel is added, which triggers the useEffect that updates
+    // _map.current and a rerender that updates the panel view components within the same render cycle
+    // leading to a panel on the list without a corresponding map entry
+    const forceUpdate =
```
genehynson, line 219: Could we memoize this so we don't repeat it every time if the panels are unchanged? Seems like it could introduce a lot of overhead of there are a lot of panels

drdelambre: thankfully the overhead of the sort and string match was minimal up to the 100 panels / notebook i was testing at (i think the prod median is 4, but our numbers are real shoddy around that), especially compared to the parse -> format_from_js_file cycle this is replacing. storing the result sets for visualization is what starts crippling the ui at that point (running out of ram), so i didn't go much further. tying into the pr text, this is to shore us up for a bit while looking for a better solution. like.. there's an implicit coupling across abstractions between the list of panels, and the queries they generate, which suggests that the abstraction is wrong and by solving that, we should be able to remove the matching all together. but while typing all this out... we dont need to sort the list of ids, because it's already a sorted list, so the overhead then just becomes a string match.  lemme push that up

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3281)

genehynson (approved):

PR #3285, opened 2021-11-19
chore: remove demo data

### LINKED ISSUE #3061: Remove Demo Data Feature in UI
**Let's remove the demo data feature based on team feedback.** 

History

We either need to remove demo data altogether, or add whatever special logic is in the data explorer to notebooks.

Current state: If you go to your buckets list, and select the Website Monitoring Bucket, and have the option to go to notebooks when a bucket is selected, you get taken to the notebooks page, but the Website Monitoring Bucket isn't available in the list of buckets. If you go to data explorer, it is.

I know we do some special trick to get this bucket to show up, as it isn't owned by the org.  We either need to do the same trick in notebooks, or decide that now is the time to remove the demo data feature altogether, as we have other sample data we can use.

### LINKED ISSUE #3286: Notebooks: Remove Bird Migration Demo Data from Query Builder
The Bird Migration demo data doesn't load when selected since it's historical data and the default time range is for the past 3 hours. There is no indicator that the user would have to change the time range or what to change it to in order to see the data, so it looks like broken demo data set or bug.

To avoid overcomplicating the conditions to solve for a historical demo data set, we want to remove this from the Query Builder. 

We'll need to inform the docs team that we're removing this so they can update the demo data docs.

genehynson:
Closes  #3286, #3061

Removes demodata from the UI to encourage people to use sample data instead.

Also removes the bird migration sample data from notebooks cuz it sucks.

image

Screen Shot 2021-11-19 at 11 12 53 AM


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3285)

asalem1 (approved):
Super cathartic, super jealous. You're gonna want to remove the API definition from OpenAPI, and remove the mechanics of it from k8s and potentially IDPE, although I don't know of the specifics well enough to be able to tell you what needs to be done. Maybe @wojciechka or @mark-rushakoff might have some context on this

genehynson:
> Super cathartic, super jealous. You're gonna want to remove the API definition from OpenAPI, and remove the mechanics of it from k8s and potentially IDPE, although I don't know of the specifics well enough to be able to tell you what needs to be done. Maybe @wojciechka or @mark-rushakoff might have some context on this

Ooo thanks for the reminder! I'll make sure to leave the issue open and will add a comment to remove those

PR #3288, opened 2021-11-19
fix: sorting buckets across UI

### LINKED ISSUE #3080: standardize alphabetical order in dropdown lists within the UI
When we show a drop-down list of buckets, we are inconsistent in how we display them across the UI, e.g.:
The buckets page shows the two system buckets, followed by user buckets (in case-insensitive alphabetical order).
The notebooks page shows user buckets, followed by system. Within the user buckets it is case-insensitive alphabetical order.
The data explorer page shows upper case user buckets, followed by system buckets, followed by lower case user buckets.

Can we agree on one format and apply it consistently across the UI?  My vote would be user buckets (case-insensitive alphabetical order) followed by system buckets.

blegesse-git:
Closes #3080 

Before fix: 
buckets tab -> we have system buckets listed first and then the rest follow in alphabetical order.
In data explorer -> we have them all listed alphabetically. System & user buckets
lastly, in notebooks -> we have 3 categories. user, system, and sample. and buckets are listed in each category in alphabetical order

After fix: 
All three places in the UI where buckets are listed are now ordered alphabetically by bucket type (user buckets followed by system bucket) 
**Bucket Tab:** 
https://user-images.githubusercontent.com/66275100/142675310-c68da837-1354-4f5c-859b-3c44bba73680.mov
**Data Explorer:** 
https://user-images.githubusercontent.com/66275100/142675360-94ab9774-73bf-48fe-a2d2-6cae04e95987.mov
**Notebooks:** 
https://user-images.githubusercontent.com/66275100/142675445-9251966b-4a4c-4247-ad8f-5250a3b8e85e.mov

### DIFF
--- a/src/authorizations/components/redesigned/IndividualAccordionBody.tsx
+++ b/src/authorizations/components/redesigned/IndividualAccordionBody.tsx
@@ -28,15 +28,14 @@ export const IndividualAccordionBody: FC<Props> = props => {
   const {resourceName, permissions, onToggle, title, disabled} = props
   let sortedPermissions
 
-  if(title === 'Individual Bucket Names') {
-      // re-order buckets: user buckets first followed by system buckets
-    const systemBuckets = permissions.splice(0,2) 
+  if (title === 'Individual Bucket Names') {
+    // re-order buckets: user buckets first followed by system buckets
+    const systemBuckets = permissions.splice(0, 2)
     sortedPermissions = [...permissions, ...systemBuckets]
   } else {
     sortedPermissions = [...permissions]
   }
-  
-  
+
   const handleReadToggle = id => {
     onToggle(resourceName, id, PermissionType.Read)
   }
@@ -90,10 +89,8 @@ export const IndividualAccordionBody: FC<Props> = props => {
       <Accordion.AccordionBodyItem className="resource-accordion-body">
         {title}
       </Accordion.AccordionBodyItem>
-      {
-      permissions
-        ? 
-        Object.keys(sortedPermissions).map(key => {
+      {permissions
+        ? Object.keys(sortedPermissions).map(key => {
             return (
               <Accordion.AccordionBodyItem
                 key={sortedPermissions[key].id}
--- a/src/authorizations/components/redesigned/IndividualAccordionBody.tsx
+++ b/src/authorizations/components/redesigned/IndividualAccordionBody.tsx
@@ -28,10 +28,19 @@ export const IndividualAccordionBody: FC<Props> = props => {
   const {resourceName, permissions, onToggle, title, disabled} = props
   let sortedPermissions
 
-  if (title === 'Individual Bucket Names') {
+  if (resourceName === 'buckets') {
     // re-order buckets: user buckets first followed by system buckets
-    const systemBuckets = permissions.splice(0, 2)
-    sortedPermissions = [...permissions, ...systemBuckets]
+    const systemBuckets = []
+    const userBuckets = []
+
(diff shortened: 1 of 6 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3288)

src/authorizations/components/redesigned/IndividualAccordionBody.tsx
```diff
@@ -26,6 +26,15 @@ interface Props {
 
 export const IndividualAccordionBody: FC<Props> = props => {
   const {resourceName, permissions, onToggle, title, disabled} = props
+  let sortedPermissions
+
+  if (title === 'Individual Bucket Names') {
+    // re-order buckets: user buckets first followed by system buckets
+    const systemBuckets = permissions.splice(0, 2)
```
hoorayimhelping, line 33: do system buckets always come back at the front of the array here? what's the original source they come from? is it possible to sort them there in case they need to be sorted elsewhere in the tokens section?

if this needs to change, I'll re-review it

src/authorizations/components/redesigned/IndividualAccordionBody.tsx
```diff
@@ -26,6 +26,15 @@ interface Props {
 
 export const IndividualAccordionBody: FC<Props> = props => {
   const {resourceName, permissions, onToggle, title, disabled} = props
+  let sortedPermissions
+
+  if (title === 'Individual Bucket Names') {
```
hoorayimhelping, line 31: is it possible to `resourceName === 'buckets'` here? I think it would be a less volatile property. A designer or UX expert may decide that `Individual Bucket Names` isn't what we want to call this section, and if we need to change the text and forget to update it here, we'd have a bug where this part of the app doesn't sort buckets consistently.

if it is possible to use `resourceName === 'buckets'`, you don't have to get a re-review with this change.

hoorayimhelping (approved):
nice work! looks good to me! one of the comments doesn't need a re-review, one probably does

src/authorizations/components/redesigned/IndividualAccordionBody.tsx
blegesse-git: gooood catch!! you're absolutely right. I was looking for a variable that would differentiate between buckets and telegrafs. I was looking else where when the answer was right under my nose

src/authorizations/components/redesigned/IndividualAccordionBody.tsx
blegesse-git: Screen Shot 2021-11-22 at 5 13 01 PM
the bucket objects are pushed inside an array and then passed to this Filter component you see in the screenshot. then the functions inside Filter sort the buckets by name and return them to IndividualAccordionBody. My thought process here was - We know system buckets are the only buckets that start with '_'. User buckets can't be named with an underscore. So when being sorted, the two strings that start with '_' will always take precedence....

blegesse-git: I can also loop through permissions array inside `IndividualAccordionBody` and filter out system buckets. and then re-order them in a new array. 
that way we don't have to rely on "system buckets will always take the first and second index" theory.

PR #3305, opened 2021-11-23
test: deep clone dashboard

karel-rehor:
Replaces current dashboard clone test with clone test that includes views, cells and variables.


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3305)

cypress/support/commands.ts
```diff
@@ -398,6 +399,20 @@ export const createMapVariable = (
   })
 }
 
+export const createMapVariableFromFix = (
```
hoorayimhelping, line 402:
```suggestion
export const createMapVariableFromFixture = (
```

`fromFix` makes me think bugfix, not fixture.

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -252,6 +253,183 @@ describe('Dashboards', () => {
...
+      cy.getByTestID('nav-item-dashboards').click()
+
+      // Verify labels
+      cy.getByTestID('dashboard-card')
+        .eq(1)
+        .within(() => {
+          cy.getByTestID(`label--pill ${labelName}`).should('be.visible')
+          cy.getByTestID('dashboard-card--name').should(
+            'contain.text',
+            `${localDashName} (clone 1)`
```
hoorayimhelping, line 390: i don't know all the intricacies of what's on the screen at this point in the test, but it might be more straightforward to do something like:

```ts
cy.contains(`${localDashName} (clone 1)`)
```

or

```ts
cy.getByTestID('dashboard-card--name').contains(
  `${localDashName} (clone 1)`
)
```


[this SO answer provides some context](https://stackoverflow.com/a/50049358/92446)

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -252,6 +253,183 @@ describe('Dashboards', () => {
...
+      // TODO verify graph and variable in original once #3287 is fixed
+
+      cy.getByTestID('variable-dropdown-input-typeAhead--Power').should('have.value','base')
+      makeGraphSnapshot().shouldBeSameAs(viewGraphCopy, false)
+
+       */
+      cy.getByTestID('cell--view-empty markdown').should(
+        'contain.text',
+        'The cat went here and there'
+      )
```
hoorayimhelping, line 429: same here using

```ts
cy.contains('The cat went here and there')
```

hoorayimhelping (commented):
Looks good to me - a couple of small comments. Let's wait for Gene to weigh in on whether the `beforeEach` in the 'cloning' section is potentially flaky or not.

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -252,6 +253,183 @@ describe('Dashboards', () => {
...
+                                body.id,
+                                cell2Resp.body.id,
+                                'sampleNote'
+                              )
+                              .then(() => {
+                                cy.fixture('routes').then(({orgs}) =>
+                                  cy
+                                    .get<Organization>('@org')
+                                    .then(({id}: Organization) => {
+                                      cy.wait(1000)
```
genehynson, line 300: any way we could avoid using a `cy.wait(1000)` here? If the page isn't loading fast enough, you can add a `timeout` argument to `cy.get` to make it wait longer. The difference being that cy.wait will always make the test last 1 second longer. But if the page load times improve in the future, increasing the `timeout` only lengthens the test if necessary

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -252,6 +253,183 @@ describe('Dashboards', () => {
...
+
+      // Verify original is decoupled from clone
+      cy.getByTestID('dashboard-card')
+        .first()
+        .within(() => {
+          cy.getByTestID('dashboard-card--name').click()
+        })
+
+      /*
+      // TODO verify graph and variable in original once #3287 is fixed
```
genehynson, line 420: can we update this comment to include the full link to the issue? There are many repos and I know I'm going to forget which one 3287 belongs to a few weeks from now 😅

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -252,6 +253,183 @@ describe('Dashboards', () => {
...
+      return cy.get<Organization>('@org').then(({id}: Organization) =>
+        cy.createDashboard(id, localDashName).then(({body}) =>
+          cy.createAndAddLabel('dashboards', id, body.id, labelName).then(() =>
+            cy
+              .createLabel(secondLabelName, id, {
+                description: 'I hate mieces',
+                color: '#00ff44',
+              })
+              .then(() => {
+                cy.createCell(body.id).then(cell1Resp =>
```
genehynson, line 276: be carful of handling async tasks. Using the `{` here means that cypress is expecting a declared `return` statement. To avoid the before each ending prematurely, I would recommending removing the `{` and just using `=>` instead.

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -252,6 +253,183 @@ describe('Dashboards', () => {
...
+                          })
+                          .then(cell2Resp =>
+                            cy
+                              .createView(
+                                body.id,
+                                cell2Resp.body.id,
+                                'sampleNote'
+                              )
+                              .then(() => {
+                                cy.fixture('routes').then(({orgs}) =>
```
genehynson, line 296: here's another place where I would remove the `{`

genehynson (changes requested):
left some minor change requests. thanks for this!

cypress/support/commands.ts
karel-rehor: refactored

cypress/e2e/shared/dashboardsIndex.test.ts
karel-rehor: replaced with contains

cypress/e2e/shared/dashboardsIndex.test.ts
karel-rehor: replaced with contains

cypress/e2e/shared/dashboardsIndex.test.ts
karel-rehor: wait moved to timeout

cypress/e2e/shared/dashboardsIndex.test.ts
karel-rehor: added comment with URL to issue

cypress/e2e/shared/dashboardsIndex.test.ts
karel-rehor: curly braces removed

cypress/e2e/shared/dashboardsIndex.test.ts
karel-rehor: curly braces removed

cypress/e2e/shared/dashboardsIndex.test.ts
karel-rehor: On second though I've decided to restore the `.should()` method.  `.should()` is used when making assertions, while `.contains()` is used more like a selector in a chain.  At this point in the test we just want to assert that the card was cloned with the correct name.

cypress/e2e/shared/dashboardsIndex.test.ts
karel-rehor: On second though decided to restore the assertion `.should()`.  See reasons in previous comment.  At his point we just want to assert that the note cell is unchanged, that it contains the original text and that the note is in no way coupled with any changes in the clone.

genehynson (approved):
thanks!

PR #3315, opened 2021-11-24
feat: move clone and delete

### LINKED ISSUE #3251: add cloning and deleting to the notebook page
currently in notebooks, there is a heavy cost to cloning and deleting of a notebook. since the user must check to see the contents of a notebook before deleting, to ensure they're not deleting the wrong one, the user flow follows:
- navigate a list by name, find a potential match and click on it
- if it's not the correct notebook, take note of the name of the notebook, return to the list, find where you left off by matching a list item to that name, and continue the search
- if it is the correct notebook, take note of the name of the notebook, return to the list, find where you left off by matching a list item to that name, then select the delete button, and confirm

we can save a lot of that cognitive overhead and back and forth if we remove one of the validation cycles and allow the user to delete right from the notebook they're currently in. By reducing the overhead of deleting one off explorations, where the user hasn't even been introduced to list navigation, we should also be able to remove the need for ephemeral notebooks (dont wanna keep it? just delete it)

Design:

drdelambre:
Closes #3251 
allows for cloning and deleting from the main notebooks page

### DIFF
--- a/src/flows/components/header/index.tsx
+++ b/src/flows/components/header/index.tsx
@@ -1,11 +1,13 @@
 // Libraries
 import React, {FC, useContext, useState, useEffect} from 'react'
-import {Link} from 'react-router-dom'
+import {useHistory, Link} from 'react-router-dom'
 import {useSelector} from 'react-redux'
 
 // Contexts
 import {FlowContext} from 'src/flows/context/flow.current'
+import {FlowListContext} from 'src/flows/context/flow.list'
 import {AppSettingProvider} from 'src/shared/contexts/app'
+import {deletePinnedItemByParam} from 'src/shared/contexts/pinneditems'
 
 // Components
 import {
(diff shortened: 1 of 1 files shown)


src/flows/components/header/index.tsx
```diff
@@ -145,6 +152,21 @@ const FlowHeader: FC = () => {
...
 
+  const handleClone = async () => {
+    event('clone_notebook')
+    const clonedId = await clone(flow.id)
+    history.push(
+      `/orgs/${orgID}/${PROJECT_NAME_PLURAL.toLowerCase()}/${clonedId}`
+    )
+  }
+
+  const handleDelete = () => {
```
asalem1, line 163: Question to ponder on about. Do we want to send an event to remove the data from redis? What’s the user story here for the other person on this notebook?

asalem1 (approved):
I think this brought up a lot of questions with regards to shared editing and what happens when a notebook is deleted

src/flows/components/header/index.tsx
drdelambre: we def will need to handle that case

PR #3317, opened 2021-11-24
test: attempt to fix flaky task test by separating should and contains

hoorayimhelping:
I couldn't get the test failures seen on CI locally to begin with, so I'm not sure how effective this will be.

I ran this in a loop of 10 runs a few times and had all greens.


genehynson (approved):

hoorayimhelping:
>so I'm not sure how effective this will be.

turns out the effectiveness is somewhere completely ineffective and not very effective at all.

Issue #3322, opened 2021-11-29
Debug Rudderstack Reporting Issue


subirjolly:
It seems that snowflake has been receiving data for a long while now. Rudderstack flag (`rudderstackReporting`) has been turned off and still the data has been reaching snowflake. We need to investigate, what's sending(leaking) data to rudderstack in order to get some clarity on the functionality which wasn't intended to be turned on yet.

jzahnd:
We will investigate if this happens again.

PR #3323, opened 2021-11-29
chore: segregate track and event for trackCancellations feat

subirjolly:
Part of: #3322

Issue: Snowflake has been receiving data for a long while now. Rudderstack flag (rudderstackReporting) has been turned off and still the data has been reaching snowflake. We need to investigate, what's sending(leaking) data to rudderstack in order to get some clarity on the functionality which wasn't intended to be turned on yet.

One of the examples of the data in Snowflake which isn't being sent by `track` is: `loadTagSelectorValues function`. This is being sent only by `event`. I need to figure out if `/api/v2/app-metrics` is sending data to snowflake. I have reached out to Bucky as well as Ana and seems like they aren't sending data to snowflake. Ana will further investigate.

**Interim Solution:** In the meantime, I would like to segregate `event` and `track` to get clarity if event will be sending data to snowflake or not when rudderstack is turned off. Currently, both these events are happening in the condition where `rudderstackReporting` and `trackCancellations` are enabled. So, moving out event and sending it in all cases.

### DIFF
--- a/src/billing/components/PayAsYouGo/CancellationOverlay.tsx
+++ b/src/billing/components/PayAsYouGo/CancellationOverlay.tsx
@@ -56,6 +56,7 @@ const CancellationOverlay: FC<Props> = ({onHideOverlay}) => {
       reason: VariableItems[reason],
       canContactForFeedback: canContactForFeedback ? 'true' : 'false',
     }
+    event('CancelServiceExecuted Event', payload)
 
     if (
       isFlagEnabled('rudderstackReporting') &&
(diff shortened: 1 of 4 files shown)


src/billing/components/PayAsYouGo/CancellationOverlay.tsx
```diff
@@ -56,6 +56,7 @@ const CancellationOverlay: FC<Props> = ({onHideOverlay}) => {
       reason: VariableItems[reason],
       canContactForFeedback: canContactForFeedback ? 'true' : 'false',
     }
+    event('CancelServiceExecuted Event', payload)
```
drdelambre, line 59: does this work? i dont think the event -> line protocol translator will handle nested objects like `suggestions`

subirjolly: `suggestions` is a string. If can `JSON.stringify(payload)`, if you'd like!

drdelambre: make sure you need that string and that you're not blowing up cardinality with putting this in a tag, but g2g other than that

drdelambre (approved):

Issue #3327, opened 2021-11-29
notebooks: Notebook crashes when accessing the queryd_operations bucket in tools


barbaranelson:
I was playing around in tools, and selected the queryd_operations bucket, and it launches the notebook and then instantly errors with 'an internal error has occurred' and the console shows the stack trace below. This doesn't happen for other buckets, and the queryd_operations bucket can be queried in data explorer.

Also, when I select a new blank notebook, and then look at the list of buckets in the dropdown list, queryd_operations is not there.  Could be related to this: https://github.com/influxdata/ui/issues/839

```
react-dom.production.min.js:198 TypeError: Cannot read properties of undefined (reading 'type')
    at context.tsx:324
    at CardList.tsx:123
    at ia (react-dom.production.min.js:199)
    at Ma (react-dom.production.min.js:218)
    at c (scheduler.production.min.js:11)
    at d (scheduler.production.min.js:12)
    at MessagePort.F.port1.onmessage (scheduler.production.min.js:16)
ra @ react-dom.production.min.js:198
i.componentDidCatch.n.callback @ react-dom.production.min.js:211
Ki @ react-dom.production.min.js:193
Xi @ react-dom.production.min.js:193
La @ react-dom.production.min.js:216
Ba @ react-dom.production.min.js:220
(anonymous) @ react-dom.production.min.js:250
t.unstable_runWithPriority @ scheduler.production.min.js:18
xs @ react-dom.production.min.js:250
As @ react-dom.production.min.js:249
Cs @ react-dom.production.min.js:248
vs @ react-dom.production.min.js:245
Ma @ react-dom.production.min.js:218
c @ scheduler.production.min.js:11
d @ scheduler.production.min.js:12
F.port1.onmessage @ scheduler.production.min.js:16
```

https://influxdb.aws.influxdata.io/notebook/from/bucket/queryd_operations

PR #3332, opened 2021-11-30
fix: stop bucket bleed

drdelambre:
addresses #3327 

might not actually close the issue, but i think it might address some crashes from a schema change that didnt get migrated over


asalem1 (approved):
What schema changes lead to this and are there any considerations around explicit bucket schemas that we should take into consideration as well?

drdelambre:
this was entirely centered around notebooks. when i added the sample data to the bucket list, i forgot about this mechanism

PR #3336, opened 2021-11-30
test(cypress): add time range function test

Dfldr:
This PR adds a test which ensures functionality of both fixed time range values and custom time range values.


cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -761,4 +762,147 @@ describe('Dashboards', () => {
...
+    cy.writeLPData({
+      lines: genCurve({points: 87000, freq: 10}),
+      offset: `${30 * 24 * 60}m`,
+      stagger: `${(30 * 24 * 60) / 87000}m`,
+    })
+    cy.intercept('POST', '/api/v2/query?*', req => {
+      req.alias = 'loadQuery'
+    })
+    cy.get<Organization>('@org').then(({id}: Organization) =>
+      cy.createDashboard(id, dashName).then(({body}) => {
```
genehynson, line 776: remove the curl brace here to since we're not explicitly declaring a `return` to handle the cypress async tasks correctly

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -761,4 +762,147 @@ describe('Dashboards', () => {
       })
     })
   })
+  it('changes time range', () => {
+    const dashName = 'dashboard'
+    cy.writeLPData({
+      lines: genCurve({points: 87000, freq: 10}),
+      offset: `${30 * 24 * 60}m`,
+      stagger: `${(30 * 24 * 60) / 87000}m`,
+    })
```
genehynson, line 771: the following lines should be within a `.then` to handle this cypress async task correctly

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -761,4 +762,147 @@ describe('Dashboards', () => {
...
+    cy.getByTestID('dropdown-item-past24h').click()
+    cy.wait('@loadQuery')
+
+    const snapshot8 = makeGraphSnapshot()
+    snapshot8.shouldBeSameAs(snapshot7, false)
+
+    // past 2d
+    cy.getByTestID('timerange-dropdown').click()
+    cy.getByTestID('dropdown-item-past2d').click()
+    cy.wait(500)
```
genehynson, line 866: please remove all `cy.wait()` calls. This is a bad practice https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting . Try to use `cy.intercept` or assert the visibility of an element instead

genehynson (changes requested):
Added some comments regarding async events and cy.wait

cypress/e2e/shared/dashboardsIndex.test.ts
genehynson: actually just add `.should('equal', 'success')` to this line and I think you're good

subirjolly (approved):

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -718,4 +719,157 @@ describe('Dashboards', () => {
...
+    cy.getByTestID('timerange-dropdown').click()
+    cy.getByTestID('dropdown-item-past5m').click()
+    cy.wait('@loadQuery')
+
+    const snapshot3 = makeGraphSnapshot()
+    snapshot3.shouldBeSameAs(snapshot2, false)
+
+    /*
+    // past 15m
+    cy.getByTestID('timerange-dropdown').click()
```
genehynson, line 783: Remove commented code please

cypress/e2e/shared/dashboardsIndex.test.ts
genehynson: This is still an issue. Please address it. Thanks

cypress/e2e/shared/dashboardsIndex.test.ts
```diff
@@ -718,4 +719,157 @@ describe('Dashboards', () => {
...
+            )
+          )
+        )
+      })
+    )
+    cy.getByTestID('dashboard-card--name').click()
+    cy.getByTestID('cell-context--toggle').click()
+    cy.getByTestID('cell-context--configure').click()
+    cy.getByTestID('selector-list curve').click()
+    cy.getByTestID('selector-list p').click()
```
genehynson, line 757: There is a lot of clicking here with no assertions of visibility. This will cause the test to be flaky. I'd recommend adding some assertions of visibility between the clicks

cypress/e2e/shared/dashboardsIndex.test.ts
Dfldr: I pushed another commit where the code has been added to the test. This particular feature caused the test to fail on CircleCI. After moving the 15m time range test to the bottom, the issue didn't appear anymore.

cypress/e2e/shared/dashboardsIndex.test.ts
Dfldr: I removed the curl brace. Is that the issue you were reffering to?

cypress/e2e/shared/dashboardsIndex.test.ts
Dfldr: Commented code was a part which made the test fail. I moved it to the end of the test. It doesn't cause any problems anymore.

cypress/e2e/shared/dashboardsIndex.test.ts
Dfldr: I pushed another commit which adds assertion that ensures visibility of new window's page header and another assertion that checks giraffe layer visibility. Thanks for the recommendation

genehynson (approved):
looks great!

PR #3337, opened 2021-11-30
test: restore cron test

### LINKED ISSUE #3333: Fix flaky tests
The `can create a cron task` https://github.com/influxdata/ui/blob/4ffe4df203232a00a2c2d83024784ebb999f0e12/cypress/e2e/shared/tasks.test.ts#L68 test is failing and blocking CI. Fix it after skippingit.

jrenee42:
Closes #3333

@drdelambre  actually fixed the tests; this just restores the skipped test and adds more information if it continues to fail.

### DIFF
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -523,10 +523,13 @@ export const saveNewScript = (script: string, preamble: string) => async (
     dispatch(notify(copy.taskCreatedSuccess()))
     dispatch(checkTaskLimits())
   } catch (error) {
+    console.error('error while saving task script: ', error)
+    const message = getErrorMessage(error)
+    console.error(`error message: ${message}`)
     if (isLimitError(error)) {
+      console.error('is limit error')
       dispatch(notify(copy.resourceLimitReached('tasks')))
     } else {
-      const message = getErrorMessage(error)
       dispatch(notify(copy.taskNotCreated(message)))
     }
   }
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -523,13 +523,10 @@ export const saveNewScript = (script: string, preamble: string) => async (
     dispatch(notify(copy.taskCreatedSuccess()))
     dispatch(checkTaskLimits())
   } catch (error) {
-    console.error('error while saving task script: ', error)
-    const message = getErrorMessage(error)
-    console.error(`error message: ${message}`)
     if (isLimitError(error)) {
-      console.error('is limit error')
       dispatch(notify(copy.resourceLimitReached('tasks')))
     } else {
+      const message = getErrorMessage(error)
       dispatch(notify(copy.taskNotCreated(message)))
     }
   }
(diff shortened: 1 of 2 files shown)


src/tasks/actions/thunks.ts
```diff
@@ -523,10 +523,13 @@ export const saveNewScript = (script: string, preamble: string) => async (
     dispatch(notify(copy.taskCreatedSuccess()))
     dispatch(checkTaskLimits())
   } catch (error) {
+    console.error('error while saving task script: ', error)
+    const message = getErrorMessage(error)
+    console.error(`error message: ${message}`)
     if (isLimitError(error)) {
+      console.error('is limit error')
```
hoorayimhelping, line 530: We don't want `console.error` calls - they lower our lighthouse scores: https://web.dev/errors-in-console/ and cause warnings to show up in jest tests

hoorayimhelping (changes requested):

src/tasks/actions/thunks.ts
jrenee42: i added them because i couldn't tell why the test was failing: the notification error is unreadable in the video (the text is too grainy).  

you still want me to remove it?  (I at least want to wait until the ci run is over ; to keep that in until the tests pass)

jrenee42: (it did pass, btw)

hoorayimhelping (approved):
as per a slack conversation:

1. JIll to remove the console.error in this pr
2. Jill to do some tests to verify that the `console.error` calls do show up; and where exactly

drdelambre:
we shouldn't merge this until we run the tests a bunch of times

jrenee42:
> we shouldn't merge this until we run the tests a bunch of times

agreed.

genehynson:
> as per a slack conversation:
> 
> 1. JIll to remove the console.error in this pr
> 2. Jill to do some tests to verify that the `console.error` calls do show up; and where exactly

if you want to log something @jrenee42 , use `cy.log` so it will show up in the video

jrenee42:
> > as per a slack conversation:
> > 
> > 1. JIll to remove the console.error in this pr
> > 2. Jill to do some tests to verify that the `console.error` calls do show up; and where exactly
> 
> if you want to log something in a test @jrenee42 , use `cy.log` so it will show up in the video

@genehynson  the problem is that the errors are generated deeper in the code....they don't get up to the test level.  
but still thanks for the info!

genehynson:
> > > as per a slack conversation:
> > > 
> > > 1. JIll to remove the console.error in this pr
> > > 2. Jill to do some tests to verify that the `console.error` calls do show up; and where exactly
> > 
> > 
> > if you want to log something in a test @jrenee42 , use `cy.log` so it will show up in the video
> 
> @genehynson the problem is that the errors are generated deeper in the code....they don't get up to the test level. but still thanks for the info!

ohhhh gotcha!

drdelambre:
this has already been applied to master

PR #3344, opened 2021-12-01
test: cypress edit and verify check card values

karel-rehor:
Enhances existing test ''can edit the check card" in the specification "shared/checks.test.ts" 

   * Modifies and verifies the check name
   * Modifies and verifies the check description
   * Implicitly verifies the value in copy-resource-id
   * Verifies sanity of values in last-completed and last-updated fields
   * Verifies notification after copying task-id to clipboard


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3344)

cypress/e2e/shared/checks.test.ts
```diff
@@ -519,6 +519,14 @@ describe('Checks', () => {
     })
 
     it('can edit the check card', () => {
+      const newDescription =
+        'русский дворянин, лечившийся четыре года в Швейцарии от эпилепсии.'
+      const newName = 'Князь Мышкин'
```
genehynson, line 524: could we change this to english? 😅

cypress/e2e/shared/checks.test.ts
```diff
@@ -537,8 +545,96 @@ describe('Checks', () => {
...
 
+      // implicitly verify id value while setting up intercept
+      cy.getByTestID('copy-resource-id').then(elem => {
+        const checkID = elem
+          .text()
+          .replace(/^\D+/g, '')
+          .substring(0, 16)
+        return cy
+          .intercept('PATCH', `/api/v2/checks/${checkID}*`)
+          .as('patchCheck')
```
genehynson, line 556: I think you already declared a `patchCheck` intercept above on line 527 right?

cypress/e2e/shared/checks.test.ts
```diff
@@ -537,8 +545,96 @@ describe('Checks', () => {
...
+          expect(now.getTime() - cdate.getTime())
+            .to.be.at.least(0)
+            .and.to.be.below(Cypress.config('requestTimeout'))
+        })
+        cy.contains('Last updated').then((elem: any) => {
+          const interval = parseInt(elem.text().replace(/\D+/g, ''))
+          // assert last update value is sane
+          expect(interval)
+            .to.be.at.least(0)
+            .and.to.be.at.most(Cypress.config('requestTimeout') / 1000)
```
genehynson, line 589: I noticed here you're dividing by 1000 but on line 582 you're not. Is that intentional?

genehynson (changes requested):
just a couple very minor things - thanks for this test!

cypress/e2e/shared/checks.test.ts
karel-rehor: Yes, I'll change this to english.  In the past I've worked on projects where internationalization was important, so I developed a habit of setting up test values that use non-latin unicode sets.  But, I think this is covered in other tests.

cypress/e2e/shared/checks.test.ts
karel-rehor: Yes, but the declaration on line 527 was a mistake.  It has been removed.

cypress/e2e/shared/checks.test.ts
karel-rehor: Yes.  On line 582 the comparison is of the value of the last completed time in date format, which is converted to a date object, from which the time in milliseconds is retrieved and compared to now.
```javascript
          expect(now.getTime() - cdate.getTime())
            .to.be.at.least(0)
            .and.to.be.below(Cypress.config('requestTimeout'))
```
While here the value is from the field 'Last updated _n_ seconds ago', so the value is already the difference parsed to int and represents seconds.

genehynson: got it, cool!

genehynson (approved):

PR #2533, opened 2021-09-02
chore(tests): reuse users throughout test suite

genehynson:
The `ui-test-provisioner` has been updated to provide the ability to provision and flush individual users by passing them into the query params like `?user=23mcfly@influxdata.com`. This PR uses this ability so we can reuse users across all the tests in a single test suite. This will keep us from running into the "ran out of 1250 provisioned users" issue

PR #3350, opened 2021-12-01
fix: cypress walk through

drdelambre:
in centralizing some of the monaco editing, i decided to test some of my assumptions about cypress

removes some noops
doesn't try to autofocus if you already have focus
waits for graphs before checking on graphs

### DIFF
--- a/cypress/e2e/shared/about.test.ts
+++ b/cypress/e2e/shared/about.test.ts
@@ -1,16 +1,15 @@
 import {Organization} from '../../../src/types'
 
 describe('About Page', () => {
-  beforeEach(() =>
-    cy.flush().then(() =>
-      cy.signin().then(() => {
-        cy.get('@org').then(({id}: Organization) => {
-          cy.visit(`/orgs/${id}/about`)
-          cy.getByTestID('about-page--header').should('be.visible')
-        })
-      })
-    )
-  )
+  beforeEach(() => {
+    cy.flush()
+    cy.signin()
+
+    cy.get('@org').then((org: Organization) => {
+      cy.visit(`/orgs/${org.id}/about`)
+      cy.getByTestID('about-page--header').should('be.visible')
+    })
+  })
 
   it('should display everything correctly', () => {
     cy.getByTestID('common-ids--panel').within(() => {
--- a/cypress/e2e/cloud/pinned.test.ts
+++ b/cypress/e2e/cloud/pinned.test.ts
@@ -3,21 +3,16 @@ import {Organization} from '../../../src/types'
 describe('Pinned Items', () => {
   let orgID: string
   beforeEach(() => {
-    cy.flush().then(() =>
-      cy.signin().then(() =>
-        cy.fixture('routes').then(() => {
-          cy.get('@org').then(({id}: any) => {
-            orgID = id
-            cy.setFeatureFlags({
-              pinnedItems: true,
-              docSearchWidget: true,
-            }).then(() => {
-              cy.getByTestID('tree-nav')
-            })
-          })
-        })
-      )
-    )
+    cy.flush()
+    cy.signin()
+    cy.get('@org').then(({id}: any) => {
+      orgID = id
+      cy.setFeatureFlags({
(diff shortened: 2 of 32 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3350)

cypress/e2e/shared/about.test.ts
```diff
@@ -1,16 +1,15 @@
...
-        cy.get('@org').then(({id}: Organization) => {
-          cy.visit(`/orgs/${id}/about`)
-          cy.getByTestID('about-page--header').should('be.visible')
-        })
-      })
-    )
-  )
+  beforeEach(() => {
+    cy.flush()
+    cy.signin()
```
genehynson, line 6: I intentionally chained these requests in [this PR](https://github.com/influxdata/ui/pull/2533) because it was causing flaky behavior

genehynson (changes requested):
I noticed you're removing a bunch of chaining. Why is that? I left a comment referencing a PR I made that introduced the chaining to avoid flaky behavior

cypress/e2e/shared/about.test.ts
drdelambre: that's not how cypress works though, these are all noops. they're not js promises, just promise... like. every time you call `cy.` or interact with a child chainable instance, you're immediately invoking a call to cypress' internal scheduler. it's why i can say `cy.visit(url); cy.get('selector').click();` and it'll wait until the page has rendered and updated and found the matching element before clicking, and also why annoying things like this exist, to re-engage with the scheduler's scope: https://github.com/influxdata/ui/blob/master/cypress/e2e/shared/telegrafs.test.ts#L274

drdelambre: they hint at it in posts like: https://docs.cypress.io/guides/guides/debugging#Debug-just-like-you-always-do
but it's def not something that is intuitive

genehynson: right but I think the difference is that with `cy.flush` and `cy.signin` they perform requests with `cy.request`. And I think there are two ways to handle waiting for the response: 1) chaining with `.then` or 2) setting an alias and waiting with `cy.wait('@alias')`. Otherwise cypress will continue on because the 'promise' is not being handled. See this: https://github.com/cypress-io/cypress/discussions/9377#discussioncomment-142617

genehynson: oh you know what. Now that I refactored the `flush` and `signin` commands to return the cypress promise, you might actually be right that we no longer need to chain off of them https://docs.cypress.io/guides/references/best-practices#Unnecessary-wait-for-cy-request

genehynson: could we test this theory by making the change to not chain flush and signin in one test file on a new branch? That would trigger the CI to run that test 20x and could tell us if it will flake or not

drdelambre: i dont understand what you're trying to say with #9377, that seems like it has more to do with the example i posted from `telegrafs.test.ts` than something specific about `cy.request`. I've run the flush / signin cycle a few hundred times while untangling some of these implicit timing couplings both locally and in CI (there's some legit things popping up while i'm refactoring), idk what another 20 would do?.

genehynson: cool if you've already run it a bunch of times and haven't seen flakes, then I'm fine with giving this a shot. I think you are correct that they don't need to be chained because the function now returns the cypress chainable promise. I just know for a fact that it did cause flaky behavior before and that's why I chained them all to begin with. But, in that same PR I refactored those functions which might have been all that was needed to fix the flakiness

genehynson (approved):
thanks for doing this!

PR #3352, opened 2021-12-01
fix: typo in billing stats while loading

### LINKED ISSUE #3062: Usage page - title looks like a typo
While the usage page is loading, you see the title: "Billing Stats for to Today".  At first I thought it was a typo, then realized it corrected itself after the page loaded to say "Billing Stats for 10/1/2021 to Today".  Can we change the text so that it just says "Billing Stats" until we have figured out the date range?

genehynson:
Closes #3062

with @appletreeisyellow

While the billing data is loading, we want the header to say "Billing Stats" instead of "Billing stats for to today" which looks like a typo. 

Loading:
image

Done loading:
image

### DIFF
--- a/src/usage/BillingStatsPanel.tsx
+++ b/src/usage/BillingStatsPanel.tsx
@@ -41,7 +41,9 @@ const BillingStatsPanel: FC = () => {
           data-testid="usage-billing--title"
         >
           <h4 className="usage--billing-date-range">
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):
>instead of "Billing stats for to today" which looks like a typo.

it looks so much like a typo that I thought _you_ made a typo in the PR description. Took me a second to wrap my head around this 😆

PR #3368, opened 2021-12-03
feat: Account Operator Pages

jrenee42:
add name and change header of the account operator  page
 also add 'account name' to the account list page

Closes influxdata/quartz#4737

### DIFF
--- a/cypress/e2e/cloud/buckets.test.ts
+++ b/cypress/e2e/cloud/buckets.test.ts
@@ -280,8 +280,8 @@ describe('Explicit Buckets', () => {
       checkContents
     )
   })
-  // also skipping this flaky test for now; fixes are being worked on
-  it.skip('should be able to create an explicit bucket and add csv schema file during editing', function() {
+
+  it('should be able to create an explicit bucket and add csv schema file during editing', function() {
     const origFileContents = `name,type,dataType
 time,timestamp,
 host,tag,
@@ -303,6 +303,7 @@ fsRead,field,float`
       checkContents
     )
   })
+
   it('should be able to create an explicit bucket and update the existing schema file during editing', function() {
     cy.getByTestID('Create Bucket').click()
     cy.getByTestID('create-bucket-form').should('be.visible')
--- a/cypress/e2e/cloud/buckets.test.ts
+++ b/cypress/e2e/cloud/buckets.test.ts
@@ -280,8 +280,8 @@ describe('Explicit Buckets', () => {
       checkContents
     )
   })
-  // alse skipping this flaky test for now; fixes are being worked on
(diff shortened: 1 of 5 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3368)

TCL735 (approved):

src/operator/account/AccountView.tsx
```diff
@@ -15,11 +15,19 @@ import PageSpinner from 'src/perf/components/PageSpinner'
 const AccountView: FC = () => {
   const {account, accountStatus} = useContext(AccountContext)
 
+  let accountTitle = ''
+  if (account?.name) {
+    const accountIdPart = account.id ? `(${account.id})` : ''
+    accountTitle = `${account.name} ${accountIdPart} `
+  } else {
+    accountTitle = `Account ${account?.id}`
```
hoorayimhelping, line 23: two questions: why not just set `accountTitle` to `Account ${account?.id}` on line 18 and remove this extra `else` statement? that way three lines and some extra syntactic parts are removed.

what happens if `id` doesn't exist? will it just say `Account `? is it even possible for `account.id` not to exist?

hoorayimhelping (approved):

src/operator/account/AccountView.tsx
jrenee42: *if* account is there, then the id will be there, but the account isn't there initially.

because the 'account' variable isn't set yet when the page first loads.  so either need the code above, or something like this:

``` let accountTitle = 'loading....'
  if (account) {
    if (account?.name) {
      accountTitle = `${account.name} (${account.id})`
    } else {
      accountTitle = `Account ${account.id}`
    }
  } ```

File: cypress/e2e/cloud/buckets.test.ts [hunk 2]
```diff
@@ -260,8 +260,8 @@ describe('Explicit Buckets', () => {
       checkContents
     )
   })
-
-  it('should be able to create an explicit bucket and add csv schema file during editing', function() {
+  // also skipping this flaky test for now; fixes are being worked on
+  it.skip('should be able to create an explicit bucket and add csv schema file during editing', function() {
```
TCL735, line 264: A pull request was just merged to fix the flakiness with all Buckets tests. Please do not skip any tests.

TCL735 (changes requested):

PR #3373, opened 2021-12-03
fix: inactive tokens should display greyed out resource card

### LINKED ISSUE #3366: tokens: Inactive tokens should be greyed out
In the list of tokens, there is no visual distinction between the inactive and active tokens. This is inconsistent with the rest of the UI, e.g. in the list of tasks the inactive tasks are shaded (see screenshot).  The inactive tokens should be shaded.
Screen Shot 2021-12-02 at 6 01 45 PM

Palakp41:
Closes #3366

Kapture 2021-12-03 at 11 01 54

### DIFF
--- a/src/authorizations/components/TokenRow.tsx
+++ b/src/authorizations/components/TokenRow.tsx
@@ -50,6 +50,7 @@ class TokenRow extends PureComponent<Props> {
       <ResourceCard
         contextMenu={this.contextMenu}
         testID={`token-card ${auth.description}`}
+        disabled={!this.isTokenEnabled}
       >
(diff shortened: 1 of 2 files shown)


hoorayimhelping (approved):
nice one!

for consistency, do we also want to consider enabling and disabling from the list view?

Palakp41:
@hoorayimhelping as part of the new design, it was required we have the `toggle` inside the EditTokenOverlay. If we move it out and the token name is also editable, then there's no need for EditTokenOverlay and would have to be turned into ViewTokenOverlay. I think it should be a discussion with product.

PR #3376, opened 2021-12-06
chore: update cypress to 8.2.0

asalem1:
This PR is meant to address some security concerns around a dependency of cypress's. More specifically, `json-schema` is giving us a vulnerability warning. This is a first pass at updating cypress to the latest version

### DIFF
--- a/cypress/e2e/cloud/home.test.ts
+++ b/cypress/e2e/cloud/home.test.ts
@@ -1,8 +1,12 @@
+import {Organization} from '../../../src/types'
+
+let orgID = ''
 describe('Home Page Tests', () => {
   beforeEach(() =>
     cy.flush().then(() =>
       cy.signin().then(() => {
-        cy.get('@org').then(() => {
+        cy.get('@org').then(({id}: Organization) => {
+          orgID = id
           cy.getByTestID('home-page--header').should('be.visible')
           cy.setFeatureFlags({
             alertsActivity: true,
@@ -39,7 +43,7 @@ describe('Home Page Tests', () => {
 ,,0,2021-08-05T20:54:31.642101603Z,Alpha,07f205512a38f000,Alpha,ok
 ,,0,2021-08-05T20:54:00Z,Check: Beta is: ok,07f2055f7a12d000,Beta,ok`
 
-    cy.intercept('POST', '/api/v2/query?*', req => {
+    cy.intercept('POST', `/api/v2/query?orgID=${orgID}`, req => {
       if (req.body.query.includes('r._measurement == "statuses"')) {
         req.alias = 'statusesQuery'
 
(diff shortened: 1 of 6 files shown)


cypress/e2e/cloud/home.test.ts
```diff
@@ -39,7 +43,7 @@ describe('Home Page Tests', () => {
 ,,0,2021-08-05T20:54:31.642101603Z,Alpha,07f205512a38f000,Alpha,ok
 ,,0,2021-08-05T20:54:00Z,Check: Beta is: ok,07f2055f7a12d000,Beta,ok`
 
-    cy.intercept('POST', '/api/v2/query?*', req => {
+    cy.intercept('POST', `/api/v2/query?orgID=${orgID}`, req => {
```
asalem1, line 46: the crux of the migration is twofold:

1. Intercepts must now abide by strict path matching, no more wildcards
2. Intercepts now go in the the opposite order they were declared for .... reasons (check the docs)

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3376)

cypress/e2e/cloud/home.test.ts
```diff
@@ -1,8 +1,12 @@
+import {Organization} from '../../../src/types'
+
+let orgID = ''
```
genehynson, line 3: I think the best practice for this kind of thing is to use `cy.wrap`. https://docs.cypress.io/api/commands/wrap#Syntax Otherwise it's possible this variable might not get assigned when you think it should because of how cypress handles its async events with its internal scheduler

asalem1: not sure i follow, since the operation should be set prior to the test running, unless i'm misunderstanding beforeEach. I'll take a look at the docs though since wrapping in a get had caused issues with intercepting requests

drdelambre: i also dont understand why you're using this technique and not the `cy.get('@org')` around the intercept

genehynson: https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Variables It looks like it's ok in some cases - so maybe this situation is alright? But typically you're not supposed to set global variables like you're doing here with orgID

genehynson: oh lol you're right @drdelambre - it's already been wrapped. So `cy.get('@org')` should be all that's needed

genehynson: here's an example where using global variables was causing flaky behavior: https://github.com/influxdata/ui/pull/3377

asalem1: Right, I used the get and it messed with intercept is what I'm saying

genehynson (approved):

PR #3393, opened 2021-12-07
chore: remove clients.users.me usage

### LINKED ISSUE #3383: Replace client.users.me with /me
Related to #2902

We want to remove `@influxdata/influx` as a dependency as much as we can given the current build system. We should first start to convert existing usages that don't interfere with the build process with generated functions that are already available. For this issue we should replace all usages of `clients.users.me` with the `/me` request that's generated in `generatedRoutes` for OSS and Cloud builds

This should basically be a slimmed down version of this:

https://github.com/influxdata/ui/pull/3256

appletreeisyellow:
Closes #3383

Replace all usages of `clients.users.me` with the `getMe` request that's generated in generatedRoutes.

### DIFF
--- a/src/Signin.tsx
+++ b/src/Signin.tsx
@@ -30,7 +30,7 @@ import {
 
 // Types
 import {RemoteDataState} from 'src/types'
-import {getMe} from './client'
+import {getMe} from 'src/client'
 
 interface State {
   loading: RemoteDataState
--- a/src/Signin.tsx
+++ b/src/Signin.tsx
@@ -3,8 +3,6 @@ import React, {ReactElement, PureComponent} from 'react'
 import {Switch, Route, RouteComponentProps} from 'react-router-dom'
 import {connect, ConnectedProps} from 'react-redux'
 
-import {client} from 'src/utils/api'
-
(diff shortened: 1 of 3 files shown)


asalem1 (approved):
LGTM!

src/Signin.tsx
```diff
@@ -32,6 +30,7 @@ import {
 
 // Types
 import {RemoteDataState} from 'src/types'
+import {getMe} from './client'
```
hoorayimhelping, line 33: it's probably better to make this an absolute path, i.e. `src/client` rather than a relative one. It's more consistent with the rest of the imports in the file and the codebase and it helps keep this file more portable if it ever moves locations.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3393)

PR #3409, opened 2021-12-09
feat: show selected filters when searching in notebooks querybuilder

### LINKED ISSUE #2866: query builder should always show selected options
i was poking around the frontendservices bucket with the querybuilder in notebooks, searched for one measurement, selected it, explored it.... then thought "lemme check this other thing" so i typed in the filter to find the other measurement, selected it, and then i got a bunch of weird tables come back, because.... i had not unselected the original measurement. but i had no idea it was still selected because it was being filtered.

if an option in one of the query builder cards is selected, it should always be visible

subirjolly:
Closes #2866 

**BEFORE**
BEFORE

**AFTER**
AFTER

### DIFF
--- a/src/flows/pipes/QueryBuilder/CardList.tsx
+++ b/src/flows/pipes/QueryBuilder/CardList.tsx
@@ -32,7 +32,6 @@ import WaitingText from 'src/shared/components/WaitingText'
 import BuilderCard from 'src/timeMachine/components/builderCard/BuilderCard'
 import SelectorList from 'src/timeMachine/components/SelectorList'
 
-
 const DEBOUNCE_TIMEOUT = 500
 const debounce_array = []
 type NOOP = () => void
--- a/src/flows/pipes/QueryBuilder/CardList.tsx
+++ b/src/flows/pipes/QueryBuilder/CardList.tsx
@@ -52,12 +52,10 @@ const Card: FC<Props> = ({idx}) => {
   const [valueSearches, setValueSearches] = useState([])
 
   const allItems = useMemo(() => {
-    const selected = new Set(card.values.selected)
-    const remaining = card.values.results.filter(
-      result => !selected.has(result)
-    )
+    const results = new Set(card.values.results)
+    const selected = card.values.selected.filter(s => !results.has(s))
 
-    return [...Array.from(selected), ...remaining]
+    return [...selected, ...Array.from(results)]
   }, [card?.values?.selected, card?.values?.results])
 
   const _remove =
--- a/src/flows/pipes/QueryBuilder/CardList.tsx
+++ b/src/flows/pipes/QueryBuilder/CardList.tsx
@@ -35,6 +35,12 @@ interface Props {
   idx: number
 }
 
+const joinAndDedupeArrays = (arr1: string[], arr2: string[]): string[] => {
+  const deduped = new Set([...arr1, ...arr2])
+
+  return Array.from(deduped)
+}
+
 const Card: FC<Props> = ({idx}) => {
   const {cards, add, update, remove, loadKeys, loadValues} = useContext(
     QueryBuilderContext
@@ -87,7 +93,6 @@ const Card: FC<Props> = ({idx}) => {
   const valueSelect = val => {
     const _vals = [...card.values.selected]
     const index = _vals.indexOf(val)
-
     if (index === -1) {
       _vals.push(val)
     } else {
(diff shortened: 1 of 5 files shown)


src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -185,7 +190,7 @@ const Card: FC<Props> = ({idx}) => {
   } else {
     _values = (
       <SelectorList
-        items={card.values.results}
+        items={joinAndDedupeArrays(card.values.results, card.values.selected)}
```
genehynson, line 193: thoughts on moving this up and wrapping it in a `useMemo`? I'm thinking this component probably gets re-rendered a lot and if there are hundreds of buckets this might be an expensive operation maybe?

src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -35,6 +35,12 @@ interface Props {
   idx: number
 }
 
+const joinAndDedupeArrays = (arr1: string[], arr2: string[]): string[] => {
+  const deduped = new Set([...arr1, ...arr2])
+
+  return Array.from(deduped)
+}
+
```
genehynson, line 43: I think you could move this to `src/flows/utils.ts`

src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -35,6 +35,12 @@ interface Props {
   idx: number
 }
 
+const joinAndDedupeArrays = (arr1: string[], arr2: string[]): string[] => {
+  const deduped = new Set([...arr1, ...arr2])
```
drdelambre, line 39: this doesn't seem to be setting the correct order. data explorer is floating the selected, but not available in results, items to the top. this seems to be random?

drdelambre (changes requested):
the order needs to be deterministic and you're missing the number of selected. see:
Screen Shot 2021-12-09 at 12 59 39 PM

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3409)

src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -50,6 +51,15 @@ const Card: FC<Props> = ({idx}) => {
   const [keySearches, setKeySearches] = useState([])
   const [valueSearches, setValueSearches] = useState([])
 
+  const allItems = useMemo(() => {
+    const selected = new Set(card.values.selected)
+    const remaining = card.values.results.filter(
```
drdelambre, line 56: almost. the logic seems to be slightly reversed from this: filter out all results from _selected_ that exist in _results_, and prepend whatever is left from that to the _results_ list

src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -44,6 +59,13 @@ const Card: FC<Props> = ({idx}) => {
   const [keySearches, setKeySearches] = useState([])
   const [valueSearches, setValueSearches] = useState([])
 
+  const allItems = useMemo(() => {
```
genehynson, line 62: niceeeeeee

src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -44,6 +59,13 @@ const Card: FC<Props> = ({idx}) => {
...
+    const results = new Set(card.values.results)
+    const selected = card.values.selected.filter(s => !results.has(s))
```
genehynson, line 64: do we need `?` here? e.g. `card?.values?.selected?.filter...`?

src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -245,6 +263,9 @@ const Card: FC<Props> = ({idx}) => {
             menuTestID="tag-selector--dropdown-menu"
             options={card.keys.results}
           />
+          {!!card.values.selected.length && (
```
genehynson, line 266: `?`s here maybe?

genehynson (approved):

PR #3423, opened 2021-12-13
chore: replacing client.telegrafConfigs with internal endpoints

### LINKED ISSUE #3384: Replace client.telegrafConfigs
Related to #2902

Similar to #3383 

This issue might be a bit more intense than #3383 since it's poorly scoped and may contain a lot of gotchas, especially with regard to typescript and OpenAPI. The gist of this is that we want to audit the `client.telegrafConfigs` usage and replace all existing usages of it with generated functions. The purpose of this is to limit our usage of the client library so that we can slowly transition away from it as a dependency (as best we can under the current constraints of the build system).

This issue might require a bit more work to integrate full, but we basically want to try:

- [x] Identify all usages of `client.telegrafConfigs.*`
- [x] Replace `client.telegrafConfigs.*` with existing generated endpoints
- [x] Identify any gaps between what we are currently generating, and what we are using with the client library
- [x] Update OpenAPI yaml definitions to bridge the gap between those
- [x] Replace any types related to the `client.telegrafConfigs` with the generated types

appletreeisyellow:
Closes #3384

This PR replaced all the influx dependency for `client.telegrafConfigs.*` and bridged any gaps in between.


TCL735 (changes requested):
This doesn't work. Not sure what you are doing with this PR but it BREAKS creating Telegraf configurations.

src/telegrafs/actions/thunks.ts
```diff
@@ -149,8 +164,29 @@ export const addTelegrafLabelsAsync = (
   labels: Label[]
 ): AppThunk<Promise<void>> => async (dispatch): Promise<void> => {
   try {
-    await client.telegrafConfigs.addLabels(telegrafID, labels as ILabel[])
-    const telegraf = await client.telegrafConfigs.get(telegrafID)
+    if (labels.length === 0) {
+      throw new Error('No label found to add')
+    }
```
asalem1, line 169: I don't think we need this

src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -42,11 +41,7 @@ import {
   TelegrafEntities,
   Telegraf,
 } from 'src/types'
-import {
-  TelegrafRequest,
-  TelegrafPluginOutputInfluxDBV2,
-  Permission,
-} from '@influxdata/influx'
+import {TelegrafRequest, Permission} from '@influxdata/influx'
```
asalem1, line 44:
```suggestion
import {Permission} from '@influxdata/influx'
```

src/telegrafs/actions/thunks.ts
```diff
@@ -149,8 +164,29 @@ export const addTelegrafLabelsAsync = (
   labels: Label[]
 ): AppThunk<Promise<void>> => async (dispatch): Promise<void> => {
   try {
-    await client.telegrafConfigs.addLabels(telegrafID, labels as ILabel[])
-    const telegraf = await client.telegrafConfigs.get(telegrafID)
+    if (labels.length === 0) {
```
asalem1, line 167: Super weird that this is accepting an array of labels. I don't think that's possible from the UI. I would check to see where this being used so that we can:

1. Identify whether we're only ever passing in 1 label or an array

If we find that we are passing in an array, we're going to want to add all the labels, not just the first one.

src/telegrafs/actions/thunks.ts
```diff
@@ -149,8 +164,29 @@ export const addTelegrafLabelsAsync = (
...
+      throw new Error('No label found to add')
+    }
+
+    const response = await postTelegrafsLabel({
+      telegrafID,
+      data: {labelID: labels[0].id},
```
asalem1, line 173: this relates to the previous comment about needing to add all the labels, not just the first one

src/telegrafs/actions/thunks.ts
```diff
@@ -168,8 +204,28 @@ export const removeTelegrafLabelsAsync = (
   labels: Label[]
 ): AppThunk<Promise<void>> => async (dispatch): Promise<void> => {
   try {
-    await client.telegrafConfigs.removeLabels(telegrafID, labels as ILabel[])
-    const telegraf = await client.telegrafConfigs.get(telegrafID)
+    if (labels.length === 0) {
+      throw new Error('No label found to delete')
+    }
```
asalem1, line 209: Again, I would check to see if the UI only ever passes in a label or if we anticipate an array and just refactor this from there

src/telegrafs/actions/thunks.ts
```diff
@@ -168,8 +204,28 @@ export const removeTelegrafLabelsAsync = (
...
+    const response = await deleteTelegrafsLabel({
+      telegrafID,
+      labelID: labels[0].id,
+    })
```
asalem1, line 213: same gist as before about wanting to remove them all if it's an array

src/telegrafs/actions/thunks.ts
```diff
@@ -168,8 +204,28 @@ export const removeTelegrafLabelsAsync = (
...
+
+    if (response.status !== 204) {
+      throw new Error(response.data.message)
+    }
+
+    const res = await apiGetTelegraf({
+      telegrafID,
+      headers: {Accept: 'application/json'},
```
asalem1, line 221: we should investigate why you need to do this vs why I'm not running into this same issue

asalem1 (changes requested):
Also what Timmy said. Let's make sure we run through the existing telegraf flows to see that everything works the way it should

src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -376,43 +372,25 @@ export const createOrUpdateTelegrafConfigAsync = () => async (
...
-      return acc
-    },
-    [influxDB2Out]
-  )
 
   if (telegrafConfigID) {
-    const telegraf = await client.telegrafConfigs.update(telegrafConfigID, {
-      name: telegrafConfigName,
-      description: telegrafConfigDescription,
-      plugins,
```
appletreeisyellow, line 413: By testing Load Data > Sources page by selecting a Telegraf Plugin, we still need plugins

asalem1: 🤯 . @appletreeisyellow are we making a request to the `telegraf/plugin` endpoint or is it the regular `POST` to `/telegraf`?

appletreeisyellow: Here in the original code it is just a regular `POST`/`PUT` to `/telegraf`, but the influx API has the functionality to update the plugins. We might want to make a request to both `POST`/`PUT` `/telegraf` and `telegraf/plugin` for the new endpoint. 

Yes, we need the `plugins` for both the creation AND update step.

Seems like we need

asalem1: Not sure i follow, I don't see any addition stuff happening in the client API that updates the plugins besides just passing them along. Let's revert these changes for now, make the changes requested in other parts and deal with the post/put separately in a follow-up PR

src/telegrafs/actions/thunks.ts
```diff
@@ -76,17 +85,17 @@ export const getTelegrafs = () => async (
 }
 
 export const createTelegraf = (telegraf: Telegraf) => async (
-  dispatch: Dispatch<Action>,
-  getState: GetState
+  dispatch: Dispatch<Action>
 ) => {
   try {
-    const state = getState()
-    const labels = getLabels(state, telegraf.labels)
```
appletreeisyellow, line 84: A new telegraf config doesn't expect to have any labels, so no need to call `getLabels`.

src/telegrafs/actions/thunks.ts
```diff
@@ -100,17 +109,19 @@ export const createTelegraf = (telegraf: Telegraf) => async (
 }
 
 export const updateTelegraf = (telegraf: Telegraf) => async (
-  dispatch: Dispatch<Action>,
-  getState: GetState
+  dispatch: Dispatch<Action>
 ) => {
   try {
-    const state = getState()
-    const labels = getLabels(state, telegraf.labels)
```
appletreeisyellow, line 108: Same as above, updating an existing telegraf config doesn't update labels, so no need to call `getLabels`.

src/dataLoaders/actions/dataLoaders.ts
appletreeisyellow: Sorry it was confusing. We were making a request to the regular `POST` to `/telegraf`.

appletreeisyellow: We are able to include the plugins in the request body of POST `/telegrafs` and PUT `/telegrafs/${params.telegrafID}` after a [fix](https://github.com/influxdata/openapi/pull/246) on OpenAPI. Thanks @glinton!!

src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -407,12 +408,21 @@ export const createOrUpdateTelegrafConfigAsync = () => async (
...
+    const response = await putTelegraf({
+      telegrafID: telegrafConfigID,
+      data: {
+        name: telegrafConfigName,
+        description: telegrafConfigDescription,
+        // TODO: interface TelegrafPlugin does not have property 'type'
+        plugins,
```
appletreeisyellow, line 417: We may want to update the type of `plugins` which is interface `TelegrafPlugin`.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3423)

src/telegrafs/actions/thunks.ts
appletreeisyellow: Does it has anything to do with an updated version of webpack dependencies? I haven't have updated it recently.

asalem1: It shouldn't have anything to do with that since it hinges upon the `generate` command. If you want to try merging in master and seeing if restarting the UI generates the expected functions without those `headers` being set

src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -407,12 +409,21 @@ export const createOrUpdateTelegrafConfigAsync = () => async (
...
-    const telegraf = await client.telegrafConfigs.update(telegrafConfigID, {
-      name: telegrafConfigName,
-      description: telegrafConfigDescription,
-      plugins,
+    const response = await putTelegraf({
+      telegrafID: telegrafConfigID,
+      data: {
+        name: telegrafConfigName,
+        description: telegrafConfigDescription,
+        // TODO: interface TelegrafPlugin does not have property 'type'
```
TCL735, line 417: I don't understand this comment. Can you explain what you mean by this TODO?

The actual TypeScript warning here is `Object literal may only specify known properties, and 'plugins' does not exist in type 'TelegrafRequest'`

TCL735 (dismissed):

asalem1 (approved):
Great job cleaning things up. It would be super cool to follow-up about the headers being set on the `get` since that _shouldn't_ be required - not really sure what's going on there. If it doesn't work without it then I'd say keep it in there and we can revisit it later

src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -395,7 +396,8 @@ export const createOrUpdateTelegrafConfigAsync = () => async (
     },
   }
 
-  const plugins = telegrafPlugins.reduce(
+  // TODO: convert this into a generic type
```
asalem1, line 399: Can we update this TODO with a link to an issue so we follow-up on this typecasting?

src/dataLoaders/actions/dataLoaders.ts
appletreeisyellow: This TODO is related to the typecasting in another TODO on line # 399.

'plugins' did not exist in type 'TelegrafPluginRequest.' After a [fix](https://github.com/influxdata/openapi/pull/246) in OpenAPI, and an update of SHA in `package.json` for `generate` script, 'plugins' is a valid field now. 

I think I should remove the TODO here because the warning is gone now. I will update the TODO on line # 399 to make it more clear. Just FYI, I wrote an issue #3433 related to this.

src/telegrafs/actions/thunks.ts
appletreeisyellow: I tried to merge in master, `yarn install`, and restarting UI on both Chrome and Firefox. I still have to set `headers` in order to get json objects. Very wired. We might want to revisit this in the future. But for now, we want to set the `headers` in order to receive json object.

appletreeisyellow:
I opened an issue above to follow up if we want to remove `headers`. For now, we still want to keep the `headers` set in order to receive a json object from '/telegrafs/{telegrafID}'.

PR #3429, opened 2021-12-14
test: configuration in checks

karel-rehor:
* Reorganizes checks.test.ts 
      * with groups for threshold and deadman checks
      * by promoting some test objects to higher scope for reuse
   * Adds tests for reconfiguring threshold and deadman checks
   * In status history tests replaces excessive rerouting, which is prone to timeout in firefox, with clicking through the nav menu


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3429)

jzahnd:
@subirjolly would you mind taking a look at this one when you have time?

cypress/e2e/shared/checks.test.ts
```diff
@@ -115,6 +116,129 @@ describe('Checks', () => {
...
+
+    it('can reconfigure a threshold check', () => {
+      const every = '5m'
+      const offset = '1m'
+      const testTags: {key: string; value: string}[] = [
+        {key: 'Prima', value: 'Smith'},
+        {key: 'Astaire', value: 'Rodgers'},
+        {key: 'Gable', value: 'Lombard'},
+      ]
+      const newMsg =
```
subirjolly, line 128: it's usually a good practice to initialize the variable nearer to where it's used.

cypress/e2e/shared/checks.test.ts
```diff
@@ -115,6 +116,129 @@ describe('Checks', () => {
...
+      cy.getByTestID('dropdown-menu--contents').within(() => {
+        cy.contains('5m').click()
+      })
+      cy.getByTestID('offset-options').click()
+      cy.getByTestID('dropdown-menu--contents').within(() => {
+        cy.contains('1m').click()
+      })
+      // Tags
+
+      for (let i = 0; i < testTags.length; i++) {
```
subirjolly, line 160: Suggestion: should be able to use `testTags.forEach` instead

cypress/e2e/shared/checks.test.ts
```diff
@@ -115,6 +116,129 @@ describe('Checks', () => {
...
+      cy.getByTestID('status-message-textarea')
+        .clear()
+        .type(newMsg, {parseSpecialCharSequences: false})
+
+      // Thresholds
+      // remove one threshold
+      cy.getByTestID('panel-OK').within(() => {
+        cy.getByTestID('add-threshold-condition-OK').should('not.exist')
+        cy.getByTestID('dismiss-button').click()
+        cy.wait(500)
```
subirjolly, line 181: `wait` is usually looked down upon. So, either move the dependent logic in here or just try moving this block's contents out and see if that works?

cypress/e2e/shared/checks.test.ts
```diff
@@ -115,6 +116,129 @@ describe('Checks', () => {
...
+      // change value of one threshold
+      cy.getByTestID('panel-WARN').within(() => {
+        cy.getByTestID('input-field-WARN')
+          .clear()
+          // seem to be encountering cypress issue here https://github.com/cypress-io/cypress/issues/3817
+          .type(` ${rangeTop}{del}`) // workaround
+      })
+      cy.getByTestID('save-cell--button').click()
+      cy.wait('@putCheck').then(({response}) => {
+        if (response) {
```
subirjolly, line 220: Suggestion: To reduce cyclomatic complexity, you can just do this:
```
if (!response) {
    fail('No response after update check')
    return
}

expect(response.body.tags).to.deep.equal(testTags)
expect(response.body.every).to.equal(every)
expect(response.body.offset).to.equal(offset)
expect(response.body.statusMessageTemplate).to.equal(newMsg)
expect(
  response.body.thresholds.filter((th: any) => th.level === 'INFO')[0]
    .min
).to.equal(parseFloat(rangeBtm))
expect(
  response.body.thresholds.filter((th: any) => th.level === 'INFO')[0]
    .max
).to.equal(parseFloat(rangeTop))
expect(
  response.body.thresholds.filter((th: any) => th.level === 'WARN')[0]
    .value
).to.equal(parseFloat(rangeTop))
```

cypress/e2e/shared/checks.test.ts
subirjolly: possibly, try using `cy.focused()` instead if the upper suggestion doesn't work?

subirjolly:
Any update on this?

cypress/e2e/shared/checks.test.ts
karel-rehor: That's debatable.  When coding in C, C++ back in the 90's it was good practice to have a variable declaration/definition block, then a logic block, especially when the variables can be reused in a number of places.  By declaring variables at the start of the namespace, I know what I'm then going to be working with. However, if you insist, this can be moved.

cypress/e2e/shared/checks.test.ts
karel-rehor: The wait was an artefact from some debugging that I forgot to remove.  It is now removed.

cypress/e2e/shared/checks.test.ts
karel-rehor: replaced with forEach

cypress/e2e/shared/checks.test.ts
karel-rehor: Refactored as per suggestion

karel-rehor:
Hi Subir, 

Just got back from holiday retreat.  I hope you had a nice break as well.  I've made most of your suggested fixes.  Replies to comments are above.  Changes should be pushed once stability is verified.

karel-rehor:
@subirjolly 

Hi Subir, 

Any reaction to the changes made two weeks ago?

cypress/e2e/shared/checks.test.ts
subirjolly: Absolutely! It's not 90s anymore. 3 decades later, it's usually what I said. Regardless, if you don't want to change it, ok!

subirjolly (approved):

PR #3443, opened 2021-12-16
chore: remove duplicate dependency from package.json

asalem1:
Pretty straightforward clean-up. Removing a dependency that appears to have been accidentally added in the process of an upgrade


hoorayimhelping (approved):
did this change yarn.lock?

asalem1:
> did this change yarn.lock?

It didn’t

PR #3464, opened 2021-12-17
fix: disable clockface CSS filtering to improve visual performance

TCL735:
Addresses the performance issues in Safari with the Load Data > Sources page:

- reverts a previous attempt which only made plugins smaller and did nothing for visual peformance
- disables the CSS filtering which is the real culprit of the slow rendering

https://user-images.githubusercontent.com/10736577/146468321-7656140e-9144-4729-b8ad-b1b1083ba478.mp4

### DIFF
--- a/src/writeData/components/WriteDataItem.scss
+++ b/src/writeData/components/WriteDataItem.scss
@@ -5,6 +5,12 @@
   color: $cf-white;
 }
 
+.write-data--item {
+  .cf-selectable-card--children {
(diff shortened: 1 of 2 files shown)


hoorayimhelping (approved):
let's also write up a ticket to address any cleanup to avoid any jank after we fix the EAR

TCL735:
Only took 20 seconds to reach the bottom of the page. Which is a whole lot better than never.

PR #3480, opened 2021-12-20
fix: connect the API changes to the UI to get maps to display in shared notebooks

### LINKED ISSUE #3245: notebook sharing: maps don't render correctly in a shared notebook
If you share a notebook with a map, the map will not display and you get the error "we are having issues connecting to the Maps Server. Please try again later". The request GET https://us-west-2-1.aws.cloud2.influxdata.com/api/v2/maps/mapToken fails with a 401.

Note when trying to reproduce - For some users the notebook will render correctly, if they happen to have the maps token in their session cookie.  Need to test in an incognito window.

asalem1:
Closes #3245

This PR (along with a corresponding API change), fixes the issue where maps in shared notebooks can't be displayed. The API PR goes into more detail as to what the issue is, and how it's resolved

### DIFF
--- a/src/visualization/types/Map/view.tsx
+++ b/src/visualization/types/Map/view.tsx
@@ -51,7 +51,14 @@ const GeoPlot: FC<Props> = ({result, properties}) => {
   })
 
   useEffect(() => {
-    if (CLOUD) {
+    const token = properties?.layers?.find(
+      layer => layer?.tileServerConfiguration?.tileServerUrl
+    )
+    if (CLOUD && token) {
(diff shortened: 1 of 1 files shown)


Palakp41 (approved):
Hm looks good to me! I didn't realize, that token already existed when view is rendered through notebooks.

asalem1:
> Hm looks good to me! I didn't realize, that token already existed when view is rendered through notebooks.

It's not. I'll link you to the API side where this is going to happen

genehynson (approved):

PR #3490, opened 2021-12-21
feat: time panel

drdelambre:
sometimes i want to look at a subset of data before running a longer query, or compare WoW on my graphs. our current range selector is a bit limited on what it allows, so i added a panel that lets me run a notebook across multiple time ranges

run `influx.set('flow-panel--time', true)` in the console to see the new panel

Peek 2021-12-21 15-43


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3490)

asalem1 (commented):
I mean, sure? Allows the user to iterate over a panel rather than being blocked by the `convert to flux` and changing the timerange in a subsequent copy pasta? It would be super cool if the timerange component just tweaked the previous query and output the results so that the user didn't have to add a panel, make the query again, and add a visualization? See what I'm saying? I know it would mean AST parsing and such, but if the intent is to iterate over the same query with a different timerange than that'd be :fire:

If the intent is just segment different parts of the notebook into segments truths for inspection and comparison then this works

drdelambre:
you mean like this?

Peek 2021-12-22 08-45

asalem1 (approved):
It does what I wanted it to do.

PR #3499, opened 2021-12-23
chore: Remove @influxdata/influx dependency

### LINKED ISSUE #2902: Remove @influxdata/influx as a dependency
We're currently using the `@influxdata/influx` for primarily two things:

1. Getting and assigning TS types
2. Making requests to the `/me` endpoint to check for user sessions.

The thing is, we have a yml files that we reference as our source of truth for types, and we use the generated functions from those yml definitions to make requests to the API. That means we might have a situation where a `type Label !== type Label` because TS and all that jazz. Furthermore, we run the risk of drifting between types as defined by the API AND we leave ourselves exposed by our peer dependencies (namely, axios, but maybe there are more). 

### Solution

We should first:

1. Check the `influxdata/influx` repo to see how those type definitions are being generated and try and bridge the gaps between those representations with what we currently have in the openapi yml.
2. Swap out TS definitions that are used from `influxdata/influx` AND swap out usages of the `Client` class to an oats-generated approach

### LINKED ISSUE #3496: Replace client.scrapers
Related to #2902

This should round out most of the usages for the usage of client. Basically we're gonna want to try and replace the usage of the client.scrapers with conditionally rendered OSS functions for the client.scrapers code. This has a lot of question marks around it since there may be some weirdness in the build and unit tests based on conditional imports based on the environment.

asalem1:
Closes #3496 
Closes #2902

This PR finishes off the replacement of the `client.*` usage through the app to make API requests based on an outdated internal dependency. In doing so, we resolve one of the critical dependency issues we've been having with axios as a peer dependency of the version locked client library we were using. 

### Notes for developers

In order to resolve some weirdness around OSS interfaces that cannot be conditionally imported, I went ahead and just declared the types/interfaces on the UI as a way of moving forward with this implementation.


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3499)

genehynson (approved):

PR #3509, opened 2022-01-04
feat: multiAccount- adding switch button and supporting UI

jrenee42:
amends the menu, shows the account page with the name of the current logged in/active account.

if there is only one account available, then the switch button does not show. (there is a test for this situation).

when the 'switch' dialog is opened, the "switch" button is only active/enabled when an account other than the currently logged in account is selected.
when the account is actually switched, the browser is redirected to the location: {CLOUD_URL}/accounts/${newAccountId}
this is not possible to test locally, but this is what @randycoulman  has directed me to do.

the menu change (and thus the page) is only accessible when the feature flag 'multiAccount' is set to true.

here's the new page:
Screen Shot 2022-01-05 at 4 09 17 PM

here is a screencap: 

https://user-images.githubusercontent.com/3900960/148290624-b2ccdc13-49ae-4e2d-b078-387d4e4a9741.mov

note:  to simulate quartz; need to run quartz-mock locally to get to these pages locally.  see the quartz-mock readme for details:  https://github.com/influxdata/quartz-mock

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -617,6 +617,15 @@ export const flush = (): Cypress.Chainable<Cypress.Response<any>> => {
   }
 }
 
+/**
+ * numAccounts: this is the number of accounts that a particular user has.
+ * (if there is more than one account, then show the switch button,
+ * if only one account don't show the switch button)
+ *
+ * if this number is not supplied, then quartz-mock has a default value of 1 (1 account)
+ *
+ *  this is for the new multi-user/multi-account capability (adding to the UI in early 2022)
+ * */
 export type ProvisionData = {
   accountType?: AccountType
   hasData?: boolean
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -624,6 +624,7 @@ export type ProvisionData = {
   isOperator?: boolean
   operatorRole?: string
   isRegionBeta?: boolean
+  numAccounts?: number
 }
 
 /*
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -12,14 +12,8 @@ import {UserAccountContext} from 'src/accounts/context/userAccount'
 import {SwitchAccountOverlay} from 'src/accounts/SwitchAccountOverlay'
 
 const AccountAboutPage: FC = () => {
-  const {userAccounts, defaultAccountId, activeAccountId} = useContext(
-    UserAccountContext
-  )
+  const {userAccounts} = useContext(UserAccountContext)
   const [isSwitchAccountVisible, setSwitchAccountVisible] = useState(false)
-
-  console.log('got userAccounts???', userAccounts)
-  console.log('arghh, default account id?', defaultAccountId)
-  console.log('active acct id???', activeAccountId)
 
   const activeAcctName =
     userAccounts && userAccounts.filter(acct => acct.isActive)[0].name
@@ -61,8 +55,6 @@ const AccountAboutPage: FC = () => {
 }
 
 const AccountPage: FC = () => {
-  //todo:  look at userlistcontainer for a tabbed example!
-
   return (
     <Page titleTag={pageTitleSuffixer(['Account Settings Page'])}>
       <UserAccountProvider>
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -4,8 +4,8 @@ import {Button, IconFont, Overlay, Page} from '@influxdata/clockface'
 
 // Utils
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
-import UserAccountProvider from './context/userAccount'
-import AccountTabContainer from './AccountTabContainer'
+import UserAccountProvider from 'src/accounts/context/userAccount'
+import AccountTabContainer from 'src/accounts/AccountTabContainer'
 
 import {UserAccountContext} from 'src/accounts/context/userAccount'
 
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -12,11 +12,12 @@ import {UserAccountContext} from 'src/accounts/context/userAccount'
 import {SwitchAccountOverlay} from 'src/accounts/SwitchAccountOverlay'
 
 const AccountAboutPage: FC = () => {
-  const {userAccounts, defaultAccountId} = useContext(UserAccountContext)
+  const {userAccounts, defaultAccountId, activeAccountId} = useContext(UserAccountContext)
   const [isSwitchAccountVisible, setSwitchAccountVisible] = useState(false)
 
   console.log('got userAccounts???', userAccounts)
   console.log('arghh, default account id?', defaultAccountId)
+  console.log('active acct id???', activeAccountId)
 
   const switchAccount = () => {
     console.log('would switch account here.....')
@@ -43,7 +44,6 @@ const AccountAboutPage: FC = () => {
           <Overlay visible={isSwitchAccountVisible}>
             <SwitchAccountOverlay
               onDismissOverlay={handleDismissOverlay}
-              userAccounts={userAccounts}
             />
           </Overlay>
         </>
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -34,9 +34,6 @@ const AccountAboutPage: FC = () => {
     <Page titleTag={pageTitleSuffixer(['About', 'Account'])}>
       <AccountTabContainer activeTab="about">
         <>
-          <h1 data-testid="account-about--header">
-            hello world on the account settings page
-          </h1>
           {userAccounts && userAccounts.length >= 2 && (
             <Button
               text="Switch Account"
@@ -45,6 +42,10 @@ const AccountAboutPage: FC = () => {
               testID="user-account-switch-btn"
             />
           )}
+          <hr/>
+          <h2  data-testid="account-settings--header" > Account Details </h2>
+           <div> ack ack ack ack (4) </div>
+
 
           <Overlay visible={isSwitchAccountVisible}>
             <SwitchAccountOverlay onDismissOverlay={handleDismissOverlay} />
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -14,22 +14,27 @@ import UserAccountProvider from './context/userAccount'
 import {UserAccountContext} from './context/userAccount'
 
 const AccountAboutPage: FC = () => {
+  const {userAccounts, defaultAccountId} = useContext(UserAccountContext)
+
+  console.log('got userAccounts???', userAccounts)
+  console.log('arghh, default account id?', defaultAccountId)
+
   return (
-    <UserAccountProvider>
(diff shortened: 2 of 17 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3509)

src/accounts/SwitchAccountOverlay.tsx
```diff
@@ -0,0 +1,123 @@
...
+  InputToggleType,
+  Overlay,
+  Toggle,
+} from '@influxdata/clockface'
+
+// Metrics
+import {event} from 'src/cloud/utils/reporting'
+
+import {UserAccountContext} from './context/userAccount'
+import {ComponentStatus} from '../clockface'
```
ChitlangeSahas, line 21: Maybe this could come from `@influxdata/clockface`?

src/accounts/SwitchAccountOverlay.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const [newAccountId, setNewAccountId] = useState<number>(null)
+  const [buttonStatus, setButtonStatus] = useState(ComponentStatus.Disabled)
+
+  const {activeAccountId} = useContext(UserAccountContext)
+
+  const doSwitchAccount = () => {
+    onDismissOverlay()
+    event('multiAccount.switchAccount')
+
+    window.location.href = `${CLOUD_URL}/accounts/${newAccountId}`
```
ChitlangeSahas, line 90: Is this the same as using `history.push()`? If yes, do we have a preference between the two? I have seen the useHistory hook being used more often in our code. I could be wrong.

jrenee42: no, this is a redirect; to go to quartz-land and they do their magic.  this is what @randycoulman  directed me to do

ChitlangeSahas: Got it!

jrenee42:
also, am putting in a 'vanilla' no-multiAccount version of the b illing page in (what the page looks like before this change). and adjusting the tests as well.

src/accounts/AccountPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
+import UserAccountProvider from './context/userAccount'
+import AccountTabContainer from './AccountTabContainer'
+
+import {UserAccountContext} from 'src/accounts/context/userAccount'
+
+import {SwitchAccountOverlay} from 'src/accounts/SwitchAccountOverlay'
+
+const AccountAboutPage: FC = () => {
+  const {userAccounts} = useContext(UserAccountContext)
```
hoorayimhelping, line 15: what's the reasoning for using context here rather than a reducer?

src/accounts/AccountPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const [isSwitchAccountVisible, setSwitchAccountVisible] = useState(false)
+
+  const activeAcctName =
+    userAccounts && userAccounts.filter(acct => acct.isActive)[0].name
```
hoorayimhelping, line 19: is there guaranteed to be at least one active account? is it guaranteed to have a name?

src/accounts/SwitchAccountOverlay.tsx
```diff
@@ -0,0 +1,123 @@
...
+  InputLabel,
+  InputToggleType,
+  Overlay,
+  Toggle,
+} from '@influxdata/clockface'
+
+// Metrics
+import {event} from 'src/cloud/utils/reporting'
+
+import {UserAccountContext} from './context/userAccount'
```
hoorayimhelping, line 21:
```suggestion
import {UserAccountContext} from 'src/accounts/context/userAccount'
```

src/accounts/SwitchAccountOverlay.tsx
```diff
@@ -0,0 +1,123 @@
...
+  // need to change it to a number, so that the equality for
+  // active and isChecked works properly
+  // it is a string because that is how the toggle works
+  const onChange = strValue => {
+    const numAcct = parseInt(strValue)
+    onClickAcct(numAcct)
+    setSelectedAcct(numAcct)
+  }
+
+  const style = {marginBottom: 7}
```
hoorayimhelping, line 45: 👍

src/accounts/SwitchAccountOverlay.tsx
```diff
@@ -0,0 +1,123 @@
...
+const ToggleGroup: FC<ToggleProps> = ({onClickAcct}) => {
+  const {userAccounts, activeAccountId} = useContext(UserAccountContext)
+
+  const [selectedAcct, setSelectedAcct] = useState(activeAccountId)
+
+  // need to change it to a number, so that the equality for
+  // active and isChecked works properly
+  // it is a string because that is how the toggle works
+  const onChange = strValue => {
+    const numAcct = parseInt(strValue)
```
hoorayimhelping, line 40:
```suggestion
    const numAcct = parseInt(strValue, 10)
```

probably safe in 2022, but there's no guarantee that the user's JavaScript engine will parse it in decimal and this would be a really nefarious bug to track down.

src/accounts/context/userAccount.tsx
```diff
@@ -0,0 +1,97 @@
+// Libraries
+import React, {FC, useCallback, useEffect, useState} from 'react'
+// import {useParams, useHistory} from 'react-router-dom'
```
hoorayimhelping: ✂️

src/accounts/context/userAccount.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+        const activeAcctArray = data.filter(line => line.isActive)
+        if (activeAcctArray && activeAcctArray.length === 1) {
+          const activeId = activeAcctArray[0].id
+          setActiveAccountId(activeId)
+        }
+      }
+    } catch (error) {
+      // leaving this in for now while developing; will remove when feature is done
+      console.error('caught error...', error)
```
hoorayimhelping, line 72: I think merging this into master is good enough reason to remove the console log. This is probably better served as a notification to the user, provided it's an actionable error. If it's not actionable, a metrics `event` is probably more appropriate.

src/accounts/context/userAccount.tsx
```diff
@@ -0,0 +1,97 @@
...
+  userAccounts: UserAccount[]
+  handleGetAccounts: () => void
+
+  defaultAccountId: number
+  activeAccountId: number
+}
+//   todo: add to above when implementing:
+//    setDefaultAccountId: (id: number) => void
+
+// isActive: true is for the currently logged in/active account
```
hoorayimhelping, line 25: what does this comment refer to? I don't see `isActive` as part of this component

src/accounts/context/userAccount.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const [activeAccountId, setActiveAccountId] = useState<number>(null)
+
+  const dispatch = useDispatch()
+
+  const handleGetAccounts = useCallback(async () => {
+    try {
+      const resp = await getAccounts({})
+      if (resp.status !== 200) {
+        // set user account status to error;...TODO
+        throw new Error(resp.data.message)
```
hoorayimhelping, line 52: almost universally, we're doing these kinds of requests in an `api/` layer so the view layer can be related to rendering and interacting with the user. Some examples for inspiration:

https://github.com/influxdata/ui/blob/master/src/organizations/apis/index.ts
https://github.com/influxdata/ui/blob/master/src/dashboards/apis/index.ts
https://github.com/influxdata/ui/blob/master/src/templates/api/index.ts

This being at this layer leads me to reiterate my previous question: why context instead of redux? We typically use context for front-end related constructs (like date display preferences) and redux for api-level data.

src/accounts/context/userAccount.tsx
```diff
@@ -0,0 +1,97 @@
...
+        activeAccountId,
+        handleGetAccounts,
+      }}
+    >
+      {children}
+    </UserAccountContext.Provider>
+  )
+})
+
+export default UserAccountProvider
```
hoorayimhelping, line 96: this is already being exported. Kind of confusing to export it twice in different ways. Can you either remove this line and import the named export (which is what I prefer since it leaves less opportunity for ambiguity and makes code easier to search for globally since there's only one canonical name to be exported or imported) or only export it as the default export?

hoorayimhelping (commented):
had some questions / suggestions

cypress/support/commands.ts
```diff
@@ -624,6 +624,7 @@ export type ProvisionData = {
   isOperator?: boolean
   operatorRole?: string
   isRegionBeta?: boolean
+  numAccounts?: number
```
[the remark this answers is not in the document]
jrenee42: this is the number of accounts that a particular user has.  for provisioning data into quartz-mock
(if there is more than one account, then show the switch button, if only one account don't show the switch button)

i will add that to the documentation

src/accounts/AccountPage.tsx
jrenee42: i was following the pattern of the other items /pages that use quartz (billing, operator ro).  I use it in multiple places.

src/accounts/context/userAccount.tsx
jrenee42: done already

src/accounts/AccountPage.tsx
jrenee42: i believe there is guaranteed to be at least one  account with a name (quartz-mock guarantees this). will confirm with randy

src/accounts/context/userAccount.tsx
jrenee42: isActive is one of the flags returned with each 'account' object

src/accounts/context/userAccount.tsx
jrenee42: i chose context because all of the other quartz components  (billing, operator account, etc) use contexts

if you really want me to change it, i can but i am following a pattern.

src/accounts/context/userAccount.tsx
jrenee42: fixed

src/accounts/context/userAccount.tsx
jrenee42: ..moved to where the comment is relevant.

src/accounts/context/userAccount.tsx
jrenee42: switched to a metric

cypress/support/commands.ts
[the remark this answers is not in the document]
jrenee42: ...added

src/accounts/AccountPage.tsx
jrenee42: confirmed with @grace and @distortia that there is guaranteed to be at least one account (since the user has to be logged in to get to this call); and each account is guaranteed to have a name.  

I will add this note as a comment to the code.

hoorayimhelping (approved):

PR #3516, opened 2022-01-05
feat: add light mode toggle for shared and regular notebooks

asalem1:
This is still a WIP that needs to be feature flagged and cleaned up


src/shared/components/MarkdownMonacoEditor.tsx
```diff
@@ -45,7 +46,7 @@ const MarkdownMonacoEditor: FC<EditorProps> = ({
     <div className="markdown-editor--monaco" data-testid="markdown-editor">
       <MonacoEditor
         language={LANGID}
-        theme={THEME_NAME}
+        theme={theme ?? THEME_NAME}
```
asalem1, line 49: pretty sure i need to do this for all of em

PR #3533, opened 2022-01-08
feat: add error thresholds setting for visualization panels on notebooks

asalem1:
This PR is an initial attempt at allowing users to have UI-based alerting based on specific thresholds set in the UI. The story here is that we want a way to engage a user beyond the current alerting system so that, in the event a user is watching an updating notebook live in presentation mode, they will be able to somehow access that panel directly and take some sort of actionable steps to identifying and remediating that issue beyond the regular alerting system. We do this by allowing the user to set specific thresholds that will generate a QR code that can be scanned and deep-linked to in a notebook whenever the threshold appears.

As part of this, we will need to also integrate auto-refresh of notebooks so that notebooks can be run at a regular interval. In addition, in order to get this feature out live, we will need to rework the notebook sidebar so that it allows the user to interact with the UI in a meaningful way.

### DIFF
--- a/src/flows/components/Sidebar.scss
+++ b/src/flows/components/Sidebar.scss
@@ -1,7 +1,7 @@
 @import '@influxdata/clockface/dist/variables.scss';
 
 .flow-sidebar {
-  flex: 0 0 300px;
+  flex: 0 0 300px; // TODO(ariel): the sidebar needs more real estate for this to work. 550px is a good min, but a redesign is likely needed
   width: 300px;
   padding: 0 $cf-space-2xs;
 
@@ -42,6 +42,7 @@
 
 .flow-sidebar--submenu-wrapper {
   width: calc(100% - 16px);
+  padding-bottom: 20px;
 }
 
 .flow-sidebar--dropdownmenu {
--- a/src/flows/components/Sidebar.scss
+++ b/src/flows/components/Sidebar.scss
@@ -1,8 +1,8 @@
 @import '@influxdata/clockface/dist/variables.scss';
 
 .flow-sidebar {
-  flex: 0 0 300px; // TODO(ariel): the sidebar needs more real estate for this to work. 550px is a good min, but a redesign is likely needed
-  width: 300px;
+  flex: 0 0 calc(420px - $cf-space-2xs);
+  width: 420px - $cf-space-2xs;
   padding: 0 $cf-space-2xs;
 
   .sidebar--item {
--- a/src/flows/pipes/Visualization/ErrorThresholds.tsx
+++ b/src/flows/pipes/Visualization/ErrorThresholds.tsx
@@ -18,21 +18,12 @@ import {
 } from '@influxdata/clockface'
 import {PipeContext} from 'src/flows/context/pipe'
 import {event} from 'src/cloud/utils/reporting'
+import {
+  ThresholdFormat,
+  Threshold,
+} from 'src/flows/pipes/Notification/Threshold'
 
-enum ThresholdFormat {
-  Value = 'value',
-  Range = 'range',
-}
-
-type Threshold = {
-  value: number
-  type: string
-  field: string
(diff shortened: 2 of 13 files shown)


src/flows/components/Sidebar.scss
```diff
@@ -1,7 +1,7 @@
 @import '@influxdata/clockface/dist/variables.scss';
 
 .flow-sidebar {
-  flex: 0 0 300px;
+  flex: 0 0 300px; // TODO(ariel): the sidebar needs more real estate for this to work. 550px is a good min, but a redesign is likely needed
```
drdelambre, line 4: make this the same size as data explorer for now

src/flows/pipes/Visualization/ErrorThresholds.tsx
```diff
@@ -0,0 +1,378 @@
...
+
+type Threshold = {
+  value: number
+  type: string
+  field: string
+  max?: number
+  min?: number
+}
+
+export const THRESHOLD_TYPES = {
```
drdelambre, line 35: move this to a shared location to reuse the definitions from the alert panel, or, if you feel that there's isn't enough overlap in the concepts (that threshold means something slightly different here), remove the alert specific information (like condition)

src/flows/pipes/Visualization/view.tsx
```diff
@@ -152,6 +162,164 @@ const Visualization: FC<PipeProp> = ({Context}) => {
...
+
+    if (triggeredErrorThresholdMessage) {
+      // show the QR code if it exists
+      const url = new URL(
+        `${window.location.origin}${window.location.pathname}?panel=${id}`
+      ).toString()
+      const doc = document.getElementById(
+        `${triggeredErrorThresholdMessage}-${id}`
+      )
+      const size = doc?.offsetHeight * 0.4 ?? 100
```
drdelambre, line 263: what determines these consts?

src/flows/pipes/Visualization/view.tsx
```diff
@@ -152,6 +162,164 @@ const Visualization: FC<PipeProp> = ({Context}) => {
...
+                >
+                  <QRCode
+                    value={url}
+                    size={size}
+                    fgColor={InfluxColors.Fire}
+                    bgColor={InfluxColors.White}
+                    imageSettings={{
+                      src:
+                        'https://www.influxdata.com/wp-content/uploads/Favicon-blue-200x200.png',
+                      width: size * 0.2,
```
drdelambre, line 306: find a qr code generator that scales or wrap this logic inside of a component that updates this value automatically based on a ref's client rect instead of setting off a docs. the later bakes in unrelated styling relationships of the dom node's hierarchy, where as the former is informed  from them.

src/flows/pipes/Visualization/view.tsx
asalem1: This is for the size of the QR code. Totally arbitrary with regards to the size, just using a height relative to the offset of the pipe to generate the size of the QR code. Figured design might have input when we come to that point

src/flows/pipes/Visualization/view.tsx
asalem1: I mean, we can just have the size be set if that's your concern. Right now the QR code does scale to the size that we're passing in. Is your concern that we need to derive the size in the first place?

ui-ci-bot-influx:
Metric | Score | Level
performance | null | :hankey:
accessibility | null | :hankey:
seo | null | :hankey:
pwa | null | :hankey:
best-practices | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3533)

drdelambre (approved):

PR #3555, opened 2022-01-11
feat: allow self removal from account with ff on

eatondustin1:
Closes influxdata/quartz#4650

This allows users to remove themselves from an account when the `selfRemovalFromAccount` feature flag is on.

### DIFF
--- a/src/shared/selectors/flags.ts
+++ b/src/shared/selectors/flags.ts
@@ -45,7 +45,6 @@ export const OSS_FLAGS = {
   boardWithFlows: false,
   createWithFlows: false,
   leadWithFlows: false,
-  selfRemovalFromAccount: false,
 }
 
 export const CLOUD_FLAGS = {
@@ -92,7 +91,6 @@ export const CLOUD_FLAGS = {
   boardWithFlows: false,
   createWithFlows: false,
   leadWithFlows: false,
-  selfRemovalFromAccount: false,
 }
 
 export const activeFlags = (state: AppState): FlagMap => {
--- a/src/shared/selectors/flags.ts
+++ b/src/shared/selectors/flags.ts
@@ -45,6 +45,7 @@ export const OSS_FLAGS = {
   boardWithFlows: false,
   createWithFlows: false,
   leadWithFlows: false,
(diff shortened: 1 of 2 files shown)


eatondustin1:
This is my first PR for the UI, so let me know if there's anything I'm missing.

src/shared/selectors/flags.ts
```diff
@@ -45,6 +45,7 @@ export const OSS_FLAGS = {
   boardWithFlows: false,
   createWithFlows: false,
   leadWithFlows: false,
+  selfRemovalFromAccount: false,
```
lexun, line 48: @hoorayimhelping It looks like we only use this flag in `UserListItem`, which as far as I can tell is not in OSS. Should we omit the attribute here?

hoorayimhelping: @lexun yep, if this isn't used in OSS, it can be removed.

src/shared/selectors/flags.ts
```diff
@@ -91,6 +92,7 @@ export const CLOUD_FLAGS = {
...
```
hoorayimhelping, line 95: without getting into too much detail and history of what this file is, you don't need to add this here since `selfRemovalFromAccount` is a flag in ConfigCat. This file is more historical and was used before we had a legitimate feature flagging service. Flags here will only be settable via the user's localStorage, so they're machine dependent and can't be changed in the ConfigCat UI.

hoorayimhelping (approved):
I think you can safely remove the changes to `selectors/flags.ts`. Everything else looks good. to me though. I left a comment about the flagging process in your slack channel since it links to internal documents.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3555)

eatondustin1:
alchemy please merge

Issue #2867, opened 2021-10-12
CSV file upload doesn't work as I would expect it to


barbaranelson:
I select 'data' in the left nav, and then select sources, then file upload.  If I pick either flux-annotated CSV or line protocol, I get an option to select a bucket, and then a drag/drop interface to upload the file. If I pick CSV I get an option to select a bucket and then a whole bunch of instructions/examples of CSV that I can copy to clipboard, but I don't get an option to actually upload my file.  I seem to have to use the CLI and influx write.

Why is that?  If we don't let people upload a CSV file through the UI, then this should not be an option under 'file upload'.

The fix for this bug will require a slight redesign as follows:
- remove verbiage under `File Upload` section that says `at the click of a button` so that CSV panel is not misleading
- Change `CSV Data` title to include some kind of CLI indicator (needs design)
- Inside `CSV Data` tile, remove bucket selector
- Inside `CSV Data` tile, add CLI description (needs design)
- Add metric for when the `CSV Data` tile is clicked (if not already available) to distinguish from CSV Flux uploader

PR #3557, opened 2022-01-11
fix: file upload section fix for CSV Data 

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/2867

Following was addressed in this PR:

- remove verbiage under File Upload section that says at the click of a button so that CSV panel is not misleading
- Change CSV Data title to include some kind of CLI indicator (needs design)
- Inside CSV Data tile, remove bucket selector
- Inside CSV Data tile, add CLI description (needs design)
- Add metric for when the CSV Data tile is clicked (if not already available) to distinguish from CSV Flux uploader

https://user-images.githubusercontent.com/18511823/149013084-0cc92917-a824-49a9-970c-b08677683b21.mov

### DIFF
--- a/src/writeData/constants/contentFileUploads.ts
+++ b/src/writeData/constants/contentFileUploads.ts
@@ -23,20 +23,20 @@ export const WRITE_DATA_FILE_UPLOADS: FileUpload[] = [
     markdown: AnnotatedCSVMarkdown,
   },
   {
-    id: 'csv',
-    name: 'CSV Data',
-    image: CSVLogo,
-    markdown: CSVMarkdown,
-  },
-  {
     id: 'lp',
     name: 'Line Protocol',
     image: LPLogo,
     markdown: LPMarkdown,
+  },
+  {
+    id: 'csv',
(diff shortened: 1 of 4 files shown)


src/writeData/constants/contentFileUploads.ts
```diff
@@ -22,21 +22,21 @@ export const WRITE_DATA_FILE_UPLOADS: FileUpload[] = [
...
+    name: 'CSV Data (CLI)',
+    image: CSVLogo,
+    markdown: CSVMarkdown,
+  },
 ]
 
 export const search = (term: string): FileUpload[] =>
   WRITE_DATA_FILE_UPLOADS.filter(item =>
     item.name.toLowerCase().includes(term.toLowerCase())
-  ).sort((a, b) => a.name.toLowerCase().localeCompare(b.name.toLowerCase()))
```
ChitlangeSahas, line 42: We only have three choices, we can force an order for them, which we want to do with the CSV data tile.

hoorayimhelping (approved):

Issue #3477, opened 2021-12-20
Color Map: update colorMapping from the TimeMachine VEO to dashboard cell flow


ChitlangeSahas:
Going from the `VEO` (create or edit mode) to the dashboard cell, upon saving, we need to pass down the colorMapping properties from the VEO to the cells. i.e update the properties to the idpe from the VEO. Just like we would for when the color family changes.

PR #3561, opened 2022-01-11
feat: color mapping for when the cell is edited through the time machine

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/3477

Video:

https://user-images.githubusercontent.com/18511823/149031181-751dfc6d-0eea-42a4-98b1-0a19e89a0852.mov

### DIFF
--- a/src/timeMachine/components/Vis.tsx
+++ b/src/timeMachine/components/Vis.tsx
@@ -33,7 +33,7 @@ import {RemoteDataState, AppState, ViewProperties} from 'src/types'
 
 // Selectors
 import {getActiveTimeRange} from 'src/timeMachine/selectors/index'
-import {setViewProperties} from '../actions'
+import {setViewProperties} from 'src/timeMachine/actions'
 
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = ReduxProps
@@ -56,6 +56,8 @@ const TimeMachineVis: FC<Props> = ({
   cellID,
 }) => {
   const {type} = viewProperties
+  const dispatch = useDispatch()
+
   // If the current selections for `xColumn`/`yColumn`/ etc. are invalid given
   // the current Flux response, attempt to make a valid selection instead. This
   // fallback logic is contained within the selectors that supply each of these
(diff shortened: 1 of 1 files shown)


src/timeMachine/components/Vis.tsx
```diff
@@ -25,12 +25,15 @@ import {
...
 import {getTimeRangeWithTimezone} from 'src/dashboards/selectors'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {getColorMappingObjects} from 'src/visualization/utils/colorMappingUtils'
 
 // Types
 import {RemoteDataState, AppState, ViewProperties} from 'src/types'
 
 // Selectors
 import {getActiveTimeRange} from 'src/timeMachine/selectors/index'
+import {setViewProperties} from '../actions'
```
hoorayimhelping, line 36: minor nit:

```suggestion
import {setViewProperties} from 'src/timeMachine/actions'
```

inconsistencies, even minor ones, tend to draw attention to themselves, and someone reading the code might assume there is more meaning here than there should be.

src/timeMachine/components/Vis.tsx
```diff
@@ -90,6 +93,26 @@ const TimeMachineVis: FC<Props> = ({
       giraffeResult.table.getColumnType('_value') !== 'number' &&
       !!giraffeResult.table.length)
 
+  const dispatch = useDispatch()
```
hoorayimhelping, line 96: i'd say move this down into the 

```
if (isFlagEnabled('graphColorMapping')) {
```

block since it's only used there. I think this will make it slightly easier to follow since it's used where it's needed, and it will move a variable initialization out of code that doesn't need it.

hoorayimhelping (commented):
a couple of small tweaks

src/timeMachine/components/Vis.tsx
ChitlangeSahas: As per our slack discussion, this would cause the linter to throw an error because dispatch needs to be called from the top of the component.

hoorayimhelping (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3561)

PR #3578, opened 2022-01-13
fix: update Giraffe to update formatStatValue

### LINKED ISSUE #3519: InfluxDB Cloud: number format
The number format in the UI is broken in my instance of InfluxDB Cloud. A "Single Stat"cell for instance shows the following value:

111.474.245.00

The thousands separator is german style. Fine with me. But the decimal separator should be a "," instead of the ".".
Am I missing something?

Best,
Keith

TCL735:
Closes #3519 

- updates Giraffe to update `formatStatValue`
- copies `formatStatValue` from Giraffe for posterity
     _note_: `formatStatValue` should not be implemented by UI and will be removed in a separate pull request.

### DIFF
--- a/src/visualization/utils/formatStatValue.ts
+++ b/src/visualization/utils/formatStatValue.ts
@@ -35,18 +35,13 @@ export const formatStatValue = (
 
   digits = Math.min(digits, MAX_DECIMAL_PLACES)
 
+  const formatter = Intl.NumberFormat(undefined, {
+    minimumFractionDigits: digits,
+    maximumFractionDigits: digits,
+  })
+
   if (typeof value === 'number') {
-    const [wholeNumber, fractionalNumber] = Number(value)
-      .toFixed(digits)
-      .split('.')
-
-    localeFormattedValue = Number(wholeNumber).toLocaleString(undefined, {
-      maximumFractionDigits: MAX_DECIMAL_PLACES,
-    })
-
-    if (fractionalNumber) {
-      localeFormattedValue += `.${fractionalNumber}`
-    }
+    localeFormattedValue = formatter.format(Number(value))
   } else if (typeof value === 'string') {
     localeFormattedValue = value
   } else {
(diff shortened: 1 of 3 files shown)


src/visualization/utils/formatStatValue.ts
```diff
@@ -35,19 +35,14 @@ export const formatStatValue = (
...
+  const formatter = Intl.NumberFormat(undefined, {
+    minimumFractionDigits: digits,
+    maximumFractionDigits: digits,
+  })
+
+  if (isNumber(value)) {
+    localeFormattedValue = formatter.format(Number(value))
+  } else if (isString(value)) {
     localeFormattedValue = value
   } else {
```
TCL735, line 47: This is not needed to fix the issue, however, we are adding this change for posterity. We want this change in the history before the whole file is removed (in a separate pull request).

TCL735: Also, there is a minor difference. Giraffe uses utils `isNumber` and `isString` while here in the UI we simply use `typeof` to check the type.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3578)

asalem1:
Is there an issue to not have this in the UI do that we don’t lose connect later on?

src/visualization/utils/formatStatValue.ts
hoorayimhelping: >We want this change in the history before the whole file is removed (in a separate pull request).

why do we want that?

hoorayimhelping (approved):
code looks fine, just had a question about the process involved

TCL735:
> Is there an issue to not have this in the UI do that we don’t lose connect later on?

`formatStatValue` supports only SingleStat and Gauge. So, it should have been removed when those two visualizations' implementations got removed from the UI. (They live in Giraffe.)

We can reference that issue to complete the removal of `formatStatValue`

PR #3582, opened 2022-01-13
fix: fixes the issue that toggling the timerange didnt update the queries timerange

### LINKED ISSUE #3526: notebooks: When change time range, query runs with old time range not new one.
With a basic notebook with a query builder and a visualization, every time I change the time range, the notebook refreshes with the time previously selected - e..g used to be 24 hours, change it to 3 hours, will run with 24 hours.  Then change to 7 days, will run with 3 hours. 

(I actually thought it didn't rerun the query, and filed another issue to capture that, but the problem is that it runs it with the wrong time range.)

asalem1:
Closes #3526 

This PR addresses an issue where toggling the time range of a notebook would run the query for that notebook based on the previously selected timerange. This PR also cleans up some implementation details where we were setting the timerange of the notebook in two separate places, which was causing some weirdness with state syncing.

### DIFF
--- a/src/flows/context/flow.query.tsx
+++ b/src/flows/context/flow.query.tsx
@@ -1,4 +1,4 @@
-import React, {FC, useContext, useEffect, useState, useRef} from 'react'
+import React, {FC, useContext, useEffect, useRef} from 'react'
 import {useDispatch, useSelector} from 'react-redux'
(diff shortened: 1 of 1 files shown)


drdelambre (approved):
🔥

PR #3586, opened 2022-01-13
chore: update bucket selector to fetch all buckets

### LINKED ISSUE #3585: Make all Buckets available in Buckets Selector in Flows Query Builder
In file: src/flows/context/bucket.scoped.tsx, we are fetching 100 buckets which(buckets) are in-turn used across the flows. As we are only fetching top 100, there should be a problem for bigger clients that buckets over 100 will not be visible in the QueryBuilder. 

Get feature parity in Data Explorer and Flows query builders.

subirjolly:
Closes #3585

Pull down all the buckets in the BucketSelector using `getSortedBuckets` selector.

### DIFF
--- a/src/flows/context/bucket.scoped.tsx
+++ b/src/flows/context/bucket.scoped.tsx
@@ -2,16 +2,15 @@
 import React, {
   FC,
   createContext,
-  useContext,
   useEffect,
   useState,
   useMemo,
   useRef,
 } from 'react'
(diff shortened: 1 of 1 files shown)


drdelambre (changes requested):
this doesn't handle resource fetching correctly and will start in a broken state for all notebooks deeplinked to

Issue #3530, opened 2022-01-08
Tokens: Create token from client library pages takes you to the wrong page


barbaranelson:
If you start in Load Data, then select a client library (e.g. Python) and then select 'generate API token', and pick 'read/write token' it takes you to the list tokens page, and doesn't actually generate the token.

I think this is a mixup between the old and the new token UI, but we need to fix the client library pages to take you to the new token generation UI.

PR #3589, opened 2022-01-13
fix: use the redesigned generateTokensDropdown in WriteDataHelperTokens

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/3530

Uses the redesigned component in the `WriteDataHelperTokens`.

https://user-images.githubusercontent.com/18511823/149421961-0d584767-976b-45b7-aa55-4414dca502c2.mov

### DIFF
--- a/src/writeData/components/WriteDataHelperTokens.tsx
+++ b/src/writeData/components/WriteDataHelperTokens.tsx
@@ -28,6 +28,9 @@ import {AppState, ResourceType, Authorization} from 'src/types'
 
 import GenerateTokenDropdown from 'src/authorizations/components/GenerateTokenDropdown'
 
+import GenerateTokenDropdownRedesigned from 'src/authorizations/components/redesigned/GenerateTokenDropdown'
+import {isFlagEnabled} from '../../shared/utils/featureFlag'
+
 const WriteDataHelperTokens: FC = () => {
   const tokens = useSelector((state: AppState) =>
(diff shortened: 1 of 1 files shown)


TCL735 (approved):

src/writeData/components/WriteDataHelperTokens.tsx
```diff
@@ -102,7 +105,11 @@ const WriteDataHelperTokens: FC = () => {
         className="write-data--details-widget-title"
       >
         Token
-        <GenerateTokenDropdown />
+        {isFlagEnabled('tokensUIRedesign') ? (
```
Palakp41, line 108: I think we can remove this `isFlagEnabled` because we have this flag on in Production and I am right now in the process of removing this flag from everywhere. So, we can just render `GenerateTokenDropdownRedesigned`. What do you think?

ChitlangeSahas: I can do that! Do you think from a commit-history point, would it be better to remove this part after we've officially removed the flag? Just in case?

Palakp41: Hmm yeah, let's commit this, and I can pull latest and remove the flag in my pr :)

ChitlangeSahas: Or, maybe I can just wait until your flag removal PR is in and then merge this code without the flag? Let me know what you think!

PR #3610, opened 2022-01-18
chore: make the autorefresher responsible for the timeout functionality

asalem1:
This PR does some clean-up around the source of truth of controlling the timeout for the AutoRefresher within the dashboard by making the GlobalRefresher responsible for the timeout logic rather than having the view layer responsible for it. We're going to need to iterate on this design to handle multiple timers and intervals so that a user can have them set for multiple notebooks and dashboards without conflicts

### DIFF
--- a/src/dashboards/components/DashboardContainer.tsx
+++ b/src/dashboards/components/DashboardContainer.tsx
@@ -32,7 +32,6 @@ import {useRouteMatch} from 'react-router-dom'
 import {event} from 'src/cloud/utils/reporting'
 
 const DashboardContainer: FC = () => {
-  const timer = useRef(null)
   const dispatch = useDispatch()
   const {autoRefresh, dashboardID} = useSelector((state: AppState) => {
     const dashboardID = state.currentDashboard.id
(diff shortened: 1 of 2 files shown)


src/dashboards/components/DashboardContainer.tsx
```diff
@@ -46,25 +45,39 @@ const DashboardContainer: FC = () => {
...
+  }
 
+  const registerListeners = useCallback(() => {
     window.addEventListener('load', registerListeners)
     document.addEventListener('mousemove', registerListeners)
     document.addEventListener('keypress', registerListeners)
-  }, [dashboardID, autoRefresh.inactivityTimeout])
+  }, [])
+
+  const registerStopListeners = useCallback(() => {
```
drdelambre, line 62: i dont understand why you made this distinction

drdelambre: is this missing a globalautorefesher.stop or something?

drdelambre: omg nm. my brain read these as the same two functions. sorry

asalem1: Moving it up or simplifying it? Moved it up because the timeoutFunction below was complaining about invoking a function prior to it being declared. Besides that I stripped it down so it just removes the listeners

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3610)

Issue #3568, opened 2022-01-12
Update About, Billing, Users, and Usage layout


gunnaraasen:
With the new subnav layout for https://github.com/influxdata/ui/issues/3463, we will also want to update the layout for the affected pages.

![Account Billing](https://user-images.githubusercontent.com/628831/149198142-c198ee9a-65bc-4d28-b86e-a0cb030ff848.png)

![Org Settings](https://user-images.githubusercontent.com/628831/149198203-ef133f38-8349-459e-b3aa-56a801d61144.png)

![Org Users](https://user-images.githubusercontent.com/628831/149198238-264b64cf-2282-4232-aa28-168a56d10d97.png)

![Org Usage](https://user-images.githubusercontent.com/628831/149198298-0be93876-b5a9-4a3e-b5f6-6166860f7572.png)

[Figma design](https://github.com/influxdata/ui/issues/3463)

PR #3612, opened 2022-01-18
chore: adjust page/tabbing for multi account

### LINKED ISSUE #3463: Update submenu layout to include Accounts page
To expose the new account settings page (https://github.com/influxdata/ui/issues/3255), we would like to change the layout of the navbar's submenu to display info and links for Account/Org/User.

The navbar submenu for the user profile should show:

- A heading of "Account" with links to two pages for "Billing" and "Settings".
    - "Billing" should link the the existing billing page.
    - "Settings" should link to the Account settings page (https://github.com/influxdata/ui/issues/3255)
- A heading of "Organization" with links to three pages for "Members", "Usage", and "Settings".
    - The "Settings" page should link to the "About" page in the existing submenu.
- A heading with the current logged in user's email address with a link to logout.
    - If the length of the email address exceeds what can be displayed in the width of the submenu, we should truncate the end of the email address to fit within the submenu. (Also add a  "..." if the email is truncated.)

Nav Menu

[Updated Figma design](https://www.figma.com/file/BkocQXFhElANIPHt8ypxJL/Multi-Account-Redesign?node-id=2%3A848)

Existing submenu for reference:

jrenee42:
Closes #3463 #3568

renames pages,
(Organization: users->members, about->settings)
and puts the "Usage" page under the organization tabbed page

only shows if the flag is on.

just cosmetic; no functional changes.

### DIFF
--- a/cypress/e2e/cloud/billing.test.ts
+++ b/cypress/e2e/cloud/billing.test.ts
@@ -67,6 +67,16 @@ describe('Billing Page PAYG Users', () => {
             }).then(() => {
               cy.visit(`/orgs/${id}/billing`)
               cy.getByTestID('billing-page--header').should('be.visible')
+
+              // note:  when the 'multiAccount' feature flag is on/removed (so always on)
+              // remove the above line, replace with:
+
+              // cy.getByTestID('accounts-billing-tab').should('be.visible')
+              //
+              // cy.getByTestID('accounts-billing-tab').should(
+              //     'have.class',
+              //     'cf-tabs--tab__active'
+              // )
             })
           })
         })
--- a/src/organizations/containers/OrgProfilePage.tsx
+++ b/src/organizations/containers/OrgProfilePage.tsx
@@ -15,10 +15,10 @@ import DeleteOrgOverlay from 'src/organizations/components/DeleteOrgOverlay'
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
 import {getQuartzMe} from 'src/me/selectors'
 import DeleteOrgProvider from 'src/organizations/components/DeleteOrgContext'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Constants
 import {CLOUD} from 'src/shared/constants'
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 const OrgProfilePage: FC = () => {
   const quartzMe = useSelector(getQuartzMe)
--- a/src/organizations/containers/OrgProfilePage.tsx
+++ b/src/organizations/containers/OrgProfilePage.tsx
@@ -18,13 +18,16 @@ import DeleteOrgProvider from 'src/organizations/components/DeleteOrgContext'
 
 // Constants
 import {CLOUD} from 'src/shared/constants'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 const OrgProfilePage: FC = () => {
   const quartzMe = useSelector(getQuartzMe)
 
+  const subTitle = isFlagEnabled('multiAccount') ? 'Settings' : 'About'
+
   return (
     <>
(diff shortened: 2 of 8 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3612)

src/organizations/containers/OrgProfilePage.tsx
```diff
@@ -18,13 +18,16 @@ import DeleteOrgProvider from 'src/organizations/components/DeleteOrgContext'
 
 // Constants
 import {CLOUD} from 'src/shared/constants'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
```
ChitlangeSahas, line 21: Just a nit pick, but this could go under the `//Utils`

src/usage/UsagePage.tsx
```diff
@@ -9,19 +9,35 @@ import LimitChecker from 'src/cloud/components/LimitChecker'
 
 // Utils
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import OrgHeader from '../organizations/components/OrgHeader'
+import OrgTabbedPage from '../organizations/components/OrgTabbedPage'
```
ChitlangeSahas, line 14: and these could go under the `// components` and maybe with absolute path as that seems to be the pattern (?)

ChitlangeSahas (approved):
It looks good to me, just added a few comments about minor import reorders 👍

cypress/e2e/cloud/billing.test.ts
```diff
@@ -67,6 +67,16 @@ describe('Billing Page PAYG Users', () => {
...
+
+              // note:  when the 'multiAccount' feature flag is on/removed (so always on)
+              // remove the above line, replace with:
+
+              // cy.getByTestID('accounts-billing-tab').should('be.visible')
+              //
+              // cy.getByTestID('accounts-billing-tab').should(
+              //     'have.class',
+              //     'cf-tabs--tab__active'
+              // )
```
hoorayimhelping, line 79: I think instead of having comments here, it would be much more useful to add an issue to add an e2e test for the new functionality, which is the system we have to track future work. It makes the work needed visible to people who aren't in this code, like managers and directors and VPs.

Adding commented out code to be removed at some time in the future can lead to a lot of confusion and has a tendency to add tech debt without much benefit. If the system changes, or what needs to happen changes, this comment will be wrong, but it's unlikely it'll be updated.

src/usage/UsagePage.tsx
hoorayimhelping: >absolute path as that seems to be the pattern

➕

we have aliases in webpack to make absolute paths work. Most of the imports in the codebase follow this pattern.

Here's how you can update your IDE to default to absolute paths: https://blog.jetbrains.com/webstorm/2020/07/configuring-the-style-of-imports-in-javascript-and-typescript/

hoorayimhelping (commented):
Mostly LGTM, but a couple of suggestions

cypress/e2e/cloud/billing.test.ts
jrenee42: i tried adding e2e tests; and then they got removed because the feature flag setting is flaky.  
i'm worried that the issue will become lost, but if you feel strongly i can create an issue with this text in it.

jrenee42: @hoorayimhelping  and I discussed on slack; he does feel strongly, so in a new PR will be removing the comments, and put in an issue here: https://github.com/influxdata/ui/issues/3651

Issue #1587, opened 2021-06-02
Secrets: Link to Secrets List Page for Flux Editing


jzahnd:
We would like to link back to the secrets list page anywhere Flux can be edited from with in the application. 

**Acceptance Criteria**

* Anywhere a user can edit Flux, we provide a link back to the secrets list page. 

Design: https://www.figma.com/file/XulFKIibFgx9xtSL92rHlY/Injecting-Secrets-in-Notebooks?node-id=39%3A721

jzahnd:
@daviddkkim I was unable to find a design for this, do we have a mock somewhere that I passed over?

daviddkkim:
Chatted with @jzahnd - will provide design for https://app.zenhub.com/workspaces/applicationsui-unity-606b1041215d5f0017898847/issues/influxdata/ui/2213  and then this as a follow on !

PR #3624, opened 2022-01-19
feat: use secrets in notebook

### LINKED ISSUE #2539: Notebooks: Utilize Secrets in Notebooks
Now that we have the ability to add secrets in the UI, we would like to both surface this interface and allow users to inject secrets from within high impact areas of the application. 

Given that we are pushing for more update and usage on Notebooks in the UI, it would be a prime candidate for piloting the ability to inject secrets that have been created from the secrets interface

Design: https://www.figma.com/file/XulFKIibFgx9xtSL92rHlY/Injecting-Secrets-in-Notebooks?node-id=43%3A1060

wiedld:
Closes #2539, #1587 

## Overview of user tasks:
1. Within notebooks, inject a secret into the flux panel.
2. Choose dropdown option “Inject Secret” in the panel submenu.
3. Side toolbar, is modeled/based off of how the functions injection looks like.
4. Include ability to Add Secret from the side toolbar.

## Video:
Caveat: non-final video. (Is not the version to be posted in the slack channel.) Use google sso:
https://drive.google.com/file/d/1YEZS_E38oFxmYsUpX6ASQK9G6FbWzypM/view?usp=sharing

## Done:
- Cleaned up the RawFluxEditor's interactions with it's sidebar children.
    - renamed children to be more explicit List components.
    - `functions.tsx` --> `GroupedFunctionsList.tsx`
    - `dynamicFunctions.tsx` --> `FunctionsList.tsx`
    - took uniform view elements, and editor injection logic, and pulled into shared code.
- new shared view components:
    - `FluxInjectionOption.tsx`. rawFluxEditor-specific, shared component.
    - `FilterList.tsx`. shared component for src/flow.
- shared monaco editor logic --> into new EditorContext
    - can now inject new text into editor via shared inject() method
    - can make declarative enum choices. e.g. "same line" versus "own line" for injection
- Added new sidebar menu option: `SecretsList.tsx`.
    - used register() into the sidebar context, for RawFluxEditor.
- Extracted new shared component: `CreateSecretsOverlay`.
    - shared across app. so into src/shared
    - uses the OverlayController
- Confirmed that the behavior of inject functions, (which now shares context and components with inject secrets), is still the same. 
    - for inject-functions, tested with and without the flag 'fluxDynamicDocs'
- feature flagged: https://github.com/influxdata/idpe/pull/12879

## TODO (future):
- Add new cypress tests.


src/flows/components/SecretsList.tsx
```diff
@@ -0,0 +1,166 @@
...
+            </>
+          )}
+        </FilterSecrets>
+      </div>
+    ),
+    [searchTerm, secrets]
+  )
+}
+
+const mdtp = {
```
drdelambre, line 159: since this is a functional component, remove this connector pattern in favor of using the `useDispatch` hook within the render function

src/flows/components/Sidebar.tsx
```diff
@@ -233,6 +244,23 @@ const Sidebar: FC = () => {
           title: 'Export to Client Library',
           menu: <ClientList />,
         },
+        {
+          title: 'Inject Secret',
+          disable: () => {
+            const {type} = flow.data.byID[id]
+
+            if (type === 'rawFluxEditor' && !!editorContext.editor) {
```
drdelambre, line 252: instead of registering this item for all panel types and selectively hiding the option, use the register option for the sidebar in the rawfluxquery panel's view.tsx to attach the item when we use the flux panel. this allows for a cleaner separation between the panels, allowing us to depreciate entire folders of our code without tracing down where the code integrates with the rest of the system inn places like this. as a byproduct, it also cleans up the lazy loading bundles a bit. you can see that pattern in action here:
https://github.com/influxdata/ui/blob/master/src/flows/pipes/Visualization/view.tsx#L106-L127

src/flows/components/FlowPipe.tsx
```diff
@@ -12,15 +13,17 @@ const FlowPipe: FC<FlowPipeProps> = ({id}) => {
...
-        <Pipe
-          Context={props => {
-            const _props = {
-              ...props,
-              id,
-            }
-            return createElement(FlowPanel, _props)
-          }}
-        />
+        <EditorProvider id={id}>
```
drdelambre, line 16: only panels that have a monaco editor have the concept of having an editor, and only the flux editor has the ability to inject code into the monaco editor. this context would be better scoped to the rawfluxquery panel type and not to all panels

src/flows/components/SecretsList.tsx
```diff
@@ -0,0 +1,166 @@
...
+  context: editorContext,
+  onDismissOverlay,
+  onShowOverlay,
+}) => {
+  const [searchTerm, setSearchTerm] = useState<string>('')
+  const dispatch = useDispatch()
+  const secrets = useSelector(getAllSecrets)
+  const {data} = useContext(PipeContext)
+  const {editor, calcInsertPosition, updateText} = editorContext
+  const {queries, activeQuery} = data || {queries: []}
```
drdelambre, line 69: i'm starting to get the feeling that this component is so tightly coupled with the flux editor (right here we have a hard assumption about the format that the data is being stored in, a format specific to the flux editor), that we should move this component into that panel's definition and not in the shared `src/flows/components` folder

src/flows/components/SecretsList.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const query = queries[activeQuery]
+
+  const handleCreateSecret = () => {
+    event('Create Secret Modal Opened')
+    onShowOverlay('create-secret', null, onDismissOverlay)
+  }
+
+  const inject = useCallback(
```
drdelambre, line 77: is there enough overlap between this and the function injection to warrant some extra abstraction to merge those two mechanisms?

src/flows/context/editor.tsx
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  return useMemo(
+    () => (
+      <EditorContext.Provider
+        value={{
+          pipeId: id,
+          editor,
+          register,
+          calcInsertPosition,
```
drdelambre, line 96: seems like an `inject` function would have more value than grabbing the insert position and having to reimplement the injection logic using that info?

src/flows/pipes/RawFluxEditor/FunctionToolTipContent.tsx
```diff
@@ -0,0 +1,60 @@
+// Libraries
```
drdelambre, line 1: was this file just renamed or were there changes made to it?

src/secrets/components/SecretsTab.tsx
```diff
@@ -158,4 +154,11 @@ const SecretsTab: FC = () => {
   )
 }
 
-export default SecretsTab
+const mdtp = {
+  onShowOverlay: showOverlay,
+  onDismissOverlay: dismissOverlay,
+}
+
+const connector = connect(null, mdtp)
```
drdelambre, line 162: remove this connector pattern in favor of `useDispatch`

src/shared/components/secrets/CreateSecretOverlay.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const {onClose} = useContext(OverlayContext)
+
+  return (
+    <Overlay.Container maxWidth={750}>
+      <Overlay.Header
+        title="Add Secret"
+        onDismiss={onClose}
+        testID="create-secret-overlay-header"
+      />
+      <GetResources resources={[ResourceType.Secrets]}>
```
drdelambre, line 26: rough

drdelambre (changes requested):
should move the editor context closer to the monaco instance, tighten up the injection abstraction, remove the connect pattern, and self register to the sidebar to help with module independence.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3624)

src/flows/components/SecretsList.tsx
wiedld: done

src/flows/pipes/RawFluxEditor/FunctionToolTipContent.tsx
wiedld: the previous file had two components:
1.  component `FunctionTooltipContents` was unchanged, and placed inside a newly [explicitly] named file.
2. component `ToolbarFunction` was a reusable element, used for popover/injection element for each list item. this became a new shared component `FluxInjectionOption`.

The choice of keeping these two separate is so that the `FunctionTooltipContents` can be modified/extract into a new shared FilteredList component for src/flows:
* Then both inject-formula and inject-secrets will use this new FilterList.
* But the FilterList will take any child component (not just a FluxOption to be injected). So that we can reuse it again in src/flows.

That was the plan...and I can do that now (in this PR).

src/secrets/components/SecretsTab.tsx
wiedld: done

src/flows/components/SecretsList.tsx
wiedld: I was thinking the same thing. Making a generalized "inject `onOwnLine` versus `withinSameLine`" logic choice, and then all the text creation etc is pulled into the EditorContext functions.

But then the lose the ability to do bespoke changes.

Instead, I may create an inject() method in the EditorContext, which refers the helper functions (calcPosition and updateText). Then based on the option based to this new inject() function, it does `onOwnLine` versus `withinSameLine`.

src/flows/context/editor.tsx
wiedld: agreed. see comment above.

src/flows/components/SecretsList.tsx
wiedld: agreed. will do.

src/shared/components/secrets/CreateSecretOverlay.tsx
wiedld: good catch. copy/pasta of old component (a.k.a. slop).
is not used in the child component (which has the newer approach to dispatching).

src/flows/components/Sidebar.tsx
wiedld: done

src/flows/components/SecretsList.tsx
wiedld: done

src/flows/components/FlowPipe.tsx
wiedld: done.

src/flows/pipes/RawFluxEditor/FunctionToolTipContent.tsx
wiedld: Turns out that we had 3 lists (used in the RawFluxEditor sidebar):
1. SecretsList
2. functions
3. dynamicFunctions

`functions` was a filtered "list" which was not an array. Rather, it is grouped functions (by category) with each category have it's own list -- which is filtered. Therefore the logic for both filtering & rendering is different for the `functions`.  Plus it's name is confusing.

Here is the updated components:
1. FilterList in `src/flows/shared`. Handles search filtered arrays.
2. SecretList in `src/flows/pipes/RawFluxEditor`. uses FilterList.
3. functions is now GroupedFunctionsList, in `src/flows/pipes/RawFluxEditor`. filtering is object based, not array. So it's own code.
4. dynamicFunctions is now FunctionsList, in `src/flows/pipes/RawFluxEditor`. uses FilterList.

src/overlays/components/OverlayController.tsx
```diff
@@ -36,6 +36,7 @@ import AutoRefreshOverlay from 'src/dashboards/components/AutoRefreshOverlay'
 import CellCloneOverlay from 'src/shared/components/cells/CellCloneOverlay'
 import CustomApiTokenOverlay from 'src/authorizations/components/redesigned/CustomApiTokenOverlay'
 import DisplayTokenOverlay from 'src/authorizations/components/redesigned/DisplayTokenOverlay'
+import CreateSecretOverlay from 'src/shared/components/secrets/CreateSecretOverlay'
```
drdelambre, line 39: the rest of these (with the exception of cell cone for some reason) are all scoped to their functional folders, why move this stuff out of the existing `src/secrets` folder into a shared folder?

src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -39,86 +46,50 @@ const FluxMonacoEditor = lazy(() =>
...
+  useEffect(() => {
+    if (isFlagEnabled('fluxInjectSecrets')) {
+      register(id, [
+        {
+          title: 'RawFluxEditor actions',
+          actions: [
+            {
+              title: 'Inject Secret',
+              disable: () => false,
+              menu: <SecretsList context={editorContext} />,
```
drdelambre, line 67: just pass the inject here instead of the entire context

src/flows/context/editor.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+interface InjectionOptions {
+  header?: string | null
+  text: string
+  type: InjectionType
+}
+
+export interface EditorContextType {
+  editor: EditorType | null
+  register: (editor: EditorType) => void
```
drdelambre, line 33: this function isn't being used?

wiedld: register gets imported and renamed to `setEditorInstance`:
https://github.com/influxdata/ui/blob/a492f43537efa08e248ce662ca0dafd122607b34/src/flows/pipes/RawFluxEditor/view.tsx#L54

which then is the callback handed to the monaco editor:
https://github.com/influxdata/ui/blob/a492f43537efa08e248ce662ca0dafd122607b34/src/flows/pipes/RawFluxEditor/view.tsx#L135

src/flows/pipes/RawFluxEditor/view.tsx
wiedld: done

src/overlays/components/OverlayController.tsx
wiedld: done

src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -147,27 +119,33 @@ const Query: FC<PipeProp> = ({Context}) => {
...
   )
 
-  return (
-    <Context controls={controls}>
-      <Suspense
-        fallback={
-          <SpinnerContainer
-            loading={RemoteDataState.Loading}
-            spinnerComponent={<TechnoSpinner />}
+  return useMemo(
```
wiedld, line 122: Twas failing tests due to timeout. Was rendering 3 times on load.
Got it down to 2 loads (the minimum possible with the editor context callback) --> tests are now passing.

cypress/e2e/cloud/userAccounts.test.ts
```diff
@@ -113,12 +113,12 @@ describe('Account Page; user with two accounts', () => {
 
     cy.getByTestID('input--active-account-name').should('have.value', 'Influx')
 
+    cy.getByTestID('input--active-account-name').clear()
+    cy.getByTestID('input--active-account-name').should('have.value', '')
+
```
wiedld, line 118: removing race condition, which was causing circleci tests to fail.

src/flows/context/editor.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const [editor, setEditor] = useState<EditorType>(null)
+  const {data, update} = useContext(PipeContext)
+  const {queries, activeQuery} = data
+  const query = queries[activeQuery]
+
+  // Downstream components use these callbacks.
+  // Deeply nested, memoized components.
+  // Avoid vulnerability to memoization ref checks.
+  // (e.g. Sidebar remaining open).
+  const queryText = useRef(() => query.text)
```
drdelambre, line 60: this looks suss. what happens when you remove the useeffect and useref and access query.text directly?

src/flows/context/editor.tsx
```diff
@@ -0,0 +1,211 @@
...
+  useEffect(() => {
+    queryText.current = () => query.text
+  }, [query.text])
+
+  const register = (instance: EditorType) => {
```
drdelambre, line 65: instead of wrapping and mapping this, just expose it in the provider under a different name

src/flows/context/editor.tsx
```diff
@@ -0,0 +1,211 @@
...
+        })
+      }
+
+      editor.executeEdits('', edits)
+      updateText(editor.getValue())
+    },
+    [editor, calcInjectiontPosition, updateText]
+  )
+
+  return useMemo(
```
drdelambre, line 195: can remove this?

src/flows/pipes/RawFluxEditor/FunctionsList.tsx
```diff
@@ -0,0 +1,39 @@
...
+      onClick={onSelect}
+      extractor={fn => (fn as FluxToolbarFunction).name}
+      key={`${fn.name}_${fn.desc}`}
+      option={fn}
+      testID={fn.name}
+      ToolTipContent={FunctionTooltipContent}
+    />
+  )
+
+  return useMemo(
```
drdelambre, line 25: can remove this?

src/flows/context/editor.tsx
drdelambre: can you remove the renaming in line 54 and call the function directly?

src/flows/context/editor.tsx
wiedld: it gets the older (last) version of the text. and does the positional calculation from the last version.

src/flows/context/editor.tsx
wiedld: we also destruct the sidebar `register()` function. So one of the contexts, sidebar or editor, needs the register to be renamed on destructing. Or directly all it.

src/flows/context/editor.tsx
wiedld: will just pass out `setEditor` then

src/flows/context/editor.tsx
wiedld: instead, directly passing out `setEditor` prop.

src/flows/context/editor.tsx
wiedld: done.

drdelambre (approved):
:shipit:

PR #3639, opened 2022-01-19
fix: update flakey test

### LINKED ISSUE #3621: test `shared/dashboardsRefresh` disabled
This test has been disabled.

asalem1:
Closes #3621

This should resolve some flakey behavior under the current PR by removing an unnecessary assertion for the test. The failures have been happening when an event is dispatched from the test runner, which is causing some weirdness with state setting as it relates to the UI and the test runner


drdelambre (approved):
feels like the contains above handles this as well

PR #3641, opened 2022-01-19
chore: make the autorefresher responsible for registering events

asalem1:
The intent of this PR is the final clean up prior to integrating the AutoRefresh work with notebooks. The crux of what we're doing is making the AutoRefresher responsible for registering and unregistering the window events rather than having the view layer be responsible for it. The reason being that the view layer should define the conditions that those events should be triggered by, rather than defining the events and their conditions

### DIFF
--- a/src/dashboards/components/DashboardContainer.tsx
+++ b/src/dashboards/components/DashboardContainer.tsx
@@ -46,38 +46,21 @@ const DashboardContainer: FC = () => {
     '/orgs/:orgID/dashboards/:dashboardID/cells/:cellID/edit'
   )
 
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3641)

drdelambre (approved):
weird function names, but good change

PR #3643, opened 2022-01-20
fix: weird ts trick

drdelambre:
no idea why this passed all the ci pipelines and still broke production builds once in master, but that'll be for tomorrow


src/shared/components/DashboardRoute.tsx
```diff
@@ -65,10 +65,10 @@ class DashboardRoute extends PureComponent<Props> {
     }
 
     const params = new URLSearchParams(window.location.search)
-    const vars = (Array.from(params.entries()) || [])
-      .filter(([k, _v]) => /(^vars\[)(.+)(\]$)/g.test(k))
-      .reduce((acc, [k, v]) => {
-        acc[k.slice(5, -1)] = v
+    const vars = Array.from(params.keys())
```
drdelambre, line 68: none of this was needed, just the dom.iterable changes in the tsconfig. but after i wrote it, i liked it better, so here we are

TCL735 (approved):

wiedld (approved):
if it passes tests...then I assume the route still works.

PR #3658, opened 2022-01-20
feat: add ability to download notebook in full presentation as PDF/PNG

### LINKED ISSUE #3325: Notebooks: Add ability to download presentation mode as PDF/PNG
Currently, our exports as referenced in #3273 just take a screenshot of the page for Notebooks. We would like to provide an export view that is in presentation mode. It likely makes most sense to start this off with a PDF and Image view for sharing for users that customers cannot or do not want to add to their org access. 

Design: 
[Figma](https://www.figma.com/file/OEeEqe0M7eOr8r87CuXfFD/Download-PDF%2FImage-Presentation?node-id=26%3A1085)

Edit Mode
Notebook Edit Mode (4)

Presentation Mode
Notebook Presentation Mode (1)

appletreeisyellow:
Closes #3325 

Currently, our exports only support downloading single cells for Notebooks. We would like to provide an ability to export Notebooks in full presentation. This PR starts this off with a PDF and Image view for sharing for users that customers cannot or do not want to add to their org access.

## Demo

Download PDF:PNG


src/flows/components/header/index.tsx
```diff
@@ -139,6 +199,81 @@ const FlowHeader: FC = () => {
       .catch(err => console.error('failed to delete share', err))
   }
 
+  const canvasOptions = {
+    backgroundColor,
+    onclone: cloneDoc => {
+      cloneDoc.getElementById(currentID).style.paddingLeft = '12px'
+      cloneDoc.getElementById('pipe-list-hidden-header').style.display = 'block'
```
drdelambre, line 206: we should generalize this mechanism to using something like `data-download-show` and `data-download-hide` on the components, then using this part of the code to run `cloneDoc.querySelectorAll('data-download-hide').forEach(d => d.style.display = 'none')` so that we can decouple the header from the components that get rendered from the thing doing the rendering. for example, as we scale this out in it's current form, we'd have to keep adding element IDs here to cover all of the different elements we want to toggle and which way we want to toggle them. this creates an implicit relationship between this part of the code matching an id defined somewhere in a different file. when we change that file or delete it, we have to know to always come and check this file as well. by switching all of these to communicate our intended change during download mode through a common interface, we can decouple the definition and mutation parts and scale this to whatever complexity you want.

asalem1 (commented):
Boatie said it way better than I ever could, a couple nits beyond that

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3658)

drdelambre (approved):

PR #3659, opened 2022-01-20
feat: handle deep linking for a logged out user

### LINKED ISSUE #3569: Deep Linking: Handle deep linking without being logged in
### Background
Deep linking has trouble following the links after logging in. It fails on this line:

```ts
'/me/alerts': `/orgs/${this.props.org.id}/alerting`,
```

with the error: `this.props.org is null`. The issue seems to be getting access to the org id from redux in the 404 handler if the user hasn't visited a page. The good news is the org id is stored in local storage at this point, so it appears to be a bug in our state management, rather than a hard blocker.

### Acceptance Criteria
- make sure `props.org.id` is not null in the NotFound component after logging in

hoorayimhelping:
Closes #3569

Should handle deep linking for a user who is logged out and logs in. If `org` is null, which happens directly after they log in and hit the 404 handler, it will fetch the org and redirect them to the correct place.

Was kind of hard to test locally since we don't have an exact analogue for logging in locally. Will be able to do a better test behind a feature flag in production.

### DIFF
--- a/src/shared/components/NotFound.tsx
+++ b/src/shared/components/NotFound.tsx
@@ -16,6 +16,8 @@ import React, {Component, FC} from 'react'
 import {connect} from 'react-redux'
 import {getOrg} from 'src/organizations/selectors'
 
+import {getOrg as fetchOrg} from 'src/organizations/apis'
+
 import {withRouter, RouteComponentProps} from 'react-router-dom'
 
 import {AppState, Organization} from 'src/types'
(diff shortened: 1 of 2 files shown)


src/shared/components/NotFound.tsx
```diff
@@ -142,25 +144,44 @@ const NotFoundOld: FC = () => (
 )
 
 class NotFound extends Component<Props> {
-  componentDidMount() {
+  state = {
+    isFetchingOrg: false,
+  }
```
ChitlangeSahas, line 149: Just wondering, is this the same as setting the state in the `constructor()`. I was wondering if re-rendering the component will re-write the state in this case (?) 

This could be a confusion on my part as I am not totally sure how the lifecycle methods work.

ChitlangeSahas (approved):
Looks good to me! Just one comment out of curiosity on the `state` variable in the class.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3659)

src/shared/components/NotFound.tsx
hoorayimhelping: @ChitlangeSahas good question.

>Just wondering, is this the same as setting the state in the `constructor()`

Yep, it's the same. It's just shorthand for setting state when the component is constructed if you don't need access to `props`.

>By using an alternative class syntax, you can leave out the constructor and initialize the state as class field declaration. However, you don't have access to the `props` anymore.

From this blog post: https://www.robinwieruch.de/react-state-without-constructor/


>I was wondering if re-rendering the component will re-write the state in this case (?)

Nope, it won't re-write the state. The only thing that will rerun this code block is the component be reinitialized (not remounted) by react. I actually don't know the complete internals of react, but I believe a constructor is only called once by react, whereas `componentDidMount` can be called multiple times.

ChitlangeSahas: Got it! Thank you for clarifying!!

PR #3663, opened 2022-01-21
feat: disallow removal of last user from account

eatondustin1:
When there is only one user in an account, do not allow that user to be removed from the account.

### DIFF
--- a/src/users/components/UserList.tsx
+++ b/src/users/components/UserList.tsx
@@ -1,5 +1,6 @@
 // Libraries
 import React, {FC, useContext, useState} from 'react'
+import {useSelector} from 'react-redux'
 
 // Components
 import {Columns, Grid, IndexList} from '@influxdata/clockface'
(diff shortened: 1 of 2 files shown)


src/users/components/UserList.tsx
```diff
@@ -24,6 +32,9 @@ const UserList: FC = () => {
 
   const filteredInvites = filter(invites, ['email', 'role'], searchTerm)
 
+  const isDeletable = user =>
+    selfRemovalFromAccount ? users.length > 1 : user.id !== currentUserId
```
hoorayimhelping, line 36: if `currentUserId` ends being `undefined` due to `useSelector(getMe)` not returning anything with an id, that is still acceptable right?

hoorayimhelping (approved):
one question, but I'm assuming I know the answer.

src/users/components/UserList.tsx
eatondustin1: Yeah that should match the existing behavior from before we added the feature flag, so it should be as safe as it was before.

PR #3670, opened 2022-01-21
chore: react select type ahead var integration

### LINKED ISSUE #3657: use react-select instead of homegrown select for variable type-ahead dropdowns
the current typeAhead variable dropdown has a bunch of issues:

1) it doesn't do well with large (>200) items
2) appears to hang when it has an initial value when it is loading
3) it does show 'loading' when there is no initial value, but an animated spinner would be a good addition.

Using react-select should fix all these problems.

jrenee42:
Closes #3657 

accessible via the feature flag:  reactSelectVariableDropdown
will put in a separate idpe PR to add this (as well as put into config cat) before merging this PR

I went over the coloring and styles with andy moon.

the styling has been adjusted to look as similar as possible to the existing component.

when doing the review, compare the new file against the older, original component: https://github.com/influxdata/ui/blob/master/src/variables/components/TypeAheadVariableDropdown.tsx

screencap video:    (there are 900+ entries in the dropdown in the video below, btw)

https://user-images.githubusercontent.com/3900960/150604449-6c109c5d-e7a8-4b7c-a12c-e5b3b8201c52.mov

### DIFF
--- a/src/variables/components/VariableDropdownReactSelect.tsx
+++ b/src/variables/components/VariableDropdownReactSelect.tsx
@@ -5,6 +5,11 @@ import {connect, ConnectedProps} from 'react-redux'
 // Components
 import Select from 'react-select'
 
+import {
+  InfluxColors,
+} from '@influxdata/clockface'
+
+
 // Actions
 import {selectValue} from 'src/variables/actions/thunks'
 
@@ -90,7 +95,7 @@ class VariableDropdownReactSelect extends PureComponent<Props, MyState> {
     const inputBackground = 'black'
     const background = '#2c2d35'
     const focusColor = '#46454e'
-    const selectColor = '#ce58eb'
+    const selectColor = InfluxColors.Comet // '#ce58eb'
     const dropdownBackgroundFocused = '#5d5f6f'
     const dropdownBackgroundNotFocused = '#828497'
     const clearIndicatorHoverColor = 'white'
--- a/src/variables/components/VariableDropdownReactSelect.tsx
+++ b/src/variables/components/VariableDropdownReactSelect.tsx
@@ -25,7 +25,6 @@ type Props = OwnProps & ReduxProps
 
 interface MyState {
   shownValues: string[]
-  selectHappened: boolean
   loaded: boolean
 }
 
@@ -35,7 +34,6 @@ class VariableDropdownReactSelect extends PureComponent<Props, MyState> {
 
     const defaultState = {
       shownValues: props.values,
-      selectHappened: false,
       loaded: false,
     }
 
@@ -48,18 +46,15 @@ class VariableDropdownReactSelect extends PureComponent<Props, MyState> {
   }
 
   // set the 'shownValues' after loading, and
-  componentDidUpdate(prevProps, prevState) {
+  componentDidUpdate(prevProps) {
     const {values, status} = this.props
     const {values: prevVals, status: prevStatus} = prevProps
-    const {loaded, selectHappened} = this.state
-    const {selectHappened: prevSelectHappened} = prevState
+    const {loaded} = this.state
 
     let newState = {}
 
     const justLoaded =
       status === RemoteDataState.Done && prevStatus !== RemoteDataState.Done
-
-    const justSelected = selectHappened && !prevSelectHappened
 
     // this is for updating the values:
     // (only want this to run *once* when the values get loaded)
@@ -74,11 +69,6 @@ class VariableDropdownReactSelect extends PureComponent<Props, MyState> {
         shownValues: values,
         loaded: true,
       }
-    }
-
-    // for updating the selected value:
-    if (justSelected) {
-      newState = {...newState, selectHappened: false}
     }
 
     this.setState(newState)
@@ -251,14 +241,6 @@ class VariableDropdownReactSelect extends PureComponent<Props, MyState> {
     if (onSelect) {
       onSelect()
     }
-
-    const newState = {
-      ...this.state,
-      actualVal: selectedValue,
-      selectHappened: true,
-    }
-
-    this.setState(newState)
   }
 
   // show the 'loading' or 'no values' as a string (no input field yet!)
--- a/src/variables/components/VariableDropdownReactSelect.tsx
+++ b/src/variables/components/VariableDropdownReactSelect.tsx
@@ -30,7 +30,6 @@ type Props = OwnProps & ReduxProps
 // the typed val is what the user types in
 interface MyState {
   actualVal: string
-  selectIndex: number
   shownValues: string[]
   selectHappened: boolean
   menuOpen: MenuStatus
@@ -43,7 +42,6 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
 
     const defaultState = {
       actualVal: '',
-      selectIndex: -1,
       shownValues: props.values,
       selectHappened: false,
       menuOpen: null,
@@ -285,7 +283,6 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
       typedValue: selectedValue,
       actualVal: selectedValue,
       selectHappened: true,
-      selectIndex: -1,
     }
 
     if (closeMenuNow) {
--- a/src/variables/components/VariableDropdownReactSelect.tsx
+++ b/src/variables/components/VariableDropdownReactSelect.tsx
@@ -73,10 +73,10 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
 
     let newState = {}
 
-    const justLoaded = () =>
+    const justLoaded =
       status === RemoteDataState.Done && prevStatus !== RemoteDataState.Done
 
-    const justSelected = () =>
+    const justSelected =
       selectHappened &&
       !prevSelectHappened &&
       actualVal &&
@@ -84,7 +84,7 @@ class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
 
     // this is for updating the values:
     // (only want this to run *once* when the values get loaded)
-    if (justLoaded()) {
+    if (justLoaded) {
       newState = this.getInitialValuesAfterLoading()
     } else if (
       !loaded &&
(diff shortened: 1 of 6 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3670)

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+    if (shownValues && shownValues.length) {
+      realVals = shownValues.map(val => ({value: val, label: val}))
+    }
+    const placeHolderText = this.getPlaceHolderText('Select a Value')
+
+    const foreground = 'white'
+    const inputBackground = 'black'
+    const background = '#2c2d35'
+    const focusColor = '#46454e'
+    const selectColor = '#ce58eb'
```
hoorayimhelping, line 154: where do these hex values come from? Clockface vends a set of color variables in the `InfluxColors` enum. `ce58eb` is represented by `Comet` and `#ffffff` is represented by `White`, but I didn't find mention of `#2c2d35` or `#46454e` or `#5d5f6f` or `#828497`

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+        return {
+          ...provided,
+          color: foreground,
+          backgroundColor,
+          padding: 10,
+        }
+      },
+      clearIndicator: provided => {
+        // console.log("clear indicator...state??? ACK", state)
+        // const foregroundColor = state.isFocused? 'yellow' : 'cyan'
```
hoorayimhelping, line 173: ✂️

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+      placeHolder: provided => ({
+        ...provided,
+        color: foreground,
+        backgroundColor: inputBackground,
+      }),
+      indicatorSeparator: () => ({
+        display: 'none',
+      }),
+      dropdownIndicator: (provided, state) => {
+        const hoverBackground = state.isFocused ? '#5d5f6f' : '#828497'
```
hoorayimhelping, line 203: can you pull these two hex values out into variables at the top of the file along with the other colors please?

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+          color: foreground,
+          backgroundColor: inputBackground,
+        }
+      },
+      menu: provided => ({
+        ...provided,
+        zIndex: 15,
+        backgroundColor: background,
+      }),
+    }
```
hoorayimhelping, line 231: it looks like this structure can be pulled out into a separate function rather than being created in the render function since it it looks like the only dependencies it has on the color variables created before the start of it. That would be preferable since we wouldn't need to re-create this structure on every render

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+    const onChange = selection => {
+      if (selection && selection.value) {
+        this.handleSelect(selection.value)
+      }
+    }
+    // primary50 is the color that is used when an item is clicked on to change the selection
+    // have not found an alternative way (via the customStyles) to change that color)
+
+    //    const style={width:200}
```
hoorayimhelping, line 241: ✂️

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+  //   const longestItemWidth = Math.floor(
+  //     allVals.reduce(function(a, b) {
+  //       return a.length > b.length ? a : b
+  //     }, '').length * 10
+  //   )
+  //
+  //   const widthLength = Math.max(192, longestItemWidth)
+  //   const widthStyle = {width: `${widthLength}px`}
+  //   return widthStyle
+  // }
```
hoorayimhelping, line 281: ✂️

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+      shownValues: values,
+      loaded: true,
+    }
+  }
+
+  render() {
+    const {selectedValue, name, status} = this.props
+    const {shownValues} = this.state
+
+    const isDisabled = !shownValues || shownValues.length === 0
```
hoorayimhelping, line 140: Since we're checking length and calling `map` on shownValues, it's probably slightly safer to do:

```suggestion
    const isDisabled = !Array.isArray(shownValues) || shownValues.length === 0
```

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+  noValuesPresent = () => {
+    const {values} = this.props
+    return this.internalNoValuesPresent(values)
+  }
+
+  internalNoValuesPresent = valsToUse => {
+    const {status} = this.props
+
+    return (
+      status === RemoteDataState.Done && (!valsToUse || valsToUse.length === 0)
```
hoorayimhelping, line 343: same comment here about being a little bit safer to use

```suggestion
      status === RemoteDataState.Done && (!Array.isArray(valsToUse) || valsToUse.length === 0)
```

hoorayimhelping (commented):
I had some questions about things and suggestions about removing comments and using existing structures.

One thing I noticed is that this component seems to have a lot of state and composite state that is made up of smaller parts of state. Is it possible to simplify / cut down on the amount of state this component maintains?

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+class TypeAheadVariableDropdown extends PureComponent<Props, MyState> {
+  constructor(props) {
+    super(props)
+
+    const defaultState = {
+      actualVal: '',
+      selectIndex: -1,
+      shownValues: props.values,
+      selectHappened: false,
+      menuOpen: null,
```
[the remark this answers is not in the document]
jrenee42: null means that the menu status is not changed; just leave it alone.

src/variables/components/VariableDropdownReactSelect.tsx
jrenee42: i removed the // const style
line; but not the comments about primary50

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+     * then nothing will happen- the menu will not open)
+     */
+
+    if (menuOpen !== prevMenuOpen && menuOpen !== null) {
+      newState['menuOpen'] = null
+    }
+
+    // need to have this, as the 'onClickAwayHere' gets triggered *before*
+    // the selected value is set to the actualValue (it keeps re-using the original
+    // property)
```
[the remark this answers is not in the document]
jrenee42: referred to code removed (from the original implomentation); good catch!  removed the comment

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+      selectHappened: prevSelectHappened,
+      menuOpen: prevMenuOpen,
+    } = prevState
+
+    let newState = {}
+
+    const justLoaded = () =>
+      status === RemoteDataState.Done && prevStatus !== RemoteDataState.Done
+
+    const justSelected = () =>
```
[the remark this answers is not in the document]
jrenee42: changed to vars

jrenee42:
> I had some questions about things and suggestions about removing comments and using existing structures.
> 
> One thing I noticed is that this component seems to have a lot of state and composite state that is made up of smaller parts of state. Is it possible to simplify / cut down on the amount of state this component maintains?

i wrote this 2 months ago, based on code I wrote almost a year ago.  i can take a look to simplify it a little, but there is a lot of interaction that needs to take place and all the state is there for a reason.

src/variables/components/VariableDropdownReactSelect.tsx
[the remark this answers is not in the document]
jrenee42: actually...that code is moot.  i lost sight of why i was making this change...which is to have the component manage the state instead of the wrapper manage the state.  and a lot of the extra state that you mentioned can be discarded.  thanks for that insightful comment.  am removing *at least* the selectIndex and menuStatus/menuOpen right now.

src/variables/components/VariableDropdownReactSelect.tsx
jrenee42: i don't remember...i was going through and comparing colors and making the component look the same (as much as possible) as the existing component

src/variables/components/VariableDropdownReactSelect.tsx
```diff
@@ -0,0 +1,369 @@
...
+  onSelect?: () => void
+}
+
+type ReduxProps = ConnectedProps<typeof connector>
+type Props = OwnProps & ReduxProps
+
+// actualVal keeps track of the actual, selected variable
+// it allows the component to return to that if something was typed that is not in the selected list
+
+// the typed val is what the user types in
```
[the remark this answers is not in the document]
jrenee42: moot; removed

src/variables/components/VariableDropdownReactSelect.tsx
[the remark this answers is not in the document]
jrenee42: and removed the unneeded one as well

src/variables/components/VariableDropdownReactSelect.tsx
jrenee42: used influxcolors where I could; the rest I got from screencaps (matching colors)

hoorayimhelping (approved):
ah, this makes a lot more sense now that you explained what was going on. LGTM!

PR #3687, opened 2022-01-24
chore: removed some unused flags

asalem1:
This PR just cleans up some unused feature flags from the UI


src/shared/components/cells/Cell.tsx
```diff
@@ -168,10 +167,6 @@ class CellComponent extends Component<Props, State> {
       />
     )
   }
-
```
drdelambre, line 171: 🤣

drdelambre (approved):

PR #3696, opened 2022-01-25
chore: soft-delete underlying exported tasks.

### LINKED ISSUE #3303: Notebooks: When deleting a panel, soft delete the dependent task
When we delete a panel in Notebooks, we should soft delete the dependent task as the association would be broken with the Notebook / Panel. 

### Acceptance Criteria

* When a Notebook panel is deleted, any dependent tasks are soft deleted

subirjolly:
Closes #3303 

Set the status of underlying exported tasks from Notebooks to `inactive` asynchronously


drdelambre (changes requested):
TLDR; right code, wrong place 

adding the patch request to deactivate the task is totally how to soft delete a task, but this code is panel type specific, which the shared sidebar component shouldn't know anything about. we need some abstraction here to decouple the shared interaction and what happens specifically for the schedule panel. what we're trying to avoid here is a giant switch statement in all of the shared actions becoming a pattern and having all panels have to register a universal action (to reduce the problem of having a panel forget to add one)

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3696)

src/flows/context/flow.current.tsx
```diff
@@ -304,6 +305,19 @@ export const FlowProvider: FC = ({children}) => {
   }
 
   const removePipe = (id: string) => {
+    if (currentFlow.data.byID[id].type === 'schedule') {
```
drdelambre, line 308: Looks like you just moved the same problem somewhere else

src/flows/context/flow.current.tsx
```diff
@@ -304,6 +305,14 @@ export const FlowProvider: FC = ({children}) => {
   }
 
   const removePipe = (id: string) => {
+    const definition = PIPE_DEFINITIONS[currentFlow.data.byID[id].type]
+
+    // FIXME: Should i do this check or should i make sure that
```
drdelambre, line 310: Remove this comment

src/flows/pipes/Schedule/remove.ts
```diff
@@ -0,0 +1,14 @@
+import {patchTask} from 'src/client'
```
drdelambre, line 1: Nit: this should be in the definition, no need for another file

drdelambre (approved):
Clean up the comment

src/flows/pipes/Schedule/remove.ts
subirjolly: updated

PR #3699, opened 2022-01-25
chore: remove the AutoRefreshDropdown Component

### LINKED ISSUE #3647: Remove the AutoRefreshDropdown Component
This is fairly straightforward housekeeping. It looks like we're still referencing the AutoRefreshDown component in certain parts of the app, but it's currently disabled in every context, instead being toggled by the new overlay for setting the auto refresh context. We should remove the component, in addition to all references to it from the codebase.

Files / lines to remove:

- [x] https://github.com/influxdata/ui/blob/master/src/shared/components/dropdown_auto_refresh/AutoRefreshDropdown.tsx
- [x] https://github.com/influxdata/ui/blob/master/src/shared/components/dropdown_auto_refresh/AutoRefreshDropdown.scss
- [x] https://github.com/influxdata/ui/blob/master/src/style/chronograf.scss#L42
- [x] https://github.com/influxdata/ui/blob/master/src/timeMachine/components/RefreshDropdown.tsx
- [x] https://github.com/influxdata/ui/blob/master/src/timeMachine/components/Queries.tsx#L70
- [x] https://github.com/influxdata/ui/blob/master/src/timeMachine/components/Queries.tsx#L10
- [x] https://github.com/influxdata/ui/blob/master/src/dashboards/components/DashboardHeader.tsx#L251-L256
- [x] https://github.com/influxdata/ui/blob/master/src/dashboards/components/DashboardHeader.tsx#L7

As part of this removal, we should see if there's any other code associated with the old dropdown that can be removed.

appletreeisyellow:
Closes #3647

This PR removes all the code that is using AutoRefreshDropdown component, including test code.


cypress/e2e/shared/flows.test.ts
```diff
@@ -239,10 +239,6 @@ describe('Flows', () => {
       .should('be.visible')
     cy.getByTestID('table-cell cool').should('not.exist')
 
-    // This is a random validator that the autorefresh option doesn't pop up
-    // In Flows again without explicit changes
-    cy.getByTestID('autorefresh-dropdown--button').should('not.exist')
```
asalem1, line 244: we might want to update this to account for the new auto refresh work, but we'd need to set the feature flag for that at the outset of the test

cypress/e2e/shared/dashboardsRefresh.test.ts
```diff
@@ -572,7 +572,6 @@ describe('Dashboard refresh', () => {
           }
         })
 
-        cy.getByTestID('autorefresh-dropdown-refresh').click()
```
asalem1, line 575: see the other comment with regards to keeping this test for the new auto refresh work

appletreeisyellow: does it mean, instead of `.click()`, we put `.should(not.exist')` on `autorefresh-dropdown-refresh` for now?

asalem1: Not quite, since the autorefresh button does exist in dashboards and is not feature flagged, we should still have access to clicking on the button. What was removed in this PR was the old AutoRefresh button

cypress/e2e/shared/flows.test.ts
appletreeisyellow: do we want to test if the new auto refresh work here? or we just check if the new auto refresh button exists? 

i think we just want to check if the new button exists. it's better to test the functionality in another PR since this PR mainly removes the old auto refresh button.

asalem1: Just the button's existence, we can always work on adding some tests for the refresh functionality in the notebook's context later on

asalem1 (approved):
LGTM. A little confused by the additions in the dashboardsRefresh test, but if it doesn't cause flakey behavior then I'm ok with it

PR #3702, opened 2022-01-25
feat: add read-only tokens, remove token selection

### LINKED ISSUE #3283: Notebook sharing: generate a new read-only token when click on the share link
When https://github.com/influxdata/idpe/issues/12378 is implemented, it will break the current notebook sharing implementation, as we will no longer be able to get the value of the token from the API.  We need to have moved away from the current solution (to getting a JWT to use for the shared notebook queries) before this IDPE issue can be completed.

subirjolly:
Closes #3283 

### BEFORE
BEFORE

### AFTER
AFTER


subirjolly:
> #3283

https://github.com/influxdata/ui/issues/3283#issuecomment-1021476317

subirjolly:
> Where did we land with the auto-generated names of the tokens since that transition from sharing to confirming looks pointless without something significant happening

https://github.com/influxdata/ui/issues/3283#issuecomment-1021476317

drdelambre:
yeah why is there just a big blank step in that workflow now?

jzahnd:
@juliajames12 I'm with @drdelambre after the user clicks the share button, it is a really weird interaction. Mabe, the share button should go directly to the generation of the share link or a yes or no dialog (or explanation text)? But displaying the caution triangle with a yes and no button with no context feels strange.

drdelambre:
on the code side, why is the token generation the responsibility of the client and not the server when it creates the share link? then we wouldnt have a cleartext token on the client at all, as well as covering async failure cases better

src/flows/components/header/index.tsx
```diff
@@ -133,8 +108,25 @@ const FlowHeader: FC = () => {
...
+
+    return await createReadOnlyAllAuthorization(orgID, authName)
+  }
+
+  const generateLink = async () => {
     setLinkLoading(RemoteDataState.Loading)
+    let token: Authorization
+
+    try {
+      token = await createAndGetReadOnlyToken()
```
asalem1, line 122: why do we need to store the token response?

subirjolly: There's no particular reason. I will remove it.

asalem1:
> on the code side, why is the token generation the responsibility of the client and not the server when it creates the share link? then we wouldnt have a cleartext token on the client at all, as well as covering async failure cases better

I think this is the end goal right? As in, the API doesn't currently support that so we'd need to do that here first, then move it to the API once it's supported?

src/flows/components/header/index.tsx
asalem1: K, we also never set the linkLoading to done. Do we want to do that?

src/flows/components/header/index.tsx
```diff
@@ -133,13 +108,31 @@ const FlowHeader: FC = () => {
...
     setLinkLoading(RemoteDataState.Loading)
+    let token: string
+
+    try {
+      token = await createAndGetReadOnlyToken()
+    } catch (e) {
+      console.error('failed to create read-only token')
+      setLinkLoading(RemoteDataState.Error)
+
+      return
```
asalem1, line 128: i think my previous comment got removed by a force push, but shouldn't we be setting a `setLinkLoading` to Done at some point? Also, not sure why we're storing the `token` if it's never being used. Lastly, should we send an event when the read-only token fails to generate?

asalem1 (changes requested):

src/flows/components/header/index.tsx
subirjolly: I didn't do a force push. So, not sure what happened to your comment.

`setLinkLoading` is getting set to done [here](https://github.com/influxdata/ui/pull/3702/files#diff-b8e90c3282c7a31096b6bbfc1a34c6bb24040e637e7fcafbd1cecc5942b0210aR140) after this in `postNotebooksShare.then` and it's being set to error in `catch`..

subirjolly: ah that comment is still there.. on the resolved comment.

asalem1: K, that checks out. What about sending an event when the notebook fails to generate a token rather than just swallowing the error in a console.error?

asalem1 (approved):
Would be nice to have some events around the failure

drdelambre (changes requested):
wait, you guys missed the point of the api conversation

subirjolly:
> Would be nice to have some events around the failure

yup. was in a meeting. pushing this now.

drdelambre:
i'm not talking about solving interservice authorizations between services, migrations, or any api changes outside of this team's responsibilities. But that we have a single endpoint under our perview that already handles the mapping of the notebook to the token and the share link when the link is created. the only reason tokens were generated on the frontend in the first place was because we were given the product mandate that we were not to generate arbitrarily named tokens behind the scenes on the user's behalf, so they must select one themselves. now that we've dropped that product requirement, the correct place to put this is on that api endpoint so that any failure on subtasks related to the generation of a share link (generation of magic tokens being one) that result in the failed creation of the link, are surfaced to the user as a failure of the system to create that share link and not as a failed background task they knew nothing about.

subirjolly:
> 

Sure, that makes sense and I will create an issue to accomplish that as a Fast Follow-up.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3702)

subirjolly:
Folks, Need re-review!!

drdelambre:
having a hard time visualizing the code, but were the design changes to remove the empty token section integrated? seems like getting rid of the tokens would change the interaction to: clicking on the share button, create a share if no share exists, display the share link with the error message next to it.

subirjolly:
> 

That's what it is for now. There's a follow-up coming from Julia.

The current interaction is:
image

subirjolly:
@drdelambre is this good to go now?

subirjolly:
> having a hard time visualizing the code, but were the design changes to remove the empty token section integrated? seems like getting rid of the tokens would change the interaction to: clicking on the share button, create a share if no share exists, display the share link with the error message next to it.

Updated per our discussion.

interaction

drdelambre (approved):

src/flows/components/header/index.tsx
```diff
@@ -97,54 +88,44 @@ const FlowHeader: FC = () => {
...
+        setLinkDeleting(RemoteDataState.Error)
+        console.error('failed to delete share', err)
+      })
   }
 
   const generateLink = () => {
+    event('Show Share Menu', {share: !!share ? 'sharing' : 'not sharing'})
+
+    if (!!share) {
+      setSharing(true)
```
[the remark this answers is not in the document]
subirjolly: this is needed as you won't be able to see if the share already exists. If you are alluding to automatically showing the link if the share already exists, then it will always show the share link by default(if share exists). Is that what we want?

PR #3712, opened 2022-01-26
fix: add some defensive programming in there

drdelambre:
i can't reproduce it locally, but there's a case where the task is being passed into the taskrunscard as null. i decided to patch up those cases where task can be null to see if it helps alleviate the problem.


asalem1 (approved):

appletreeisyellow (approved):
looks good to me!

PR #3719, opened 2022-01-27
✨ Add redirect through Quartz after self-removal

abshierjoel:
After a user removes themself from an account, they need to be redirected back through quartz to be sent to the right cluster URL of the account they will move to. After deletion, they will be redirected to the `CLOUD_URL` environment variable, which will make them hits Quartz for their new cluster URL.


src/users/context/users.tsx
```diff
@@ -217,6 +221,8 @@ export const UsersProvider: FC<Props> = React.memo(({children}) => {
         window.dataLayer.push({
           event: GTM_USER_REMOVED,
         })
+
+        if (userId == currentUserId) window.location.href = CLOUD_URL
```
hoorayimhelping, line 225: I think you'll get dinged by the linter on this one - we require brackets on all control statements. I'm not sure if there's a lint rule for it but if there isn't we strongly prefer no single line control expressions, so instead of

```ts
if (userId == currentUserId) { window.location.href = CLOUD_URL }
```

```ts
if (userId == currentUserId) {
  window.location.href = CLOUD_URL
}
```

abshierjoel: Thanks much -- I got that changed now!

hoorayimhelping (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3719)

PR #3750, opened 2022-02-01
Replace RavenDB Logo

mateuszbartosik:
The two-colored logo was an old version. Within this PR, I update it to the current one.


TCL735:
For posterity, here is the old logo:
Screen Shot 2022-02-01 at 12 07 59 PM

TCL735:
And here is the new logo:
Screen Shot 2022-02-01 at 12 14 28 PM

TCL735 (approved):

PR #3754, opened 2022-02-01
chore: update potentially unsafe _blank links

hoorayimhelping:
Closes https://github.com/influxdata/idpe/issues/12694

- Adds two helpers:
  - A React Component wrapper around `<a>` to be used when the anchor tag has a `target` of `"_blank"` which adds `rel="noopener noreferrer"`
  - A function wrapper around `window.open` that has a `target` of `'_blank'` and adds `noopener,noreferrer`

- Updates any last parts of UI code that open targets with these helpers

Does **not** update existing links that open with a `rel` of `noopener` or `noreferrer`  to use the new helpers; this leaves those alone.


src/buckets/components/lineProtocol/LineProtocolHelperText.tsx
```diff
@@ -1,21 +0,0 @@
...
-        href={`https://docs.influxdata.com/influxdb/${DOCS_URL_VERSION}/reference/syntax/line-protocol/`}
-        target="_blank"
-      >
-        See Documentation
-      </a>
-    </p>
-  )
-}
-
-export default LineProtocolHelperText
```
hoorayimhelping, line 21: this file was not in use anywhere

TCL735 (approved):

PR #3758, opened 2022-02-01
fix: allow all three safari users to select simpleTable text

### LINKED ISSUE #3650: Unable to select text in data table in Safari
When I toggle "View Raw Data" in Explore and get the data table view, I'm unable to select any of the content in the table in Safari. After digging a bit more I discovered that this style is prohibiting me from doing so: https://github.com/influxdata/ui/blob/57744d0fc09c88cd9ae9c0a27135cc1e25d5d0bb/src/timeMachine/components/TimeMachine.scss#L6-L10

Here's the realized style as it's seen in my browser:

Screen Shot 2022-01-20 at 11 23 08 AM

When I disable this style, I'm able to select content in the table without issue.

hoorayimhelping:
Closes #3650

Allows Safari users to select text from simple tables

Screen Shot 2022-02-01 at 3 20 35 PM

### DIFF
--- a/src/visualization/types/SimpleTable/style.scss
+++ b/src/visualization/types/SimpleTable/style.scss
@@ -7,6 +7,7 @@
   display: flex;
   flex-direction: column;
   font-family: $cf-code-font;
+  -webkit-user-select: text;
   user-select: text;
 
   .cf-dapper-scrollbars {


TCL735:
> allow all three safari users to select simpleTable text

Four safari users. Three devs + one real user

TCL735 (approved):

PR #3767, opened 2022-02-02
fix(3523/3522/3524): python code gen formatting

### LINKED ISSUE #3522: Python Code Generator for Executing a Flux Query Not Properly Escaped
The "Execute a Flux Query" section of the Python code generator generates this line of Python:
```python
query = 'from(bucket: "rick+book's Bucket") |> range(start: -1h)'
```

This doesn't work because the apostrophe in the bucket name is not escaped. This is particularly bad because default bucket names contain apostrophes.

### LINKED ISSUE #3523: Python Generator Write Options Don't Work on Paste because they lack tabs
In the Python code generator, the first batch of code uses a ```with``` construct to scope the resources for a write operation. 

However, the following code blocks are not indented so do not actually work when pasted.

By default you get:
```python
from datetime import datetime

from influxdb_client import InfluxDBClient, Point, WritePrecision
from influxdb_client.client.write_api import SYNCHRONOUS

# You can generate an API token from the "API Tokens Tab" in the UI
token = "zDfcaZVAp3xMwI0hqlHwrEO4wfZzW8q5oHmX-qV9T3gsjVlEyPCHMTwMJTxlom5T9pvJ9wDmaQyMZi9xdSFnVg=="
org = "rick+book@influxdata.com"
bucket = "rick+book's Bucket"

with InfluxDBClient(url="https://eastus-1.azure.cloud2.influxdata.com", token=token, org=org) as client:
write_api = client.write_api(write_options=SYNCHRONOUS)

data = "mem,host=host1 used_percent=23.43234543"
write_api.write(bucket, org, data)

```
which obviously won't work because the last 3 lines are not indented. If you are using a python shell, you have to copy each line individually.

### LINKED ISSUE #3524: Python Code generator, Query Code Not indented properly, incomplete
It appears that the query code is meant to be scoped with a ```with``` construct, but the ```with``` is not included, and also the ensuing code is not indented. This is the code that works:

```python
with InfluxDBClient(url="https://eastus-1.azure.cloud2.influxdata.com", token=token, org=org) as client:
...     query = 'from(bucket: "rick+book\'s Bucket") |> range(start: -1h)'
...     tables = client.query_api().query(query, org=org)
...     for table in tables:
...             for record in table.records:
...                     print(record)
```

but this is the code that is provided:

```python
query = 'from(bucket: "rick+book's Bucket") |> range(start: -1h)'
tables = client.query_api().query(query, org=org)
for table in tables:
    for record in table.records:
        print(record)
```

(note also the lack of escaping mentioned in #3522 )

wiedld:
Closes #3523
Closes #3522
Closes #3524

### DONE:
- inter-related tickets.
- 3524 and 3523: all python block expressions, after initialization, needs to be indented.
- 3522: make sure to triple string quote the query, so no need to worry about escaping.
- copy/paste into python shell to make sure it worked.
      - from all of the options on the client libraries page.
      - from the export-to-client option off the flow pipes.

### NOT DONE:
- could not re-create one of the situations seen in https://github.com/influxdata/ui/issues/3524.
      - I assume this is the "export to client" from the flow pipes. 
      - The only way this could happen would be if we were missing templates for both the `def.initialize` and `def.executeFull` in the ClientCodeCopyPage.
      - not sure if I should do any hypothetical defensive coding here. I noted this on the ticket and pinged the issue reporter for more details.

### Screenshot of client libraries page
Screen Shot 2022-02-02 at 12 14 40 PM

### DIFF
--- a/src/writeData/clients/Python/dispose.example
+++ b/src/writeData/clients/Python/dispose.example
@@ -1,1 +1,1 @@
-client.close()
+    client.close()
--- a/src/writeData/clients/Python/execute.example
+++ b/src/writeData/clients/Python/execute.example
@@ -1,5 +1,5 @@
-query = """<%= query %>"""
-tables = client.query_api().query(query, org=org)
-for table in tables:
-    for record in table.records:
-        print(record)
+    query = """<%= query %>"""
+    tables = client.query_api().query(query, org=org)
+    for table in tables:
+        for record in table.records:
+            print(record)
--- a/src/writeData/clients/Python/execute.example
+++ b/src/writeData/clients/Python/execute.example
@@ -1,4 +1,4 @@
-query = '<%= query %>'
+query = """<%= query %>"""
 tables = client.query_api().query(query, org=org)
 for table in tables:
     for record in table.records:
--- a/src/writeData/clients/Python/write.0.example
+++ b/src/writeData/clients/Python/write.0.example
@@ -1,5 +1,5 @@
-write_api = client.write_api(write_options=SYNCHRONOUS)
+    write_api = client.write_api(write_options=SYNCHRONOUS)
 
-data = "mem,host=host1 used_percent=23.43234543"
-write_api.write(bucket, org, data)
+    data = "mem,host=host1 used_percent=23.43234543"
+    write_api.write(bucket, org, data)
 
--- a/src/writeData/clients/Python/write.1.example
+++ b/src/writeData/clients/Python/write.1.example
@@ -1,7 +1,7 @@
-point = Point("mem") \
-  .tag("host", "host1") \
-  .field("used_percent", 23.43234543) \
-  .time(datetime.utcnow(), WritePrecision.NS)
+    point = Point("mem") \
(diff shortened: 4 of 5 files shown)


wiedld:
Not related to these issues, but noted in the code:

We have a derivation in the token handling for the executeFull template, versus the initialize template.
- the initialize template uses the env variable `token = os.getenv("INFLUX_TOKEN")`.
- the executeFull template uses the token in the running app state, but the default is `<INFLUX_TOKEN>`.

I would suggest making the default be the same:
`os.getenv("INFLUX_TOKEN")`

Which means changing this:
https://github.com/influxdata/ui/blob/f06032ccf3265643d06f29820a8983e287828bde/src/writeData/components/WriteDataDetailsContext.tsx#L59
(As well as making the string quotes be added a bit differently.)

drdelambre (approved):
doesnt look like this'll bring down prod

wiedld:
ok. will make a ticket for the token handling inconsistency across code samples.

Issue #3404, opened 2021-12-08
Band graphs broken/do not render.


philjb:
## About the bug
Band graphs do not seem to work. Getting an error.

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. Query for data in explorer page
2. Switch from "Graph" to "Band" graph plot type
3. Observe that an "error" message appears and there are errors in the console.

**Expected behavior:**
Band graph plots. No error.

**Actual behavior:**
Error seen.

**Visual Proof:**

![Screen Shot 2021-12-08 at 11 44 20 AM](https://user-images.githubusercontent.com/13472206/145274411-e9a98225-1f34-4533-92a4-0362d1dec72f.png)

![Screen Shot 2021-12-08 at 11 44 09 AM](https://user-images.githubusercontent.com/13472206/145274478-84062c48-e101-4f13-ba35-99473b0ff79d.png)

## About your environment

**Environment info:**
cloud2, us-west-2

**Config:**
na

**Logs:**
na

**Performance:**
na

TCL735:
Step 1 is vague here. I can complete the steps and not produce an error.

@philjb is there a more specific step 1 that can reproduce this error consistently? In other words, do you have a query that can reproduce this error? Thanks in advance for your help.

philjb:
I think i have a reproducible scenario.

Using this line protocol for a single point:
```
weather,location=nashville,county=davison,state=tn,tz=central,music=country temp=7000i
```

and this query for it

```
from(bucket: "phil")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "weather")
  |> filter(fn: (r) => r["_field"] == "temp")
  |> filter(fn: (r) => r["location"] == "nashville")
//   |> yield(name: "mean")
```

Then band graph has an error but then if you uncommen thte yield the band graph plots fine.

I get the same results on safari anf firefox. It appears that the band chart needs the `yield` in the flux.

TCL735:
Band graph must have a yield. Furthermore, the yielded name must be in the options (open Customize) under the "Main Column Name". The Script Editor does not automatically update "Main Column Name" with the yielded name.

This is not a bug. Closing this.

philjb:
[...]

PR #3776, opened 2022-02-04
feat: update all versions of fromFlux

TCL735:
Part 1 of #3404 

- updates Giraffe to use the new `fromFlux`
- updates all other versions of `fromFlux` to minimize confusion
- the new property `resultColumnNames` can be safely ignored if not needed

### DIFF
--- a/src/shared/contexts/csv.worker.ts
+++ b/src/shared/contexts/csv.worker.ts
@@ -1,5 +1,6 @@
 import {csvParse, csvParseRows} from 'd3-dsv'
 
+const RESULT_COLUMN_INDEX = 1
 interface FromFluxResult {
   error?: Error
   // The single parsed `Table`
@@ -7,6 +8,9 @@ interface FromFluxResult {
 
   // The union of unique group keys from all input Flux tables
   fluxGroupKeyUnion: string[]
+
+  // All possible values of the "result" column
+  resultColumnNames: string[]
 }
 
 type Column =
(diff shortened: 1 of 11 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3776)

src/shared/contexts/csv.worker.ts
```diff
@@ -101,11 +106,29 @@ const fromFlux = (fluxCSV: string): FromFluxResult => {
     let columnDefault: any = ''
 
     for (const chunk of chunks) {
-      tableText = chunk
-        .split('\n')
-        .filter(line => !line.startsWith('#'))
-        .join('\n')
-        .trim()
+      const splittedChunk = chunk.split('\n')
```
hoorayimhelping, line 109: this looks like the logic from giraffe - is it possible to reuse that?

hoorayimhelping (approved):

src/shared/contexts/csv.worker.ts
TCL735: No, because this is the "native" version in the UI. There's a larger problem of having too many `fromFlux` parsers in the UI.

My goal here is simply to make the differences smaller, and not solve the larger problem of removing extraneous `fromFlux` parsers.

TCL735:
There is a minor update coming soon. The `fromFlux` parsers in the UI are fine, but the one in Giraffe makes an assumption that the `result` column is always the first column. According to the Flux team, the spec is that `result` can actually be in any position even though in the current implementation it is the first column.

PR #3801, opened 2022-02-07
feat: get all the version history for the user and have a hover state with a glimpse at the publish by and publish at

asalem1:
So KAP was giving me a hard time with a gif, here's a picture:

Screen Shot 2022-02-07 at 2 29 07 PM

### Background

As part of the publish lifecycle work, we're going to want to allow users to scroll through their previous version history in order to potential revert / clone from a specific version of a notebook's published history. 

The purpose of this PR is to allow a simple interface for users to be able to see at a glance their published history (by who and when). This does not fully integrate the full lifecycle of a user viewing their published history, but merely introduces the first step at that.

### DIFF
--- a/src/flows/components/header/PublishedVersions.tsx
+++ b/src/flows/components/header/PublishedVersions.tsx
@@ -1,0 +1,47 @@
+import React, {FC, useCallback, useContext, useEffect, useState} from 'react'
+import VersionBullet from 'src/flows/components/header/VersionBullet'
+import {FlowContext} from 'src/flows/context/flow.current'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {
+  getNotebooksVersions,
+  VersionHistories,
+} from 'src/client/notebooksRoutes'
+
+const PublishedVersions: FC = () => {
+  const {flow} = useContext(FlowContext)
+  const [versions, setVersions] = useState<VersionHistories>([])
+
+  const handleGetNotebookVersions = useCallback(async () => {
+    try {
+      const response = await getNotebooksVersions({id: flow.id})
+
+      if (response.status !== 200) {
+        throw new Error(response.data.message)
+      }
+
+      setVersions(response.data)
+    } catch (error) {
+      console.error({error})
+    }
+  }, [flow])
+
+  useEffect(() => {
+    if (isFlagEnabled('flowPublishLifecycle')) {
+      handleGetNotebookVersions()
+    }
+  }, [handleGetNotebookVersions])
+
+  if (versions.length === 0) {
+    return null
+  }
+
+  return (
+    <>
+      {versions.map(version => (
+        <VersionBullet key={version.id} version={version} />
+      ))}
+    </>
+  )
+}
+
+export default PublishedVersions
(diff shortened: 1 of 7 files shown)


src/flows/components/header/VersionBullet.scss
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.publish-version-bullet--button::before {
+  content: '';
+  height: 25px;
+  width: 4px;
+  transform: translate(-480%, 0%) rotate(-90deg);
+
+  @media screen and (max-width: $cf-grid--breakpoint-md) {
+    transform: translate(-380%, 0%) rotate(-90deg);
```
asalem1, line 16: the media is here because the component automatically adjusts sizes at the breakpoint.

src/flows/components/header/PublishedVersions.tsx
```diff
@@ -0,0 +1,47 @@
...
+    } catch (error) {
+      console.error({error})
+    }
+  }, [flow])
+
+  useEffect(() => {
+    if (isFlagEnabled('flowPublishLifecycle')) {
+      handleGetNotebookVersions()
+    }
+  }, [handleGetNotebookVersions])
```
asalem1, line 32: this doesn't refetch after a notebook has been published. Besides prop drilling or an emitting an event. This needs to be worked on in order to trigger a call of the version history so that we can refetch the notebook published history when a user publishes a new version. This might just be using context, but it might be something else

src/flows/components/header/index.tsx
```diff
@@ -153,7 +154,7 @@ const FlowHeader: FC = () => {
       try {
         const response = await postNotebooksVersion({id: flow.id})
 
-        if (response.status !== 204) {
+        if (response.status < 200 && response.status > 299) {
```
asalem1, line 157: not a fan of this, but we can deal with the API quirk of it returning a 200 later

drdelambre: Yeah, why are you still feeling like pushing this to prod?

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3801)

drdelambre (commented):
this needs to separate the concerns of the view layer and the data layer. we currently do this by using `useContext` to generate a provider that defines the interface for the data layer. the publishedversions shouldn't be holding on to the list of versions or be responsible for fetching them

src/flows/components/header/PublishedVersions.tsx
asalem1: Moved everything to context so we can refetch there

src/flows/components/header/PublishedVersions.tsx
```diff
@@ -0,0 +1,30 @@
...
+import VersionBullet from 'src/flows/components/header/VersionBullet'
+import {VersionPublishContext} from 'src/flows/context/version.publish'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
+const PublishedVersions: FC = () => {
+  const {versions, handleGetNotebookVersions} = useContext(
+    VersionPublishContext
+  )
+
+  useEffect(() => {
```
drdelambre, line 11: oh make the provider responsible for fetching

asalem1: K

drdelambre (approved):

PR #3806, opened 2022-02-08
fix: save a notebook if it is ephemeral before linking

drdelambre:
we've been waiting on trying the new task publish lifecycle because it didn't work when the notebook isn't saved. this isn't the default in dev mode, and i'm pretty sure we're inverting this pattern to make the publish button responsible for exporting resources, but it's currently the way prod works.

SO WE'RE SHORING IT UP UNTIL THEN

### DIFF
--- a/src/flows/pipes/Schedule/ExportTaskButton.tsx
+++ b/src/flows/pipes/Schedule/ExportTaskButton.tsx
@@ -1,5 +1,6 @@
 // Libraries
 import React, {FC, useContext} from 'react'
+import {useHistory} from 'react-router-dom'
 import {useDispatch, useSelector} from 'react-redux'
 import {
   ButtonType,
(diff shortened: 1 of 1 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3806)

src/flows/pipes/Schedule/ExportTaskButton.tsx
```diff
@@ -100,24 +104,41 @@ const ExportTaskButton: FC<Props> = ({
...
+              if (onCreate) {
+                onCreate(resp.data)
+              }
 
-        if (onCreate) {
-          onCreate(resp.data)
+              if (id !== flow.id) {
+                history.replace(`/orgs/${org.id}/notebooks/${id}`)
+              }
+            })
```
asalem1, line 141: sorry, my brain is having a hard time. Shouldn't we be updating the flow.list here?

drdelambre: the `add(flow)` is updating the list once theres a successful post. this just redirects to the new url so that it's not `/notebooks/from/default` up there when people start clicking around

asalem1 (approved):

PR #3816, opened 2022-02-09
chore(ecommerce): Removing delete account operator overlay

djcarpe:
Feature flag conditional for the `Delete Account` feature from the Account operator page. 

https://user-images.githubusercontent.com/59489977/153293867-bb714595-278f-4ae8-9e9b-c1e366c46ae3.mov


ChitlangeSahas (approved):
Looks good to me! I think we should link the issue that this PR is trying to address, in the description.

randycoulman (changes requested):
I believe that removing this button is gated on us having turned on the `freeAccountCancellation` feature flag everywhere first.

Given that, I think we either need to conditionally remove this button based on the feature flag value for now and actually delete the code later, or we need to hold off on merging this PR until we've completely turned on `freeAccountCancellation` and removed all code based on that feature flag.

djcarpe:
Good catch. Rather than letting this sit until we're ready I took your advice and feature flagged it so we can still test in staging.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3816)

src/operator/account/AccountViewHeader.tsx
```diff
@@ -12,12 +12,14 @@ import {
...
 } from '@influxdata/clockface'
 import {Link} from 'react-router-dom'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Contexts
 import {AccountContext} from 'src/operator/context/account'
 import {OperatorContext} from 'src/operator/context/operator'
 
 const AccountViewHeader: FC = () => {
+  const hideDeleteAccountButton = !isFlagEnabled('freeAccountCancellation')
```
randycoulman, line 22: This feels backwards to me if I'm reading this right (and with the logic below). Should this be named `showDeleteAccountButton` instead?

randycoulman (commented):
I'll let a UI team member give final approval, but one comment below.

randycoulman (commented):
This seems OK, but given that we're not sure we're moving forward with free account cancellation, it may not be worth merging at this time because we'll just have to remove it again later if we ultimately decide not to move forward.

Maybe let's wait until we finish scoping out what we're doing with cancellation before we merge this.

djcarpe:
This work is out dated.

PR #3819, opened 2022-02-10
fix: expose env var to build

drdelambre:
the amplitude key isn't showing up on the other side of the build. seeing if this is the issue


drdelambre:
yeah, that's why i thought it would work! but i guess i mixed up my local dev builds with my local prod builds

Issue #3630, opened 2022-01-19
Alerts: HTTP Notification endpoint still sending notifications after deactivation 


karel-rehor:
## About the bug

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. In AWS setup a simple express HTTP listener (example below)

In Influxdb UI

1. Add some test datapoints to a bucket with some timestamps going into the future for at least several minutes
2. Create a Threshold check that will trigger a Crit status based on the datapoints
3. Create an HTTP Notification Endpoint directed to the listener in AWS 
4. Create a notification rule, that binds together the check and the endpoint
5. Wait to verify 
   5.1. in AWS console that notifications are received by the listener 
   5.2 and in the UI that they are part of the notification history
6. In Influxdb UI deactivate the Notification Endpoint 
7. Add some more test datapoints with timestamps in the future that will trigger a Critical status notification 
8. Wait for ten minutes or for the next scheduled run of the Notification Rule

**Expected behavior:**

Expected the deactivated endpoint to stop sending notifications.  

**Actual behavior:**

After roughly ten minutes (default scheduler) notifications are appearing in the console of the AWS listener.
When then checking the notification history, these post-deactivation notifications are listed. 

**Visual Proof:**

## About your environment

**Environment info:**

Testing against K8S-IDPE remocal with up to date copy of UI project 

Latest commit in UI 
```
commit 019ec01d3ca48826c02777043ddd76cb47ac7416 
Author: karel-rehor <karel-rehor@users.noreply.github.com>
Date:   Wed Jan 19 10:13:01 2022 +0100
```

Latest commit in K8S-IDPE
```
commit ead1db508e2afa1c713e3872b6419b3e8c771be7 
Author: Hercules Mango Churchill <alchemy@influxdata.com>
Date:   Wed Jan 19 09:08:24 2022 +0000

```

**Config:**

**Logs:**

### Additional artifacts

*Simple Express listener*
Deployed on AWS EC2 Micro
```javascript
const express = require('express')
const app = express()
const port = 3000

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.post('/', (req,res) => {
        console.log('POST')

        let body = ''

        req.on('data', (data) => {

                console.log(`DATA: ${data}`)
                body += data
        })

        req.on('end', () => {
                res.writeHead(200, {'Content-Type': 'text'})
                res.end(`received: \n ${body} \n`)
        })

//      res.send('POST received\n')
})

[...]

PR #3832, opened 2022-02-11
fix(3629): cannot create rules from deactive endpoints

### LINKED ISSUE #3629: Alerts: deactivated endpoints are available when defining notification rule
## About the bug

**Steps to reproduce:**

In Alerts
1. Create a Threshold Check 
2. Create one or more endpoints
3. Deactivate an endpoint
4. Open Notification Rules
5. Create a new Rule, opens a popup
6. Click on the Endpoints selector in the popup

Alternatively - since it appears all Alert items are fetched together when opening the Alerts page

1. Create a Threshold Check 
2. Create one or more endpoints
3. Deactivate an endpoint
4. Switch to another page in the UI, e.g. Buckets
5. Return to Alerts - To refresh all JSON documents
6. Open Notification Rules
7. Create a new Rule, opens a popup
8. Click on the Endpoints selector in the popup

**Expected behavior:**

Expected that deactivated endpoint would not be selectable, and that either
   * It would not appear in the selector for available endpoints
   * Or, that it would appear in the list as greyed or otherwise disabled

**Actual behavior:**

I am able to select the deactivated endpoint. 

**Visual Proof:**

NotificationEndpDisabled01

## About your environment

**Environment info:**

Running against latest k8s-idpe remocal on Ubuntu 20.04. 

Latest UI commit 
```
commit 019ec01d3ca48826c02777043ddd76cb47ac7416
Author: karel-rehor <karel-rehor@users.noreply.github.com>
Date:   Wed Jan 19 10:13:01 2022 +0100
```
Latest K8s-IDPE
```
commit ead1db508e2afa1c713e3872b6419b3e8c771be7
Date:   Wed Jan 19 09:08:24 2022 +0000

```

**Config:**

**Logs:**

**Performance:**

```sh
iostat -xd 1 30 > iostat.txt
```

wiedld:
Closes #3629 
Problem was that in the Alerts part of the application, a disabled endpoint was still available to create a Notification Rule.

### Done:
* make selector for getAllActiveEndpoints()
* use the new state connector pattern in the components touched (NewRuleOverlay, RuleMessage, RulesColumn, AlertsColumn)
* use our new overlay controller part in the overlay touched (NewRuleOverlay).
* confirmed worked (see video below).
* confirmed could still create function e2e alert notifications.

### Proof that works:
https://user-images.githubusercontent.com/10232835/153638262-5c4a79e7-4aa3-4ff0-9f5e-1ffd4bfd5569.mov

### DIFF
--- a/src/alerting/components/AlertingIndex.tsx
+++ b/src/alerting/components/AlertingIndex.tsx
@@ -17,7 +17,6 @@ import GetAssetLimits from 'src/cloud/components/GetAssetLimits'
 import RateLimitAlert from 'src/cloud/components/RateLimitAlert'
 import GetResources from 'src/resources/components/GetResources'
 import EditCheckEO from 'src/checks/components/EditCheckEO'
-import NewRuleOverlay from 'src/notifications/rules/components/NewRuleOverlay'
 import EditRuleOverlay from 'src/notifications/rules/components/EditRuleOverlay'
 import NewEndpointOverlay from 'src/notifications/endpoints/components/NewEndpointOverlay'
 import EditEndpointOverlay from 'src/notifications/endpoints/components/EditEndpointOverlay'
@@ -168,7 +167,6 @@ const AlertingIndex: FunctionComponent = () => {
           path={`${alertsPath}/checks/:checkID/edit`}
           component={EditCheckEO}
         />
-        <Route path={`${alertsPath}/rules/new`} component={NewRuleOverlay} />
         <Route
           path={`${alertsPath}/rules/:ruleID/edit`}
           component={EditRuleOverlay}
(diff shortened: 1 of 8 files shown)


asalem1 (approved):
LGTM. Way to go on cleaning up some old code and grasping the overlay context usage. Out of curiosity, what happens when a user creates an endpoint and the corresponding rule has been deactivated?

wiedld:
> LGTM. Way to go on cleaning up some old code and grasping the overlay context usage. Out of curiosity, what happens when a user creates an endpoint and the corresponding rule has been deactivated?

Response to @asalem1 

**Endpoints** and Alert **Checks** are two separate entities, which exist separately from each other.
Notification **Rules** act to tie the Endpoint to the Alert Check; it does so by creating another task (under the cover).
Inactivation/Deletion of the notification rule do not delete the endpoint itself. It only removes the rule.

You are hitting on very important use cases, such as when an Endpoint is deactivated --> does the notification rule also stop? The product answer is that the rule should halt (because the endpoint is deactivated). However, we have a platform bug: 
see the UI ticket (where we confirmed that it occurs): https://github.com/influxdata/ui/issues/3630
and then kicked over to idpe: https://github.com/influxdata/idpe/issues/13160

Issue #3757, opened 2022-02-01
Multi Account: Nav bar Account Information design fix


ChitlangeSahas:
The Nav bar on the left that displays the account info needs to be tweaked. 

**Expected**:
Account Name
Org Info

**Current:**
Account Email
Org Info

ChitlangeSahas:
Done.

PR #3834, opened 2022-02-11
fix: update the avatar name when account is switched

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/3757
Depends on: https://github.com/influxdata/idpe/issues/13230

Bug fix is behind a feature flag. `avatarNameForMultiAccountFix`.

https://user-images.githubusercontent.com/18511823/153654020-c95c9abd-1795-45d2-96c4-7f5505532669.mov

### DIFF
--- a/src/me/actions/thunks/index.ts
+++ b/src/me/actions/thunks/index.ts
@@ -5,7 +5,7 @@ import {Dispatch} from 'react'
 
 // API
 import {getMe as apiGetApiMe} from 'src/client'
-import {getMe as apiGetQuartzMe} from 'src/client/unityRoutes'
+import {getAccounts, getMe as apiGetQuartzMe} from 'src/client/unityRoutes'
 // Utils
 import {gaEvent, updateReportingContext} from 'src/cloud/utils/reporting'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3834)

src/me/actions/thunks/index.ts
```diff
@@ -28,12 +28,27 @@ export const getMe = () => async (
...
+    if (
+      isFlagEnabled('multiAccount') &&
+      isFlagEnabled('avatarNameForMultiAccountFix')
+    ) {
+      const resp = await getAccounts({})
+
+      if (resp.status !== 200) {
+        throw new Error(resp.data.message)
+      }
+      user = resp.data.filter(account => account.isActive)[0]
```
hoorayimhelping, line 42: i think [`find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) is made specifically for this.

note, if for some reason there's no active account at this stage, this might cause errors / crashes since code below this relies on `user` existing and having an `id` on it

hoorayimhelping (commented):
just one comment - you have move context on this than I do though, so my concerns may not be a big deal

src/me/actions/thunks/index.ts
ChitlangeSahas: Just pushed the `find` fix. 

For the issue of none of the account being active, there is an assumption throughout that atleast one account will be active. 

This comment confirms that: https://github.com/influxdata/ui/blob/master/src/accounts/AccountPage.tsx#L60-L68

hoorayimhelping (approved):

PR #3835, opened 2022-02-11
fix: token issue when creating specific/all buckets token

### LINKED ISSUE #3600: Client library pages: don't update when create a new bucket or token
When I go into a client library page, the UI defaults to selecting one of my existing tokens and one of my existing buckets. If I choose to create a new bucket, or a new token, the bucket or token is created, but the UI continues to select the previous token or bucket. It should change to selecting the new bucket or token, as given the context we can assume the user wants to use the new token or bucket.  See video for details.

https://user-images.githubusercontent.com/13143006/149570228-d39a7a8a-c5c2-4a51-93e0-de6a9ee0282c.mov

Palakp41:
Closes #3600 
When creating a custom api token and a user selects `read` permissions for a specific bucket but then toggles `write` permissions for all buckets, it creates the token for `bucket: read` and `all buckets: write` and fails horribly when trying to view that token. For now, I have added the logic that when all buckets `read/write` toggle happens, we want to keep the specific read/write selections and the accordion open. We will collapse the accordion only when both toggle all for read and write are checked.. Just trying to fix this issue instead of failing terrible in the UI. This is going to be discussed with Balaji and Daniel to get proper design and a new issue will be created to track the progress.

Screen Shot 2022-02-15 at 1 28 59 PM

### DIFF
--- a/src/authorizations/components/CustomApiTokenOverlay.tsx
+++ b/src/authorizations/components/CustomApiTokenOverlay.tsx
@@ -122,13 +122,11 @@ const CustomApiTokenOverlay: FC<Props> = props => {
     let name = resourceName.replaceAll(/\s/g, '')
     name = name.charAt(0).toLowerCase() + name.slice(1)
     const newPermValue = newPerm[name][permission]
-    const oppPermission = permission === 'write' ? 'read' : 'write'
     if (newPerm[name].sublevelPermissions) {
       Object.keys(newPerm[name].sublevelPermissions).forEach(key => {
-        newPerm[name].sublevelPermissions[key].permissions[permission] = false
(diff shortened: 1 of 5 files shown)


hoorayimhelping (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3835)

TCL735 (approved):

PR #3881, opened 2022-02-16
chore: trying some console log statements for getMe call

ChitlangeSahas:
Part of : https://github.com/influxdata/ui/issues/3757

This function is behaving very differently in different enviornments so we need to print the outputs to see what is going on.


asalem1:
Not sure about what this achieves. What are you trying to debug? I might be able to provide some insight instead of waiting for this to bubble up to production. Also, FWIW console.warn won’t throw a lint error

TCL735 (approved):

djcarpe (approved):
Nice! Excited to see what's going on with these flags

PR #3900, opened 2022-02-16
fix: remove old token and bucket linking

drdelambre:
we used to have a mechanism that would autoselect the first bucket and auth token, then load it for people into the template for their client code examples. we removed that feature, but kept in the mechanism, leading to weird behavior as the page lived on and tokens were loaded behind the scene. this removes those mechanisms.

### DIFF
--- a/src/writeData/components/WriteDataDetailsContext.tsx
+++ b/src/writeData/components/WriteDataDetailsContext.tsx
@@ -13,20 +13,25 @@ import {
   transform,
 } from 'src/shared/components/CodeSnippet'
 // Utils
+import {getAll} from 'src/resources/selectors'
 import {getOrg} from 'src/organizations/selectors'
 
 // Types
-import {AppState, ResourceType} from 'src/types'
+import {AppState, ResourceType, Bucket} from 'src/types'
 
 const DEFAULT_TOKEN = '<INFLUX TOKEN>'
 const DEFAULT_BUCKET = '<BUCKET>'
 
 interface WriteDataDetailsContextType {
+  bucket: Bucket
+  changeBucket: (bucket: Bucket) => void
   query: string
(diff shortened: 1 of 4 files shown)


src/writeData/components/WriteDataDetailsContext.tsx
```diff
@@ -101,33 +73,27 @@ const WriteDataDetailsProvider: FC = ({children}) => {
     })
   }, [variables, org?.name, update])
   useEffect(() => {
-    if (bucket?.name === variables.bucket) {
+    if (DEFAULT_BUCKET === variables.bucket) {
```
asalem1, line 76: something about these checks seem weird. Wouldn't it make more sense to listen for changes to the `variables.bucket` rather than run a bunch of useEffects just to update a specific property of the variables? I don't know, something about the useEffects sending some flags

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3900)

src/writeData/components/WriteDataDetailsContext.tsx
drdelambre: the control of this is inverted i think to what you expect. the component this connects to doesn't consume the provider like most of our other contexts, but instead needs to call the update function of it's parent provider with the value of variables. templates just have variables, this thing adds a structure to that open ended option. these useEffects just say "no matter what you think you're passing, the answer is DEFAULT_*"

asalem1 (approved):

PR #3905, opened 2022-02-17
fix: scope user resources correctly for custom api token

### LINKED ISSUE #3849: Bug: Custom Token uses incorrectly scoped permission for users
There is a discrepancy between how we are creating All Access Tokens and Custom Tokens in regards to the `users` permission.

All Access Token:

image

Custom Token with all boxes checked:

image

Notice how the `user` permission for the All Access token contains the user ID (aka `id`) but the Custom Token contains the orgID.  This is a very important difference. From @wbaker85:

> the custom one gets you read/write:org/<org_id>/users yeah and the all-access one gives you read/write:users/<user_id>. the problem is that creating an authorization very explicitly requires one and not the other. specifically, the requesting token must have precisely write:users/user_id for it to work

Basically if I create a Custom Token, I _cannot_ use that token to create other tokens. I don't think this is an intentional difference between these token generation methods and we should resolve it to avoid confusion from users who expect to be able to use their Custom Tokens to create other tokens.

TCL735:
Closes #3849 

User permissions are a special case, it seems.

Here's the custom api token being created and the network response.

https://user-images.githubusercontent.com/10736577/154382984-f0a1ae4e-adaf-4c0f-a492-2f55281374d1.mp4


TCL735:
A force push here to ensure we have the latest updates to unrelated e2e tests.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3905)

src/authorizations/utils/permissions.ts
```diff
@@ -185,6 +185,14 @@ export const formatApiPermissions = (permissions, orgID, orgName) => {
...
           },
         })
+      } else if (key === 'users') {
+        apiPerms.push({
+          action: 'read',
+          resource: {
+            id: meID,
+            type: key,
+          },
+        })
```
genehynson, line 195: @wbaker85 this looks right to you, yeah?

genehynson (approved):

src/authorizations/utils/permissions.ts
wbaker85: The "all access" token also has the user `name` field in the permission. Will this one have that? I see from the video that it does so it must be getting added somewhere else?

wbaker85:
Just as a sanity check, would it be possible to try using the custom API token and verifying that token as the authorizing token is able to create another token, if you haven't already done that?

TCL735:
> Just as a sanity check, would it be possible to try using the custom API token and verifying that token as the authorizing token is able to create another token, if you haven't already done that?

How do i use a custom token to create another token? If this is done through the cli then i am unfamiliar with that.

wbaker85:
> > Just as a sanity check, would it be possible to try using the custom API token and verifying that token as the authorizing token is able to create another token, if you haven't already done that?
> 
> How do i use a custom token to create another token? If this is done through the cli then i am unfamiliar with that.

You should be able to do a `curl` like this, replacing values as needed:
```
curl --location --request POST '<your remocal url>/api/v2/authorizations' \
--header 'Authorization: Token <the created token>' \
--header 'Content-Type: application/json' \
--data-raw '{
    "orgID": "<remocal org id>",
    "description": "newToken",
    "permissions": [
        {
            "action": "read",
            "resource": {
                "type": "buckets",
                "orgID": "<remocal org id>"
            }
        }
    ]
}'
```

TCL735:
It works! Here's the output:
```
{
	"id": "08ee40f16f183000",
	"token": "2sWSDfOXrMMM3Nbr5ULarx0zYaPCeXCSR0ADZLthmuWAt0Eg8NlwrpkrWZKAmc4uw5BHBzpKsAeA1cDDQRt80Q==",
	"status": "active",
	"description": "newToken",
	"orgID": "9e577c9af4fe3d69",
	"org": "dev",
	"userID": "08e1982cb86af000",
	"user": "mcfly@influxdata.com",
	"permissions": [
		{
			"action": "read",
			"resource": {
				"type": "buckets",
				"orgID": "9e577c9af4fe3d69",
				"org": "dev"
			}
		}
	],
	"links": {
		"self": "/api/v2/authorizations/08ee40f16f183000",
		"user": "/api/v2/users/08e1982cb86af000"
	},
	"createdAt": "2022-02-17T18:51:27.548371542Z",
	"updatedAt": "2022-02-17T18:51:27.548371542Z"
}
```

TCL735:
...and then the newToken shows up in the UI 🎉

wbaker85 (approved):
LGTM

PR #3906, opened 2022-02-17
fix: allow All Access Tokens to use all resources

### LINKED ISSUE #3836: Bug: All Access Token created with a hardcoded list of permissions
I tried to create an All Access token using a new permission that was recently added (`subscriptions`). If I create a Custom Token, I can see this type as an option. But if I create an All Access Token, this permission is not assigned to my token.

I did a little bit of digging in the UI code and it appears that the `CustomApiTokenOverlay` component uses the `/api/v2/resources` API to get the latest resources and converts them into permissions. I believe this is the correct solution. However, the `AllAccessTokenOverlay` component uses this hardcoded list of permissions: https://github.com/influxdata/ui/blob/master/src/authorizations/utils/permissions.ts#L7 

I think the `AllAccessTokenOverlay` component should also use the resources API to ensure it's always creating tokens with the complete set of available permissions for that environment.

TCL735:
Closes #3836 

Allows All Access Tokens to get all resources and use the tokens' `allResources` list to populate all the permissions.

https://user-images.githubusercontent.com/10736577/154393911-bcf63b2a-8131-4935-b412-42523ef171b0.mp4

### DIFF
--- a/src/authorizations/components/GenerateTokenDropdown.tsx
+++ b/src/authorizations/components/GenerateTokenDropdown.tsx
@@ -30,9 +30,15 @@ const GenerateTokenDropdown: FC<ReduxProps & GenerateTokenProps> = ({
 
   const customApiOption = 'Custom API Token'
 
-  const handleAllAccess = () => {
-    showOverlay('add-master-token', null, dismissOverlay)
-    event('generate_token_dropdown.all_access_overlay.opened')
+  const handleAllAccess = async () => {
+    try {
+      await getAllResources()
+      showOverlay('add-master-token', null, dismissOverlay)
+      event('generate_token_dropdown.all_access_overlay.opened')
+    } catch (e) {
+      dispatch(notify(getResourcesTokensFailure()))
+      event('generate_token_dropdown.all_access_overlay.failed')
+    }
   }
 
   const handleCustomApi = async () => {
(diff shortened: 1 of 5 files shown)


ChitlangeSahas (approved):

src/authorizations/components/GenerateTokenDropdown.tsx
```diff
@@ -30,9 +30,15 @@ const GenerateTokenDropdown: FC<ReduxProps & GenerateTokenProps> = ({
...
-  const handleAllAccess = () => {
-    showOverlay('add-master-token', null, dismissOverlay)
-    event('generate_token_dropdown.all_access_overlay.opened')
+  const handleAllAccess = async () => {
+    try {
+      await getAllResources()
+      showOverlay('add-master-token', null, dismissOverlay)
+      event('generate_token_dropdown.all_access_overlay.opened')
+    } catch (e) {
+      dispatch(notify(getResourcesTokensFailure()))
```
blegesse-git, line 39: ohh just realized this; in the event `getAllResources` fails to fetch the data and `getResourcesTokensFailure` notification is fired, the message that pops up  will read `Failed to fetch all resources for creating custom api token`. I think the message should say something like Failed to fetch all resources for creating an _all access_ api token.

blegesse-git (commented):
other than my one suggestion, everything else looks great!

src/authorizations/components/GenerateTokenDropdown.tsx
blegesse-git: Screen Shot 2022-02-23 at 5 43 30 PM
maybe we should create another function that handles failure notification specifically for all access tokens?

TCL735: Good catch! Yes, we need the correct message for the All Access Token.

PR #3916, opened 2022-02-18
3875/export only valid alerts to task

### LINKED ISSUE #3875: Notebook(Alert): Clicking Export Alert Task fails to generates Success/Failure notification
## About the bug

**Steps to reproduce:**

1. Setup an echo listener to receive HTTP messages including Basic Authentication verification. Headers and body should be simply echoed to the console once authentication succeeds. 
3. Add test data to a bucket
4. Create an Alert Notebook
5. Use the test data to define the query
6. In the alert section choose HTTP
7. Click Authorization Basic
8. Add the URL to the listener, the username and the password
9. Click "Export Alert Task"

**Expected behavior:**

Expected to see a notification message that the transaction ended either successfully or in failure. Such a notification is provided, for example, when clicking "Test Alert".  Creating a running task is an even more important transaction and the user should be informed through notifications as to the success of the operation. 

**Actual behavior:**

No notification message is displayed. The user needs to open the tasks view to see the results.  

**Visual Proof:**

NBExportAlertTaskNoNotification01

## About your environment

**Environment info:**

Testing in K8SIDPE Remocal

latest commit in project UI
```
commit c5a2b5d154a2a8913ee2274a86c3c232a038be70 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Author: wiedld <wiedld@users.noreply.github.com>
Date:   Mon Feb 14 14:24:46 2022 -0800
```
latest commit in K8SIDPE
```
commit d65bf8f0c3fe9a84637ea06a446d65f7051282dd (HEAD -> master, origin/master, origin/alchemy/automatic, origin/HEAD)
Date:   Wed Feb 16 04:16:35 2022 +0000
```

wiedld:
Closes #3875 

Issue is that the "export task" button, in the notebook alerts panel, gives no feedback. Neither success, nor failure.

### Done:
* the create CRUD api for tasks does not validate the posted data. So we pre-run the validation with a call to the query api.
* fail validation -->  get failure notice.
* succeed validation --> continue with POST to create task --> then give success notice after post
* Also did minor refactoring. Pulled the `ExportTask.tsx` (>300 lines) from the main `/Notification/view.tsx`. Because it was getting messy.
* Also did some refactoring within the logic for ExportTask.
     * There are now common root methods for generateTask(), validateTask(), and handleTaskCreation(). 
     * Then separate helper methods (which are still the original code) for generateDeadmanTask() and generateThresholdTask().

### Visual Proof:

https://user-images.githubusercontent.com/10232835/154601557-cfddbf1e-e123-4e78-b9ca-b5e3fb244c76.mov


src/flows/pipes/Schedule/ExportTaskButton.tsx
```diff
@@ -97,58 +106,68 @@ const ExportTaskButton: FC<Props> = ({
       }
     }
 
-    postTask({data: {orgID: org.id, flux: query}})
-      .then(resp => {
-        if (resp.status !== 201) {
-          throw new Error(resp.data.message)
-        }
+    const resp = await postTask({data: {orgID: org.id, flux: query}})
+    try {
```
asalem1, line 110: nit: just in case the org is undefined for some reason
```suggestion
    try {
      const resp = await postTask({data: {orgID: org.id, flux: query}})
```

src/flows/pipes/Schedule/ExportTaskButton.tsx
```diff
@@ -97,58 +106,68 @@ const ExportTaskButton: FC<Props> = ({
...
+        throw new Error(resp.data.message)
+      }
 
-        if (isFlagEnabled('createWithFlows')) {
-          new Promise(resolve => {
-            if (flow.id) {
-              resolve(flow.id)
-              return
-            }
+      if (isFlagEnabled('createWithFlows')) {
```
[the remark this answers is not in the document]
asalem1: not sure what `onCreate` does from this context so i can't say

src/flows/pipes/Notification/view.tsx
```diff
@@ -704,11 +478,7 @@ ${ENDPOINT_DEFINITIONS[data.endpoint]?.generateTestQuery(data.endpointData)}`
         </FlexBox>
         <Panel.Footer justifyContent={JustifyContent.FlexEnd}>
           <FlexBox margin={ComponentSize.Medium}>
-            <ExportTaskButton
-              generate={generateTask}
-              text="Export Alert Task"
-              type="alert"
```
asalem1, line 710: since we're removing some of the props from this, I would check out other usages of this button (not the new one) to see if we can clean this up a bit more. If it's only being used in one other place, we could reduce the props we're sending down and isolate things a little more

asalem1 (approved):
LGTM. I like the segmentation of logic so that the button is tasked with generating the stuff. Ideally we could break that component up more so that the dropdowns are isolated too but that's way out of scope

drdelambre:
give me a second to get through this. the vast majority of this PR is effectively just changing whitespace or shuffling things from one place to another. For future PRs, please perform those steps in two operations as it's taking me way longer than it should to validate what effective product changes were made here.

src/flows/pipes/Notification/view.tsx
```diff
@@ -259,243 +255,21 @@ const Notification: FC<PipeProp> = ({Context}) => {
...
-    data.every,
-    data.offset,
-    data.endpointData,
-    data.endpoint,
-    data.thresholds,
-    data.message,
-  ])
-
-  const generateTask = useCallback(() => {
-    if (data.thresholds[0].type === deadmanType) {
```
drdelambre, line 378: note: moving this was one change

src/flows/pipes/Notification/view.tsx
```diff
@@ -259,243 +255,21 @@ const Notification: FC<PipeProp> = ({Context}) => {
...
-import "strings"
-import "regexp"
-import "influxdata/influxdb/monitor"
-import "influxdata/influxdb/schema"
-import "influxdata/influxdb/secrets"
-import "experimental"
-${ENDPOINT_DEFINITIONS[data.endpoint]?.generateImports()}
-
-check = {
-	_check_id: "${id}",
```
drdelambre, line 430: seems like this formatting was wrong too 😬

src/flows/pipes/Notification/view.tsx
```diff
@@ -259,243 +255,21 @@ const Notification: FC<PipeProp> = ({Context}) => {
...
-import "strings"
-import "regexp"
-import "influxdata/influxdb/schema"
-import "influxdata/influxdb/secrets"
-import "experimental"
-${ENDPOINT_DEFINITIONS[data.endpoint]?.generateTestImports()}
 
-${ENDPOINT_DEFINITIONS[data.endpoint]?.generateTestQuery(data.endpointData)}`
+    const queryText = `
+      import "strings"
```
drdelambre, line 262: the whitespace here is fine as we're not exposing it to the user

src/flows/pipes/Notification/ExportTask.tsx
```diff
@@ -0,0 +1,303 @@
...
+    event('Alert Panel (Notebooks) - Export Alert Task Clicked')
+
+    if (data.thresholds[0].type === deadmanType) {
+      return generateDeadmanTask()
+    } else {
+      return generateThresholdTask()
+    }
+  }, [generateDeadmanTask, generateThresholdTask, data.thresholds])
+
+  const validateTask = async (queryText: string): Promise<boolean> => {
```
drdelambre, line 272: straddling this validation between the two components is weird. and all essentially for an event name to stay collocated in a button press? feels like the utility of sharing that component has been stretched to it's breaking point here. thoughts on breaking it up here?

src/flows/pipes/Notification/view.tsx
drdelambre: i missed where you reimplemented this check between the threshold types

src/flows/pipes/Schedule/ExportTaskButton.tsx
wiedld: good catch. done.

src/flows/pipes/Notification/view.tsx
wiedld: The same logic is not there.

Only difference is that the root method generateTask() now does the hand off to the helper methods.
```
  const generateTask = useCallback(() => {
    event('Alert Panel (Notebooks) - Export Alert Task Clicked')

    if (data.thresholds[0].type === deadmanType) {
      return generateDeadmanTask()
    } else {
      return generateThresholdTask()
    }
  }, [generateDeadmanTask, generateThresholdTask, data.thresholds])
```

src/flows/pipes/Notification/ExportTask.tsx
```diff
@@ -0,0 +1,303 @@
...
+        return _acc
+      }, acc)
+
+      return acc
+    }, {})
+
+    const conditions = THRESHOLD_TYPES[deadmanType].condition(deadman)
+
+    const newQuery = `
+    import "strings"
```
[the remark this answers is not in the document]
wiedld: I can remove that extra newline.

src/flows/pipes/Notification/view.tsx
drdelambre: https://github.com/influxdata/ui/pull/3916/files#diff-8061db6ad62c27e1ef312918e57f5cfe61c64cfab70532e30477c0d8314ae72bR265  there it is

drdelambre (approved):
looks like we need to do some WASM work to get this in line with my mental model, but this'll shore up the alerts great while we wait to move this logic to notebooksd for the publish lifecycle. thanks for cleaning up the whitespace artifacts and pointing me to that switch.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3916)

src/flows/pipes/Notification/ExportTask.tsx
wiedld: we spoke offline.
This is a good compromise on what we have now...with the correct solution not being possible at this time.

src/flows/pipes/Notification/view.tsx
wiedld: good catch. made the format correct.

PR #3924, opened 2022-02-18
fix: (multi account) call getMe once more after flags are loaded

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/3757

The issue is that we're referencing the feature flags inside [getMe](https://github.com/influxdata/ui/blob/8c3ba72b41034e6faa35dffe167570094ae4fc80/src/me/actions/thunks/index.ts#L40-L44) (`multiAccount` and `avatarWidgetMultiAccountInfo`) before the flags are even loaded. The fix envolves calling the `getMe` one more time again when the `GetFlags` is finished loading the flags. This lets us use the flags inside the `getMe` function [here](https://github.com/influxdata/ui/blob/8c3ba72b41034e6faa35dffe167570094ae4fc80/src/me/actions/thunks/index.ts#L40-L44)

https://user-images.githubusercontent.com/18511823/154751210-9c4d361b-5239-4668-bb97-f10801290dfd.mov

### DIFF
--- a/src/me/actions/thunks/index.ts
+++ b/src/me/actions/thunks/index.ts
@@ -30,18 +30,31 @@ export const getMe = () => async (
   try {
     let user
 
+    // TODO: remove these logs, just for testing.
+    // eslint-disable-next-line no-console
+    console.log(
+      'Flag evaluation: ',
+      isFlagEnabled('multiAccount'),
+      isFlagEnabled('avatarWidgetMultiAccountInfo')
+    )
+
     if (
       isFlagEnabled('multiAccount') &&
       isFlagEnabled('avatarWidgetMultiAccountInfo')
     ) {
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3924)

asalem1 (changes requested):
Offlined with Sahas, the solution isn't to add another request but to rather expose the flag pre authentication and reduce the infrastructure overhead for users

Issue #3824, opened 2022-02-10
Bug: Helper error message mentions string being used even though int is used


abalone23:
The following query with `quantile()` requires a float be used rather than int, but the popup error message indicates a string is being used:
```js
import "array"

data = array.from(rows: [
  {_time: 2020-01-01T00:00:00Z, _field: "exampleField", _value: 3, foo: "bar"}
])

data
  |> quantile(column: "_value", q: 0.99, method: "estimate_tdigest", compression: 1000.0)
```
>It looks like you're trying to apply a number-based aggregate function to a string, which cannot be processed. You can fix this by selecting the Aggregate Function "Last"

It also mentions using `last()` which doesn't make sense. It should say to convert to float before using `quantile()` with the `estimate_tdigest` method. Note that the `exact_selector` method does work with an int.

<img width="1804" alt="Screen Shot 2022-02-10 at 12 09 24 PM" src="https://user-images.githubusercontent.com/330044/153488444-c3072082-e185-4173-9935-0f076008733f.png">

Luckette:
@AyeAndy - Do you have design recommendations on how to word/present this error message?

ChitlangeSahas:
@AyeAndy : Yes! So we decided that we will just show the flux error. And if we wanted a better error message, we'll reach out to the flux team and have them change the wording, in the future. We don't want to hard code error messages in the front end.

About the button margins, I agree, except the whole notification is gone so we are fine 👍

PR #3932, opened 2022-02-21
fix: remove hard-coded error message for Aggregate Type error

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/3824

The problem was that the error handling didn't have the knowledge of whether it is in the script editor mode (Advanced) or the query builder mode (builder). This caused it to throw this hard-coded error message:

Screen Shot 2022-02-21 at 3 53 41 PM

which would be helpful for the builder mode, because it has a good call to action button, but not in the advanced mode. 

https://user-images.githubusercontent.com/18511823/155035428-0f21ed10-ec8f-4af8-8a47-78eb2f9f200d.mov


hoorayimhelping (approved):

barbaranelson:
My two cents on this one is that in general I don't like hard-coded UI error messages that mask the flux error message. I would prefer that we just return the flux error message, and if we feel it isn't clear enough, then we can have the flux team change the wording.  In this example, the flux message would be: 
`runtime error @8:6-8:90: quantile: unsupported aggregate column type int`
which has the benefit of giving the line number, and I think is probably as good a message as what we are showing.

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3932)

src/timeMachine/actions/queries.ts
```diff
@@ -15,22 +15,16 @@ import {notify} from 'src/shared/actions/notifications'
...
-  updateAggregateType,
-} from 'src/shared/copy/notifications'
+import {rateLimitReached, resultTooLarge} from 'src/shared/copy/notifications'
 
 // Utils
 import fromFlux from 'src/shared/utils/fromFlux'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {buildUsedVarsOption} from 'src/variables/utils/buildVarsOption'
 import {findNodes} from 'src/shared/utils/ast'
-import {isAggregateTypeError} from 'src/utils/aggregateTypeErrors'
```
hoorayimhelping, line 29: i just searched the codebase for `isAggregateTypeError` and only found it used here. Go ahead and delete `src/utils/aggregateTypeErrors` from since it's not being used anymore.

src/timeMachine/actions/queries.ts
```diff
@@ -15,22 +15,16 @@ import {notify} from 'src/shared/actions/notifications'
...
 import {event} from 'src/cloud/utils/reporting'
 import {asSimplyKeyValueVariables, hashCode} from 'src/shared/apis/queryCache'
 import {filterUnusedVarsBasedOnQuery} from 'src/shared/utils/filterUnusedVars'
-import {getAggregateTypeErrorButton} from 'src/shared/components/notifications/NotificationButtons'
```
hoorayimhelping, line 33: `getAggregateTypeErrorButton` is only being used here, so you can delete that method from `src/shared/components/notifications/NotificationButtons`.

hoorayimhelping (approved):

PR #3939, opened 2022-02-22
chore: replace `react-copy-to-clipboard` with native JS

### LINKED ISSUE #3591: Remove `react-copy-to-clipboard` for native implementations
This one should be straightforward. We can use native JS here to swap out the dependency to replicate the functionality without the overhead of an additional dependency

appletreeisyellow:
Closes #3591 

This PR removed the dependency library `react-copy-to-clipboard` and replaced it with native JS.

### DIFF
--- a/src/shared/components/CopyToClipboard.tsx
+++ b/src/shared/components/CopyToClipboard.tsx
@@ -4,14 +4,14 @@ import copy from 'copy-to-clipboard'
 
 interface Props {
   text: string
-  onCopy: (copiedText: string, copyWasSuccessful: boolean) => void
-  children?: JSX.Element
+  children: JSX.Element
+  onCopy?: (copiedText: string, copyWasSuccessful: boolean) => void
 }
 
-const CopyToClipboard: FC<Props> = ({text, onCopy, children, ...props}) => {
-  const onClick = (event: ChangeEvent<HTMLInputElement>) => {
-    const elem = React.Children.only(children)
+const CopyToClipboard: FC<Props> = ({text, children, onCopy, ...props}) => {
+  const elem = React.Children.only(children)
 
+  const onClick = (event: ChangeEvent<HTMLInputElement>) => {
     const result = copy(text)
 
     if (onCopy) {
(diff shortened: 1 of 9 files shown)


asalem1:
We should find a polyfill for the clipboard API so that we don't lose support for IE

src/shared/components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,48 @@
...
+  try {
+    if (navigator.clipboard) {
+      // All browsers except IE
+      await navigator.clipboard.writeText(text)
+    } else {
+      // IE
+      const successful = document.execCommand('copy')
+      if (!successful) {
+        throw new Error('Copy command was unsuccessful using execCommand')
+      }
```
appletreeisyellow, line 15: @asalem1 `document.execCommand` is a polyfill for the clipboard API in IE

asalem1: Gotcha. I misunderstood what you had mentioned in yesterday's stand-up about IE support. Ya i looked at the docs and it said it was deprecated and might be discontinued at any moment. I'm wondering if there's anything else we can do rather than this or if this is our only option

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3939)

asalem1 (approved):
If there's no other support besides using the deprecated document.execCommand then let's ship it and see if someone says something

src/shared/components/CopyToClipboard.tsx
appletreeisyellow: I think this is our only option for supporting IE to copy to clipboard. 

This is the main browser API that `copy-to-clipboard` library used in their code for copying text to clipboard. `copy-to-clipboard` is a dependency of `react-copy-to-clipboard` that we want to remove in this PR. Since we use the recommended method to copy text to clipboard for other browsers except IE, i.e. `navigator.clipboard.writeText`, we should be fine here. 

Gary confirmed with the product team that we do not support InfluxData Cloud for IE, so I think we can either remove this part of the code for IE or keep it here. What do you think? @asalem1

asalem1: Sounds good. Let's do it :)

PR #3943, opened 2022-02-23
fix: update telegraf cloning to include labels

TCL735:
Part 1 (telegrafs) of idpe 13138

- Cloning telegraf configurations will now also clone the labels

https://user-images.githubusercontent.com/10736577/155422775-5ae8b9b3-add4-4d43-b38a-3939464773ea.mp4

### DIFF
--- a/src/telegrafs/actions/thunks.ts
+++ b/src/telegrafs/actions/thunks.ts
@@ -18,7 +18,6 @@ import {
 // Constants
 import {
   addTelegrafLabelFailed,
-  cloneTelegrafLabelsSuccess,
   cloneTelegrafSuccess,
   getTelegrafConfigFailed,
   removeTelegrafLabelFailed,
@@ -133,7 +132,7 @@ const cloneTelegrafLabels = (
         'An error occurred cloning the labels for this telegraf config'
       )
     }
-    dispatch(notify(cloneTelegrafLabelsSuccess()))
+    dispatch(notify(cloneTelegrafSuccess()))
   } catch {
     dispatch(notify(addTelegrafLabelFailed()))
   }
@@ -153,7 +152,6 @@ export const cloneTelegraf = (telegraf: Telegraf) => async (
     }
 
     clonedTelegraf = response.data
-    dispatch(notify(cloneTelegrafSuccess()))
   } catch (error) {
     console.error(error)
     dispatch(notify(telegrafCreateFailed()))
--- a/src/telegrafs/actions/thunks.ts
+++ b/src/telegrafs/actions/thunks.ts
@@ -108,6 +108,61 @@ export const createTelegraf = (telegraf: Telegraf) => async (
   }
 }
 
+export const cloneTelegraf = (telegraf: Telegraf) => async (
+  dispatch: Dispatch<Action>
+) => {
+  try {
+    // Step 1: create a new telegraf
+    const response = await postTelegraf({data: telegraf})
+
+    if (response.status !== 201) {
+      throw new Error(response.data.message)
+    }
+
+    const clonedTelegraf = response.data
+
+    // Step 2: clone the labels and post them with the new telegraf's id
+    const pendingLabels = telegraf.labels.map(labelID =>
+      postTelegrafsLabel({
+        telegrafID: clonedTelegraf.id,
+        data: {labelID},
+      })
+    )
+
+    const mappedLabels = await Promise.all(pendingLabels)
+
+    if (
+      mappedLabels.length &&
+      mappedLabels.some(label => label.status !== 201)
+    ) {
+      throw new Error(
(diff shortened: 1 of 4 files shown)


src/telegrafs/actions/thunks.ts
```diff
@@ -108,6 +108,61 @@ export const createTelegraf = (telegraf: Telegraf) => async (
...
+) => {
+  try {
+    // Step 1: create a new telegraf
+    const response = await postTelegraf({data: telegraf})
+
+    if (response.status !== 201) {
+      throw new Error(response.data.message)
+    }
+
+    const clonedTelegraf = response.data
```
hoorayimhelping, line 122: I think we should have multiple `try/catch` blocks here, since there are multiple operations. As it is now, if the telegraf cloning passes, but a label cloning fails, we'll dispatch an action that said the telegraf clone failed. With two `try/catch` blocks, we can have separate error messages for each situation, e.g. "we failed to clone your label 'foo', but your telegraf configuration was cloned successfully"

src/telegrafs/actions/thunks.ts
```diff
@@ -108,6 +108,61 @@ export const createTelegraf = (telegraf: Telegraf) => async (
...
+
+    const mappedLabels = await Promise.all(pendingLabels)
+
+    if (
+      mappedLabels.length &&
+      mappedLabels.some(label => label.status !== 201)
+    ) {
+      throw new Error(
+        'An error occurred cloning the labels for this telegraf config'
+      )
```
hoorayimhelping, line 140: It might be helpful to include the name of the label that failed to be cloned. Although it may be obvious if there are only a couple of labels, if there a bunch of them, and only one failed, it would make things easier to know which one failed.

src/telegrafs/actions/thunks.ts
```diff
@@ -108,6 +108,61 @@ export const createTelegraf = (telegraf: Telegraf) => async (
...
+    }
+
+    // Step 3: get the cloned telegraf
+    const updatedResponse = await apiGetTelegraf({
+      telegrafID: clonedTelegraf.id,
+      headers: {Accept: 'application/json'},
+    })
```
hoorayimhelping, line 147: this might be another place to have another `try/catch`, since this is only fetching the cloned telegraf, and if it fails, we can tell the user to reload the page, rather than telling them the whole operation failed

hoorayimhelping (commented):
I have a suggestion of splitting up the operations into multiple `try/catch` blocks since there are multiple (I count three) operations and we can tailor the message to the operation that failed. What are your thoughts on that?

TCL735:
Updated with distinct steps for better error handling and messaging.

The new video shows two notifications, one for cloning a telegraf and one for cloning the labels.

src/telegrafs/actions/thunks.ts
TCL735: Good idea, but we don't have the label's name or any other information about the label when the status is not a success. We get back only an error object.

src/telegrafs/actions/thunks.ts
```diff
@@ -108,6 +111,61 @@ export const createTelegraf = (telegraf: Telegraf) => async (
...
+
+    clonedTelegraf = response.data
+    dispatch(notify(cloneTelegrafSuccess()))
+  } catch (error) {
+    console.error(error)
+    dispatch(notify(telegrafCreateFailed()))
+  }
+
+  // Step 2: clone the labels
+  cloneTelegrafLabels(telegraf, clonedTelegraf)(dispatch)
```
hoorayimhelping, line 163: I think you can also do:

```ts
dispatch(cloneTelegrafLabels(telegraf, clonedTelegraf))
```

hoorayimhelping (approved):
Looks good to me. One comment about `dispatch(thunk())` vs `thunk()(dispatch)`.

I think we should rope Amy in on this and see if she has an opinion on showing two notifications. Not sure if the user would care about getting notified that the labels were cloned

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3943)

src/telegrafs/actions/thunks.ts
TCL735: oh nice, yeah that's better. it's consistent. i will make the change.

TCL735: oops. no, unfortunately we can't. i just tried it and TypeScript is telling i am passing in the wrong thing to `dispatch`.

PR #3947, opened 2022-02-23
fix: set share to null to reset the state

### LINKED ISSUE #3766: Notebook: wrong URL in share of a cloned notebook
Ref: https://github.com/influxdata/EAR/issues/2970

If you have already generated a share link for a notebook, and then you clone the notebook, when you go to share the cloned notebook it gives you the same share link as the original. If you then leave the clone and come back, it makes you generate a new share URL.

The share URL for a notebook should not be copied to the clone.

asalem1:
Closes #3766

This is a weird edge case where navigating to a newly cloned notebook after sharing and cancelling out of the shared state would keep a reference to the original share in the state. Clearing it out seems to do the trick, but this kind of seems indicative that there's some sort of memoization happening where data isn't being properly cleared when the state changes for the component.

### DIFF
--- a/src/flows/components/header/index.tsx
+++ b/src/flows/components/header/index.tsx
@@ -317,6 +317,7 @@ const FlowHeader: FC = () => {
   const handleClone = async () => {
     event('clone_notebook')
     const clonedId = await clone(flow.id)
+    setShare(null)
     history.push(
       `/orgs/${orgID}/${PROJECT_NAME_PLURAL.toLowerCase()}/${clonedId}`
     )


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3947)

drdelambre (approved):
i think the only way to fix this is to not have this abstraction in the first place, so i'd hold off on that rewrite. it's totally code smell here though. like... we should just use the notebook id, and depending on your login status, choose to show you read only or not, meaning ... you share the cloneid, you get the clone data

PR #3954, opened 2022-02-23
chore(ecommerce): Add commas to Operator UI Limits

abshierjoel:
In the Operator UI for organization limits, some larger numbers are difficult to read without commas to break apart the places. For example, one million is easier to read as `1,000,000` than `1000000`. 

This adds these separators, based on region, to the rate limits screen.

### DIFF
--- a/src/operator/LimitsField.tsx
+++ b/src/operator/LimitsField.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FC, useContext} from 'react'
+import React, {FC, useContext, useState} from 'react'
 import {Input, InputType} from '@influxdata/clockface'
 import {get} from 'lodash'
 import {set} from 'lodash/fp'
@@ -18,13 +18,17 @@ interface Props {
 }
 
 const LimitsField: FC<Props> = ({type, name, limits, onChangeLimits}) => {
+  const [hasFocus, setHasFocus] = useState(false)
   const {hasWritePermissions} = useContext(OperatorContext)
   const value = get(limits, name, '')
+  const csv = isNaN(value)
+    ? value
+    : Intl.NumberFormat(navigator.language).format(value)
 
   if (!hasWritePermissions) {
     return (
       <p data-testid={`limits-${name}--p`} className="operator-limits-label">
-        {value}
+        {csv}
       </p>
     )
   }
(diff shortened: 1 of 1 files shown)


src/operator/LimitsField.tsx
```diff
@@ -18,13 +18,17 @@ interface Props {
 }
 
 const LimitsField: FC<Props> = ({type, name, limits, onChangeLimits}) => {
+  const [hasFocus, setHasFocus] = useState(false)
   const {hasWritePermissions} = useContext(OperatorContext)
   const value = get(limits, name, '')
+  const csv = isNaN(value)
```
hoorayimhelping, line 24: Nit: you might consider changing this from `csv` - I thought it was an actual .csv file. And technically, it'll only be comma separated in some regions.

One thing that stands out is that we're doing two different comparisons. This code checks against `isNaN` while the code below checks the type. Is it possible to make these type checks consistent?

src/operator/LimitsField.tsx
```diff
@@ -36,12 +40,35 @@ const LimitsField: FC<Props> = ({type, name, limits, onChangeLimits}) => {
     onChangeLimits(newLimits)
   }
 
+  if (type == InputType.Number) {
```
hoorayimhelping, line 43:
```suggestion
  if (type === InputType.Number) {
```

src/operator/LimitsField.tsx
```diff
@@ -36,12 +40,35 @@ const LimitsField: FC<Props> = ({type, name, limits, onChangeLimits}) => {
...
+    return hasFocus ? (
+      <Input
+        type={type}
+        name={name}
+        value={value}
```
hoorayimhelping, line 48: shouldn't this also be `csv` (or whatever you change the name to, if you change it) since the code is already checking the type above?

src/operator/LimitsField.tsx
```diff
@@ -36,12 +40,35 @@ const LimitsField: FC<Props> = ({type, name, limits, onChangeLimits}) => {
...
+      <Input
+        type={InputType.Text}
+        name={name}
+        value={csv}
+        onChange={onChange}
+        onFocus={() => setHasFocus(true)}
+        testID={`limits-${name}--input`}
+      />
+    )
+  }
```
hoorayimhelping, line 63: I'm a little bit confused about the intent here. If I'm reading it correctly, it looks like the top level `if` checks if the type is a number. If it is, it checks if the component has focus. If it doesn't, it drops down into the bottom case, and sets the type to be `text`. Is this intended?

hoorayimhelping (commented):
Sorry for lambasting you with comments, but I was a little unclear about the intent of some of the parallel construction of the code. I'd just like some clarification on this :)

src/operator/LimitsField.tsx
abshierjoel: In this case it needs to be the value, since a number input can't have commas in it. I'm swapping the input to a text input `onBlur` so they can see the value with commas (if large enough).

src/operator/LimitsField.tsx
abshierjoel: All set

src/operator/LimitsField.tsx
abshierjoel: Good call on both. I changed this place to check the `InputType`.

hoorayimhelping (approved):
Thanks for the explanations!

abshierjoel:
@hoorayimhelping 
> Thanks for the explanations!

For sure! Thanks for the feedback!

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3954)

Issue #3941, opened 2022-02-22
Amplitude: Answer Initial Questions for Notebooks and UI


jzahnd:
We have a list of initial questions that we would like to answer via Amplitude. These questions will be answered by one or more tools with in Amplitude and, in some cases, updates to our eventing where needed. 

We should break out each question to determine if any new events or tooling need to be added and break out the report / deliverable in Amplitude.

#### Follow up questions
- How do we determine the user is "done" interacting in query builder to measure the flip to script editor? 
#### Measurement Assumptions 
- We will only capture users who interact with the query builder AND interacted with script editor.

---

#### Measurement Assumptions 
- Currently, all users land on the homepage when they login. We would want to filter that step out (at least for now) and report on behavior after that page load.

---

#### Follow up questions
- What does it mean to have "no results"? Is it an error, timeout or empty dataset?
- Can we expand on the overall goal of this measurement?

---

PR #3961, opened 2022-02-24
fix: auditing some events

drdelambre:
as part of this: https://github.com/influxdata/ui/issues/3941 and a greater question about user interaction with the query builder, i found out that we dont have a lot of data around the query builder interaction. While i was auditing in there, i found that we also dont have great events around the data sources page (the current version of the event prevents any kind of grouping), which triggered some rework of the module layout to allow for the different types of interactions to be segmented properly. 
also moved metric selector back behind a flag because product needed to review it

### DIFF
--- a/src/writeData/components/ClientLibrarySection.tsx
+++ b/src/writeData/components/ClientLibrarySection.tsx
@@ -1,7 +1,13 @@
 // Libraries
 import React, {useContext} from 'react'
-import {CLIENT_LIBS} from 'src/shared/constants/routes'
+import {useSelector} from 'react-redux'
+import {useHistory} from 'react-router-dom'
+import {ORGS, CLIENT_LIBS} from 'src/shared/constants/routes'
 import {searchClients} from 'src/writeData'
+
+// Utils
+import {getOrg} from 'src/organizations/selectors'
+import {event} from 'src/cloud/utils/reporting'
 
 // Contexts
 import {WriteDataSearchContext} from 'src/writeData/containers/WriteDataPage'
@@ -19,6 +25,9 @@ import WriteDataItem from 'src/writeData/components/WriteDataItem'
 const ClientLibrarySection = () => {
   const {searchTerm} = useContext(WriteDataSearchContext)
   const items = searchClients(searchTerm)
+
+  const history = useHistory()
+  const org = useSelector(getOrg)
 
   if (!items.length) {
     return null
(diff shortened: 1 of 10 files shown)


src/writeData/components/WriteDataItem.tsx
```diff
@@ -60,34 +41,15 @@ const WriteDataItem: FC<Props> = ({
...
     <SquareGrid.Card key={id}>
       <SelectableCard
         id={id}
         formName="load-data-cards"
         label={name}
-        testID={`load-data-item ${id}`}
-        selected={false}
-        onClick={handleClick}
+        selected={selected}
+        onClick={onClick}
```
TCL735, line 51: Thanks for fixing this so that there's only one way to use it.

`onClick` is no longer optional. It should be updated in the interface to be required or we need a fallback here.

TCL735 (changes requested):

TCL735 (approved):

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3961)

src/writeData/components/ClientLibrarySection.tsx
```diff
@@ -44,15 +53,24 @@ const ClientLibrarySection = () => {
...
-        {items.map(item => (
-          <WriteDataItem
-            key={item.id}
-            id={item.id}
-            name={item.name}
-            image={item.logo}
-            url={`${CLIENT_LIBS}/${item.id}`}
-          />
-        ))}
+        {items.map(item => {
```
Palakp41, line 56: Should we pull this out into a function or component since it's been written 3 times in 3 components?

Palakp41 (approved):
Just one comment, everything else looks good!

src/writeData/components/ClientLibrarySection.tsx
drdelambre: there's some weirdness where its sometimes items.logo and sometimes items.image, or how the urls are all different. that strange distinction coupled with how simple it is led me to believe that these snippets dont necessarily do the same thing, but very similar things. if it was more complicated, enough to validate normalizing the input and abstracting the link info out of it, i think they would totally warrant some reusability patterns, but for this case it seems like the line count to support the reusability aspect is pretty close to just copying

PR #3976, opened 2022-02-28
chore: remove event we arent looking at

drdelambre:
we are hitting some resource limits at amplitude, and there's some events we're not currently optimizing, so lets stop reporting them


asalem1 (approved):
Might wanna see if TimeSeries and the view thunk have eventing to include in this

drdelambre:
someone did a good job centralizing. looks like we're good

PR #3985, opened 2022-03-01
feat: always on insert

drdelambre:
people aren't adding cells into their notebooks. this has a bunch of problems like having people lean on the limited templating system and limiting a lot of the potential of notebooks to morph with people's needs and follow them along the journey of their data discovery. It turns out that 84% of users will never click on that little purple plus button. The edit queries and run them, but dont extend them. Of those who DO click on that button, 70% of the users find something useful in there.
Screen Shot 2022-03-01 at 11 30 11 AM

this PR exposes the panel insertion interface right on the top level, which doesn't just make insertion a lot easier / quicker, but will hopefully get those 80% of users exposed to whatever those limited users found was useful in there. i'd expect the total number of users who add a panel to raise above 11.6% and even see the little plus button see more love once people know you can actually add panels in a notebook somehow.

looks like this
Kapture 2022-03-01 at 11 25 53

### DIFF
--- a/src/flows/components/PipeList.tsx
+++ b/src/flows/components/PipeList.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FC, useContext} from 'react'
+import React, {FC, Fragment, useContext} from 'react'
 import {useSelector} from 'react-redux'
 import ReactGridLayout, {WidthProvider, Layout} from 'react-grid-layout'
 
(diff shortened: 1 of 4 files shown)


asalem1 (approved):

drdelambre:
change seems to be working as intended. we're seeing greater than 2x engagement on panel insertion week over week since making this change. here's a graph showing how many unique users have added a new panel to their notebook.
Screenshot 2022-03-07 9 42 32 AM

the 7 and 2 on the graph represent interactions on a sunday, when few people are on the system.

where as previously only ~9% of our users had even discovered that they could add a panel, now all can see it and nearly 20% are engaging with the interaction.

Issue #3590, opened 2022-01-13
Remove `memoizeone` in favor of react memoization


asalem1:
This one should be pretty straight-forward, albeit tedious. `memoize-one` is a dependency we have sprinkled throughout the app for memoization, but we can do that kind of stuff with React. We should lean into react for this and not have the overhead of another dependency

appletreeisyellow:
Since there are many components that are using `memoize-one` library, I'm going to push changes iteratively. The following is a list of files/components need to be refactored:

- [ ] dashboards
    - src/dashboards/components/dashboard_index/DashboardsIndexContentsPaginated.tsx
    - src/dashboards/components/dashboard_index/DashboardCards.tsx
- [ ] buckets
    -  src/buckets/pagination/BucketList.tsx
    -  src/buckets/components/BucketList.tsx
- [ ] tasks
    -  src/tasks/components/TasksList.tsx
    -  src/tasks/components/TaskRunsList.tsx
- [ ] telegrafs collector list
    -  src/telegrafs/components/CollectorList.tsx
- [x] token list #3990
    -  src/authorizations/components/TokenList.tsx
- [ ] label list
    - src/labels/components/LabelList.tsx
- [ ] member list
    -  src/members/components/MemberList.tsx
- [ ] scraper list
    -  src/scrapers/components/ScraperList.tsx

jzahnd:
We are blocking this so that we can divide it into multiple issues and document some concerns. We will discuss as a team on next steps.

appletreeisyellow:
Waiting to hear from the dumplings team since they were in the pagination code recently. We'll continue to refactor the `memoize-one` once we hear an update from them.

PR #3990, opened 2022-03-02
chore: replace `memoize-one` with React memoization in `TokenList`

appletreeisyellow:
Part of #3590 

This PR removes the third-party library `memoize-one` in `TokenList` by converting the class component into a functional component with the use of React memoization.


asalem1:
Let's close this for ... reasons

PR #3995, opened 2022-03-02
fix: region issues

drdelambre:
when importing from tools this week, i found that the cross region panel wasn't populating default values correctly for the scope. so i fixed that. I also moved those context panels into their own thing until we decide what to do with them


asalem1 (approved):
Should we need to make any changes to the API for notebook sharing?

drdelambre:
totally should! expect to get pinged later today

Issue #3880, opened 2022-02-16
Notebook(Alerts): HTTP Basic Auth Test Alert with bad credentials ends with notification that looks like success


karel-rehor:
## About the bug

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. Setup an echo listener to receive HTTP messages including Basic Authentication verification. Headers and body should be simply echoed to the console once authentication succeeds. Authentication failure should follow the RFC and return HTTP Status 401.
2. Go to Influxdbv2 UI (note that during this test the feature flag 'notebooksNewEndpoints' was set)
3. Add test data to a bucket
4. Create an Alert Notebook
5. Use the test data to define the query
6. In the alert section choose HTTP
7. Click Authorization Basic
8. Add the URL to the listener
9. Add a bad name and/or a bad password
10. Click "Test Alert"

**Expected behavior:**

Expected to see a notification that the test failed. The notification should not leave the impression that the alert test succeeded when the server returned an HTTP Code of 400, 401, 500 etc.  

**Actual behavior:**

The notification is colored green, as if the test succeeded.  The text "A test alert has been sent to HTTP" does not indicate the final state of the entire test process.  An inattentive user might conclude that the configuration is correct and the test is a success. 

Note. Not all servers will be configured to provide clear information as to why or how the message was received or rejected, however they should return 401 when authentication fails and the user should be informed of this. 

**Visual Proof:**

![NBBasicAuthAlertBadCreds01](https://user-images.githubusercontent.com/25881301/154309022-97d904d4-09b5-4800-99dd-7723d21806bc.gif)

## About your environment

**Environment info:**

Testing in K8SIDPE Remocal

latest commit in project UI
```
commit c5a2b5d154a2a8913ee2274a86c3c232a038be70 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Author: wiedld <wiedld@users.noreply.github.com>
Date:   Mon Feb 14 14:24:46 2022 -0800
```
latest commit in K8SIDPE
```
commit d65bf8f0c3fe9a84637ea06a446d65f7051282dd (HEAD -> master, origin/master, origin/alchemy/automatic, origin/HEAD)
Author: Hercules Mango Churchill <alchemy@influxdata.com>
Date:   Wed Feb 16 04:16:35 2022 +0000
```

karel-rehor:
Related to this would also be a test of the endpoint that fails to receive any response within a reasonable timeout.

[...]

PR #3996, opened 2022-03-03
fix(3880): surface errors from query endpoint response, to the testNotification panel

wiedld:
Part of #3880 

### Issue:
* errors are surfaced when using bad creds for HTTP post notification
      * cannot fix this specific one. the backend returns a 2xx, and nothing significant in the response body.
* general case: when the backend returns an error, we are not surfacing that specific error.
      * note: backend has updated itself to throw when there is a badURL used.
      * we need to surface that exact error returned.   

### Done:
* error from the API call is already processed here: https://github.com/influxdata/ui/blob/master/src/shared/contexts/query.tsx#L509-L522
* then is caught, and only the message (which we want) is passed up: https://github.com/influxdata/ui/blob/master/src/shared/contexts/query.tsx#L562-L569
* therefore, we should only have to pass along this message to a UI view element

.

### Visual evidence:

https://user-images.githubusercontent.com/10232835/156675016-fd3d7ed5-03ef-45cb-899a-824f9889bab5.mov

### DIFF
--- a/src/flows/pipes/Notification/view.tsx
+++ b/src/flows/pipes/Notification/view.tsx
@@ -55,6 +55,7 @@ import {notify} from 'src/shared/actions/notifications'
 import {
   testNotificationSuccess,
   testNotificationFailure,
+  invalidFluxQuery,
 } from 'src/shared/copy/notifications'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {getSecrets} from 'src/secrets/actions/thunks'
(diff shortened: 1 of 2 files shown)


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=3996)

drdelambre (approved):

PR #4004, opened 2022-03-03
fix: flaky tests fix: flows, operator

subirjolly:
**Flaky Tests**
flows.test.ts
operator.test.ts

**Broken Tests Videos**
https://user-images.githubusercontent.com/1637395/156644486-0b2621c7-cfcb-4606-b775-d7963a202a08.mp4
https://user-images.githubusercontent.com/1637395/156644502-6ba7dcd1-3a2b-4ebf-9d92-c8907185f536.mp4

### DIFF
--- a/cypress/e2e/cloud/operator.test.ts
+++ b/cypress/e2e/cloud/operator.test.ts
@@ -49,10 +49,13 @@ describe('Operator Page', () => {
       .invoke('attr', 'placeholder')
       .should('contain', 'Filter accounts by email')
 
-    cy.getByTestID('operator-resource--searchbar').type('asalem', {
-      force: true,
-      delay: 50,
-    })
+    cy.focused()
+    cy.getByTestID('operator-resource--searchbar')
+      .scrollIntoView()
+      .type('asalem', {
+        force: true,
+        delay: 50,
+      })
 
     cy.getByTestID('table-body').within(() => {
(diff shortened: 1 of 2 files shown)


cypress/e2e/cloud/operator.test.ts
```diff
@@ -49,10 +49,13 @@ describe('Operator Page', () => {
       .invoke('attr', 'placeholder')
       .should('contain', 'Filter accounts by email')
 
-    cy.getByTestID('operator-resource--searchbar').type('asalem', {
-      force: true,
-      delay: 50,
-    })
+    cy.focused()
```
hoorayimhelping, line 52: i don't think this does anything by itself if you don't use the yielded element. i think `cy.focused` by itself is like doing `array.map(operation)` and not storing the return value in an array

https://docs.cypress.io/api/commands/focused#Usage

drdelambre (approved):
lets give it a shot!

cypress/e2e/cloud/operator.test.ts
```diff
@@ -49,10 +49,13 @@ describe('Operator Page', () => {
...
+    cy.getByTestID('operator-resource--searchbar')
+      .scrollIntoView()
```
hoorayimhelping, line 54: will this fix it? in the failed test, what was typed was `alm`. it's not typing certain characters.

PR #4013, opened 2022-03-04
fix: make cloud/operator test more reliable

TCL735:
Makes cloud/operator test more reliable

- use `cy.intercept` because every character in the search bar triggers an api call
- wait on each call and assert at will
- a few other minor fixes


hoorayimhelping:
>use `cy.intercept` because every character in the search bar triggers an api call

wow, huge find. Also, an api call on every keystroke is a bit excessive and should probably be tweaked.

hoorayimhelping (approved):

PR #4018, opened 2022-03-04
fix: unblock users for the weekend

drdelambre:
based on tanking conversion numbers for alert generation while notebook utilization continues to climb, the front side validation of our notebooks alerts panel that was recently added seems to be preventing users from generating alerts over the past week. this pr is to unblock those customers trying to create an alert while we riff on the implementation to capture the intent of the original implementation.

### DIFF
--- a/src/flows/pipes/Notification/ExportTask.tsx
+++ b/src/flows/pipes/Notification/ExportTask.tsx
@@ -1,4 +1,4 @@
-import React, {FC, useContext, useCallback, useState} from 'react'
+import React, {FC, useContext, useCallback} from 'react'
 import {useDispatch} from 'react-redux'
 import {parse, format_from_js_file} from '@influxdata/flux-lsp-browser'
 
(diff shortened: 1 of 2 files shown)


wiedld:
Will approve, if (and only if) we make a ticket and place it on our sprint board. To handle the lack of validation, and feedback to the user, when creating invalid flux in the alert panel.

Otherwise, we keep getting the same theme of tickets (found issues) from bonito. Where we are exporting to tasks an invalid flux query. Can we agree on this compromise?

wiedld:
> Will approve, if (and only if) we make a ticket and place it on our sprint board. To handle the lack of validation, and feedback to the user, when creating invalid flux in the alert panel.
> 
> Otherwise, we keep getting the same theme of tickets (found issues) from bonito. Where we are exporting to tasks an invalid flux query. Can we agree on this compromise @drdelambre ?

made a ticket:
https://github.com/influxdata/ui/issues/4020

wiedld (approved):
Approved with hesitations.
We do need to unblock users.
But we are now re-introducing old bugs.
anyone curious can read here: https://github.com/influxdata/ui/issues/4020

PR #4019, opened 2022-03-04
feat: unpinning an item from the resource location

### LINKED ISSUE #3862: Allow the unpinning of item from its recourse location
** User Story **
As an end user, I want to be able to unpin an item from the same location where I have pinned it, so that I don't have to move to the home page in order to unpin it.

** Acceptance Criteria **

*Given* that
   I have pinned an item in my resource list (for instance, Notebooks)
*When*
   I select the gear icon
 *Then*
 I will see an option to unpin and can select it to unpin it

*Given* that
   I have unpinned an item in my resource list (for instance, Notebooks)
*When*
   I select the gear icon
 *Then*
 I will again see an option to pin it and can select it to pin it

appletreeisyellow:
Closes #3862 

For pinned items, the current way to remove a pinned item is to unpin it from the home page. This PR enables another way of unpinning an item from the resource list location. This applies to Notebooks, Dashboards, and Tasks.

Here is a demo for Notebook:

https://user-images.githubusercontent.com/14298407/156846371-883fa1db-b50a-4a6b-8d87-b5f390c96d32.mov

### DIFF
--- a/src/dashboards/components/dashboard_index/DashboardCard.tsx
+++ b/src/dashboards/components/dashboard_index/DashboardCard.tsx
@@ -67,6 +67,7 @@ interface OwnProps {
   labels: string[]
   onFilterChange: (searchTerm: string) => void
   onPinDashboard: () => void
+  onUnpinDashboard: (DashboardID: string) => void
   isPinned: boolean
 }
 
@@ -163,23 +164,10 @@ class DashboardCard extends PureComponent<Props> {
     }
   }
 
-  private onDeletePinnedItem = async () => {
-    const {id} = this.props
-
-    // delete from pinned item list
-    try {
-      await deletePinnedItemByParam(id)
-      this.props.sendNotification(pinnedItemSuccess('dashboard', 'deleted'))
-      this.props.onPinDashboard()
-    } catch (err) {
-      this.props.sendNotification(pinnedItemFailure(err.message, 'delete'))
-    }
-  }
-
   private handlePinDashboard = () => {
-    const {isPinned} = this.props
+    const {isPinned, id} = this.props
     if (isPinned) {
-      this.onDeletePinnedItem()
+      this.props.onUnpinDashboard(id)
     } else {
       this.onAddPinnedItem()
     }
--- a/src/dashboards/components/dashboard_index/DashboardCard.tsx
+++ b/src/dashboards/components/dashboard_index/DashboardCard.tsx
@@ -129,7 +129,7 @@ class DashboardCard extends PureComponent<Props> {
         updatePinnedItemByParam(id, {name})
         this.props.sendNotification(pinnedItemSuccess('dashboard', 'updated'))
       } catch (err) {
-        this.props.sendNotification(pinnedItemFailure(err.message, 'dashboard'))
+        this.props.sendNotification(pinnedItemFailure(err.message, 'update'))
       }
     }
   }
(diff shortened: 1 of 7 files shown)


src/dashboards/components/dashboard_index/DashboardCard.tsx
```diff
@@ -129,7 +130,7 @@ class DashboardCard extends PureComponent<Props> {
         updatePinnedItemByParam(id, {name})
         this.props.sendNotification(pinnedItemSuccess('dashboard', 'updated'))
       } catch (err) {
-        this.props.sendNotification(pinnedItemFailure(err.message, 'dashboard'))
+        this.props.sendNotification(pinnedItemFailure(err.message, 'update'))
```
asalem1, line 133: nice catch

src/dashboards/components/dashboard_index/DashboardCardsPaginated.tsx
```diff
@@ -85,6 +86,14 @@ class DashboardCards extends PureComponent<OwnProps & StateProps> {
...
     )
   }
+
+  public handlePinDashboard = () => {
+    getPinnedItems()
+      .then(res => {
+        this.setState({pinnedItems: res})
+      })
+      .catch(err => console.error(err))
+  }
```
asalem1, line 96: I would expected this function to be called in the `componentDidMount`

src/flows/components/FlowContextMenu.tsx
```diff
@@ -54,19 +54,30 @@ const FlowContextMenu: FC<Props> = ({id, name, isPinned}) => {
...
+          userID: me.id,
+          metadata: {
+            flowID: id,
+            name,
+          },
+          type: PinnedItemTypes.Notebook,
+        })
+        dispatch(notify(pinnedItemSuccess('notebook', 'added')))
+      } catch (err) {
+        dispatch(notify(pinnedItemFailure(err.message, 'create')))
```
asalem1, line 79: same nit as above about having this be two separate functions that are called.

src/tasks/components/TaskCard.tsx
```diff
@@ -127,20 +128,35 @@ export class TaskCard extends PureComponent<
...
+          orgID: org.id,
+          userID: me.id,
+          metadata: {
+            taskID: task.id,
+            name: task.name,
+          },
+          type: PinnedItemTypes.Task,
+        })
+        this.props.sendNotification(pinnedItemSuccess('task', 'added'))
+        this.props.onPinTask()
```
asalem1, line 156: Same point about the two functions & having `onPInTask` potentially do this

src/tasks/components/TasksList.tsx
```diff
@@ -158,6 +158,15 @@ export default class TasksList extends PureComponent<Props, State>
     )
   }
 
+  public handlePinTask = () => {
+    // update pinned item list
+    getPinnedItems()
+      .then(res => {
+        this.setState({pinnedItems: res})
+      })
+      .catch(err => console.error(err))
```
asalem1, line 167: I would expect this to be called in the componentDidMount

asalem1 (approved):
Overall really great work. One thing that I'd love to see get done prior to merging this in would be to consolidate the componentDidMount logic with the handlePinTask / Dashboard so that we don't have to keep track of the same logic in two different places. 

Bonus points would be having one function that deals with adding / deleting passed on the `isPInned` status in the parent component that we can just call in the parent

ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=4019)

src/dashboards/components/dashboard_index/DashboardCardsPaginated.tsx
appletreeisyellow: Good catch! I was thinking of the same thing since I didn't want to duplicate code. I tried to call the `handlePinDashboard` in the `componentDidMount` and I kept getting this React warning saying:
```
Warning: Can't perform a React state update on an unmounted component. 
This is a no-op, but it indicates a memory leak in your application.
```

Here are the things I tried:

**Attempt 1**

```TypeScript
public componentDidMount() {
  if (isFlagEnabled('pinnedItems') && CLOUD) {
      if (this._isMounted) {
        this.handlePinDashboard()
      }
    }
}
```

**Attempt 2**

```TypeScript
public componentDidMount() {
  if (isFlagEnabled('pinnedItems') && CLOUD) {
    this.handlePinDashboard(this._isMounted)
  }
}

public handlePinDashboard = (isMounted: boolean) => {
  getPinnedItems()
    .then(res => {
      if (isMounted) {
        this.setState({pinnedItems: res})
      }
    })
    .catch(err => console.error(err))
}
```

Both implementations give the above warning. Seems like `this._isMounted` should be checked in the callback after `getPinnedItems()` is done, like [this](https://github.com/influxdata/ui/blob/f59f4b5e7b2502dc586b50494aae6d8b3fc818a5/src/dashboards/components/dashboard_index/DashboardCardsPaginated.tsx#L43). I'm not quite sure why this is the case.

I'm open to other suggestions!

asalem1: Ya, let's do that. The reason why is because the state can't be set on an unmounted component (as the error suggest). Since we only want to update the state when the component has been mounted, that's fine for our purposes.

appletreeisyellow: So the two attempts are both valid in our case even though the React warning keeps showing up?

asalem1: Sort of, you might want to try:


```
public componentDidMount() {
  if (isFlagEnabled('pinnedItems') && CLOUD) {
    this.handlePinDashboard()
  }
}

public handlePinDashboard = (isMounted: boolean) => {
 if (this._isMounted) {
  getPinnedItems()
    .then(res => {
      this.setState({pinnedItems: res})
    })
    .catch(err => console.error(err))
}
```

asalem1 (approved):
LGTM! Great work

PR #4037, opened 2022-03-08
feat: make design changes for publish lifecycle

asalem1:
This PR _most_ of the design updates for the publish lifecycle work that's been ongoing. Some important things to note are:

- Ephemeral notebooks are here to stay, even with the publish lifecycle flag enabled
- We need to separate the flow.current from the flow.list

The latter part of this will be tackled elsewhere and should complete the redesign work. More specifically, by separating out the flow from the list, we can engage in a more practical pattern of updating a flow and getting a flow's source of truth from the API, rather than constantly updating the flow.list and using the local state to manage the rest of the list  / current flow


ui-ci-bot-influx:
Metric | Score | Level
seo | 90 | :green_circle:
pwa | null | :hankey:


[Compare your scores with master branch](https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/073c4b95f1089000?lower=now%28%29%20-%202d&vars%5Blighthouse_score_branch%5D=4037)

src/flows/components/VersionPage.tsx
```diff
@@ -42,17 +41,14 @@ const ReadOnlyFlowPage: FC = () => {
     <Page titleTag={flow.name + ' (Shared) | InfluxDB Cloud'}>
       <VersionHeader />
       <Page.Contents fullWidth={true} scrollable={false} className="flow-page">
-        <PopupProvider>
```
drdelambre, line 45: you sure these are all gone?

src/flows/components/header/SaveState.tsx
```diff
@@ -29,13 +29,14 @@ const SaveState: FC = () => {
...
 
-  if (
-    !isFlagEnabled('ephemeralNotebook') ||
-    isFlagEnabled('flowPublishLifecycle')
-  ) {
+  if (!isFlagEnabled('ephemeralNotebook')) {
     return null
   }
 
+  if (isFlagEnabled('flowPublishLifecycle')) {
```
drdelambre, line 36: move this to being above ephemeral notebook

asalem1: k

src/flows/components/header/index.tsx
```diff
@@ -72,8 +71,10 @@ const backgroundColor = '#07070E'
 
 type MenuItemType = {
   title: string
-  onClick: () => void
-  icon: IconFont
+  onClick?: () => void
```
drdelambre, line 74: what is the point of defining an interface is everything is optional

asalem1: Right?

drdelambre: rephrased another way:
```
interface ButtonProp {
  menuItems: Array<MenuItemType | MenuDividerType>
}
```

src/flows/components/VersionSidebar.tsx
```diff
@@ -0,0 +1,229 @@
...
+    }
+  }
+
+  const handleClone = async () => {
+    event('clone_notebook_version')
+    try {
+      const {flows} = await getAllAPI(orgID)
+
+      const allFlowNames = Object.values(flows).map(value => value.name)
+      const clonedName = incrementCloneName(allFlowNames, flow.name)
```
drdelambre, line 90: i know this is how we generate clone names, and it's great that you're not reusing the list context here, but it seems wasteful to fetch everything just to get some names that might not exist. like imagine you have thousands of notebooks and they all have unique names. you'd  be grabbing all of them for nothing. idk if we should push back on product on this or encompass it in an api so that it's at least removing the overhead on the client. thoughts?

asalem1: Totally. I'm not a fan but I can't think of anything besides `name (clone <timestamp>)` in order to bypass that

src/flows/components/VersionSidebar.tsx
```diff
@@ -0,0 +1,229 @@
...
+import {Flow} from 'src/types'
+
+interface Props {
+  version: VersionHistory | Flow
+}
+
+const VersionSidebarListItem: FC<Props> = ({version}) => {
+  const triggerRef: RefObject<HTMLButtonElement> = createRef()
+  const history = useHistory()
+  const {notebookID, id} = useParams<{notebookID: string; id: string}>()
```
drdelambre, line 50: can we explore notebooks without having a current notebook context? if not, we should be decoupling the view from the route as much as possible and using the notebook id from the flow context. i think most of your code reflects this already, so imma ask that you remove  this line, but i might be missing something? seems like we have two sources of "current notebook" and "current version" being used here

src/flows/components/VersionSidebar.tsx
drdelambre: that's prolly a good suggestion and cheaper than reimplementing clone logic on the backend everywhere. imma make a ticket for product

drdelambre (approved):
talked about my concerns. most of it was  predicated on a misunderstanding that we're overloading the current flow in the flowcontext, be we're not, just reusing that interface with a different provider. changes look good.

PR #4043, opened 2022-03-09
feat(ecommerce): Validate Email Addresses

abshierjoel:
This PR adds validation to the following places we collect email addresses: checkout, billing notification settings, and multi-account invitations. 

I thought I would bring in [validatorjs](https://github.com/validatorjs/validator.js). I thought it might be valuable to use it, since it's well-maintained and we wouldn't need to keep regex floating around. It is also tree-shakable, so we should only need to pull in the `isEmail` function from it. However, @hoorayimhelping did mention to me that validation is usually more one-off. So if we feel like this isn't a useful change, I could write up our own email validation and put it in `utils` or something.

### DIFF
--- a/package.json
+++ b/package.json
@@ -214,7 +214,6 @@
     "rudder-sdk-js": "^1.0.16",
     "s2-geometry": "^1.2.10",
     "use-local-storage-state": "^9.0.2",
-    "validator": "^13.7.0",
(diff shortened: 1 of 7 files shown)


asalem1 (approved):
Looks great. Thanks!

asalem1 (approved):

PR #4044, opened 2022-03-09
feat: fetch flux functions via fluxdocsd

blegesse-git:
Closes #


blegesse-git:
broke pr into smaller branches.

PR #4067, opened 2022-03-10
fix(4229/4021/3707/4080): notebook alert threshold validations

### LINKED ISSUE #3707: Notebooks(Alert): Alert messages are being received as spam
## About the bug

**Steps to reproduce:**

1. In AWS EC2 set up a micro instance with a simple HTTP listener that echos messages back to the console
2. Open Influxdbv2 (note that during this test the feature flag 'notebooksNewEndpoints') was set
3. Add test data to a bucket with data points that will trigger a critical threshold alert
4. Create a new Alert Notebook
5. Create a query based on the test data
6. In the alerts section define a threshold designed to trigger an alert notification 
7. In the endpoints section select HTTP and set the URL to the EC2 instance.  In this case plain HTTP with no tokens or authentication. 
8. Generate more test data with timestamps in the future and values to trigger the notification
9. Export the Notebook as a task 
10. Monitor the EC2 instance console for messages

**Expected behavior:**

Expected to receive one notification, just as when a similar Check-Endpoint-Rule is created with classic alerts. 

**Actual behavior:**

The listener is receiving scores of alerts.  This spam is repeated at a pulse equal to the value of the "Check Every" field in the Notebook alert definition. It keeps repeating well after the last data point has passed. 

From the message content, it appears a message is generated for every point in the data set.

**Visual Proof:**

**Console before creating/activating task**
ClassicCheckHTTPNotifiyListenerConsole01
Note one message from the traditional Check-Endp-Rule configuration in Alerts. 

**The Alert Definition**
NotebooksAlert01
Associated data set to be added below. 

**Console after notfication triggered(first time)**
NotebookGeneratedCheckNotifications01

**Note Similar Behavior with Slack Notification**
NotebookGeneratedCheckNotificationsSlack01
This was also encountered when exploring notifications with Slack a couple of days ago. 

## About your environment

**Environment info:**

Testing in K8S-IDPE remocal. 

last UI commit
```
commit a918e664f19a21d33b21fe2c19c7cac8aa01c573 (upstream/master, origin/master, origin/HEAD, master)
Author: Palakp41 <palak@influxdata.com>
Date:   Tue Jan 25 17:16:27 2022 -0800
```
last K8S-IDPE commit
```
commit 84c1d4fd7f71ca41759a5f31e3c245c59f3bd4da (HEAD -> master, origin/master, origin/HEAD)
Date:   Wed Jan 26 02:00:44 2022 +0000
```

**Data Set**
```
jsoucno,jev=haf mag=8.1
jsoucno,jev=haf mag=8.75
jsoucno,jev=haf mag=7.99
jsoucno,jev=haf mag=7.3
jsoucno,jev=haf mag=6.67
jsoucno,jev=haf mag=6.5
jsoucno,jev=haf mag=6.398
jsoucno,jev=haf mag=6.158
jsoucno,jev=haf mag=6.0631
jsoucno,jev=haf mag=6.61
jsoucno,jev=haf mag=6.72
jsoucno,jev=haf mag=7.28
jsoucno,jev=haf mag=8.25
jsoucno,jev=haf mag=8.6
jsoucno,jev=haf mag=9.34
jsoucno,jev=haf mag=8.76
jsoucno,jev=haf mag=7.88
jsoucno,jev=haf mag=8.094
jsoucno,jev=haf mag=32.25
jsoucno,jev=haf mag=6.06
```
Second to last point intended to trigger alert. 

**LIstener Used**
for convenience

```javascript
const fs = require('fs')
const express = require('express')
const app = express()
const port = 3000
const https = require('https')
const privateKey = fs.readFileSync('./ssl/express-selfsigned.key')
const certificate = fs.readFileSync('./ssl/express-selfsigned.crt')

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.post('/', (req,res) => {
        console.log('POST')

        let body = ''

        console.log(`HEADERS: ${JSON.stringify(req.headers)}`)

        req.on('data', (data) => {

                console.log(`DATA: ${data}`)
                body += data
        })

        req.on('end', () => {
                res.writeHead(200, {'Content-Type': 'text'})
                res.end(`received: \n ${body} \n`)
        })

//      res.send('POST received\n')
})

app.listen(port, '0.0.0.0', () => {
  console.log(`Example app listening at http://0.0.0.0:${port}`)
})

const httpsServer = https.createServer({key: privateKey, cert: certificate}, app)
httpsServer.listen(443, () => {
   console.log(`Example app listening for HTTPS`)
})


```

### LINKED ISSUE #4021: Notebook(Alerts): multiple variables in Alert definition generates uncompilable flux in exported task
## About the bug

**Steps to reproduce:**

1. Setup an echo listener to receive HTTP messages 
3. Add test data with at least two numeric fields to a bucket
4. Create an Alert Notebook
5. Using the test data define the query to use all numeric fields 
6. In the alert section define a trigger using the first numeric field AND the second numeric field. 
7. In the alert section choose basic HTTP with the endpoint from step 1
8. Click "Export Alert Task"
9. Locate the generated task in Tasks from the nav bar. e.g. Tasks > Notebook Task for local_XXXXX
10. Click Run. The task is scheduled. 
11. Return to the tasks list via the Nav Bar. And check Last Run Status 

**Expected behavior:**

Expected the task to run successfully.

**Actual behavior:**

Last Run Status is failed.  
The logs show the message. 

```
Unexpected error from query; Err: error @26:36-26:56: expected bool but found (r:{A with ixs:B}) => bool (function): error @26:36-26:56: expected bool but found (r:{A with ixs:B}) => bool (function)
```

Edit the task and the flux script shows a compilation error on line 26. 

```
trigger = (r) => r["dur"] > 20 and (r) => r["ixs"] > 28
```

**Visual Proof:**

NBAlert2Vars01

The generated flux script in the task editor
NBAlert2VarsGenFlux01

## About your environment

**Environment info:**

Testing against K8SIDPE Remocal 

Latest commit in project UI

```
commit 22850c6b4a45fa63c8e8cf267439f90cc8e58460 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Author: Jakub Bednář <jakub.bednar@gmail.com>
Date:   Mon Mar 7 08:19:59 2022 +0100
```

Latest commit in K8SIDPE

```
commit 7f2c8fcedb1170828a73670afccd8565685fcf74 (HEAD -> master, origin/master, origin/alchemy/automatic, origin/HEAD)
Date:   Mon Mar 7 08:17:01 2022 +0000
```

wiedld:
Closes #4021 
Part of #4229 
Closes #3707 
Part of #4080 

## Issues/Bugs
* bug 4229. We have no validation of the Alerts panel form inputs.
    * partial fixed here --> form validation for the thresholds. 
* bug 4021. The addition of multiple thresholds, resulted in invalid flux being exported to a task.
     * flux was writing `(r) => r["a"] > 0 and (r) => r["a"] = 0` . This is an incorrect lambda.
* Bug found. We were setting the default `threshold.field` to a string value `"_value"`.
     * this was occurring even without a user selecting a field.
     * Screen Shot 2022-03-09 at 7 04 58 PM
     * produced flux with `r["_value]`
     * therefore our form validation could not tell if the user had selected a field, or if they already had a field named "_value"
* bug 3707: alerts are being received as spam
     * the is because the `|> filter()` was not being applied before the monitor.notify()
     * also because we have multiple rows piped to monitor.notify()
* Bug 4080. We export invalid alert tasks, if a single measurement is not chosen before the monitor.check(). 
     * see docs requirement: https://docs.influxdata.com/flux/v0.x/stdlib/influxdata/influxdb/monitor/check/
     * the result was a obtuse error of
         * ```
            Unexpected error from query; Err: runtime error @42:8-42:74: check: failed to evaluate map function: unsupported binary expression invalid > int: runtime error @42:8-42:74: check: failed to evaluate map function: unsupported binary expression invalid > int
            ```

.

## Done:
* commits are labeled with the issue they address.
* housekeeping: make enum type for the ThresholdType
   * use enum in threshold definitions
   * use enum in switch statement for validations
* bug 4229:
   * validate the threshold, based upon the ThresholdType
   * do not explicitly set the `threshold.field` value ourselves. (should only be set by user interaction.)
* bug 4021:
   * handle the AND joins across multiple predicates/thresholds.
* bug 4080:
   * applied the measurement filter, before the monitor.check(). To have only 1 measurement.
   * Across all endpoints.
* bug 3707:
   * applied the threshold filter, a grouping (`|> keep`), and a limit on the results.
   * before the monitor.notify()
   * Across all endpoints.
   * confirmed with flux team that we should not have unintended behaviors.

.

## Visual proof:
https://user-images.githubusercontent.com/10232835/159588986-583bda89-1fbe-4932-b3a4-6105fbd9c430.mp4



Resultant query:
<img width="1205" alt="Screen Shot 2022-03-22 at 3 49 29 PM" src="https://user-images.githubusercontent.com/10232835/159589065-dbc2474e-f85d-4d7d-a59c-cf451cefe6b3.png">


Confirmed is being send received, and only once.
(note that the data refresh rate is >5 minutes. But see unique time stamps.)
<img width="350" alt="Screen Shot 2022-03-22 at 4 52 01 PM" src="https://user-images.githubusercontent.com/10232835/159595434-0d2f6e88-cd89-43f5-b472-e2d65109e0bd.png">



Confirmed error threshold still work:

https://user-images.githubusercontent.com/10232835/159591929-9cb96145-98e2-4b13-a6e7-cdaabfacdb48.mov



Confirmed deadman task still works:
<img width="1430" alt="Screen Shot 2022-03-22 at 4 50 51 PM" src="https://user-images.githubusercontent.com/10232835/159595340-49a1010c-33d5-469b-a5fb-dd53528acd2a.png">




.

## Next PR:
* add more schema usage, in the alert panel. (finish out 4080)
* tests

### DIFF
--- a/src/flows/pipes/Notification/ExportTask.tsx
+++ b/src/flows/pipes/Notification/ExportTask.tsx
@@ -72,6 +72,11 @@ const ExportTask: FC = () => {
       return acc
     }, {})
 
+    if (!data.measurement) {
+      throw new Error('please select a measurement.')
+    }
+    const measurement = `(r) => r["_measurement"] == "${data.measurement}"`
+
     const conditions = THRESHOLD_TYPES[deadmanType].condition(deadman)
     const imports = parse(`
 import "strings"
@@ -114,7 +119,10 @@ task_data = ${format_from_js_file(ast)}
 trigger = ${conditions}
 messageFn = (r) => ("${data.message}")
 
-${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(data.endpointData)}
+${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(
+  data.endpointData,
+  measurement
+)}
 |> monitor["deadman"](t: experimental["subDuration"](from: now(), d: ${
       deadman.deadmanCheckValue
     }))`
@@ -155,6 +163,7 @@ ${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(data.endpointData)}
     data.offset,
     data.endpointData,
     data.endpoint,
+    data.measurement,
     data.thresholds,
     data.message,
   ])
@@ -195,12 +204,15 @@ ${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(data.endpointData)}
       return acc
     }, {})
 
+    if (!data.measurement) {
+      throw new Error('please select a measurement.')
+    }
+    const measurement = `(r) => r["_measurement"] == "${data.measurement}"`
+
     const conditions = data.thresholds
       .map(threshold => THRESHOLD_TYPES[threshold.type].condition(threshold))
       .map((cond, i) => (i > 0 ? cond.split(lambdaPrefix)[1] : cond))
       .join(' and ')
-
-    const measurement = `(r) => r["_measurement"] == "${data.measurement}"`
 
     const imports = parse(`
 import "strings"
@@ -283,6 +295,7 @@ ${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(
     data.offset,
     data.endpointData,
     data.endpoint,
+    data.measurement,
     data.thresholds,
     data.message,
   ])
--- a/src/flows/pipes/Notification/ExportTask.tsx
+++ b/src/flows/pipes/Notification/ExportTask.tsx
@@ -16,7 +16,7 @@ import {
   deadmanType,
   lambdaPrefix,
   THRESHOLD_TYPES,
-} from 'src/flows/pipes/Visualization/threshold'
+} from 'src/flows/shared/threshold'
 import {ImportDeclaration} from 'src/types/ast'
 
 // Utils
--- a/src/flows/pipes/Notification/ExportTask.tsx
+++ b/src/flows/pipes/Notification/ExportTask.tsx
@@ -200,6 +200,8 @@ ${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(data.endpointData)}
       .map((cond, i) => (i > 0 ? cond.split(lambdaPrefix)[1] : cond))
       .join(' and ')
 
+    const measurement = `(r) => r["_measurement"] == "${data.measurement}"`
+
     const imports = parse(`
 import "strings"
 import "regexp"
@@ -240,7 +242,10 @@ task_data = ${format_from_js_file(ast)}
 trigger = ${conditions}
 messageFn = (r) => ("${data.message}")
 
-${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(data.endpointData)}`
+${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(
+  data.endpointData,
+  measurement
+)}`
 
     const newAST = parse(newQuery)
 
--- a/src/flows/pipes/Notification/ExportTask.tsx
+++ b/src/flows/pipes/Notification/ExportTask.tsx
@@ -14,6 +14,7 @@ import {remove} from 'src/shared/contexts/query'
 // Types
 import {
   deadmanType,
+  lambdaPrefix,
   THRESHOLD_TYPES,
 } from 'src/flows/pipes/Visualization/threshold'
 import {ImportDeclaration} from 'src/types/ast'
@@ -196,6 +197,7 @@ ${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(data.endpointData)}
 
     const conditions = data.thresholds
       .map(threshold => THRESHOLD_TYPES[threshold.type].condition(threshold))
+      .map((cond, i) => (i > 0 ? cond.split(lambdaPrefix)[1] : cond))
       .join(' and ')
 
     const imports = parse(`
--- a/src/flows/pipes/Notification/Threshold.tsx
+++ b/src/flows/pipes/Notification/Threshold.tsx
@@ -24,7 +24,7 @@ import {
   Threshold,
   ThresholdFormat,
   THRESHOLD_TYPES,
-} from 'src/flows/pipes/Visualization/threshold'
+} from 'src/flows/shared/threshold'
 
 interface Props {
   readOnly?: boolean
@@ -67,7 +67,8 @@ const Threshold: FC<Props> = ({readOnly}) => {
             type: deadmanType,
             deadmanCheckValue: '5s',
             deadmanStopValue: '90s',
-            field: threshold.field || 'Select a numeric field',
+            field: threshold.field || 'Select a field',
+            fieldType: '',
           },
         ]
       } else {
@@ -160,7 +161,7 @@ const Threshold: FC<Props> = ({readOnly}) => {
             !!readOnly ? ComponentStatus.Disabled : ComponentStatus.Default
           }
         >
-          {threshold?.field || 'Select a numeric field'}
+          {threshold?.field || 'Select a field'}
         </Dropdown.Button>
       )
       return <Dropdown menu={menu} button={menuButton} />
--- a/src/flows/pipes/Notification/Threshold.tsx
+++ b/src/flows/pipes/Notification/Threshold.tsx
@@ -54,7 +54,7 @@ const Threshold: FC<Props> = ({readOnly}) => {
       }
 
       threshold.type = type
-      threshold.field = threshold.field || '_value'
+      threshold.field = threshold?.field
 
       let updatedThreshold = thresholds
 
@@ -67,7 +67,7 @@ const Threshold: FC<Props> = ({readOnly}) => {
             type: deadmanType,
             deadmanCheckValue: '5s',
             deadmanStopValue: '90s',
-            field: threshold.field || '_value',
+            field: threshold.field || 'Select a numeric field',
           },
         ]
       } else {
--- a/src/flows/pipes/Notification/readOnly.tsx
+++ b/src/flows/pipes/Notification/readOnly.tsx
@@ -16,6 +16,7 @@ import {
   TechnoSpinner,
   SpinnerContainer,
 } from '@influxdata/clockface'
+import Measurement from 'src/flows/pipes/Notification/Measurement'
 import Threshold from 'src/flows/pipes/Notification/Threshold'
 
 import {PipeContext} from 'src/flows/context/pipe'
@@ -99,6 +100,7 @@ const ReadOnly: FC<PipeProp> = ({Context}) => {
   return (
     <Context>
       <div className="notification">
+        <Measurement readOnly />
         <Threshold readOnly />
         <FlexBox margin={ComponentSize.Medium} style={{padding: '24px 0'}}>
           <FlexBox.Child grow={1} shrink={1}>
--- a/src/flows/pipes/Visualization/ErrorThresholds/ErrorThresholds.tsx
+++ b/src/flows/pipes/Visualization/ErrorThresholds/ErrorThresholds.tsx
@@ -66,49 +66,47 @@ const ErrorThresholds: FC = () => {
           testID="component-spacer"
           stretchToFitWidth
         >
-          {data?.errorThresholds?.map(
-            (threshold: Threshold, index: number) => (
-              <FlexBox
-                direction={FlexDirection.Row}
-                margin={ComponentSize.Medium}
-                stretchToFitWidth
-                testID="component-spacer"
-                key={`${threshold.type}_${index}`}
-              >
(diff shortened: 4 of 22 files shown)


asalem1 (commented):
Some of this with the error threshold stuff. Let’s chat later

src/flows/pipes/Visualization/threshold.ts
```diff
@@ -4,85 +4,141 @@ export enum ThresholdFormat {
   Deadman = 'deadman',
 }
 
+export enum ThresholdType {
```
drdelambre, line 7: not agreeing with it existing in the visualization panel, but why isnt this using the same source of truth for types as is generating the thresholds? `keyof COMMON_THRESHOLD` or whatever
https://github.com/influxdata/ui/blob/master/src/flows/pipes/Visualization/threshold.ts#L21

src/flows/pipes/Visualization/threshold.ts
```diff
@@ -4,85 +4,141 @@ export enum ThresholdFormat {
...
+  deadmanStopValue: string // e.g. '5 minutes'
 }
 
 export interface ErrorThreshold extends Threshold {
   fieldType: string
 }
 
+const existenceCheck = unknown => unknown == 0 || !!unknown
+
+export function validateThreshold(t: Threshold): boolean {
```
drdelambre, line 37: if all types have a validation, why isn't validation a part of the type specification?

wiedld: Additional schema-based validation will be a part of the next PR.
Which will confirm:
* what the monitor.check() will permit in the lambda. (a.k.a. flux accepts string equivalency...but want to check for specifics with the check method)
* grabbing the implicit schema. probably similar to how MetricsSelector does it.

src/flows/pipes/Visualization/threshold.ts
wiedld: not sure what you mean.

I would need to keyof another type definition (interface definition).
https://www.typescriptlang.org/docs/handbook/2/keyof-types.html

src/flows/pipes/Notification/ExportTask.tsx
```diff
@@ -71,6 +72,11 @@ const ExportTask: FC = () => {
       return acc
     }, {})
 
+    if (!data.measurement) {
+      throw new Error('please select a measurement.')
+    }
+    const measurement = `(r) => r["_measurement"] == "${data.measurement}"`
+
```
wiedld, line 79: This is the deadman alert task, in notebooks.
I could not figure out how in the UI to trigger this type of alert --> to ensure it still works.

src/flows/pipes/Visualization/ErrorThresholds/ErrorThresholds.tsx
```diff
@@ -66,49 +66,47 @@ const ErrorThresholds: FC = () => {
...
-                  icon={IconFont.Trash_New}
-                  size={ComponentSize.Small}
-                  onClick={() => handleRemoveThreshold(index)}
-                  color={ComponentColor.Tertiary}
-                />
-              </FlexBox>
-            )
-          )}
+          {data?.errorThresholds?.map((threshold: Threshold, index: number) => (
+            <FlexBox
```
wiedld, line 70: this block here --> is all a linter required whitespace change.
(Once I updated the import statement, the linter then checked this file & made changes.)

src/flows/pipes/Notification/Threshold.tsx
```diff
@@ -67,7 +67,8 @@ const Threshold: FC<Props> = ({readOnly}) => {
             type: deadmanType,
             deadmanCheckValue: '5s',
             deadmanStopValue: '90s',
-            field: threshold.field || '_value',
+            field: threshold.field || 'Select a field',
+            fieldType: '',
```
asalem1, line 71: what's fieldType used for?

src/flows/pipes/Notification/readOnly.tsx
```diff
@@ -99,6 +100,7 @@ const ReadOnly: FC<PipeProp> = ({Context}) => {
   return (
     <Context>
       <div className="notification">
+        <Measurement readOnly />
```
asalem1, line 103: nice

src/flows/pipes/Notification/Threshold.tsx
wiedld: good question.
ErrorThreshold and Threshold are the same type signature, except for `fieldType`.
`fieldType` is currently utilized in ErrorThreshold only, not yet in Threshold.
The next PR (where the implicit data schema will be fetched), will allow `fieldType` to be added to Threshold as well.

The use of `fieldType` in ErrorThreshold is by defining it as 'number' or 'not-number'. Thresholds should probably have the same options...but undecided. Either way, will make it consistent for all Thresholds (be it Error or not).

asalem1: Gotcha.

asalem1 (approved):
LGTM.

src/flows/pipes/Visualization/threshold.ts
wiedld: unless what you are asking for, is something else?

src/flows/pipes/Visualization/threshold.ts
drdelambre: you added a lot of changes to the threshold definition object to tie it to the enum so that you could later use the enum values to validate a collective type. seems like a lot of effort to now couple these two concepts (the type definitions and the thing that validates to a key in that object) through a new intermediary type definition. how you have this defined, you also have to update the enum anytime you want to modify the object, which feels like an inversion of responsibility. it'd be quicker, easier to maintain, and involve less coupling to use the original object that serves to define the threshold types as the source of truth and utilize the typescript functionality of `keyof` to generate the type definition that this enum fills, but automatically. the definition then changes from:
```
export enum ThresholdType = { ... }
```

to:
```
export type ThresholdType = keyof typeof COMMON_THRESHOLD_TYPES
```

i was wondering if there was a use case i missed that would justify using the enum vs the keyof

src/flows/pipes/Visualization/threshold.ts
drdelambre: oh definitely something else. so we had this problem we first identified within our redux reducers where switches grow... to unmaintainable levels and in places where it's hard to predict (more _which_ switch will grow vs "you add another case"). You add twelve types, now this switch needs the twelve types added, the file gets too long so you split the file in two, and now you dont have the  co-locality of the code enforcing the updating of the switch statement, things silently start to fall into the default case  because new engineers dont know to check this other file, now it's a whole mess. fictional but representative. so where applicable, we've been trying to move these interfaces into the type definition at the point of definition. in this case it would be moving these switch cases into `validate` functions within each of the COMMON_THRESHOLD_TYPES and having this function just call that validate function based on the type of the threshold being passed in.

wiedld: If we have the validate statements in the object COMMON_THRESHOLD_TYPES, then we have to maintain each threshold type (9 types) separately. Instead of a single switch statement with only 3 branches (grouping all numerics, versus all non-numerics, versus the deadman).

As for the number of cases possible -- that is limited by the number of possible operators (e.g. >=. =<) which are far fewer than in a redux switch statement.

But if we want to make sure that type-checking confirms that all cases are handled (a.k.a. exhaustive switch), then we can do that with use of the `never` default case. I just pushed a commit showing this ability: https://github.com/influxdata/ui/pull/4067/commits/5e6f6d975d490168c3c5a91a14b02b0c8e18815d

wiedld: Your approach (to keep everything in that object) would make sense if we had dynamic operators, and are loading an object of dynamic operators that were defined at runtime. Or we were passing around dynamically defined closures.

But instead we have a set of statically defined operators, without the ability to have field schema checking prior to query runtime (hitting the storage layer). We do have the ability, however, to pull an implicit schema (as per the MetricsSelector). The next PR will be pulling this schema per measurement. At this point, we can have the validateThreshold() method handle the enum/operator set (which is statically defined) in comparison to this implicit schema.

Using the switch statement approach allows us to quickly mix/match (in the very near future) the different operator sets, with the schema type. **This will give us UI predicate validation** on user form input...which we don't have currently. And do so in a concise method.

Writing this logic within the COMMON_THRESHOLD_TYPES object would translate to 9 different points of potential drift (versus a single function for validation control flow). And our single point of validation can turn on/off abilities more quickly (e.g. deciding if `>` operator is a numeric only, or also non-numeric).

wiedld:
Breaking this into separate PRs.
To see if we can get merged in the changes which are in agreement...while leaving these last tickets as open:

https://github.com/influxdata/ui/issues/4229
https://github.com/influxdata/ui/issues/4080

PR #4068, opened 2022-03-10
chore: breakdown giant 1000+ line notifications file into subtyped module files

wiedld:
Closes #`annoyed-dev`

Because reusability of notifications breaks down when all of the notifications are in a giant 1000+ line file.


asalem1 (approved):
Sure, but how do we enforce this change after the fact. What’s to stop another developer from just doing the same thing accidentally?

wiedld:
> Sure, but how do we enforce this change after the fact. What’s to stop another developer from just doing the same thing accidentally?

Agreed. We don't have any enforcement, outside of the normal code review process. (Similar to the shared react components...we have to make sure people use those, instead of re-inventing the wheel.)

PR #4075, opened 2022-03-11
feat: less visual weight

drdelambre:
now that we're encouraging people to insert more panels in their notebooks, and they're getting introduced to the concept, lets dedicate less visual space to the edge cases of that process. 

query development in notebooks is predominately a top to bottom kind of thing. we want them to start, find something interesting, move on to finding something else interesting, and have a record of that whole exploration. if we think of the notebook as a chronological exploration and not as an artifact, inserting panels between each other becomes an edge case, predominately used for polishing an exploration into an artifact. because we want to focus on that initial exploratory interaction, even though we still need to give the user the affordances to do these actions for later, we dont need to make it a first level interaction.

starts to look a bit like this:
Screenshot 2022-03-11 8 38 58 AM

hover state makes it pop a little more  when you get close to the button:
Screenshot 2022-03-11 8 49 47 AM

### DIFF
--- a/src/flows/components/panel/InsertCellButton.tsx
+++ b/src/flows/components/panel/InsertCellButton.tsx
@@ -52,10 +52,7 @@ const InsertCellButton: FC<Props> = ({id}) => {
       dividerRef.current.classList.remove('flow-divider__popped')
   }
 
-  if (
-    isFlagEnabled('showLastInsert') &&
-    index === flow.data.allIDs.length - 1
-  ) {
+  if (index === flow.data.allIDs.length - 1) {
     return (
       <FlexBox
         direction={FlexDirection.Column}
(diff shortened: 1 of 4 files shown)


src/flows/components/panel/InsertCellButton.tsx
```diff
@@ -69,6 +66,65 @@ const InsertCellButton: FC<Props> = ({id}) => {
...
+              direction={FlexDirection.Column}
+              alignItems={AlignItems.Stretch}
+              margin={ComponentSize.Small}
+              className="insert-cell-menu"
+            >
+              <p className="insert-cell-menu--title">Insert Panel Here</p>
+              <AddButtons index={index} onInsert={onHide} />
+            </FlexBox>
+          )}
+        />
```
appletreeisyellow, line 123: seems like popover is not showing up

appletreeisyellow: nvm, after i pull in the updated code, it's working now!

drdelambre: did you try with the latest code push? i had the targetref attached to the button like it was before, so if you click on the purple button you get the popup, but when the hover state changes the text color like it does, you expect the text to be clickable too. i tried clicking on the text... no popup... so i changed the targetref to include it

appletreeisyellow (approved):
Looks good to me!

asalem1 (commented):
Does this remove the plus sign on the left?

src/flows/components/panel/InsertCellButton.tsx
drdelambre: lol while i was typing

PR #4077, opened 2022-03-11
fix: add eventing around controversal dropdown

drdelambre:
there seems to be some contention around whether or not people find the timezone dropdown useful. instead of blindly yoloing that call, lets see if our users are currently using this feature and where they engage with it most


asalem1 (approved):
YOLO

PR #4099, opened 2022-03-14
chore: remove the feature flag for error thresholds

asalem1:
This PR is follow-up on the error threshold work for notebooks. We've got eventing to track the feature usage and it's been stable for a while. Removing the feature flag and committing to the feature


drdelambre (approved):
:emoji of excitement:

Issue #4102, opened 2022-03-14
Fix styling of Rename Organization modal


juliajames12:
Fixes needed: 
- Use Clockface modal component to fix padding
- Remove "X" icon from Cancel button

<img width="535" alt="Screen Shot 2022-03-14 at 4 39 23 PM" src="https://user-images.githubusercontent.com/81173353/158257362-26ad0dac-1631-42a6-8e69-28e82cb1f1ff.png">

PR #4103, opened 2022-03-14
fix: rename org overlay styling

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4102

Changed the button text to say "save changes" 

Screen Shot 2022-03-14 at 2 43 38 PM


juliajames12:
This looks much better. And I like the change to the button name. Thanks for the quick fix!

PR #4108, opened 2022-03-15
test(multi-account): add remove account from org

hoorayimhelping:
Adds a test for removing an account from an organization

This is identical to https://github.com/influxdata/ui/pull/4072, but that branch got closed and deleted when I renamed the branch.


hoorayimhelping:
this test doesn't test the functionality we want tested.

PR #4122, opened 2022-03-16
test: fix flows cloning flakiness

TCL735:
- intercept the specific api call that updates the notebook spec
- hit enter after typing the name to trigger the api call


hoorayimhelping:
🤞

hoorayimhelping (approved):

PR #4132, opened 2022-03-18
fix: allow StaticLegend to work with Notebooks pipe context

### LINKED ISSUE #3962: Bug: Static Legend: has default 0 height in Notebooks
Reported by https://community.influxdata.com/t/static-legends-dont-actually-appear-on-my-notebooks/21943

Currently,
- Static Legend defaults to a non-zero (visible) height for the Data Explorer
- Static Legend defaults to a zero (not visible) height for Notebooks

We should offer a sane (visible) height in all places.

TCL735:
Closes #3962 

Static Legend now selects the correct context, either Notebooks or Time Machine, and does its magic in auto calculating the initial height for the size of the legend box.

https://user-images.githubusercontent.com/10736577/158919514-a808583d-f3ca-4bbb-8e93-5392e3ffbc39.mp4

### DIFF
--- a/src/visualization/components/View.test.tsx
+++ b/src/visualization/components/View.test.tsx
@@ -67,6 +67,19 @@ jest.doMock('src/visualization/contextLoader', () => {
   }
 })
 
+jest.mock('src/flows/index', () => {
+  return {
+    PROJECT_NAME: 'NoteBook',
+    DEFAULT_PROJECT_NAME: 'Untitled Notebook',
+  }
+})
+
+jest.mock('src/flows/templates/index', () => {
+  return {
+    TEMPLATES: {},
+  }
+})
+
 import {View, SUPPORTED_VISUALIZATIONS} from 'src/visualization'
 
 const setup = properties => {
--- a/src/visualization/utils/useStaticLegend.ts
+++ b/src/visualization/utils/useStaticLegend.ts
@@ -91,7 +91,6 @@ export const useStaticLegend = (properties): StaticLegendConfig => {
   if (id) {
     updateStaticLegendProperties = (updatedProperties: StaticLegendAPI) =>
       update({
-        ...data,
         properties: {
           ...notebookViewProperties,
           staticLegend: {...notebookStaticLegend, ...updatedProperties},
--- a/src/visualization/utils/useStaticLegend.ts
+++ b/src/visualization/utils/useStaticLegend.ts
@@ -1,6 +1,9 @@
 // Libraries
-import {useMemo, useCallback} from 'react'
+import {useMemo, useContext, useCallback} from 'react'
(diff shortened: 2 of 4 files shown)


hoorayimhelping (approved):

src/visualization/components/View.test.tsx
```diff
@@ -67,6 +67,19 @@ jest.doMock('src/visualization/contextLoader', () => {
...
+    DEFAULT_PROJECT_NAME: 'Untitled Notebook',
+  }
+})
+
+jest.mock('src/flows/templates/index', () => {
+  return {
+    TEMPLATES: {},
+  }
+})
+
```
TCL735, line 82: The mocks here are because of the changes to `useStaticLegend` which pulls in the `PipeContext` from Notebooks. This cascades into other dependencies being pulled in, eventually leading to a `require.context` which Jest cannot handle.

By mocking out the two dependencies that call `require.context`, this test file runs correctly again, just like before.

TCL735:
Add mocks and e2e tests @hoorayimhelping

ChitlangeSahas (approved):

src/visualization/utils/useStaticLegend.ts
```diff
@@ -79,6 +84,23 @@ export const useStaticLegend = (properties): StaticLegendConfig => {
...
     [dispatch, properties.staticLegend]
   )
+
+  let updateStaticLegendProperties
+  const notebookViewProperties = data?.properties || {}
+  const notebookStaticLegend = notebookViewProperties.staticLegend || {}
+  if (id) {
+    updateStaticLegendProperties = (updatedProperties: StaticLegendAPI) =>
+      update({
+        ...data,
```
drdelambre, line 94: you might have issues with stale data by implementing the spread operator here if multiple update requests are made in the same render cycle, so i'd delete this line. the same'll happen for the properties as well, but we currently dont have a mechanism for mitigating nested update collisions as it'll happen at less of a chance.

asalem1 (commented):
Will this complain about different hook calls between render cycles? Like, if the id doesn't exist we aren't using a hook and react will break?

TCL735:
> Will this complain about different hook calls between render cycles? Like, if the id doesn't exist we aren't using a hook and react will break?

No, because the hook is `useCallback` which is always called. We always create `timeMachineUpdate` because we cannot do it conditionally.

`update` from `PipeContext` is not a hook.

src/visualization/utils/useStaticLegend.ts
TCL735: Thanks, I'll remove `...data`.

Yeah, the `properties` update, we will have to take a chance on. Human users are probably not fast enough to cause the collision from selecting different view properties.

PR #4133, opened 2022-03-18
feat(4083/4084): flux doc functions, and conversion of fluxDocs into a redux resource

### LINKED ISSUE #4083: Flux Dynamic Help Panels: (Notebooks) list all flux packages and functions
As a Flux user, I want Notebooks Script Editor Function box to list all Flux functions alphabetically by Flux package first, and then alphabetically within that package. (e.g. date.second, date.week, geo.filterRows).

### LINKED ISSUE #4084: Flux Dynamic Help Panels: (Notebooks) search for a function
As a Flux user, I would like the ability to select/search for any current Flux function from the Script Editor Function box.

Current behavior allows search. Verify that this will continue to work for the new documentation. Ensure that the searching behavior is performant.

- remove unnecessary sorting where appropriate
- memoize where appropriate

wiedld:
Closes #4083 
Closes #4084 

## Done:
* updated the DynamicFunctionsList to use a sorted list, with fluxDocs fetched via network.
     * also make new child component (enough changes to not be dupe) of the tooltip contents.
* convert the fluxDocs redux store --> into a resources (global resources for this application).
     * this also means that any future use of a fluxDocs `id`, can be used to fetch-by-id: `getState().resources.fluxDocs.byID[fluxDocId]` 
* have timeMachine consume the fluxDocs resource
     * this also means that it doesn't have to handle loading state etc, because the resources pattern includes that
* note: tried to breakup those^^ tasks into separate commits, to make it easier to follow. 
  
.

## Blocking feature release:
1. There is a disconnect between the hardcoded local `FLUX_FUNCTIONS`, and the network payload
     * local has a `fn.example` which we use when injecting the function into the flux script editor
        * Screen Shot 2022-03-17 at 4 01 45 PM

     * networked payload does not include an example.
        * Screen Shot 2022-03-17 at 4 01 28 PM
     
     * This means the `|>  undefined` was being injected into the flux script.
        * Screen Shot 2022-03-17 at 5 50 25 PM

1. Since we don't have a fn.example:
     * I used `${fn.name}()` for now.
        * But this incorrect, because many require that a package be called with the function name.
        * e.g. `|> aggregate.rate()` has a package `aggregate` and a name `rate`.
     * release considerations:
        * the updates are behind a feature flag for external release.
        * But, upon merge would be released to internal influxdata users. (And then they would create broken flux.)
   
1. The ui code expects that all of the objects returned in the network payload, are indeed only function objects.
      * can we confirm that we are not getting any other objects?


src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -74,12 +74,15 @@ const Query: FC<PipeProp> = ({Context}) => {
...
-      if (fn.name === 'from' || fn.name === 'union') {
-        text = `${fn.example}`
-      } else {
-        text = `  |> ${fn.example}`
+    (fn: FluxToolbarFunction | FluxFunction): void => {
+      // FIXME: now that we are using FluxDocs, not hardcoded FLUX_FUNCTIONS
+      // we are missing the example in the http request payload
+      let text = !!(fn as FluxToolbarFunction).example
+        ? (fn as FluxToolbarFunction).example
+        : `${fn.name}()`
```
wiedld, line 82: As mentioned in the description, this is incorrect.
Need to decide what/how to handle.

src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -102,15 +102,18 @@ const TimeMachineFluxEditor: FC = () => {
       row = currentRange.endLineNumber + 1
     }
 
-    let text = ''
+    // FIXME: now that we are using FluxDocs, not hardcoded FLUX_FUNCTIONS
+    // we are missing the example in the http request payload
+    let text = !!(func as FluxToolbarFunction).example
+      ? (func as FluxToolbarFunction).example
+      : `${func.name}()`
+
```
wiedld, line 110: Another instance of this incorrect use of `fn.name`.
Must come up with a better solution.

src/types/shared.ts
```diff
@@ -33,6 +34,11 @@ export interface FluxToolbarFunction {
   link: string
 }
 
+// FIXME: confirm if this is really FluxFuntions?
+// or also contains other docs for additional flux language features.
+export type FluxDocs = FluxdocsRouteT
+export type FluxFunction = FluxDocs
```
wiedld, line 40: Wrote the types to make it explicit.
That we are assuming that the FluxDocs are all of flux functions.

src/flows/pipes/RawFluxEditor/view.tsx
wiedld: should really be `${fn.package}.${fn.name}`

src/timeMachine/components/TimeMachineFluxEditor.tsx
wiedld: `${func.example}.${func.name}()`

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
```diff
@@ -1,33 +1,65 @@
...
+  onSelect: (fn: FluxFunction) => void
 }
 
 const DynamicFunctionsList: FC<Props> = ({onSelect}) => {
+  const dispatch = useDispatch()
+  const fluxFunctions: FluxFunction[] = useSelector(getAllFluxFunctions)
+
+  useEffect(() => {
+    if (fluxFunctions.length === 0) {
+      dispatch(getFluxPackages())
```
wiedld, line 35: update to bve the action which fetches and dumps into whatever redux subtree

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
```diff
@@ -1,33 +1,65 @@
...
+}
 
 interface Props {
-  onSelect: (fn: FluxToolbarFunction) => void
+  onSelect: (fn: FluxFunction) => void
 }
 
 const DynamicFunctionsList: FC<Props> = ({onSelect}) => {
+  const dispatch = useDispatch()
+  const fluxFunctions: FluxFunction[] = useSelector(getAllFluxFunctions)
```
wiedld, line 31: update selector to grab from whatever redux subtree.

wiedld:
commits were cherry picked and merged in another PR.
closing this one.

Issue #4140, opened 2022-03-18
firstMile: Add a Filter for measurement per Query


Luckette:
**ACs**
- On both the simple flux query page and the aggregate query page, add a filter for a measurement

**Dev Notes**
|> filter(fn: (r) => r._measurement == "measurement1")

Consult @rickspencer3 and @pierwill - if needed

Luckette:
FYI @kristinarobinson

PR #4142, opened 2022-03-18
feat: add filter to ExecuteAggregateQuery.tsx

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4140


hoorayimhelping (commented):
I think you'll also want to add the filter on the `Execute a Flux Query` page as well

ChitlangeSahas:
> I think you'll also want to add the filter on the `Execute a Flux Query` page as well

Yes! That was done in this PR. I just wanted to make a new one to separate the two pages 👍

hoorayimhelping (approved):

PR #4156, opened 2022-03-21
chore: remove trackCancellations flag

### LINKED ISSUE #4166: remove `trackCancellations` feature flag
remove `trackCancellations` feature flag. `trackCancellations` was turned on in production on November 23rd and now it's time to remove it from the code and configcat.

subirjolly:
closes #4166 

`trackCancellations` got enabled on Nov 23rd 2021 in production. Thus we need to remove this flag.

### DIFF
--- a/src/accounts/context/userAccount.tsx
+++ b/src/accounts/context/userAccount.tsx
@@ -3,12 +3,11 @@ import React, {FC, useCallback, useEffect, useState} from 'react'
 import {useDispatch} from 'react-redux'
 
 // Types
-import {Account as UserAccount, postCancel} from 'src/client/unityRoutes'
+import {Account as UserAccount} from 'src/client/unityRoutes'
 
 // Notifications
 import {notify} from 'src/shared/actions/notifications'
 import {
-  accountCancellationError,
   accountDefaultSettingError,
   accountDefaultSettingSuccess,
   accountRenameError,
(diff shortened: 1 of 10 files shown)


asalem1 (commented):
What does handleCancelAccount do and are there any events under the postSignout in our metrics? Meaning does that event ever fire consistently?

subirjolly:
> 

Good point. `handleCancelAccount` doesn't hold any significance anymore in the context. Will move to the code where it's being invoked from..

According to the code, there don't seem to be any event reporting. I can add a wrapper. Does that sound okay?

src/accounts/context/userAccount.tsx
```diff
@@ -88,10 +88,6 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
       if (resp.status !== 204) {
         throw new Error(resp.data.message)
       }
-
-      if (!isFlagEnabled('trackCancellations')) {
-        history.push(`/logout`)
```
asalem1, line 93: What’s the different between postCancel and postSignout?

subirjolly: I think `postCancel` handles the cancellation of Subscription and `postSignout` is just responsible to sign out(oss/cloud). Also, `postCancel` is in Quartz and `postSignout` is in idpe.

asalem1 (approved):

PR #4157, opened 2022-03-21
feat: add gh workflow to emit merge events at target influx

GeorgeMac:
This is supporting some deployment and release metrics.

The purpose of this change is to have merges to master emit a point at a target influx.

This point will track:

- a rough timestamp of when each commit landed in master
- a best-effort attempt at annotating the source PR number
- the repo it came from
- the new SHA added

With this, we can measure the time it takes to deliver this change into target environments.
It will also allow us to alert when we exceed a threshold of time to deploy.

The goal is for the deployments team to detect and respond promptly to blocked deployments of UI changes.

Outstanding:

- [ ] I need to add the secrets referenced below to this project.

### DIFF
--- a/.github/workflows/deployments_track_push.yml
+++ b/.github/workflows/deployments_track_push.yml
@@ -1,0 +1,49 @@
+name: Track pushes to master in Influx
+
+on:
+  push:
+    branches:
+      - 'master'
+
+jobs:
+  track-new-commit-on-master:
+    name: Write event to Influx Cloud
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download and Install Influx Client
+        env:
+          INFLUXDB_RELEASES_PATH: https://dl.influxdata.com/influxdb/releases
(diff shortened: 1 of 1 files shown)


hoorayimhelping:
this looks reasonable to me, but I'll let Alex and/or Gene weigh in since they're more familiar with our deployment process

genehynson (commented):
@GeorgeMac I like this idea and don't see any harm in merging it in. But I would advise y'all compare the `Cloud-Master` github tag to what's deployed in order to figure out if a UI image should be pending deployment. Our CI updates the `Cloud-Master` tag in the UI repo whenever the _deploy_ job runs from the monitor-ci pipeline after a merge into the master branch. And the _deploy_ job only runs if those conditions are met as @wiedld pointed out.

Just comparing merges to master isn't accurate since 1) our CI has to finish on the merge commit before deploying and 2) if a newer commit is added, that merge to master will not be deployed

PR #4168, opened 2022-03-21
feat: remove feature flag from working alerts and add eventing for successful exports

### LINKED ISSUE #4159: Only support Notebooks Alert Endpoints for which we have production credentials
Given the issues that we have had testing Notebooks Alerting Endpoints, we have decided to only support Endpoints for which we have production credentials that we can test with (trial accounts are not a reliable method of testing). As such, we should segment (with a feature flag) or remove the endpoints that we will not be supporting. 

We can support the following endpoints:

- Mailgun
- AWS SES
- Slack
- Pagerduty
- HTTP Post

All other endpoint development / troubleshooting should cease and we should remove those endpoints or put them under a different feature flag (open to team decision).

asalem1:
Closes #4159

This PR enables a few endpoints that are proven to work in the UI by removing their feature flags. It also adds eventing around successfully exported alerts so that we can gain an understanding of which exported alerts users are gravitating towards

### DIFF
--- a/src/flows/pipes/Notification/ExportTask.tsx
+++ b/src/flows/pipes/Notification/ExportTask.tsx
@@ -291,6 +291,9 @@ ${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(data.endpointData)}`
   }, [generateDeadmanTask, generateThresholdTask, data.thresholds])
 
   const handleTaskCreation = _ => {
+    event('Alert panel exported', {
+      type: data.endpoint,
+    })
     dispatch(notify(exportAlertToTaskSuccess(data.endpoint)))
   }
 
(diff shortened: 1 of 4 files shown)


src/flows/pipes/Notification/ExportTask.tsx
```diff
@@ -291,6 +291,9 @@ ${ENDPOINT_DEFINITIONS[data.endpoint]?.generateQuery(data.endpointData)}`
   }, [generateDeadmanTask, generateThresholdTask, data.thresholds])
 
   const handleTaskCreation = _ => {
+    event('Alert panel exported', {
```
drdelambre, line 294: can you move this to the Export Modal Skipped event? we already have some things key'd off of that

drdelambre (approved):

PR #4170, opened 2022-03-21
feat: auto-populate variables for first mile python

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4160

Uses the variables to auto-populate the variables in the script.

https://user-images.githubusercontent.com/18511823/159373138-13de82d1-84cb-4881-9650-5551eee82c05.mov

### DIFF
--- a/src/homepageExperience/components/steps/WriteData.tsx
+++ b/src/homepageExperience/components/steps/WriteData.tsx
@@ -29,13 +29,13 @@ const logDocsOpened = () => {
 }
 
 type WriteDataProps = {
-  handleSelectBucket: (bucketName: string) => void
+  onSelectBucket: (bucketName: string) => void
 }
 
 export const WriteDataComponent = (props: WriteDataProps) => {
   const org = useSelector(getOrg)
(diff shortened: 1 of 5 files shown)


src/homepageExperience/components/steps/WriteData.tsx
```diff
@@ -114,10 +120,10 @@ export const WriteDataComponent = () => {
   )
 }
 
-export const WriteData = () => {
+export const WriteData = props => {
   return (
     <WriteDataDetailsContextProvider>
-      <WriteDataComponent />
+      <WriteDataComponent {...props} />
```
hoorayimhelping, line 126: we don't need to use spread props here since this is a custom component. The typical use case is when a react component wraps an html component that takes props, like a `<InfluxButton>` wrapping a `<button>`. It's fine to pass a single prop down like this, it's more explicit and self-documenting

hoorayimhelping (approved):
one nit about using spread props when they aren't needed, but everything else LGTM

PR #4176, opened 2022-03-22
chore: divide checkout tests for better debugging

subirjolly:
Related Issue: https://github.com/influxdata/idpe/issues/12913


subirjolly:
> 

That's correct.

PR #4177, opened 2022-03-22
chore: add event metrics to flux dynamic help panels

### LINKED ISSUE #4107: Flux Dynamic Help Panels: eventing
When users perform certain actions with this feature, we want metrics on what they are doing:

- search the functions list for a name (will need to define how long to wait on user typing)
- inject a function into the Script Editor

blegesse-git:
closes #4107 
Changes are under feature flag influx.toggle('fluxDynamicDocs')

- added eventing for script editor in Notebooks and Data Explorer. 
- uses a shared fluxDoc reducer (shared between notebooks and explorer) 
- currently, since example property doesn't exist from the flux object we get back from the service, injecting a function into editor returns `undefined`. Pr provides a **temporary** fix by injecting flux function name.


src/flows/pipes/RawFluxEditor/FunctionsList/perFunction/FluxDocsTooltipContent.tsx
```diff
@@ -7,11 +7,16 @@ import {DapperScrollbars} from '@influxdata/clockface'
...
 
+// Utils
+import {event} from 'src/cloud/utils/reporting'
+
 interface TooltipProps {
   item: Fluxdocs
 }
 
 const FluxDocsTooltipContent: FC<TooltipProps> = ({item: func}) => {
+  event('Flux function tooltip overlay opened')
```
blegesse-git, line 18: @TCL735 - I added the event metrics inside the tooltip component. so basically, whenever user hovers over the lists of flux functions, aka the tooltip component renders, it records the event. Let me know if there is a better way to do this.

src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -102,15 +102,16 @@ const TimeMachineFluxEditor: FC = () => {
       row = currentRange.endLineNumber + 1
     }
 
-    let text = ''
+    let text = !!(func as FluxToolbarFunction).example
+      ? (func as FluxToolbarFunction).example
```
blegesse-git, line 106: this is a temporary fix until we parse flux signature from the flux type and name properties. Prior this 'fix', injecting a function (with dynamic flux feature flag on) returns undefined. at least now it will say flux.name()

src/flows/pipes/RawFluxEditor/FunctionsList/perFunction/FluxDocsTooltipContent.tsx
TCL735: The event should be fired with some information about which function the user is looking at.

TCL735: The number of times the even should be fired is a good question. A user may hover over the list back and forth several times. Let's get feedback and guidance on how many times the event for each function should be fired.

blegesse-git: yeah, i should add the flux package or name to the event.l good call

TCL735:
What if a user does not inject any functions after searching? It looks like no events are fired in this case.

TCL735 (changes requested):
What if a user does not inject any functions after searching? It looks like no events are fired in this case.

blegesse-git:
> What if a user does not inject any functions after searching? It looks like no events are fired in this case.

that's an edge case I didn't think about.. then this takes us back to our original question; how can we determine when user has stopped typing in the search box. naturally, one way to know when search is "complete" is when the tooltip content pops up. We can pass searchterm to the tooltip component and say "if (searchTerm) { fire event }", but this will cause event to be recorded every time tooltip renders. Let me think on this longer

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
```diff
@@ -26,18 +26,26 @@ const sortFuncs = (a, b) => {
...
 
-  const handleSelectItem = useCallback((func: FluxFunction) => {
-    onSelect(func)
-    event('Inject FluxDoc function into Flux Script')
-  }, [])
+  const handleSelectItem = useCallback(
+    (func: FluxFunction) => {
+      onSelect(func)
+      event('flux.function.injected', {name: `${func.package}.${func.name}`})
+      if (searchTerm) {
```
drdelambre, line 43: You might wanna add this as meta data on the first event (with the name) and not make a distinct event

src/timeMachine/components/dynamicFluxFunctionsToolbar/ToolbarFunction.tsx
```diff
@@ -16,20 +16,33 @@ import {
...
+  testID,
+  searchTerm,
+}) => {
   const functionRef = createRef<HTMLDListElement>()
   const handleClickFunction = () => {
     onClickFunction(func)
+    if (searchTerm) {
+      event('flux.function.searched', {searchTerm: searchTerm})
+    }
+    event('flux.function.injected', {name: `${func.package}.${func.name}`})
```
drdelambre, line 45: Same here

TCL735:
> > What if a user does not inject any functions after searching? It looks like no events are fired in this case.
> 
> that's an edge case I didn't think about.. then this takes us back to our original question; how can we determine when user has stopped typing in the search box. naturally, one way to know when search is "complete" is when the tooltip content pops up. We can pass `searchterm` to the tooltip component and say `if (searchTerm) { fire event }`, but this will cause event to be recorded every time tooltip renders. Let me think on this longer

I think we were going to go with: fire an event for each function in the list on hover, but only the first time and not repeatedly if the user hovers over it again. On top of this, once the search term changes, the same function is allowed to fire an event again (and only on the first hover). In other words, each function gets one hover event per search term.

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
```diff
@@ -26,18 +26,36 @@ const sortFuncs = (a, b) => {
...
-    event('Inject FluxDoc function into Flux Script')
-  }, [])
+  useEffect(() => {
+    if (tooltipPopup && eventSearchTerm !== termRecorded) {
+      event('flux.function.searched', {searchTerm: eventSearchTerm})
+      setTermRecorded(eventSearchTerm)
+    }
+    setTooltipPopup(false)
+  }, [eventSearchTerm, tooltipPopup])
+
```
blegesse-git, line 48: I set up a state to keep track of tooltip pop up status. when tooltip component renders, tooltip pop up state gets set to true. When true, we check if search term is a new search term we haven't recorded before, if so, we can the search value to event metrics. Keeping termRecorded state is necessary because, it ensures search term will be sent to metrics once.  that is, until we have a new search term

blegesse-git: @TCL735 what are your thoughts on this?

blegesse-git: If we like this approach, I will add this to data explorer flux help panel

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
blegesse-git: I figured a separate event is needed here because user might not have searched for a function when they inject onto the editor. If `searchTem` is empty, then there wouldn't be a reason to send an empty value to events. But now I'm thinking about it, since we have a function that handles event metrics for what user searches for, it would be redundant to have it here again. In other words, if user searches for a function, then starts hovering over the filtered list to inject the function,  search term gets recorded. So, i think i'll remove it (the second event) entirely.

src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -102,15 +102,14 @@ const TimeMachineFluxEditor: FC = () => {
...
+      text = `\n  |> ${text}`
+    } else if (functionRequiresNewLine(func.name)) {
+      text = `\n${text}\n`
     } else {
-      text = `  |> ${func.example}\n`
-    }
-
-    if (functionRequiresNewLine(func.name)) {
-      text = `\n${func.example}\n`
+      text = ` |> ${text}\n`
```
TCL735, line 112: This is causing the test failure

TCL735 (dismissed):

TCL735 (approved):

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
```diff
@@ -24,19 +24,44 @@ const sortFuncs = (a, b) => {
...
+      hoveredFunctions.clear()
+      event('flux.function.searched', {searchTerm: eventSearchTerm})
+      setTermRecorded(eventSearchTerm)
+    }
+    setTooltipPopup(false)
+    if (tooltipPopup) {
+      hoveredFunctions.add(hoverdFunction)
+
+      if (recordedFunctions.size !== hoveredFunctions.size) {
+        event('hovered.flux.function', hoveredFunctions)
```
TCL735, line 56: Something like this:
```
event('hovered.flux.function', {hoveredFunction, searchTerm: eventSearchTerm})
```

but make sure it's at the moment of hover

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
```diff
@@ -24,19 +24,44 @@ const sortFuncs = (a, b) => {
...
+        recordedFunctions = new Set([...hoveredFunctions])
```
TCL735, line 57: i think `recordedFunctions.add(hoveredFunction)` is easier, right? why re-create the whole set again just to add a hovered function?

TCL735 (changes requested):

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
TCL735: also, the name of the event might need tweaking. maybe something like `flux.function.hover`

blegesse-git: it's firing right on hover, so we're good there. But adding eventSearchTerm there would mean we will be sending an empty string. If that's not an issue, i can easily add it. My thought process was; I have a code that handles firing event for search term, so we dont need it here.

src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList.tsx
blegesse-git: yes much easier than what i have! done!!

TCL735 (approved):

PR #4182, opened 2022-03-22
fix: adding navigation events

drdelambre:
we had no visibility into how many people are clicking the upgrade button or where they were when they click it, so i added that event so that we an start looking closer at those funnels. hoping we can optimize that workflow a bit. also added some events to the tree nav as we have very little insight into how many people are using that crucial piece of navigation.

sparked by the observation that the "happy path" for our users who want to sign up and aren't sold by the "trust us it's easy" text on the top of the checkout form is a loop to nowhere.

d65be39125


helenosheaa (approved):
lgtm!

asalem1 (approved):

PR #4187, opened 2022-03-23
feat: restrict panel adding based on previous panels

### LINKED ISSUE #3777: Notebook: Some panels don't work correctly when added as a first panel
If you add an alert panel as the first panel, it does not display correctly and instead shows a confusing error that "This cell will display results from the previous cell after selecting Run."  We need some way to correctly handle panels that need previous results to run correctly, either by not allowing them to be an option when adding a panel at the top of the notebook, or by giving a more specific error message that explains the problem.

asalem1:
Closes #3777 

### Problem

Any time of panel could be added anywhere, regardless of whether a panel is dependent upon previous query results or not

### Solution 

Conditionally disabled certain panel type additions so that users are forced into a flow of valid panel insertions based on the previous pipes.

add-panel

### DIFF
--- a/src/flows/components/AddButtons.tsx
+++ b/src/flows/components/AddButtons.tsx
@@ -84,9 +84,7 @@ const AddButtons: FC<Props> = ({index, onInsert}) => {
 
   const pipeIds = flow.data.allIDs
   const firstQueryIndex = pipeIds.findIndex(
-    id =>
-      flow.data.byID[id].type === 'queryBuilder' ||
-      flow.data.byID[id].type === 'rawFluxEditor'
+    id => PIPE_DEFINITIONS[flow.data.byID[id].type].family === 'inputs'
   )
 
   const cellFamilies = SUPPORTED_FAMILIES.filter(fam =>
--- a/src/flows/components/AddButtons.tsx
+++ b/src/flows/components/AddButtons.tsx
@@ -2,7 +2,7 @@
 import React, {FC, useContext} from 'react'
 
 // Components
-import {Button, ComponentColor} from '@influxdata/clockface'
+import {Button, ComponentColor, ComponentStatus} from '@influxdata/clockface'
 import CellFamily from 'src/flows/components/CellFamily'
 
 // Constants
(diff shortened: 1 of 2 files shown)


src/flows/components/AddButtons.tsx
```diff
@@ -80,15 +82,27 @@ const AddButtons: FC<Props> = ({index, onInsert}) => {
     return acc
   }, {})
 
+  const pipeIds = flow.data.allIDs
+  const firstQueryIndex = pipeIds.findIndex(
+    id =>
+      flow.data.byID[id].type === 'queryBuilder' ||
```
drdelambre, line 88: can please key off the family here instead of the type in order to allow for more types to be added?

src/flows/components/AddButtons.tsx
```diff
@@ -80,15 +82,27 @@ const AddButtons: FC<Props> = ({index, onInsert}) => {
     return acc
   }, {})
 
+  const pipeIds = flow.data.allIDs
+  const firstQueryIndex = pipeIds.findIndex(
```
drdelambre, line 86: a bit torn about this. seems like it should be run at point of consumption, but maybe this is fine

src/flows/components/AddButtons.tsx
```diff
@@ -80,15 +82,27 @@ const AddButtons: FC<Props> = ({index, onInsert}) => {
...
+      flow.data.byID[id].type === 'queryBuilder' ||
+      flow.data.byID[id].type === 'rawFluxEditor'
+  )
+
   const cellFamilies = SUPPORTED_FAMILIES.filter(fam =>
     pipeFamilies.hasOwnProperty(fam.family)
   ).map(fam => {
     const pipes = pipeFamilies[fam.family].map(def => {
+      let status = ComponentStatus.Default
+      if (index < firstQueryIndex && NEED_QUERY_PANELS.includes(def.type)) {
```
drdelambre, line 97: a note to future me and not to ari: i dont like how this is keying off of the type and not something more generic like the family, or a property of the panels themselves, but that's the specific ask of the product, so deal

src/flows/components/AddButtons.tsx
asalem1: done and done!

drdelambre (approved):

PR #4189, opened 2022-03-23
feat: subscriptions list

helenosheaa:
Closes https://github.com/influxdata/data-acquisition/issues/95

Adds filter by name and sort by updated at/ name to subscriptions list view.
Adds last modified and status info to details card.

### DIFF
--- a/src/shared/components/resource_sort_dropdown/generateSortItems.ts
+++ b/src/shared/components/resource_sort_dropdown/generateSortItems.ts
@@ -401,15 +401,15 @@ export const generateSortItems = (
           sortDirection: Sort.Descending,
         },
         {
-          label: 'Description (Ascending)',
-          sortKey: 'description',
-          sortType: SortTypes.String,
+          label: 'Created (Oldest)',
+          sortKey: 'createdAt',
+          sortType: SortTypes.Date,
           sortDirection: Sort.Ascending,
         },
         {
-          label: 'Description (Descending)',
-          sortKey: 'description',
-          sortType: SortTypes.String,
+          label: 'Created (Newest)',
+          sortKey: 'createdAt',
+          sortType: SortTypes.Date,
           sortDirection: Sort.Descending,
         },
       ]
--- a/src/shared/components/resource_sort_dropdown/generateSortItems.ts
+++ b/src/shared/components/resource_sort_dropdown/generateSortItems.ts
@@ -401,14 +401,14 @@ export const generateSortItems = (
           sortDirection: Sort.Descending,
         },
         {
-          label: 'Created (Oldest)',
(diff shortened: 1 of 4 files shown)


src/writeData/subscriptions/components/SubscriptionCard.tsx
```diff
@@ -1,5 +1,6 @@
 // Libraries
 import React, {FC, useContext} from 'react'
+import moment from 'moment'
```
genehynson, line 3: I believe we've moved away from `moment` in the UI since it has been [deprecated](https://github.com/moment/moment). It looks like it's only available in package.json as a devDep.

src/writeData/subscriptions/components/SubscriptionCard.tsx
```diff
@@ -22,6 +23,7 @@ interface Props {
 
 const SubscriptionCard: FC<Props> = ({subscription}) => {
   const {deleteSubscription} = useContext(SubscriptionListContext)
+  const timeSince = moment(subscription.updatedAt).fromNow()
```
genehynson, line 26: Instead of moment I think you can use `luxon` to calculate relative datetimes.

src/shared/components/resource_sort_dropdown/generateSortItems.ts
```diff
@@ -401,15 +401,15 @@ export const generateSortItems = (
...
+          sortType: SortTypes.Date,
           sortDirection: Sort.Ascending,
         },
         {
-          label: 'Description (Descending)',
-          sortKey: 'description',
-          sortType: SortTypes.String,
+          label: 'Updated (Newest)',
+          sortKey: 'updatedAt',
+          sortType: SortTypes.Date,
```
genehynson, line 412: this looks like a shared component. Will changing this affect other pages?

genehynson (commented):
looks good - just need to swap moment for luxon I think

src/shared/components/resource_sort_dropdown/generateSortItems.ts
helenosheaa: if you look on line 389 this is just the definition for subscriptions

src/writeData/subscriptions/components/SubscriptionCard.tsx
helenosheaa: ah ok i'll change that out. I just searched moment and saw it imported without checking further

src/writeData/subscriptions/components/SubscriptionCard.tsx
helenosheaa: good catch, done!

genehynson (approved):

PR #4193, opened 2022-03-24
fix: use Giraffe's prefix formatters correctly

TCL735:
Waiting on a new version of Giraffe


TCL735:
The unit test will pass when the new version of Giraffe is released.

Issue #4194, opened 2022-03-24
Usage page doesn't work with `graphColorMapping` flag on


ChitlangeSahas:
The usage page doesn't work when the color mapping flag is on.

Issue #4031, opened 2022-03-07
InfluxDB Cloud organization usage metrics dashboard not showing


natalya-patrikeeva:
## About the bug

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior. Refreshing the page. 

1. In the UI, go to Organization -> Usage to see the usage metrics dashboard

**Expected behavior:**
See the usage metrics dashboard with information like the number of queries, MB of data, etc.

**Actual behavior:**
Only see the error message - An InfluxDB error has occurred. Please report the issue here.

**Visual Proof:**
Chrome:
<img width="762" alt="screen-shot" src="https://user-images.githubusercontent.com/10470710/157105982-e04d2534-de47-471c-af89-b7e0ce30a89b.png">
Safari:
<img width="923" alt="Screen Shot 2022-03-07 at 1 48 25 PM" src="https://user-images.githubusercontent.com/10470710/157107118-5327cea3-7c00-4b15-8440-629c58b5a87c.png">

## About your environment

**Environment info:**

- System info:
```bash
uname -srm
Darwin 21.3.0 x86_64
```
- Running InfluxDB Cloud on us-west-2-2.aws.cloud2 
- Same bug in Google Chrome Version 98.0.4758.102 (Official Build) (x86_64) and Safari Version 15.3 (17612.4.9.1.8)

jzahnd:
This was a bug that has since been fixed. Closing.

Issue #4011, opened 2022-03-03
usage not working


robertoctf:
## About the bug

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. ...tried to access usage
2. ...
3. ...

**Expected behavior:**

**Actual behavior:**

**Visual Proof:**

## About your environment

**Environment info:**

**Config:**

**Logs:**

**Performance:**
Generate profiles with the following commands for bugs related to performance, locking, out of memory (OOM), etc.

```sh
# Commands should be run when the bug is actively.
# Note: This command will run for at least 30 seconds.
curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all?cpu=true"
curl -o vars.txt "http://localhost:8086/debug/vars"
iostat -xd 1 30 > iostat.txt
# Attach the `profiles.tar.gz`, `vars.txt`, and `iostat.txt` output files.
```

kristinarobinson:
@robertoctf - can you verify whether you are a free tier user on Cloud, a PAYG customer on cloud, or using our OSS product?

jzahnd:
This was a bug that we have fixed, closing for now.

PR #4195, opened 2022-03-24
fix: usage page make it work with `graphColorMapping` flag

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4194
Closes https://github.com/influxdata/ui/issues/4011
Closes https://github.com/influxdata/ui/issues/4031
Among multiple of similar usage page issues reported.

https://user-images.githubusercontent.com/18511823/159864353-f9e788bc-f5d8-4de6-a27c-84f86e0671bf.mov

### DIFF
--- a/src/visualization/utils/colorMappingUtils.ts
+++ b/src/visualization/utils/colorMappingUtils.ts
@@ -159,7 +159,14 @@ export const generateSeriesToColorHex = (
   const cgMap = {...columnGroupMap}
   cgMap.mappings.forEach((graphLine, colorIndex) => {
     const id = getSeriesId(graphLine, columnGroupMap.columnKeys)
-    const colors = properties.colors ?? DEFAULT_LINE_COLORS
+    let colors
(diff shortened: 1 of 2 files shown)


src/visualization/utils/colorMappingUtils.ts
```diff
@@ -159,7 +159,14 @@ export const generateSeriesToColorHex = (
   const cgMap = {...columnGroupMap}
   cgMap.mappings.forEach((graphLine, colorIndex) => {
     const id = getSeriesId(graphLine, columnGroupMap.columnKeys)
-    const colors = properties.colors ?? DEFAULT_LINE_COLORS
+    let colors
```
ChitlangeSahas, line 162: This was coming in as an empty array `[]` and we only dealt with the cases when the `properties.colors` was either `null` or `undefined`. We needed to add the handler to deal with when it's `[]`.

hoorayimhelping (approved):

PR #4197, opened 2022-03-24
chore: audit some features that are behind flags to determine their usage

asalem1:
We've got some codepaths behind feature flags that are .... outdated. This PR adds some events to the areas of our codebase that are feature flagged in order to get a better sense of whether they're being used or not

### DIFF
--- a/src/checks/components/CheckMatchingRulesCard.tsx
+++ b/src/checks/components/CheckMatchingRulesCard.tsx
@@ -31,11 +31,15 @@ import {
 import {EmptyState, ComponentSize, RemoteDataState} from '@influxdata/clockface'
 import BuilderCard from 'src/timeMachine/components/builderCard/BuilderCard'
 import {getActiveTimeMachine} from 'src/timeMachine/selectors'
+import {event} from 'src/cloud/utils/reporting'
 
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = ReduxProps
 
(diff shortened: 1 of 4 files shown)


src/checks/components/CheckMatchingRulesCard.tsx
```diff
@@ -31,11 +31,15 @@ import {
 import {EmptyState, ComponentSize, RemoteDataState} from '@influxdata/clockface'
 import BuilderCard from 'src/timeMachine/components/builderCard/BuilderCard'
 import {getActiveTimeMachine} from 'src/timeMachine/selectors'
+import {event} from 'src/cloud/utils/reporting'
 
 type ReduxProps = ConnectedProps<typeof connector>
 type Props = ReduxProps
 
 const CheckMatchingRulesCard: FC<Props> = ({orgID, tags, queryResults}) => {
+  React.useEffect(() => {
```
drdelambre, line 40: why isn;t  this just a  `useEffect`?

asalem1: Didn't realize it was being imported at the top. It can be

drdelambre:
i really dont understand what you're trying to achieve here. the eventing subsystem is used to track user engagement /  interaction with the interface. most of  these read like "if the flag is on, throw an event".. except we know if a flag is on or not and have an audit log for them. i feel like i'm missing something?

asalem1:
Gonna close this based on the last comment

PR #4203, opened 2022-03-24
fix: styling of the documentation panel so it is a better user experience

ChitlangeSahas:
Video: 

https://user-images.githubusercontent.com/18511823/159992846-cd9a0be0-dbdf-432f-9c4c-18d2756dcac0.mov

### DIFF
--- a/src/me/components/DocSearchWidget.tsx
+++ b/src/me/components/DocSearchWidget.tsx
@@ -26,13 +26,13 @@ const supportLinks = [
   },
   {
     link: 'https://github.com/influxdata/ui/issues/new',
-    title: '🐛 Report a bug',
+    title: 'Report a bug',
   },
-  {link: 'https://community.influxdata.com', title: '💭 Community Forum'},
+  {link: 'https://community.influxdata.com', title: 'Community Forum'},
   {
     link:
       'https://github.com/influxdata/influxdb/issues/new?template=feature_request.md',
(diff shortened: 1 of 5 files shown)


src/me/components/DocSearchWidget.tsx
```diff
@@ -42,7 +42,7 @@ const DocSearchWidget: FC = () => {
       <DocSearch type={DocSearchType.Widget} />
       <p className="WidgetHelperText">Press CTRL + M on any page to search</p>
       <div className="useful-links">
-        <h4>Useful Links</h4>
+        <h4>USEFUL LINKS</h4>
```
hoorayimhelping, line 45: I think we should prefer `text-transform` over all caps: https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

Here's a good article explaining why: https://medium.com/@mandy.michael/why-you-shouldnt-write-your-content-in-uppercase-instead-use-css-b03ac2c65b99

hoorayimhelping (approved):
LGTM, but one suggestion about uppercase vs `text-transform`

src/me/components/DocSearchWidget.tsx
ChitlangeSahas: ✅

hoorayimhelping:
also, great work removing needless `!important` declarations!

PR #4205, opened 2022-03-24
feat: separate the flow list from the current flow

asalem1:
This PR separates the flows list context from the current flow context. The rationale behind this move is so that we can keep the list page a dumb list, and make each flow a source of truth in and of itself. This renders a few methods from the list page irrelevant while making sure that the current functionality still exists. 

Since the list is now no longer the source of truth, we should update the API for the getAllNotebooks so restrict the data that's sent so that we don't end up with a potentially bloated list page. 

current-flow

### DIFF
--- a/src/flows/components/FlowPage.tsx
+++ b/src/flows/components/FlowPage.tsx
@@ -1,6 +1,5 @@
 // Libraries
 import React, {FC, useContext, useEffect} from 'react'
-import {useParams} from 'react-router-dom'
 import {Page} from '@influxdata/clockface'
 import {DapperScrollbars} from '@influxdata/clockface'
 
@@ -8,7 +7,6 @@ import {DapperScrollbars} from '@influxdata/clockface'
 import CurrentFlowProvider, {FlowContext} from 'src/flows/context/flow.current'
 import QueryProvider from 'src/shared/contexts/query'
 import {FlowQueryProvider, FlowQueryContext} from 'src/flows/context/flow.query'
-import {FlowListContext} from 'src/flows/context/flow.list'
 import {PopupDrawer, PopupProvider} from 'src/flows/context/popup'
 import {ResultsProvider} from 'src/flows/context/results'
 import {SidebarProvider} from 'src/flows/context/sidebar'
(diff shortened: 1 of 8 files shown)


src/flows/components/FlowPage.tsx
```diff
@@ -29,23 +27,15 @@ import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
 import {event} from 'src/cloud/utils/reporting'
 
 const FlowFromRoute = () => {
-  const {id} = useParams<{id: string}>()
-  const {change, flows, currentID} = useContext(FlowListContext)
-
-  useEffect(() => {
-    change(id)
-  }, [id, change])
+  const {flow} = useContext(FlowContext)
```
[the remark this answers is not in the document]
asalem1: We are, in the context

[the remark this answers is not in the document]
asalem1: My bad. Turns out there's no way for that component after this change. Misunderstood the comment

src/flows/context/flow.current.tsx
```diff
@@ -54,17 +67,62 @@ export const FlowProvider: FC = ({children}) => {
...
+      }
+
+      return response.data.id
+    } catch (error) {
+      console.error({error})
+    }
+  }, [currentFlow, orgID])
+
+  const handleGetNotebook = useCallback(async notebookId => {
+    try {
```
drdelambre, line 107: i think you have to add a loading state here now that it's being fetched from someplace else

src/flows/components/header/MenuButton.tsx
```diff
@@ -45,7 +43,7 @@ const MenuButton: FC<Props> = ({handleResetShare}) => {
     event('clone_notebook', {
       context: 'notebook',
     })
-    const clonedId = await clone(flow.id)
+    const clonedId = await cloneNotebook()
```
drdelambre, line 46: if we're gonna be awaiting, we'd want the loading state for these as well so they know that their request has been sent and they dont keep smashing the submit button. idk if that's a component level thing or a context level thing

src/flows/components/header/MenuButton.tsx
```diff
@@ -57,7 +55,7 @@ const MenuButton: FC<Props> = ({handleResetShare}) => {
       context: 'notebook',
     })
     deletePinnedItemByParam(flow.id)
-    remove(flow.id)
+    deleteNotebook()
```
drdelambre, line 58: i think the same comment about waiting applies here too

src/flows/components/header/MenuButton.tsx
asalem1: k. I'd say it's a component layer thing so that the button is responsive to the load

drdelambre (approved):
outside of the couple of changes because we're moving to a more async pattern instead of running out of localstorage, this looks alright

PR #4210, opened 2022-03-25
chore: remove some dead code

asalem1:
These have been behind a UI feature flag for two years. Analyzing the metrics it looks like they've had zero engagement (meaning that since we've been reporting metrics, the flags have **never** been toggled on in a production environment) so it's time to remove them.

The flags in question are:
1.  customCheckQuery
2. fluxParser
3. matchingNotificationRules


drdelambre (approved):
note: timmy moved these tests to giraffe a bit ago so no worries about coverage loss

PR #4212, opened 2022-03-25
chore: refactor label list to remove memoize-one dependency

### LINKED ISSUE #4093: Refactor for Label list to remove memoizeone dependency
As part of https://github.com/influxdata/ui/issues/3590, we would like to refactor some areas of Label list to remove the dependency on `memoize-one`. Currently this refactor is scoped to:

* src/labels/components/LabelList.tsx

We should write tests where necessary to support this refactoring.

appletreeisyellow:
Closes #4093

This PR removes the `memoize-one` dependency in label list. Passed tests in `cypress/e2e/shared/labels.tests.ts` ✅


asalem1 (approved):
Out of scope for this, but it would be nice to have the overlay be part of the overlay context

appletreeisyellow:
Totally agree, the overlay here is a very old version, like over 3 years

PR #4217, opened 2022-03-25
chore: refactor member list to remove memoize-one dependency

### LINKED ISSUE #4094: Refactor for Member list to remove memoizeone dependency
As part of https://github.com/influxdata/ui/issues/3590, we would like to refactor some areas of Member list to remove the dependency on `memoize-one`. Currently this refactor is scoped to:

* src/members/components/MemberList.tsx

We should write tests where necessary to support this refactoring.

appletreeisyellow:
Closes #4094 

This PR removes the `memoize-one` dependency in member list


asalem1 (approved):
🥳

PR #4218, opened 2022-03-25
feat(ecommerce): Add Account Type Dropdown to Operator UI

abshierjoel:
This adds a dropdown to the operator UI which will allow for filtering based on account-type. By default, none will be selected al all accounts will be shown. After the operator selects one or many types, the results will be filtered based on the types chosen.

image

### DIFF
--- a/src/operator/ResourcesAccountType.tsx
+++ b/src/operator/ResourcesAccountType.tsx
@@ -12,7 +12,7 @@ const ResourcesAccountType: FC = () => {
   const availableTypes: AccountType[] = allAccountTypes
 
   const handleSelect = (selectedOption: string): void =>
-    accountTypes.includes(JSON.parse(selectedOption))
+    accountTypes.includes(selectedOption as AccountType)
       ? setAccountTypes(
           accountTypes.filter(x => x !== (selectedOption as AccountType))
         )
--- a/src/operator/ResourcesAccountType.tsx
+++ b/src/operator/ResourcesAccountType.tsx
@@ -12,7 +12,7 @@ const ResourcesAccountType: FC = () => {
   const availableTypes: AccountType[] = allAccountTypes
 
   const handleSelect = (selectedOption: string): void =>
-    accountTypes.includes(selectedOption)
+    accountTypes.includes(JSON.parse(selectedOption))
       ? setAccountTypes(
           accountTypes.filter(x => x !== (selectedOption as AccountType))
         )
--- a/src/operator/ResourcesAccountType.tsx
+++ b/src/operator/ResourcesAccountType.tsx
@@ -7,23 +7,27 @@ import React, {FC, useContext} from 'react'
 import {OperatorContext} from './context/operator'
 import {AccountType} from 'src/types'
 
+const allAccountTypes: AccountType[] = [
+  'cancelled',
+  'contract',
+  'free',
+  'pay_as_you_go',
+]
+
 const ResourcesAccountType: FC = () => {
   const {accountTypes, setAccountTypes} = useContext(OperatorContext)
-  const availableTypes: AccountType[] = allAccountTypes
 
-  const handleSelect = (selectedOption: string): void =>
-    accountTypes.includes(selectedOption as AccountType)
-      ? setAccountTypes(
-          accountTypes.filter(x => x !== (selectedOption as AccountType))
-        )
-      : setAccountTypes([selectedOption as AccountType, ...accountTypes])
+  const handleSelect = (selectedOption: AccountType): void =>
+    accountTypes.includes(selectedOption)
+      ? setAccountTypes(prev => prev.filter(x => x !== selectedOption))
+      : setAccountTypes(prev => [selectedOption, ...prev])
 
   return (
     <MultiSelectDropdown
       style={{width: '220px', marginRight: '20px'}}
-      options={availableTypes}
+      options={allAccountTypes}
       selectedOptions={accountTypes}
-      onSelect={handleSelect}
+      onSelect={(value: AccountType) => handleSelect(value)}
       menuTheme={DropdownMenuTheme.Onyx}
       emptyText="Account Types"
       buttonColor={ComponentColor.Primary}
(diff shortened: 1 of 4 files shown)


src/operator/ResourcesAccountType.tsx
```diff
@@ -0,0 +1,38 @@
...
+  ComponentColor,
+  DropdownMenuTheme,
+  MultiSelectDropdown,
+} from '@influxdata/clockface'
+import React, {FC, useContext} from 'react'
+import {OperatorContext} from './context/operator'
+
+const ResourcesAccountType: FC = () => {
+  const {accountTypes, setAccountTypes} = useContext(OperatorContext)
+  const availableTypes: string[] = allAccountTypes
```
asalem1, line 11: Why do we need this rather than just passing in the `allAccountTypes` as the selectedOptions

asalem1 (approved):

src/operator/ResourcesAccountType.tsx
```diff
@@ -0,0 +1,41 @@
...
+import React, {FC, useContext} from 'react'
+import {OperatorContext} from './context/operator'
+import {AccountType} from 'src/types'
+
+const ResourcesAccountType: FC = () => {
+  const {accountTypes, setAccountTypes} = useContext(OperatorContext)
+  const availableTypes: AccountType[] = allAccountTypes
+
+  const handleSelect = (selectedOption: string): void =>
+    accountTypes.includes(selectedOption as AccountType)
```
asalem1, line 15:
```suggestion
  const handleSelect = (selectedOption: AccountType): void =>
    accountTypes.includes(selectedOption)
```

src/operator/ResourcesAccountType.tsx
```diff
@@ -0,0 +1,41 @@
...
+      ? setAccountTypes(
+          accountTypes.filter(x => x !== (selectedOption as AccountType))
+        )
```
asalem1, line 18:
```suggestion
      ? setAccountTypes(prev => prev.filter(x => x !== selectedOption))
```

src/operator/ResourcesAccountType.tsx
```diff
@@ -0,0 +1,41 @@
...
+      : setAccountTypes([selectedOption as AccountType, ...accountTypes])
```
asalem1, line 19:
```suggestion
      : setAccountTypes(prev => [selectedOption, ...prev])
```

src/operator/ResourcesAccountType.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export default ResourcesAccountType
+
+const allAccountTypes: AccountType[] = [
+  'cancelled',
+  'contract',
+  'free',
+  'pay_as_you_go',
+]
```
asalem1, line 41: might wanna move this up to the top prior to declaring the functional component. Not sure if it'll complain or not about trying to access a const prior to it being declared since variable hoisting is different for `var` and `const`

src/operator/ResourcesAccountType.tsx
```diff
@@ -0,0 +1,41 @@
...
+  DropdownMenuTheme,
+  MultiSelectDropdown,
+} from '@influxdata/clockface'
+import React, {FC, useContext} from 'react'
+import {OperatorContext} from './context/operator'
+import {AccountType} from 'src/types'
+
+const ResourcesAccountType: FC = () => {
+  const {accountTypes, setAccountTypes} = useContext(OperatorContext)
+  const availableTypes: AccountType[] = allAccountTypes
```
asalem1, line 12: why are we doing this if it exists as a const outside of the component?

src/operator/context/operator.tsx
```diff
@@ -63,6 +73,7 @@ export const OperatorProvider: FC<Props> = React.memo(({children}) => {
       const resp = await getOperatorAccounts({
         query: {
           query: searchTerm,
+          accountTypes: accountTypes,
```
asalem1, line 76:
```suggestion
          accountTypes,
```

src/operator/context/operator.tsx
```diff
@@ -77,12 +88,14 @@ export const OperatorProvider: FC<Props> = React.memo(({children}) => {
...
     }
-  }, [searchTerm, dispatch])
+  }, [accountTypes, searchTerm, dispatch])
 
   const handleGetOrgs = useCallback(async () => {
     try {
       setOrgsStatus(RemoteDataState.Loading)
-      const resp = await getOperatorOrgs({query: {query: searchTerm}})
+      const resp = await getOperatorOrgs({
+        query: {query: searchTerm, accountTypes: accountTypes},
```
asalem1, line 97:
```suggestion
        query: {query: searchTerm, accountTypes},
```

asalem1 (changes requested):

src/operator/ResourcesAccountType.tsx
abshierjoel: Good call. I'll eliminate it.

src/operator/context/operator.tsx
abshierjoel: Thanks!

src/operator/context/operator.tsx
abshierjoel: Thanks!

src/operator/ResourcesAccountType.tsx
abshierjoel: Right on! I'll move it up!

src/operator/ResourcesAccountType.tsx
```diff
@@ -0,0 +1,41 @@
...
+          accountTypes.filter(x => x !== (selectedOption as AccountType))
+        )
+      : setAccountTypes([selectedOption as AccountType, ...accountTypes])
+
+  return (
+    <MultiSelectDropdown
+      style={{width: '220px', marginRight: '20px'}}
+      options={availableTypes}
+      selectedOptions={accountTypes}
+      onSelect={handleSelect}
```
asalem1, line 26:
```suggestion
      onSelect={(value as AccountType) => handleSelect(value)}
```

abshierjoel: Added! 👍

src/operator/ResourcesAccountType.tsx
abshierjoel: Because the `onSelect` event returns a string, @asalem1 offered [this solution](https://github.com/influxdata/ui/pull/4218#discussion_r846298246) to this.

asalem1 (approved):

PR #4219, opened 2022-03-25
chore: consolidate logos for Sources page

### LINKED ISSUE #4208: Telegraf plugins: too many similar logos
Telegraf plugins on the Load Data > Sources page have logos. Many of them use the same logo while keeping a copy of it as a separate file.

- [x] delete copies of the same logo
- [x] rename commonly used logos with files names that are more generic

TCL735:
Closes #4208 

Load Data > Sources has a lot of images. Many of the images are exactly the same. In such cases, let's use 1 image file rather than making duplicate files for the same image.


TCL735:
The page loads fine with no console errors



https://user-images.githubusercontent.com/10736577/160213857-c08eb08a-41f6-4d20-bf07-80dd17afa041.mp4

hoorayimhelping (approved):
just going to trust on this one

PR #4234, opened 2022-03-29
chore: upgrade cypress in package.json to 9.5.2

subirjolly:
chore: upgrade cypress in package.json to 9.5.2


TCL735 (changes requested):
Why is the diff for yarn-lock so large?

subirjolly:
> Why is the diff for yarn-lock so large?

Something was wrong with my branch. So, Recreated with the same changes.

PR #4262, opened 2022-03-30
fix: weird header

### LINKED ISSUE #4240: notebooks: Upgrade button appears in strange spot as scroll down list of notebooks
The 'upgrade now' button moves to a weird spot when the notebook buttons are resized (see screenshot).  This happens as you scroll down in a list of notebooks.
Screen Shot 2022-03-29 at 11 02 19 AM

drdelambre:
Closes #4240

i wish i could say i knew what this flag does but i do not

Peek 2022-03-30 09-45


asalem1 (approved):
Let's see some pics

PR #4263, opened 2022-03-30
fix: the way notebooks are updated

### LINKED ISSUE #4255: Notebooks try to execute the Flux at **EVERY KEYSTROKE**
I like Notebooks, but someone there at Influx has decided that customers want to see the results of their code run after **EVERY KEYSTROKE**
It literally executes the flux after typing a new line, or even after typing one "/" when you were trying to comment out a line

image

You have *got* to be kidding.  It's simply a terrible idea.
It will now take customers 100 times as long to write code.
If they are ready to execute code then there is a big fat **RUN** button.

image

Please please please undo this and restore some coding sanity!

What’s more concerning is that I’m not sure that developers are testing these features on enough customers before they are implemented them on the platform and forcing everyone to use it.
Did someone approve this idea for release ?  Did it go through any customer testing before that ?
I’m all for progress and bleeding-edge, but your developers are breaking things as fast as they are fixing them.

I’ve been using influx for 2 years now and I’ve seen a huge uptick in new “features” that break things…

:-)

### LINKED ISSUE #4258: In Notebooks, in Presentation mode, the resizing of dashboard elements is now broken.
### To Reproduce
- Create a notebook 
- Create two charts in the notebook (or a chart and a table)
- Switch to Presentation mode
- Attempt to make the first chart bigger or smaller in either the X or Y direction
image

### Observed Behaviour
- The chart appears to resize while holding the handle, 
- After releasing the handle the chart pings back to the original size
- It is *NOT* because the new size is not possible.  This used to work.  I used to use it a lot
- Resizing is now broken
- There is a workaround, it appears that by resizing the *second* chart to be really small, the *first* chart will accept *certain*  combinations of resizing.

### Expected Behaviour
Obviously resizing that is shown while the handle is being dragged should stick (assuming no rules are broken about the new arrangement, but these should be reflected in what's happening *while* the handle is being dragged).

### LINKED ISSUE #4259: In Notebooks, renaming of the Notebook does not update in certain cases
### To Reproduce
- Create a Notebook
- Rename the notebook "MY NEW NOTEBOOK" at the top of the page *BUT DO NOT PRESS ENTER*
- Tab away from the notebook name.
- Save the notebook (which will also trigger Autosave)
- Create whatever elements you need to in the Notebook
- 
image

- Close the notebook and return to Notebooks list

### Observed Behaviour
- The Notebook you just created is still called "unnamed" in the list
image

- Edit the Notebook again.  You'll see it's un-named at the top.
image
- Rename the notebook "MY NEW NOTEBOOK" at the top of the page *THIS TIME PRESS ENTER*
- Close the notebook and return to Notebooks list
- The Notebook you just created is still called "unnamed" in the list
- Try to edit it in the list of Notebooks and press ENTER
- The name changes.  But ....
- Now edit the Notebook.  You'll see that the dashboard you created is now gone/reset
(!!!!)

### Expected Behaviour
I don't know where to start.  Notebooks should stay named, wherever they are renamed from, and shouldn't loose what you have saved.

asalem1:
Closes #4258
Closes #4259
Closes #4255

This PR updates the way that notebook updates are handled. Whereas beforehand a deep copy was created and used as a reference to update the flow, this PR ensures that the only data that's being updated should be passed into the current flow so that the update registers the most recent changes and doesn't have stale data.

In addition, this issue resolves a previously unreported bug with our renaming where if a user tabs out of naming a resource, those name changes didn't get registered

### DIFF
--- a/src/flows/context/flow.current.tsx
+++ b/src/flows/context/flow.current.tsx
@@ -7,10 +7,10 @@ import {Doc} from 'yjs'
 import {WebsocketProvider} from 'y-websocket'
 import {serialize, hydrate} from 'src/flows/context/flow.list'
 import {useParams} from 'react-router-dom'
-import {getNotebook, postNotebook} from 'src/client/notebooksRoutes'
+import {getNotebook, postNotebooksClone} from 'src/client/notebooksRoutes'
 import {event} from 'src/cloud/utils/reporting'
 import {deleteNotebook} from 'src/client/notebooksRoutes'
-import {getAllAPI, pooledUpdateAPI} from 'src/flows/context/api'
+import {pooledUpdateAPI} from 'src/flows/context/api'
 import {useDispatch} from 'react-redux'
 import {notify} from 'src/shared/actions/notifications'
 import {
@@ -20,7 +20,6 @@ import {
 import PageSpinner from 'src/perf/components/PageSpinner'
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
 import {RemoteDataState} from '@influxdata/clockface'
-import {incrementCloneName} from 'src/utils/naming'
 
 const prettyid = customAlphabet('abcdefghijklmnop0123456789', 12)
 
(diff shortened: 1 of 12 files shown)


src/flows/context/flow.current.tsx
```diff
@@ -94,6 +96,9 @@ export const FlowProvider: FC = ({children}) => {
       delete _flow.data.id
 
       const response = await postNotebook(_flow)
+      // const response = await postNotebooksClone({id: currentFlow.id, data: {
+      //   orgID
+      // }})
```
TCL735, line 101: ✂️

TCL735 (changes requested):

src/flows/context/flow.current.tsx
asalem1: I'm gonna add that back in the next PR

TCL735 (approved):

Issue #4057, opened 2022-03-09
firstMile: refactor ConnectionInformation and DataListening


ChitlangeSahas:
When we are confident that the new component workarounds are working fine, we want to incorporate those changes in the files that we copied this from. 

verifyStep/DataListening
verifyStep/ConnectionInformation

PR #4264, opened 2022-03-30
refactor: take out common parts of two similar components (DataListening.tsx)

ChitlangeSahas:
Connects https://github.com/influxdata/ui/issues/4057

The two files have common functionality because they were a copy of each other, but needed to be tweaked for slightly different behavior.


src/shared/utils/dataListening.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        if (timePassed > TIMEOUT_MILLISECONDS) {
+          updateResponse(LoadingState.NotFound)
+          clearInterval(intervalId)
+        }
+      })
+    } catch {
+      updateResponse(LoadingState.Error)
+      clearInterval(intervalId)
+    }
```
hoorayimhelping, line 33: I don't think the `try/catch` will work the way you're expecting here - it requires being in an `async/await` call for it to work like non-asynchronous code. So since this is a more traditional promise using `then`, the `try/catch` should be changed to a `then().catch()` which is identical, functionally. We can also just chain the `then()` and `catch()` together without creating the `status` variable

```suggestion
    checkForData(orgID, bucket).then(value => {
      if (value === LoadingState.Done) {
        updateResponse(LoadingState.Done)
        clearInterval(intervalId)
        return
      }
      if (value === LoadingState.Error) {
        updateResponse(LoadingState.Error)
        clearInterval(intervalId)
      }
      const timePassed = Date.now() - startTime

      if (timePassed > TIMEOUT_MILLISECONDS) {
        updateResponse(LoadingState.NotFound)
        clearInterval(intervalId)
      }
    }).catch(() => {
      updateResponse(LoadingState.Error)
      clearInterval(intervalId)
    })
```

hoorayimhelping (commented):
looks good to me generally, but there's a suggestion for making the `then()` and `catch` parts work the way you expect

src/shared/utils/dataListening.ts
ChitlangeSahas: This makes sense, I added the change and it works !

hoorayimhelping (approved):

PR #4269, opened 2022-03-30
feat: transform function signature to example for injection

### LINKED ISSUE #3638: Flux Dynamic Help Panels: transform function signature to example for injection
Currently, there is a feature that allows the user to select "inject" on a Flux function and have an example of that function inserted into the Script Editor. The example code is static and only exists for the limited set of functions found in the static documentation.

We want to have an example for any function found through the Fluxdocs service. The example function should rely on the `fluxType` property which is the function signature, to create an example function. Possibly, make use of the `name` property also.  Build this functionality to allow the inject feature to continue to work with the new Fluxdocs service.

We may need to change interfaces in the UI. This should only be done in the UI because changes to this feature are easier to maintain in the UI, as this is a UI feature.

- use existing properties in the interface such as `fluxType` and `name` to create the example code
- should not require storage
- may update or create UI specific interfaces

blegesse-git:
Closes #3638 

`getFluxExample` function takes in flux function object as a parameter and parses a signature that can be injected onto script editor. 

https://user-images.githubusercontent.com/66275100/161821573-71ff8760-93a9-49e7-9e77-32074355ace4.mov


drdelambre:
i know you're still working on this and i'll be the first to admit that i have little context on this project, so please excuse my ignorance while i entertain a curiosity, but the architecture of this PR seems to imply that we're choosing to have every user construct the example from the api definition at the point of... injection i think? my question is why we chose this path instead of building an example for all functions when we construct the data source for the api so that we only have to do this logic once (when we import changes from flux) instead of... a bunch (on every user's computer everytime they interact with the interface). you all prolly talked it over so this is more for my own mental model than anything.

blegesse-git:
> i know you're still working on this and i'll be the first to admit that i have little context on this project, so please excuse my ignorance while i entertain a curiosity, but the architecture of this PR seems to imply that we're choosing to have every user construct the example from the api definition at the point of... injection i think? my question is why we chose this path instead of building an example for all functions when we construct the data source for the api so that we only have to do this logic once (when we import changes from flux) instead of... a bunch (on every user's computer everytime they interact with the interface). you all prolly talked it over so this is more for my own mental model than anything.

Hey Alex, So the json data created and maintained by the Flux team _does_ have an example property for each flux function, but these examples contain extra details that aren't needed when using these functions on the editor. And even if we were to use these examples, we would have to parse them on the front end at the point of injection (like how I'm doing on this pr). I'll let the project lead chime in here @TCL735, maybe he can give more insight on why we chose this route.

TCL735:
> i know you're still working on this and i'll be the first to admit that i have little context on this project, so please excuse my ignorance while i entertain a curiosity, but the architecture of this PR seems to imply that we're choosing to have every user construct the example from the api definition at the point of... injection i think? my question is why we chose this path instead of building an example for all functions when we construct the data source for the api so that we only have to do this logic once (when we import changes from flux) instead of... a bunch (on every user's computer everytime they interact with the interface). you all prolly talked it over so this is more for my own mental model than anything.

Originally, I did consider putting the example on the backend. But didn't do it because doing so is essentially making a hardcoded example that takes no context on how the inject button is used. As an mvp, we decided not to mess with additional, nice-to-have features. But still kept the example/placeholder creation on the frontend.

We've struggled with the placeholder since... forever. Recently, the Flux team has come up with a proposal to leverage the LSP to assist with the placeholder. That will take a while to get to fruition. When it does, the frontend will be the place that does the example/placeholder logic (or at least interact with the LSP in some way), so storing this on the backend in the Fluxdocs service would be moot.

Having the Fluxdocs service hold the example/placeholder _still_ has a lot of merit though. If we can't execute the idea in the above paragraph, we may end up with having this in the service. We'll see.

src/timeMachine/utils/insertFunction.ts
```diff
@@ -24,3 +24,110 @@ export const generateImport = (
```
blegesse-git, line 24: I'll remove the comments after the pr gets approved.

src/timeMachine/utils/insertFunction.ts
```diff
@@ -24,3 +24,110 @@ export const generateImport = (
 
   return importStatement
 }
+
```
blegesse-git, line 27: I'll remove comments once pr gets approved. Just wanted to provide context for now :)

src/timeMachine/utils/insertFunction.ts
```diff
@@ -24,3 +24,110 @@ export const generateImport = (
...
+export const getFluxExample = func => {
+  const {name, fluxType} = func
+
+  // "fluxType": "(<-tables:stream[A], every:duration, ?groupColumns:[string], ?unit:duration) => stream[B] where A: Record, B: Record"
+
+  // get copy of fluxtype signature before arrow sign
+  const arrowSign = fluxType.lastIndexOf('=') // using lastIndexOf because of edge cases like array.filter
+  const fluxParam = fluxType.slice(0, arrowSign) // fluxParam = (<-tables:stream[A], every:duration, ?groupColumns:[string], ?unit:duration)
```
TCL735, line 35: Looking at `aggregate.count` I get this when i console.log the variables out in an object
```
{
  arrowSign: 116,
  fluxParam: '(?column:A, ?fill:B) => {reduce:(state:int, values…t:(values:[D]) => int, fill:B, compute:(state:C) '
}
```

This looks wrong already. We might need to rethink this logic.

Also, can you move this util file to `shared` - since this gets used in both Time Machine and Notebooks. I don't think we should keep this in `timeMachine`. It might get mistakened in the future as Time Machine specific code which this isn't.

TCL735 (changes requested):
Let's start with addressing these first. Since this will probably affect the rest of the logic. I wanted to get a corrected version before another round of review.

src/timeMachine/utils/insertFunction.ts
blegesse-git: good idea. I moved the `getFluxExample` function into its own file in the shared folder. 

my thought process there is; we need to separate the actual parameters from the rest of `fluxType` content. It's guaranteed everything to the left of the arrow sign contains the function's parameters, and everything to the right of the arrow is irrelevant information we wont need in the parsing process. 

Initially, I was using `fluxType.indexOf('=')` to find the position of the arrow sign, but this approach didn't work for functions like `array.filter`. Some flux types have multiple arrow signs, making it difficult to know where the parameters start and stop. So I figured using `lastIndexOf` would be a viable solution for this first step. But this means, we need to add logic that can weed out or clean up `fluxParam` and make sure we parse out the correct parameters. 

Because each function's fluxType is unique in some way, it's been difficult to write code that takes these nuances into account. I found it easier to have different blocks of code that can act like "barriers" and by the time the parameters get to the end, only the required parameters remain. 
With that being said, I'm absolutely all ears if you have suggestions here. What other approach can we take in order to separate the parameters from the rest of the content in the `fluxType` value?

blegesse-git: good idea. I moved the `getFluxExample` function into its own file in the shared folder. 

my thought process there is; we need to separate the actual parameters from the rest of `fluxType` content. It's guaranteed everything to the left of the arrow sign contains the function's parameters, and everything to the right of the arrow is irrelevant information we wont need in the parsing process. 

Initially, I was using `fluxType.indexOf('=')` to find the position of the arrow sign, but this approach didn't work for functions like `array.filter`. Some flux types have multiple arrow signs, making it difficult to know where the parameters start and stop. So I figured using `lastIndexOf` would be a viable solution for this first step. But this means, we need to add logic that can weed out or clean up `fluxParam` and make sure we parse out the correct parameters. 

Because each function's fluxType is unique in some way, it's been difficult to write code that takes these nuances into account. I found it easier to have different blocks of code that can act like "barriers" and by the time the parameters get to the end, only the required parameters remain. 
With that being said, I'm absolutely all ears if you have suggestions here. What other approach can we take in order to separate the parameters from the rest of the content in the `fluxType` value?

src/shared/utils/fluxExample.ts
```diff
@@ -0,0 +1,25 @@
...
+        if (parameter.required) {
+          // add a comma if the current injected list is not all spaces
+          if (injectedParameters.trim().length !== 0) {
+            injectedParameters += ', '
+          }
+          injectedParameters += `${parameter.name}: `
+        } else {
+          injectedParameters += ' '
+        }
+        return injectedParameters
```
hoorayimhelping, line 20: if possible, don't mix concatenation and string interpolation - concatenation can be confused for the mathematical addition operator.

We could also reformulate this to have less syntax and fewer variable assignments (and thus be easier to follow):

```ts
if (parameter.required) {
  // add a comma if the current injected list is not all spaces
  if (injectedParameters.trim().length !== 0) {
    injectedParameters += ', '
  }
  return `${injectedParameters}${parameter.name}: `
}
return `${injectedParameters} `
```

This is a personal preference thing - I personally find it easier to set defaults and not deal with `else` clauses if possible.

I'd also argue that using `reduce` provides more overhead for not a lot of benefit, since we're overwriting strings anyway, which kind of defeats the purpose of using immutable functional paradigms.

```ts
let injectedParameters
for (const parameter of fluxParameters) {
  if (parameter.required) {
    // add a comma if the current injected list is not all spaces
    if (injectedParameters.trim().length !== 0) {
      injectedParameters += ', '
    }
    inectedParameters = `${injectedParameters}${parameter.name}: `
    continue
  }
  injectedParameters = `${injectedParameters} `
}

example = `${packageName}.${name}(${injectedParameters})`
```

but this is more of a preference thing that we don't have standards on. My personal preference is that `reduce` should be used to reduce a collection of values into a single value, not used to convert an array to an object or do string manipulation. Your choice on this comment, since it's a preference thing.

hoorayimhelping (approved):
Looks good. Have a stylistic suggestion that you don't need to follow if you feel strongly about how you structured your code.

src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -98,7 +87,6 @@ const Query: FC<PipeProp> = ({Context}) => {
         header: !!fn.package ? `import "${fn.package}"` : null,
       }
       inject(options)
-      event('Inject function into Flux Script', {fn: fn.name})
```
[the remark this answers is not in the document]
blegesse-git: Screen Shot 2022-04-07 at 12 16 04 PM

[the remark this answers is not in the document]
blegesse-git: the event inside `view.tsx` is actually a duplicate. We already have an event for injecting functions inside `DynamicFunctionsList.tsx`. That's why I removed it

src/shared/utils/fluxExample.ts
TCL735: This is a good observation. Sometimes, I have gotten so used to and familiar with certain utilities that I forget that not everyone may see the usage and understand immediately. Readability is better for maintenance.

PR #4271, opened 2022-03-30
chore: use resolutions for sec fixes

jdstrand:
Since (some of) the less urgent updates for the dependabot alerts have been lingering due to our dependencies not pulling in updated versions, I decided to explore other options and came across: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/. This should resolve all dependabot alerts except `clean-css`. There are 3 different commits:

* chore: resolve minimist to 1.2.6
* chore: resolve ansi-regex to 4.1.1 (ansi-regex is used in a number of places. The 4.0.0 series should use 4.1.1 though, so use 'yarn why ansi-regex' to pick out what to resolve to the various differently fixed versions). The others use 3.0.1, 5.0.1 and 6.0.1 since those are the ones that have the fix
* chore: resolve nanoid to 3.1.31 (first with the fix)

Each commit consisted of a) updating `resolutions` in `package.json` for the dependency, running `yarn install`, examining `git diff` and then committing.

Please be advised that the mimimist update resolves a high issue (but I've not triaged it to see if it is actually high for us yet). I did not test this beyond eyeballing that the semvers looked right.

### DIFF
--- a/package.json
+++ b/package.json
@@ -227,6 +227,7 @@
     "mocha-junit-reporter/**/ansi-regex": "^3.0.1",
     "strip-ansi/**/ansi-regex": "^5.0.1",
     "table/**/ansi-regex": "^4.1.1",
-    "webpack-dev-server/**/ansi-regex": "^6.0.1"
+    "webpack-dev-server/**/ansi-regex": "^6.0.1",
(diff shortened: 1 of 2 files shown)


asalem1 (approved):

asalem1:
@jdstrand you might want to rebase the las test changes from master to get this PR to pass

PR #4274, opened 2022-03-31
chore: remove unused actions from Collectors component

Palakp41:
Removes unused actions from Collectors component and it's tests


hoorayimhelping (approved):
lgtm!

these are still being used [here](https://github.com/influxdata/ui/blob/f647b7c231efff819e94fd6e88083ac6efc6570f/src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx) right? which is what is preventing us from deleting them and the associated reducers completely, right?

Palakp41:
> lgtm!
> 
> these are still being used [here](https://github.com/influxdata/ui/blob/f647b7c231efff819e94fd6e88083ac6efc6570f/src/writeData/components/PluginAddToExistingConfiguration/Configure.tsx) right? which is what is preventing us from deleting them and the associated reducers completely, right?

Exactly! These actions are still in use but not in the `Collectors.tsx`, so cleaned this one component up. Thanks for quick review :)

Issue #3896, opened 2022-02-16
firstMile: End to End Testing


Luckette:
Onboarding: End to End Testing

PR #4280, opened 2022-03-31
test: basic e2e tests for firstMile

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/3896

First pass of e2e tests for the first mile / python part of it.

### DIFF
--- a/cypress/e2e/shared/firstmile.tests.ts
+++ b/cypress/e2e/shared/firstmile.tests.ts
@@ -21,35 +21,35 @@ describe('First mile experience', () => {
 
     it("should be able to run through the python's wizard pages", function() {
       // first page is overview
-      cy.getByTestID('overview-page-header').should('exist')
+      cy.contains('Hello, Time-Series World!')
 
       // second page is install dependencies
       cy.getByTestID('next-button').click()
-      cy.getByTestID('install-dependencies-page-header').should('exist')
+      cy.contains('Install Dependencies')
 
       // third page is create token
       cy.getByTestID('next-button').click()
-      cy.getByTestID('tokens-page-header').should('exist')
+      cy.contains('Create a Token')
 
       // fourth page is initalize client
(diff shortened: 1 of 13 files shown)


cypress/e2e/shared/firstmile.tests.ts
```diff
@@ -0,0 +1,55 @@
...
+    cy.signin()
+    cy.setFeatureFlags({firstMile: true})
+  })
+
+  it('should load the main home-page that has the language tiles', function() {
+    cy.getByTestID('home-page--header').should('exist')
+    cy.getByTestID('language-tiles--scrollbox').should('exist')
+  })
+
+  describe('Python', () => {
```
hoorayimhelping, line 13: let's add another describe block for the subway navigation. The things it should check:

- doing something like `cy.contains('Create a Token').click()` takes you to the tokens page
  - this is a case where clicking a nav link with the text `Create a Token` and then checking that the page it goes to has the text `Create a Token` is probably a good behavior to test, since from a product standpoint, we want the header text to match the link text
- clicking `cy.contains('Next')` on a page takes you to the next page, similar with `cy.contains('Previous')`
  - test that the navigation doesn't let you go past the last page and before the first page.

hoorayimhelping (changes requested):
I'm worried that this is going to create more work for us as we start to iterate on this and potentially change things. I think we should wait to assert on the content of the pages until the feature stabilizes and gets closer to being turned on in production.

One worry is that this couples the tests to the current page order and current test ids of the pages, so that if  a page changes, we'll also have to change this test since it's ordered in a specific way.

I also think we don't need to add test ids to the buttons, since they already have stable text we can target with `cy.contains` (`Next` and `Previous`).

blegesse-git (approved):
looks good to me

hoorayimhelping (approved):

PR #4286, opened 2022-04-01
feat: limiting down to a single measurement

### LINKED ISSUE #4253: Notebooks - Force Measurement Selection

<p></p>

The current Notebooks interface allows for users (after selecting a bucket) to select a  _measurement or a _field.   This then allows them to choose a field without also requiring them to choose a measurement which causes the metadata lookup to need to grab metadata from all measurements.  This has been reported to cause performance issues with large datasets.

<p></p>

But as a SaaS service provider, I want to force my users to choose a measurement so that I can narrow that metadata and give the user a better user experience.

<p></p>

## Suggested Implementation

<p></p>

- Remove the group/filter option from the first panel the user sees after selecting a bucket.

- Give that panel a title of "Measurement"

- Show the user only a list of measurements to choose from within that panel

- The list of measurements is the complete list for that bucket, not filtered by time range

- Still allow the user to search for a measurement, but don't include fields or tags (only measurements)

- Subsequent panels work as they do today, except there is no longer a _measurement option in them

<p></p>

Design from design team:

## Acceptance Criteria

<p></p>

- User will see a measurement selection panel after selecting a bucket

- User will see a list of measurements in that panel

- User will be able to search through the list of measurements in that panel, or scroll through them

- All measurements will be shown (not just the measurements in the current time range (a change of current behavior)

- User will be able to select a measurement

- Field Selection won't be available until after a measurement has been selected

- User won't be able to add more panels until after they have made a measurement selection

- Measurement panel won't be available until after bucket selection; it could be done like we currently do with bucket selection:  

(but in this case, it would say to select a measurement first)

- User will NOT be able to select more than one measurement

- Once a measurement is selected the next panel will not give an option for _measurement

- When building the flux query, the query will have only one measurement in the script

- When converting to flux the user will only have one measurement selected (if they have selected one)

- User doesn't see a filter dropdown in the measure selection panel

- For pre-existing notebooks that already have more than one measurement, the old UI will be shown for backward compatibility

## Design Notes

<p></p>

- The current design has a number showing the number of measurements selected; this needs final determination as some feel that more than 1 measurement should never be allowed

<p></p>

<p></p>

<p></p>

<p></p>

drdelambre:
Closes #4253 

makes it where you only have the option to select a single measurement in the notebooks query builder. it's behind a flag so we can see how users feel about it before going all in

looks just like the figma!
Screenshot 2022-04-01 1 23 37 PM


src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -114,16 +115,35 @@ const Card: FC<Props> = ({idx}) => {
     })
   }
 
+  const isCompliant =
+    isFlagEnabled('newQueryBuilder') &&
+    idx === 0 &&
+    (card.keys.selected.length === 0 ||
+      (card.keys.selected.length === 1 &&
+        card.keys.selected[0] === '_measurement')) &&
+    card.aggregateFunctionType !== 'group' &&
```
asalem1, line 124: why does the aggregateFunctionType matter?

drdelambre: most of these are for handling migrations. like if someone put a group on a tag right up front, we still need the frontend to support that

asalem1: So on the topic of migrations, this should be compatible with `/from` dashboards, etc... and all existing notebooks should still work if the flag gets turned on right?

drdelambre: yes

asalem1 (approved):

drdelambre:
tested in prod,  looks like we have to clear the query everytime a measurement is selected, when the measurement card is in compliance

Issue #3603, opened 2022-01-14
Bug Duty: Telegraf Plugins - updating via script should include markdown files


TCL735:
Currently, when the `telegraf-plugins:update` script is run, only the .conf files for plugins are checked and updated for all plugins.

The above script should also include markdown files, because all markdown content rendered in the UI at `/orgs/{orgId}/load-data/telegraf-plugins/{pluginName}` is (should be) a copy-paste of the same markdown asset found in the [Telegraf repository](https://github.com/influxdata/telegraf/tree/master/plugins/inputs) for each plugin.

By including the markdown files for updating via script, we cut down on the amount of manual work required when updating the Telegraf plugins list in the UI whenever there is a new release of Telegraf.

TCL735:
One related issue is the way links are written in markdown. There are a handful of links written as relative links. This does not work for the UI, because we are consuming the READMEs in a different file structure than where they originate (the Telegraf repo).

We need all absolute links, preferably.

PR #4290, opened 2022-04-01
fix: update Telegraf plugin READMEs to use absolute paths to github links

TCL735:
Follow-up to #3603 

Per the Telegraf team, they want to stick to relative paths because it works better for them when they are looking at their own documentation in their IDE. Since they are the maintainers for the READMEs and relative paths allow them to work efficiently, this speeds up the process of updating READMEs.

Hence, I have updated the script here to change the relative paths for github links to absolute paths, so that users can click on them properly in the UI.

Also, I ran the script again to pick up the few cases where relative paths are used.


src/writeData/utils/updateTelegrafPlugins.mjs
```diff
@@ -348,13 +348,28 @@ const formatConfigurationText = configurationText => {
...
 
   return readmeText
     .map(line => {
-      return line.replace(relativeLink, replacement)
+      return [...relativeLinkReplacements.entries()].reduce(
+        (modifiedLine, replacementPair) => {
+          return modifiedLine.replace(replacementPair[1], replacementPair[0])
+        },
+        line
+      )
```
hoorayimhelping, line 372: can we get some unit tests for this functionality, since if it doesn't work, it'll break documentation?

hoorayimhelping (approved):
lgtm, but the function that makes the links relatives could use a few assertions in a unit test

src/writeData/utils/updateTelegrafPlugins.mjs
TCL735: Added.

src/writeData/utils/updateTelegrafPlugins.mjs
```diff
@@ -321,44 +326,6 @@ const fetch = (path, parsedPath) => {
...
-  const replacement =
-    '(https://github.com/influxdata/telegraf/tree/master/plugins/'
-
-  return readmeText
-    .map(line => {
-      return line.replace(relativeLink, replacement)
-    })
-    .join()
-}
-
```
TCL735, line 361: Moved into another file because of the `.mjs` extension here, which makes it annoyingly difficult to write Jest unit tests.

hoorayimhelping (approved):
thanks for adding those tests in!

src/writeData/utils/updateTelegrafPlugins.mjs
TCL735: Ended up moving this back into this file but at a different location.

Issue #4154, opened 2022-03-19
LSP UI Improvements


garylfowler:
## SPIKE Description

<p></p>

The query experience project is leaning heavy towards a flux editor focused user interface.  As such, it is expected that users will be writing flux manually (with some UI based injection help).  Because of this LSP may be a very important aspect of the solution.

<p></p>

It has been communicated that the current way that the cloud2 interface is utilizing LSP is not as optimal as it could be, which is causing some of the functionality that can be used in other Influx interfaces that utilize it to not be as valuable within the cloud2 UI.  

<p></p>

This spike is for seeing what changes can be made to use all of the features of the  LSP.

<p></p>

The results of this spike will then be evaluated to determine what should be included in the new Query experience being looked at by the Query Experience committee.

<p></p>

## Spike Deliverables

<p></p>

- ~~Proof-of-concept/Prototype of Query Editor with maximum LSP usage~~

- Simple document outlining learnings from the process and how difficult it may be for implementation into an existing or new Query builder

<p></p>

<p></p>

garylfowler:
Link to feature: https://influxdata.productboard.com/feature-board/planning/features/12552818

wiedld:
Had an initial conversation with the LSP team. There is some question as to how the UI is interacting with the LSP communication stream, as well as monaco editor versioning. Will likely require a bit of UI research (e.g. if/why interrupting the stream) and code changes (e.g. allow the pass-through to monaco).

wiedld:
Spike in progress. Implementation steps:
- [ ] Refactor that part of the codebase, such that it's more approachable for UI developers.
- [ ] Version bump latest libs, and access feature availability of the libs we use to provide the LSP interaction.
- [ ] Update currently interaction with LSP, for additional features.
- [ ] Document learnings, and impact on Query Experience.

PR #3997, opened 2022-03-03
refactor: make typing in the monaco editor less aggressive

TCL735:
The use of `force` as an option should be reserved for cases when we do not want Cypress to act like a user.

In all our tests that involve typing into the Monaco Editor, we _do_ want Cypress to act like a user. We want to ensure that the elements are actionable. No forcing.

Make typing in the Monaco Editor less aggressive:
- remove force clicking
- remove force typing
- do not focus
- remove chaining to avoid detached elements when Cypress retrys

PR #4293, opened 2022-04-02
chore(4154): update monaco packages

### LINKED ISSUE #4299: LSP: assess and upgrade libraries used.
Version bump latest libs, and access lib feature availability (a.k.a what LSP features these libs supports).

Part of this spike:
https://github.com/influxdata/ui/issues/4154

wiedld:
Part of #4154
Closes #4299 

## Done:
* Bumped `monaco-editor`.
    * set to the most recent version supported by `react-monaco-editor`
    * required 1 line change for KeyCode defintions. (Doc link in commit.)
    * required webpack build to use babel preset-env for the monaco-editor ESM. (Doc link in commit.)
    * Required bump in `monaco-editor-webpack-plugin`. (Also referenced in commit).
* Bumped `monaco-editor-textmate` and `onigasm`.
* Bumped `monaco-languageclient`.
    * This is how our flux-lsp is integrated into the editor.
* See all individual commit messages for why/how lib versions were chosen.

## Quality control:
* confirmed cross compatibilities in versions
* manual QA’ed could still use the editor, with syntax highlighting
* confirmed clean web pack build (no errors)
   * we do now get a warning, because we import and use the global monaco-editor worker script without then creating the worker.
   * I am unclear of the source of this warning, since the example docs never have explicit creation of the worker (instead, the monaco-editor lib handles it.) But we are breaking the docs pattern already...and I was intending to explore this further in the next PR. 
   * **NOTE:** Is only a webpack warning, and not breaking -- as in, no loss of current functionality.
    ```
       WARNING in new Worker() will only be bundled if passed options that include { type: 'module' }.
       Received: new Worker()(null, {"name":null})
        @ ./node_modules/monaco-editor/esm/vs/editor/browser/services/editorWorkerService.js 88:0-85 691:32-52
        @ ./node_modules/monaco-editor/esm/vs/editor/browser/services/webWorker.js
        @ ./node_modules/monaco-editor/esm/vs/editor/standalone/browser/standaloneEditor.js
        @ ./node_modules/monaco-editor/esm/vs/editor/editor.api.js
        @ ./node_modules/monaco-editor/esm/vs/editor/editor.api.js
        @ ./node_modules/react-monaco-editor/lib/index.js
        @ ./src/shared/components/FluxMonacoEditor.tsx
     ```

## Visual confirmation:
syntax highlighting
Screen Shot 2022-04-01 at 7 04 39 PM

suggestions
Screen Shot 2022-04-01 at 7 05 01 PM


wiedld:
Have two (non-blocking questions).

### Question 1:

We are using textmate for our default syntax handling in the monaco editor, and then we add on top the additional grammars per editor (e.g. flux):
* we removed the use of the simple wireTmGrammars() interface, which can handle multiple grammars
* instead placed in our own custom Tokenizer. Why?
* commit: https://github.com/influxdata/ui/commit/5c21941c62f3f60ae632e1704e56762120d2e999 
* PR: https://github.com/influxdata/influxdb/pull/18257 

.


### Question 2:

We have the separate editor components per language (toml, markdown, flux), which is good. But then we have 2 different flux editors: FluxMonacoEditor versus NotificationMonacoEditor.

FluxMonacoEditor:
* hooked into LSP
* hooked into the autogrow handling 
* used almost everywhere

NotificationMonacoEditor:
* not hooked into the LSP
* not hooked into the autogrow handing 
* provides a slightly different set of formatting params to the react-monaco-editor child component.
* used only for the `src/flows/pipes/Notification`

Is there a historical reason for the NotificationMonacoEditor to be different?

wiedld:
Difference in bundle size:

before 415kb transferred size:
Screen Shot 2022-04-01 at 9 05 27 PM

after 746kb transferred size:
Screen Shot 2022-04-01 at 9 05 59 PM

src/flows/pipes/Notification/NotificationMonacoEditor.tsx
```diff
@@ -59,7 +60,10 @@ const NotificationMonacoEditor: FC<Props> = ({
           scrollBeyondLastLine: false,
           readOnly: !!readOnly,
         }}
-        editorDidMount={editorDidMount}
+        editorDidMount={ed => {
+          pointToCorrectWorkers()
+          editorDidMount(ed)
```
wiedld, line 65: Always reference the same editor worker instance:
    * can get multiple instances with ESM loading (webpack plugin)
    * plus we have 2 stages of webpack builds (`webpack.vendor.ts`, and `webpack.common.ts`) --> but at runtime want to have only 1 instance.
    * simple solution: both resolve to `filename: '[name].worker.js'` on build
    * HERE it's referencing the correct, single instance workers

src/languageSupport/languages/flux/monaco.flux.syntax.ts
```diff
@@ -1,4 +1,9 @@
 import register from 'src/languageSupport/monaco.onigasm'
+import * as monaco from 'monaco-editor/esm/vs/editor/editor.api'
+
+if (!window.monaco) {
+  window.monaco = monaco
+}
```
wiedld, line 6: There is longer a global monaco instance on window.monaco.
https://github.com/TypeFox/monaco-languageclient/pull/248#issuecomment-774026638

src/languageSupport/monaco.onigasm.ts
```diff
@@ -95,6 +96,9 @@ async function loader() {
 }
 
 export default async function register(scope, definition) {
+  if (!window.monaco) {
+    window.monaco = monaco
+  }
```
wiedld, line 101: Sometimes this gets loaded first. So set here as well.

webpack.common.ts
```diff
@@ -46,6 +52,17 @@ module.exports = {
...
+        test: /\.m?js$/,
+        include: MONACO_DIR,
+        use: {
+          loader: 'babel-loader',
+          options: {
+            presets: ['@babel/preset-env'],
+            plugins: ['@babel/plugin-transform-runtime'],
+          },
+        },
+      },
```
wiedld, line 65: Webpack plugin now uses ESM loading of the monaco-editor. This requires a separate loading rules if encountered errors (which we did).

webpack.common.ts
```diff
@@ -147,52 +164,8 @@ module.exports = {
...
-        '!toggleTabFocusMode',
-        '!transpose',
-        '!unusualLineTerminators',
-        '!viewportSemanticTokens',
-        '!wordHighlighter',
-        '!wordOperations',
-        '!wordPartOperations',
-      ],
+      languages: ['json', 'markdown'],
+      filename: '[name].worker.js',
```
wiedld, line 168: Same filename output, and same language settings, for both `webpack.common.ts` and `webpack.vendor.ts`. In the end, we'll still end up with only a single copy -- across the two, separate webpack build steps.

webpack.common.ts
```diff
@@ -46,6 +52,17 @@ module.exports = {
...
   module: {
     rules: [
+      {
+        test: /\.m?js$/,
+        include: MONACO_DIR,
+        use: {
+          loader: 'babel-loader',
+          options: {
+            presets: ['@babel/preset-env'],
+            plugins: ['@babel/plugin-transform-runtime'],
```
wiedld, line 62: Webpack plugin now uses ESM loading of the monaco-editor. This requires separate loading rules if encountered errors (which we did).

webpack.common.ts
```diff
@@ -147,52 +164,8 @@ module.exports = {
...
-        '!toggleTabFocusMode',
-        '!transpose',
-        '!unusualLineTerminators',
-        '!viewportSemanticTokens',
-        '!wordHighlighter',
-        '!wordOperations',
-        '!wordPartOperations',
-      ],
+      languages: ['json', 'markdown'],
+      filename: '[name].worker.bundle.js',
```
wiedld, line 168: Same filename output, and same language settings, for both `webpack.common.ts` and `webpack.vendor.ts`. In the end, we'll still end up with only a single copy -- across the two, separate webpack build steps.

note our build step, with the 2 webpack builds:
`yarn generate && yarn build:vendor && webpack --config webpack.prod.ts`

src/languageSupport/languages/flux/monaco.flux.syntax.ts
```diff
@@ -1,4 +1,9 @@
 import register from 'src/languageSupport/monaco.onigasm'
+import * as monaco from 'monaco-editor/esm/vs/editor/editor.api'
+
+if (!self.monaco) {
+  self.monaco = monaco
+}
```
wiedld, line 6: There is longer a global monaco instance on window.monaco.
https://github.com/TypeFox/monaco-languageclient/pull/248#issuecomment-774026638

I choose to set it explicitly, since we are already utilizing this global object across the application. The global instance is set per language syntax file (toml, markdown, flux), as that file usually gets loaded first.

wiedld: Also note that web workers use `self`, not window. So have to change global used.

cypress/support/commands.ts
```diff
@@ -70,12 +70,12 @@ export const loginViaDexUI = (username: string, password: string) => {
...
   {prevSubject: true},
-  (subject, text: string) => {
+  (subject, text: string, force = true, delay = 10) => {
     return cy.wrap(subject).within(() => {
       cy.get('.monaco-editor .view-line:last')
         .click({force: true})
         .focused()
-        .type(text, {force: true, delay: 10})
+        .type(text, {force, delay})
     })
```
wiedld, line 79: Need the ability to type slower, and without force, when the specific test relies on monaco editor autoComplete.

src/languageSupport/languages/flux/monaco.flux.syntax.ts
```diff
@@ -12,7 +13,10 @@ async function addSyntax() {
     }),
   }))
 
-  window.monaco.languages.setLanguageConfiguration(LANGID, {
+  if (!self.monaco) {
+    self.monaco = monaco
+  }
+  self.monaco.languages.setLanguageConfiguration(LANGID, {
```
wiedld, line 19: window.monaco global object is no longer present.
Now use the self.monaco, for the web workers.

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -167,7 +167,7 @@ describe('Dashboard', () => {
     const headerPrefix2 = '-'
 
     cy.getByTestID('note-editor--overlay').monacoType(
-      `${headerPrefix2} ${noteText2}`
+      `\n${headerPrefix2} ${noteText2}`
```
wiedld, line 170: the `- Text` has to be on a newline, in order to trigger the markdown conversion to a list.

tests were failing due to the absence of this newline (paired with the editor version bump):
https://app.circleci.com/pipelines/github/influxdata/monitor-ci/10633/workflows/1f173ccd-f236-499e-bc1c-9e8848004d22/jobs/558131/tests#failed-test-0

webpack.common.ts
wiedld: Per discussion with @drdelambre , at the end of the final debugging (tests passing) I tested that these webpack updates are required. They are.

 Error without this rule:
```
ERROR in ./node_modules/monaco-editor/esm/vs/language/css/monaco.contribution.js 29:15
Module parse failed: Unexpected token (29:15)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| // src/language/css/monaco.contribution.ts
| var LanguageServiceDefaultsImpl = class {
>   _onDidChange = new monaco_editor_core_exports.Emitter();
|   _options;
|   _modeConfiguration;
 @ ./node_modules/monaco-editor/esm/vs/editor/editor.main.js 2:0-45
 @ include-loader!./node_modules/monaco-editor/esm/vs/editor/editor.main.js
 @ dll vendor
```

Error without the plugin '@babel/plugin-transform-runtime':
Screen Shot 2022-04-06 at 5 05 50 PM

src/languageSupport/monaco.utils.ts
```diff
@@ -0,0 +1,17 @@
...
+    }
+    if (label === 'html' || label === 'handlebars' || label === 'razor') {
+      return './html.worker.bundle.js'
+    }
+    if (label === 'typescript' || label === 'javascript') {
+      return './ts.worker.bundle.js'
+    }
+    return './editor.worker.bundle.js'
+  }
+}
```
wiedld, line 17: per
https://github.com/microsoft/monaco-editor/blob/main/samples/browser-esm-webpack-typescript-react/src/components/Editor.tsx

src/languageSupport/monaco.utils.ts
```diff
@@ -0,0 +1,17 @@
...
+    if (label === 'css' || label === 'scss' || label === 'less') {
+      return './css.worker.bundle.js'
+    }
+    if (label === 'html' || label === 'handlebars' || label === 'razor') {
+      return './html.worker.bundle.js'
+    }
+    if (label === 'typescript' || label === 'javascript') {
+      return './ts.worker.bundle.js'
+    }
+    return './editor.worker.bundle.js'
```
drdelambre, line 15: this needs a cache busting mechanism attached to it

drdelambre:
> * of the simple wireTmGrammars() interface

turns out it wasn't that simple to achieve what design required once we started displaying tomls next to flux editors next to markdown editors in the ui, each with their own styling requirements. it's been a couple years, but iirc the example for wireTmGrammars didn't work properly and adding flux support turned into... well what you see today. in the linked PR, you'll see the themeing support issue referenced in the monaco repo that caused havoc when trying to allow easy theme support for languages and early explorations of mitigating that by utilizing the suggested path of normalizing the built in tokenizations to some kind of shared proto language created a crazy explosion of relationships between keywords in different contexts to different themes.

cypress/support/commands.ts
drdelambre: checked with @TCL735 as I remember him ripping something similar to these changes out recently as he chased down some flaky test patterns. he was like "totes cool yolo :shipit:" but reserved the right to be upset if it reintroduced issues. so... :+1:

global.d.ts
```diff
@@ -1,17 +1,19 @@
...
 //
 
 declare let monaco: MonacoType
+declare let MonacoEnvironment: MonacoEnvironmentType
 
 declare global {
   let Z: ZuoraClient
   let _abcr: AdMinds
   interface Window {
     monaco: MonacoType
```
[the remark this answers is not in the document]
wiedld: typescript searches for `self.*` on `Window & typeof globalThis`.

src/languageSupport/monaco.utils.ts
wiedld: These files are from monaco-editor, pegged to a minor version (only patches allowed) -- because compatibility with the associated libs are dependent on minor versions.

Does this need to be cache busted?

wiedld:
> iirc the example for wireTmGrammars didn't work properly and adding flux support turned into... well what you see today. in the linked PR, you'll see the themeing support issue referenced in the monaco repo that caused havoc when trying to allow easy theme support for languages...

Thank you for the detailed explanation. I looked at the linked monaco-editor issue, and it appears it's still unresolved. They have made a roadmap decision to not support different themes per language, and the solutions are either: (1) the tokenizer or (2) loading different css styles per language.

cypress/support/commands.ts
wiedld: Yeah...I have worries. This pattern is used twice to test autocomplete functionality. In this particular test example (for pinned items), it has not been flaky in the branch tests.

However, for the change in `cypress/e2e/shared/explorer.test.ts`, it has been a tad flaky in OSS tests. I just extended the time from 30ms to 50ms delay between char typing @TCL735 . Hopefully that one will now behave.

TCL735: @wiedld if you want to change the `monacoType` command, then the new way shouldn't involve some arbitrary amount of delay. The changes should look something like this: https://github.com/influxdata/ui/pull/3997/files


```
Cypress.Commands.add(
  'monacoType',
  {prevSubject: true},
  (subject, text: string) => {
    return cy.wrap(subject).within(() => {
      cy.get('.monaco-editor .view-line:last').should('be.visible')
      cy.get('.monaco-editor .view-line:last').click()
      cy.get('.monaco-editor .view-line:last').type(text)
    })
  }
)
```

TCL735: I had previously refrained from merging the above mentioned PR because i didn't want to introduce so many changes to typing until we were certain that e2e tests were stable again.

e2e tests are generally stable now. But i had forgotten about this PR.

wiedld:
> Did you test this on a production build? i can't see how the `./*.worker.bundle.js` mechanism would work, but i might be lacking imagination.

Yes:
* The CI tests run on the prod build (refer to `build-image` job in monitor-ci). 
* I've also been using the same prod build (`quay.io/influxdb/ui-acceptance`) built in the CI pipeline, when deploying my remocal cluster. If you go to my currently deployed cluster right now, you can inspect the remocal deployment with the latest ui-acceptance image.
https://twodotoh-dev-denisewiedl20220405130506.remocal.influxdev.co

.

> please verify that the worker is engaged in a production build and that we're not regressing to the fall back mechanism.

So here is where we run into the difference in how we are connecting to the worker for flux (using `monaco-languageclient` bridge), versus the toml and markdown editors. In all cases, using the production build image (`quay.io/influxdb/ui-acceptance`) the workers are present and registered in the browser.

Screen Shot 2022-04-07 at 3 54 56 PM

However, the worker is only engaged for the non-flux editors. 
(Below is a toml editor.)
Screen Shot 2022-04-07 at 3 55 13 PM

What I did not address in this PR (to be addressed in the next one), is how we tie in the LSP to the monaco-editor over the `monaco-languageclient` bridge:
* Looking at the code, it looks like we are running our own custom handling (our own LSP server) instead of the monaco-editor worker.
* This is (I suspect) why we are missing much of the new flux LSP functionality. Because we are overwriting the monaco-editor feature list to call out to our own LSP server (in `src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts`).
* This is the code pattern used for custom handling (or custom workers), instead of using the monaco-editor's worker. 

I can fix the LSP integration in this PR also. But I've been cognizant of requests to keep PRs smaller, so I'm trying to move that into the next PR. (Goal of this PR was version bumping, whilst maintaining the same functionality we currently have.)

.

> also you mentioned the size of the bundle changing, but from a user experience perspective, what does that do to an already frustratingly slow load time of the editor when we lazy load it in for the first time?

I am load testing now with the remocal deployment + ui-acceptance image. Will update this comment with results when finished.

cypress/support/commands.ts
wiedld: Had an offline convo about how the tests are using the delay for use of the autocomplete feature of the monaco-editor. The use of the `{rightarrow}` works pretty well with autocomplete and a delayed typing.

However, the explorer test uses an editor cursor pause (to inject a variable) with the autocomplete closing paran. And that is more sensitive. Will change the approach on this one.

webpack.common.ts
```diff
@@ -30,6 +32,10 @@ module.exports = {
       src: path.resolve(__dirname, 'src'),
       assets: path.resolve(__dirname, 'assets'),
       react: path.resolve('./node_modules/react'),
+      vscode: path.resolve(
```
drdelambre, line 35: what does this one do?

webpack.common.ts
drdelambre: k one more time.  now that we have removed the random worker that was being loaded and the only assets going through have  been verified as being compiled, meaning there's no need  for the runtime as it's baked in.... is this still needed?

webpack.common.ts
wiedld: Was a requirement of `monaco-languageclient` in order to set the correct vscode resolution. Their `0.19.0-next.1` version (which we are not using yet) explicitly does this for us. But earlier versions (like the 0.18.x that we are using) requires us to define this resolution in our own webpack config.

Testcase: the build had been failing without it as recently as `chore: final code cleanup` commit (when I confirmed all webpack changes were still required, by incrementing them on/off).

So it's required...
...until we switched to the `globalAPI` option today. Now everything appears to build.

wiedld: error if not present:
```
ERROR in ./node_modules/vscode-languageclient/lib/common/typeDefinition.js
Module not found: Error: Can't resolve 'vscode' in '/Users/denisewiedl/influxdata/ui/node_modules/vscode-languageclient/lib/common'
 @ ./node_modules/vscode-languageclient/lib/common/typeDefinition.js 8:17-34
 @ ./node_modules/monaco-languageclient/lib/monaco-language-client.js
 @ ./node_modules/monaco-languageclient/lib/index.js
 @ dll vendor
```

drdelambre (approved):
please double check the babel additions after the learnings and changes from yesterday. it'd be best to remove any additional requirements if they're not really needed, but not a blocker if they have to stay

webpack.common.ts
wiedld: yes. Same error.

Still required for loading ESM modules for the bumped monaco-editor.

PR #4320, opened 2022-04-05
feat: start/stop functionality subscriptions

helenosheaa:
Closes https://github.com/influxdata/data-acquisition/issues/203

Enables the user to start and stop their subscription.

Screen Shot 2022-04-05 at 2 09 57 PM
Screen Shot 2022-04-05 at 2 10 13 PM

### DIFF
--- a/src/writeData/subscriptions/context/subscription.update.tsx
+++ b/src/writeData/subscriptions/context/subscription.update.tsx
@@ -1,6 +1,10 @@
 // Libraries
 import React, {FC, useState, useCallback, useContext, useEffect} from 'react'
-import {updateAPI} from 'src/writeData/subscriptions/context/api'
+import {
+  updateAPI,
+  updateStatusAPI,
+  getByIDAPI,
+} from 'src/writeData/subscriptions/context/api'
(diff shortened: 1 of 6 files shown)


src/writeData/subscriptions/context/subscription.update.tsx
```diff
@@ -144,6 +153,27 @@ export const SubscriptionUpdateProvider: FC = ({children}) => {
...
 
+  const setStatus = (isActive: boolean): any => {
+    setLoading(RemoteDataState.Loading)
+    const params = {id: currentID, data: {isActive}}
+    updateStatusAPI(params)
+      .then(() => {
+        getSubscription()
+      })
+      .catch(() => {
+        dispatch(notify(subscriptionStatusUpdateFail()))
```
genehynson, line 164: I think you need to `setLoading(RemoteDataState.Done)` here so the spinner doesn't live forever if there is an error

helenosheaa: good point, i think there's other places where i was doing that too, so fixed in latest commit

genehynson (approved):

PR #4321, opened 2022-04-05
fix: atomic query builder actions

drdelambre:
Got some feedback on the "always select a measurement first" product change made in #4286. measurement selection should mirror bucket selection for a product perspective in that it should clear the query and start over again when you select a new one.

this involved some clean up on bucket selection as well. we used some useEffects to sniff on changes to the data.buckets property to see if the bucket had changed, and thus we should perform some action on the data. Now that we wanted that action to also be connected to a measurement, but only if there was a flag on, the useEffect got a bit... tangled. In order to make the code a bit more straight forward, the useEffect logic for clearing a query when a bucket gets selected got moved to it's own function on the context. once this proved a bit more forward, since we wanted a product change that mirrored this behavior, the technique was mirrored for measurement selection. this mechanism is a bit more rough because, for backwards compatibility reasons, we need to say that a measurement is the first card, and only if deemed 'compliant'.

### DIFF
--- a/src/flows/pipes/QueryBuilder/AddButton.tsx
+++ b/src/flows/pipes/QueryBuilder/AddButton.tsx
@@ -4,6 +4,7 @@ import {SquareButton, IconFont} from '@influxdata/clockface'
 import {RemoteDataState} from 'src/types'
 
 import {QueryBuilderContext} from 'src/flows/pipes/QueryBuilder/context'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 import {event} from 'src/cloud/utils/reporting'
 
@@ -16,6 +17,14 @@ const AddButton: FC = () => {
   }, [add])
 
   if (!cards.length) {
+    return null
+  }
+
+  if (
+    isFlagEnabled('newQueryBuilder') &&
+    !cards[0]?.values?.selected?.length &&
+    cards.length < 2
+  ) {
     return null
   }
 
--- a/src/flows/pipes/QueryBuilder/BucketSelector.tsx
+++ b/src/flows/pipes/QueryBuilder/BucketSelector.tsx
@@ -5,25 +5,18 @@ import React, {FC, useState, useContext, useEffect} from 'react'
 import {Input, ComponentSize, List, Gradients} from '@influxdata/clockface'
 import {Bucket, RemoteDataState} from 'src/types'
 import BuilderCard from 'src/timeMachine/components/builderCard/BuilderCard'
+import {QueryBuilderContext} from 'src/flows/pipes/QueryBuilder/context'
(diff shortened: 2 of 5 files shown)


src/flows/pipes/QueryBuilder/AddButton.tsx
```diff
@@ -19,6 +20,14 @@ const AddButton: FC = () => {
     return null
   }
 
+  if (
+    isFlagEnabled('newQueryBuilder') &&
```
asalem1, line 24: it's like the old one, but newer

src/flows/pipes/QueryBuilder/BucketSelector.tsx
```diff
@@ -67,7 +38,14 @@ const BucketSelector: FC = () => {
       </BuilderCard>
     )
   }
-
+  const _selectBucket = (bucket?: string) => {
+    if (bucket) {
+      event('Query Builder Bucket Selected', {
+        usedSearch: !!search.length ? 'yus' : 'nah',
```
asalem1, line 44:
```suggestion
        usedSearch: !!search.length,
``` 
I've got a sense of humor somewhere, but shouldn't this just be fine?

src/flows/pipes/QueryBuilder/CardList.tsx
```diff
@@ -125,23 +131,27 @@ const Card: FC<Props> = ({idx}) => {
...
+      if (index === -1) {
+        event('Query Builder Value Selected')
+        selectMeasurement(val)
       } else {
-        _vals.push(val)
+        event('Query Builder Value Unselected')
+        selectMeasurement(null)
       }
+
+      return
```
asalem1, line 146: confused on this early return. Shouldn't we update the values for this idx anyways to clear it out? Or is that handled elsewhere?

asalem1 (approved):
Some questions but nothing sus.

src/flows/pipes/QueryBuilder/AddButton.tsx
drdelambre: wait for the `newerQueryBuilder` flag

src/flows/pipes/QueryBuilder/BucketSelector.tsx
drdelambre: gotta be a string, and stop being boring

src/flows/pipes/QueryBuilder/CardList.tsx
drdelambre: yeah, so all of the stuff done below is moved out into the context function. but the context  function is only used for  the  _measurement card

asalem1: Cool

Issue #3866, opened 2022-02-15
Filter field endlessly displays "...Loading" in FILTER box (Query Builder in a Notebook)


gefaila:
**Steps to reproduce:**
I have a Notebook that starts with a Query Builder.
I filter by a handful of _fields that I've selected, followed by a table and some charts.
Whenever I load the Notebook it runs correctly.

*HOWEVER* the UI refuses to display the list of _fields that were originally selected in the *FILTER* box of the Query Builder

As shown below (note the clock bottom right)
Note that the Notebook has already run and returned the correct data!
![image](https://user-images.githubusercontent.com/69196723/154025673-3fe07acd-90ce-40ca-b38b-36406775bb4c.png)
 
2 minutes later it has still not stopped "....Loading" the tags
![image](https://user-images.githubusercontent.com/69196723/154025849-3b6847f6-90fd-4696-b910-eb1d46f8a6f2.png)

After refreshing the page it still displays "...Loading".
![image](https://user-images.githubusercontent.com/69196723/154026129-ab4e277c-085e-4af9-8233-c528e6084749.png)

Only by typing something in the box does it now update and display the original tags *AND* the tags it's found due to what was typed.
![image](https://user-images.githubusercontent.com/69196723/154026769-1fe26191-3e1f-4cf1-bf14-dfaac8810f74.png)

**Expected behavior:**
All filter boxes should be populated when the Notebook is opened / run.

**Config:**
https://eu-central-1-1.aws.cloud2.influxdata.com/orgs/faac9e69ed04ee08/notebooks/5fa821932b640000

asalem1:
Hello @gefaila is this still an issue? I can't seem to reproduce it. I'm wondering if this may have been resolved by some previous work that's been done to enhance the user query experience in notebooks.

asalem1:
@gefaila I just merged in what _might_ be a fix (the reason I say maybe is that it's been super hard to track down the issue and duplicate it to some degree of consistency). The new UI image is going to be deployed later today, please let me know if you're still seeing the issue. I'm going to leave the issue open until we can confirm its gone for good

jzahnd:
@gefaila any feedback on the update above? We are happy to follow up but want to check in on the latest updates.

gefaila:
It's still happening,
Just hadn't had time to follow up

Cheers

Andrew

gefaila:
I see this when I go to a saved notebook.
[...]

PR #4324, opened 2022-04-05
fix: potentially fixes the loading state forever bug

asalem1:
Related #3866

I'm not confident that we've seen the last of this bug since it's a difficult one to reproduce locally, but it is reproducible. I _think_ that the crux of the issue here is that we're referencing stale cardMeta when making these state updates, and so we might be setting stale data and never getting past the point of loading as a result of that. By removing the useCallback here, I'm hoping that we can resolve the stale state issue whereby we always reference the state at that point and aren't necessarily relying upon cached data. I was able to inconsistently reproduce the bug linked here in the query builder when I would select a bunch of measurements from a bucket and start adding and removing cards from the list, then refreshing the page. I would be able to reproduce this 1/10 times. After making these changes I did the same operation about 20 times and wasn't able to reproduce it so 🤷🏽‍♂️

### DIFF
--- a/src/flows/pipes/QueryBuilder/context.tsx
+++ b/src/flows/pipes/QueryBuilder/context.tsx
@@ -4,7 +4,6 @@ import React, {
   createContext,
   useContext,
   useState,
-  useCallback,
   useMemo,
   useEffect,
 } from 'react'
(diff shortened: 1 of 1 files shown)


drdelambre (approved):
what a weird diff, but seems to check out

PR #4326, opened 2022-04-06
chore: remove the ephemeral notebooks and publish lifecycle feature flags

asalem1:
Now that the publish lifecycle work has been on in the UI for a while, we want to make the break with ephemeral notebooks official by remove the flags for ephemeral notebooks and the publish lifecycle.


drdelambre (approved):
i forget all the weird additional stuff we needed to make ephemeral work, but this is definitely the brunt of it. thanks for putting it to bed

Issue #4335, opened 2022-04-07
Bug: Datepicker format changes to UTC when user selects a date from the calendar


ChitlangeSahas:
https://user-images.githubusercontent.com/18511823/162236077-ddafe7bc-1f65-4d03-816c-a6d82f89c774.mov

The time format should remain the default when the user changes the calendar selection.

Issue #2350, opened 2021-08-12
Bug Duty: Integration tests for Date Picker


ChitlangeSahas:
Now that we have the datepicker streamlined a little bit more, integration tests for it would be a nice addition.

PR #4336, opened 2022-04-07
fix: datepicker time format on input should stay the same when calendar selection is made

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4335
Connects https://github.com/influxdata/ui/issues/2350

* the `overrideInputState` is not necessary and was causing console log errors so I removed it
* made the default `inputFormat` state of the component to be `YYYY-MM-DD HH:mm:ss`
* added two integration tests

https://user-images.githubusercontent.com/18511823/162237629-349fb188-9c43-4910-beef-f91df0503954.mov

### DIFF
--- a/src/shared/components/dateRangePicker/DatePicker.tsx
+++ b/src/shared/components/dateRangePicker/DatePicker.tsx
@@ -77,7 +77,7 @@ class DatePicker extends PureComponent<Props, State> {
   private inCurrentMonth: boolean = false
   state = {
     inputValue: null,
-    inputFormat: 'YYYY-MM-DD HH:mm',
+    inputFormat: 'YYYY-MM-DD HH:mm:ss',
   }
 
   public componentDidUpdate() {
--- a/src/shared/components/dateRangePicker/DatePicker.tsx
+++ b/src/shared/components/dateRangePicker/DatePicker.tsx
@@ -77,7 +77,7 @@ class DatePicker extends PureComponent<Props, State> {
   private inCurrentMonth: boolean = false
   state = {
     inputValue: null,
-    inputFormat: null,
+    inputFormat: 'YYYY-MM-DD HH:mm',
   }
 
   public componentDidUpdate() {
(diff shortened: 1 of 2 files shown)


src/shared/components/dateRangePicker/DatePicker.tsx
```diff
@@ -77,7 +77,7 @@ class DatePicker extends PureComponent<Props, State> {
   private inCurrentMonth: boolean = false
   state = {
     inputValue: null,
-    inputFormat: null,
+    inputFormat: 'YYYY-MM-DD HH:mm:ss',
```
hoorayimhelping, line 80: tip for the future: before using a string like a dateformat, or instuctrions, do a global search for it. I found [this](https://github.com/influxdata/ui/blob/master/src/utils/datetime/constants.ts#L1), which should probably be used here 👍

src/shared/components/dateRangePicker/DateRangePicker.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {pastHourTimeRange} from 'src/shared/constants/timeRanges'
+import {renderWithReduxAndRouter} from 'src/mockState'
+
+import {screen} from '@testing-library/dom'
+import {fireEvent} from '@testing-library/react'
+
+jest.useFakeTimers().setSystemTime(new Date('2022-03-10 20:00:00').getTime())
+
+describe('Date Range Picker', function() {
+  test('should have input values in the correct format after user changes the date selection', () => {
```
hoorayimhelping, line 14: if you use `it` instead of `test`, it's in theory, supposed to read like a sentence

```ts
it('should have input values in the correct format after user changes the date selection')
```

is meant to be read like

>it should have input values in the correct format after user changes the date selection

hoorayimhelping (approved):
other than using a hardcoded string when a global variable exists for it, lgtm!

good work getting that test up

PR #4352, opened 2022-04-08
feat: add save as notebook option from data explorer

asalem1:
This work stems from frustration trying to debug the differences between the way data explorer handles the gauge and histogram visualizations vs notebooks. I figured if I could save the data correctly, then things should work as a like-for-like replacement. 
save_as

### DIFF
--- a/src/dataExplorer/components/SaveAsNotebookForm.tsx
+++ b/src/dataExplorer/components/SaveAsNotebookForm.tsx
@@ -75,14 +75,13 @@ const SaveAsNotebookForm: FC<Props> = ({dismiss}) => {
       for (let i = 0; i < draftQueries.length; i++) {
         const draftQuery = draftQueries[i]
         let pipe: any = {
-          family: 'inputs',
           id: `local_${nanoid()}`,
           visible: !draftQuery.hidden,
         }
         if (draftQuery.editMode === 'builder') {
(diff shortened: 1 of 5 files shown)


src/dataExplorer/components/SaveAsNotebookForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+                name: draftQuery.builderConfig.buckets[0],
+                orgID,
+                type: 'user',
+              },
+            ],
+          }
+        } else {
+          pipe.title = 'Query to Run'
+          pipe.queries = properties.queries
+          pipe.activeQuery = i
```
drdelambre, line 90: idk how well this is supported in notebooks. it's _kind of_ in there, but we never exposed the interface to switch between queries so there prolly haven't been a lot of eyes on it for a bit. might be safer to always run `activeQuery = 0; queries = properties.queries[i]`

asalem1: sweet. Good to know

drdelambre (approved):
cool cool cool. kinda breaks the notebooks format, but idk how that "import all the queries" thing is gonna work, so lets keep it loose and save the choice on which direction to go for later when we've clicked around a bit

PR #4357, opened 2022-04-08
chore: add guardrails in case there arent fields

asalem1:
There's an assumption that fields will exist in results from a query. This adds guardrails against that. We should iterate on this so that we can make sure the error threshold work is still applicable after this gets in. For context, check out the conversation around the issue [here](https://influxdata.slack.com/archives/CLHATC50E/p1649348887581009)

### DIFF
--- a/src/flows/pipes/Visualization/ErrorThresholds/FieldColumnDropdown.tsx
+++ b/src/flows/pipes/Visualization/ErrorThresholds/FieldColumnDropdown.tsx
@@ -37,8 +37,9 @@ const FieldColumnDropdown: FC<Props> = ({threshold, index}) => {
         return c.name === '_value'
       })
 
-      const fields = (results.parsed.table.columns['_field']
-        .data as any).reduce((acc, curr, index) => {
+      const fields = (
+        (results.parsed.table.columns['_field'].data as any) ?? []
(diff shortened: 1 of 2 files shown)


wiedld (approved):

src/flows/pipes/Visualization/ErrorThresholds/FieldColumnDropdown.tsx
```diff
@@ -37,8 +37,9 @@ const FieldColumnDropdown: FC<Props> = ({threshold, index}) => {
         return c.name === '_value'
       })
 
-      const fields = (results.parsed.table.columns['_field']
-        .data as any).reduce((acc, curr, index) => {
+      const fields = (
+        (results.parsed.table.columns['_field'].data as any) ?? []
```
subirjolly, line 41: shouldn't it be an Array<any>?

asalem1: it should, but a base arrays without a type don't make TS happy, and declaring `string[]` would be a lie because it could be undefined. At this point `any` suffices

subirjolly (approved):

Issue #4356, opened 2022-04-08
Alert error messages are inaccessible in the UI


genehynson:
Currently our alert error messages are inaccessible in the UI unless you perform some hacky maneuvers. However, I think it's an easy fix to resolve this issue.

To reproduce:
1. create an alert with a message that references a field/column that does NOT exist
2. save the alert and observe that it fails
3. observe that while the alert is in a failure state, there is no way to collect information on WHY the alert is failing

Notes:
I noticed that you can see the logs of the alert's associated task to get the error message by copying the task ID and navigating to `/tasks/{id}/runs`. However, this task ID does not show up in the list of tasks on the Tasks page (intentionally). This is because we do not want users to be able to edit the task associated to the alert because they could break it. However, there is no harm in them seeing the logs.

In my mind this is a critical issue because if users run into an issue while creating an Alert, they have no way to debug the issue further. When I experienced this bug, my Alert was failing because I had `null` values in my query's results. I had absolutely no idea what was causing the failure and it took me over an hour to figure it out - extremely frustrating!

Proposed fix:
1. Clicking the red triangle on the failed alert should take you to `/tasks/{taskID}/runs` page in the UI
2. The `/tasks/{taskID}/runs` page should only show the "Edit" button IF the provided task ID exists in the response from /api/v2/tasks

![image](https://user-images.githubusercontent.com/6411855/162487914-49beb383-be2a-44d5-87bb-c2adc174b541.png)

genehynson:
Enable `navToTaskRuns` flag to see my proposed solution

wiedld:
This is currently on in prod. Should we remove this flag (and make it the norm)? Or leave as-is and close issue?

PR #4359, opened 2022-04-08
fix(alerts): navigate to task runs

genehynson:
Helps https://github.com/influxdata/ui/issues/4356

Ever had a failed alert and wanted to know why? Well today's your lucky day! Now you can click the warning symbol and be taken to the associated task's runs page and view the logs!

If the user navigates to the task runs page, we now confirm the task is editable before letting them edit it. We can confirm the task is editable by checking if it exists in the response from the `/tasks?name={task name}` API. Tasks created by Alerts are not returned in this API response. I used the `name` filter because the Task created by an Alert always shares the same name as the Alert. From this list, I set `isTaskEditable=true` iff I can find a task in the response with the matching task ID from the Alert.

All of this is behind flag `navToTasksRuns`

https://user-images.githubusercontent.com/6411855/162501555-3c91f852-ca19-4733-a17c-31d38c262a06.mov

### DIFF
--- a/src/checks/components/CheckCard.tsx
+++ b/src/checks/components/CheckCard.tsx
@@ -44,6 +44,8 @@ import {Check, Label} from 'src/types'
 // Utilities
 import {relativeTimestampFormatter} from 'src/shared/utils/relativeTimestampFormatter'
 import ErrorBoundary from 'src/shared/components/ErrorBoundary'
+import {event} from 'src/cloud/utils/reporting'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 interface OwnProps {
   check: Check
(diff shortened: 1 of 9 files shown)


asalem1 (approved):
I'll own this

Luckette:
FYI @kristinarobinson - Can we look at this during bug scrub?

kristinarobinson:
I think it would be better to show the LOG from the failed task, rather than the Task Run page.  As it stands, the user must click another button from the Task Run page (finding the failed task and clicking View Logs).  When an alert fails, please display the `Run Logs` modal of the Task Runs page, as found from this page using the failed task ID and  clicking on View Logs:

https://us-central1-1.gcp.cloud2.influxdata.com/orgs/<orgId>/tasks/<taskid>/runs 

Sample:
TaskRunLog

PR #4361, opened 2022-04-08
fix: naming issues for notebooks referencing stale state

### LINKED ISSUE #4360: Cannot rename notebooks
I noticed that I can no longer rename my notebooks (see video). 

https://user-images.githubusercontent.com/6411855/162526513-0b1cf725-48ab-4688-89a7-d631f6d04853.mov

asalem1:
Closes #4360 

This PR closes two bugs surrounding renaming a notebook. The first, when a user renaming a notebook in the list page wouldn't have their changes reflected immediately since the state wasn't setting correctly based on the user's input. The second one being that if a user renamed a notebook from the notebook view & immediately went off to perform another update (like updating a flux script in a panel), then the user's secondary patch (with an out of sync notebook name) would overwrite the original changes.

### DIFF
--- a/src/flows/components/FlowCard.tsx
+++ b/src/flows/components/FlowCard.tsx
@@ -45,7 +45,8 @@ const FlowCard: FC<Props> = ({id, isPinned}) => {
   )
 
   const handleRenameNotebook = async (name: string) => {
-    await update(id, name)
+    flow.name = name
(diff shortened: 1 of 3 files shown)


drdelambre (approved):
good luck pushing late on a friday!

PR #4369, opened 2022-04-11
chore: remove cloneToOtherBoards feature flag

asalem1:
This feature has been enabled for quite some time in the UI. It's time to rip out the feature flag now that we're certain that this is a feature we want to maintain


appletreeisyellow (approved):
🎉

PR #4382, opened 2022-04-12
feat(HelpBar): Display support overlays for payg and free accounts 

### LINKED ISSUE #3450: Help Bar: Display a PAYG overlay for when Contact Support is clicked
When clicked on Contact Support, check if the user is PAYG, then display a PAYG Overlay.

blegesse-git:
Closes #3450 

- adds `PayGSupportOverlay` and `FreeAccountOverlay` to overlay controller file. So when `Contact Support` is clicked, depending on the account type, we direct user to the correct support overlay.

### DIFF
--- a/src/pageLayout/containers/TreeNav.tsx
+++ b/src/pageLayout/containers/TreeNav.tsx
@@ -2,7 +2,7 @@
 import React, {FC, useContext, useEffect} from 'react'
 import {Link} from 'react-router-dom'
 import {useSelector} from 'react-redux'
-import {connect, ConnectedProps, useDispatch} from 'react-redux'
+import {connect, ConnectedProps} from 'react-redux'
 import {withRouter, RouteComponentProps} from 'react-router-dom'
 
 // Components
@@ -26,7 +26,6 @@ import {SafeBlankLink} from 'src/utils/SafeBlankLink'
 import {NavItem, NavSubItem} from 'src/pageLayout/constants/navigationHierarchy'
 import {AppState} from 'src/types'
 
-import {getStore} from 'src/store/configureStore'
 import {showOverlay, dismissOverlay} from 'src/overlays/actions/overlays'
 
 type ReduxProps = ConnectedProps<typeof connector>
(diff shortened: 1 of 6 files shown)


src/support/components/FreeAccountSupportOverlay.tsx
```diff
@@ -0,0 +1,35 @@
+import React, {FC, useContext} from 'react'
```
blegesse-git, line 1: I'm not too sure about the file structure here. Right now, I have both Overlays saved in their own folder; `src/support/components`...

src/pageLayout/containers/TreeNav.tsx
```diff
@@ -22,12 +24,27 @@ import {SafeBlankLink} from 'src/utils/SafeBlankLink'
...
+}) => {
   const {presentationMode, navbarMode, setNavbarMode} = useContext(
     AppSettingContext
   )
+  const userAccount = (state: AppState): boolean => {
+    const {quartzMe} = state.me
+    const accountType = quartzMe?.accountType ?? 'free'
+    return accountType !== 'free'
+  }
+  const isPayGCustomer = useSelector(userAccount)
```
hoorayimhelping, line 45: I think the naming and usage of this was a little bit confusing to me. It took me a second to realize that `userAccount` is a selector function. 

I think if we're going to take the time to write a selector, we should put it in a place where it's reusable. I'm not sure if this needs to be reusable yet though - there's a tension between the overhead of writing and importing a selector to be used in a single place and just using the data you already have access to in the component. I usually default to creating a selector if I have to access this data in multiple different components.

There are two options how to do this:
- if you want to keep it as a selector, put it in `me/selectors/index` and name it something that indicates it gets the payg status
- just set the `isPayGCustomer` (great variable name btw!) in this block of code without making a selector function

```ts
const accountType = state.me.quartzMe?.accountType ?? 'free'
const isPayGCustomer = accountType !== 'free'
```

also since there are subtle differences between `??` and `||`, if you haven't already, make sure that `??` is the operator you want, rather than `||` :)

hoorayimhelping (approved):
lgtm! just one comment about using a selector

src/pageLayout/containers/TreeNav.tsx
blegesse-git: hmm.. looking back at it, I can see how `userAccount` can be a bit confusing to understand. I removed the selector function code. Setting the `PayGCustomer` like you mentioned is a much easier route.

PR #4389, opened 2022-04-13
chore(monitor-ci-345): remove bundle analyzer from prod build

wiedld:
Part of https://github.com/influxdata/monitor-ci/issues/345


hoorayimhelping (approved):
This works for me. I don't think we need to include this in the prod build, since we can use it as a [cli tool](https://www.npmjs.com/package/webpack-bundle-analyzer#user-content-usage-as-a-cli-utility) without shipping it to prod on every build. I've used it occasionally in that manner, so I think we should keep the dependency around.

drdelambre:
who is this for?

wiedld:
> who is this for?

See second comment here. Is a direct request:
https://github.com/influxdata/monitor-ci/issues/345#issuecomment-1048297575

PR #4390, opened 2022-04-13
feat: subscriptions single page

helenosheaa:
Helps with https://github.com/influxdata/data-acquisition/issues/240

This adds a single page view for subscriptions/details view behind the feature flag: subscriptionsSinglePage which is default to false. This allows us to get feedback on two different designs for edit/details view.

Video:

https://user-images.githubusercontent.com/38860767/163249315-9084ed62-da69-4139-9d9c-b549413b21f1.mov


src/writeData/subscriptions/components/BrokerDetails.tsx
```diff
@@ -66,44 +157,60 @@ const BrokerDetails: FC<Props> = ({
...
-              color={ComponentColor.Success}
-              onClick={() => {
-                history.push(`/orgs/${org.id}/notebooks`)
-              }}
-              type={ButtonType.Button}
-              testID="update-broker-form--view-data"
-              status={ComponentStatus.Default}
-            />
-          </Overlay.Footer>
+          {!singlePage ? (
```
drdelambre, line 160: at this point would it be better to have two separate components?

src/writeData/subscriptions/components/JsonParsingForm.tsx
```diff
@@ -141,7 +143,7 @@ const JsonParsingForm: FC<Props> = ({formContent, updateForm}) => {
                   )
                 }
                 status={status}
-                maxLength={16}
+                maxLength={56}
```
genehynson, line 146: just curious, why was 56 chars chosen?

genehynson (commented):
Good stuff. I agree with Boatie - I think there might be a couple places where it would make sense to isolate the single page / non-single page stuff into their own components to help with readability. I don't expect both UI's (single page / multi page) to stay in the product for long tho so it doesn't need to be perfect

src/writeData/subscriptions/components/SubscriptionDetails.tsx
```diff
@@ -70,56 +81,61 @@ const SubscriptionDetails: FC<Props> = ({
...
+          {!singlePage ? (
+            <Overlay.Footer>
+              <Button
+                text="Close"
+                color={ComponentColor.Tertiary}
+                onClick={() => {
+                  event('close button clicked', {}, {feature: 'subscriptions'})
+                  history.push(`/orgs/${org.id}/${LOAD_DATA}/${SUBSCRIPTIONS}`)
+                }}
+                titleText="Cancel update and return to Subscriptions list"
```
genehynson, line 94: I think this footer could be put into a component, maybe? It looks like the code is duplicated in each of the pages right now and is wrapped with the `{!singlePage ? ...}` flag

src/writeData/subscriptions/components/JsonParsingForm.tsx
helenosheaa: this was already in the string parsing one so i was just making it longer based on what gary said this morning but open to suggestions

src/writeData/subscriptions/components/SubscriptionDetails.tsx
helenosheaa: yeah i could make it a single component where the parsing info for what to navigate to next is handed in

helenosheaa: ok i've pulled it out, ended up being more faffy as the parsing form has different buttons as well.

helenosheaa:
ok cool, thanks for the suggestions. I split out the broker one and then pulled out the footer buttons into a single component. I ended up having to change the tests as it's one and the same component for the buttons.

src/writeData/subscriptions/components/JsonParsingForm.tsx
genehynson: Looking at the db schema, the Json ones can be unlimited in length. The string ones have a max length of 255 chars. So I think we should probably mirror that here too?

genehynson (approved):
Love it! It's so much cleaner now - thanks :)

genehynson (approved):

PR #4393, opened 2022-04-13
feat: shared bucket context

drdelambre:
The interesting part of this PR is in the new bucket context over in `src/shared/contexts/buckets.tsx` where i try to address some of the short comings of the current bucket context that keep it scoped to singular panels within a notebook. It was a requirement from the old context to always fetch the list of bucket resources to ensure that the bucket list you received was up to date and properly scoped to the region / org your panel was pointing to. By moving the data model to a unified location whos data model is scoped behind a hash that describes this limitation, we can now instantiate as many bucket providers as we want on a page, and the state layer will manage if we can reuse an already stored list of buckets or if we have to refetch them. And since the data storage + retrieval mechanism is now independent of the view layer, providers can also be nested as deeply as we require. This is an issue that hasn't shown up in notebooks yet, but seriously hampers code reusability within the rest of the application. The `GetResources` pattern we currently employ creates a coupling between components that is really obfuscated from each other through the reducer. Up to this point, attempts to colocate resource requirements with the components that required them (ie: i need to show a list of buckets, lets make sure we've fetched the list of buckets) in order to reduce side effects of not having called a resource's fetching mechanism (because it's defined on another part of the page) and the maintainability of having to know which resources are required for a page to run (all components that require a bucket have been removed, what removes the resource fetcher?) have resulted in crazy infinite loading loops or inefficient duplication of requests. Hopefully this is a solution to that problem.

the vast majority of the line count for this one came in duplicating the explicit bucket creation mechanism from data explorer. the current one was... built to only be used in data explorer and not reused anywhere else. Once we get some eyes on the duplication effort to ensure it's completion, the intent is to merge these mechanisms to reduce efforts in writing the same thing in notebooks and data explorer. all of that code is hidden behind the `sharedBucketCreator` flag for testing


src/flows/pipes/MetricSelector/view.tsx
```diff
@@ -35,7 +35,7 @@ const DataSource: FC<PipeProp> = ({Context}) => {
   }
 
   return (
-    <BucketProvider>
+    <BucketProvider scope={scope}>
```
appletreeisyellow, line 38: does the `scope` refer to notebook, data explorer, and wherever the `BucketProvider` is used?

drdelambre: yeah. so data explorer has a fixed scope of `{ org: org.id, region: window.location }`, but for notebooks you can query across regions so each panel needs to have a bucket provider with a potentially different scope

src/shared/contexts/buckets.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+    bucks.system.sort((a, b) =>
+      `${a.name}`.toLowerCase().localeCompare(`${b.name}`.toLowerCase())
+    )
+    bucks.user.sort((a, b) =>
+      `${a.name}`.toLowerCase().localeCompare(`${b.name}`.toLowerCase())
+    )
+    bucks.sample.sort((a, b) =>
+      `${a.name}`.toLowerCase().localeCompare(`${b.name}`.toLowerCase())
+    )
```
appletreeisyellow, line 293: there is only one bucket passed into `addBucket` function, so line 282 will construct an object with one key instead of three keys (i.e `system`, `user`, `sample`). it seems like any two of these lines will have a run-time error like `bucks.system` is undefined

drdelambre: 274 is operating on the bucket cache (`buckets`) and adding the new bucket (`bucket`) by prepopulating it into the result set of that reduce function with line 282.

drdelambre: turns out that notebooks is calling this with a cold cache before the bucket list has loaded. only affects existing notebooks before a new notebook has been created

appletreeisyellow (approved):
Thanks for making bucket sharable and optimizing the performance!

PR #4401, opened 2022-04-14
fix: InitalizeClient code snippet for python

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4402

Need to update the code snippet so that it that works. The previous one is missing or has incorrect import statements.

Figma says the code should be : 
Screen Shot 2022-04-13 at 5 24 07 PM

After fix the code snippet says this (identical to the latest figma) : 
Screen Shot 2022-04-13 at 5 38 43 PM


TCL735 (approved):

hoorayimhelping (approved):
this worked for me locally in the python shell

PR #4405, opened 2022-04-14
chore: isolate bad test

genehynson:
Isolates the dashboardView test that was causing pipelines to fail. This should help us debug it further


TCL735 (approved):

genehynson:
this is being worked in another PR: https://github.com/influxdata/ui/pull/4409

PR #4420, opened 2022-04-18
(HelpBar): create Feedback & Questions overlay with text field

### LINKED ISSUE #3456: Help Bar: Feedback form will open an overlay form
For Feedback Form Overlay, create a simple form has a text field for description, with a limit of 2500 characters

blegesse-git:
Closes #3456 

Pr creates a Feedback Form Overlay with a simple form that has a text field for description, with a limit of 2500 characters

https://user-images.githubusercontent.com/66275100/163853454-1f3d5c29-9f36-41f4-a5ec-8824729de741.mov

### DIFF
--- a/src/support/components/FeedbackQuestionsOverlay.tsx
+++ b/src/support/components/FeedbackQuestionsOverlay.tsx
@@ -23,14 +23,14 @@ interface OwnProps {
 
 const FeedbackQuestionsOverlay: FC<OwnProps> = () => {
   const {onClose} = useContext(OverlayContext)
-  const [input, setInput] = useState('')
+  const [feedbackText, setFeedbackText] = useState('')
 
   const handleSubmit = () => {
     // handle form submit
   }
 
   const handleInputChange = (e: ChangeEvent<HTMLTextAreaElement>) => {
-    setInput(e.target.value)
+    setFeedbackText(e.target.value)
   }
 
   const handleValidation = (value: string): string | null => {
@@ -39,17 +39,14 @@ const FeedbackQuestionsOverlay: FC<OwnProps> = () => {
     }
 
     if (value.length >= 2500) {
-      return 'Must be 2500 characters or less'
+      return 'Must be 2500 characters or fewer'
     }
     return null
   }
 
-  const submitButtonStatus = (): ComponentStatus => {
-    if (!input.length) {
-      return ComponentStatus.Disabled
-    }
-    return ComponentStatus.Default
-  }
+  const submitButtonStatus = feedbackText.length
+    ? ComponentStatus.Default
+    : ComponentStatus.Disabled
 
   return (
     <Overlay.Container maxWidth={600}>
@@ -59,12 +56,12 @@ const FeedbackQuestionsOverlay: FC<OwnProps> = () => {
         onDismiss={onClose}
       />
       <ErrorBoundary>
-        <Overlay.Body>
-          <Form onSubmit={handleSubmit}>
+        <Form onSubmit={handleSubmit}>
+          <Overlay.Body>
             <Form.ValidationElement
               label="Description"
               required={true}
-              value={input}
+              value={feedbackText}
               validationFunc={handleValidation}
             >
               {status => (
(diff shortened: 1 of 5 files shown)


src/support/components/FeedbackQuestionsOverlay.tsx
```diff
@@ -0,0 +1,105 @@
...
+                testID="payg-contact-support--cancel"
+              />
+              <Button
+                text="Submit"
+                color={ComponentColor.Success}
+                type={ButtonType.Submit}
+                testID="payg-contact-support--submit"
+                status={submitButtonStatus()}
+              />
+            </Overlay.Footer>
```
ChitlangeSahas, line 97: Is it possible to put this outside the `Overlay.Body` so that the buttons align to the right as per the design standard?

src/support/components/FeedbackQuestionsOverlay.tsx
```diff
@@ -0,0 +1,105 @@
...
+// Types
+import ErrorBoundary from 'src/shared/components/ErrorBoundary'
+
+interface OwnProps {
+  onClose: () => void
+}
+
+const FeedbackQuestionsOverlay: FC<OwnProps> = () => {
+  const {onClose} = useContext(OverlayContext)
+  const [input, setInput] = useState('')
```
hoorayimhelping, line 26: `input` is a bit of an overloaded term. it could refer to an html text input in a form, or a concept of input. I think it's generally better to name variables after the concept it represents in the UI. `description` is what the label is for the field, so that's a start. it might aslo be called something like `feedbackText` for example - both of these things are what the text input represents; `description` is a little more vague than `feedbackText`, and in general, I find clarity in variable names to be easier to follow and reason about.

src/support/components/FeedbackQuestionsOverlay.tsx
```diff
@@ -0,0 +1,105 @@
...
+            >
+              {status => (
+                <TextArea
+                  status={status}
+                  rows={10}
+                  testID="support-description--textarea"
+                  name="description"
+                  value={input}
+                  onChange={handleInputChange}
+                  placeholder="Describe your feeback (like/dislikes with reasoning, bug you found, what could be improved, etc.) or question (i.e. product pricing) in detail."
```
hoorayimhelping, line 78:
```suggestion
                  placeholder="Describe your feeback (like/dislikes with reasoning, bug you found, what could be improved, etc.) or question (e.g. product pricing) in detail."
```

should this be `e.g.`, which generally means 'for example' instead of `i.e.`, which generally means 'in other words'?

https://www.merriam-webster.com/words-at-play/ie-vs-eg-abbreviation-meaning-usage-difference

src/support/components/FeedbackQuestionsOverlay.tsx
blegesse-git: ahh, good call. The difference is small but the overall design looks better when the footer is outside the body.

src/support/components/FeedbackQuestionsOverlay.tsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+    return null
+  }
+
+  const submitButtonStatus = (): ComponentStatus => {
+    if (!input.length) {
+      return ComponentStatus.Disabled
+    }
+    return ComponentStatus.Default
+  }
```
hoorayimhelping, line 52: this doesn't need to be a function, it can just be set as a variable, so it doesn't have to be called down below:

```ts
const submitButtonStatus = input.length? ComponentStatus.Default : ComponentStatus.Disabled
```

hoorayimhelping (approved):
clean, well-scoped PR! 👍

just a couple of suggestions

src/support/components/FeedbackQuestionsOverlay.tsx
blegesse-git: much cleaner alternative!

src/support/components/FeedbackQuestionsOverlay.tsx
```diff
@@ -0,0 +1,105 @@
...
+    setInput(e.target.value)
+  }
+
+  const handleValidation = (value: string): string | null => {
+    if (value.trim() === '') {
+      return 'This field cannot be empty'
+    }
+
+    if (value.length >= 2500) {
+      return 'Must be 2500 characters or less'
```
hoorayimhelping, line 42: I think very technically, this should be:

>Must be 2500 characters or fewer

but I don't think we care that much. https://www.merriam-webster.com/words-at-play/fewer-vs-less

src/support/components/FeedbackQuestionsOverlay.tsx
blegesse-git: `e.g.` seems more appropriate here.

src/support/components/FeedbackQuestionsOverlay.tsx
blegesse-git: haha! "very technically"

Issue #4153, opened 2022-03-19
Query Experience Toggle


garylfowler:
## EPIC Description

<p></p>

The query experience team is leaning towards a flux edit focused User Interface for Query as a replacement for Data explorer.  Its focus will be on helping the user get started with flux vs. the existing visual builder as an alternative to flux editing.

<p></p>

Because this will be a big change in user experience, it will be too risky to turn it on for and have it be changed for everyone.

<p></p>

The alternative it to take an approach, similar to those done by Atlassian in the Jira product, or Microsoft with Outlook, or Google with Gmail:

<p></p>

**Give the user the chance to choose between the old view and new view.**

<p></p>

This will then give us the opportunity to get feedback and look at analytics for which view is getting used by percentage.

<p></p>

## Acceptance Criteria

<p></p>

- A visual indication will be available within Data Explorer to indicate there is a new view to try out

- When a user chooses the "new view" they will be moved to an alternate data explorer window (specs TBD after Query Experience Project)

- When in the new view, a visual indication will be available within Data Explorer to indicate they can choose to go back to the old view

- When a user chooses the "old view" they will be moved back to the old Data Explorer view

- If feasible (won't be known until after the new view is designed), any query criteria already specified should move to the new window

- An amplitude Event is generated for any toggle selection

- Feature is initially hidden behind a feature flag

<p></p>

<p></p>

## Notes

<p></p>

<p></p>

If possible, we will want to try to be able to track in amplitude how long a user stays in the new view before toggling back (and if they toggle back)

<p></p>

Would also like to be able to keep stats from time of query start to successful query with each view, but existing DE will probably need some additional instrumentation to do this.

<p></p>

<p></p>

<p></p>

<p></p>

<p></p>

<p></p>

garylfowler:
Link to feature: https://influxdata.productboard.com/feature-board/planning/features/12533460

juliajames12:
Here's the design for the toggle:

Toggle Description: [✨](https://emojipedia.org/sparkles/) Try new Data Explorer

Toggle Off:
<img width="1513" alt="Screen Shot 2022-04-11 at 3 30 40 PM" src="https://user-images.githubusercontent.com/81173353/162826847-8cd2141f-6a61-4b7b-80a8-6072b6475010.png">

Toggle On: 
[...]

PR #4425, opened 2022-04-18
feat: adding query experience toggle

drdelambre:
Closes https://github.com/influxdata/ui/issues/4153

uses the existing app user setting infrastructure to add in support for a user opting into the new query experience project. Currently behind the `influx.set('newQueryBuilder', true)` flag

Peek 2022-04-18 14-22


src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -80,19 +106,23 @@ const DataExplorerPage: FC = () => {
...
+              <TimeZoneDropdown />
+              <SaveAsButton />
+            </Page.ControlBarRight>
+          </Page.ControlBar>
+        )}
+        {!newDataExplorer && (
+          <Page.Contents fullWidth={true} scrollable={false}>
+            <DataExplorer />
+          </Page.Contents>
+        )}
```
asalem1, line 125: Wait what? Is this supposed to be a ternary?

asalem1 (commented):
So ... I know we set state onto local storage somewhere in this app, and that the goal of this feature is a stop gap for us to have the toggle state stored in local storage, but we do this thing where when a user logs out we clear out that local storage. Should this exist outside of that framework? It simplifies it in that we just get a kv pair from local storage based without needing the overhead that integrating it into the existing app state provides?

src/dataExplorer/components/DataExplorerPage.tsx
drdelambre: nope, those are two places that display nothing when the setting is set to true because we haven't written it yet

asalem1: So wouldn't it make more sense to have these be 1 thing that's wrapped by a fragment? Right now it seems like a mistake. It might make it easier to understand as its own component so we just do `!newDataExplorer && <TheWholeShabang />`

drdelambre:
> It simplifies it in that we just get a kv pair from local storage

it also adds complications when people are using the same computer or jumping between users / orgs because we have their account configured like that. it think it's best to keep with our current "user setting" mechanism and not invest in it until we have proper support on the backend to connect users to their settings without bleeding across these local state boundaries, or to build an actual preference subsystem which should be a larger effort

asalem1 (approved):

PR #4426, opened 2022-04-19
feat: use kv caching for fields and measurements

### LINKED ISSUE #4254: Notebooks - Show all Fields and Values

<p></p>

The current Notebooks interface shows the user a list of fields and values after making in the group/filter panel of the query builder.  Currently it only shows fields and values that appear in the data within the current time range set by the user in the interface.

<p></p>

But as a SaaS service provider, I want to let my users see all of the possible fields/values so that I don't have to filter the metadata when gathering it and so that I don't confuse the user and think they have data missing.

<p></p>

## Suggested Implementation

<p></p>

- When showing a list of fields/values - show the complete list, not filtered by date range

## Acceptance Criteria

<p></p>

- When user looks at the drop down list after selecting a bucket and measurement, they will see a complete list of field/tag/values; not filtered by their time range

<p></p>

## Notes

<p></p>

- Another user story handles changing the behavior to show complete list of measurements instead of filtering them by data

<p></p>

<p></p>

<p></p>

<p></p>

subirjolly:
Closes #4254 

Instead of using flux queries, use `schema.tagValues` and `schema.tagKeys` so that we can get keys and values without a time range and utilize the caching provided by the backend on these responses.


src/utils/datetime/constants.ts
```diff
@@ -1,3 +1,7 @@
 export const DEFAULT_TIME_FORMAT = 'YYYY-MM-DD HH:mm:ss'
 export const STRICT_ISO8061_TIME_FORMAT = 'STRICT_ISO8061_TIME_FORMAT'
 export const RFC3339_PATTERN = /(\d\d\d\d)(-)?(\d\d)(-)?(\d\d)(\s|T)?(\d\d)(:)?(\d\d)(:)?(\d\d)(\.\d+)?(Z|([+-])(\d\d)(:)?(\d\d))/
+
+// Hard-coded dates required by the backend for caching purposes
```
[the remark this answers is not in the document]
wolffcm: > These Exact dates are needed for caching.

Yes, that's right. We could at some point make these default arguments or provide constants in the Flux standard library, but I think we should wait on that. That would encourage users to start using storage-layer caching of metadata more broadly, and the implications of that for the platform are still unclear.

If everything goes well, and we turn the feature flag on and we see an improvement and no ill effects, then we can look at exposing storage-layer metadata caching in a more idiomatic way in Flux. Does that make sense?

src/flows/pipes/QueryBuilder/context.tsx
```diff
@@ -268,20 +266,35 @@ export const QueryBuilderProvider: FC = ({children}) => {
...
+    |> keys()
+    |> keep(columns: ["_value"])
+    |> distinct()${searchString}${previousTagString}
+    |> filter(fn: (r) => r._value != "_time" and r._value != "_start" and r._value !=  "_stop" and r._value != "_value")
+    |> sort()
+    |> limit(n: ${limit})`
+
+    if (
+      data.buckets[0].type !== 'sample' &&
+      isFlagEnabled('MeasurementFieldsNoTime')
```
wolffcm, line 281: I'm curious, does this need to be defined somewhere, or will it just get a default value of `false`?

Could this feature flag be named something more informative to indicate that it affects the query builder? Like `QueryBuilderUseMetadataCaching` or something like that?

src/flows/pipes/QueryBuilder/context.tsx
```diff
@@ -363,21 +376,35 @@ export const QueryBuilderProvider: FC = ({children}) => {
...
+      import "influxdata/influxdb/schema"`
+      queryText = `${_source}
+  schema.tagValues(
+    bucket: "${data.buckets[0].name}",
+    tag: "${cards[idx].keys.selected[0]}",
+    predicate: (r) => ${tagString},
+    start: ${CACHING_REQUIRED_START_DATE},
+    stop: ${CACHING_REQUIRED_END_DATE},
+  )${searchString}
+  |> group()
```
wolffcm, line 402: Is this call to `group()` needed here? I think that the result of `tagValues` will already be grouped altogether. If for some reason it's needed, I'm curious what it is.

```suggestion
```

src/flows/pipes/QueryBuilder/context.tsx
```diff
@@ -268,20 +266,35 @@ export const QueryBuilderProvider: FC = ({children}) => {
...
+      queryText = `${_source}
+  schema.tagKeys(
+    bucket: "${data.buckets[0].name}",
+    predicate: (r) => ${tagString},
+    start: ${CACHING_REQUIRED_START_DATE},
+    stop: ${CACHING_REQUIRED_END_DATE},
+  )
+    |> distinct()${searchString}${previousTagString}
+    |> filter(fn: (r) => r._value != "_time" and r._value != "_start" and r._value !=  "_stop" and r._value != "_value")
+    |> sort()`
```
wolffcm, line 294: I see that there was a `limit` appended to the old query that's omitted here, you need it to get the same answer when the feature flag is toggled on/off, yes?

src/flows/pipes/QueryBuilder/context.tsx
```diff
@@ -363,21 +376,35 @@ export const QueryBuilderProvider: FC = ({children}) => {
...
+  schema.tagValues(
+    bucket: "${data.buckets[0].name}",
+    tag: "${cards[idx].keys.selected[0]}",
+    predicate: (r) => ${tagString},
+    start: ${CACHING_REQUIRED_START_DATE},
+    stop: ${CACHING_REQUIRED_END_DATE},
+  )${searchString}
+  |> group()
+  |> sort()
+    |> limit(n: ${limit})`
```
wolffcm, line 404: I see that this reverses the order of `sort` and `limit` from the old query. This version here is consistent with the query for getting tag keys above, and it makes sense to me. Just want to call it out in case it wasn't intended.

wolffcm (changes requested):
This looks great. I'm requesting a few changes to the Flux queries to ensure we can get the most performance out of them.

I had a few other cosmetic suggestions, feel free to take or leave them.

src/flows/pipes/QueryBuilder/context.tsx
subirjolly: 👍

src/flows/pipes/QueryBuilder/context.tsx
subirjolly: Good catch. I think it makes more sense to have `limit` and then `sort`. So, I have made sure that it's the same in these 4 places.

src/flows/pipes/QueryBuilder/context.tsx
wolffcm: Curious why you prefer `limit |> sort` over `sort |> limit`?

The former means that you are truncating a result to the first `n` rows, and then sorting them. This result might not be deterministic; it could be different every time. I'm actually not sure what order storage will choose to return tag values and tag keys. It is better for performance though, since you are sorting fewer things in Flux.

Doing `sort |> limit` will be deterministic, but if a lot of tag keys or values are being returned it could take longer. I don't think it would make that much of a difference though.

My sense is that users would prefer to have a deterministic result. Just my $.02.

wolffcm (approved):
Looks great, I just had a question about ordering of limit vs. sort, but that's really up to what you want to do.

subirjolly:
> Looks great, I just had a question about ordering of limit vs. sort, but that's really up to what you want to do.

I was thinking about performance but I think I will switch it out so that we have sorted list, paginated(limit).

src/flows/pipes/QueryBuilder/context.tsx
philjb: > I'm actually not sure what order storage will choose to return tag values and tag keys.

should be lexically sorted - the only way we can merge results from multiple shards. 

In any case, sort -> limit is usually the pattern for determinism as Chris mentions.

subirjolly: yup, I was done for the day. So, was going to change that today: https://github.com/influxdata/ui/pull/4426#issuecomment-1103011806

subirjolly: If it's already sorted, then should I remove sort all together from these new queries?

drdelambre: wait a sec, i think the limit |> sort was done for a reason. a dumb reason. but a reason

drdelambre: if there is a huge data set, doesnt sort |> limit cause the query to iterate over the entire dataset before limiting the result? wouldn't switching to sort |> limit cause additional timeouts for our users? it's something we're constantly sensitive about. we should def test @philjb's assumption as just adding a limit would be the best of all worlds, but i dont think that query was constructed like that for deterministic results but to save the sorting from happening on the frontend

subirjolly: @philjb

drdelambre (approved):

drdelambre (approved):

src/flows/pipes/QueryBuilder/context.tsx
wolffcm: @drdelambre As Phil says, the sorting and limiting happens in the Flux engine, and in addition we have an optimization where if we see a `sort |> limit` we will do a special sort that only needs to track the top _n_ records. 

So while it is still expensive to sort many things, it may be faster than you think, especially if _n_ is something small. In my experiments, UI metaqueries seem to do `limit(n: 200)` which is quite reasonable; I wouldn't be that concerned about performance in this case.

subirjolly:
I added a basic test with flag on. `schema.tagValues` was also having the same issue. So, I switched `sort` and `limit` again as in the previous query.

PR #4430, opened 2022-04-19
feat: add messaging for the $250 credit experiment

### LINKED ISSUE #4310: 250 Credit: Rate Limit Notification
**Feature: Rate-Limit Notification**

    ACs
    - The rate-limit notification is updated to match the figma
    - The rate-limit notification remains on the screen until the user closed it (clicks the 'x') (Remove the auto-hide functionality)
    - Re-show the rate-limit notification if the user starts a new session, keep it on the screen until the user closes it. 
    
## Design

TCL735:
Closes #4310 

Adds the messaging for the $250 credit experiment, including minor styling updates where necessary.

Screen Shot 2022-04-19 at 4 30 51 PM

Screen Shot 2022-04-19 at 4 30 56 PM

Screen Shot 2022-04-19 at 4 31 09 PM

Screen Shot 2022-04-19 at 4 31 37 PM

### DIFF
--- a/src/cloud/components/RateLimitAlert.tsx
+++ b/src/cloud/components/RateLimitAlert.tsx
@@ -24,6 +24,7 @@ import {
   extractRateLimitStatus,
 } from 'src/cloud/utils/limits'
 import {event} from 'src/cloud/utils/reporting'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Constants
 import {CLOUD} from 'src/shared/constants'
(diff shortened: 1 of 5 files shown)


src/shared/components/notifications/Notifications.tsx
```diff
@@ -52,7 +52,9 @@ const Notifications: FC = () => {
               style={{maxWidth: '600px', alignItems: 'center'}}
             >
               <span style={styles}>
-                <span className="notification--message">{message}</span>
+                {message && (
+                  <span className="notification--message">{message}</span>
+                )}
```
TCL735, line 57: This is a subtle and necessary change.

When `message` is an empty string, previously the span would be rendered (due to styling) as an inline block, which means an invisible newline is inserted causing extra spacing at the top.

With this change, when `message` is an empty string, we don't render the span at all. Nothing is inserted and spacing is consistent.

src/cloud/components/RateLimitAlert.tsx
```diff
@@ -89,11 +90,16 @@ const RateLimitAlert: FC<Props> = ({alertOnly, className, location}) => {
         )
       }
     }
-  }, [showUpgrade, status])
+  }, [showUpgrade, status]) // eslint-disable-line react-hooks/exhaustive-deps
```
randycoulman, line 93: Why is this lint override needed? It doesn't look like any of the related code changed, so I'm curious why the linter would start complaining suddenly.

blegesse-git (approved):

src/cloud/components/RateLimitAlert.tsx
TCL735: It has always complained. No one else minded. But I mind. 😄

PR #4431, opened 2022-04-20
fix: add checks and fallbacks for optional properties

TCL735:
Adds optional chaining and fallbacks for properties that are not required.

When the server response is not what you expect, console errors begin to appear due to undefined properties that are assumed to be present but are defined in the interface as optional. This can be reproduced by trying to save again after a non-200 response.

The above happens because when the form fails to save, certain fields have already been sanitized or deleted. Attempting to save again will try to sanitize properties that are nested in non-existent fields.


TCL735:
When the form fails to save, certain fields have already been sanitized or deleted. Attempting to save again will try to sanitize properties that are nested in non-existent fields.

helenosheaa (approved):
lgtm, thanks for adding this!

PR #4436, opened 2022-04-20
fix(alert): keep alert description saved after an alert is edited

### LINKED ISSUE #4348: Alert description vanishes after I edit the alert
I noticed that the description of my alert gets erased after I edit the alert. To reproduce the bug you actually have to modify something in the alert (in the video I modified the message) and refresh the page. Check out the video.

https://user-images.githubusercontent.com/6411855/162329056-859263f5-0a88-41d4-a414-d30cf83cfe13.mov

appletreeisyellow:
Closes #4348 

This PR fixed a bug in alert description where the description of an alert got erased after the alert was edited.

## Before

https://user-images.githubusercontent.com/14298407/164305163-96ca83fa-a927-45d7-a811-fc0851f6c074.mov

## After

https://user-images.githubusercontent.com/14298407/164304776-ebd0df5e-caf1-439d-83c7-59cf41017a23.mov

### DIFF
--- a/src/alerting/reducers/alertBuilder.test.ts
+++ b/src/alerting/reducers/alertBuilder.test.ts
@@ -59,6 +59,7 @@ const mockState = (): AlertBuilderState => ({
   level: 'OK',
   thresholds: [],
   status: RemoteDataState.Done,
+  description: 'mock test',
 })
 
 describe('alertBuilderReducer', () => {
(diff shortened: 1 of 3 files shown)


asalem1 (approved):
That's really funny that that was never included and the issue was only reported internally after being live for a couple years. Great work with finding the solution!

Issue #4251, opened 2022-03-30
firstMile: Wizard Step 8: Update Finish Elements for Python, Javascript, and Go (What are the users' next steps?)


Luckette:
## Figma

See new Figma from @AyeAndy: https://www.figma.com/file/JfGYCOb7ZyQ8JHSqeu4fma/Onboarding-Exploration?node-id=714%3A2554)

## ACs

- The finish page has a micro-survey (thumbs up / thumbs down question)
- The finish page has the following boxes as optional next steps for the user:
           - Sample App
           - Key Concepts
           - InfluxDB University
- Add eventing for the sample app box, the key concepts box, and the InfluxDB University box. Make sure this data is tracked in amplitude per user

PR #4438, opened 2022-04-20
feat: updated Finish Page for first mile

ChitlangeSahas:
Connects https://github.com/influxdata/ui/issues/4251

Does not add user pilot links. yet. @AyeAndy said he will do it after this is merged.

Added confetti. 🎉 

https://user-images.githubusercontent.com/18511823/164521928-6a8c728a-f5e7-44db-99ed-84a54ea098e7.mov


src/homepageExperience/components/FeedbackBar.tsx
```diff
@@ -0,0 +1,38 @@
...
+  wizardEventName: string
+}
+export default class FeedbackBar extends React.Component<OwnProps> {
+  private handleThumbsUpClick = () => {
+    // handle event here
+  }
+  private handleThumbsDownClick = () => {
+    // handle event here
+  }
+  render() {
```
ChitlangeSahas, line 15: @AyeAndy we need to add the amplitude links here.

src/homepageExperience/components/steps/Finish.tsx
```diff
@@ -3,18 +3,17 @@ import {
...
-  Icon,
-  IconFont,
   ResourceCard,
 } from '@influxdata/clockface'
 
 import {
-  InfluxDBUniversityIcon,
-  VSCodePluginIcon,
+  BookIcon,
+  CodeTerminalIcon,
```
ChitlangeSahas, line 11: These icons can be generically named.

src/homepageExperience/containers/PythonWizard.tsx
```diff
@@ -116,10 +116,16 @@ export class PythonWizard extends PureComponent<null, State> {
             />
           </div>
         </aside>
-        <div className="homepage-wizard-container--main">
+        <div
+          className={classnames('homepage-wizard-container--main', {
+            showConfetti: currentStep === HOMEPAGE_NAVIGATION_STEPS.length,
+          })}
+        >
```
ChitlangeSahas, line 123: show confetti on last page.

src/homepageExperience/containers/PythonWizard.tsx
```diff
@@ -116,10 +116,16 @@ export class PythonWizard extends PureComponent<null, State> {
...
           <div
             className={classnames('homepage-wizard-container--main-wrapper', {
-              overviewSection: this.state.currentStep === 1,
+              verticallyCentered:
+                this.state.currentStep === 1 ||
+                this.state.currentStep === HOMEPAGE_NAVIGATION_STEPS.length,
```
ChitlangeSahas, line 128: vertically center first and last page.

src/homepageExperience/components/steps/Finish.tsx
```diff
@@ -28,46 +27,36 @@ export const Finish = (props: OwnProps) => {
...
-      <p>Curious to learn more? Try these next steps!</p>
+      <FeedbackBar wizardEventName={props.wizardEventName} />
+      <p style={{marginTop: '80px'}}>
+        Curious to learn more? Try these next steps!
+      </p>
       <FlexBox margin={ComponentSize.Medium} alignItems={AlignItems.Stretch}>
         <ResourceCard className="homepage-wizard-next-steps">
-          <SafeBlankLink href="https://influxdbu.com/">
-            <h4>{InfluxDBUniversityIcon}InfluxDB University</h4>
+          <SafeBlankLink href="">
```
ChitlangeSahas, line 36: just needs a link for the sample app.

src/homepageExperience/components/FeedbackBar.tsx
AyeAndy: Just to verify though, these thumbs up and down buttons have amplitude tracking with them? We would still need to have these events classified in amplitude

src/homepageExperience/components/FeedbackBar.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import {ComponentSize, FlexBox, JustifyContent} from '@influxdata/clockface'
+import {ThumbsDown, ThumbsUp} from './HomepageIcons'
```
hoorayimhelping, line 3:
```suggestion
import {ThumbsDown, ThumbsUp} from 'src/homepageExperience/components/HomepageIcons'
```

just in case we need to move this file around

src/homepageExperience/components/FeedbackBar.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type OwnProps = {
+  wizardEventName: string
+}
+export default class FeedbackBar extends React.Component<OwnProps> {
+  private handleThumbsUpClick = () => {
+    // handle event here
```
hoorayimhelping, line 10: anything keeping us from adding these events in now?

src/homepageExperience/components/FeedbackBar.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+  render() {
+    const feedbackBarStyle = {
+      background:
+        'linear-gradient(256.11deg, rgba(147, 148, 255, 0.2) 0%, rgba(81, 60, 198, 0.2) 100%)',
+      paddingLeft: '16px',
+      paddingRight: '16px',
+      maxWidth: '493px',
+    }
```
hoorayimhelping, line 22: let's move this into the homepageExperience css file since it seems pretty static

src/homepageExperience/components/HomepageExperience.scss
```diff
@@ -43,7 +43,7 @@ h1#first-mile--header {
     font-size: $cf-text-base-1;
   }
 
-  &.overviewSection {
+  &.verticallyCentered {
```
hoorayimhelping, line 46: 👍 good change

src/homepageExperience/components/steps/Finish.tsx
```diff
@@ -24,50 +25,81 @@ export const Finish = (props: OwnProps) => {
...
+      origin: {y: 0.7},
+    }
+
+    const fire = (particleRatio, options) => {
+      confetti(
+        Object.assign({}, defaults, options, {
+          particleCount: Math.floor(count * particleRatio),
+        })
+      )
+    }
```
hoorayimhelping, line 41: fyi, since we make use of the spread operator, this can be changed to:

```ts
confetti({
  ...defaults,
  ...options,
  particleCount: Math.floor(count *  particleRatio)
})
```

src/homepageExperience/components/steps/Finish.tsx
```diff
@@ -24,50 +25,81 @@ export const Finish = (props: OwnProps) => {
   useEffect(() => {
     event(`firstMile.${props.wizardEventName}.finished`)
   }, [])
+
+  const fireConfetti = () => {
```
hoorayimhelping, line 29: I think we should do two things here:

First, move this function definition outside the component so it's not getting recreated every time this component renders - nothing depends on props, or state, or anything in a component so it can be safely moved outside the render method.

Second, let's create a boolean variable outside the function component with `let` that tells us whether the confetti has been fired yet. Then before calling `fireConfetti`, check to make sure that variable is false. If it's false, fire the confetti, then set the flag to true. This way, if the user goes back and forth between the finish page and another page, they don't have a confetti shower raining down on them every time, which might get annoying. What do you think?

hoorayimhelping (changes requested):
Have a couple of small suggestions, what do you think?

src/homepageExperience/components/FeedbackBar.tsx
ChitlangeSahas: ✅

src/homepageExperience/components/FeedbackBar.tsx
ChitlangeSahas: I can add them, I just wasn't sure if we wanted to or not with this PR.

src/homepageExperience/components/FeedbackBar.tsx
ChitlangeSahas: Done!

src/homepageExperience/components/FeedbackBar.tsx
ChitlangeSahas: added, and added some color feedback to the buttons.

src/homepageExperience/components/steps/Finish.tsx
ChitlangeSahas: ✅

src/homepageExperience/components/steps/Finish.tsx
ChitlangeSahas: I changed the first 👍

I'll slack about the second suggestion, I have an idea that i think might work. Adding a boolean to the file isn't seem to be working because react probably unmounts the component when we switch back and forth.

src/homepageExperience/containers/PythonWizard.tsx
```diff
@@ -28,18 +28,24 @@ import {event} from 'src/cloud/utils/reporting'
 interface State {
   currentStep: number
   selectedBucket: string
+  finishStepCompleted: boolean
```
ChitlangeSahas, line 31: keep track whether finish step was opened during this session.

hoorayimhelping (approved):
huzzah!!!

PR #4440, opened 2022-04-21
feat: make opening links with modifiers work cross-platform

hoorayimhelping:
Closes https://github.com/influxdata/clockface/issues/723

Previously, we only recognized the `metaKey` being pressed when handling click events. This maps to `command` on MacOS, which allows links to be opened in new tabs. This doesn't work for Windows or Linux though. This change recognizes the modifier keys (ctrl+click and clicking with the middle-mouse button on Windows and Linux) for those platforms and modifies existing `metaKey` handlers to work cross platform. It also adds the proper browser context menu for links, by passing in an `href` argument.

I tested this on my personal Windows PC and got Will Cooke to test it on his Linux machine.

- Adds a function `shouldOpenLinkInNewTab` which abstracts checking the current platform and key combination on click events
  - Makes existing handlers for `metaKey` use `shouldOpenLinkInNewTab` to work cross-platform
    - `DashboardCard`, `TaskCard`, `FlowsCard`
  - Adds a handler for `CheckCard` to be opened with a `ctrl`|`cmd` click
- Adds unit tests for `shouldOpenLinkInNewTab`
- Pulls in Clockface `3.12.0` which adds an `href` property to `ResourceCard.EditableName`, which adds the link context menu and allows `cmd`|`ctrl` clicking and middle mouse clicking

Screen Shot 2022-04-19 at 12 14 13 PM

### DIFF
--- a/src/dashboards/components/dashboard_index/DashboardCard.tsx
+++ b/src/dashboards/components/dashboard_index/DashboardCard.tsx
@@ -39,6 +39,8 @@ import {DEFAULT_DASHBOARD_NAME} from 'src/dashboards/constants'
 
 // Utilities
 import {relativeTimestampFormatter} from 'src/shared/utils/relativeTimestampFormatter'
+import {shouldOpenLinkInNewTab} from 'src/utils/crossPlatform'
+import {safeBlankLinkOpen} from 'src/utils/safeBlankLinkOpen'
 
 import {
   pinnedItemFailure,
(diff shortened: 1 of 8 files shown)


src/dashboards/components/dashboard_index/DashboardCard.tsx
```diff
@@ -232,16 +244,16 @@ class DashboardCard extends PureComponent<Props> {
       },
     } = this.props
 
-    let dest = `/${PROJECT_NAME.toLowerCase()}/from/dashboard/${id}`
+    let dashboardUrl = `/orgs/${orgID}/dashboards/${id}`
 
-    if (!isFlagEnabled('boardWithFlows')) {
-      dest = `/orgs/${orgID}/dashboards/${id}`
+    if (isFlagEnabled('boardWithFlows')) {
```
hoorayimhelping, line 249: changing the order of this `if` statement so that the change that's been flagged is the exceptional codepath for the `if` statement, not the default codepath.

blegesse-git (approved):
learned a few things about mouse event and testing from this pr. looks good to me😎

8none1:
This is a beautiful thing, thank you!

PR #4444, opened 2022-04-21
chore(Dockerfile.chronograf.prod): use node:lts-alpine

jdstrand:
Currently Dockerfile.chronograf.prod specifies node:lts-alpine3.12, but
according to https://hub.docker.com/_/node this is no longer supported.
Instead of hard-coding an alpine version, simply specify lts-alpine
(which currently is the same as lts-alpine3.15 (16-alpine3.15).

Successfully built an image using:

 $ docker build -f ./docker/Dockerfile.chronograf.prod .

I have not tested this beyond building the image.

PR #4445, opened 2022-04-21
chore(package.json): use node engine 14.17.0

jdstrand:
Per https://nodejs.org/en/about/releases/, nodejs v12 is going EOL on
2022-04-30. package.json currently specifies >=12.13.0, so update it to
14.17.0 (14 is also an LTS and supported until April 2023).
    
Ideally we would update this to the latest LTS (16.14.2), but images
used in CI/CD generated from this repo's docker/Dockerfile.cypress
currently specify quay.io/influxdb/cypress-slim:9.5.2-included which
uses:
```
FROM cypress/base:14.17.0 AS original
```
When https://github.com/influxdata/cypress-slim/blob/main/Dockerfile#L2
is updated to use node 16, then we can update the node engine in this
repo's package.json accordingly.

'yarn clean && yarn build' succeeds. I have not tested this beyond
these yarn commands.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "github:influxdata/ui"
   },
   "engines": {
-    "node": ">=12.13.0",
+    "node": ">=14.17.0",
     "yarn": ">=1.16.0"
   },
   "alias": {


jdstrand:
The failures seem unrelated. Eg, https://app.circleci.com/pipelines/github/influxdata/monitor-ci/11022/workflows/f828fead-502d-4e0a-a37e-62998f5a6fba/jobs/596576:
```
#!/bin/bash -eo pipefail
"../k8s-idpe/remocal-ci/cd-app-e2e-test-remocal-wait-for-jobs.bash" --token "${CIRCLECI_STATUS_API_TOKEN}" --org "influxdata" --project "monitor-ci" --pipeline-number "11022" --workflow "build" --exclude "remocal-finalize" --job-prefix "remocal-" --timeout 30m

cat: can't open '/tmp/workspace/namespace.txt': No such file or directory

Exited with code exit status 1

CircleCI received exit code 1
```

package.json
```diff
@@ -9,7 +9,7 @@
     "url": "github:influxdata/ui"
   },
   "engines": {
-    "node": ">=12.13.0",
+    "node": ">=14.17.0",
```
drdelambre, line 12: why this version and not LTS (16.14.2)?

jdstrand:
> @drdelambre after merging PR #4444 and bumping back to 16.14.2, CI/CD fails with:
> 
> ```
> [1/5] Validating package.json...
> error influxdb-ui@2.0.0: The engine "node" is incompatible with this module. Expected version ">=16.14.2". Got "14.17.5"
> error Found incompatible module.
> ```
> 
> I recall now that I saw this before. However, I'll try again a bit later because I suspect like with PR #4444 there is an interplay with quay.io and perhaps it isn't caught up yet? Please advise if you know of something else I should be doing.

This is coming from `quay.io/influxdb/cypress-slim:9.5.2-included`:
```
$ docker run --entrypoint /bin/bash -it --rm quay.io/influxdb/cypress-slim:9.5.2-included
root@c18561f91c5c:/# node --version
v14.17.0
```

drdelambre (approved):
Thanks for being so thorough on that dive! Seems like we got some other tires to kick before getting to lts, but that shouldn't hold this PR up.

PR #4447, opened 2022-04-22
feat: add $250 credit message to upgrade button

### LINKED ISSUE #4309: 250 Credit: Persistent 'Upgrade Now' Button & Text
**Feature: Persistent Upgrade Now Button & Text**
    
    ACs
    - "Get $250 free credit" text displays next to the 'Upgrade Now' Button on every main tab AND any sub-tab where the button currently appears 
 
## Design

TCL735:
Closes #4309 

Adds "Get $250 free credit" to the Upgrade Now button in most, but not all, contexts.

Screen Shot 2022-04-21 at 4 54 21 PM

Screen Shot 2022-04-21 at 4 54 30 PM

Screen Shot 2022-04-21 at 4 56 08 PM

Screen Shot 2022-04-21 at 4 56 15 PM

Screen Shot 2022-04-21 at 4 57 28 PM

Screen Shot 2022-04-21 at 5 45 18 PM

Screen Shot 2022-04-21 at 5 45 24 PM

### DIFF
--- a/src/cloud/components/RateLimitAlertContent.tsx
+++ b/src/cloud/components/RateLimitAlertContent.tsx
@@ -93,7 +93,7 @@ export const UpgradeContent: FC<UpgradeProps> = ({
         <UpgradeMessage {...{limitText, link, type}} />
         <CloudUpgradeButton
           className="upgrade-payg--button__rate-alert"
-          hidePromoMessage={true}
+          showPromoMessage={false}
           metric={() => event(`user.limits.${type}.upgrade`, {location})}
         />
       </FlexBox>
--- a/src/cloud/components/RateLimitAlertContent.tsx
+++ b/src/cloud/components/RateLimitAlertContent.tsx
@@ -93,6 +93,7 @@ export const UpgradeContent: FC<UpgradeProps> = ({
         <UpgradeMessage {...{limitText, link, type}} />
         <CloudUpgradeButton
           className="upgrade-payg--button__rate-alert"
+          hidePromoMessage={true}
           metric={() => event(`user.limits.${type}.upgrade`, {location})}
         />
       </FlexBox>
--- a/src/shared/components/CloudUpgradeButton.tsx
+++ b/src/shared/components/CloudUpgradeButton.tsx
@@ -20,16 +20,16 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 interface OwnProps {
   buttonText?: string
   className?: string
-  hidePromoMessage?: boolean
   metric?: () => void
+  showPromoMessage?: boolean
   size?: ComponentSize
 }
 
 const CloudUpgradeButton: FC<OwnProps> = ({
   buttonText = 'Upgrade Now',
   className,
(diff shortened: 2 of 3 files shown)


src/shared/components/CloudUpgradeButton.tsx
```diff
@@ -15,19 +15,22 @@ import CloudOnly from 'src/shared/components/cloud/CloudOnly'
...
 // Utils
 import {shouldShowUpgradeButton} from 'src/me/selectors'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 interface OwnProps {
-  className?: string
   buttonText?: string
-  size?: ComponentSize
+  className?: string
+  hidePromoMessage?: boolean
```
TCL735, line 23: We are doing it this way to minimize the number of changes. Instead of asking all consumers of `CloudUpgradeButton` to pass a new prop, we are only asking those components who need to opt-out of the promo messaging (**even when the feature is on, to avoid redundancy**) to pass a new prop.

src/cloud/components/RateLimitAlertContent.tsx
```diff
@@ -93,6 +93,7 @@ export const UpgradeContent: FC<UpgradeProps> = ({
         <UpgradeMessage {...{limitText, link, type}} />
         <CloudUpgradeButton
           className="upgrade-payg--button__rate-alert"
+          hidePromoMessage={true}
```
TCL735, line 96: There is already a separate part of this component that contains the promo message.

src/shared/components/CloudUpgradeButton.tsx
```diff
@@ -44,22 +47,32 @@ const CloudUpgradeButton: FC<OwnProps> = ({
...
+          <Button
+            icon={IconFont.CrownSolid_New}
+            className={cloudUpgradeButtonClass}
+            size={size}
+            shape={ButtonShape.Default}
+            onClick={handleClick}
+            text={buttonText}
+            style={{
+              background:
+                'linear-gradient(45deg, rgb(52, 187, 85) 0%, rgb(0, 163, 255) 100%)',
```
hoorayimhelping, line 71: might be worth considering moving this into a scss file, if it's possible

hoorayimhelping (approved):
LGTM! just a suggestion about inverting `hidePromoMessage` so we can avoid double negatives

src/shared/components/CloudUpgradeButton.tsx
TCL735: Yeah, this makes sense.

Originally, I avoided `showPromoMessage` as the prop name because when it is undefined, it has be defaulted to true.

This seems a bit unexpected for a prop, because why should a prop be truthy when it isn't passed in. Especially for a prop that is considered optional. In other words, here is an optional prop that shows something. In my mind that means, if i don't use the prop, i don't want that something shown. But surprise! That's not case. This optional prop will default to true if you don't pass it in.

But i guess that's not really that big of an issue. We will go with `showPromoMessage`

Issue #4442, opened 2022-04-21
firstMile: Code snippet syntax highlighting


hoorayimhelping:
Explore whether using https://highlightjs.org/usage/ will allow us to syntax highlight code snippets

Luckette:
Does this integrate with monaco code editor? Denise had some good questions and feedback during the ui demo on 05/13/22

PR #4450, opened 2022-04-22
feat: syntax highlighting for firstmile python

ChitlangeSahas:
Connects https://github.com/influxdata/ui/issues/4442

Bringing color syntax highlighting to python in the first mile. 

https://user-images.githubusercontent.com/18511823/164610739-64a050ba-ee3c-4e86-9e3c-bbde612a54fe.mov

### DIFF
--- a/src/shared/components/CodeSnippet.tsx
+++ b/src/shared/components/CodeSnippet.tsx
@@ -2,8 +2,11 @@
 import React, {FC, createContext, useContext, useEffect, useState} from 'react'
 import {useDispatch} from 'react-redux'
 
+import SyntaxHighlighter from 'react-syntax-highlighter'
+import {stackoverflowDark as syntaxTheme} from 'react-syntax-highlighter/dist/esm/styles/hljs'
+
 // Components
-import {DapperScrollbars} from '@influxdata/clockface'
+import {DapperScrollbars, InfluxColors} from '@influxdata/clockface'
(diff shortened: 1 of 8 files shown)


src/shared/components/CodeSnippet.tsx
```diff
@@ -109,9 +114,19 @@ const CodeSnippet: FC<Props> = ({
...
-            <code>{_text}</code>
-          </pre>
+          {language === null ? (
+            <pre>
+              <code>{_text}</code>
+            </pre>
+          ) : (
+            <SyntaxHighlighter
+              language={language}
+              style={syntaxTheme}
```
ChitlangeSahas, line 124: If the component has a `language` specified, use it, else keep the default `purple` font.

hoorayimhelping (approved):
code looks good to me, but let's hear if design thinks we should tweak the colors

brandenTenbrink (commented):
The colors seem fine for me. My only initial question was whether or not this supported more than python, but it seems it does, so it seems g2g to me.

ChitlangeSahas:
> 



> The colors seem fine for me. My only initial question was whether or not this supported more than python, but it seems it does, so it seems g2g to me.

@brandenTenbrink Yes, all other languages are also supported.

hoorayimhelping (approved):

PR #4452, opened 2022-04-22
Update Account to UserAccount

hansenreid:
Updates `Account` to `UserAccount` in accordance with the openapi change made here: https://github.com/influxdata/openapi/commit/b983e6ecc75f9b9852401e5585f2706890b10bb3

`Account` was already being aliased as `UserAccount` so the only change needed was in the type.

The plural `Accounts` was not being used so no change was needed to reflect the openapi rename from `Accounts` to `UserAccounts`. (UI just uses an array of `UserAccount`).

Part of quartz #6138 changes for multi org.


hoorayimhelping (approved):
_nice_

PR #4461, opened 2022-04-22
fix: insert before yield

drdelambre:
turns out there's some weirdness with just appending `|> limit` to things, specifically yields here. we have encountered similar issues, and the query experience project is working to more holistically remove the need for these types of patterns, so i'm seeing this as some duct tape until that work is done.


asalem1 (approved):
Key word here is duct tape for a solution that should be improved by computational infrastructure and linguistic improvements

PR #4462, opened 2022-04-22
fix: fix broken test

subirjolly:
fix: fix broken test

### DIFF
--- a/cypress/e2e/cloud/flows.test.ts
+++ b/cypress/e2e/cloud/flows.test.ts
@@ -317,6 +317,7 @@ describe('Flows with queryBuilderUseMetadataCaching flag on', () => {
     )
 
     const flowName = 'Flowbooks'
+    const flowCloneNamePrefix = `${flowName} (cloned at `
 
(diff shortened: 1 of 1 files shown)


wiedld (approved):
Nice, fast work.

cypress/e2e/cloud/flows.test.ts
```diff
@@ -411,7 +411,7 @@ describe('Flows with queryBuilderUseMetadataCaching flag on', () => {
 
     cy.getByTestID('time-machine-submit-button').should('be.visible')
 
-    const clone = `${flowName} (cloned at 11-30-2018:00:00:00)`
+    const clone = `${flowName} (cloned at 2018-11-30 00:00:00)`
```
TCL735, line 414: Why do you need to validate a specific date?

The clone's name is a valid date, not a specific date. Test for a valid date, which can be any date. Not a specific date.

TCL735 (changes requested):

TCL735 (approved):

PR #4465, opened 2022-04-23
fix(4463): millisecond discrepancy was causing wrong Duration Literal calculation

### LINKED ISSUE #4463: Notebooks/DataExplorer: misbehaviors when setting time range.
From slack convo. Summarizing here:
* The v.windowPeriod seems to jump around between two different values for each selected time range.
* in notebooks and data explorer. 
* examples:
   * If you run the same query with the same time range (past 15 min), the window period will mostly be 10S, but sometimes will be 2S, 500MS or 10S
   * if the time range is the past 5min the window period will either be 10s or 833MS
   * if the time range is the past 1min the window period will either be 1s or 166MS.
* "I don’t think I was switching between a custom time range (though it is hard to reproduce and at one stage I did try a custom time range). I was mostly switching between different selected ranges - 1m, 5m, 6hr etc."
* Intermittent error. It doesn’t happen very often.
   * When it does pick the other time range that is deterministic (i.e. if it doesn’t pick 1s for the 1 minute range, it will pick 166ms every time).

wiedld:
Closes #4463 

See ticket details, for how tracked down this bug.

### DIFF
--- a/src/shared/contexts/query.tsx
+++ b/src/shared/contexts/query.tsx
@@ -113,6 +113,8 @@ export const remove = (node: File, test, acc = []) => {
 }
 
 const _addWindowPeriod = (ast, optionAST): void => {
+  const NOW = Date.now()
+
   const queryRanges = find(
     ast,
     node =>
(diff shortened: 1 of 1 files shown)


drdelambre (approved):
Good hunting. Feels safe enough to deploy on a Friday after hours. Go nuts, just verify in prod

wiedld:
> Good hunting. Feels safe enough to deploy on a Friday after hours. Go nuts, just verify in prod

😆 will do.

TCL735:
Brilliant. Good eye.

PR #4466, opened 2022-04-23
feat: cancel queries in query builder when toggling between values

### LINKED ISSUE #4282: Performance: Notebooks Query Builder
Right now toggling between buckets / measurements / fields / tags on a bucket list, the UI fires off a request and doesn't abort the request once the original selection has been changed. We should cancel the query that's running in the background so that the user doesn't load in a bunch of unnecessary data

asalem1:
Closes #4282

### Background

There's currently an issue where whenever a user toggles between buckets / measurements / fields / tags in the query builder, a query will fire off and continue processing even when a user toggles away from that previously selected value. 

### Solution

This PR aims to resolve this by canceling existing pending queries when they are toggled away from. This is achieved by:

1. Prior to returning the promised fromFlux result in the query context, we set the `cancel` property onto it (since this was previously being wiped away in the promise chain sequence)
2. We set that promise (with the cancel property) onto the query builder context state for the specific key / value that's being loaded so that we have access to it in the event that we need to cancel the query
3. If the CardList unmounts and there's a pending promise for a given card, we fire off the cancel query
4. Updating the `selectValue` method to remove the cardList from the cards if the selected value is singular and being deselected

Visual Proof:

cancel-queries

### DIFF
--- a/src/flows/pipes/QueryBuilder/CardList.tsx
+++ b/src/flows/pipes/QueryBuilder/CardList.tsx
@@ -133,8 +133,7 @@ const Card: FC<Props> = ({idx}) => {
     card.values.selected.length <= 1
 
   const valueSelect = val => {
-    const _vals = [...card.values.selected]
-    const index = _vals.indexOf(val)
+    const index = card.values.selected.indexOf(val)
 
     if (isCompliant) {
       if (index === -1) {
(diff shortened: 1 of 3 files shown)


drdelambre (approved):
funneled the cancel on the promise for a few jumps and chased down some reference + encapsualted state issues

PR #4467, opened 2022-04-25
fix(Bug): remove unnecesary state setting to stop multiple re-renderings

### LINKED ISSUE #4451: Bug: bad setState() call inside XYPlot
Steps to reproduce 

- Go to Data Explorer 
- Add the following dummy data (or any data) to script editor and click submit 
```
import "array"
 
arrData = [
  {_time: 2022-03-09T15:00:00Z, _field: "foo", _value: 0},
  {_time: 2022-03-09T15:00:10Z, _field: "foo", _value: 10},
  {_time: 2022-03-09T15:00:20Z, _field: "foo", _value: 0},
  {_time: 2022-03-09T15:00:30Z, _field: "foo", _value: 0},
  {_time: 2022-03-09T15:00:40Z, _field: "foo", _value: 10},
  {_time: 2022-03-09T15:00:50Z, _field: "foo", _value: 10},
  {_time: 2022-03-09T15:01:00Z, _field: "foo", _value: 0},
  {_time: 2022-03-09T15:01:10Z, _field: "foo", _value: 0},
  {_time: 2022-03-09T15:01:20Z, _field: "foo", _value: 10},
  {_time: 2022-03-09T15:01:40Z, _field: "foo", _value: 0},
]
 
array.from(rows: arrData)
```
- click on `Customize` button 
- Check out the warning in the console
Screen Shot 2022-04-19 at 12 18 40 PM

Environment - error only exists locally. Tools, staging, and prod don't throw this error in console

blegesse-git:
Closes #4451 

problem pr fixes: function inside `useStaticLegend` component gets executed (unnecessarily) each time a render happens. Causing a bad setState error in the console. 

solution: wrap function in a useEffect hook so state only gets updated when one of the three values in dependency array change. 

on the second rendering (when `XYPlot` isn't rendered yet), `updateStaticLegends` gets executed one too many times.
On each render, the three values in the dependency array, `[isViewingVisOptions, show, heightRatioget]` get evaluated and in turn cause `updateStaticLegends` to get executed. This execution happens during a particular render where `XYPlot` was not already mounted on the DOM. This here is why pr executed this inside a `useEffect` so the only way function gets called is if one of the variables in the dependency array change.

### DIFF
--- a/src/visualization/utils/useStaticLegend.ts
+++ b/src/visualization/utils/useStaticLegend.ts
@@ -1,6 +1,6 @@
 // Libraries
-import {useMemo, useContext, useCallback, useEffect} from 'react'
-import {useDispatch, useSelector} from 'react-redux'
+import {useMemo, useContext, useCallback} from 'react'
+import {useDispatch} from 'react-redux'
 
 // Context
 import {PipeContext} from 'src/flows/context/pipe'
@@ -12,15 +12,10 @@ import {setStaticLegend} from 'src/timeMachine/actions'
 import {StaticLegend as StaticLegendConfig} from '@influxdata/giraffe'
 import {StaticLegend as StaticLegendAPI} from 'src/types'
 
-// Utils
-import {getActiveTimeMachine} from 'src/timeMachine/selectors'
-
 // Constants
 import {
   LEGEND_COLORIZE_ROWS_DEFAULT,
   LEGEND_OPACITY_DEFAULT,
-  LEGEND_OPACITY_MINIMUM,
-  LEGEND_OPACITY_MAXIMUM,
   LEGEND_ORIENTATION_THRESHOLD_DEFAULT,
   LEGEND_ORIENTATION_THRESHOLD_HORIZONTAL,
   LEGEND_ORIENTATION_THRESHOLD_VERTICAL,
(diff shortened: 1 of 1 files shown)


src/visualization/utils/useStaticLegend.ts
```diff
@@ -73,6 +73,64 @@ export const useStaticLegend = (properties): StaticLegendConfig => {
...
+      legendOrientationThreshold > 0
+        ? LEGEND_ORIENTATION_THRESHOLD_HORIZONTAL
+        : LEGEND_ORIENTATION_THRESHOLD_VERTICAL
+
+    updateStaticLegendProperties({
+      colorizeRows: legendColorizeRows,
+      opacity: validOpacity,
+      orientationThreshold: validThreshold,
+    })
+  }, [isViewingVisOptions, show, heightRatio])
```
TCL735, line 132: This is out of order. It looks like `updateStaticLegendProperties` is defined later.

TCL735 (commented):
So, I took a look at this more deeply. The troublesome block of code is actually not needed anymore. It was originally there to handle very old, existing dashboard cells that didn't have a static legend. The opacity would default to 0 which makes the legend invisible.

The user can fix it with the slider, but we also wanted an easier, quicker way of doing that.

However, that was many many months ago. Very likely, the vast majority of today's dashboard cells already have the proper default for the opacity.

I will push up a proper fix that will get rid of the setState warning.

src/visualization/utils/useStaticLegend.ts
TCL735: No need to address this, just wanted to comment that your proposed fix had things out of order.

TCL735: No need to address this, just wanted to comment that your proposed fix had things out of order.

hoorayimhelping:
>However, that was many many months ago. Very likely, the vast majority of today's dashboard cells already have the proper default for the opacity.

what happens if the peopel don't have the proper default for opacity and this change goes out?

TCL735:
>  what happens if the peopel don't have the proper default for opacity and this change goes out?

The static legend would be invisible if they try to show it. But there's a slider and it would be at the far left. Once the user adjusts the slider, the static legend would become visible.

hoorayimhelping:
>However, that was many many months ago. Very likely, the vast majority of today's dashboard cells already have the proper default for the opacity.

what happens if the peopel don't have the proper default for opacity and this change goes out?

hoorayimhelping:
>The static legend would be invisible if they try to show it. But there's a slider and it would be at the far left. Once the user adjusts the slider, the static legend would become visible.

If that's the case and you think it's okay to remove this (I think you're basically the PM of this feature at this point), it might be a good idea to reach out to some support people in case this does affect anyone, or it affects more people than we estimate it will

hoorayimhelping (approved):
Like I said [in this comment](https://github.com/influxdata/ui/pull/4467#issuecomment-1108923069) it might be worthwhile to consider proactively reaching out to support and letting them know this is changing

PR #4480, opened 2022-04-25
fix(3707/4021/4479): notebooks alerts multiple predicates, don't hold pipe threshold state after pipe deletion, and have flux not spam the endpoint.

### LINKED ISSUE #3707: Notebooks(Alert): Alert messages are being received as spam

### LINKED ISSUE #4021: Notebook(Alerts): multiple variables in Alert definition generates uncompilable flux in exported task

### LINKED ISSUE #4479: Notebook(alerts): threshold state is being maintained across deletion/re-creation
Found bug, and including fix in already ongoing PR. But documenting here for what it was.

https://user-images.githubusercontent.com/10232835/165152312-51962d2d-aa44-4c95-9658-d649d0ac87ed.mov

wiedld:
Closes #3707 
Closes #4021 
Closes #4479
Also fixed one other found bug.

## Issues/Bugs:
* bug 4021. The addition of multiple thresholds, resulted in invalid flux being exported to a task.
     * flux was writing `(r) => r["a"] > 0 and (r) => r["a"] = 0` . This is an incorrect lambda.
* Bug found. We were setting the default `threshold.field` to a string value `"_value"`.
     * this was occurring even without a user selecting a field.
     * Screen Shot 2022-03-09 at 7 04 58 PM
     * produced flux with `r["_value]`
* bug 3707: alerts are being received as spam
     * the filter/threshold-predicate was being given to the monitor.check() method, but never applied as a transformation in the current query
     * additionally, we have multiple rows piped to monitor.notify() --> instead of limiting to 1. For each row, a notification gets sent.
* bug 4479:
    * make threshold panel. Delete panel --> state was being persisted.
    * FIX: Delete panel --> thresholds are deleted --> re-make panel, now get blank thresholds (clean state).

## Visual proof:

https://user-images.githubusercontent.com/10232835/165164243-d6c5d10c-8771-4d87-9301-13564c27ea14.mp4

* re-check (on this new branch/PR) that we still have functional:
   * deadman checks
   * error thresholds in graph 

## Resultant Query:
Parts of the query updated/changed. (Not the full task query. See vid for that.)

```
trigger = (r) => r["co"] > 0.2 and r["humidity"] < 20 and (r["temperature"] > 0 and r["temperature"] < 100)

task_data
    |> schema["fieldsAsCols"]()
    |> set(
        key: "_notebook_link",
        value:
            "https://twodotoh-dev-denisewiedl20220405130506.remocal.influxdev.co/orgs/9c5955fc99a60b8f/notebooks/6014cd2c58952000",
    )
    |> filter(fn: (r) => r["_measurement"] == "airSensors")
    |> monitor["check"](data: check, messageFn: messageFn, crit: trigger)
    |> filter(fn: trigger)
    |> keep(columns: ["_value", "_time", "_measurement"])
    |> limit(n: 1, offset: 0)
    |> monitor["notify"](
        data: notification,
        endpoint:
            http.endpoint(url: "https://www.example.com/endpoint")(
                mapFn: (r) => {
                    body = {r with _version: 1}

                    return {headers: {"Content-Type": "application/json"}, data: json.encode(v: body)}
                },
            ),
    )
``` 

## Future work (not done in this PR):
* form validation for Notebook Alerts (issue 4229)
* use of schema (issue 4080) for:
   * form validation
   * only can select fields, which belong to the measurement chosen in the Alerts panel

### DIFF
--- a/src/flows/pipes/Notification/Measurement.tsx
+++ b/src/flows/pipes/Notification/Measurement.tsx
@@ -1,4 +1,4 @@
-import React, {useState, FC, useCallback, useContext} from 'react'
+import React, {FC, useCallback, useContext} from 'react'
 import {
   ComponentSize,
   FlexBox,
@@ -21,7 +21,7 @@ interface Props {
 
 const Measurement: FC<Props> = ({readOnly}) => {
   const {data, update, results} = useContext(PipeContext)
-  const [measurement, setMeasurement] = useState(data.measurement || null)
+  const {measurement} = data
 
   const measurements = Array.from(
     new Set(results.parsed.table.columns['_measurement']?.data as string[])
@@ -30,7 +30,6 @@ const Measurement: FC<Props> = ({readOnly}) => {
   const onSelect = useCallback(
     (measurement: string) => {
       event('Changed Notification Measurement')
-      setMeasurement(measurement)
       update({measurement})
     },
     [measurements, update]
--- a/src/flows/pipes/Notification/Measurement.tsx
+++ b/src/flows/pipes/Notification/Measurement.tsx
@@ -38,6 +38,7 @@ const Measurement: FC<Props> = ({readOnly}) => {
 
   const menuItems = measurements.map(key => (
     <Dropdown.Item
+      testID="dropdown-item--measurement"
       key={key}
       value={key}
       onClick={() => onSelect(key)}
@@ -55,6 +56,7 @@ const Measurement: FC<Props> = ({readOnly}) => {
 
   const menuButton = (active, onClick) => (
     <Dropdown.Button
+      testID="dropdown--measurements"
       onClick={onClick}
       active={active}
       size={ComponentSize.Medium}
--- a/src/flows/pipes/Notification/Measurement.tsx
+++ b/src/flows/pipes/Notification/Measurement.tsx
@@ -1,0 +1,87 @@
+import React, {useState, FC, useCallback, useContext} from 'react'
+import {
+  ComponentSize,
+  FlexBox,
+  AlignItems,
+  TextBlock,
+  Dropdown,
+  FlexDirection,
+  ComponentStatus,
+} from '@influxdata/clockface'
+
+import {PipeContext} from 'src/flows/context/pipe'
+import 'src/flows/pipes/Notification/Threshold.scss'
+
+// Utils
+import {event} from 'src/cloud/utils/reporting'
+
+interface Props {
+  readOnly?: boolean
+}
+
+const Measurement: FC<Props> = ({readOnly}) => {
+  const {data, update, results} = useContext(PipeContext)
+  const [measurement, setMeasurement] = useState(data.measurement || null)
+
+  const measurements = Array.from(
+    new Set(results.parsed.table.columns['_measurement']?.data as string[])
+  )
+
+  const onSelect = useCallback(
+    (measurement: string) => {
+      event('Changed Notification Measurement')
+      setMeasurement(measurement)
+      update({measurement})
+    },
+    [measurements, update]
+  )
+
+  const menuItems = measurements.map(key => (
+    <Dropdown.Item
+      key={key}
+      value={key}
+      onClick={() => onSelect(key)}
+      selected={key === measurement}
+      title={key}
+      disabled={!!readOnly}
+    >
+      {key}
+    </Dropdown.Item>
+  ))
+
+  const menu = onCollapse => (
+    <Dropdown.Menu onCollapse={onCollapse}>{menuItems}</Dropdown.Menu>
+  )
+
+  const menuButton = (active, onClick) => (
+    <Dropdown.Button
+      onClick={onClick}
+      active={active}
+      size={ComponentSize.Medium}
+      status={!!readOnly ? ComponentStatus.Disabled : ComponentStatus.Default}
+    >
+      {measurement || 'Select a measurement'}
+    </Dropdown.Button>
+  )
+
+  return (
+    <FlexBox
+      direction={FlexDirection.Row}
+      margin={ComponentSize.Medium}
+      alignItems={AlignItems.FlexStart}
+      testID="component-spacer"
+      style={{padding: '24px 0'}}
+    >
+      <TextBlock
+        testID="measurement-value-text-block"
+        text="For"
+        style={{minWidth: 56, textAlign: 'center'}}
+      />
+      <FlexBox.Child grow={0}>
+        <Dropdown menu={menu} button={menuButton} />
+      </FlexBox.Child>
+    </FlexBox>
+  )
+}
+
+export default Measurement
--- a/src/flows/pipes/Notification/endpoints/AWS/index.ts
+++ b/src/flows/pipes/Notification/endpoints/AWS/index.ts
@@ -36,6 +36,8 @@ export default register => {
 		crit: trigger,
 	)
   |> filter(fn: trigger)
+  |> keep(columns: ["_value", "_time", "_measurement"])
+  |> limit(n: 1, offset: 0)
 	|> monitor["notify"](
     data: notification,
     endpoint: ((r) => {
--- a/src/flows/pipes/Notification/endpoints/AWS/index.ts
+++ b/src/flows/pipes/Notification/endpoints/AWS/index.ts
@@ -26,14 +26,16 @@ export default register => {
       ['array', 'http', 'influxdata/influxdb/secrets']
         .map(i => `import "${i}"`)
         .join('\n'),
-    generateQuery: data => `task_data
+    generateQuery: (data, measurement) => `task_data
 	|> schema["fieldsAsCols"]()
       |> set(key: "_notebook_link", value: "${window.location.href}")
+  |> filter(fn: ${measurement})
 	|> monitor["check"](
 		data: check,
 		messageFn: messageFn,
 		crit: trigger,
 	)
+  |> filter(fn: trigger)
 	|> monitor["notify"](
     data: notification,
     endpoint: ((r) => {
--- a/src/flows/pipes/Notification/endpoints/HTTP/index.ts
+++ b/src/flows/pipes/Notification/endpoints/HTTP/index.ts
@@ -62,6 +62,8 @@ export default register => {
             crit: trigger,
           )
           |> filter(fn: trigger)
+          |> keep(columns: ["_value", "_time", "_measurement"])
+          |> limit(n: 1, offset: 0)
           |> monitor["notify"](data: notification, endpoint: http.endpoint(url: "${data.url}")(
             mapFn: (r) => {
                 body = {r with _version: 1}
--- a/src/flows/pipes/Notification/endpoints/HTTP/index.ts
+++ b/src/flows/pipes/Notification/endpoints/HTTP/index.ts
@@ -23,7 +23,7 @@ export default register => {
       ['array', 'http', 'influxdata/influxdb/secrets', 'json']
         .map(i => `import "${i}"`)
         .join('\n'),
-    generateQuery: data => {
+    generateQuery: (data, measurement) => {
       const headers = [['"Content-Type"', '"application/json"']]
       let prefixSecrets = ''
 
@@ -55,11 +55,13 @@ export default register => {
         task_data
           |> schema["fieldsAsCols"]()
               |> set(key: "_notebook_link", value: "${window.location.href}")
+          |> filter(fn: ${measurement})
           |> monitor["check"](
             data: check,
             messageFn: messageFn,
             crit: trigger,
           )
+          |> filter(fn: trigger)
           |> monitor["notify"](data: notification, endpoint: http.endpoint(url: "${data.url}")(
             mapFn: (r) => {
                 body = {r with _version: 1}
--- a/src/flows/pipes/Notification/endpoints/PagerDuty/index.ts
+++ b/src/flows/pipes/Notification/endpoints/PagerDuty/index.ts
@@ -21,14 +21,16 @@ export default register => {
       ['pagerduty', 'influxdata/influxdb/secrets']
         .map(i => `import "${i}"`)
         .join('\n'),
-    generateQuery: data => `task_data
+    generateQuery: (data, measurement) => `task_data
 	|> schema["fieldsAsCols"]()
-  |> set(key: "_notebook_link", value: "${window.location.href}")  
+  |> set(key: "_notebook_link", value: "${window.location.href}")
+  |> filter(fn: ${measurement})
   |> monitor["check"](
 		data: check,
 		messageFn: messageFn,
     crit: trigger,
 	)
+  |> filter(fn: trigger)
 	|> monitor["notify"](data: notification, endpoint: pagerduty.endpoint()(mapFn: (r) => ({ r with
         routingKey: "${data.key}",
         client: "influxdata",
--- a/src/flows/pipes/Notification/endpoints/PagerDuty/index.ts
+++ b/src/flows/pipes/Notification/endpoints/PagerDuty/index.ts
@@ -31,6 +31,8 @@ export default register => {
     crit: trigger,
 	)
   |> filter(fn: trigger)
+  |> keep(columns: ["_value", "_time", "_measurement"])
+  |> limit(n: 1, offset: 0)
 	|> monitor["notify"](data: notification, endpoint: pagerduty.endpoint()(mapFn: (r) => ({ r with
         routingKey: "${data.key}",
         client: "influxdata",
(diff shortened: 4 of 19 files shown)


src/flows/pipes/Notification/endpoints/HTTP/index.ts
```diff
@@ -55,11 +55,15 @@ export default register => {
         task_data
           |> schema["fieldsAsCols"]()
               |> set(key: "_notebook_link", value: "${window.location.href}")
+          |> filter(fn: ${measurement})
```
wiedld, line 58: monitor.check() will fail if we have >1 measurement in the piped data.

Even once the query builder forces a single measurement, it will still be possible (in the raw flux editor) to have >1 measurement. Therefore, this requirement in the Alerts Panel will not go away.

src/flows/pipes/Notification/endpoints/AWS/index.ts
```diff
@@ -26,14 +26,18 @@ export default register => {
...
       |> set(key: "_notebook_link", value: "${window.location.href}")
+  |> filter(fn: ${measurement})
 	|> monitor["check"](
 		data: check,
 		messageFn: messageFn,
 		crit: trigger,
 	)
+  |> filter(fn: trigger)
+  |> keep(columns: ["_value", "_time", "_measurement"])
+  |> limit(n: 1, offset: 0)
```
wiedld, line 40: FIX: reduce cardinality by selecting columns (keep()) --> then apply limit()

src/flows/pipes/Visualization/threshold.ts
```diff
@@ -18,61 +18,54 @@ export interface ErrorThreshold extends Threshold {
...
+  },
+  'not-equal': {
+    name: 'not equal to',
+    format: ThresholdFormat.Value,
+    condition: data => `${lambdaPrefix} (r["${data.field}"] != ${data.value})`,
+  },
+}
+
 export const COMMON_THRESHOLD_TYPES = {
+  ...EQUALITY_THRESHOLD_TYPES,
```
wiedld, line 37: We should have a single point of definition for `equal` and `not-equal`.
Rather than having 2 points of definition.

src/flows/pipes/Notification/Measurement.tsx
```diff
@@ -0,0 +1,87 @@
...
+// Utils
+import {event} from 'src/cloud/utils/reporting'
+
+interface Props {
+  readOnly?: boolean
+}
+
+const Measurement: FC<Props> = ({readOnly}) => {
+  const {data, update, results} = useContext(PipeContext)
+  const [measurement, setMeasurement] = useState(data.measurement || null)
```
drdelambre, line 24: please dont mirror state within the view layer like this as it causes weird bugs. just reference `data.measurement` and update the state with the `update` function

src/flows/pipes/Notification/endpoints/PagerDuty/index.ts
```diff
@@ -21,14 +21,18 @@ export default register => {
...
-  |> set(key: "_notebook_link", value: "${window.location.href}")  
+  |> set(key: "_notebook_link", value: "${window.location.href}")
+  |> filter(fn: ${measurement})
   |> monitor["check"](
 		data: check,
 		messageFn: messageFn,
     crit: trigger,
 	)
+  |> filter(fn: trigger)
+  |> keep(columns: ["_value", "_time", "_measurement"])
```
drdelambre, line 34: i dont understand these additions. what were you trying to address?

src/flows/pipes/Notification/index.ts
```diff
@@ -1,5 +1,10 @@
 import View from './view'
 import ReadOnly from './readOnly'
+import {PipeData} from 'src/types'
+
+const removeNotificationThresholds = data => {
+  delete data.thresholds
```
drdelambre, line 6: this is not what that function is for. what are you trying to do here? the next step in the chain is that `data` gets deleted,  so  i'm unsure the purpose of removing a field from it before removing the whole object

src/flows/pipes/Visualization/threshold.ts
```diff
@@ -18,31 +18,28 @@ export interface ErrorThreshold extends Threshold {
...
 }
 
+export const lambdaPrefix = '(r) =>'
+
 export const COMMON_THRESHOLD_TYPES = {
   greater: {
     name: 'greater than',
     format: ThresholdFormat.Value,
-    condition: data => `(r) => (r["${data.field}"] > ${data.value})`,
+    condition: data => `${lambdaPrefix} (r["${data.field}"] > ${data.value})`,
```
drdelambre, line 41: i'm not following. this work feels unrelated, why was this change  made in this PR?

drdelambre:
why is it that we can only have one measurement?

src/flows/pipes/Notification/endpoints/PagerDuty/index.ts
wiedld: Notification endpoints were being spammed. This is because the alerts were being triggered:
1. without the filter being applied at all
   * this is because alert.check() creates a separate task.
   * the `trigger` was only being applied within the alert check task.
   * it was NOT being applied in the current flux query. Therefore, monitor.notify() was always occurring (because `|> filter(fn: trigger)` was not occurring)

2. even once the filter is applied, there is several rows of data. Each row of data --> triggers a separate notification in the `monitor.notify()` 
   * use `keep()` to reduce cardinality of the data.
   * use `limit()` to then have a single row of data

A full description of what we are doing, and with the feedback from the flux team that this^^ query is correct, is available here:  https://influxdata.slack.com/archives/C0AD7FFS9/p1647972735067469

src/flows/pipes/Notification/index.ts
wiedld: This error was occurring:
https://github.com/influxdata/ui/issues/4479

That exact change, fixed the error.
It was occurring for the javascript pointer object (the array), but not the primitive (measurement).

src/flows/pipes/Visualization/threshold.ts
wiedld: See how the `lambdaPrefix` is then used in the:
```
THRESHOLD_TYPES[threshold.type].condition(threshold))
      .map((cond, i) => (i > 0 ? cond.split(lambdaPrefix)[1] : cond))
      .join(' and ')
```

The purpose is to fix this bug: https://github.com/influxdata/ui/issues/4021

wiedld:
> why is it that we can only have one measurement?

Because that is a requirement of monitor.check(). It will fail if you provide >1 measurement.

src/flows/pipes/Notification/Measurement.tsx
wiedld: done.

src/flows/pipes/Notification/index.ts
drdelambre: please remove this improperly added memo instead. it's preventing the threshold allocation from being properly recycled with garbage collection
https://github.com/influxdata/ui/blob/master/src/flows/pipes/Notification/Threshold.tsx#L40

src/flows/pipes/Visualization/threshold.ts
drdelambre: k so `lambdaPrefix` is just a const (should be `LAMBDA_PREFIX` in that case, and should be applied more uniformly across the codebase in case we change it). But if the intent is to just apply it on locations that might later have to be passed through this split to match this conditional rendering, then we're creating an ill defined coupling between the definition of the condition and code that conditionally mutates it. while you've updated the current dealings, there's nothing enforcing that relationship down the line, meaning these have a good chance of drifting.

since it doesn't seem like you actually want / need the `(r) =>` to be configurable (correct me if i read this wrong), a better interface might be to modify the condition functions to handle the rendering themselves by updating the function signature to something like `condition: (data: Record<"field"|"value", string>, prefix?: string) => string`, or to remove the prefix from the condition interface altogether and just conditionally apply the prefix to the whole string, so... 
```
const thresholds  = '(r) => ` + (THRESHOLD_TYPES[threshold.type].condition(threshold))
      .map((cond, i) => (i > 0 ? cond.split(lambdaPrefix)[1] : cond))
      .join(' and '))
```

src/flows/pipes/Notification/index.ts
wiedld: It's not the memoize. (That component is completely unmounted.)

Tested:
1. reverted the commit with the `removeNotificationThresholds()`
2. removed the memo.
  * Screen Shot 2022-04-26 at 10 01 34 AM

3. the issue/bug is still present (without the memo).


https://user-images.githubusercontent.com/10232835/165354223-7a3c24a8-bbef-47af-be1d-3b03ea81d0ad.mov

drdelambre: does that mean that we're not properly deep cloning from the `initial` state?

wiedld: agreed. it's a pass by reference error somewhere.
I was forcing deletion of the reference, to prevent this error.

We could alternatively enforce a deep cloning of the init state whenever we addPipe. I like that idea. Will do.

drdelambre:
please remove the `.map((cond, i) => (i > 0 ? cond.split(lambdaPrefix)[1] : cond))` mechanism. it is fragile and causes unnecessary implicit coupling of concepts

src/flows/pipes/Visualization/threshold.ts
wiedld: removed per request.

wiedld:
> please remove the `.map((cond, i) => (i > 0 ? cond.split(lambdaPrefix)[1] : cond))` mechanism. it is fragile and causes unnecessary implicit coupling of concepts

removed per request.
Note that there is still a coupling between these parts of the codebase. But it's no longer surfaced as a coupled concept...will just have to catch drift as reported bugs.

src/flows/pipes/Notification/Measurement.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+interface Props {
+  readOnly?: boolean
+}
+
+const Measurement: FC<Props> = ({readOnly}) => {
+  const {data, update, results} = useContext(PipeContext)
+  const {measurement} = data
+
+  const measurements = Array.from(
```
drdelambre, line 26: not for this pr, as it extends to the edge case, but it'd be nice if this component didn't render when `measurements.length === 1`, just to reduce the number of choices

src/flows/pipes/Notification/Measurement.tsx
```diff
@@ -0,0 +1,88 @@
...
+    </Dropdown.Button>
+  )
+
+  return (
+    <FlexBox
+      direction={FlexDirection.Row}
+      margin={ComponentSize.Medium}
+      alignItems={AlignItems.FlexStart}
+      testID="component-spacer"
+      style={{padding: '24px 0'}}
```
drdelambre, line 74: it'd be nice to see these moved out into a class, but not a blocker

drdelambre (approved):

PR #4503, opened 2022-04-27
fix: wrap upgrade button and messaging with Google Optimize

TCL735:
Follow-up for #4309 and #4310 

Wraps the feature as a Google Optimize experiment

Screen Shot 2022-04-19 at 4 30 51 PM

Screen Shot 2022-04-19 at 4 30 56 PM

Screen Shot 2022-04-19 at 4 31 09 PM

Screen Shot 2022-04-19 at 4 31 37 PM

Screen Shot 2022-04-21 at 4 54 21 PM

Screen Shot 2022-04-21 at 4 54 30 PM

Screen Shot 2022-04-21 at 4 56 08 PM

Screen Shot 2022-04-21 at 4 56 15 PM

Screen Shot 2022-04-21 at 4 57 28 PM

Screen Shot 2022-04-21 at 5 45 18 PM

Screen Shot 2022-04-21 at 5 45 24 PM


asalem1 (approved):

asalem1:
I would check the billing page for free accounts to see how it looks over there, since there's an upgrade button there. Also, didn't there used to be one on the global nav bar?

src/shared/components/CloudUpgradeButton.tsx
```diff
@@ -47,28 +50,43 @@ const CloudUpgradeButton: FC<OwnProps> = ({
...
+    <Button
+      className={cloudUpgradeButtonClass}
+      icon={IconFont.CrownSolid_New}
+      onClick={handleClick}
+      shape={ButtonShape.Default}
+      size={size}
+      testID="cloud-upgrade--button"
+      text={buttonText}
+    />
+  ) : null
```
hoorayimhelping, line 63: if we're going to return `null` from an if statement in the context of a `render` function, this can just be shortened to:

```ts
conditional && (
  returnValue
)
```

to avoid an empty `else` block. `null`, `undefined`, and boolean values are all equivalent when it comes to rendering JSX. https://stackoverflow.com/a/42083262/92446

src/cloud/components/RateLimitAlertContent.tsx
```diff
@@ -75,6 +62,32 @@ const UpgradeMessage: FC<UpgradeMessageProps> = ({limitText, link, type}) => {
...
+      original={original}
+      variants={[
+        <span className="upgrade-message" key="1">
+          You hit the{' '}
+          <a
+            href={link}
+            className="rate-alert--docs-link"
+            target="_blank"
+            rel="noreferrer"
+          >
```
hoorayimhelping, line 78: there's a component, [`SafeBlankLink`](https://github.com/influxdata/ui/blob/master/src/utils/SafeBlankLink.tsx) to handle this for you.

[here](https://github.com/influxdata/ui/blob/master/src/homepageExperience/components/steps/nodejs/WriteData.tsx#L108-L113) is an example of it being used

src/cloud/components/RateLimitAlertContent.tsx
```diff
@@ -75,6 +62,32 @@ const UpgradeMessage: FC<UpgradeMessageProps> = ({limitText, link, type}) => {
...
+              : 'query write'}
+          </a>{' '}
+          limit {limitText ?? ''} and your data stopped writing. Upgrade to get
+          a free $250 credit for the first 30 days.
+        </span>,
+      ]}
+    />
+  ) : (
+    original
+  )
```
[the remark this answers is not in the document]
hoorayimhelping: +1

hoorayimhelping (approved):
looks good, my comments are more about style than actually fixing any big problems

TCL735:
> I would check the billing page for free accounts to see how it looks over there, since there's an upgrade button there. Also, didn't there used to be one on the global nav bar?

Billing Page has its own story for this epic. Will definitely be checking to ensure it looks good and proper.

The upgrade button on the nav bar got removed, i think.

Issue #4283, opened 2022-04-01
Performance: Iterating across published versions can be slow


asalem1:
In cases where a lot of data is being queried, iterating through the published versions of a notebook can be slow. We should we what we can do to make the experience a bit more responsive and quick

subirjolly:
I'd like to start identifying the performance issues I am seeing and things I have worked on. Those fixes might be removed, if we think they aren't as useful or needed.

**Performance Issues**
1. Version List Delay - Fixed by Caching - Not needed(Per Review)
**Before** 
![image](https://user-images.githubusercontent.com/1637395/165550913-daba7c96-4334-4715-a356-1244b3d9d294.png)
**After**
![image](https://user-images.githubusercontent.com/1637395/165551021-a3a6cd2d-73f8-4cc9-b748-2136d6690e4a.png)

2. Ability to cancel queries. What I understood was that when we are moving away from one version to another or from a version back to the notebook, the longer running or high data fetching queries might make browser a little unstable. So, I added the cancellation of queries when we interact with other version(s) or go back to notebook.

I am still trying to identify performance issues. Will update this issue when I encounter more cases.

subirjolly:
https://github.com/influxdata/ui/pull/4504 - Closed PR - No need for this.
https://github.com/influxdata/ui/pull/4505 - Merged.

subirjolly:
**Performance Issue #: 3**
I have noticed that the entire UI gets stalls or hangs when a big query is running. It does periodically give control back to the mouse/browser and other components but the entire tab and mouse seems to be stuck in a particular state.

I am trying to figure out if it's possible to keep the processing localized to the components making calls.

subirjolly:
@asalem1 has tackled the `fromFlux` issue. The performance looks a lot better. There's still some hang which is happening because of the 2nd slower area(`for` loop), specifically `csvParse` to an extent.

asalem1:
Closing this since it's being addressed with performance enhancements on the fromFlux parser in giraffe

PR #4504, opened 2022-04-27
chore: first dig at performance - add persistent caching

subirjolly:
Related Issue: #4283 

This issue is vague and needs more exploration and experimentation. I have taken a first dig at this by adding persistent caching instead of moving the versions context one or two levels up and having a common provider ancestor. by adding persistent caching, initial load of the versions is almost instantaneous. Along with that, the cache is almost immediately updated as soon as the in-flight request resolves which fetches new/latest versions. The user might see a blip(maybe for a second) where the user doesn't see their latest version added in MAX last 5 minutes.

This caching also works for multiple notebooks, if the user navigates between notebooks and accesses versions.

NOTE: As mentioned above, this is the first dig and I will still continue to look at the other possible solutions along with looking at the backend code(just in case).

**Before**
BEFORE

**After**
AFTER

### DIFF
--- a/src/flows/context/version.publish.tsx
+++ b/src/flows/context/version.publish.tsx
@@ -9,6 +9,7 @@ import React, {
 } from 'react'
 import {useDispatch} from 'react-redux'
 import {useParams} from 'react-router-dom'
+import {createLocalStorageStateHook} from 'use-local-storage-state'
 
 // Context
 import {FlowContext} from 'src/flows/context/flow.current'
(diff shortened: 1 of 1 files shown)


asalem1 (changes requested):
Caching is rarely if ever the answer. This might be helpful for users to have access to, but if someone is downloading 28mb of CSVs with each version (where they’re hitting performance issues), this won’t speed things up. In fact, their browser will buckle under the amount of data that’s being stored.

PR #4505, opened 2022-04-27
chore: cancel all in-flight queries when switching versions

subirjolly:
Cancel all in-flight queries when switching notebooks versions

**My Query Time**
Query

**Working Query Cancellation**
AFTER

### DIFF
--- a/src/flows/components/VersionSidebar.tsx
+++ b/src/flows/components/VersionSidebar.tsx
@@ -53,6 +53,8 @@ const VersionSidebarListItem: FC<Props> = ({version}) => {
 
   const handleClick = () => {
     cancel()
+
+    event('click version history')
     history.push(
       `/orgs/${orgID}/${PROJECT_NAME_PLURAL.toLowerCase()}/${
         flow.id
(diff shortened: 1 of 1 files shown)


asalem1:
Is there a clip that we can verify this works or not?

subirjolly:
> Is there a clip that we can verify this works or not?

Added the gifs

subirjolly:
> > Is there a clip that we can verify this works or not?
> 
> Added the gifs

More Examples with Network Console Tab
**Example 1**
example1

**Example 2**
example2

asalem1:
It should emit a cancellation error when the query has been cancelled

subirjolly:
> It should emit a cancellation error when the query has been cancelled

if i remember correctly, we catch the cancellation error and chew it. let me confirm it.

asalem1:
> > It should emit a cancellation error when the query has been cancelled
> 
> if i remember correctly, we catch the cancellation error and chew it. let me confirm it.

It'll look red and say it was aborted. Here's how it looks:

cancel-queries

subirjolly:
> > > It should emit a cancellation error when the query has been cancelled
> > 
> > 
> > if i remember correctly, we catch the cancellation error and chew it. let me confirm it.
> 
> It'll look red and say it was aborted. Here's how it looks:
> 
> cancel-queries

**Query Aborted**
abortgif

subirjolly:
**Multiple Cancellations**
multiplecancellations

asalem1 (approved):
This is a great step in helping users access their version history in a smoother fashion. We should iterate on this to see if there are any other performance boosts we can get out of this to make the querying experience a lot smoother

PR #4507, opened 2022-04-27
feat(notebook/view): add post query statistics to tables

### LINKED ISSUE #4169: Post Query Statistics in Notebooks
## User Story

<p></p>

As a user, I would like to see some information about the query I just ran:

<p></p>

- How many tables were returned

- How may rows were returned

- How long the query took

<p></p>

So that I can better understand what happened during the query and the query results.

<p></p>

## Acceptance Criteria

<p></p>

- An additional row will be shown at the top of the validate data window within Notebooks

- The row will contain a small-font (right justified?)line  that will state, "X tables, and X  number of rows Returned, Query Time:  x ms".   **review with Julia**

- When a query is re-run, the text will go away until the re-run completes

- When a query is re-run and it completes, the information will update to reflect the latest query info

- Since we default to "last 100", the line will state "X tables, last 100 rows Returned, Query Time:  x ms" if there are 100 rows.  When less than 100, it will show the actual number as indicated previously

appletreeisyellow:
Closes #4169

This PR adds a line of post query statistics for Notebook tables.

## After

Post Query Statistics 1

https://user-images.githubusercontent.com/14298407/165555575-fabb99d9-0029-427d-a390-36e07fc58e13.mov

### DIFF
--- a/src/flows/pipes/Table/view.tsx
+++ b/src/flows/pipes/Table/view.tsx
@@ -20,7 +20,6 @@ import {View} from 'src/visualization'
 // Types
 import {RemoteDataState, SimpleTableViewProperties} from 'src/types'
 import {PipeProp} from 'src/types/flows'
-import {FromFluxResult} from '@influxdata/giraffe'
 
 import {PipeContext} from 'src/flows/context/pipe'
 import {FlowQueryContext} from 'src/flows/context/flow.query'
@@ -32,17 +31,13 @@ import {downloadTextFile} from 'src/shared/utils/download'
 // Constants
 import {UNPROCESSED_PANEL_TEXT} from 'src/flows'
 
-interface QueryStatProp {
-  result?: FromFluxResult
-  loading?: RemoteDataState
-}
-
-const QueryStat: FC<QueryStatProp> = ({result, loading}) => {
+const QueryStat: FC = () => {
+  const {loading, results} = useContext(PipeContext)
   const queryStart = useRef(0)
   const [processTime, setProcessTime] = useState(0)
   let tableNum = 0
 
-  const tableColumn = result?.table?.getColumn('table') || []
+  const tableColumn = results.parsed.table?.getColumn('table') || []
   const lastTableValue = tableColumn[tableColumn.length - 1]
 
   if (typeof lastTableValue === 'string') {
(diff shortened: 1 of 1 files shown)


asalem1:
Are the number of rows tied to the number of rows displayed or the total number of rows from the data that's returned? Also, did we want this design to apply to the simple table that's displayed for the dataa explorer?

src/visualization/components/View.tsx
```diff
@@ -26,6 +28,68 @@ interface Props {
...
+    }
+
+    if (loading === RemoteDataState.Done && queryStart !== 0) {
+      const timePassed = Date.now() - queryStart // ms
+      setQueryStart(0)
+      setProcessTime(timePassed)
+      return
+    }
+
+    setQueryStart(0)
```
drdelambre, line 70: might wanna use a `useRef` for `queryStart` as you're not expecting it to update the render cycle

appletreeisyellow:
> Are the number of rows tied to the number of rows displayed or the total number of rows from the data that's returned? Also, did we want this design to apply to the simple table that's displayed for the data explorer?

@asalem1 
- The number of rows are ties to the total number of rows from the data that's returned. 
- From what Boatie said above, seems like we want to only show the statistics in Notebook. I'm still checking with Gary.

appletreeisyellow:
@asalem1 FYI

Gary would like to have the statistics in Notebooks first. If we get positive feedback, then add it to data explorer. I think in this PR, we just want to show the query statistics in Notebook. 

Here is his original answer https://github.com/influxdata/ui/issues/4169#issuecomment-1111575195

src/flows/pipes/Table/view.tsx
```diff
@@ -21,6 +32,68 @@ import {downloadTextFile} from 'src/shared/utils/download'
 // Constants
 import {UNPROCESSED_PANEL_TEXT} from 'src/flows'
 
+interface QueryStatProp {
+  result?: FromFluxResult
+  loading?: RemoteDataState
+}
+
+const QueryStat: FC<QueryStatProp> = ({result, loading}) => {
```
asalem1, line 40: nit: can we remove the prop drilling in favor of getting the data via context?

asalem1 (approved):
Looks good to me. Minor preference for avoiding prop drilling since it becomes harder to keep track of what’s used where as components drift apart

PR #4523, opened 2022-04-29
feat: force measurement selection for data explorer

### LINKED ISSUE #4288: Data Explorer - Force Measurement Selection

<p></p>

The current Data Explorer interface allows for users (after selecting a bucket) to select a  _measurement or a _field.   This then allows them to choose a field without also requiring them to choose a measurement which causes the metadata lookup to need to grab metadata from all measurements.  This has been reported to cause performance issues with large datasets.

<p></p>

But as a SaaS service provider, I want to force my users to choose a measurement so that I can narrow that metadata and give the user a better user experience.

<p></p>

## Suggested Implementation

<p></p>

- Remove the group/filter option from the first panel the user sees after selecting a bucket.

- Give that panel a title of "Measurement"

- Show the user only a list of measurements to choose from within that panel

- The list of measurements is the complete list for that bucket, not filtered by time range

- Still allow the user to search for a measurement, but don't include fields or tags (only measurements)

- Subsequent panels work as they do today, except there is no longer a _measurement option in them

<p></p>

Design from design team:

## Acceptance Criteria

<p></p>

- User will see a measurement selection panel after selecting a bucket

- User will see a list of measurements in that panel

- User will be able to search through the list of measurements in that panel, or scroll through them

- All measurements will be shown (not just the measurements in the current time range (a change of current behavior)

- User will be able to select a measurement

- Field Selection won't be available until after a measurement has been selected

- User won't be able to add more panels until after they have made a measurement selection

- Measurement panel won't be available until after bucket selection; it could be done like we currently do with bucket selection:  

(but in this case, it would say to select a measurement first)

- User will NOT be able to select more than one measurement

- Once a measurement is selected the next panel will not give an option for _measurement

- When building the flux query, the query will have only one measurement in the script

- When converting to flux the user will only have one measurement selected (if they have selected one)

- User doesn't see a filter dropdown in the measure selection panel

<p></p>

## Design Notes

<p></p>

- The current design has a number showing the number of measurements selected; this needs final determination as some feel that more than 1 measurement should never be allowed. **UPDATE:  No display of # of measurements need to be shown as only 1 should be allowed.**

<p></p>

<p></p>

<p></p>

<p></p>

asalem1:
Closes #4288 

### Background

As we iterate on the query building experience, we want to narrow the scope of user queries by enforcing a measurement selection prior to issuing and iterating on a query within the context of the query builder. This PR translates the work that's already been done in notebooks and transfers that same logic to the data explorer.


src/timeMachine/actions/queryBuilderThunks.ts
```diff
@@ -321,10 +322,15 @@ export const selectTagValue = (index: number, value: string) => (
...
     newValues = [value]
+  } else if (isFlagEnabled('newQueryBuilder') && index === 0) {
+    newValues = [value]
   } else {
     newValues = [...values, value]
   }
 
+  if (newValues.length === 0 && index === 0) {
+    dispatch(removeTagSelector(index + 1))
+  }
```
asalem1, line 333: this condition removes the following column if the measurement has been selected and deselected. I wasn't sure if we should just reset the entire selection list, but settled on this since it emptying the list might lead to valid queries being invalidated, and this just cleans up a potentially empty query

src/timeMachine/components/SelectorList.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
   wrapText?: boolean
 }
 
-const SelectorList: SFC<Props> = props => {
+const SelectorList: FC<Props> = props => {
```
asalem1, line 17: Stateless Functional Components are deprecated. Replaced for a Functional component

src/timeMachine/components/SubmitQueryButton.tsx
```diff
@@ -150,7 +150,11 @@ const mstp = (state: AppState) => {
   const queryStatus = getActiveTimeMachine(state).queryResults.status
 
   const activeQueryText = getActiveQuery(state).text
-  const submitButtonDisabled = activeQueryText === ''
+  const tags = getActiveQuery(state).builderConfig.tags
+  const measurementTag = tags?.[0]
+  const submitButtonDisabled =
+    activeQueryText === '' ||
+    (measurementTag && measurementTag.values.length === 0)
```
asalem1, line 157: This adds a condition so that the first tag has to have a selected value in order to be a valid runnable query

drdelambre (approved):
code seems ok? gonna have to click on it to verify the product logic, but that can be done in production

PR #4527, opened 2022-04-29
fix(bug): use path property to import nested package

### LINKED ISSUE #4525: Incorrect import paths injected for nested Flux packages
## About the bug

**Steps to reproduce:**

1. Go to the data Explorer.
2. Find `usage.from` in the list of Flux functions and click **INJECT**.
3. Fill out the required `start` and `stop` parameters.
4. Run the query.

**Expected behavior:**
I expect that when injecting a function within a package, the injected import statement should be correct.

**Actual behavior:**
The package import path injected with the function is incorrect. This is a nested package. The correct import path should be `experimental/usage`.

**Visual Proof:**
image

**Solution:**
When looking at the `flux-docs-short.json` artifact the UI Flux docs are built from, it looks like it's using the package `name` when injecting the import statement. It should use the package `path`.

## About your environment

**Environment info:**

- InfluxDB version: InfluxDB Cloud
- Other relevant environment details: Production

blegesse-git:
Closes #4525 

For nested flux packages, the import statement needs to use the path property instead of func.package. One way to check if a package is nested is by checking out the path property. If path contains `/` sign, it means it's nested. If so, pr re-assigns the importStatement variable to use the path value instead of package. 

Before fix: 

https://user-images.githubusercontent.com/66275100/165997033-8ccc5ca6-95ba-4f85-b282-6ea649d4e4fa.mov

After fix: 

https://user-images.githubusercontent.com/66275100/165997250-0f9a5b66-b098-4c92-8ee3-890002ff0797.mov

### DIFF
--- a/src/flows/pipes/RawFluxEditor/view.tsx
+++ b/src/flows/pipes/RawFluxEditor/view.tsx
@@ -82,12 +82,12 @@ const Query: FC<PipeProp> = ({Context}) => {
         text = `  |> ${fn.example}`
       }
 
-      const getHeader = (fn) => {
-        let importStatement 
+      const getHeader = fn => {
+        let importStatement
 
         if (fn.package) {
           importStatement = `import "${fn.package}"`
-          if(isFlagEnabled('fluxDynamicDocs') && fn.path.includes('/')) {
+          if (isFlagEnabled('fluxDynamicDocs') && fn.path.includes('/')) {
             importStatement = `import "${fn.path}"`
           }
           return importStatement
@@ -98,7 +98,7 @@ const Query: FC<PipeProp> = ({Context}) => {
       const options = {
         text,
         type: InjectionType.OnOwnLine,
-        header: getHeader(fn)
+        header: getHeader(fn),
       }
       inject(options)
     },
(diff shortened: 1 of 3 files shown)


src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -82,10 +82,23 @@ const Query: FC<PipeProp> = ({Context}) => {
...
+        if (fn.package) {
+          importStatement = `import "${fn.package}"`
+          if (isFlagEnabled('fluxDynamicDocs') && fn.path.includes('/')) {
+            importStatement = `import "${fn.path}"`
+          }
+          return importStatement
+        }
+        return null
+      }
+
```
TCL735, line 97: Would this version be clearer and easier to read?

```
      const getHeader = fn => {
        let importStatement = null

        if (fn.package) {
          importStatement = `import "${fn.package}"`
          if (isFlagEnabled('fluxDynamicDocs') && fn.path.includes('/')) {
            importStatement = `import "${fn.path}"`
          }
        }
        return importStatement
      }
```

TCL735 (changes requested):

src/flows/pipes/RawFluxEditor/view.tsx
blegesse-git: oh for sure!

blegesse-git: oh for sure!

TCL735 (approved):

TCL735 (approved):

PR #4546, opened 2022-05-10
feat: inject pipe-forward operator only when necessary 

### LINKED ISSUE #4535: Not all injected Flux functions should include a pipe-forward operator
## About the bug

**Steps to reproduce:**

1. In the **Data Explorer**, select **Script Builder** and search for `uint()` in the list of functions.
2. `universe.uint()` is returned.
3. Click **Inject** next to the function.

**Expected behavior:**
`universe.uint()` does not accept piped-forward data, so it should not include the pipe-forward operator (`|>`).

**Actual behavior:**
All functions are injected with a pipe-forward operator, even those that do not accept piped-forward data.

**Visual Proof:**
image

## Proposed solution
I think we could look at the `flux_type` for each function and look for a pipe-receive operator (`<-`). If the type signature doesn't include a pipe-receive operator, the function should not be injected with a pipe-forward operator.

blegesse-git:
Closes #4535 

In both current and older flux panel UI, all flux functions get injected in the editor with a pipe-forward operator `|>`. 
This isn't correct. Not all functions require `|>`. Pr leverages the `fluxType` property for each flux function to determine whether pipe-forward sign is needed. All functions that require this operator contain a pipe-receive operator `<-` in the fluxType property. 

Before: 

https://user-images.githubusercontent.com/66275100/167695593-fad8c98e-ab12-4971-8428-ac8b2d4a269f.mov

After: 

https://user-images.githubusercontent.com/66275100/167695446-a3ac26e0-e7e1-408e-b8cf-be18115584a5.mov

### DIFF
--- a/src/flows/pipes/RawFluxEditor/view.tsx
+++ b/src/flows/pipes/RawFluxEditor/view.tsx
@@ -76,10 +76,15 @@ const Query: FC<PipeProp> = ({Context}) => {
   const injectIntoEditor = useCallback(
     (fn): void => {
       let text = ''
-      if (fn.name === 'from' || fn.name === 'union') {
-        text = `${fn.example}`
+      if (isFlagEnabled('fluxDynamicDocs')) {
+        // only fluxTypes with <- sign require a pipe forward sign
+        text = fn.fluxType[1] === '<' ? `  |> ${fn.example}` : `${fn.example}`
       } else {
-        text = `  |> ${fn.example}`
+        if (fn.name === 'from' || fn.name === 'union') {
+          text = `${fn.example}`
+        } else {
+          text = `  |> ${fn.example}`
+        }
       }
 
       const getHeader = fn => {
--- a/src/flows/pipes/RawFluxEditor/view.tsx
+++ b/src/flows/pipes/RawFluxEditor/view.tsx
@@ -76,9 +76,11 @@ const Query: FC<PipeProp> = ({Context}) => {
   const injectIntoEditor = useCallback(
     (fn): void => {
       let text = ''
-      if (isFlagEnabled('fluxDynamicDocs')) {
+      if (CLOUD && isFlagEnabled('fluxDynamicDocs')) {
         // only fluxTypes with <- sign require a pipe forward sign
-        text = fn.fluxType.startsWith('<-', 1) ? `  |> ${fn.example}` : `${fn.example}`
+        text = fn.fluxType.startsWith('<-', 1)
+          ? `  |> ${fn.example}`
+          : `${fn.example}`
       } else {
         if (fn.name === 'from' || fn.name === 'union') {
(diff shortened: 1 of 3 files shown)


src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -76,10 +76,15 @@ const Query: FC<PipeProp> = ({Context}) => {
   const injectIntoEditor = useCallback(
     (fn): void => {
       let text = ''
-      if (fn.name === 'from' || fn.name === 'union') {
-        text = `${fn.example}`
+      if (isFlagEnabled('fluxDynamicDocs')) {
+        // only fluxTypes with <- sign require a pipe forward sign
+        text = fn.fluxType[1] === '<' ? `  |> ${fn.example}` : `${fn.example}`
```
hoorayimhelping, line 81: you might consider something like:

```ts
fn.fluxType.includes('<-') ? ` |> ${fn.example}` : `${fn.example}`
```
Seems slightly less brittle and more explicit in the business rule (it will only work with `<-`, rather than `<afasdfasdf` (which may not be an issue). https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes.

Can also consider [`startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) if the `<-` sequence needs to be at the beginning of the string

hoorayimhelping (commented):
just a comment on considering alternatives to checking the index of the string

src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -76,10 +76,15 @@ const Query: FC<PipeProp> = ({Context}) => {
   const injectIntoEditor = useCallback(
     (fn): void => {
       let text = ''
-      if (fn.name === 'from' || fn.name === 'union') {
-        text = `${fn.example}`
+      if (isFlagEnabled('fluxDynamicDocs')) {
```
TCL735, line 79: Is `CLOUD` already guaranteed here? Maybe we should be explicit and check to make sure we in `CLOUD`

src/flows/pipes/RawFluxEditor/view.tsx
blegesse-git: but I don't see any problems with using `startsWith`. I'll update the code with that. Thanks bucky :)

src/flows/pipes/RawFluxEditor/view.tsx
blegesse-git: we explicitly check if we're in `CLOUD` in `TimeMachine` component, so it would make sense to include it here as well

TCL735 (approved):

hoorayimhelping (approved):

PR #4550, opened 2022-05-10
chore: remove config for semantic-pull-requests, add call to influx w…

tysonkamp:
…orkflow to replace.

We are removing semantic-pull-requests org wide. Influxdata has its own semantic validator that runs as a workflow triggered by GH Actions.  To address the settings in the semantic-pull-requests config file, the following will be done:

allowMergeCommits: true -> this is already allowed
allowRevertCommits: true -> this is already allowed
titleOnly: true -> this is configurable and will be noted in the commit referring to COMMITS_HISTORY: 0

If the team would prefer that the PR title was only validated when merging into master, this is also achievable.  Please comment and mention @tysonkamp to have this functionality configured (it would be in  the semantic.yml).

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,4 +1,0 @@
-# docs: https://github.com/probot/semantic-pull-requests#configuration
-allowMergeCommits: true
-allowRevertCommits: true
-titleOnly: true
--- a/.github/workflows/semantic.yml
+++ b/.github/workflows/semantic.yml
@@ -1,0 +1,10 @@
+---
+name: "Semantic PR and Commit Messages"
+
+on:
+  pull_request:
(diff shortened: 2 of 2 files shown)


asalem1 (approved):

hoorayimhelping:
@tysonkamp Thanks for the explanation. It's kind of alarming that it's still using such an old version of nodejs, and it makes sense we'd want something more secure. I'm not super worried about this change per-se, it doesn't seem like anything too disruptive or crazy; However, it was a bit of a surprise, and it got me thinking about the precedent this sets for a future change that we're not aware of and does happen to be disruptive. I just want to make sure we're aware of upcoming changes in the future

tysonkamp:
Would it be appropriate per the team's practices for me to squash and merge at this point?

PR #4551, opened 2022-05-11
feat: add Credit 250 panel to Billing Page for experiment

### LINKED ISSUE #4308: 250 Credit: Billing Page
**Feature: Billing Page**

    ACs
    - Lower Panel around the upgrade now button includes additional text and is re-created to match the figma

## Design

TCL735:
Closes #4308  

Screen Shot 2022-05-10 at 5 33 27 PM

### DIFF
--- a/src/billing/components/Free/PAYGConversion.tsx
+++ b/src/billing/components/Free/PAYGConversion.tsx
@@ -17,6 +17,8 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {GoogleOptimizeExperiment} from 'src/cloud/components/experiments/GoogleOptimizeExperiment'
 import {CREDIT_250_EXPERIMENT_ID} from 'src/shared/constants'
 
+const ONE_MILLION = 1_000_000
+
 export const Credit250PAYGConversion: FC = () => {
   if (isFlagEnabled('credit250Experiment')) {
     return (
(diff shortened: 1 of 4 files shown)


src/billing/components/Free/PAYGConversion.tsx
```diff
@@ -12,8 +12,64 @@ import {
...
+            <Grid.Column widthSM={Columns.Eight}>
+              <div className="conversion-panel--list">
+                <ul className="credit-250-conversion-panel--benefits">
+                  <li>Unlimited buckets to store your data</li>
+                  <li>Unlimited storage retention</li>
+                  <li>Unlimited dashboards</li>
+                  <li>Unlimited tasks</li>
+                  <li>Unlimited alert checks and notification rules</li>
+                  <li>HTTP and PagerDuty notifications</li>
+                  <li>Up to 1,000,000 series cardinality</li>
```
hoorayimhelping, line 60: Let's wrap this in a `new Intl.NumberFormat()`. I'm not sure if we have requirements that this be formatted in American english, but even if we hard code the locale to `en-US`, this will give us a bit of future proofing if we ever decide to go international. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format.

Something like

```ts
const numberFormatter = new Intl.NumberFormat('en-US')
...
<li>Up to {numberFormatter.format(1000000)} series cardinality</li>
```

PR #4531, opened 2022-05-07
fix: Missing comma on tickscript.join - Fixed

Xoffio:
Missing comma added on `tickscript.join` function.

If you `Inject` the function `tickscript.join` on the ui. You will get the next error:

```
compilation failed: error @19:6-23:2: expected comma in property list, got COLON

error @21:7-21:18: missing property key 
```

PR #4560, opened 2022-05-11
fix: fix bug with the example code and test our CI pipeline for weirdness

asalem1:
This issue was originally brought to our attention in [this PR](https://github.com/influxdata/ui/pull/4531/). For some reason, that one is failing in the CI pipeline. It's worth investigating the root cause of that failure. As such, I'm making the same PR to try and isolate the issue


asalem1:
Closing this since it seems like the tests are built around false assumptions

PR #4576, opened 2022-05-12
feat: turn on queries for new DE

drdelambre:
turns on the ability to run queries in the experimental DE

chrome-capture-2022-4-12


appletreeisyellow (approved):
Great!! I like how neat the current `NewDataExplorer` is!

PR #4578, opened 2022-05-12
chore: use Intl.NumberFormat

TCL735:
This should have been included in https://github.com/influxdata/ui/pull/4551

When the feature is off, we still want to use `Intl.NumberFormat` to display large numbers that use grouping.

Screen Shot 2022-05-12 at 3 49 49 PM

### DIFF
--- a/src/billing/components/Free/PAYGConversion.tsx
+++ b/src/billing/components/Free/PAYGConversion.tsx
@@ -109,7 +109,10 @@ export const PAYGConversion: FC = () => {
                       <li>Unlimited tasks</li>
                       <li>Unlimited alert checks and notification rules</li>
                       <li>HTTP and PagerDuty notifications</li>
-                      <li>Up to 1,000,000 series cardinality</li>
+                      <li>
+                        Up to {Intl.NumberFormat().format(ONE_MILLION)} series
+                        cardinality
+                      </li>
                     </ul>
                   </div>
                 </div>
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):

src/billing/components/Free/PAYGConversion.tsx
```diff
@@ -109,7 +109,10 @@ export const PAYGConversion: FC = () => {
                       <li>Unlimited tasks</li>
                       <li>Unlimited alert checks and notification rules</li>
                       <li>HTTP and PagerDuty notifications</li>
-                      <li>Up to 1,000,000 series cardinality</li>
+                      <li>
+                        Up to {Intl.NumberFormat().format(ONE_MILLION)} series
```
hoorayimhelping, line 113: the constant `ONE_MILLION` is pretty much the same as using the magic number `1,000,000` (except it might be worse, because it's a layer of indirection). Assigning numerical values to constants is less about assigning the value of the number to a constant and more about naming the constant in a way that explains what its purpose is. What does the number mean, not what is the identity of the number. The constant `ONE_MILLION` will always be `1000000`, so we might as well just use the literal value `1000000`

Instead of being called `ONE_MILLION`, this might be more valuable if it was called `CARDINALITY_LIMIT`. That way, anywhere we need the system-wide cardinality limit, we could just use the variable. If in the future, after IOX comes out let's say, we increase the cardinality limit to 10,000,000, we just increase the value of the variable and everywhere else works. If the constant is called `ONE_MILLION` and we increase the cardinality limit to 10,000,000, we'd have to create a new variable and change all the places `ONE_MILLION` is used to represent cardinality.

PR #4579, opened 2022-05-13
feat(client_libs): add Scala WriteAPI example

bednar:
Related to https://github.com/influxdata/influxdb-client-java/pull/347

1. Added examples how to use Scala's `WriteAPI`:

image

2. Updated client versions


appletreeisyellow (approved):
Thank you for updating and adding client libraries to make InfluxData more diverse! 🎉

bednar:
@appletreeisyellow thanks for approving 👍

Issue #4464, opened 2022-04-22
250 Credit: send the signup date to google optimize


Luckette:
Optimize will handle the logic: 
- 50% of all free accounts that signed up AFTER Mar 8 will be part of the experimental group that sees the $250 credit offer
- 100% of "legacy" free accounts, that is, those that signed up PRIOR to Mar 8 will be part of the experimental group that sees the $250 credit offer

PR #4582, opened 2022-05-13
fix: wrap experiment properties under identity object

TCL735:
Follow-up to #4464 

Wraps the properties for the Credit 250 Experiment under an `identity` object before pushing to the `dataLayer`

### DIFF
--- a/src/shared/containers/GetOrganizations.tsx
+++ b/src/shared/containers/GetOrganizations.tsx
@@ -22,7 +22,8 @@ import {getQuartzMe as apiGetQuartzMe} from 'src/me/actions/thunks'
 import RouteToOrg from 'src/shared/containers/RouteToOrg'
 
 // Selectors
-import {getQuartzMe} from 'src/me/selectors'
+import {getAllOrgs} from 'src/resources/selectors'
+import {getMe, getQuartzMe} from 'src/me/selectors'
 
 // Constants
 import {CLOUD} from 'src/shared/constants'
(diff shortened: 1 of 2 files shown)


src/shared/containers/GetOrganizations.tsx
```diff
@@ -61,16 +64,31 @@ const GetOrganizations: FunctionComponent = () => {
...
+          account_type,
+          account_created_at,
+          id: meId,
+          email,
+          organization_id: orgId,
+        },
+      })
     }
-  }, [dispatch, me, quartzMeStatus])
+  }, [quartzMeStatus, status]) // eslint-disable-line react-hooks/exhaustive-deps
```
TCL735, line 91: If we use all of the dependencies suggested by the linter, then the `identity` gets pushed again to the `dataLayer` unnecessarily. We just want the status of the two api calls to be `RemoteDataState.Done` before executing the push.

src/shared/containers/GetOrganizations.tsx
```diff
@@ -89,10 +107,10 @@ const GetOrganizations: FunctionComponent = () => {
               />
               <Route path="/orgs" component={App} />
               <Route exact path="/" component={RouteToOrg} />
-              {CLOUD && canAccessCheckout(me) && (
+              {CLOUD && canAccessCheckout(quartzMe) && (
```
hoorayimhelping, line 110: what's the difference between using `quartzMe` and `me` here?

hoorayimhelping (approved):

src/shared/containers/GetOrganizations.tsx
TCL735: It has always been an object related to `quartzMe` but was previously named `me`. Yes, very confusing.

PR #4590, opened 2022-05-16
fix: add fallback OrgsState for getAllOrgs

TCL735:
Part of #4464 

Adds an error state to `OrgsState` when `resources.orgs` cannot be found. Initially, `OrgsState` is set by the reducer, which should not use this fallback unless `resources.orgs` was modified by something else.

### DIFF
--- a/src/resources/selectors/index.ts
+++ b/src/resources/selectors/index.ts
@@ -79,5 +79,5 @@ export const getAllOrgs = (state: AppState): OrgsState => {
     status: RemoteDataState.Error,
     org: {id: ''},
   }
-  return get(state, 'resources.orgs', errorOrgsState) || errorOrgsState
+  return get(state, 'resources.orgs', errorOrgsState)
 }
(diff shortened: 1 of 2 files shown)


src/resources/selectors/index.ts
```diff
@@ -74,5 +74,10 @@ export const getLabels = (state: AppState, labelIDs: string[]): Label[] => {
...
 
-export const getAllOrgs = (state: AppState): OrgsState =>
-  get(state, 'resources.orgs', {}) || {}
+export const getAllOrgs = (state: AppState): OrgsState => {
+  const errorOrgsState = {
+    status: RemoteDataState.Error,
+    org: null,
+  }
+  return get(state, 'resources.orgs', errorOrgsState) || errorOrgsState
+}
```
TCL735, line 83: I'm actually not sure if this is even necessary. Maybe the change in the other file is sufficient.

blegesse-git (approved):

hoorayimhelping (approved):

PR #4595, opened 2022-05-17
fix(flux-4720): temporary fix to hardcode top20 examples, plus add cursor awarenss in function injection

wiedld:
Temporary fix for https://github.com/influxdata/flux/issues/4720

## Bug:
* Pro: have dynamic list of available flux functions, served from the backend.
* Con: we lost the hardcoded examples for flux functions.
  * suggestions from LSP are not sufficient. 

## Solution:
* Hardcode the top 20 flux functions, with an example.
  * future plans are to have the backend (or LSP) provide data for examples. But not yet ready. 
* Make sure the example insertion considers the cursor position, versus the type of insertion (e.g. on the same line, or another line).
    * This just meant using existing code for `calcInjectionPosition()`.
    * knows when the flux function should be injected on the same line, vs different line. Then does the math.
    * For different line (on own line), injects based on where cursor is in line --> will inject in line before or after (cursor awareness).
* After insert, poke the LSP for suggestions.
    * We can readily turn this off, if it’s not liked by people. See the vids.

## Tradeoffs made:
* Did not include the top20 function `universe.from`, since we don’t use it as an injected function.
   * Injected from is only package-scoped `csv.from()`, and not the universe.from().
* Did not include all possible parameter arguments (completeness) per example.
    * Otherwise, we have a lot of `tables: [table1, table2]` for many functions.
    * And this may imply (incorrectly) that these arguments are mandatory for the function.
* Instead, used the previously hardcoded functions as a guide of what is useful.

## Future work (not this PR):
* The TimeMachineFluxEditor is not yet using the editor.context.
    * It’s almost to the point that it can use this context. 
    * Except that we have the EditorContext still consuming from the PipeContext. So cannot use in timeMachine yet.
* As a result of this^^ we have:
    * Tight coupling between the `src/shared/utils/fluxFunctions` and the `src/flows/context/editor`

## Visual evidence:

timeMachine, in top 20:

https://user-images.githubusercontent.com/10232835/168706296-d3f612ec-36e4-4d43-9437-92453af45d45.mp4

timeMachine, not top 20:

https://user-images.githubusercontent.com/10232835/168706177-00c38a6e-1cbd-4018-b14d-5db1cd14edce.mp4

notebooks, in top 20:

https://user-images.githubusercontent.com/10232835/168707206-3a8d63ce-a226-4079-9b34-88caa095f0a5.mp4

### DIFF
--- a/cypress/e2e/shared/explorerVisualizations.test.ts
+++ b/cypress/e2e/shared/explorerVisualizations.test.ts
@@ -89,6 +89,7 @@ describe('visualizations', () => {
           cy.log('check to see if new aggregate rate is at the bottom')
           cy.get('.view-line')
             .last()
+            .prev()
             .contains('aggregate.')
           cy.getByTestID('flux-editor').should('exist')
           cy.getByTestID('flux-editor').monacoType(`yoyoyoyoyo`)
--- a/src/flows/context/editor.tsx
+++ b/src/flows/context/editor.tsx
@@ -9,8 +9,8 @@ import {EditorType} from 'src/types'
 import {PipeContext} from 'src/flows/context/pipe'
 import {
   InjectionType,
-  InjectionPosition,
   calcInjectionPosition,
+  moveCursorAndTriggerSuggest,
 } from 'src/shared/utils/fluxFunctions'
 export {InjectionType, InjectionPosition} from 'src/shared/utils/fluxFunctions'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
@@ -26,7 +26,6 @@ export interface EditorContextType {
   editor: EditorType | null
   setEditor: (editor: EditorType) => void
   inject: (options: InjectionOptions) => void
-  calcInjectiontPosition: (type: InjectionType) => Partial<InjectionPosition>
   updateText: (t: string) => void
 }
 
@@ -34,7 +33,6 @@ const DEFAULT_CONTEXT: EditorContextType = {
   editor: null,
   setEditor: _ => {},
   inject: _ => {},
-  calcInjectiontPosition: _ => ({}),
   updateText: _ => {},
 }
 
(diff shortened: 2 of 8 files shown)


cypress/e2e/shared/explorerVisualizations.test.ts
```diff
@@ -89,6 +89,7 @@ describe('visualizations', () => {
           cy.log('check to see if new aggregate rate is at the bottom')
           cy.get('.view-line')
             .last()
+            .prev()
             .contains('aggregate.')
```
wiedld, line 93: we have a newline `\n` in the last view-line.
I could make it:
```
.last()
.should() // test equal to 2 nested, empty spans
.prev()
.contains('aggregate.')
```

src/shared/utils/fluxFunctions.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+export interface InjectionPosition {
+  row: number
+  column: number
+  shouldStartWithNewLine: boolean
+  shouldEndInNewLine: boolean
+}
+
+export function calcInjectionPosition(
```
wiedld, line 36: this function is exactly the same, only extracted to be used with any monaco-editor (not just the editor.context used by notebooks).

src/shared/utils/fluxFunctions.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  const cursorInMiddleOfText = textAheadOfCursor && textBehindCursor
+  if (type == InjectionType.OnOwnLine && cursorInMiddleOfText) {
+    row++
+    shouldEndInNewLine = true
+  }
+  if (type == InjectionType.OnOwnLine) {
+    column = 1
+  }
```
wiedld, line 87: changed from
```
  if (type == InjectionType.OnOwnLine && cursorInMiddleOfText) {
    row++
    column = 1
    shouldEndInNewLine = true
  }
```

src/flows/context/editor.tsx
```diff
@@ -7,23 +7,19 @@ import React, {
...
-  column: number
-  shouldStartWithNewLine: boolean
-  shouldEndInNewLine: boolean
-}
+import {
+  InjectionType,
+  InjectionPosition,
+  calcInjectionPosition,
+} from 'src/shared/utils/fluxFunctions'
+export {InjectionType, InjectionPosition} from 'src/shared/utils/fluxFunctions'
```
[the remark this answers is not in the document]
wiedld: Note that making a generalized editor.context will be also required for part of the QE flux builder. (Because all the schema injections and function injections will use the same logic). So this follow-up PR will definitely be occurring. 😸

rockstar (approved):

appletreeisyellow (approved):

PR #4600, opened 2022-05-18
fix: update usage panel to use Credit 250 experiment

### LINKED ISSUE #4385: 250 Credit: Update Progress Widget with new experiment ID (usage panel in the UI)

TCL735:
Closes #4385 
Also part of quartz 6164

- Updates the Usage Panel to work with the Credit 250 Experiment

Screen Shot 2022-05-17 at 5 03 32 PM

### DIFF
--- a/src/usage/context/usage.test.ts
+++ b/src/usage/context/usage.test.ts
@@ -1,0 +1,47 @@
+import {calculateCreditDaysUsed} from 'src/usage/context/usage'
+
+describe('calculateCreditDaysUsed', () => {
+  const ONE_DAY_MILLSECONDS = 1000 * 60 * 60 * 24
+
+  it('handles falsy values as invalid date', () => {
+    expect(calculateCreditDaysUsed(undefined)).toEqual(-1)
+    expect(calculateCreditDaysUsed(null)).toEqual(-1)
+    expect(calculateCreditDaysUsed('')).toEqual(-1)
+  })
+
+  it('handles valid dates in the past', () => {
+    const yesterday = new Date(Date.now() - ONE_DAY_MILLSECONDS).toISOString()
+    const twoDaysAgo = new Date(
+      Date.now() - 2 * ONE_DAY_MILLSECONDS
+    ).toISOString()
+    const oneWeekAgo = new Date(
+      Date.now() - 7 * ONE_DAY_MILLSECONDS
+    ).toISOString()
+    const thirtyDaysAgo = new Date(
+      Date.now() - 30 * ONE_DAY_MILLSECONDS
+    ).toISOString()
+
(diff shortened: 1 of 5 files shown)


src/usage/context/usage.tsx
```diff
@@ -78,6 +78,17 @@ export const UsageContext = React.createContext<UsageContextType>(
   DEFAULT_CONTEXT
 )
 
+export const calculateCreditDaysUsed = (creditStartDate: string): number => {
+  const MILLISECONDS_IN_ONE_DAY = 1000 * 60 * 60 * 24
```
hoorayimhelping, line 82: It's not essential for this issue, but I think it's a good idea to move this constant to the [datetime constants](https://github.com/influxdata/ui/blob/master/src/utils/datetime/constants.ts) file. That way, it doesn't need to be defined in this file and redefined in the tests. If you don't move it to that file in this PR, at the very least I think this should be moved out of the function declaration and higher up into the file since it's a constant and it doesn't depend on any argument passed into the function. No need to redeclare this constant every time the function is called.

src/usage/context/usage.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    const thirtyDaysFromNow = new Date(
+      Date.now() + 30 * ONE_DAY_MILLSECONDS
+    ).toISOString()
+
+    expect(calculateCreditDaysUsed(tomorrow)).toEqual(-1)
+    expect(calculateCreditDaysUsed(twoDaysFromNow)).toEqual(-2)
+    expect(calculateCreditDaysUsed(oneWeekFromNow)).toEqual(-7)
+    expect(calculateCreditDaysUsed(thirtyDaysFromNow)).toEqual(-30)
+  })
+})
```
hoorayimhelping, line 47: thank you for adding in these tests. nice work

hoorayimhelping (approved):
Code looks good to me. A couple of considerations that don't need to block this work.

One question I couldn't answer: how does this look with `firstMile` enabled? I couldn't get the usage api panel to show up locally, even after toggling `newUsageAPI`

src/usage/context/usage.tsx
```diff
@@ -78,6 +78,17 @@ export const UsageContext = React.createContext<UsageContextType>(
...
+  if (!creditStartDate) {
+    return -1
+  }
+  const startDate = new Date(creditStartDate)
+  const current = new Date()
+  const diffTime = current.getTime() - startDate.getTime()
+  return Math.floor(diffTime / MILLISECONDS_IN_ONE_DAY)
```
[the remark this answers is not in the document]
TCL735: It could be. Seems like a simple operation though. And I've only run across this once so far.

hoorayimhelping (approved):

PR #4605, opened 2022-05-18
fix: add missing events for node.js wizard and go wizard

### LINKED ISSUE #4584: firstMile: BUG - Missing events for whether users choose the node.js wizard and the GO wizard
We have an event for if the users click on the Python Wizard Icon. 

However, we do not have events for if the users click on the node.js Wizard Icon and the GO Icon 

Screen Shot 2022-05-16 at 12 25 56 PM

wdoconnell:
Closes  #4584

Eventing in onboarding was set up for users clicking the python wizard icon, but not for users clicking the node wizard or go wizard icons.

Fix adds eventing, and changes function name to be more semantic: e.g., `logPythonWizardClick` instead of `logPythonEvent`

Screen Shot 2022-05-18 at 12 28 45 PM

### DIFF
--- a/src/homepageExperience/containers/HomepageContainer.tsx
+++ b/src/homepageExperience/containers/HomepageContainer.tsx
@@ -54,7 +54,15 @@ export const HomepageContainer: FC = () => {
   const squareGridCardSize = '200px'
 
   // events handling
-  const logPythonEvent = () => {
+  const logGoWizardClick = () => {
+    event('firstMile.goWizard.clicked')
+  }
+
+  const logNodeWizardClick = () => {
+    event('firstMile.goWizard.clicked')
+  }
+
+  const logPythonWizardClick = () => {
     event('firstMile.pythonWizard.clicked')
   }
 
@@ -91,7 +99,7 @@ export const HomepageContainer: FC = () => {
                       <Link
                         to={pythonWizardLink}
                         style={linkStyle}
-                        onClick={logPythonEvent}
+                        onClick={logPythonWizardClick}
                       >
                         <div
                           className="homepage-wizard-language-tile"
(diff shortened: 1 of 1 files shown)


src/homepageExperience/containers/HomepageContainer.tsx
```diff
@@ -54,7 +54,15 @@ export const HomepageContainer: FC = () => {
   const squareGridCardSize = '200px'
 
   // events handling
-  const logPythonEvent = () => {
+  const logGoWizardClick = () => {
+    event('firstMile.goWizard.clicked')
+  }
+
+  const logNodeWizardClick = () => {
+    event('firstMile.goWizard.clicked')
```
ChitlangeSahas, line 62: looks good to me!  should this be `firstMile.goWizard.clicked`?

ChitlangeSahas (approved):

src/homepageExperience/containers/HomepageContainer.tsx
hoorayimhelping, line 62: copypasta! (it says `goWizard` not `nodejsWizard`). Also, make sure it's `nodejsWizard` for consistency - all the other event names are `nodejsWizard`, rather than `nodeWizard`)

hoorayimhelping (approved):
fix up that copy/paste mistake and send it!

src/homepageExperience/containers/HomepageContainer.tsx
wdoconnell: Thanks! Copypasta fixed.

wdoconnell: Made function names consistent.

wdoconnell:
@Luckette Asked for an additional event for the 'More' button to detect when the user isn't using the Go, Python, or NodeJS wizards -- added that in latest commit.

hoorayimhelping (approved):

PR #4607, opened 2022-05-18
fix: do not display provider if account type is PAYG

### LINKED ISSUE #4597: Bug: PAYG Org Settings - Stop displaying Billing Provider under "Provider"

Image

wdoconnell:
Closes #4597 

Organization settings page currently displays Provider information (zuora, aws, gcm, azure) when user has a pay-as-you-go account type. The valid strings for type `AccountType` are `'cancelled' | 'contract' | 'free' | 'pay_as_you_go'` (src/client/unityRoutes.ts).
168848872-c3b39112-eb6c-42b9-b6a8-e045008cfa94

The fix adjusts conditional render of Provider component (`<LabeledData label="Provider" />`), so that the component does not render if quartz identifies the user's `accountType` as `pay_as_you_go'.

Screen Shot 2022-05-18 at 1 42 38 PM

### DIFF
--- a/src/organizations/components/OrgProfileTab/index.tsx
+++ b/src/organizations/components/OrgProfileTab/index.tsx
@@ -85,9 +85,10 @@ const OrgProfileTab: FC = () => {
             stretchToFitWidth={true}
             style={{width: '85%'}}
           >
-            {me.quartzMe?.billingProvider && (
+            {me.quartzMe?.billingProvider &&
+            me.quartzMe?.accountType !== 'pay_as_you_go' ? (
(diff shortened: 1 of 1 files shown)


ChitlangeSahas (approved):

wdoconnell:
Further updated this based on discussion with @Luckette.

wdoconnell:
#4720 is being deployed as the fix for this issue, as it uses the new endpoint, rather than the `billing_provider` field, which is not the correct source of the cloud provider information.

wdoconnell:
Closing this PR because it will be fixed by a separate ticket utilizing the data from `state.identity` that is pulled from `/quartz/identity`.

Issue #4313, opened 2022-04-04
250 Credit: Measure Success - Setup Eventing


Luckette:
**Measure Success - Setup Eventing**  

  ACs
   - Track # users that upgrade via upgrade now button per tab
   - Track # users that upgrade via billing page
   - Track # users that upgrade via the query-write limit notification (If they click via this, it should supersede the event for upgrade now button per tab)
   - Track # users that cancel on the upgrade during checkout
   - Track # of users on the same account that can see both the experimental group and control group (date, time, email, user ID, account ID, browser, test variation vs control) - include this in the 90 days logging
   
  Dev Notes
    - Ask eCommerce to create an app event that can be picked up downstream
    - Dumplings can use Amplitude and/or base logging 
    - Ask Jessica if we can add the eCommerce page to Amplitude

## Design

[Link to Prototype in Figma](https://www.figma.com/file/dkSQvZquqfoXd5iiDfBQpY/Checkout-%26-Post-Checkout?node-id=71%3A894)

ALuckette:
Related to: https://github.com/influxdata/Analytics/issues/282

TCL735:
`billing.conversion.payg.credit-250.upgrade` - the upgrade button in the body of the Billing page. This is the bottom button.

`checkout.cancel.credit-250.upgrade` - the cancel button on the checkout page.

`${resourceName}.alert.limit.credit-250.upgrade` - when a resource is exceed, this upgrade button appears related to that resource.

`${assetName}.overlay.limit.credit-250.upgrade` - when an action performed exceeds a resource, an overlay appears to warn the user that the resource is exceeded, with a button to upgrade.

`${location}.alert.credit-250.upgrade` - the "Upgrade Now" button at the top right of the screen. The location is where the user is in the navigation when the button is pressed.

`user.limits.${type}.credit-250.upgrade` - when series cardinality or query write is exceeded, this upgrade button appears.

`help-bar.overlay.free-account.credit-250.upgrade` - the button in the overlay that appears when the user selects Contact Support in the help navigation.

`subscriptions.parsing-form.credit-250.upgrade` - the button within the body of the form for the Subscriptions page. **Note: this button is hidden by default**. Instead, the global header has the same upgrade button, although the even for that button might be slightly different. See one of the above events.

PR #4610, opened 2022-05-18
fix: force value to string before calling trim

TCL735:
Part of #4313 

Fixes the `a.trim is not a function` by guaranteeing the value is a string before calling `.trim()`

Screen Shot 2022-05-18 at 2 16 15 PM


src/cloud/utils/reporting.ts
```diff
@@ -70,7 +70,7 @@ export const cleanTags = (data: Point): Point => {
       }
 
       // if it's made it this far, it's a string, cast it explicitly so typescript will stfu
```
hoorayimhelping, line 72: this comment is lying, let's remove it.

hoorayimhelping (approved):

Issue #4113, opened 2022-03-15
Bulk Action (API Tokens Bulk Delete): Bulk Delete


Luckette:
# User Story
As a user, I want to be able to bulk delete API tokens so that I can save time deleting API tokens and work more efficiently. 
    
    Acceptance Criteria
    - The API Tokens page has a Global Checkbox component
    - The API Tokens page has a Checkbox component on each Token component
    - The user should be able to select all via the Global Checkbox component
    - The user should be able to select 1 via the individual checkbox component per token
    - Once at least one token is selected, show the "Delete Selected" Button
    - After the user clicks the "Delete Selected" Button, display a hover box asking "Are you sure you want to delete?"
    - If all items are de-selected, then the "Delete Selected" button should disappear 

    Dev Notes
    - See Figma Prototype: https://www.figma.com/file/KvIPhGVlAoG5CCBB7LFema/Misc-Mocks?node-id=0%3A1

Luckette:
Designers: @taramk @brandenTenbrink

PR #4616, opened 2022-05-18
feat: bulk action delete for tokens

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4113

Video:

https://user-images.githubusercontent.com/18511823/169667107-c645ff9e-863f-4c2f-b2ed-13303c744fdd.mov


src/authorizations/components/TokenList.tsx
```diff
@@ -45,6 +51,7 @@ export default class TokenList extends PureComponent<Props, State> {
   public currentPage: number = 1
   public rowsPerPage: number = 10
   public totalPages: number
+  private enableBulkActionDelete = isFlagEnabled('bulkActionDelete')
```
asalem1, line 54: What’s the point of having this be a property of the class? Why not just use the isFlagEnabled instead?

src/authorizations/components/TokenRow.tsx
```diff
@@ -55,18 +57,22 @@ const formatter = createDateTimeFormatter(UPDATED_AT_TIME_FORMAT)
 class TokensRow extends PureComponent<Props> {
   public render() {
     const {description} = this.props.auth
-    const {auth} = this.props
+    const {auth, tokenIsSelected, onCheckboxClick} = this.props
     const date = new Date(auth.createdAt)
+    const enableBulkActionDelete = isFlagEnabled('bulkActionDelete')
```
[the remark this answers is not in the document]
ChitlangeSahas: I thought it'd save us calls to the function but I agree, not needed. Changed to using the function call instead. 👍

src/authorizations/actions/thunks.ts
```diff
@@ -168,6 +168,30 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
   }
 }
 
+export const bulkDeleteAuthorizations = (tokenIds: string[]) => async (
+  dispatch: Dispatch<Action | NotificationAction>
+) => {
+  try {
+    for (const id of tokenIds) {
+      try {
+        const resp = await api.deleteAuthorization({authID: id})
```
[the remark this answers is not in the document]
asalem1: Totally, I'm not saying it won't work. The implementation doesn't take advantage of concurrent requests that browsers allow, hence one of the reasons behind the [no-await-in-loop](https://eslint.org/docs/rules/no-await-in-loop) lint rule. 

As an aside, doing this could trigger a heap of dispatches that can cause slowiness in the process. This same iteration & dispatch combo was previously used in the dashboard list page. One thing that could be done to improve performance here would be to reduce the number of dispatches to one by pushing in the successful IDs into an array, and dispatching a new event that would handle removing an array of values rather than dispatching multiple events.

src/authorizations/actions/thunks.ts
```diff
@@ -168,6 +168,30 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
...
+
+        if (resp.status !== 204) {
+          throw new Error(resp.data.message)
+        }
+        event('token.delete.success', {id})
+        dispatch(removeAuthorization(id))
+      } catch (e) {
+        event('token.delete.failure', {id})
```
asalem1, line 185: what's the value added here for including the ID in the event? This type of pattern creates a unique value for every event, id event. I don't see the point, is this a product ask or is simply knowing that a token has been deleted valuable enough?

ChitlangeSahas: This is true, that was a copy paste error / leftover. I have it fixed and will push soon!

src/authorizations/components/TokenRow.tsx
```diff
@@ -41,10 +40,13 @@ import {
 import {relativeTimestampFormatter} from 'src/shared/utils/relativeTimestampFormatter'
 import {setCloneName} from 'src/utils/naming'
 import {event} from 'src/cloud/utils/reporting'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 interface OwnProps {
   auth: Authorization
   onClickDescription: (authID: string) => void
+  tokenIsSelected?: boolean
+  onCheckboxClick?: (token: Authorization) => void
```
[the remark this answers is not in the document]
ChitlangeSahas: I renamed it to `onSelectForBulkAction` so we can accommodate for future bulk actions like labelling or anything else we decide we want to do.

src/authorizations/components/TokensTab.tsx
```diff
@@ -101,25 +123,61 @@ class TokensTab extends PureComponent<Props, State> {
...
-          onSelect={this.handleSort}
-          width={238}
-        />
+        {isFlagEnabled('bulkActionDeleteTokens') && (
+          <Toggle
+            type={InputToggleType.Checkbox}
+            checked={tokensSelectedForBatchOperation.length > 0}
+            style={{
+              marginLeft: '24px',
+              marginRight: '24px',
```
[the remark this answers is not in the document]
ChitlangeSahas: Fixed!

src/authorizations/actions/thunks.ts
```diff
@@ -168,6 +168,30 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
...
+        console.error(e)
+      }
+    }
+    dispatch(notify(authorizationDeleteSuccess()))
+  } catch (e) {
+    dispatch(notify(authorizationDeleteFailed('Bulk')))
```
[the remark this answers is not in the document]
ChitlangeSahas: Updated with the newer `Promise.allSettled` approach 👍

ChitlangeSahas:
https://user-images.githubusercontent.com/18511823/169667107-c645ff9e-863f-4c2f-b2ed-13303c744fdd.mov

src/authorizations/actions/thunks.ts
```diff
@@ -168,6 +170,35 @@ export const deleteAuthorization = (id: string, name: string = '') => async (
...
+export const bulkDeleteAuthorizations = (tokenIds: string[]) => async (
+  dispatch: Dispatch<Action | NotificationAction>
+) => {
+  const deletePromises = tokenIds.map(async (tokenId: string) => {
+    const resp = await api.deleteAuthorization({authID: tokenId})
+
+    if (resp.status !== 204) {
+      return Promise.reject(new Error(resp.data.message))
+    }
+    dispatch(removeAuthorization(tokenId))
```
asalem1, line 182: since we're already mapping through and creating a promise, it would be great to optimize this to reduce the number of dispatches when this is called from N to 1. We can do this by simply returning the `tokenId` here, then dispatching the `results` below (i.e. all the tokenIds that have been successfully deleted and all the promises settled). This would require a new action to `removeAuthorizations` (plural) that basically does the same thing a the `removeAuthorization`, except for an array of ids rather than 1 id. The benefit here is that we reduce the number of render cycles and make this cool new feature more reactive.

hoorayimhelping: I don't think it's necessary to change redux to do a batch operation at this point. Users performing an operation on a list of items are going to have different expectations of the page's behavior than when they navigate to a page displaying resources. This is an experimental feature that we're testing the waters on (and trying to get out quickly and with minimal changes) and I don't want to add redux operations until we're sure this will be used throughout the app and we have a better idea of how the bulk actions should behave.

Technically this function does return a promise, but it doesn't include all the request information the original does, just the resolution status of `resolved` or `rejected`. Academically I think being explicit and returning the promise is the 'correct' course of action, but pramgatically _as it stands now_, it doesn't matter. The value isn't being used by the consuming function (`Promise.allSettled`'s function) and since the thunk dispatches a redux evenut and the UI doesn't use the promise, it's not necessary.

I think it's a best practice to return `resp` so that if we change how it's used in the future, we're not left scratching our heads wondering why it's not working the way we expect.

src/authorizations/components/TokensTab.tsx
```diff
@@ -191,6 +271,59 @@ class TokensTab extends PureComponent<Props, State> {
...
+
+  private getGlobalBatchSelectionToggleIcon = () => {
+    const {tokensOnCurrentPage, tokensSelectedForBatchOperation} = this.state
+
+    if (tokensSelectedForBatchOperation.length < tokensOnCurrentPage.length) {
+      return IconFont.Subtract
+    }
+    if (tokensSelectedForBatchOperation.length === tokensOnCurrentPage.length) {
+      return IconFont.Checkmark
+    }
```
hoorayimhelping, line 312: nice touch on the ui here 👍

hoorayimhelping (approved):
A couple of small suggestions, mainly around the language of the notifications, but I think this is ready to try out! Good work!

src/authorizations/actions/thunks.ts
asalem1: Hmmm, ok. If you’re ok with that then I’m not gonna block it. If this is experimental, should we be feature flagging it? Also, we should we write up an issue to address optimizations so that we don’t end up scratching our heads if the bulk actions ever grows in scope and we’re left with performance issues?

ChitlangeSahas: We already have the feature flag! `bulkActionDeleteTokens` used in the PR! I can write up an issue for the future!

src/shared/copy/notifications/categories/auth-and-tokens.ts
```diff
@@ -51,6 +51,18 @@ export const authorizationDeleteFailed = (desc: string): Notification => ({
...
+export const bulkAuthorizationDeleteSuccess = (
+  numberOfTokens: number
+): Notification => ({
+  ...defaultSuccessNotification,
+  message: `${numberOfTokens} API tokens were deleted successfully`,
+})
+
+export const bulkAuthorizationDeleteFailed = (desc: string): Notification => ({
+  ...defaultErrorNotification,
+  message: `Failed to delete API tokens: "${desc}"`,
```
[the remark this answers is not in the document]
ChitlangeSahas: Fixed this! I'll run this by Amy too.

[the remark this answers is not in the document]
AmyLuckette: Yeah that sounds much better. Go with Bucky's idea.

PR #4625, opened 2022-05-19
fix: queryBuilder tag key & value gets

asalem1:
This is just me cleaning up a sloppy copy job on my part and not bothering to look at the root of what the existing query was doing. The solution here was a few things:

1. not prepending a `(r) => ` onto the predicate since the tagFilters already does that through a function
2. Changing the `tagKeys` to `tagValues` for the values 🤦🏽 
3. Adding in the selected tag `key` for the tagValues

Visual proof here is that the query doesn't error out:

query-builder-fix

### DIFF
--- a/src/timeMachine/apis/queryBuilder.ts
+++ b/src/timeMachine/apis/queryBuilder.ts
@@ -75,7 +75,7 @@ import "influxdata/influxdb/schema"
 
 schema.tagKeys(
   bucket: "${bucket}",
-  predicate: (r) => ${tagFilters},
+  predicate: ${tagFilters},
   start: ${CACHING_REQUIRED_START_DATE},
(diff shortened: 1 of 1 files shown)


appletreeisyellow (approved):
Nice clean up!

> Changing the `tagKeys` to `tagValues` for the values

I just ran into this this afternoon that I supposed to copy `tagValues` but copied `tagKeys`... I wish I could see this PR earlier!

asalem1:
> Nice clean up!
> 
> > Changing the `tagKeys` to `tagValues` for the values
> 
> I just ran into this this afternoon that I supposed to copy `tagValues` but copied `tagKeys`... I wish I could see this PR earlier!

Ugh, so sorry to hear! Wish I sent it up well to begin with

PR #4626, opened 2022-05-19
feat: convert account_created_at to epoch time

TCL735:
Part of #4313 

Change `account_created_at` to epoch time, because it is easier to manage as a number in the Google Optimize experiment dashboard.

Screen Shot 2022-05-19 at 2 37 46 PM

### DIFF
--- a/src/shared/containers/GetOrganizations.tsx
+++ b/src/shared/containers/GetOrganizations.tsx
@@ -30,6 +30,7 @@ import {CLOUD} from 'src/shared/constants'
 
 // Utils
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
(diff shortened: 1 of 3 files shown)


asalem1 (approved):

src/shared/containers/GetOrganizations.tsx
```diff
@@ -81,7 +82,7 @@ const GetOrganizations: FunctionComponent = () => {
       window.dataLayer.push({
         identity: {
           account_type,
-          account_created_at,
+          account_created_at: convertStringToEpoch(account_created_at),
```
wdoconnell, line 85: Is there anything we'd have to change in the Google Optimize dashboard since the data type is changing to `number`?

TCL735: Good question but nope.

PR #4628, opened 2022-05-20
fix: make eventing for firstMile buttons identify stage it was clicked at

### LINKED ISSUE #4608: firstMile: BUG - All eventing for the previous and next buttons needs to specify the page on which it was clicked, not just the language.
**Problem:** 
As of 05/18/2022, we are missing events for specifically which page users clicked the previous and next buttons. Knowing that they're clicked in a wizard in general is only minimally helpful. We also need to know specifically which step that navigated back to and forward to. 

**ACs**
- All eventing for the previous and next buttons needs to specify the page on which it was clicked, not just the language.

wdoconnell:
Closes #4608

Next/Previous buttons on Go, NodeJS, and Python wizards were calling `event` but only passing it a `measurement` (firstMile.wizardName.next.clicked'), but no identifying `meta` (tags) or `values`, preventing identifying stage at which the user decided to move forward, stop, or go back.

Added two values identifying the stage at which the button was clicked (and where user is now).

Values
--
   `clickedButtonAtStep`: stage of the wizard at which the user clicked the button
   `currentStep`: stage of the wizard the user is at now

@Luckette Do these value names look good to you?

EDIT: Removed tag names as unnecessary.

https://user-images.githubusercontent.com/91283923/169563040-0b416be2-7b8e-4556-9fd8-bf517a4f1235.mov


src/homepageExperience/containers/GoWizard.tsx
```diff
@@ -70,7 +70,14 @@ export class GoWizard extends PureComponent<null, State> {
         ),
       },
       () => {
-        event('firstMile.goWizard.next.clicked')
+        event(
+          'firstMile.goWizard.next.clicked',
+          {clickType: 'next'},
```
ChitlangeSahas, line 75: Do we need to pass `clickType`? given that it is part of `'firstMile.goWizard.next.clicked'` 

Rest looks good!

wdoconnell: Since it's param 2 out of 3 we'll need to pass something now, but the default value in the code is just an empty `object`, so we could do that if it's not needed!

hoorayimhelping: >Since it's param 2 out of 3 we'll need to pass something now

it's okay to send an empty object to the second argument. [Example here](https://github.com/influxdata/ui/blob/eb134fded2a37b637d7c39f8067d966cc6350c3d/src/timeMachine/actions/queries.ts#L277)

```ts
event(
  'firstMile.goWizard.next.clicked',
  {},
  {
    clickedButtonAtStep: this.state.currentStep - 1,
    currentStep: this.state.currentStep,
  }
)
```

wdoconnell: Per above, removed the tags, and am now passing in an empty object. One thing I noticed when re-testing: even if the `streamEvents` flag is turned on, it won't console.log event values unless the event has a tag. Fixed the `if` blocks so that streamEvents still provides console.logs if the event has values but no tag.

Let's wait on @Luckette to make sure this format looks good to her.

https://user-images.githubusercontent.com/91283923/169574222-89c529b1-f8ff-4437-9b28-653e131458e1.mov

hoorayimhelping (approved):

wdoconnell:
Removed changes to logging in src/cloud/utils/reporting.ts @hoorayimhelping

PR #4629, opened 2022-05-20
feat(4615): generalize editor context

### LINKED ISSUE #4615: Editor: generalize the editor.context for use across the application.
In notebooks, we use the editor context to manage code injection (e.g. inject functions, variables, etc).

As we have been making updates to timeMachine, we have begun extracting out helper functions from the editor context --> in order to provide the same functionality to timeMachine (outside of notebooks).
* we did this for faster bug fixing
* But these helpers are very tightly coupled to the editor context.

We should be able to generalize the editor context rather easily. Such that it can be used in notebooks, timeMachine, and anywhere else.  (Removal of tech debt, while quickly enabling new editor features/debugging across all the editors in the application.)

wiedld:
Closes #4615 

## Goal:
* make editor context the go-to state manager for the monaco-editor 
  * used anywhere we use the monaco-editor, with incoming affordances from UI interactions
  * e.g. notebooks, timeMachine
* what this is not:
  * does not manage editor code injection from other sources (e.g. the LSP)
* what this is: 
  * for injections triggered from the UI components outside the editor
  * or other (non-injection) updates pushed from the UI, into the editor
* specific state management options, currently offered by the editor context (can be expanded too):
  * injection
  * injection of variables (keeping the same rules across the app)
  * injection of functions (keeping the same rules across the app)


src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -172,6 +145,7 @@ const Query: FC<PipeProp> = ({Context}) => {
       updateText,
       editorContext.editor,
       variables,
+      launcher,
```
wiedld, line 148: Was having a memoize issue for the launcher not grabbing the latest `showId`. Hence, these minor changes for the launcher.

src/shared/contexts/editor/injection.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (shouldEndInNewLine) {
+    columnOffset++
+  }
+  setTimeout(() => {
+    editor.focus()
+    editor.setPosition({
+      lineNumber: hasHeader ? row + 1 : row,
+      column: column + textLength - columnOffset,
+    })
+    editor.trigger('', 'editor.action.triggerSuggest', {})
```
wiedld, line 100: For future reference (when we do our injection hackathon next week):
this editor trigger event is a way that we kick the editor to communicate with the LSP.

Depending on what approaches people try next week, they may also want to trigger editor actions.

src/shared/contexts/editor/injection.ts
```diff
@@ -0,0 +1,102 @@
+export enum InjectionType {
```
wiedld, line 1: This entire file was just moved functions. No change in the code used.

src/shared/contexts/editor/injection.ts
wiedld: I should clarify.

Every time we type into the monaco-editor, the editor chats with the LSP.

When we directly have the UI do things to the editor (which is not typing the in editor window)...the editor triggers some (but not all) communication with the LSP. Hence when using these editor triggers are useful.

src/shared/contexts/editor/insertFunction.ts
```diff
@@ -24,11 +18,11 @@ export const generateImport = (
   }
 
   if (
-    !funcPackage ||
-    funcPackage === 'universe' ||
+    !func.package ||
+    func.package === 'universe' ||
     script.includes(importStatement)
   ) {
-    return false
```
asalem1, line 31: Why the change here?

src/shared/contexts/editor/injection.ts
asalem1: Should we / do we debounce this process?

src/shared/contexts/editor/insertFunction.ts
wiedld: This method is now combining two separate (but functionally equivalent) ways to generate the import statement. One approach produced a string or null, whereas the other produced a string or false. 

It's cleaner to use the former option, because typescript accepts a returned null as unifying with a string. (Versus having to provide a `string | boolean` type.) And we do a null-pointer check downstream anyways.

src/shared/contexts/editor/injection.ts
wiedld: This manually triggering we are doing here, is in the scenario when we manually insert a block of text (e.g. a function injection)...but then at the end we once (and only once) kick the editor to act like we have been typing in text.

As such, it should not need to be debounced.

asalem1 (approved):
It would be nice not to have to prop drill more into the secret list but that would require overhauling the codebase 😅

PR #4630, opened 2022-05-20
chore: fix the loading state for the query builder

asalem1:
This PR addresses inconsistencies with the loading state for the newQueryBuilder forced measurement selector. Whereas previously the filter / group field was available while the measurements were loading, this implementation aligns more closely with the product intent by limiting the user manipulation of queries within the query builder experience by removing that option for users. Effectively this looks like:

loading-de

### DIFF
--- a/src/timeMachine/components/TagSelector.tsx
+++ b/src/timeMachine/components/TagSelector.tsx
@@ -78,12 +78,7 @@ class TagSelector extends PureComponent<Props> {
   }
 
   public render() {
-    return (
-      <BuilderCard>
-        {this.header}
-        {this.body}
-      </BuilderCard>
-    )
+    return <BuilderCard>{this.body}</BuilderCard>
   }
 
   private get isCompliant() {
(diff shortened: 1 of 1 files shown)


src/timeMachine/components/TagSelector.tsx
```diff
@@ -97,29 +92,12 @@ class TagSelector extends PureComponent<Props> {
     return (
       isFlagEnabled('newQueryBuilder') &&
       index === 0 &&
-      selectedKey === '_measurement' &&
+      (selectedKey === '' || selectedKey === '_measurement') &&
```
asalem1, line 95: The core of what this is doing here is determining whether the first tag selector (index === 0) is the empty or the measurement tag. The reason being that the measurement tag isn't populate as the selected field until the data is returned from the API. The problem this introduces for us is that when a user is loading data from a bucket with a lot of data, the loading state ends up in this weird in-between

src/timeMachine/components/TagSelector.tsx
```diff
@@ -148,39 +126,60 @@ class TagSelector extends PureComponent<Props> {
...
-                    testID="tag-selector--dropdown"
-                    buttonTestID="tag-selector--dropdown-button"
-                    menuTestID="tag-selector--dropdown-menu"
-                    options={keys}
-                  />
-                )}
-              </ErrorBoundary>
-              {this.selectedCounter}
-            </FlexBox>
+          {!this.isCompliant && (
```
asalem1, line 154: made this align with notebooks so that we have consistent UI across the app. Hopefully this can be better implemented with the new query builder experience by reusing component architecture

appletreeisyellow (approved):
Nice clean up! Thanks for making the UI across notebook and data explorer consistent!

PR #4640, opened 2022-05-23
fix: Add missing eventing to firstMile video

### LINKED ISSUE #4623: firstMile: BUG - mIssing eventing for all three languages on whether or not the time series video was clicked
As a PM and Data Analyst, I want to know when users click to watch the Time Series video. 

Image

wdoconnell:
Closes #4623

Eventing was not present for the tutorial 'what is time series' video in the Node, Go, and Python wizards. 

`<iframe>` can't accept onClick or similar events [because iframe represents a new nested browsing context](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe). This fix is adapted from solution found [here](https://github.com/springload/react-iframe-click), which infers the click from an onBlur event in `window` when the iframe is the then-active DOM element.

This fix fires the event on every click where the user changes context to the `<iframe>`. If this behavior is only desired on first click (e.g., shouldn't refire if user pauses, clicks out, then later clicks back in), we should remove the event listener after the first fire.

https://user-images.githubusercontent.com/91283923/169846461-c58abadb-46ec-49b2-8a2c-9327b02ba7d0.mov

### DIFF
--- a/src/homepageExperience/components/steps/Overview.tsx
+++ b/src/homepageExperience/components/steps/Overview.tsx
@@ -9,7 +9,7 @@ type Props = {
 export const Overview: FC<Props> = ({wizard}) => {
   const videoFrame = useRef<null | HTMLIFrameElement>(null)
 
-  // onClick (and related events) are unavailable for media embedded in iframes because they represented a nested browsing context. This solution, which infers a click from an onBlur event where the iframe is the active element on the DOM, is adapted from https://github.com/springload/react-iframe-click
+  // iFrames represent separate browsing context, requiring this workaround. For more, see resolution of this ticket (https://github.com/influxdata/ui/issues/4623)
 
   useEffect(() => {
     const checkVidClick = () => {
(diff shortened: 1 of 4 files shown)


src/homepageExperience/components/steps/Overview.tsx
```diff
@@ -1,39 +1,59 @@
...
-            In the next 5 minutes, you will set up InfluxData on your machine
-            and write and execute some basic queries.
-          </p>
+  useEffect(() => {
+    const checkVidClick = () => {
+      if (
+        document.activeElement &&
+        document.activeElement.nodeName.toLowerCase() === 'iframe' &&
+        videoFrame.current &&
+        videoFrame.current === document.activeElement
```
hoorayimhelping, line 20: Gold star for checking on existence before using it! ⭐

But we have access to [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) so this can be shortened to

```ts
document?.activeElement?.nodeName?.toLowerCase() === 'iframe'
```
and
```ts
videoFrame?.current === document.activeElement
```

hoorayimhelping (approved):

wdoconnell:
Per @Luckette, decision is to generate a new click if the user clicks out of the video and clicks back in (not just the first time it's clicked when visiting the page). Event should be generated "every time - Data analysts will be able to filter for unique clicks". Thanks to @alexjw3 for reminding us to document this.

PR #4642, opened 2022-05-23
fix: regex for timestamp

helenosheaa:
Closes: https://github.com/influxdata/data-acquisition/issues/320
Related: https://github.com/influxdata/data-acquisition/issues/322
Related: [#14547](https://github.com/influxdata/idpe/pull/14547)

Add timestamp to the util func for string parsing. Add all string measurement ,fields ,tags and timestamps to the util func for update parsing.

We were seeing inconsistencies in the way the regex parsing was working, plus it adding a timestamp property when it was the path was not given by the user - this resolves this.


cannonpalms (commented):
Do these `.replace(...)`s do anything? 

```ts
f.pattern = f.pattern?.replace(/\\\\/g, '\\')
```

It's replacing all occurrences of `\\` with `\\`, no? What am I missing?

helenosheaa:
when a user types in \\ in a regex path react form decides to escapes them and puts \\\\, this then breaks our nifi parsing

cannonpalms:
Gotcha, so you want to replace four slashes with 2? (`\\\\` -> `\\`) The current `replace(...)` code is replacing 2 slashes with 2 slashes (`\\` -> `\\`)

helenosheaa:
without my replace regex:
```
stringMeasurement:
name: "measurement"
pattern: "a=(\\\\d)"
```

with:
```
stringMeasurement:
name: "measurement"
pattern: "a=(\\d)"
```

which becomes in nifi extract text processor:
`a=(\d)`

cannonpalms:
Can't argue with that. I'm getting something wrong, clearly. I'm going to go find out what it is.

cannonpalms (approved):

PR #4648, opened 2022-05-23
feat(HelpBar): add confirmation overlay 

blegesse-git:
Closes [#3465](https://github.com/influxdata/ui/issues/3465) 
Closes [#3466](https://github.com/influxdata/ui/issues/3466)

Pr adds a shared confirmation overlay for both payg support and feedback and questions. 

https://user-images.githubusercontent.com/66275100/170743388-f0318097-6508-4372-984a-ea6b23f56295.mov


src/support/components/FeedbackQuestionsOverlay.tsx
```diff
@@ -114,4 +123,11 @@ const FeedbackQuestionsOverlay: FC<OwnProps> = () => {
...
 
-export default FeedbackQuestionsOverlay
+const mdtp = {
+  showOverlay,
+  dismissOverlay,
+}
+
+const connector = connect(null, mdtp)
+
+export default connector(FeedbackQuestionsOverlay)
```
hoorayimhelping, line 133: we can skip all this by using the `useDispatch` hook from `react-redux`. 

[here's](https://github.com/influxdata/ui/blob/master/src/cloud/components/RateLimitAlert.tsx#L68) an example of `useDispatch` being used in combination with `showOverlay`

src/support/components/FeedbackQuestionsOverlay.tsx
```diff
@@ -28,20 +31,26 @@ import {event} from 'src/cloud/utils/reporting'
...
   const {onClose} = useContext(OverlayContext)
   const {id: orgID} = useSelector(getOrg)
   const {id: meID} = useSelector(getMe)
   const [feedbackText, setFeedbackText] = useState('')
 
-  const handleSubmit = () => {
-    // handle form submit
+  const handleSubmit = (e): void => {
+    const {showOverlay} = props
+    e.preventDefault()
```
hoorayimhelping, line 47: I'd prefer we call this `evt` since `event` is already used in this scope. It's a little bit more descriptive than `e`  (which is commonly used to represent `event` or `error`) and is very handy when searching the whole codebase for errors or events.

src/support/components/PayGSupportOverlay.tsx
```diff
@@ -1,5 +1,5 @@
 import React, {FC, ChangeEvent, useContext, useState} from 'react'
-import {useSelector} from 'react-redux'
+import {connect, useSelector} from 'react-redux'
```
hoorayimhelping, line 2: same comments here about `e` vs `evt` and `connect` vs `useDispatch`

hoorayimhelping (changes requested):
Couple of small change requests

hoorayimhelping (approved):

PR #4649, opened 2022-05-23
feat(ecommerce): Add Cloud Provider Filters

abshierjoel:
This adds filter dropdowns (see below) to the operator UI  for filtering organizations by their cloud Provider and Regions. 

The OpenAPI specs to support this is being added in: influxdata/openapi#328

operator-ui-providers-filter

### DIFF
--- a/src/operator/ResourcesCloudProvider.tsx
+++ b/src/operator/ResourcesCloudProvider.tsx
@@ -1,0 +1,86 @@
+import {
+  ComponentColor,
+  DropdownMenuTheme,
+  MultiSelectDropdown,
+} from '@influxdata/clockface'
+import React, {FC, useContext} from 'react'
+import {OperatorContext} from './context/operator'
+import {OperatorRegions} from 'src/types'
+
+const ResourcesCloudProvider: FC = () => {
+  const {
+    providerInfo,
+    providers,
+    regions,
+    setProviders,
+    setRegions,
+  } = useContext(OperatorContext)
+
+  const hasSelectedProvider = providers.length > 0
+
+  const availableRegions: OperatorRegions = providers
+    .map(p => getRegions(p, providerInfo.regions))
+    .flat()
+
+  const providerOptions = providerInfo.providers.map(p => p.provider)
+  const regionOptions = availableRegions.flat().map(r => r.region)
+
+  const handleSelectProvider = (selectedOption: string): void => {
+    setRegions([])
+    providers.includes(selectedOption)
+      ? setProviders(prev => prev.filter(x => x !== selectedOption))
+      : setProviders(prev => [selectedOption, ...prev])
+  }
+
+  const handleSelectRegion = (selectedOption: string): void =>
+    regions.includes(selectedOption)
+      ? setRegions(prev => prev.filter(x => x !== selectedOption))
+      : setRegions(prev => [selectedOption, ...prev])
+
+  return (
+    <>
+      <MultiSelectDropdown
+        style={{width: '220px', marginRight: '10px'}}
+        options={providerOptions}
+        selectedOptions={providers}
+        onSelect={handleSelectProvider}
+        menuTheme={DropdownMenuTheme.Onyx}
+        emptyText="Cloud Provider"
+        buttonColor={ComponentColor.Primary}
+      />
+      {hasSelectedProvider && (
+        <MultiSelectDropdown
+          style={{width: '220px', marginRight: '20px'}}
+          options={regionOptions}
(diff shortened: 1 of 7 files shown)


src/operator/ResourcesCloudProvider.tsx
```diff
@@ -0,0 +1,86 @@
+import {
+  ComponentColor,
+  DropdownMenuTheme,
+  MultiSelectDropdown,
+} from '@influxdata/clockface'
+import React, {FC, useContext} from 'react'
+import {OperatorContext} from './context/operator'
+import {OperatorRegions} from 'src/types'
+
+const ResourcesCloudProvider: FC = () => {
```
hoorayimhelping, line 10: I think the name of this should change, since in the context system we use, *Provider refers to a [context provider](https://reactjs.org/docs/context.html#contextprovider). Usually something with the postfix of `Provider` wraps components that consume a react context.

src/operator/ResourcesCloudProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const handleSelectRegion = (selectedOption: string): void =>
+    regions.includes(selectedOption)
+      ? setRegions(prev => prev.filter(x => x !== selectedOption))
+      : setRegions(prev => [selectedOption, ...prev])
+
+  return (
+    <>
+      <MultiSelectDropdown
+        style={{width: '220px', marginRight: '10px'}}
```
hoorayimhelping, line 43: Is this component going to be rendered on every operator page? If so, I think this style object literal should be set into a variable, since object literals passed as props to components cause unnecessary renders. If not, it's probably not a huge deal, but it's a nice-to-have.

https://www.debugbear.com/blog/react-rerenders#passing-objects-as-props

src/operator/ResourcesTable.tsx
```diff
@@ -29,9 +29,10 @@ import {OperatorRoutes} from 'src/operator/constants'
 // Types
 import {OperatorOrg, OperatorAccount} from 'src/types'
 import ResourcesAccountType from './ResourcesAccountType'
+import ResourcesCloudProvider from './ResourcesCloudProvider'
 
 const ResourcesTable: FC = () => {
-  const {pathname, accounts, organizations, status} = useContext(
+  const {accounts, organizations, pathname, status} = useContext(
```
hoorayimhelping, line 35: 👍

hoorayimhelping (changes requested):

src/operator/ResourcesCloudProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+  } = useContext(OperatorContext)
+
+  const hasSelectedProvider = providers.length > 0
+
+  const availableRegions: OperatorRegions = providers
+    .map(p => getRegions(p, providerInfo.regions))
+    .flat()
+
+  const providerOptions = providerInfo.providers.map(p => p.provider)
+  const regionOptions = availableRegions.flat().map(r => r.region)
```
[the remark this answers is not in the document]
abshierjoel: Thank you! Yeah, the 2nd `flat()` was definitely unnecessary. I changed this now to a `flatMap()` followed by a `map()` to retrieve the titles. That'll eliminate the intermediate variable.

src/operator/ResourcesCloudProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const handleSelectProvider = (selectedOption: string): void => {
+    setRegions([])
+    providers.includes(selectedOption)
+      ? setProviders(prev => prev.filter(x => x !== selectedOption))
```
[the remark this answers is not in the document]
abshierjoel: Good call! I changed those to their respective names.

src/operator/ResourcesCloudProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+    case 'Azure':
+      return regions.Azure
+    case 'AWS':
+      return regions.AWS
+    case 'GCP':
+      return regions.GCP
+    default:
+      return []
+  }
+}
```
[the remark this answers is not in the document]
abshierjoel: Moved it! Thanks!

src/operator/ResourcesCloudProvider.tsx
abshierjoel: Ah yeah, I didn't think about the context providers. I went ahead and changed this to `CloudClusters` instead.

hoorayimhelping (approved):

ChitlangeSahas (approved):

PR #4650, opened 2022-05-23
feat: add basic schema browser in new data explorer

### LINKED ISSUE #4510: Basic Schema Browser
## User Story

<p></p>

As a user, I want to be able to see the buckets, measurements, fields, tag keys, and tag values that are available to me to query so that I understand the shape of my data and I know what I can use in my query.

<p></p>

## Acceptance Criteria

<p></p>

- Schema Browser will show a list of buckets available to the user

- Schema Browser will allow one (and only one) bucket to be selected from the list

- Schema Browser, upon bucket selection, will show a list of measurements to the user; this will be a list not bound by the time range selection

- Schema Browser will allow the user to select one (and only one) measurement

- Schema Browser, upon measurement selection, will show a list of fields, and tag keys to the user

- Schema Browser will allow an expansion of Tag keys into a list of tag values

- Schema Browser will allow the one or more tag values to be selected

- The first 10 items (or less if less than 10 exist) in any list (buckets, measurements, fields, tag keys, tag values initially so that the user sees there are more types of things to choose from

- When there are more than 10 items in any list, then the user will be shown a "Load More" entry where they can choose to load more

- When Load More is chosen, up to 25 additional entries will be shown to the user; if more than 25 more exists in the list, the user will again have the option to load more which will result in an additional 25

- User will be able to "search" buckets which will then filter the list of buckets dynamically to match what has been entered into the search area (for instance, if "b" is entered - then the list will only show buckets that start with b" - if "bl" is entered, then only buckets that start with "bl" will be shown

- User will be able to search measurements, fields, tags, and tag values in the same fashion

- Schema browser will be shown within the "Try new Data Explorer" view

- Schema browser will be hold the far left side of the interface

- Schema browser will be vertically scrollable

- Schema browser will not be horizontally scrollable, but the size of the schema browser can be resized larger and smaller horizontally

- Schema browser will have a basic tree structure indentation to show a hierarchical view of their data:  Bucket at the top level, then measurements, then fields and tag keys, then tag values under tag keys

- A title will appear above the various lists/selection area to clearly show the user what those elements are:  Bucket, Measurements, Fields, and Tag Keys (tag values will appear only as the values themselves under the tag key - so no additional title necessary)

- For fields and tag values, a "select all" option will appear at the top of the list that can be selected

- For bucket, measurement, field, and and tag keys, and information button will be available next to the title that can be selected to provide an overlay description of the element and link to documentation for it (content to be defined separately, for this story - placeholder text can be used)

- Only one bucket can be selected and when selected, the selection is shown and the list is collapsed

- Only one measurement can be selected and when selected, the selection is shown and the list is collapsed

<p></p>

## Notes

<p></p>

The story does not include requirements for injection of the selections into a flux script; that will be done in a separate story

<p></p>

## UI Design

<p></p>

<p></p>

See the data selection section from this FIGMA mockup:

<p></p>

<https://www.figma.com/proto/0qAntPk5LVangAHguWT74X/Query-Experience-Project?page-id=191%3A13710&node-id=567%3A21115&viewport=241%2C48%2C0.15&scaling=min-zoom&starting-point-node-id=567%3A21115>

<p></p>

Note:  The selection checkboxes shown here are no longer required and don't need to be included:

<p></p>

<p></p>

<p></p>

<p></p>

<p></p>

<p></p>

<p></p>

<p></p>

appletreeisyellow:
Closes #4510

This PR adds the MVP version of the basic schema browser in new data explorer ✨ It is behind feature flag `newDataExplorer`.

## Before
Screen Shot 2022-05-23 at 5 05 26 PM

## After
https://user-images.githubusercontent.com/14298407/170143922-2a64610c-1faa-422a-baf6-6070d1f21071.mov


src/dataExplorer/components/BucketSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const [searchTerm, setSearchTerm] = useState('')
+
+  const handleSelectBucket = (option: string) => {
+    const bucket = buckets.filter(b => b.name === option)
+    if (!bucket.length) {
+      // TODO: any error message?
+      return
+    }
+    selectBucket(bucket[0])
+  }
```
asalem1, line 24:
```suggestion
  const handleSelectBucket = (name: string) => {
    const bucket = buckets.find(b => b.name === name)
    if (!bucket) {
      // TODO: any error message?
      return
    }
    selectBucket(bucket)
  }
```

src/dataExplorer/components/FieldSelector.tsx
```diff
@@ -0,0 +1,95 @@
+import React, {FC, useContext, useEffect, useState} from 'react'
+import {concat, slice} from 'lodash'
```
asalem1, line 2:
```suggestion
```

src/dataExplorer/components/FieldSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [loadMore, setLoadMore] = useState(false)
+  const [index, setIndex] = useState(LOCAL_LIMIT)
+
+  useEffect(() => {
+    if (fields.length === 0) {
+      setFieldsToShow([])
+      setLoadMore(false)
+      setIndex(LOCAL_LIMIT)
+    } else {
+      const newFieldsToShow = slice(fields, 0, index)
```
asalem1, line 33:
```suggestion
      const newFieldsToShow = fields.slice(0, index)
```

src/dataExplorer/components/TagSelector.tsx
```diff
@@ -0,0 +1,172 @@
...
+  )
+
+  return (
+    <Accordion className="tag-selector-value">
+      <Accordion.AccordionHeader className="tag-selector-value--header">
+        <div onClick={() => onSelectTagKey(tagKey)}>
+          <SelectorTitle title={tagKey} />
+        </div>
+      </Accordion.AccordionHeader>
+      <div className="container-side-bar">
```
asalem1, line 104: based on the video it looks like the list text is a bit squished, any chance we can add a larger margin-bottom to this?

src/dataExplorer/components/FieldSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  const handleLoadMore = () => {
+    const newIndex = index + LOCAL_LIMIT
+    if (loadMore) {
+      // Add more field values to show
+      const newFieldsToShow = concat(
+        fieldsToShow,
+        slice(fields, index, newIndex)
+      )
```
asalem1, line 68: this should have the same effect and be a bit more simplified:

```suggestion
      const newFieldsToShow = fields.slice(0, newIndex)
```

IF we're never invalidating the original cropped data set, the start should always be the same and we're just pushing more data onto it right?

src/dataExplorer/components/FieldSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [fieldsToShow, setFieldsToShow] = useState([])
+  const [loadMore, setLoadMore] = useState(false)
+  const [index, setIndex] = useState(LOCAL_LIMIT)
+
+  useEffect(() => {
+    if (fields.length === 0) {
+      setFieldsToShow([])
+      setLoadMore(false)
+      setIndex(LOCAL_LIMIT)
+    } else {
```
asalem1, line 32: is this necessary? Can we just return if this is the case?

src/dataExplorer/components/FieldSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <SelectorTitle title="Fields" info="Test info" />
+      </Accordion.AccordionHeader>
+      <div className="container-side-bar">
+        {list}
+        {loadMoreButton}
+      </div>
+    </Accordion>
+  )
+}
+
```
asalem1, line 94: we should be able to simplify the logic of this component to remove the need for setting the `index` here. More specifically, it looks like `index` being used to determine the ending boundary of the cropped list. This _should_ also be derived by simply using the length of the `fieldsToShow`. In doing so, we reduce the complexity of this component and reduce potential drift between two sources of truth.

The same point goes for the `loadMore` logic. In theory, we should only want to hide the button when the `fieldsToShow.length !== fields.length` right?

src/dataExplorer/components/MeasurementSelector.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  const handleChangeSearchTerm = (value: string) => {
+    setSearchTerm(value)
+  }
+
+  return useMemo(() => {
+    if (isEmpty(selectedBucket)) {
+      return null
+    }
```
asalem1, line 32: preference for not using lodash methods here. If there's a way we can derive this natively that would be 🔥

src/dataExplorer/components/Schema.tsx
```diff
@@ -0,0 +1,95 @@
...
+    searchTerm,
+    setSearchTerm,
+  } = useContext(NewDataExplorerContext)
+
+  const handleSearchFieldsTags = (searchTerm: string): void => {
+    setSearchTerm(searchTerm)
+  }
+
+  return useMemo(() => {
+    if (isEmpty(selectedBucket) || !selectedMeasurement) {
```
asalem1, line 45: same comment as before re: lodash methods

src/dataExplorer/components/TagSelector.tsx
```diff
@@ -0,0 +1,172 @@
...
+    <Accordion className="tag-selector-key" expanded={true}>
+      <Accordion.AccordionHeader className="tag-selector-key--header">
+        <SelectorTitle title="Tag Keys" info="Test info" />
+      </Accordion.AccordionHeader>
+      <div className="container-side-bar">{list}</div>
+    </Accordion>
+  )
+}
+
+export default TagSelector
```
asalem1, line 172: similar points here as about the `FieldSelector` w/r/t reducing the usage of stateful logic here in favor of deriving properties

asalem1 (changes requested):
Overall really great work. There are a couple of things that would be nice to clean up prior to pushing this up. I understand it's a POC that will be iterated upon, but these things are easy to get cemented once they're in the codebase. Specific requests that would be cool to see cleaned up:

1. Using native methods vs `lodash`
2. Reducing state logic in favor of derived properties
3. Splitting the `newDataExplorer` context up
4. Writing up some issues for the TODOs so that we can share context and track what still needs to be done, and potentially work on that stuff in parallel rather than having it all on your plate.

src/dataExplorer/components/FieldSelector.tsx
appletreeisyellow: Yes! This is simpler!

src/dataExplorer/components/TagSelector.tsx
```diff
@@ -0,0 +1,172 @@
...
+      selectTagKey(key)
+    }
+  }
+
+  const handleSelectTagValue = (value: string) => {
+    // TODO
+    /* eslint-disable no-console */
+    console.log(value)
+    /* eslint-disable no-console */
+  }
```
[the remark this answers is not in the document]
appletreeisyellow: Eventually inject the tag value into the flux script editor. I think Gary, Adam, and the engineer team will brainstorm on this later this week to see how we want to show users when they select tag values and field values

src/dataExplorer/components/TagSelector.tsx
```diff
@@ -0,0 +1,172 @@
...
+  )
+
+  const tagKeys: string[] = Object.keys(tags)
+
+  const handleSelectTagKey = (key: string) => {
+    if (tags[key].length === 0) {
+      // Only need to get tag values if currently no values
+      selectTagKey(key)
+    }
+  }
```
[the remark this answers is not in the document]
appletreeisyellow: Yes, here is the product requirement from #4510 

> Schema Browser will allow the one or more tag values to be selected

I think we will handle multiple tag values in `handleSelectTagValue` function. The function `selectTagKey` is used to fetch tag values on demand. Sorry the function name is not explicit on what it does and it is confusing 😅 I changed the name so that it's clearer

src/dataExplorer/context/newDataExplorer.tsx
```diff
@@ -0,0 +1,442 @@
...
+import {Bucket, QueryScope, RemoteDataState} from 'src/types'
+
+// Utils
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
+const IMPORT_REGEXP = 'import "regexp"\n'
+const IMPORT_INFLUX_SCHEMA = 'import "influxdata/influxdb/schema"'
+const SAMPLE_DATA_SET = (bucketID: string) =>
+  `import "influxdata/influxdb/sample"\nsample.data(set: "${bucketID}")`
+const FROM_BUCKET = (bucketName: string) => `from(bucket: "${bucketName}")`
```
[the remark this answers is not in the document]
appletreeisyellow: Sorry it's confusing here 😅 I try to replicate the flux query from flow. Hopefully we can cleanup the query to make it more consistent in other PR

src/shared/components/SearchableDropdown.tsx
```diff
@@ -101,6 +104,7 @@ const SearchableDropdown: FC<Props> = ({
         >
           <div className="searchable-dropdown--input-container">
             <Input
+              icon={iconOn && IconFont.Search_New}
```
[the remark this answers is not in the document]
appletreeisyellow: Just talk to Julia and she will bring this up to the design team in their team meeting today

src/dataExplorer/components/FieldSelector.tsx
appletreeisyellow: Great suggestion! After re-writing the logic and eliminating unnecessary states, the code is simplified by 70% here!! 🚀

src/shared/components/SearchableDropdown.tsx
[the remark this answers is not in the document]
appletreeisyellow: Update - confirmed with Julia and the design team that we want this icon on and no need for the conditional logic!

src/dataExplorer/components/Schema.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return useMemo(() => {
+    if (!selectedBucket || !selectedMeasurement) {
+      return null
+    }
+
+    return (
+      <div className="container-side-bar">
+        <SearchWidget
+          placeholderText="Search fields and tags..."
```
asalem1, line 52: There’s a PR up to remove ellipses, did we want them here?

src/dataExplorer/context/newDataExplorer.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+export const NewDataExplorerProvider: FC = ({children}) => {
+  // Contexts
+  const {getMeasurements} = useContext(MeasurementContext)
+  const {getFields, resetFields} = useContext(FieldsContext)
+  const {getTagKeys, resetTags} = useContext(TagsContext)
+
+  // States
+  const [selectedBucket, setSelectedBucket] = useState(null)
+  const [selectedMeasurement, setSelectedMeasurement] = useState('')
```
asalem1, line 59: Why is this a concern of this context?

asalem1 (approved):
Overall it’s looking really great. I am curious as to why the selectedMeasurements aren’t a concern of the measurement context. I understand why (selecting a measurement should reset the fields and tags), but that seems like a consequence that should cascade within the context of the field and tags rather than something we do explicitly here. I’m not gonna block this PR for that since it might be a bit more work than is necessary simply for a demo, but I’d love to track that in an issue so that we can scale this out without these explicit actions

src/dataExplorer/components/Schema.tsx
appletreeisyellow: Nice catch!

appletreeisyellow:
@asalem1 I have been ponder on this question -

> why the selectedMeasurements aren’t a concern of the measurement context

Here are the reasons I can think of why we should keep the `selectedMeasurement` in `NewDataExplorerContext` instead of moving it to `MeasurementContext`:

1. I think `MeasurementContext` is a place to keep track of *all* the measurements instead of which *one* of the measurements is selected by a user. Sorry, I should change the name to `MeasurementsContext` with plural to avoid confusion 😅

2. There will be a deadlock if we move `selectedMeasurement` into measurements context. As you can see the structure of the schema browser [here](https://github.com/influxdata/ui/blob/master/src/dataExplorer/components/Schema.tsx#L72-L92), it's like this:
```js
<MeasurementProvider>
  <FieldsProvider>
     <TagsProvider>
              ...
     </TagsProvider>
  </FieldsProvider>
</MeasurementProvider>
```
If we move `selectedMeasurement` into `MeasurementContext`, then it doesn't have the way to communicate to update the fields, tag keys, and tag values. This is how we do it [now](https://github.com/influxdata/ui/blob/master/src/dataExplorer/context/newDataExplorer.tsx#L83-L84).

If we restructure the providers, like this:
```js
<FieldsProvider>
  <TagsProvider>
    <MeasurementProvider>
    </MeasurementProvider>
  </TagsProvider>
</FieldsProvider>
```
Then FieldsContext and TagsContext loss the access to rest themselves whenever a selected measurement is changed.

3. Currently, we keep track of the `selectedBucket` in [`NewDataExplorerContext`](https://github.com/influxdata/ui/blob/master/src/dataExplorer/context/newDataExplorer.tsx#L19-L31). Since the [bucket context is shared](https://github.com/influxdata/ui/blob/master/src/dataExplorer/components/BucketSelector.tsx#L10) among the entire UI, we store the `selectedBucket` in `NewDataExplorerContext` that is only specific to the schema browser. In the future for code injection implementation, we might want to keep track of the selected fields, tag keys, and tag values. It makes more sense to keep all the selected values (bucket, measurement, fields, tag keys, and tag values) in the same context.

PR #4652, opened 2022-05-24
WIP: correct windowPeriod calculation via query ranges and scope appropriately, make variable abstractions more clear, and define a windowPeriod duration vs timeRange duration.

### LINKED ISSUE #4646: Windows Period v. parameters not getting calculated correctly.
While investigating a customer issue, @nathanielc found that the UI was not correctly injecting the v.windowPeriod when working with non simplified flux queries (for instance, ones that had multiple from range filters).

@wiedld  investigated and found problems with both the calculation and injection of v.windowPeriod.  

This issue is to track the work associated with that problem identification.

### LINKED ISSUE #4836: Scoping of v.windowPeriod.
Client examples included scoped use of windowPeriod declarations. The basic pattern of scoping is this:
* has the user already hardcoded what value of `v.windowPeriod` that they want?
  * yes --> do not calculate one for them.
  * no --> calculate and provide with query. 
* for scoped calculation:
  * where is the `v.windowPeriod` used?
  * what is the scope at the moment of it's use? Which timeRange is it in?
  * also be aware that user declaration of `v.windowPeriod` can be scoped.

## Summary of the current test cases, and the expected outcome.

```
// TEST CASE  --> use declared v.windowPeriod
// EXPECT: nothing injected
// IN PROD: error @5:1-5:25: variable "v" conflicts with option of same name
import "array"
v = {windowPeriod: 2444}
array.from(rows: [{x: v.windowPeriod}])

// TEST CASE  --> use outerScope windowPeriod, without a range in the query
// EXPECT: inject v.windowPeriod
// IN PROD: injects windowPeriod undefined
import "array"
array.from(rows: [{x: v.windowPeriod}])

// TEST CASE  --> use v.windowPeriod extracted from range transformation
// EXPECT: inject v.windowPeriod, using 120000
// IN PROD: no error. Does correct injection of v.windowPeriod = 120000
import "influxdata/influxdb/sample"
sample.data(set: "airSensor")
  |> range(start: -12h, stop: now())
  |> yield(name: "${v.windowPeriod}")

// TEST CASE  --> use v.windowPeriod from inner scope
// EXPECT: nothing injected
// IN PROD: it injects the -12h...not the 2444
import "influxdata/influxdb/sample"
sample.data(set: "airSensor")
  |> range(start: -12h, stop: now())
  |> map(fn: (r) => {
    v = {windowPeriod: 2444}
    return ({ r with _value: v.windowPeriod })
  })

// TEST CASE  --> use v.windowPeriod from inner scope, does not overwrite outer scope
// EXPECT: injects windowPeriod from outer UI variables
// IN PROD: incorrectly injects from the inner scoped range
import "influxdata/influxdb/sample"
import "array"
sample.data(set: "airSensor")
  |> range(start: -12h, stop: now())
  |> map(fn: (r) => {
    v = {windowPeriod: 2444}
    return ({ r with _value: "foo" })
  })
array.from(rows: [{x: v.windowPeriod}])

// TEST CASE  --> detect v.windowPeriod in the inner lambda
// EXPECT: inject v.windowPeriod, using the range -12h
// IN PROD: is correct
import "influxdata/influxdb/sample"
sample.data(set: "airSensor")
  |> range(start: -12h, stop: now())
  |> map(fn: (r) => {
    return ({ r with _value: "${v.windowPeriod}" })
  })

// TEST CASE: timeRange variable bound by VariableAssignment statement
// EXPECT: inject v.timeRangeStop, v.windowPeriod 119833ms
// IN PROD: option v = {timeRangeStop: now(), windowPeriod: 119833ms}
import "influxdata/influxdb/sample"
import "date"
import "experimental"
truncated = date.truncate(t:v.timeRangeStop,unit:1m)
splitPoint = -1m
sample.data(set: "airSensor")
  |> range(start: -12h, stop: splitPoint)
  |> yield(name: "${v.windowPeriod}")

// TEST CASE: two timeRanges used. Should grab the proper one for windowPeriod.
// EXPECT: inject v.windowPeriod using the range -12h
// IN PROD: injects for the shortest range. In this case, the -15m.
import "influxdata/influxdb/sample"
sample.data(set: "airSensor")
  |> range(start: -12h, stop: now())
  |> yield(name: "${v.windowPeriod}")
sample.data(set: "airSensor")
  |> range(start: -15m, stop: now())
```

### LINKED ISSUE #4839: windowPeriod Variable has a cycle in the UI.
## Found issue:
* We had a cycle.
   1. use TimeRange to create UI Variable
   2. set windowPeriod UI Variable
   3. provide all UI Variables in outer scope
   4. then decide if we really need windowPeriod Variable based on flux query (inner scope)
   5. remove then from UI variables? or not? override the UI Variable? what is the src of truth here?
* We don't need the UI windowPeriod Variable
   * we don't use this Variable as a UI typed Variable obj.
      * We only use it at "flux runtime" --> externs and variable injection into flux code.
      * Which is the VariableAssignment ast node.
   * so why are we doing all this stuff with a UI windowPeriod Variable?

## Proposed solution?
* remove use of windowPeriod as a constructed UI variable.
* only create windowPeriod at the moment of use. And only as an AST node (VariableAssignment).
* proposed change is made here:
  * https://github.com/influxdata/ui/pull/4652

wiedld:
# WIP
**IN PROGRESS.**

Much of the code which extracts `v.windowPeriod` is 2-3 years old, with some maintenance as needed. It handles a very simple use case, which is no longer sufficient for current needs.

Additionally, the variable handling code needs some maintenance. This WIP is the first step in that direction.
Closes #4646
Closes #4836
Closes #4839

## Issue 4646: fix windowPeriod calculation
* The `v.windowPeriod` is not calculated, unless a timeRange is coded into the flux query.
    * Requirement for time range:
        * https://github.com/influxdata/ui/blob/582e53c01d2a2460e0825ad60b4536d9894cf58b/src/shared/utils/getMinDurationFromAST.ts#L32-L36
    * Only provide timeRange variables, if they are used:
        * https://github.com/influxdata/ui/blob/582e53c01d2a2460e0825ad60b4536d9894cf58b/src/variables/utils/getWindowVars.ts#L126
* Additionally, the timeRanges are then used to calculate the windowPeriod using this set of rules:
    * Summarized steps:
        * Grab all `start` and `stop` across the entire query.
        * Takes the minimum possible windowPeriod as a cartesian product.
            * This answer is only correct when there is only 1 timeRange.
    * Comment and key code:
        * https://github.com/influxdata/ui/blob/582e53c01d2a2460e0825ad60b4536d9894cf58b/src/shared/utils/getMinDurationFromAST.ts#L26-L30
        * https://github.com/influxdata/ui/blob/582e53c01d2a2460e0825ad60b4536d9894cf58b/src/shared/utils/getMinDurationFromAST.ts#L38-L40
* Finally, the `findNodes()` method simply returns all node which match a given condition.
    * There is no scoping. Everything is effectively global (if it’s found).

## Issue 4839: remove use of windowPeriod as a UI Variable
* We had a cycle.
   1. use TimeRange to create UI Variable
   2. set windowPeriod UI Variable
   3. provide all UI Variables in outer scope
   4. then decide if we really need windowPeriod Variable based on flux query (inner scope)
   5. remove then from UI variables? or not? override the UI Variable? what is the src of truth here?
* We don't need the UI windowPeriod Variable
   * we don't use this Variable as a UI typed Variable obj.
      * We only use it at "flux runtime" --> externs and variable injection into flux code.
      * Which is the VariableAssignment ast node.
   * so why are we doing all this stuff with a UI windowPeriod Variable?
   * **CHANGE HERE**: only create windowPeriod at the moment of use. And only as an AST node (VariableAssignment).

## Cleanup:
*  Overlapping concepts:
    * `Variable` vs `VariableAssignment`.
        * definition:
            * `Variable` is our UI type
            * `VariableAssignment` is the flux AST node
        * Done here:
            * A bit more explicit wording, when we use which.
    * `Duration` is used for two different meanings:
        * (1) Total time in range (end-start).
        * (2) `windowPeriod.duration`:
            * A smaller time window.
            * Determined by two possible methods:
                * constant SELECTABLE_TIME_RANGES
                * Math.round((end-start)/DESIRED_POINTS_PER_GRAPH)
        * Done here:
            * Pulled duration utils into a single dir.
            * Tried to make the conversion more explicit.
* Break into submodules:
    * Move tightly coupled, AST-related methods to a single `src/shared/utils/ast/`
    * Move all the various “duration” code to the single dir `src/shared/utils/duration`

## TODO:
- [x] ~Double use of the `v` namespace.~ **Future work. Pushed into an issue ticket.** 
    * https://github.com/influxdata/ui/issues/4691
- [x] Notebooks is remaining on legacy code (wrong calculation of the windowPeriod).
    * notebooks ignores the existing Variable typing and abstractions.
       * in some cases, notebooks code uses the common helper methods & abstracts --> so can get the right calculations.
       * in other cases, it does not use these type abstractions --> so is not using the helpers.
    * Therefore notebooks is not getting fixed.
- [x] Make issue ticket: spike on how to handle:
    * for `v.timeRangeStart`|`v.timeRangeStop` as runtime Expr
    * https://github.com/influxdata/ui/issues/4695
- [x] figure out which key nodes are missing from `findNodeScope()`
    * ~update `src/types/ast`~
    * Instead, handle default case with unknown nodes are just an end to recursing. Return acc.
- [ ] Write tests, based on use cases tested.
- [ ] Handoff deployed remocal instance to Product, to perform QA


src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -58,37 +57,10 @@ const ResultsPane: FC = () => {
...
-      }
-
-      if (!timeRange.upper) {
-        timeRangeStop = 'now()'
-      } else if (isNaN(Date.parse(timeRange.upper))) {
-        timeRangeStop = null
-      } else {
-        timeRangeStop = new Date(timeRange.upper).toISOString()
-      }
-    }
```
wiedld, line 83: This code is only used to extract a `timeRangeStart` and `timeRangeStop` from the TimeRange abstraction. So can override an override.
* there is lower down code, which incorrectly overrides the TimeRange abstraction.

src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -58,37 +57,10 @@ const ResultsPane: FC = () => {
...
-
     setStatus(RemoteDataState.Loading)
     const time = Date.now()
     query(text, {
-      vars: {
-        timeRangeStart,
-        timeRangeStop,
-      },
+      vars: !!timeRange ? getTimeRangeVars(timeRange) : [],
```
wiedld, line 63: give the proper extraction

src/dataExplorer/components/SaveAsTaskForm.tsx
```diff
@@ -136,7 +137,7 @@ const mstp = (state: AppState) => {
   const activeQuery = getActiveQuery(state)
   const org = getOrg(state)
   const userDefinedVars = getAllVariables(state)
-    .map(v => asAssignment(v))
+    .map(v => asAssignmentNode(v))
```
wiedld, line 140: A bit of explicit naming updates.
`Variable` == UI abstraction/type
`VariableAssignment` == flux AST for a Variable

src/flows/context/flow.query.tsx
```diff
@@ -238,19 +243,31 @@ export const FlowQueryProvider: FC = ({children}) => {
   }
 
   const query = (text: string, override?: QueryScope): Promise<FluxResult> => {
+    // DLW FIXME TODO:
+    // this override can be the individual range on a pipe
+    // but it's not a TimeRange time. have to reach further into the code.
+    console.log('override', override)
+    console.log('flow.query > query', flow.range)
```
wiedld, line 250: reminder to myself. Have to finish updating all the referring code, which are not using the TimeRange abstraction.

src/flows/context/variables.tsx
```diff
@@ -37,8 +36,7 @@ export const VariablesProvider: FC = ({children}) => {
       getRangeVariable(TIME_RANGE_START, range),
       getRangeVariable(TIME_RANGE_STOP, range),
     ]
-    const windowVar = getWindowPeriodVariableFromVariables(source, timeVars)
-    setVariables(!!windowVar ? timeVars.concat(windowVar) : timeVars)
```
wiedld, line 41: windowVars (for the editor) are now handled in the prelude to the LSP.
Benefiting all editors across the app.

src/shared/apis/queryCache.ts
```diff
@@ -229,18 +220,14 @@ export const getCachedResultsOrRunQuery = (
...
-    windowVars = getWindowVarsFromVariables(query, variables)
-  }
+  const windowVar = getWindowPeriodVariableFromVariables(query, variables)
 
   // otherwise query & set results
-  const extern = buildUsedVarsOption(query, variables, windowVars)
+  const extern = buildUsedVarsOption(
+    query,
+    windowVar ? variables.concat(windowVar) : variables
+  )
```
wiedld, line 230: `buildUsedVarsOption` 
* already has a contract to de-dupe based upon what variables are in the query.
* it then appends the optional 3rd function parameter.

src/shared/contexts/query.tsx
```diff
@@ -192,7 +129,9 @@ export const simplify = (text, vars = {}) => {
     )
       .map(node => node.property.name)
       .reduce((acc, curr) => {
-        acc[curr] = vars[curr]
+        if (vars[curr]) {
+          acc[curr] = vars[curr]
+        }
```
wiedld, line 134: this code was incorrect.

What it was doing:
*detecting a used variable (`v.myVar`) in the query string
* then expected that it must be defined externally in the vars function parameter
* However, if it's defined in the query code itself, and not the vars, it was injecting into the flux `option v = {myVar: undefined}`

src/shared/utils/ast/extractors.ts
```diff
@@ -84,6 +56,8 @@ function rangeTimes(ast: any, rangeNode: CallExpression): [number, number] {
   return [start, end]
 }
 
+// FIXME: this should not use lookupVariable(ast, value.name)
+// instead should lookup value in scope map
```
wiedld, line 60: cannot fix all variable handling in this PR.
comment left for future reference/work.

src/shared/utils/ast/nodes.ts
```diff
@@ -0,0 +1,25 @@
+import {get} from 'lodash'
```
wiedld, line 1: This file has only 2 new functions.
* isWindowPeriodVariableNode, isVariableDeclaration
* the rest is moved code

src/shared/utils/ast/visitors.ts
```diff
@@ -0,0 +1,338 @@
...
+  [variableIdentifier: string]: Node
+}
+const SCOPED_WINDOW_PERIOD = `v.${WINDOW_PERIOD}`
+
+/*
+  Find the first node matching the halt `predicate` function.
+  Return the variable scope for this node.
+  Follow flux scoping rules.
+*/
+export const findNodeScope = (
```
wiedld, line 51: new function.

src/shared/utils/duration/getDurationFromTimeRange.ts
```diff
@@ -0,0 +1,23 @@
+import {TimeRange} from 'src/types'
+import {
+  millisecondsToDuration,
+  calcWindowPeriodForRangeDuration,
+} from 'src/shared/utils/duration'
+
+export function getDurationFromTimeRange(timeRange: TimeRange) {
```
wiedld, line 7: this is function is logic extracted from repeated code.

src/shared/utils/duration/index.ts
```diff
@@ -0,0 +1,19 @@
+import {DESIRED_POINTS_PER_GRAPH} from 'src/shared/constants'
+import {SELECTABLE_TIME_RANGES} from 'src/shared/constants/timeRanges'
+
+export function calcWindowPeriodForRangeDuration(rangeDuration) {
+  return Math.round(rangeDuration / DESIRED_POINTS_PER_GRAPH) // duration in ms
+}
+
+export function convertTimeRangeDurationToWindowPeriodDuration(rangeDuration) {
```
wiedld, line 8: repeated code, now made into common helper func.

src/shared/utils/duration/timeRangeLabels.ts
```diff
@@ -78,13 +78,13 @@ export const durationToMilliseconds = (duration: Duration[]): number =>
...
+Convert an amount of milliseconds to a duration string.
 
-  The returned duration string will use the largest units possible, e.g.
+The returned duration string will use the largest units possible, e.g.
 
-      millisecondsToDuration(9_000_000)
+    millisecondsToDuration(9_000_000)
 
-  Will return `2h30m` rather than `9000000ms`.
+Will return `2h30m` rather than `9000000ms`.
```
wiedld, line 87: linter changes, because I moved the file.

src/types/ast.ts
```diff
@@ -174,7 +179,7 @@ export interface BinaryExpression extends BaseNode {
 export interface CallExpression extends BaseNode {
   callee: Expression
   arguments?: Expression[]
-  type: 'CallExpression'
+  type?: 'CallExpression' // observed in ASTs. type not always present.
```
wiedld, line 182: I am hesitant about this.
Found for range expressions. The rest of the node had the proper structure...but it was not typed.

src/types/flows.ts
```diff
@@ -141,7 +141,10 @@ export interface VariableMap {
 
 // TODO: type this better. there are required properties, that have types, but
 // we also need this to stay open for panels to register whatever they want
-export interface QueryScope {
+interface VariableOuterScope {
+  vars?: Variable[]
+}
+export interface QueryScope extends VariableOuterScope {
```
wiedld, line 147: This use of `[props: string]: any` means that the TimeRange variables (and other variables) are easily overriden with anything. And the dev cannot lean on typescript. 😭

src/variables/utils/buildVarsOption.ts
```diff
@@ -32,7 +30,8 @@ export const buildVarsOption = (variables: VariableAssignment[]): File => ({
   type: 'File',
   package: null,
   imports: null,
-  body: [
+  // only claim namespace `v` if needed
+  ...(!variables.length ? {body: []} : {body: [
```
wiedld, line 34: We have two choices:
1. only send the extern when optional variables are present. It means having changes be correct in >=5 places across the codebase.
2. always create the extern. But only claims the namespace `v` if variables are provided via extern.

src/variables/utils/convertVariables.ts
```diff
@@ -0,0 +1,95 @@
...
+  TIME_RANGE_START,
+  TIME_RANGE_STOP,
+  WINDOW_PERIOD,
+} from 'src/variables/constants'
+
+// Types
+import {VariableAssignment} from 'src/types/ast'
+import {Variable} from 'src/types'
+
+export const asAssignmentNode = (variable: Variable): VariableAssignment => {
```
wiedld, line 12: existing function. Moved only. No changes.

src/variables/utils/getTimeRangeVars.ts
```diff
@@ -9,14 +9,19 @@ import {TIME_RANGE_START, TIME_RANGE_STOP} from 'src/variables/constants'
...
-export const getTimeRangeVars = (
+export const getTimeRangeVarAssignments = (
   timeRange: TimeRange
 ): VariableAssignment[] => {
+  return getTimeRangeVars(timeRange)
+    .map(v => asAssignmentNode(v))
+    .filter(v => !!v)
+}
+
+export const getTimeRangeVars = (timeRange: TimeRange): Variable[] => {
```
wiedld, line 20: break into 2 methods.
* return ui Variable abstraction
* return flux node VariableAssignment.

src/variables/utils/getWindowVars.test.ts
```diff
@@ -1,78 +1,78 @@
-import {getWindowPeriodVariable} from 'src/variables/utils/getWindowVars'
-import {defaultVariableAssignments} from 'src/variables/mocks'
+// import {getWindowPeriodVariable} from 'src/variables/utils/getWindowVars'
+// import {defaultVariableAssignments} from 'src/variables/mocks'
```
wiedld, line 2: TODO (not yet done in this WIP).
Updating/adding tests.
Have a full set of test cases with scoping rules etc.

src/variables/utils/getWindowVars.ts
```diff
@@ -2,152 +2,58 @@
...
+    console.warn('error', error)
     return null
   }
 }
 
-export const getWindowPeriodVariable = (
+/*
+  Extract the UI abstraction Variable for `v.windowPeriod`
+*/
+export const getWindowPeriodVariableFromVariables = (
```
wiedld, line 52: * 2 main functions (in this file):
     * used for UI Variable
     * used for flux ast VariableAssignment.
* both now use the same core logic.

src/shared/apis/queryCache.ts
wiedld: note: the `buildUsedVarsOption` dedupe is done using the older (treat everything as global scope) pattern. But we can fix that is a followup/future PR.

src/shared/components/TimeSeries.tsx
```diff
@@ -267,6 +269,7 @@ class TimeSeries extends Component<Props, State> {
 
       let statuses = [] as StatusRow[][]
       if (check) {
+        // DLW FIXME TODO: decide how to handle
         const extern = buildUsedVarsOption(
           queries.map(query => query.text),
```
wiedld, line 274: This is multiple queries.
Some of which may define the variable internally. Some of which may not.
So how do we handle when/if to include the windowPeriod variable?

src/shared/components/TimeSeries.tsx
```diff
@@ -267,6 +269,7 @@ class TimeSeries extends Component<Props, State> {
 
       let statuses = [] as StatusRow[][]
       if (check) {
+        // DLW FIXME TODO: decide how to handle
```
wiedld, line 272: The below used of `buildUsedVarsOption` is with multiple queries at once.
Some of which may define the variable internally. Some of which may not.
So how do we handle when/if to include the windowPeriod variable?

src/flows/context/flow.query.tsx
```diff
@@ -1,21 +1,26 @@
...
+
+// Context, and State selectors
 import {FlowContext} from 'src/flows/context/flow.current'
 import {ResultsContext} from 'src/flows/context/results'
 import {QueryContext, simplify} from 'src/shared/contexts/query'
-import {event} from 'src/cloud/utils/reporting'
-import {FluxResult, QueryScope} from 'src/types/flows'
-import {PIPE_DEFINITIONS, PROJECT_NAME} from 'src/flows'
-import {notify} from 'src/shared/actions/notifications'
-import EmptyGraphMessage from 'src/shared/components/EmptyGraphMessage'
```
wiedld, line 10: A bit of tech debt fixing. Placed the imports in the correct location.

src/variables/utils/buildVarsOption.ts
```diff
@@ -6,48 +6,55 @@ import {File, Property, VariableAssignment, Variable} from 'src/types'
...
-
-// Selectors
-import {asAssignment} from 'src/variables/selectors'
+import {asAssignmentNode} from 'src/variables/utils/convertVariables'
 
 export const buildUsedVarsOption = (
   query: string | string[],
   allVariables: Variable[],
-  windowVars?: VariableAssignment[]
+  variableAssignmentNodesToAppend?: VariableAssignment[]
```
wiedld, line 14: Changed parameter naming. To make more explicit, what it does.

src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -58,37 +56,10 @@ const ResultsPane: FC = () => {
...
-    }
-
     setStatus(RemoteDataState.Loading)
     const time = Date.now()
     query(text, {
-      vars: {
-        timeRangeStart,
-        timeRangeStop,
-      },
+      variables: !!timeRange ? getTimeRangeVars(timeRange) : [],
```
wiedld, line 62: instead of using `timeRangeStart` and `timeRangeStop` as on override (or an override)...just pass the timeRange Variables.

src/flows/context/flow.query.tsx
```diff
@@ -129,6 +134,8 @@ export const FlowQueryProvider: FC = ({children}) => {
             timeRangeStart,
             timeRangeStop,
           },
+          // future use: additional Variables injected into notebooks
+          variables: flow?.range ? getTimeRangeVars(flow.range) : [],
```
drdelambre, line 138: how is this different than the variables immediately above this?

wiedld: The focus is on using the typed objects (which we already have in TimeRange), and transforming with the typed helpers (which we already have), in order to produced the typed output (Variables) which is an abstraction used in all other helpers we need downstream.

drdelambre: if that type you're looking for is `Variables`, you're going in the wrong direction. after many months of detangling broken dashboard lifecycles and hand crafting rendering rules to prevent the page from crashing if it included a variable, it was an intentional move to shift away from that implementation that conflates resource fetching / storage / selection / context of selection and all of that into a separate type that represents a variable selection and not a variable resource.

src/shared/utils/ast/visitors.ts
```diff
@@ -0,0 +1,373 @@
...
+    case 'DurationLiteral':
+    case 'FloatLiteral':
+    case 'PipeLiteral':
+    case 'RegexpLiteral':
+    case 'StringLiteral':
+    case 'UnsignedIntegerLiteral':
+      return acc
+    default:
+      return acc
+    // return assertUnreachable(node.type as never)
```
wiedld, line 308: This line can be uncommented during dev.

It forces typescript to make sure that the switch statement is exhausted. Meaning, that all possible use cases (based on all types that typescript knows about) are handled.

src/dataExplorer/components/ResultsPane.tsx
wiedld: took out the changes to the ResultsPane and notebooks. Figure we can have that discussion, after this PR goes in (and fixes everywhere else in the UI application).

src/shared/contexts/query.tsx
```diff
@@ -104,86 +100,27 @@ export const remove = (node: File, test, acc = []) => {
...
-  }
+  const windowDuration = getDurationFromAST(ast, optionAST)
 
   windowPeriod.forEach(node => {
     node.value = {
       type: 'DurationLiteral',
       values: [
         {
-          magnitude: Math.round(queryDuration / DESIRED_POINTS_PER_GRAPH),
+          magnitude: windowDuration,
```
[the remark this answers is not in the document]
wiedld: This code was left in. Even though it's wrong, per the reasons above.
But I'm leaving notebooks code alone for now, which means leaving the shared QueryContext alone.

src/shared/contexts/query.tsx
wiedld: This code was left in. Even though it's wrong, per the explanation^^ above.
But I'm leaving notebooks code alone for now, which means leaving the shared QueryContext alone.

src/shared/utils/ast/extractors.ts
wiedld: ah...I actually ended up fixing this. Because I needed to.

src/types/flows.ts
wiedld: This code was left in. Even though it's wrong, per the explanation^^ above.
But I'm leaving notebooks code alone for now, which means leaving this alone.

src/dataExplorer/components/ResultsPane.tsx
wiedld: more of the same. this is a code smell -- but leaving ResultsPane and notebooks alone for now.

src/flows/context/flow.query.tsx
wiedld: The rest of the UI application (outside of notebooks) still uses the `Variable` abstraction. Also consulted with other senior devs who are still using this abstraction. For now, going to scope these changes to the rest of the UI application outside of notebooks.

wiedld:
We are not taking this route. Instead, we are changing our general approach to using the windowPeriod.

PR #4663, opened 2022-05-24
test: debug tasks test, to unblock CI pipeline

wiedld:
Unblock CI pipeline.
Part of #4673 

Does not fix all of the shared state issues in the Tasks part of the application. But hits at single the single biggest culprit.

## Change made:
* We are using a global store to fill out the form.
   * The form gets filled out, depending on which task cypress test, as either 'Task' or '🦄ask.
   * When the tests failed locally, they did it when it was expecting 'Task' but got '🦄ask'.
* Usually I try to keep form state as local state (react hooks or stateful components), and keep it out of any global store (e.g. redux). Because it can easily cause shared state issues.
   * The specific error here was a mutable default object.


hoorayimhelping (approved):
sure why not

hoorayimhelping (approved):

PR #4678, opened 2022-05-25
chore: replace `-30d` time range with `-100y` for sample data

### LINKED ISSUE #4676: Replace time range with caching required dates
Currently in the Flux Query Builder schema browser, we are using `start: -30d, stop: now()` as the time range to get measurements, fields, and tags. 

We want to replace it with `CACHING_REQUIRED_START_DATE` and `CACHING_REQUIRED_END_DATE` to speed up the query. 

## Why?

Because it's faster with the latter way at the storage level. When the storage sees any query starts and ends with the `CACHING_REQUIRED_START_DATE` and `CACHING_REQUIRED_END_DATE`, it returns the indexed measurements, fields, and tags.

appletreeisyellow:
Closes #4676 

~~This PR replace the `-30d` time range with constant start and end dates to speed up query.~~

This PR replace the `-30d` time range with `-100y` for sample data. See [below](https://github.com/influxdata/ui/pull/4678#issuecomment-1138944434) for the reason why query sample data is different from query real data.

### DIFF
--- a/src/dataExplorer/context/fields.tsx
+++ b/src/dataExplorer/context/fields.tsx
@@ -78,7 +78,7 @@ export const FieldsProvider: FC<Prop> = ({children, scope}) => {
     }
 
     let queryText = `${_source}
-      |> range(start: ${CACHING_REQUIRED_START_DATE}, stop: ${CACHING_REQUIRED_END_DATE})
+      |> range(start: -100y, stop: now())
       |> filter(fn: (r) => (r["_measurement"] == "${measurement}"))
       |> keep(columns: ["_field"])
       |> group()
--- a/src/dataExplorer/context/fields.tsx
+++ b/src/dataExplorer/context/fields.tsx
@@ -77,9 +77,8 @@ export const FieldsProvider: FC<Prop> = ({children, scope}) => {
       _source += FROM_BUCKET(bucket.name)
     }
 
-    // TODO: is 30d large enough to get all field values?
     let queryText = `${_source}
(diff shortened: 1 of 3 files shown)


asalem1 (commented):
🤷🏽 I would run this by @nathanielc and @garylfowler to make sure this is actually a requirement. The reason the other query accepts the const dates you're injecting here are to help enable an optimization that was introduced for that particular query, and this may actually make the query in this context very ... slow. Ideally we wouldn't need to even have this query in here because the feature flagged one would be live by the time this is delivered but .... who knows

appletreeisyellow:
Let me double check on this

appletreeisyellow:
> Is it true that any query with this time range will returned the indexed values?

I checked with Paul [on slack](https://influxdata.slack.com/archives/C030G5783M3/p1653577164179439). 

For non-sample data, the answer is YES if the query uses [schema](https://docs.influxdata.com/flux/v0.x/stdlib/influxdata/influxdb/schema/) package to get measurements, fields, tag keys, and tag values with time range `CACHING_REQUIRED_START_DATE` (1677-09-21T00:12:43.145224194Z) and `CACHING_REQUIRED_END_DATE` (2262-04-11T23:47:16.854775806Z). 

However, the answer is NO for sample data, because sample data doesn't come from storage. This means we cannot use `schema` package to query sample data, which means the query won't return indexed values. Paul suggested to use `-100y` to get measurements, fields, tag keys, and tag values for sample data.

asalem1 (approved):

PR #4680, opened 2022-05-25
feat: eventing for bulk action - delete (tokens)

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/3146

Simple events to measure the use of bulk action delete for tokens

### DIFF
--- a/src/authorizations/components/TokensTab.tsx
+++ b/src/authorizations/components/TokensTab.tsx
@@ -39,6 +39,7 @@ import {bulkDeleteAuthorizations} from 'src/authorizations/actions/thunks'
 // Styles
 import './TokensTabStyles.scss'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {event} from 'src/cloud/utils/reporting'
 
 enum AuthSearchKeys {
   Description = 'description',
@@ -295,6 +296,9 @@ class TokensTab extends PureComponent<Props, State> {
     this.props.bulkDeleteAuthorizations(
       this.state.tokensSelectedForBatchOperation.map(token => token.id)
     )
+
+    const numberOfSelectedTokens = this.state.tokensSelectedForBatchOperation
+    event(`bulkAction.deleted.${numberOfSelectedTokens}.tokens`)
     // reset the list of selected tokens
     this.setState({
       tokensSelectedForBatchOperation: [],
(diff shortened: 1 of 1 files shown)


TCL735 (approved):

src/authorizations/components/TokensTab.tsx
```diff
@@ -295,6 +296,9 @@ class TokensTab extends PureComponent<Props, State> {
     this.props.bulkDeleteAuthorizations(
       this.state.tokensSelectedForBatchOperation.map(token => token.id)
     )
+
+    const numberOfSelectedTokens = this.state.tokensSelectedForBatchOperation
+    event(`bulkAction.deleted.${numberOfSelectedTokens}.tokens`)
```
hoorayimhelping, line 301: I think we should consider changing this to similar how it is below, where instead of making the count a unique name, the event is generic and the count is a tag or a value.

```ts
event('bulkAction.tokens.deleted', {}, {count: numberOfSelectedTokens})
```

This keeps the action generic, but allows us to get a count on a specific action, rather than grouping bulk actions by how many tokens were deleted

so when we look at the stats, instead of
```
bulk action deleted 5 tokens   |   5
bulk action deleted 2 tokens   |   35
bulk action deleted 55 tokens  |   1
bulk action deleted 3 tokens   |   5
```
we have some more like:
```
bulk action tokens deleted     |   46
```

src/authorizations/components/TokensTab.tsx
```diff
@@ -319,7 +323,9 @@ class TokensTab extends PureComponent<Props, State> {
 
     if (isEmpty(tokensSelectedForBatchOperation)) {
       this.setState({tokensSelectedForBatchOperation: tokensOnCurrentPage})
+      event(`bulkAction.selectAll`, {count: tokensOnCurrentPage.length})
     } else {
+      event(`bulkAction.deSelectAll`)
```
hoorayimhelping, line 328: should these be:

```ts
'bulkAction.tokens.selectAll'
'bulkAction.tokens.deselectAll'
```
so if we add more bulk actions, we know it's for tokens?

Also, no need to use the template string (\`) if there's not going to be interpolation (`${variable}`)

hoorayimhelping (commented):
I'm not super deep into the requirements for eventing here, what do you think of the suggestions?

src/authorizations/components/TokensTab.tsx
ChitlangeSahas: ✅

hoorayimhelping (approved):

PR #4681, opened 2022-05-26
fix: find the proper task-card, now that names are not shared.

wiedld:
### Observation:
Now that we are no longer sharing taskName state through the mutable `taskOption` in redux, we are instead getting more cypress errors.

### Hypothesis:
* This could be because we exposed overselves to the fact that our tests were not waiting for the newest task-card (with the unique/non-shared taskName) to be present.
* Additionally, we have a shared backend state (during CI tests) with parallel cypress tests (which each could create their own exported tasks etc). So it is possible that we could end up with additional task cards at any given timepoint.
* Therefore:
   * ensure that we have cy.get() awaiting on our specific task-card--name
   * make sure the error did not occur with an invalid creation of the task
   * when test is doing changes over time to edit the task stored in remocal DB state, make sure is using a unique name 
       * --> because we don't want a chrome vs firefox test (or any other test) only trying to mutate the same task-by-taskName.

Yes, there is plenty of speculation going on here. But the 20 parallel test runs (per browser) all pass:
https://app.circleci.com/pipelines/github/influxdata/monitor-ci/11809

### DIFF
--- a/cypress/e2e/shared/tasks.test.ts
+++ b/cypress/e2e/shared/tasks.test.ts
@@ -410,8 +410,7 @@ from(bucket: "defbuck")
     })
 
     it('can click to filter tasks by labels', () => {
-      // should be > 1 length
-      cy.getByTestID('task-card').should('not.have.length', 1)
+      cy.getByTestID('task-card').should('have.length.greaterThan', 1)
 
       cy.getByTestID(`label--pill ${newLabelName}`).click()
 
--- a/cypress/e2e/shared/tasks.test.ts
+++ b/cypress/e2e/shared/tasks.test.ts
@@ -32,7 +32,7 @@ describe('Tasks', () => {
   })
 
   it('can create a task', () => {
-    const taskName = 'Task'
+    const taskName = 'Task A'
     cy.createTaskFromEmpty(taskName, ({name}) => {
       return `import "influxdata/influxdb/v1"
 v1.tagValues(bucket: "${name}", tag: "_field")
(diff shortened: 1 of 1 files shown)


rockstar (approved):

cypress/e2e/shared/tasks.test.ts
```diff
@@ -192,8 +190,6 @@ from(bucket: "defbuck")
     })
 
     it('can delete a task', () => {
-      const TaskName = '🦄ask'
```
wiedld, line 195: now looks for the outerscope TaskName (`🦄ask ${Math.random()}`)

cypress/e2e/shared/tasks.test.ts
```diff
@@ -32,35 +32,33 @@ describe('Tasks', () => {
...
-    cy.getByTestID('task-save-btn').click()
-
-    cy.getByTestID('task-card')
-      .should('have.length', 1)
-      .and('contain', taskName)
+    cy.getByTestID('task-save-btn')
+      .should('not.be.disabled')
+      .click()
+    cy.getByTestID('notification-success--dismiss').click()
+    cy.getByTestID('task-card--name').contains(taskName)
```
hoorayimhelping, line 61: if `taskName` is a unique value and there are no other tasks called `Task B` this can be simplified to

```ts
cy.contains(taskName)
```

cypress/e2e/shared/tasks.test.ts
```diff
@@ -414,7 +410,8 @@ from(bucket: "defbuck")
     })
 
     it('can click to filter tasks by labels', () => {
-      cy.getByTestID('task-card').should('have.length', 2)
+      // should be > 1 length
+      cy.getByTestID('task-card').should('not.have.length', 1)
```
hoorayimhelping, line 414: This comment and this line can probably be combined into a more self-documenting statement:

```ts
cy.getByTestID('task-card').should('have.length.greaterThan', 1)
```

hoorayimhelping (approved):
LGTM, just a couple of small bits about Cypress

cypress/e2e/shared/tasks.test.ts
wiedld: done.

cypress/e2e/shared/tasks.test.ts
wiedld: excellent point. done.

wiedld:
I originally made the slight changes recommended in the PR comments...then the tests kept failing. Even though the changes should be insignificant.
https://app.circleci.com/pipelines/github/influxdata/ui?branch=fix%2Fdo-not-error-based-on-test-execution-order&filter=all

So I reset it back to the last SHA which passed (when the PR was approved). Is now also not passing -- where before it passed with 20 runs in each browser (20x firefox and 20x chrome) running in parallel.

It's clear we have a lot of weirdness unresolved. But that is too big of a nut to crack atm. I'll just focus on getting something to pass in all 20 (x2 browsers) runs.

wiedld:
Closing this for now. Not going to poke the resting beast (a.k.a. the Tasks test).

PR #4684, opened 2022-05-26
chore: org rename exclusively done in quartz

gspletzer:
[Quartz #6173](https://github.com/influxdata/quartz/issues/6173)

This implements the new route for renaming `org`; org renames for Cloud users will now run exclusively through Quartz. UI currently doesn't support `org description` but it remains an optional parameter in the event it becomes available.

### DIFF
--- a/cypress/e2e/shared/about.test.ts
+++ b/cypress/e2e/shared/about.test.ts
@@ -42,9 +42,7 @@ describe('About Page', () => {
     const newOrgName = `hard@knock.life${Math.random()}`
 
     const patchOrgPath =
-      Cypress.env('dexUrl') === 'CLOUD'
-        ? 'api/v2/quartz/orgs/*'
-        : 'api/v2/orgs/*'
+      Cypress.env('dexUrl') === 'OSS' ? 'api/v2/orgs/*' : 'api/v2/quartz/orgs/*'
 
     cy.intercept('PATCH', patchOrgPath).as('patchOrg')
     cy.getByTestID('create-org-name-input').type(newOrgName)
--- a/cypress/e2e/shared/about.test.ts
+++ b/cypress/e2e/shared/about.test.ts
@@ -1,4 +1,4 @@
-import {CLOUD} from 'src/shared/constants'
+import {CLOUD} from '../../../src/shared/constants'
 import {Organization} from '../../../src/types'
 
 describe('About Page', () => {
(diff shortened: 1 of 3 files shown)


cypress/e2e/shared/about.test.ts
```diff
@@ -41,14 +41,12 @@ describe('About Page', () => {
...
 
-    cy.intercept('PATCH', 'api/v2/orgs/*').as('patchOrg')
-
     cy.getByTestID('create-org-name-input').type(newOrgName)
     cy.getByTestID('rename-org-submit--button')
       .should('not.be.disabled')
       .click()
 
-    cy.wait('@patchOrg')
+    cy.intercept('PATCH', 'api/v2/quartz/orgs/*')
```
asalem1, line 49: does this exist within the quartz mock?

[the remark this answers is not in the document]
gspletzer: I changed it because when the test was running, you could see the success hitting as expected, but then the test would get stuck on the `wait` and fail. Any ideas on how to resolve that?

gspletzer: That might account for the failure I was seeing. I always forget about quartz mock.

docker/Dockerfile.cypress
```diff
@@ -9,11 +9,11 @@ COPY ./package.json .
 
 RUN yarn add cypress-circleci-reporter
 
-COPY ./cypress.json .
+COPY ./cypress.json ./cypress.json
```
hoorayimhelping, line 12: is this necessary or was this a change made while working a solution?

I'm pretty sure that

```docker
COPY ./cypress.json .
```

is identical to
```docker
COPY ./cypress.json ./cypress.json
```
so it probably doesn't make a huge deal.

hoorayimhelping (approved):

docker/Dockerfile.cypress
gspletzer: @hoorayimhelping the change was made when I was trouble shooting the cypress dockerfile. All the other paths were listed twice, so I updated that one to match. If you want it removed, I can revert that line back.

hoorayimhelping: @gspletzer since this PR is passing and it's more consistent and explicit, I think it's fine to leave it.

PR #4686, opened 2022-05-26
fix: redirect to Quartz login upon session expiration

### LINKED ISSUE #4657: Redirect to main login page after UI session expiry
The UI hits the /me endpoint every minute. When the session expires, the UI gets an unauthorized response from /me and then it decides what to do. Right now it redirects to the cluster login page (which is deprecated). The UI needs to redirect to the quartz login page. 

https://<cloud2.influxdata.com>/login (where <cloud2.influxdata.com> is the appropriate Quartz URL from which the UI is served: prod/staging/remocal/local). 

See EAR https://github.com/influxdata/EAR/issues/3012

TCL735:
Closes #4657 

When a session expires, redirect to the Quartz login page rather than the cluster's login page.

### DIFF
--- a/src/Signin.tsx
+++ b/src/Signin.tsx
@@ -112,20 +112,26 @@ export class Signin extends PureComponent<Props, State> {
       } = this.props
 
       clearInterval(this.intervalID)
-      /**
-       * We'll need this authSessionCookieOn flag off for tools until
-       * Quartz is integrated into that environment
-       */
-      if (CLOUD && isFlagEnabled('authSessionCookieOn')) {
-        const url = new URL(
-          `${window.location.origin}${CLOUD_LOGIN_PATHNAME}?redirectTo=${window.location.href}`
-        )
-        setToLocalStorage('redirectTo', window.location.href)
-        window.location.href = url.href
-        throw error
-      }
 
       if (CLOUD) {
+        if (isFlagEnabled('useQuartzLogin')) {
+          window.location.reload()
+          throw error
+        }
+
+        /**
+         * We'll need this authSessionCookieOn flag off for tools until
+         * Quartz is integrated into that environment
+         */
+        if (isFlagEnabled('authSessionCookieOn')) {
+          const url = new URL(
+            `${window.location.origin}${CLOUD_LOGIN_PATHNAME}?redirectTo=${window.location.href}`
+          )
+          setToLocalStorage('redirectTo', window.location.href)
+          window.location.href = url.href
+          throw error
+        }
+
         const url = new URL(
           `${window.location.origin}${CLOUD_SIGNIN_PATHNAME}?redirectTo=${window.location.href}`
         )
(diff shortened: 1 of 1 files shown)


TCL735:
@asalem1 does this look right to you? I'm taking a stab here and may be totally off base.

src/Signin.tsx
```diff
@@ -117,9 +113,7 @@ export class Signin extends PureComponent<Props, State> {
        * Quartz is integrated into that environment
        */
       if (CLOUD && isFlagEnabled('authSessionCookieOn')) {
-        const url = new URL(
-          `${window.location.origin}${CLOUD_LOGIN_PATHNAME}?redirectTo=${window.location.href}`
-        )
+        const url = new URL(`${CLOUD_URL}?redirectTo=${window.location.href}`)
```
asalem1, line 116: 🤷🏽 this should work w/r/t to directing the user the Quartz login page, but I don't think the `redirectTo` stuff will be relevant. Since we're dealing with authentication and signing in, I would feature flag this behind a public feature flag so that we ensure that there's some sort of stability in place prior to merging in something that may unintentionally break auth

asalem1 (changes requested):

TCL735:
Changes pending. We will turn this into a page refresh rather than redirect. By refreshing the page, Quartz will automatically handle everything: from detecting an unauthenticated user to login page to redirect back to previous page user was on.

src/Signin.tsx
```diff
@@ -112,20 +112,26 @@ export class Signin extends PureComponent<Props, State> {
...
       if (CLOUD) {
+        if (isFlagEnabled('useQuartzLogin')) {
+          window.location.reload()
+          throw error
+        }
+
+        /**
+         * We'll need this authSessionCookieOn flag off for tools until
+         * Quartz is integrated into that environment
+         */
```
hoorayimhelping, line 125: i think we can remove this comment since it's not relevant anymore

src/Signin.tsx
```diff
@@ -112,20 +112,26 @@ export class Signin extends PureComponent<Props, State> {
...
-        )
-        setToLocalStorage('redirectTo', window.location.href)
-        window.location.href = url.href
-        throw error
-      }
 
       if (CLOUD) {
+        if (isFlagEnabled('useQuartzLogin')) {
+          window.location.reload()
+          throw error
```
hoorayimhelping, line 119: like we discussed offline, I think we should remove this line since it won't really do anything and it tends to communicate that is doing something

src/Signin.tsx
```diff
@@ -112,20 +112,26 @@ export class Signin extends PureComponent<Props, State> {
...
+         * We'll need this authSessionCookieOn flag off for tools until
+         * Quartz is integrated into that environment
+         */
+        if (isFlagEnabled('authSessionCookieOn')) {
+          const url = new URL(
+            `${window.location.origin}${CLOUD_LOGIN_PATHNAME}?redirectTo=${window.location.href}`
+          )
+          setToLocalStorage('redirectTo', window.location.href)
+          window.location.href = url.href
+          throw error
```
hoorayimhelping, line 132: same here with this line

hoorayimhelping (commented):
couple of comments about removing unneeded lines

PR #4702, opened 2022-05-31
chore(HelpBar): move text inside overlay body

blegesse-git:
New info icon inside free-tier support overlay needs to be correctly aligned. 
Pr moves the `p` tag containing status text and icon inside `Overlay.Body`. Check out screenshot below to see the change. 

Before: 
icon

After: 
Screen Shot 2022-05-31 at 9 07 54 AM


ChitlangeSahas (approved):
Looks good!

PR #4703, opened 2022-05-31
chore(HelpBar): add banner panel with text

blegesse-git:
Pr adds a banner pannel for payg support confirmation overlay. 

Screen Shot 2022-05-31 at 9 55 15 AM


src/support/components/ConfirmationOverlay.tsx
```diff
@@ -28,6 +33,25 @@ const ConfirmationOverlay: FC<OwnProps> = () => {
...
+      >
+        <div>
+          For critical issues, please contact the InfluxData support team at the
+          toll-free numbers{' '}
+          <a href="tel:+18559581047" rel="noopener noreferrer" target="_self">
+            +1 855-958-1047
+          </a>{' '}
+          and{' '}
+          <a href="tel:+448081966148" rel="noopener noreferrer" target="_self">
+            +44 808-196-6148
```
hoorayimhelping, line 50: Two things:

I think we should pull these phone numbers out into `shared/constants/index` so if they need to be referenced in the future, we can reference them. And if they never get referenced, I think it's still good to document what they are. It looks like we'll probably need two variables for each phone number - one for the href and one for display.

I don't think we need a target of `_self` or a rel of `noopener noreferrer` here. A `_self` target is the default behavior, and since we're not opening a blank link, we don't need the `rel`.

hoorayimhelping (commented):
What do yo think about storing these phone numbers in constants?

src/support/components/ConfirmationOverlay.tsx
blegesse-git: totally agree with pulling the phone numbers into shared folder. 

 I added `_self` initially because I was using the `SafeBlankLink` component (which has a set `_target` property) . But like you said, since we aren't opening a new tab, there is no need for them. consider them removed

src/shared/constants/index.ts
```diff
@@ -168,3 +168,8 @@ export const PAYG_MAX_CREDIT = 250
 export const PAYG_CREDIT_EXPERIMENT_ID = process.env.PAYG_CREDIT_EXPERIMENT_ID
 export const CREDIT_250_EXPERIMENT_ID = 'JVfGg-umSiSk-JVpeh4hGA'
 export const CREDIT_250_EXPERIENCE_CUTOFF_EPOCH = 1646697600000 // 8 March 2022
+
+export const INFLUXDATA_SUPPORT_CONTACT_UK = '+44 808-196-6148'
+export const INFLUXDATA_SUPPORT_CONTACT_UK_NO_SPACE = 'tel:+448081966148'
```
hoorayimhelping, line 173: nitpick: `NO_SPACE` might be more descriptive if it's called `TELEPHONE_LINK` since that's what the `tel:` prefix is. so like 

```ts
INFLUXDATA_SUPPORT_CONTACT_UK_TELEPHONE_LINK
INFLUXDATA_SUPPORT_CONTACT_US_TELEPHONE_LINK
```

hoorayimhelping (approved):
nice work!

PR #4704, opened 2022-05-31
feat(HelpBar): connect feedback form with feedback link

blegesse-git:
Inside free tier contact support, the link directing user to feedback and questions form has an empty href. 
Pr adds an `onClick` property that opens the Feedback and Questions form. 

https://user-images.githubusercontent.com/66275100/171212858-97e0329c-ebb1-46e6-ac53-2b2a5f1b4738.mov

### DIFF
--- a/src/support/components/FreeAccountSupportOverlay.tsx
+++ b/src/support/components/FreeAccountSupportOverlay.tsx
@@ -1,6 +1,6 @@
 // Libraries
 import React, {FC, useContext} from 'react'
-import {useSelector} from 'react-redux'
+import {useDispatch, useSelector} from 'react-redux'
 
 // Components
 import {Icon, IconFont, List, Overlay} from '@influxdata/clockface'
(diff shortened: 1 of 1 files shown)


src/support/components/FreeAccountSupportOverlay.tsx
```diff
@@ -32,6 +35,15 @@ const FreeAccountSupportOverlay: FC<OwnProps> = () => {
   const {onClose} = useContext(OverlayContext)
   const isCredit250ExperienceActive = useSelector(shouldGetCredit250Experience)
 
+  const dispatch = useDispatch()
+
+  const openFeedbackOverlay = (event): void => {
+    event.preventDefault()
```
hoorayimhelping, line 41: I don't think the `event.preventDefault()` is necessary here. What happens when you remove it?

Also, do we want an `event` here?

hoorayimhelping (commented):
Nice find! A couple of questions

src/support/components/FreeAccountSupportOverlay.tsx
hoorayimhelping: @blegesse-git cool, looks like the move is to leave the `eventPrevent` :) (nice answer too heh)

do we need an app event like `event('helpBar.freeAccount.supportOverlay.shown')`

blegesse-git: haha! yes. i didn't realize i answered your question with a question. 
and i thought about that too. I think adding an event metrics there could be helpful. But that means adding event metrics for both payg and free account overlays  inside `TreeNav` component as well. Which is a-okay with me. I'll get the fix up soon

hoorayimhelping (approved):

Issue #4707, opened 2022-05-31
Flaky Test: Tasks


kristinarobinson:
Look at flaky CI tests for Tasks

PR #4711, opened 2022-05-31
test: press {enter} explicitly in tasks test to avoid invalid query in cypress test

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4707

### The cause

The problem is with the `cy.type` and how monaco editor is responding with the auto-complete. 

When cypress test types `import "` the monaco editor attempts to autocomplete with `import ""` (auto closing the double quotes). Sometimes monaco editor seems to glitch out and misplace the cursor, causing cypress to enter an invalid query. 

Screen Shot 2022-05-31 at 12 13 28 PM

Screen Shot 2022-05-31 at 11 36 08 AM

(note the missing end quote)

--- 

### The Solution: 

My attempt to fix it does two things: 

1. Force an `{enter}` after the first line so that it is explicit.
2. Make the `delay` between keystrokes to be 0, this seems to fix the autocomplete issue because it doesn't give monaco editor / lsp the time to think too much (?). The reasoning of why it works is a hypothesis of mine. 

Proof of it working: (50 runs with no flakes)

 
Screen Shot 2022-05-31 at 12 16 35 PM

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -1056,7 +1056,7 @@ export const createTaskFromEmpty = (
   interval: string = '24h',
   offset: string = '20m',
   force: boolean = true,
-  delay: number = 10
+  delay: number = 15
 ) => {
   cy.getByTestID('create-task--button')
     .first()
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -1056,7 +1056,7 @@ export const createTaskFromEmpty = (
   interval: string = '24h',
   offset: string = '20m',
   force: boolean = true,
-  delay: number = 15
+  delay: number = 0
 ) => {
   cy.getByTestID('create-task--button')
     .first()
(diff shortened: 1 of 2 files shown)


wdoconnell:
Great catch on the autocomplete! 50x runs is great proof this resolves the flakiness.

cypress/support/commands.ts
```diff
@@ -1056,7 +1056,7 @@ export const createTaskFromEmpty = (
   interval: string = '24h',
   offset: string = '20m',
   force: boolean = true,
-  delay: number = 10
+  delay: number = 0
```
hoorayimhelping, line 1059: this is also used in the [pinned items test](https://github.com/influxdata/ui/blob/f6e0f246bed706a342a322df0d90e5ae0442c7e1/cypress/e2e/cloud/pinned.test.ts#L144) so we should keep an eye out for any weird behavior there as a result of this change

hoorayimhelping (approved):

PR #4716, opened 2022-05-31
feat: when using group aggregate function, set default to all grouping tag values

### LINKED ISSUE #4635: Show default grouping, and enable users to turn off grouping in data explorer
This has been requested by Solana. They have a tag (host_id) with 1500 values, and when they use the data explorer, the series comes back grouped by host_id, so they get 1500 series back.  They want to get back a single series and do not know how to change the flux to do that.

The proposed approach is to change the UI to a) show you what tags are being included in the group, and b) allow the user to deselect all the tags so that the data is no longer grouped.

So, in the screenshot below, host_id would be checked (as it is in the default group) and if you uncheck it, we would no longer group by host_id.
image (8)

wiedld:
Closes #4635 
Applies to Dashboard and DataExplorer.

## Presumptions made:
* The `|> keys()` transformation will only return those items which are a group key.
* See video of presumed, expected interaction (a.k.a. changes to flux query constructed).

## Not touched:
* Notebooks query builder. 
* New QX flux builder.

## Video evidence:

https://user-images.githubusercontent.com/10232835/171297109-f17a06f5-d238-4460-bc1b-3823c219f286.mp4

### DIFF
--- a/src/timeMachine/reducers/index.ts
+++ b/src/timeMachine/reducers/index.ts
@@ -840,9 +840,20 @@ export const timeMachineReducer = (
           // we want to clear out any previously selected values
           draftQuery.builderConfig.tags[index].values = []
 
+          // but for the group aggregate, want to set default values
+          if (builderAggregateFunctionType === 'group') {
+            draftQuery.builderConfig.tags[
+              index
+            ].values = draftQuery.builderConfig.tags
+              .map(t => t.key)
+              .filter(x => !!x)
+          }
+
           draftQuery.builderConfig.tags[
             index
           ].aggregateFunctionType = builderAggregateFunctionType
+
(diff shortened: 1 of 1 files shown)


wiedld:
Did not do notebooks:
* the existing notebooks code treats GroupBy to apply to values, not tag keys.
   * See vid below.
   * The GroupBy option should be different from the Filter option, since it should be the keys() (not the values for the _field).
* Started to unwind/update, but decided to time box for now.

https://user-images.githubusercontent.com/10232835/171297305-049e3b1c-f7f0-421d-88bc-9d7ca74e5e31.mov

rockstar (approved):
Even smaller than I expected the patch to be!

Issue #18, opened 2020-07-21
Timezone selector is not linked to date picker


desa:
![Screen Shot 2020-07-02 at 12 17 22 PM](https://user-images.githubusercontent.com/1477539/88116417-5d92d880-cb6d-11ea-97c5-341298e0c36b.png)

WHEN viewing a dashboard
AND the user selects a timezone and date
THEN `v.timeRangeStart` and `v.timeRangeStop` should use the timezone specified in the timezone picker

desa:
cc @jacobmarble

Asalem1:
@desa is this still an issue? It should've been addressed here:

https://github.com/influxdata/influxdb/pull/19146

joelfried:
Tested this behavior locally and this problem no longer appears to be occurring.  Specifically I found a dataset with a data point that is in range when I'm on local time but out of range when I'm on UTC and changing does in fact cause the cell to report no more data available.  Thus the queries inside dashboard cells must be respecting the timezone in the drop down.

PR #4720, opened 2022-06-01
fix: use orgs/:orgId endpoint to provide correct cloud provider for organizations page

### LINKED ISSUE #4597: Bug: PAYG Org Settings - Stop displaying Billing Provider under "Provider"

wdoconnell:
Closes #4597

Also obviates need for PR #4607

The Organization/Settings page in UI has a cloud provider field, but it currently displays _billing_ provider information (_zuora_, aws, gcm, azure), not cloud provider information. The root cause is that the UI is using the `billingProvider` property on quartzMe, and the UI doesn't have the cloud `provider` property in global state, since it hasn't been requested from any endpoint.

This fix uses the new `api/v2/quartz/orgs/:orgId` endpoint to retrieve the cloud provider for the current org, and stores it on the object for the current org (`orgs.org`) in global state. The fix is dependent on [quartz-mock #18](https://github.com/influxdata/quartz-mock/pull/18), which is now merged.

The organization page uses (and continues to use) the IDPE org endpoint (`/api/v2/orgs`), _not_ the quartz endpoint, to populate orgs in state. For most use cases prior to multi-org being implemented, this continues to return an array of organizations of `length 1` with the current org, which is populated in state to `resources.orgs.org`. I have not changed this logic.

I understand that there can be edge cases where an array of more than one organization would be returned by the IDPE `/orgs` endpoint for certain users. To ensure the validity of the cloud provider-organization association and avoid confusion, I've proposed some logic to avoid populating the `provider` property when:

- we're not in a cloud environment, 
- for whatever reason, more than one org was returned by the IDPE (`/api/v2/orgs/`) endpoint, 
- or no org was returned. 

In those cases, the UI won't request the cloud provider property or render it to the organizations page. There will likely be need for a more comprehensive solution once all of the logic for multi-org is implemented.

Adding @randycoulman to make sure we're all on the same page. Thanks!

Includes two new unit tests for cloud to check that the cloud provider is being rendered.

Before fix:
168848872-c3b39112-eb6c-42b9-b6a8-e045008cfa94

After fix:
Screen Shot 2022-06-01 at 11 45 19 AM


src/organizations/actions/thunks.ts
```diff
@@ -60,9 +64,23 @@ export const getOrganizations = () => async (
     }
 
     const {orgs} = resp.data
+    const numOrgs = orgs.length
+    const orgsWithCloudProvider: Organization[] = orgs
+
+    // api.getOrgs returns all organizations associated with the currently active account.
+    // Assuming cloud user, and assuming one org returned (no multi-org in cloud yet), retrieve
+    // organization's cloud provider from new org API.
+    if (CLOUD && numOrgs === 1 && Object.keys(orgs[0]).length > 0) {
```
hoorayimhelping, line 73: this code uses an inference - it's inferring that if `Object.keys(orgs[0]).length > 0` then `orgs[0].id` exists. I think it would be better to be more explicit and just check if `orgs[0].id` exists, e.g.

```ts
if (CLOUD && numOrgs === 1 && orgs[0].id) {
```

hoorayimhelping (commented):
nice! have a thought on being explicit about checking for things existing before using it, but I may not understand the full intent.

hoorayimhelping (approved):

randycoulman (commented):
I think this looks OK, but seems like a lot of hassle for something that's going to change very soon. In cloud, the new Quartz `/identity` endpoint will give you back the ID of the current org, so you wouldn't have to fetch all of the user's orgs and hope that there's only one.

Does it make sense to start using `/identity` now for this small part of the overall use case so we don't have to redo this work in a week or two?

Another optional thing you could do here: the response from `/quartz/orgs/:id` also includes the region code and region name, so you could get that information from there instead of having to pull in `quartzMe`.

I'll leave final approval to someone more familiar with this codebase. I'm OK with what's here, but wondering if it would be worth moving more in the direction we're about to go.

hoorayimhelping:
@wdoconnell Randy's comment might actually be the right move forward. Since we need to use `/identity` anyway, and this is to fix a minor bug, it might be better to just do that

wdoconnell:
I like Randy's idea better. Let me play with it and do this with `identity.`

randycoulman:
@wdoconnell I'd like to be able to tell you that you could use Quartz' `/identity` and `/orgs/:id` APIs entirely, but you wouldn't (currently) get back the org's name that way. Soon, but not just yet. But you'd be able to use IDPE's `/orgs/:id` endpoint rather than having to fetch the list of orgs. All of this is Cloud-only, of course, though.

wdoconnell:
Closing this PR because it will be fixed by a separate ticket utilizing the data from `state.identity` that is pulled from `/quartz/identity`.

Issue #4694, opened 2022-05-27
very minor bug: graph error message hidden by "copy" button


philjb:
## About the bug

When a flux error occurs that prevents plotting, the error shows in the cell, but the copy button hides part of the message.

![Screen Shot 2022-05-27 at 11 11 55 AM](https://user-images.githubusercontent.com/13472206/170779541-7d571737-528e-464d-ba14-136cd64601f7.png)

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. write invalid flux in a cell
2. see error message

**Expected behavior:**
see full error text
**Actual behavior:**
button hides text

**Visual Proof:**
see above

## About your environment

**Environment info:**
cloud2

**Config:**

na

**Logs:**

na
**Performance:**
na

PR #4721, opened 2022-06-01
fix: better error message display in data explorer

ChitlangeSahas:
Closes [#4694](https://github.com/influxdata/ui/issues/4694)

Before: 

Screen Shot 2022-06-01 at 8 23 26 AM

After: 

https://user-images.githubusercontent.com/18511823/171439831-5e516139-236a-4edf-a810-d110e18166a2.mov

### DIFF
--- a/src/shared/components/EmptyGraphError.scss
+++ b/src/shared/components/EmptyGraphError.scss
@@ -1,6 +1,8 @@
 .empty-graph-error {
   width: 100%;
   height: 100%;
+  display: flex;
+  justify-content: space-between;
   border-radius: $cf-radius;
   background-color: $cf-grey-5;
   color: $c-dreamsicle;
@@ -25,5 +27,6 @@
 }
 
 .empty-graph-error--scroll {
+  width: 90% !important;
   margin-top: $cf-space-s;
 }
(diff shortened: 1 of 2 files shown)


src/shared/components/EmptyGraphError.scss
```diff
@@ -25,5 +27,6 @@
 }
 
 .empty-graph-error--scroll {
+  width: 90% !important;
```
hoorayimhelping, line 30: is it possible to apply this style without `!important`, by being more [specific?](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity)

For example, do

```css
div.empty-graph-error--scroll {
  width: 90%
}
```
or 

```css
.empty-graph-error--scroll pre {
  width: 90%
}
```

or

```css
.empty-graph-error--scroll pre code {
  width: 90%
}
```
work?

hoorayimhelping (approved):
lgtm, but if possible, we should try to avoid using `!important`

src/shared/components/EmptyGraphError.scss
ChitlangeSahas: I agree! and I tried those, I think the issue is that the DappersScrollBar component in clockface is overriding the styles.

One way around it is to pass `style` prop to the `DappersScrollBar` component, would that be preferred? 

Screen Shot 2022-06-01 at 8 33 21 AM

PR #4723, opened 2022-06-01
chore: use built-in functions in flux schema package to query

appletreeisyellow:
This PR updates queries to use the built-in functions in [flux schema package](https://docs.influxdata.com/flux/v0.x/stdlib/influxdata/influxdb/schema/) to get values for measurements, fields, tag keys, and tag values. 

## What got changed

1. Replace the generic function `tagKeys()` and `tagValues()` with the built-in functions `measurements()`, `measurementFieldKeys()`, `measurementTagKeys()`, `measurementTagValues()`
2. Swap the order of `sort()` and `limit()`. Now the query do sort first, then limit


rockstar (approved):
This is a great refactor. Thanks for chasing it.

PR #4724, opened 2022-06-01
fix: fix subscriptions search and add test

subirjolly:
Closes https://github.com/influxdata/data-acquisition/issues/333

## BEFORE
BEFORE

## AFTER
AFTER

### DIFF
--- a/src/writeData/subscriptions/components/SubscriptionsLanding.tsx
+++ b/src/writeData/subscriptions/components/SubscriptionsLanding.tsx
@@ -51,15 +51,30 @@ const SubscriptionsLanding: FC = () => {
     sortType: SortTypes.String,
     sortDirection: Sort.Ascending,
   })
+  const [filteredSubscriptions, setFilteredSubscriptions] = useState([])
+
   useEffect(() => {
     event('visited subscriptions page', {}, {feature: 'subscriptions'})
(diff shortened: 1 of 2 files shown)


src/writeData/subscriptions/components/SubscriptionsLanding.tsx
```diff
@@ -51,15 +51,30 @@ const SubscriptionsLanding: FC = () => {
...
   }, [])
-  const filteredSubscriptions =
-    subscriptions && search
-      ? subscriptions.filter(
-          s => s.name.toLowerCase() === search.toLowerCase().trim()
-        )
-      : subscriptions
+
+  useEffect(() => {
+    if (subscriptions === null || !subscriptions.length) {
```
genehynson, line 61: suggesting this change to handle `undefined` scenario as well as `null`.

```suggestion
    if (!subscriptions || !subscriptions.length) {
```

genehynson (commented):
just one comment - otherwise looks great! thanks for adding the test and those utility funcs

subirjolly:
> just one comment - otherwise looks great! thanks for adding the test and those utility funcs

committed 🙌

genehynson (approved):
🙌

PR #4727, opened 2022-06-02
feat: add search functionality to schema browser

### LINKED ISSUE #4712: Schema browser: add logic to search fields and tag keys
We want to add the functionality to search fields and tag keys in schema browser. 

Part of #4510 

Screen Shot 2022-05-31 at 2 25 40 PM

appletreeisyellow:
Closes #4712 

This PR adds the search functionality to search for fields and tag keys in schema browser.

## What got changed

1. Add search string in query for a backend search
2. Add debouncer for keyboard to reduce API calls
3. Move constants to a separate file

## Visual Proof

https://user-images.githubusercontent.com/14298407/171646996-003b6aca-c82b-43c1-9b36-2b4790d018ad.mov


src/dataExplorer/context/fields.tsx
```diff
@@ -61,7 +62,11 @@ export const FieldsProvider: FC<Prop> = ({children, scope}) => {
     ? EXTENDED_TAG_LIMIT
     : DEFAULT_TAG_LIMIT
 
-  const getFields = async (bucket: any, measurement: string) => {
+  const getFields = async (
+    bucket: any,
+    measurement: string,
+    searchTerm?: string
```
wiedld, line 68: Why are we typing bucket as `any` type? When we have access to the Bucket type?

appletreeisyellow: Nice catch! Totally agree that we should use type `Bucket` instead of `any`. 

However, if I changed the type to `Bucket`, VS code would shows error message on line 79 and line 96
Screen Shot 2022-06-02 at 4 45 08 PM

As you can see from the screenshot, there should be only 'user' and 'system' for `bucket.type` as [it is defined in openapi](https://github.com/influxdata/openapi/blob/3481defece3e8c70c47f21680857ac1a7c051a8b/src/common/schemas/Bucket.yml#L34-L40)

The reason why can line 79 check `bucket.type === 'sample'` is because the 'sample' buckets are hardcoded into the bucket list in the frontend [here](https://github.com/influxdata/ui/blob/master/src/shared/contexts/buckets.tsx#L152-L172)

A fix I can think of is to add a line `readonly type?: 'user' | 'system' | 'sample' `in [`Bucket` type definition](https://github.com/influxdata/ui/blob/master/src/types/buckets.ts#L5-L8) so that it's syntactically correct to use `Bucket`. 

What do you think? @wiedld

appletreeisyellow: I tired the fix I suggested above in [this commit](https://github.com/influxdata/ui/pull/4727/commits/309d391d55057b8051e93daf6a86e9a5be2276b9). Seems like lint on CircleCI is [complaining](https://app.circleci.com/pipelines/github/influxdata/ui/13546/workflows/732d3150-cd24-4a12-af40-7068932c2b2d/jobs/80137)

Screen Shot 2022-06-02 at 5 38 25 PM

So I think using type `any` is fair unless you have other suggestions. I'm open to try!

wiedld: I think this is typescript doing it's job. And pointing out that the hardcoded sample buckets are not fitting the Bucket type. 😆 

That hardcoding is taking an existing bucket payload from the backend, and then hardcoding a subset of example buckets (without it being bucket CRUD objects) in the frontend, while adding an additional "type == sample" to the frontend. Then all our special case handling (of how we get this to work) looks for:
* the featureFlag `newQueryBuilder` being present.
* AND the bucket being this special Bucket object we created (which doesn't follow the type contract).

Ramifications of this decision:
* once we take this thing live, we have to update all the other conditional paths to handle non "type == sample" buckets. 
* we are not testing (as we dev) buckets which will exist in the real world. A.k.a. the other typed buckets -- such as system buckets, user buckets, etc.
* Is this what we want?

Alternatively, may I suggest:
* in this section: https://github.com/influxdata/ui/blob/master/src/shared/contexts/buckets.tsx#L152-L172
* If we have the featureFlag `newQueryBuilder` AND the sample buckets we seek are not present:
   * seed your sample buckets as `type: user` or `type: system`.
   * type the sample buckets per the expected API contract.
   * Then make sure all downstream code does not break.

Does this make sense?

drdelambre: sample buckets are not api resources, and are not used as such for code generation which is where you see the branching off of the bucket type (cannot use a `from` but instead need to use a `sample`)

src/dataExplorer/context/newDataExplorer.tsx
```diff
@@ -8,14 +15,17 @@ import {TagsContext} from 'src/dataExplorer/context/tags'
...
-export const IMPORT_REGEXP = 'import "regexp"\n'
-export const IMPORT_INFLUX_SCHEMA = 'import "influxdata/influxdb/schema"'
-export const SAMPLE_DATA_SET = (bucketID: string) =>
-  `import "influxdata/influxdb/sample"\nsample.data(set: "${bucketID}")`
-export const FROM_BUCKET = (bucketName: string) =>
-  `from(bucket: "${bucketName}")`
+const DEBOUNCE_TIMEOUT = 500
+let timer
+type NOOP = () => void
+const debouncer = (action: NOOP): void => {
```
drdelambre, line 21: not for this PR, just for your reference... i dont think this is _technically_ a debouncer. we have a bunch of message throttle controls we just refer to as a debouncer. there's a bunch of specific names for them that better describe their behavior if you wanna go look them up. this one just waits for a lull in the user activity and forces the `action` function to hold a sense of the source of truth (proxy for the input event always returning the WHOLE value). this simplistic model helps us get around weird cases where you deleted something, but the last update event in the debouncer was to add something, and the user sees ghosts typing in their input boxes! here's an even more complicated one that combines messages instead of dropping them like this one does:
https://github.com/influxdata/ui/blob/master/src/cloud/utils/reporting.ts#L78

and here is the lodash version many libraries implement which calls a function at MOST every X ms which is good for things like tracking fast moving pointer events without clobbering a render cycle:
https://github.com/lodash/lodash/blob/master/debounce.js

you can even rewrite one of those to connect to the next frame render with `window.requestAnimationFrame` instead of chosing an arbitrary time to decide your debounce frequency! you'll find that version more in rendering libraries and the like

drdelambre (approved):

src/dataExplorer/context/fields.tsx
appletreeisyellow: Thanks @wiedld! I continue the suggestion from @drdelambre and fixed the complains from Typescript. Now `Bucket` type is working well!

src/dataExplorer/shared/utils.ts
```diff
@@ -0,0 +1,12 @@
+export const LOCAL_LIMIT = 8
+export const IMPORT_REGEXP = 'import "regexp"\n'
+export const IMPORT_INFLUX_SCHEMA = 'import "influxdata/influxdb/schema"'
+
+export const SAMPLE_DATA_SET = (bucketID: string) =>
+  `import "influxdata/influxdb/sample"\nsample.data(set: "${bucketID}")`
```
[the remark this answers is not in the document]
appletreeisyellow: You are right on non-sample buckets! Sample bucket list always has id as it is hardcoded [here](https://github.com/influxdata/ui/blob/master/src/shared/contexts/buckets.tsx#L152-L172)

src/dataExplorer/context/fields.tsx
```diff
@@ -13,28 +13,29 @@ import {
 } from 'src/shared/constants/queryBuilder'
 
 // Contexts
+import {QueryContext} from 'src/shared/contexts/query'
```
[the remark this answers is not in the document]
wiedld: Some tech debt, which I am not sure if you want to quickly update in this issue. We have the QueryProvider wrapped twice around the Schema.
* in `src/dataExplorer/components/newDatqExplorer`
* then again in in `src/dataExplorer/components/Schema`

wiedld:
Some tech debt, which I am not sure if you want to quickly update in this PR. We have the QueryProvider wrapped twice around the Schema.
* in `src/dataExplorer/components/newDataExplorer`
* then again in in `src/dataExplorer/components/Schema`

src/dataExplorer/context/newDataExplorer.tsx
appletreeisyellow: Good to know! Even though I'm still digesting on this :)

wiedld (approved):
Approving with the caveat being discussed in the slacks now. That we are treating csv as a Bucket object. (But that is a question tangential to this PR...although we may update some of this code afterwards.)

drdelambre:
the code should support as many nested query providers as you need. lemme know if that changes. same goes for the bucket provider. by externalizing how the important state for these providers is stored for these universal type contexts (where the provider is acting like a singleton), we should be able to increase code reusability / modularity. also read: you shouldn't have to know about the entire render tree to include a component, and a component shouldn't know it's entire ancestry to determine if it's resource  requirements have been fulfilled. components should say "i need buckets to work" and that should just work no matter where the component is injected. we have this issue currently with the get resource component  and need to divest from those patterns immediately as  they require every component tree  underneath them  to be sanitized in all contexts  from inclusion of resource fetches, meaning the developer needs to first know about this condition an then actively look for it / maintain that model while developing, which doesn't happen.

appletreeisyellow:
@drdelambre Good to know! 

I tested and so far removing the query providers in [recent commit](https://github.com/influxdata/ui/pull/4727/commits/01fe90eb56cbb857f216ca99643fa084a7a6b972) doesn't change anything since we are just using the `queryAPI` function from the query context

PR #4700, opened 2022-05-30
fix: wrap strings in double quotes so strings with commas can render in Table / SimpleTable

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/4583

https://user-images.githubusercontent.com/18511823/171056239-9f6b6868-9a72-4094-b3d3-9e4f3459b67d.mov

PR #4733, opened 2022-06-02
fix: handle JSON Objects in parseFromFluxResults

### LINKED ISSUE #3637: Bug Duty: Table and SimpleTable break when column is a nested object
## About the bug

When the column value is a nested object, `parseFromFluxResults` doesn't work correctly. This is seen in `SimpleTable` as well as `Table` graph. 

image

### LINKED ISSUE #3736: Dashboards - markdown editor - Emoji, strikethrough
## About the bug

**Steps to reproduce:**
1. Create a dashboard
2. Add a note
3. Use markdown editor
4. Type any :emoji:
5. Use tilde symbols to add strikethrough

**Expected behavior:**
image
image

**Actual behavior:**
image

## About your environment

**Environment info:**
Testing with remocal on a UI project (Ubuntu 20.04)

ChitlangeSahas:
Closes #3637 

In https://github.com/influxdata/ui/pull/4700, we handled a special case where the data in the column can be a string of `CSV` data. 

In this one, we handle the case when the data is a `JSON` object. 

The approach is to take the JSON string, remove all the double quotes and replace them with single quotes, (this saves the csvParser from being thrown off) and then wrap the JSON object in double quotes. 

Example
-> `{"key" : "value"}` came in from the flux query
-> `{'key': 'value'}` : we replace double quotes with single quotes
-> `"{'key': 'value'}"` : we wrap the string form of the JSON object in double quotes so we can display it.

----

### Proof of it working:

Please look at the `_value` column in all the screenshots for the demo of JSON and CSV strings working.

**Simple CSV in Simple Table:** 

Screen Shot 2022-06-02 at 4 01 23 PM

Screen Shot 2022-06-02 at 4 05 17 PM

**Simple String in Simple Table:** 

Screen Shot 2022-06-02 at 4 04 22 PM

**JSON in Simple table:** 

Screen Shot 2022-06-02 at 4 01 18 PM

Screen Shot 2022-06-02 at 4 06 50 PM

**Complex result with multiple JSON objects and normal strings**

Screen Shot 2022-06-02 at 4 01 09 PM

### DIFF
--- a/src/timeMachine/utils/rawFluxDataTable.ts
+++ b/src/timeMachine/utils/rawFluxDataTable.ts
@@ -29,16 +29,12 @@ export const parseFromFluxResults = (
   const groupSet = new Set(fluxGroupKeyUnion)
   let max = 0
 
+  // checks whether the string is valid JSON object or not
   const isJsonObject = jsonString => {
     try {
-      const o = JSON.parse(jsonString)
-
-      // Handle non-exception-throwing cases:
-      // Neither JSON.parse(false) or JSON.parse(1234) throw errors, hence the type-checking,
-      // but... JSON.parse(null) returns null, and typeof null === "object",
-      // so we must check for that, too. Thankfully, null is falsey, so this suffices:
-      if (o && typeof o === 'object') {
-        return o
+      const object = JSON.parse(jsonString)
+      if (object && typeof object === 'object') {
+        return object
       }
     } catch (_) {}
 
--- a/src/timeMachine/utils/rawFluxDataTable.ts
+++ b/src/timeMachine/utils/rawFluxDataTable.ts
@@ -29,6 +29,22 @@ export const parseFromFluxResults = (
   const groupSet = new Set(fluxGroupKeyUnion)
   let max = 0
 
+  const isJsonObject = jsonString => {
+    try {
+      const o = JSON.parse(jsonString)
+
+      // Handle non-exception-throwing cases:
+      // Neither JSON.parse(false) or JSON.parse(1234) throw errors, hence the type-checking,
+      // but... JSON.parse(null) returns null, and typeof null === "object",
+      // so we must check for that, too. Thankfully, null is falsey, so this suffices:
+      if (o && typeof o === 'object') {
+        return o
+      }
+    } catch (_) {}
+
+    return false
+  }
+
   for (let i = 0; i < tables.length; i++) {
     if (values[i] !== currVal || tables[i] !== currTable) {
       // sets the boundaries for the chunk based on different yields or tables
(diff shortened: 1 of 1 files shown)


src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -29,6 +29,18 @@ export const parseFromFluxResults = (
   const groupSet = new Set(fluxGroupKeyUnion)
   let max = 0
 
+  // checks whether the string is valid JSON object or not
+  const isJsonObject = jsonString => {
+    try {
+      const object = JSON.parse(jsonString)
+      if (object && typeof object === 'object') {
+        return object
```
hoorayimhelping, line 37: I think this should `return true`. An object is truthy, but it's good to be explicit in these cases and not rely on weird JS truthiness rules.

src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -29,6 +29,18 @@ export const parseFromFluxResults = (
...
+      }
+    } catch (_) {}
```
hoorayimhelping, line 39: I think you can just do `catch {}`

src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -72,7 +84,16 @@ export const parseFromFluxResults = (
...
       }
-      if (typeof columnData === 'string') {
+
+      // (Sahas):
+      // the columnData can have a comma (,) in two cases
+      // 1. it's a JSON Object
+      // 2. It's a string of CSV
+      if (isJsonObject(columnData)) {
+        // replace Double quotes \" with single quotes \' in a json object
+        // columnData = `"${columnData.replace(/['"]+/g, '\'')}"`
```
hoorayimhelping, line 94: this line can be ✂️, right?

hoorayimhelping (commented):
A couple of suggestions, let me know what you think (and request a re-review when you're ready for one)

src/timeMachine/utils/rawFluxDataTable.ts
ChitlangeSahas: ✅

src/timeMachine/utils/rawFluxDataTable.ts
ChitlangeSahas: Got it, fixed!

src/timeMachine/utils/rawFluxDataTable.ts
ChitlangeSahas: Yes, fixed 👍

hoorayimhelping (approved):

Issue #4591, opened 2022-05-16
Injection of Schema Selections into Flux Editor


influx-product-manager:
## User Story

<p></p>

As a user, I want my flux script started for me by using the metrics that I have chosen in the schema browser, so that I don't have to type as many lines of code.

<p></p>

## Solution Overview

<p></p>

The schema browser (at the end of the basic schema browser user story) allows the user to select a bucket and measurement and shows them a list of fields, tags, and tag values.  This story adds the ability to hover over a field, tag, or tag value and select an "Inject" button that will appear.  

<p></p>

## Acceptance Criteria

<p></p>

- When a user hovers over a field, tag, or tag value in schema browser, an "inject" button will appear

- Upon clicking on the inject button, the software will formulate the appropriate flux script lines per the following rules:

  - If no current from/range/measurement filter exists in the flux script then a from range statement will be added using the bucket that is currently selected and a filter statement will be added for the currently selected measurement

  - If a from/range statement already exists, but no measurement in the script, then a measurement filter will be added

- When the object being injected is a field, the software fill formulate the appropriate flux script line  per the following rules:

  - If a filter line exists already for a field, the injected field will be added to the existing flux filter line for that field with an "OR" and the new field name.

  - If a filter line does not already exist for a field in this search, then a filter line will be added for this field.

- When the object being injected is a Tag, the software fill formulate the appropriate flux script line  per the following rules:

  - If a filter line exists already for a tag, the injected tag will be added to the existing flux filter line for tags with an "OR" and the new tag name.

  - If a filter line does not already exist for a tag in this search, then a filter line will be added for this tag.

- When the object being injected is a Tag value, the software fill formulate the appropriate flux script line  per the following rules:

  - If a filter line exists already for a tag value, the injected tag value will be added to the existing flux filter line for tags with an "AND" and the new tag value name.

  - If a filter line does not already exist for a tag value in this search, then a filter line will be added for this tag value.

- If a user changes bucket or measurement when there is already a from range and filter for that bucket/measurement and attempts to do an inject of a field, tag, or tag value - then the user will be shown a dialog that asks them whether they want to replace the existing script or append to the existing script and based upon that selection the software will either replace or append.

- When a from range filter is injected into the script, a pivot and yield flux statement will also be inserted (need to discuss with QX team)

<p></p>

## Notes

<p></p>

- This A/C needs reviewed by QE team for accuracy.

- The Replace/Append item needs UI/UX Design

<p></p>

[...]

Issue #4658, opened 2022-05-24
QX: long term fix, for the function parameters.


wiedld:
- [ ] #4803

See original ticket here:
https://github.com/influxdata/flux/issues/4720

PR #4735, opened 2022-06-03
feat(4658/4709): sending executeCommand payload to LSP to trigger code injection

wiedld:
Part of #4591 

## Goal:
* Enable schema code injection for the new QX FluxBuilder.  
   * This PR provide the interface to be used, for schema injection.
   * `EditorContext.injectViaLsp()` for use by all downstream injections.
   * Use differently typed payloads, per what get's injected.

## Approaches assessed, and rejected, for the UI side of code injection:
* monaco-editor has commands and actions.
  * these do not seem to be a 1-to-1 mapping with the spec executeCommand and codeActions.
* editor.addCommand(keyBinding, args => stuff)
  * reasons rejected:
     * bound to keys, cannot be triggered by name/enum
* editor.executeCommand()
  * reasons rejected:
     * is about asking the editor to execute a command, from a list of existing commands
     * ICommand accepted:  https://github.com/microsoft/monaco-editor/blob/2c5c9636a276ffd00d78d8d7f9bfb13fb7740f77/website/typedoc/monaco.d.ts#L2181-L2195
* editor.addAction()
   * good:
       * does bind to provided enum/name. a.k.a. `ExecuteCommand.InjectFunction`
       * added actions are treated as commands that the monaco-editor can execute.
          * is recognized in monaco-editor `StandaloneCommandService.executeCommand()` when using editor.trigger()
   * reason rejected:
       *  although the type signature suggests a payload can be passed...it cannot.
       * not passing payload:
          * `editor.trigger('', ExecuteCommand.InjectFunction, payload)`
          * `editor.getAction(actionId).run(payload)`
       * found filed tickets with similar complaints. 

## Roundtrip of payloads:
If use injectMeasurement, then will get:
1. to LSP:
  * ``` {"jsonrpc":"2.0","method":"workspace/executeCommand","params":{"command":"injectMeasurementFilter","arguments":[{"bucket":"my-bucket","name":"my-measurement","textDocument":{"uri":"inmemory://model/2"}}],"workDoneProgressParams":{"workDoneToken":null}},"id":"0.510533937643387"} ```
2. response from LSP:
  * ``` {"jsonrpc":"2.0","method":"workspace/applyEdit","params":{"edit":{"changes":{"inmemory://model/2":[{"newText":"from(bucket: \"my-bucket\") |> filter(fn: (r) => r._measurement == \"my-measurement\")","range":{"end":{"character":83,"line":1},"start":{"character":0,"line":0}}}]}}},"id":0}```
3. to LSP:
  * ```{"jsonrpc":"2.0","id":0,"result":{"applied":true}}```

## Not done yet (future PRs):
- Scheme injection for the new QX DataExplorer. (Chunchun has a branch, I believe.)
- far future:
  * injection of user defined variables, excluding windowPeriod
  * injection of windowPeriod variable

### DIFF
--- a/src/languageSupport/languages/flux/lsp/utils.ts
+++ b/src/languageSupport/languages/flux/lsp/utils.ts
@@ -2,7 +2,6 @@ import {
   ProtocolRequestType,
   ProtocolNotificationType,
 } from 'vscode-languageserver-protocol'
-import * as MonacoTypes from 'monaco-editor/esm/vs/editor/editor.api'
 
 export const makeRequestType = (method: string) => {
   return new ProtocolRequestType<any, any, any, any, any>(method)
@@ -56,7 +55,6 @@ export const didOpen = (uri: string, text: string, version: number) => {
 interface ExecuteCommandParams {
   textDocument: {
     uri: string
-    position: MonacoTypes.Position
   }
 }
 
--- a/src/languageSupport/languages/flux/lsp/utils.ts
+++ b/src/languageSupport/languages/flux/lsp/utils.ts
@@ -2,6 +2,8 @@ import {
   ProtocolRequestType,
   ProtocolNotificationType,
 } from 'vscode-languageserver-protocol'
+import * as MonacoTypes from 'monaco-editor/esm/vs/editor/editor.api'
+import {FluxDocs} from 'src/types'
 
 export const makeRequestType = (method: string) => {
   return new ProtocolRequestType<any, any, any, any, any>(method)
@@ -19,6 +21,15 @@ const createNotification = (method: string, params: object = {}) => {
     jsonrpc: JSONRPC,
     method,
     params,
+  }
+}
+
+const createRequest = (method: string, params: object = {}) => {
+  return {
+    jsonrpc: JSONRPC,
+    method,
+    params,
+    id: Math.random().toString(),
   }
 }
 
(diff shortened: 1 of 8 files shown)


src/shared/contexts/editor/index.tsx
```diff
@@ -116,6 +141,10 @@ export const EditorProvider: FC = ({children}) => {
           ? getFluxExample(rawFn as FluxFunction)
           : (rawFn as FluxFunction)
 
+      if (isFlagEnabled('injectionViaLsp')) {
+        return injectViaLsp(ExecuteCommand.InjectFunction, fn)
+      }
```
wiedld, line 146: Since this EditorContext is used everywhere we have a monaco-editor in WRITE mode, we only need to update the injection points within this context.

Here is the function injection.
We have another EditorContext method for injecting variables.

src/dataExplorer/components/NewDataExplorer.tsx
```diff
@@ -19,20 +20,22 @@ const NewDataExplorer: FC = () => {
...
-        <DraggableResizer.Panel>
-          <Schema />
-        </DraggableResizer.Panel>
-        <DraggableResizer.Panel className="new-data-explorer-rightside">
-          <ResultsProvider>
-            <ResultsPane />
-          </ResultsProvider>
-        </DraggableResizer.Panel>
-      </DraggableResizer>
+      <EditorProvider>
```
wiedld, line 23: Only added the EditorProvider.
Nothing else changed.

src/shared/contexts/editor/index.tsx
```diff
@@ -164,6 +199,7 @@ export const EditorProvider: FC = ({children}) => {
         inject,
         injectFunction,
         injectVariable,
+        injectViaLsp,
```
wiedld, line 202: Exposed for use in new DataExplorer. For injection of whatever.

src/dataExplorer/components/NewDataExplorer.tsx
wiedld: Because all injection into the editor will come through the `EditorContext.injectViaLsp()`.
This includes schema injection.
This is the lowest point in the hierarchy which also includes the schema browser.

src/shared/contexts/editor/index.tsx
wiedld: Note that the issue ticket for function injection (#4658) is about doing LSP function injection across the application. This is why the conditional here is impacting everywhere across the UI application, once the feature flag is on.

Whereas the schema injection (and further use of `EditorContext.injectViaLsp()` will be selectively used in the new QX DataExplorer. This matches the requirements for that part of the product.

src/shared/contexts/editor/index.tsx
```diff
@@ -15,16 +15,30 @@ import {
...
+    editor: EditorType,
+    conn: React.MutableRefObject<LspConnectionManager>
+  ) => void
   inject: (options: InjectionOptions) => void
   injectFunction: (fn, cbToParent) => void
   injectVariable: (variableName, cbToParent) => void
+  injectViaLsp: (
+    cmd: ExecuteCommand,
+    data: Omit<ExecuteCommandArgument, 'textDocument'>
+  ) => void
```
wiedld, line 41: This is the method which would be used, to inject the schema.

```
const {injectViaLsp} = useContext(EditorContext)

// look at the type definitions for ExecuteCommandArgument, for the payload
injectViaLsp(ExecuteCommand.injectMeasurement, payload) 
```

Note that the LSP injection requires that the bucket, measurement, field, and value EACH be injected one at a time...in series.

src/languageSupport/languages/flux/lsp/utils.ts
```diff
@@ -22,6 +24,15 @@ const createNotification = (method: string, params: object = {}) => {
   }
 }
 
+const createRequest = (method: string, params: object = {}) => {
+  return {
+    jsonrpc: JSONRPC,
+    method,
+    params,
+    id: Math.random().toString(),
```
wiedld, line 32: Requirement is for a unique ID.
Does not need to be in the same monotonically increasing order as the unique ID from the monaco-editor-initiated events.

src/languageSupport/languages/flux/lsp/utils.ts
```diff
@@ -43,6 +53,48 @@ export const didOpen = (uri: string, text: string, version: number) => {
   })
 }
 
+interface ExecuteCommandParams {
+  textDocument: {
+    uri: string
+    position: MonacoTypes.Position
```
rockstar, line 59: Didn't we not need `position`?

src/languageSupport/languages/flux/lsp/utils.ts
```diff
@@ -58,4 +110,12 @@ export enum Methods {
...
   References = 'textDocument/references',
   Rename = 'textDocument/rename',
+  ExecuteCommand = 'workspace/executeCommand',
+}
+
+export enum ExecuteCommand {
+  InjectionMeasurement = 'injectMeasurementFilter',
+  InjectField = 'injectFieldFilter',
+  InjectTag = 'injectTagFilter',
+  InjectTagValue = 'injectTagValueFilter',
```
rockstar, line 120: I'm glad we have these. I _think_ there might be a way to simplify this serialization, which also might mean these get renamed. I would not consider this API "stable" yet, so this is good that we have these all in the same place.

rockstar (approved):

src/languageSupport/languages/flux/lsp/utils.ts
wiedld: removed.

PR #4740, opened 2022-06-03
refactor: rename `NewDataExplorer` to `FluxQueryBuilder`

### LINKED ISSUE #4674: Flux Query Builder: Rename `NewDataExplorer` to `FluxQueryBuilder`
Currently, we are using `NewDataExplorer` and `newDataExplorer` to refer to the new query experience in the codebase. We decided (on [slack](https://influxdata.slack.com/archives/C030G5783M3/p1652903981896049)) to use the new name `FluxQueryBuilder`

appletreeisyellow:
Closes #4674 

This PR names all the related code thats using `NewDataExplorer` to `FluxQueryBuilder`. Check out the issue for more info on why we want to rename it. 

Note that the feature flag `newDataExplorer` stays the same since it is already set up in idpe and ConfigCat.


asalem1 (changes requested):
Keep the feature flag the same, this isn't backwards compatible

appletreeisyellow:
Yup, the feature flag stays the same

asalem1:
> Yup, the feature flag stays the same

Totally. I can't read. Great job, my bad

asalem1 (approved):

drdelambre:
to the point ari was kinda making, even if it's not a flag, this does have the unintended side effect of turning off the switch for anyone who currently has it on and leaves a dangling state in their local storage. I doubt the impact of that is gonna be big given the projects current status (just a small subset of internal users having access to it), but as a general rule for the future, changing selectors like that should cause pause if not call for a larger migration effort. Again, fine for this PR.

PR #4750, opened 2022-06-06
feat: moving query time over to new status

drdelambre:
moves the query time status information into the space between the editor and the viz

Peek 2022-06-06 10-40


src/dataExplorer/components/ResultsContext.tsx
```diff
@@ -9,25 +9,57 @@ interface ResultsContextType {
...
+  useEffect(() => {
+    let running = false
+    const check = () => {
+      if (!running) {
+        return
+      }
+
+      setTime(Date.now() - timeStart.current)
+
+      window.requestAnimationFrame(check)
```
appletreeisyellow, line 40: so `check` is calling itself recursively until the status gets updated to non-loading?

drdelambre: yes

PR #4754, opened 2022-06-06
feat: add visualizations

drdelambre:
adds visualizations to the new data explorer. design hasn't given a placement for the view options interaction with the function browser yet, so this gives them a head start on how those'll interact

Peek 2022-06-06 13-35


asalem1 (approved):
Would be super cool to get clarity on that launcher

Issue #4709, opened 2022-05-31
Basic Injection of Flux Selections into Flux Editor


influx-product-manager:
## User Story

<p></p>

As a flux developer user, I want to be able to select flux functions and have them inject into my script so that I can more quickly develop flux code.

<p></p>

## Acceptance Criteria

<p></p>

- Functions selected from a flux search get injected into flux script, as is done with current DE (no new requirements, for now)

## Notes

<p></p>

- This will be the same behavior as before, but we need to work out where this is shown in the UI

<p></p>

<p></p>

influx-product-manager:
Link to subfeature: https://influxdata.productboard.com/feature-board/planning/features/13145386

wiedld:
Already done.

PR #4780, opened 2022-06-08
feat(4708/4709): flux browser panel in QX flux query builder

### LINKED ISSUE #4708: Basic Flux Function Search
## User Story

<p></p>

As a flux developer/user, I want to be able to search find flux functions that I may want to use in my flux code, so that I can quickly identify the function name and parameters that I want to use in my script.

<p></p>

## Acceptance Criteria

<p></p>

- An option will be available to search functions (similar to what exists in DE and Notebooks)

<p></p>

## Notes

<p></p>

- For now, this is the same search that is available today, just making it available in the new DE.

<p></p>

wiedld:
Closes #4708 
Part of #4709 

## Done:
* make 3 panels on the FluxQueryBuilder page
  * **did not do the collapsing panels** requested by design, since some questions with the implementation.
* right sidePanel:
  * generalized for any future use
  * for now, only has flux library search
* FluxBrowser
  * uses new MinimalInjectionOption. per design.  
* flux function injection into editor
  * uses the EditorContext, so will be subject to feature flagging for injection-via-LSP

## Future PRs:
* tests
* In the ResultsPane, decide if we want to use the QueryContext
  * it's pulling in a lot previous decisions on variable handling, tech debt on using a cartesian product for the WindowVars, etc.

## Video evidence:

(Updated. Does not have collapsing panels now.)

https://user-images.githubusercontent.com/10232835/172662727-815b7be4-e9da-415f-ba58-b93e0a325e3e.mov


src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -1,38 +1,105 @@
...
+  const adjustMiddlePanel = grow => {
+    if (!grow) {
+      return
+    }
+    const el = document.querySelector(
+      '[data-testid="flux-query-builder-middle-panel"]'
+    )
+    if (el) {
+      (el as any).style.flexGrow = 0.8
+    }
```
wiedld, line 43: I am not a huge fan of this approach. But the problem is mixing the DraggableResizer (which sets the flex-basis on all the panels), with a collapsing button for the leftPanel and rightPanel.

Without this manually triggering of the central panel to flexGrow, it was not expanding to fill the space:

https://user-images.githubusercontent.com/10232835/172515928-446a6ed2-e4f4-4c85-834a-500033c70f85.mov

src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -1,38 +1,105 @@
...
+              alignItems={AlignItems.FlexStart}
+              style={{height: '100%', width: '100%', position: 'absolute'}}
+            >
+              <FlexBox.Child>
+                <span
+                  onClick={() => {
+                    toggleLeftPanel(!showLeft)
+                    adjustMiddlePanel(showLeft)
+                  }}
+                  style={{marginLeft: '-20px', float: 'left'}}
```
wiedld, line 72: Move the collapse icons to appear on top of the DraggableResizer's handle. Per design.

In the right side's divider, I had to set the zIndex to make it actionable (instead of the divider handle).

src/flows/pipes/RawFluxEditor/view.tsx
```diff
@@ -28,8 +28,8 @@ import {VariablesContext} from 'src/flows/context/variables'
 
 // Components
 import SecretsList from 'src/flows/pipes/RawFluxEditor/SecretsList'
-import Functions from 'src/flows/pipes/RawFluxEditor/FunctionsList/GroupedFunctionsList'
-import DynamicFunctions from 'src/flows/pipes/RawFluxEditor/FunctionsList/DynamicFunctionsList'
+import Functions from 'src/shared/components/functions/FunctionsList/GroupedFunctionsList'
+import DynamicFunctions from 'src/shared/components/functions/FunctionsList/DynamicFunctionsList'
```
wiedld, line 32: These are already shared components. Moved to shared dir.

src/flows/pipes/RawFluxEditor/SecretsList.tsx
```diff
@@ -10,8 +10,8 @@ import {
   IconFont,
   JustifyContent,
 } from '@influxdata/clockface'
-import FilterList from 'src/flows/shared/FilterList/FilterList'
-import InjectSecretOption from 'src/flows/shared/FilterList/InjectionOption'
+import FilterList from 'src/shared/components/FilterList/FilterList'
+import InjectSecretOption from 'src/shared/components/FilterList/InjectionOption'
```
wiedld, line 14: These are already shared components. Moved to shared dir.

src/shared/components/FilterList/MinimalistInjectionOption.scss
```diff
@@ -0,0 +1,32 @@
...
+    transition-property: color;
+    &:hover {
+      background: rgba(241, 241, 243, 0.1);
+      /* Note: backdrop-filter has minimal browser support */
+      backdrop-filter: blur(42px);
+      border-radius: 2px;
+      color: $c-galaxy;
+      &:after {
+        content: '+';
+        padding-left: 10px;
```
wiedld, line 28: On hover over the InjectionOption (e.g. a flux function), it does various CSS transforms. Such as adding a `+` to the end of the text.

src/shared/components/FilterList/MinimalistInjectionOption.scss
```diff
@@ -0,0 +1,32 @@
+@import '@influxdata/clockface/dist/variables.scss';
+
+// inherits from the other flux-toolbar--list-item class
+// overwrites certain props
+.flux-toolbar--list-item {
```
wiedld, line 5: Choose to re-use the same classNames (from the shared components), and overwrite what props we need different here. Goal is to try keeping the look as much in sync as possible...except where we have explicitly updated design for this MinimalistInjectionOption.

src/shared/components/functions/perFunction/FluxDocsTooltipContent.tsx
```diff
@@ -20,8 +20,8 @@ const FluxDocsTooltipContent: FC<TooltipProps> = ({
   setHoverdFunction,
 }) => {
   useEffect(() => {
-    setToolTipPopup(true)
-    setHoverdFunction(`${func.package}.${func.name}`)
+    setToolTipPopup && setToolTipPopup(true)
+    setHoverdFunction && setHoverdFunction(`${func.package}.${func.name}`)
```
wiedld, line 24: These are optional props, not used in the MinimalistInjectionOption.
Therefore, these really need an existence check (to not error).

src/shared/components/functions/utils.ts
```diff
@@ -0,0 +1,7 @@
+export const sortFuncs = (a, b) => {
+  if (a.package.toLowerCase() === b.package.toLowerCase()) {
+    return a.name.toLowerCase() < b.name.toLowerCase() ? -1 : 1
+  } else {
+    return a.package.toLowerCase() < b.package.toLowerCase() ? -1 : 1
+  }
+}
```
wiedld, line 7: Common helper function. Extracted.

src/dataExplorer/components/FluxBrowser.tsx
```diff
@@ -0,0 +1,84 @@
...
+  useEffect(() => {
+    if (fluxFunctions.length === 0) {
+      dispatch(getFluxPackages())
+    }
+  }, [])
+
+  const handleSelectItem = useCallback(
+    (func: FluxFunction) => {
+      injectFunction(getFluxExample(func), _ =>
+        console.warn('Determine use of query context.')
```
wiedld, line 42: This is where we normally do a callback to a QueryContext.
But with our new plan for how to inject code into the editor, it may be better to just call `editor.getModel().getValue()` as the src of truth.

src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -1,38 +1,105 @@
...
+                <span
+                  onClick={() => {
+                    toggleRightPanel(!showRight)
+                    adjustMiddlePanel(showRight)
+                  }}
+                  style={{
+                    marginRight: '-20px',
+                    float: 'right',
+                    zIndex: '999',
+                    position: 'absolute',
```
wiedld, line 90: For the right side's collapsing icon, I had to set the zIndex to make it actionable above the panelDraggableHandle.

src/dataExplorer/components/FluxQueryBuilder.tsx
asalem1: Surely there's a better way

asalem1 (commented):
I might be missing context on the inject mechanism that would need clarify w/r/t to the comment about the QueryContext, but I think some of that styling might be a little unsustainable and buggy in the long run if we leave it as is. It might be worth trying to resolve some of these reactivity issues individually and breaking this PR up so that we can tackle one styling problem at a time. 

Also, not sure if the SVGs should live in the UI or in clockface but that's 🤷🏽

src/dataExplorer/components/FluxQueryBuilder.tsx
wiedld: Removed the collapsing panel. Resolved.

wiedld:
> I might be missing context on the inject mechanism that would need clarify w/r/t to the comment about the QueryContext, but I think some of that styling might be a little unsustainable and buggy in the long run if we leave it as is. It might be worth trying to resolve some of these reactivity issues individually and breaking this PR up so that we can tackle one styling problem at a time.
> 
> Also, not sure if the SVGs should live in the UI or in clockface but that's 🤷🏽

Changes in response:
* removed collapsing mechanism, since that's not the main feature anyways but was a big (understandable) part of the negative feedback.
  * No svgs now either.
* Use of QueryContext is a discussion for the team to have another day.
  * Does not have to be part of the convo for this PR. 
  * Was merely highlighting it because the downstream of this PR flow (a.k.a. the query) uses it.

src/dataExplorer/components/SidePane.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+const SidePane: FC = () => {
+  return (
+    <div className="container-right-side-bar">
+      <FluxBrowser />
+    </div>
+  )
+}
+
+export default SidePane
```
asalem1, line 17: Is this conditionally rendered when the user runs long distances?

src/shared/components/FilterList/MinimalistInjectOption.tsx
```diff
@@ -0,0 +1,78 @@
...
+          testID={`flux--${testID}--inject`}
+          text={extractor(option)}
+          size={ComponentSize.ExtraSmall}
+        />
+      </dd>
+    </>
+  )
+}
+
+InjectionOption.defaultProps = defaultProps
```
asalem1, line 76: why like this? Why not just assign the parameter?

asalem1 (approved):

src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -3,36 +3,48 @@ import React, {FC, useState} from 'react'
...
-        <DraggableResizer.Panel>
-          <Schema />
-        </DraggableResizer.Panel>
-        <DraggableResizer.Panel className="new-data-explorer-rightside">
-          <ResultsProvider>
-            <ResultsPane />
-          </ResultsProvider>
-        </DraggableResizer.Panel>
-      </DraggableResizer>
+      <EditorProvider>
```
drdelambre, line 26: having to set this so high just so other components can speak to it's injection mechanism seems to be a bit of code smell. lets say we have a button in the header that triggers the injection mechanism, do we now have to hoist the provider to the page level? does the meaning stay the same as the code floats in a different context, or is the connection between the disparate interactions constitute application logic which should be handled by a separate provider to cover the connection of these three components (fluxeditor, sidepane, schema)

wiedld: If you want the providers to be singletons, then they need to be singletons. Especially if it holds state (e.g. the connection to the editor).

src/dataExplorer/components/SidePane.tsx
```diff
@@ -0,0 +1,17 @@
+import React, {FC} from 'react'
+
+// Components
+import FluxBrowser from 'src/dataExplorer/components/FluxBrowser'
+
+// Utils
+import './SidePane.scss'
+
+const SidePane: FC = () => {
+  return (
```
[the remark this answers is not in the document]
wiedld: Design has some ideas. Leaving open the option.

src/dataExplorer/components/FluxBrowser.tsx
```diff
@@ -0,0 +1,82 @@
...
+// Utils
+import {event} from 'src/cloud/utils/reporting'
+import {getFluxExample} from 'src/shared/utils/fluxExample'
+import {sortFuncs} from 'src/shared/components/functions/utils'
+import './FluxBrowser.scss'
+
+const TOOLTIP = `The flux standard library contains several packages, \
+functions, and variables which may be useful when constructing your flux query.`
+
+const FluxBrowser: FC = () => {
```
drdelambre, line 27: why do we have three versions of this interaction? are they that different?

src/shared/components/FilterList/MinimalistInjectOption.tsx
wiedld: Was copy/pasta. Changed.

src/dataExplorer/components/FluxBrowser.tsx
wiedld: Yes, they are different.
* Each of the three component uses the same child components (e.g. FilterList).
* But then has different grandchildren components (e.g. renderItem) which is provided in this higher level component.
  * The other 2 components have moved closer together over time, but this new FluxBrowser moves further away per the new design.
* This parent components is just the callback and the renderItem.

src/dataExplorer/components/FluxQueryBuilder.tsx
wiedld: But if you have shared state to the editor access, then you really only want 1 connection to the editor. And the same provider where the editor is set via the callback from the react-monaco-editor.

src/dataExplorer/components/FluxBrowser.tsx
wiedld: The handoff was a Figma design.
I can double check (ask to message up the tree).

src/dataExplorer/components/FluxQueryBuilder.tsx
drdelambre: yeah no this isn't about the relationship between the editor context and the flux monaco editor, that one's ok, but about the relationship between the editor context and the other implementation specific components of the side panel and the schema browser. while i doubt there's going to be a product requirement to have a single injection mechanism update multiple editors, the code encapsulation should ensure that the module interdependence mechanism be resilient enough to handle an implementation change, especially since our next step is to experiment with this interaction. one proposed option explored by product is to have every button press in the schema browser call injection into a new flux query in an editor that is in preview mode, _then_ a separate button actually injects it into your working query. the interface as described in this pr is creating an implicit tight coupling between three things that should have little to do with each other and would make that option difficult to implement. there needs to be a controller layer here for encapsulating the application logic and decouple those components. Not expecting you to engineer for imaginary test cases here, just using examples to help illustrate that there are more flexible patterns to use in this case

src/dataExplorer/components/FluxBrowser.tsx
drdelambre: oh yeah. a lot can be lost in just a picture. thanks for asking

src/flows/pipes/RawFluxEditor/view.tsx
wiedld: Existing code was keeping shared components under flows, even though timeMachine also uses it. So it was a move into the shared part...without any further work to de-dupe the components (since outside scope of this PR).
.

Reasoning (which can be changed) in the dir:
* `src/shared/components/functions/`
  * the components used in the function browsers
* subdir `/FunctionsList`
  * our 2 FunctionList styles, specifically for functions.
  * are used conditionally in both timeMachine and flows.
* subdir `/perFunction`
  * reusable tooltip contents, specific for function.

.

We could probably stand to condense these ^^ even further. But at the time, there was much inter-team negotiation to get it even down to this much shared components. 
* As a follow up to this PR, we could try that process of moving them closer again.
* Especially since the additional changes since, have moved these components closer (except not the tooltips).

The question is whether or not we want to holdup this FluxBrowser PR.

src/dataExplorer/components/FluxQueryBuilder.tsx
wiedld: The provider holds state, the connection to the editor. Therefore we need a single connection to the editor, so a single provider used). The context use can be moved closer down the tree to the components, but the single connection/provider needs to be available to both the SidePane (with the FluxBrowser injecting functions) and the SchemaPane (to be injecting measurements), as well as the central pane (with the react-monaco-editor setting the editor instance).

So it's a single editor, a single editor connection, and a single injection point...via the common provider.

We don't want multiple connections to the editor, via multiple providers. (That is a recipe for race conditions etc.) We cannot entirely move our application away from any shared state...which is why we use the provider pattern for our shared state, with a single instance per provider. This matches our existing pattern and use of providers.

wiedld: I understand your point of implicit coupling. I am disagreeing with priorities for reducing bugs.

I am stating that a single src of connection to the editor instance is more important. And adding a pattern of event listeners (graph pattern into the editor) more messy and prone to bugs imho.

drdelambre: i dont understand that jump? im not advocating for a particular control pattern, just that the components need to be decoupled to fill their application and provided a list of other cases where we broke from these implicit couplings through prop drilling in order to keep components reusable and maintainable.

wiedld: I got a greenlight to merge this for now, (to not block release of the FluxBroswer into the feature flagged new QX FluxBuilder). We can then have a conversation next week in followup about how we want to use providers in this part of the application.

PR #4781, opened 2022-06-08
chore: freeze timerange changes from triggering background queries in dashboards

asalem1:
This PR resolves an issue where changing the timerange in a cell editor overlay would trigger a the background cells to update as well. Not entirely sure when this mechanism broke, but somewhere along the way this stopped work. 

The crux of the issue was that we referencing the wrong route, `cell` vs `cells`. Not sure when that change happened but it looks like this has been broken and unreported for a while.

There was a bunch of state setting that I cleaned up to just derive values off of existing information that's available in the TimeSeries component

cell-editor

### DIFF
--- a/src/buckets/components/BucketCard.tsx
+++ b/src/buckets/components/BucketCard.tsx
@@ -1,6 +1,6 @@
 // Libraries
 import React, {FC} from 'react'
-import {withRouter, RouteComponentProps} from 'react-router-dom'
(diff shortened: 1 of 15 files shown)


drdelambre (approved):
seems like mostly FC rewritres, but to reiterate the dashboard performance findings from last week, the slowness we're sensitive to at the moment isn't related to the rerunning of queries but the rerendering of already run queries

PR #4785, opened 2022-06-08
style: update hover effect to current design

appletreeisyellow:
Related #4510 

This PR updates the hover effect of the fields and tag values. 

## After

https://user-images.githubusercontent.com/14298407/172699371-a88da57e-17ae-496a-af7f-db5ae92e2f22.mov

### DIFF
--- a/src/dataExplorer/components/FluxQueryBuilder.tsx
+++ b/src/dataExplorer/components/FluxQueryBuilder.tsx
@@ -10,7 +10,7 @@ import Schema from 'src/dataExplorer/components/Schema'
 // Styles
 import './FluxQueryBuilder.scss'
 
-const INITIAL_VERT_RESIZER_HANDLE = 0.25
+const INITIAL_VERT_RESIZER_HANDLE = 0.24
 
 const FluxQueryBuilder: FC = () => {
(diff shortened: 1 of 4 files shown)


subirjolly (approved):

src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -10,7 +10,7 @@ import Schema from 'src/dataExplorer/components/Schema'
 // Styles
 import './FluxQueryBuilder.scss'
 
-const INITIAL_VERT_RESIZER_HANDLE = 0.2
+const INITIAL_VERT_RESIZER_HANDLE = 0.24
```
appletreeisyellow, line 13: There was a request (from Julia on [slack](https://influxdata.slack.com/archives/C030G5783M3/p1654096476089539?thread_ts=1654094364.248209&cid=C030G5783M3)) to increase the default width for schema browser. Since the clockface component `DraggableResizer` set default width by ratio instead of by a fix value, after calculation, Julia suggested to use 0.24

drdelambre: 🤔 🤔 🤔

drdelambre (approved):

Issue #4677, opened 2022-05-25
Error when navigating dashboards in cloud2


philjb:
## About the bug
While editing a dashboard, I received the "an error has occurred. please report here" message so I am reporting it here. 

The error seemed to be persistent as I navigated around (see the gifs) but I'm not sure how to reproduce it.

In the zip archive are two text files showing the console errors. Here's the top of them.

```
==> error1 <==
[Error] TypeError: Type error
decode
D — flux-lsp-browser_bg.js:342
__wbindgen_json_parse — bootstrap:69
wasm-stub
<?>.wasm-function[4200]
<?>.wasm-function[1895]
wasm-stub
C — flux-lsp-browser_bg.js:225
s — astim.ts:22

==> error2 <==
[Error] Unhandled Promise Rejection: SyntaxError: JSON Parse error: Unrecognized token ''
        parse (30.f5990f947f.js:2:1779771)
        D (30.f5990f947f.js:2:1779771)
        __wbindgen_json_parse (c0375f819e.js:1:544)
        wasm-stub
        <?>.wasm-function[4200]
        <?>.wasm-function[1895]
        wasm-stub
        C (30.f5990f947f.js:2:1779085)
        s (31.9bc96f602e.js:1:642599)
```

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

Unknown

**Expected behavior:**
Use cloud2 without errors.

**Actual behavior:**
Page crashed and reported than an error has occurred.

**Visual Proof:**

![ui error](https://user-images.githubusercontent.com/13472206/170364275-99860467-d028-4de0-988f-5be2a0e6ca37.gif)
<img width="902" alt="Screen Shot 2022-05-25 at 12 02 16 PM" src="https://user-images.githubusercontent.com/13472206/170364291-5962ee24-d5a7-4fca-aff7-50a0b2d02a17.png">
[Archive.zip](https://github.com/influxdata/ui/files/8774434/Archive.zip)

## About your environment

**Environment info:**
cloud2

**Config:**

na

**Logs:**

see archive above for console log
**Performance:**
na

philjb:
I got two type errors when just scrolling down my dashboard

```
==> type error <==
[Error] Unhandled Promise Rejection: TypeError: Type error
        decode (30.19444cc24f.js:2:1779773)
        D (30.19444cc24f.js:2:1779773)
        __wbindgen_json_parse (fc7b1f118e.js:1:544)
        wasm-stub
[...]

PR #4786, opened 2022-06-08
fix(4677): lsp helper methods need to be error handled.

wiedld:
Part of (possibly closes) #4677 

## Bug:
* when occurs:
  * on load of the old DataExplorer, with the queryBuilder (not the flux script editor):
    * get error in the queryBuilder loading
  * on load of an existing Dashboard, once you click to open a cell:
    * this again loads DataExplorer subcomponents, with a queryBuilder (not a flux script editor). 
* console errors:
  * lsp library error.

## Debugging steps:
1. since the script editor is not loaded in these^^ views, the error is likely in an lsp helper method.
2. found downstream code in these^^ use cases:
    * which use the lsp helper methods
    * where are not error handling
3. if I manually force throw at these points --> I see the same UI error screens.
4. These errors only occur in certain use cases:
    * based upon the user query and/or the source data loaded in the query builder
    * may be why this error does not appear everywhere for everyone. 🤷🏼 

## Post-merge/depoy followup:
* Have issue reporters double check that it resolves their use cases, for their data.
* Only part of the ticket not reproduced:
  * since we are not loading the script editor (in the screenshot test cases), we are not using the editor model.
  * This means that this call stack cannot be triggered: https://github.com/influxdata/ui/issues/4677/#issuecomment-1146400724
  * Not sure if we have a separate bug going on? Will ping for more details.

### DIFF
--- a/src/flows/pipes/Notification/view.tsx
+++ b/src/flows/pipes/Notification/view.tsx
@@ -42,6 +42,7 @@ import {
 } from 'src/flows/pipes/Notification/endpoints'
 import {SidebarContext} from 'src/flows/context/sidebar'
 import ExportTask from 'src/flows/pipes/Notification/ExportTask'
+import ErrorBoundary from 'src/shared/components/ErrorBoundary'
 const NotificationMonacoEditor = lazy(() =>
   import('src/flows/pipes/Notification/NotificationMonacoEditor')
 )
@@ -481,7 +482,9 @@ const Notification: FC<PipeProp> = ({Context}) => {
         </FlexBox>
         <Panel.Footer justifyContent={JustifyContent.FlexEnd}>
           <FlexBox margin={ComponentSize.Medium}>
-            <ExportTask />
+            <ErrorBoundary>
+              <ExportTask />
+            </ErrorBoundary>
           </FlexBox>
         </Panel.Footer>
       </div>
--- a/src/timeMachine/actions/queries.ts
+++ b/src/timeMachine/actions/queries.ts
@@ -81,15 +81,19 @@ export const getOrgIDFromBuckets = (
   text: string,
   allBuckets: Bucket[]
 ): string | null => {
-  const ast = parse(text)
-  const bucketsInQuery: string[] = findNodes(ast, isFromBucket).map(node =>
-    get(node, 'arguments.0.properties.0.value.value', '')
-  )
+  try {
+    const ast = parse(text)
+    const bucketsInQuery: string[] = findNodes(ast, isFromBucket).map(node =>
+      get(node, 'arguments.0.properties.0.value.value', '')
+    )
 
-  // if there are buckets from multiple orgs in a query, query will error, and user will receive error from query
-  const bucketMatch = allBuckets.find(a => bucketsInQuery.includes(a.name))
+    // if there are buckets from multiple orgs in a query, query will error, and user will receive error from query
+    const bucketMatch = allBuckets.find(a => bucketsInQuery.includes(a.name))
 
-  return get(bucketMatch, 'orgID', null)
+    return get(bucketMatch, 'orgID', null)
+  } catch (_) {
+    return null
+  }
 }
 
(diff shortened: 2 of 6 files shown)


src/variables/utils/astim.ts
```diff
@@ -19,11 +19,19 @@ const parseAllVariables = (ast: File): MemberExpression[] => {
 }
 
 export const parseASTIM = (query: string): ASTIM => {
-  const ast: File = parse(query)
-  // Using `any` to circumvent ts error
-  const variables: any = ast ? parseAllVariables(ast) : []
+  let ast: File = null
+  try {
```
wiedld, line 23: This method `parseASTIM()` is called in the existing DataExplorer, as well as when opening a cell on the Dashboard.

src/flows/pipes/Notification/view.tsx
```diff
@@ -481,7 +482,9 @@ const Notification: FC<PipeProp> = ({Context}) => {
         </FlexBox>
         <Panel.Footer justifyContent={JustifyContent.FlexEnd}>
           <FlexBox margin={ComponentSize.Medium}>
-            <ExportTask />
+            <ErrorBoundary>
+              <ExportTask />
+            </ErrorBoundary>
```
wiedld, line 487: ExportTask has many uses of the lsp helper methods, which are not all error handled. Therefore, adding an ErrorBoundary wrapper.

src/writeData/components/ClientCodeQueryHelper.tsx
```diff
@@ -50,18 +50,19 @@ const ClientCodeQueryHelper: FC<Props> = ({clientQuery, contentID}) => {
...
-    if (queryBucket) {
-      changeBucket({name: queryBucket} as Bucket)
-    }
-    updateBucketInAST(ast, '<%= bucket %>')
-    let query = format_from_js_file(ast)
-    if (def.querySanitize) {
-      query = def.querySanitize(query)
-    }
-    changeQuery(query)
+    try {
```
wiedld, line 53: Not in the direct path of the errors seen in this ticket, but is also instances when lsp methods are not error handled.

There are remaining instances in notebook/flows code, which are not error handled. But since the error was not seen in notebooks (and there are several more instances there), I timeboxed to code outside of noteboooks/flows.

src/timeMachine/actions/queries.ts
```diff
@@ -81,15 +81,19 @@ export const getOrgIDFromBuckets = (
   text: string,
   allBuckets: Bucket[]
 ): string | null => {
-  const ast = parse(text)
-  const bucketsInQuery: string[] = findNodes(ast, isFromBucket).map(node =>
-    get(node, 'arguments.0.properties.0.value.value', '')
-  )
+  try {
```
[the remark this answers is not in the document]
drdelambre: you can actually remove this function all together, we no longer allow for the bucket name to be an override for determining the org a query is run against (long, aweful story). it must have been skipped when we did the cleanup for removing the old sample bucket feature

src/variables/utils/astim.ts
```diff
@@ -19,11 +19,19 @@ const parseAllVariables = (ast: File): MemberExpression[] => {
...
+    ast = parse(query)
+  } catch (_) {}
```
drdelambre, line 25: can we console.error these instead of swallowing them? it helps support out later down the line

src/variables/utils/astim.ts
```diff
@@ -19,11 +19,19 @@ const parseAllVariables = (ast: File): MemberExpression[] => {
 }
 
 export const parseASTIM = (query: string): ASTIM => {
-  const ast: File = parse(query)
-  // Using `any` to circumvent ts error
```
wiedld, line 23: Removed use of `any`  type. Which indicated that the code was not handling 2 possible children of MemberExpression.

We now handle both cases, both as if else. (Some future proofing in case there is a possible AST node child of `variable.property` which we don't know about in the UI types.

asalem1 (approved):

PR #4787, opened 2022-06-08
chore: perform case-insensitive sort in schema browser

appletreeisyellow:
This PR adds case-insensitive sort in schema browser query to get the values of measurements, fields, tag keys, and tag values. 

This is a suggestion from @jsternberg (on [slack](https://influxdata.slack.com/archives/C030G5783M3/p1653681553393739))

### DIFF
--- a/src/dataExplorer/context/fields.tsx
+++ b/src/dataExplorer/context/fields.tsx
@@ -18,7 +18,7 @@ import {QueryContext} from 'src/shared/contexts/query'
 // Utils
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {
-  IMPORT_REGEXP,
+  IMPORT_STRINGS,
   IMPORT_INFLUX_SCHEMA,
   SAMPLE_DATA_SET,
   FROM_BUCKET,
@@ -80,7 +80,7 @@ export const FieldsProvider: FC<Prop> = ({children, scope}) => {
     //   Here is the source code for handling sample data:
     //   https://github.com/influxdata/flux/blob/master/stdlib/influxdata/influxdb/sample/sample.flux
     //   That is why _source and query script for sample data is different
-    let _source = IMPORT_REGEXP
+    let _source = ''
     if (bucket.type === 'sample') {
       _source += SAMPLE_DATA_SET(bucket.id)
     } else {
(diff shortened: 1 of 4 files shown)


src/dataExplorer/context/fields.tsx
```diff
@@ -80,7 +80,7 @@ export const FieldsProvider: FC<Prop> = ({children, scope}) => {
     //   Here is the source code for handling sample data:
     //   https://github.com/influxdata/flux/blob/master/stdlib/influxdata/influxdb/sample/sample.flux
     //   That is why _source and query script for sample data is different
-    let _source = IMPORT_REGEXP
+    let _source = ''
```
appletreeisyellow, line 83: `regexp` is not used anywhere in the query here, so removing

src/dataExplorer/context/tags.tsx
```diff
@@ -110,19 +110,21 @@ export const TagsProvider: FC<Prop> = ({children, scope}) => {
...
           stop: ${CACHING_REQUIRED_END_DATE},
         )
-          |> filter(fn: (r) => r._value != "_measurement" and r._value != "_field")
-          |> filter(fn: (r) => r._value != "_start" and r._value != "_stop")
-          ${searchTerm ? SEARCH_STRING(searchTerm) : ''}
-          |> sort()
-          |> limit(n: ${limit})
+        |> filter(fn: (r) => r._value != "_measurement" and r._value != "_field")
+        |> filter(fn: (r) => r._value != "_start" and r._value != "_stop")
+        ${searchTerm ? SEARCH_STRING(searchTerm) : ''}
```
drdelambre, line 123: dont we still need the regex imported for `SEARCH_STRING`?

appletreeisyellow: yes, since we are using strings package `strings.toLower()` in the next line

src/dataExplorer/context/tags.tsx
```diff
@@ -110,19 +110,21 @@ export const TagsProvider: FC<Prop> = ({children, scope}) => {
     `
 
     if (bucket.type !== 'sample' && isFlagEnabled('newQueryBuilder')) {
-      _source = `${IMPORT_REGEXP}${IMPORT_INFLUX_SCHEMA}`
+      _source = `${IMPORT_STRINGS}${IMPORT_INFLUX_SCHEMA}`
```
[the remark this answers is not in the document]
appletreeisyellow: regexp is also not used here, but the `strings` package is used in this query, so i replaced `IMPORT_REGEXP` with `IMPORT_STRINGS`

drdelambre (approved):
i dont think we need to drop the lowercase column as the results aren't exposed to the user, and there's a potential perf hit on that, but not a blocker. the rest looks great!

appletreeisyellow:
that's a good point!

PR #4795, opened 2022-06-09
chore(helpBar): usability tweaks after feedback

hoorayimhelping:
Closes https://github.com/influxdata/EAR/issues/3323
Connects https://github.com/influxdata/EAR/issues/3322

- Appends the current org id to the description
- Translates the severity we show on the dropdown into the value SF uses (e.g. `3 - Standard` becomes `Severity 3`)
- Updates the text in the confirmation overlay to say for critical issues, calling support is required
- Makes Severity 3 the default:

Screen Shot 2022-06-09 at 4 21 24 PM


blegesse-git (approved):
looks great!

PR #4796, opened 2022-06-09
fix: redirect to CLOUD but not for dev environments

TCL735:
Part of #4657 

- Refreshing doesn't work because it handles testing of edge cases poorly
- Redirecting seems the best approach, but not for dev environments
- Ignore this issue for dev environments

### DIFF
--- a/src/Signin.tsx
+++ b/src/Signin.tsx
@@ -26,6 +26,7 @@ import {
   CLOUD,
   CLOUD_LOGIN_PATHNAME,
   CLOUD_SIGNIN_PATHNAME,
+  CLOUD_URL,
 } from 'src/shared/constants'
 
 // Types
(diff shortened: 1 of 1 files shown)


wdoconnell:
LGTM. `CLOUD_URL` always points to the quartz login page (as opposed to the deprecated cluster login page)?

wdoconnell (approved):

src/Signin.tsx
```diff
@@ -114,8 +115,12 @@ export class Signin extends PureComponent<Props, State> {
       clearInterval(this.intervalID)
 
       if (CLOUD) {
-        if (isFlagEnabled('useQuartzLogin')) {
-          window.location.reload()
+        if (
+          isFlagEnabled('useQuartzLogin') &&
+          process.env.NODE_ENV &&
+          process.env.NODE_ENV !== 'development'
```
hoorayimhelping, line 121: can this be shortened to

```ts
process.env?.NODE_ENV !== 'development'
``` 
?

hoorayimhelping (approved):

TCL735:
> LGTM. `CLOUD_URL` always points to the quartz login page (as opposed to the deprecated cluster login page)?

yes

src/Signin.tsx
TCL735: We also need to handle the case when the `NODE_ENV` is not set. This case is also considered a dev environment.

PR #4797, opened 2022-06-10
fix: updated rename org setting and Host Name

### LINKED ISSUE #4052: Org Settings Page: Host Name = Cluster URL
**User Story: Org Settings**
As a user, I want to see the "Host Name" in my Org settings so that I know that it's equal to the Cluster URL. 

**Technically, this is NOT part of multi-org, but, it helps users learn that host name = Cluster URL.** 

    ACs
    - Add "Host Name" in parentheses next to "Cluster URL" so that users understand these are the same.  
    
    Dev Notes
    -

### LINKED ISSUE #4791: Copy to clipboard to org name field
There should be a Copy to Clipboard button for Org Name. As you can see, the org name is getting cut off as well.

image

david-rusnak:
Closes #4791 
Closes #4052 

Closes both of the issues above. Made an optional prop on `CopyableLabeledData` to toggle the rename org button being visible. Left the Copy button to say "Copy to Clipboard" for upcoming facelift. Also adds "(Host Name)" next to Cluster URL to make it clear to users.

https://user-images.githubusercontent.com/106361125/173078264-5e72d5c8-2696-4f8f-a037-df720b765254.mov

### DIFF
--- a/src/organizations/components/OrgProfileTab/CopyableLabeledData.tsx
+++ b/src/organizations/components/OrgProfileTab/CopyableLabeledData.tsx
@@ -29,7 +29,7 @@ interface Props {
   label: string
   src: string
   name?: string
-  renameableOrg?: boolean
+  isRenameableOrg?: boolean
 }
 
 const CopyableLabeledData: FC<Props> = ({
@@ -37,7 +37,7 @@ const CopyableLabeledData: FC<Props> = ({
   label,
   src,
   name,
-  renameableOrg,
+  isRenameableOrg,
 }) => {
   const dispatch = useDispatch()
   const org = useSelector(getOrg)
@@ -80,7 +80,7 @@ const CopyableLabeledData: FC<Props> = ({
             testID={`copy-btn--${id}`}
             onCopy={generateCopyText}
           />
-          {renameableOrg && (
+          {isRenameableOrg && (
             <Button
               testID="rename-org--button"
               text="Rename"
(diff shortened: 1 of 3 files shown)


src/organizations/components/OrgProfileTab/CopyableLabeledData.tsx
```diff
@@ -19,20 +22,35 @@ import {notify} from 'src/shared/actions/notifications'
...
+  renameableOrg?: boolean
 }
 
-const CopyableLabeledData: FC<Props> = ({id, label, src, name}) => {
+const CopyableLabeledData: FC<Props> = ({
+  id,
+  label,
+  src,
+  name,
+  renameableOrg,
```
hoorayimhelping, line 40: one suggestion on boolean variable naming:

When naming booleans, it generally helps remove ambiguity to have some kind of conjugation of the verb `to be`. `renameableOrg` could be thought of as a value, as in 
```ts
const renameableOrg = 'org1234,org567'
```
It might be read as the org that is renamable, rather than answering the question, is this org renameable.

Contrast that with `orgIsRenameable` or `isOrgRenameable` or even `isRenameableOrg`. The `is`, which is a conjugation of the verb `to be`, which indicates existence, hints that it's a boolean. It's kind of one of the only places where modern coding conventions use [Hungarian Notation.](https://en.wikipedia.org/wiki/Hungarian_notation)

So down below it would change to something like

```ts
{isOrgRenameable && (
  <Button
```

which is a small thing, but helps clue in the reader to the intent of the code.

hoorayimhelping (commented):
one comment on variable naming.

Regardless of that, solid, clean PR! Nice job

PR #4799, opened 2022-06-10
feat: persist browser

drdelambre:
a majority of this PR is centered around normalizing some of the patterns that are emerging from the query experience project into other places in the code base that are starting to diverge between the old data explorer, the new data explorer, and notebooks. things like having them all with a separate idea of how a function browser should look, and migrating them to all using the same injection mechanisms instead of re-implementing them all over the codebase (i'm looking at you secrets). This boundary around injection was also formalized to separate intent and implementation in the code, as those should be following different cadences through our next few sprints and our current patterns around it were leading to those efforts being locked at best, and neglected in practice.

BUT

at the last second and at the request of product, I added persistence to some of the user selections you can make around the new query builder (query, table vs graph, view properties, bucket and measurement selection), so that's the feat this PR is built around.

Peek 2022-06-10 13-38


src/dataExplorer/components/Results.tsx
```diff
@@ -100,13 +113,23 @@ const Results: FC = () => {
   }
 
   const dataExists = result.parsed && Object.entries(result.parsed).length
+  const setVizState = state => {
```
asalem1, line 116: wait, what is `state`?

src/shared/components/FilterList/InjectionOption.tsx
```diff
@@ -48,6 +50,21 @@ const FluxInjectionOption: FC<Props> = ({
...
+  }
+
+  const classer = [
+    ['flux-toolbar--list-item', true],
+    ['flux-toolbar--function', true],
+    ['flux-toolbar--new-style', isFlagEnabled('fluxEditorButtonStyling')],
+  ]
+    .filter(c => c[1])
+    .map(c => c[0])
+    .join(' ')
```
asalem1, line 67: but like ... why have to reason about it and not just:

`flux-toolbar--list-item flux-toolbar--function ${isFlagEnabled ? 'flux-toolbar--new-style}`

src/shared/contexts/editor/index.tsx
```diff
@@ -106,74 +114,119 @@ export const EditorProvider: FC = ({children}) => {
...
+    nextEditor.getModel().onDidChangeContent(() => {
+      if (!editor) {
+        return
+      }
+      const val = editor.getValue()
+      updateCBs.current.forEach(cb => cb(val))
+    })
+
+    setEditor(nextEditor)
+  }
```
asalem1, line 189: big fan of making the editor responsible for editor-related issues

src/shared/components/FilterList/InjectionOption.tsx
drdelambre: Oh this is a copy pasta job i started inserting a year ago instead of using classer. I like the readability of the format (makes more sense as complexity grows), but it felt weird maintaining a dependency for it

src/shared/contexts/editor/index.tsx
drdelambre: Me too. Optional provider is way more flexible than prop drilling. Of note, the code above this isn't actually used anywhere, but denise was using a callback structure instead of basing it off the editor state updating for some reason. That stuff seemed to just be double calling the updateText functions instead of picking up on some dropped state change like i originally thought, so it melted away as i tightened up the interfaces, but I wanted to keep the option there in case I missed a special condition

src/dataExplorer/components/Results.tsx
drdelambre: 'table' | 'graph', enforced on the view assignment

asalem1: maybe we name it something else since state is pretty loaded?

aanthony1243 (changes requested):
We should not merge this PR. 

First, this is way too big of a change, we need to figure out how to accomplish this incrementally.  

Second, anything involving the query editing needs to be reviewed by Paul to see if we can use the LSP instead so that we can take advantage of the formal Flux grammar features.  

Finally, if this must be broken up, any PR that requires changing 48 files and affects the whole UI will need the approval of a senior engineer from outside our team.

drdelambre:
this code has been merged in already

PR #4802, opened 2022-06-10
fix: use the explicit quartz login url

TCL735:
Part of #4657 

Production also sets `CLOUD_URL` to something other than what we want. We will have to be explicit.


wdoconnell (approved):
Looks good to me. `CLOUD_QUARTZ_URL` is within a block limited to cloud environments with the flag turned on.

PR #4809, opened 2022-06-13
chore: update giraffe

asalem1:
Closes #


ChitlangeSahas (approved):
This sounds good! 

Before merging maybe giving a heads up in the UI channel would be helpful for others to have context?

✅

Issue #4798, opened 2022-06-10
Multi Org - expired session definition


TCL735:
Currently, when an api call to`/api/v2/me` (aka `me`) is **not** a 200, the response is considered an error and treated uniformly. The user is redirected to a login page. This includes both expired sessions and temporary hiccups from the network.

Generally, the way the user is handled should be more nuanced.
 
- [ ] An expired session should continue to ask the user to re-authenticate (log in). This is indicated with a 4xx response from the `me` endpoint.

- [ ] Any 5xx response should not need to re-authenticate the user. Instead 5xx response can simply try to call the `me` endpoint again.

Issue #4587, opened 2022-05-16
Multi Org - Update polling /me endpoint (to be deprecated) to /identity endpoint


kristinarobinson:
Currently there exists in the UI a polling request every 60 secs that returns valid session/user account data from the /idpe/me endpoint.  That endpoint is being deprecated, and should be replaced with the new /quartz/identity endpoint.

wdoconnell:
Testing to make sure access permissions still work after making board private.

Luckette:
#6207

Luckette:
Heads up, @eatondustin1

wdoconnell:
@eatondustin1 FYI - note that this is still partially feature flagged. In the cloud, IDPE `/me` isn't used for polling anymore - as long as the user has access to quartz, quartz endpoints are used instead. 

But until the `quartzIdentity` flag is on for all users in the UI, `/quartz/me` will still be used in prod.

eatondustin1:
Awesome thanks for the heads up Bill, just let us know when the FF is removed and we'll remove our `/me` endpoint!

abshierjoel:
@wdoconnell 
Hi Bill, I just wanted to check if this FF was now fully enabled? I checked configCat and it looks like `quartzIdentity` is on for all users now.

wdoconnell:
@abshierjoel Correct, the FF has been on now in prod without issue for 10 days. Let me check with the team re: how much longer before we rip out the dead code behind the flag on the UI side, which will unblock you all on removing the dead `/quartz/me` code on your end.

PR #4816, opened 2022-06-14
feat(identity): create an identity api layer ahead of multi-org

hoorayimhelping:
Connects #4798
Connects #4587

Not currently in use. To be used by the code that addresses the above tickets. This PR creates an interface that multiple changes can make use of.

This creates an API layer for identity operatrions. This includes the code to hit the routes

- `/me`
- `/quartz/me`
- `/quartz/identity`

among others.

This layer provides an abstraction layer and encapsulation for making the decision on whether to fetch `me` from idpe, `me` from quartz or `identity` from quartz.

### DIFF
--- a/src/identity/apis/auth.ts
+++ b/src/identity/apis/auth.ts
@@ -1,0 +1,213 @@
+// Functions calling API
+import {
+  getAccount,
+  getIdentity,
+  getMe as getMeQuartz,
+  getOrg,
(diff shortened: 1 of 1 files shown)


wdoconnell (approved):
Looks good to me. Since we've paired on this already, my input is already incorporated.  thumbsup:

PR #4820, opened 2022-06-14
chore: add events to the bucket & measurement, and results window

asalem1:
This PR introduces eventing around bucket selection, measurement selection, and result status for queries that have been run using the FluxQueryBuilder. This does not include adding events of the same nature to the TagSelector and FieldSelector since the selection functionality is not currently implemented. We should make sure to add those events as and when the functionality is added

### DIFF
--- a/src/buckets/components/RenameBucketOverlay.tsx
+++ b/src/buckets/components/RenameBucketOverlay.tsx
@@ -1,52 +1,43 @@
 // Libraries
-import React, {PureComponent} from 'react'
-import {withRouter, RouteComponentProps} from 'react-router-dom'
+import React, {FC} from 'react'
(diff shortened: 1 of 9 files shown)


src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -99,10 +99,14 @@ const ResultsPane: FC = () => {
...
     })
       .then(r => {
+        event('resultReceived', {
+          status: r.parsed.table.length === 0 ? 'empty' : 'good',
+        })
         setResult(r)
         setStatus(RemoteDataState.Done)
       })
       .catch(() => {
+        event('resultReceived', {status: 'error'})
```
drdelambre, line 109: note to self: figure out what a timeout error looks like

Issue #4821, opened 2022-06-14
Multi Org Phase 1 - Remove any /quartz/me references after feature flag is removed


wdoconnell:
Transition from `/quartz/me` to `/quartz/identity` employs a featureFlag (`quartzIdentity`) to preserve access to the legacy route and avoid affecting functionality before multi-org is complete.

After the feature flag is removed, and all traffic is using `/quartz/identity`, the remaining legacy references to `/quartz/me` should be removed (assuming they aren't needed for any other environments).

ACs

- Confirm /quartz/me references can be removed after feature flag is off
- Remove references to /quartz/me routes
- Ensure dead code flagged with comments in `identity` thunks and reducer is removed

PR #4822, opened 2022-06-14
feat: transition from /quartz/me to /quartz/identity endpoint

### LINKED ISSUE #4587: Multi Org - Update polling /me endpoint (to be deprecated) to /identity endpoint

### LINKED ISSUE #4588: Multi Org - Update API endpoints from /me to /quartz/identity for org and account info in Global Header
Currently, account and org information can be retrieved from /idpe/me and /quartz/me api's, for slightly different data.  Replace all /me calls with the new /quartz/identity API, which is the new source of truth for all org and account lists.

wdoconnell:
Closes #4588 

**Edits after 6/15 code review:**
- Updated fix to be more explicit about `null` billingProvider values in BillingPageContents.tsx

Implementing multi-org in the UI requires transitioning from the `/quartz/me` to `/quartz/identity` endpoint to retrieve the user's `user`, `account`, and `org` data. This PR implements that fix, which is controlled by feature flag `quartzIdentity`. `quartzIdentity` is also dependent on feature flag `uiUnificationFlag`.

Primary changes:

1. Uses the new API layer for `/quartz/identity` at /src/identity/apis/auth.ts. See #4816 
2. Adds reducer and thunks for pulling state from `identity` endpoint, and `/orgs/:orgId` and `/accounts/:accountId` endpoints as needed, into state.
3. When the feature flag is enabled, `/quartz/identity` is used to populate a new state, found at `state.identity.` That state is also converted to the format used by `state.me` throughout the app.
4. When the feature flag is enabled, the app will pull additional information as needed from `/orgs/:orgId` and `/accounts/:accountId` endpoint, to the extent data that is _currently_ relied upon from `/quartz/me` is not present is `/quartz/identity`. 

Ecommerce is working on a fix to allow us to avoid having to call `/orgs/:orgId` at login.

Screen Shot 2022-06-14 at 5 10 29 PM

When the feature flag is removed, remaining references to the legacy `/quartz/me` endpoint, state, reducer, and thunks should be removed. See [#4821](https://github.com/influxdata/ui/issues/4821).

Influxers can access the Miro board describing the transition [here](https://miro.com/app/board/uXjVOuX1O-k=/).


wdoconnell:
P.S.: Review from one of the @influxdata/dumplings on my `cy.featureFlags` would be much appreciated.

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,121 @@
...
+
+// Utilities
+import {
+  convertIdentityToMe,
+  fetchQuartzIdentity,
+  fetchAccountDetails,
+  fetchOrgDetails,
+} from 'src/identity/apis/auth'
+
+// Retrieves user's quartz identity from /quartz/identity, and stores it in state.identity.
```
hoorayimhelping, line 33: I'm kind of ambivalent about this comment. It basically repeats what the code is doing. I think we should use comments to explain why the code is doing what it's doing, not what the code is doing, since the code explains what it's doing.

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,121 @@
...
+  try {
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Loading))
+
+    const quartzIdentity = await fetchQuartzIdentity()
+
+    dispatch(setQuartzIdentity(quartzIdentity))
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Done))
+
+    const legacyMe = convertIdentityToMe(quartzIdentity)
+    // Remove lines below once quartzMe is deprecated.
```
hoorayimhelping, line 46: I'm not a big fan of comments like this because they tend to overstate their authority. From what we know right now, we'll want to remove those lines when quartzMe is deprecated. But maybe something changes between now and then and this comment becomes outdated and is incorrect. If that happens, there is a vanishingly small chance we'll remember to update this comment, and then when we go to change these lines, we see this comment and wonder if the comment knows something we don't.

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,121 @@
...
+) => {
+  try {
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Loading))
+
+    const initialState = getState()
+    const accountId = initialState?.identity?.account?.id
+
+    if (typeof accountId !== 'string' && typeof accountId !== 'number') {
+      throw new Error(RetrievalError.accountIdentity)
+    }
```
hoorayimhelping, line 70: in what situation would `accountId` not be a number or string? What case is this guarding against that the `try/catch` block won't guard against?

src/identity/apis/auth.ts
```diff
@@ -0,0 +1,213 @@
...
+    paygCreditStartDate: account.paygCreditStartDate,
+    billingProvider: account.billingProvider ? account.billingProvider : null,
+
+    // Organization Data
+    clusterHost: org.clusterHost,
+    regionCode: org.regionCode ? org.regionCode : null,
+    isRegionBeta: org.isRegionBeta ? org.isRegionBeta : null,
+    regionName: org.regionName ? org.regionName : null,
+  }
+}
```
hoorayimhelping, line 213: make sure you get the latest master - this file shouldn't be in this diff, since it's already a part of master

src/identity/reducers/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    const newState = reducer(undefined, setQuartzIdentity(initialState))
+    expect(newState).toEqual(initialState)
+  })
+
+  it('can change the user identity using quartzIdentity', () => {
+    for (let i = 0; i < mockIdentities.length; i++) {
+      const expectedState = mockIdentities[i]
+      const actual = reducer(undefined, setQuartzIdentity(mockIdentities[i]))
+      expect(actual).toEqual(expectedState)
+    }
```
hoorayimhelping, line 33: these days, I'm in favor of delegating the looping logic to `forEach` for simple iteration cases like this, or `map` when you need to construct an array. 

```ts
mockIdentities.forEach(mockIdentity => {
  expect(reducer(undefined, setQuartzIdentity(mockIdentity[i])))
  .toEqual(mockIdentity)
})

File: src/identity/actions/creators/index.ts [hunk 6]
```diff
@@ -0,0 +1,46 @@
...
+  | ReturnType<typeof clearQuartzIdentity>
+  | ReturnType<typeof setQuartzIdentityStatus>
+  | ReturnType<typeof setCurrentBillingProvider>
+  | ReturnType<typeof setCurrentOrgDetails>
+
+export const setQuartzIdentity = (identity: CurrentIdentity) =>
+  ({
+    type: SET_QUARTZ_IDENTITY,
+    identity: identity,
+  } as const)
```
hoorayimhelping, line 21: [YAGNI](https://martinfowler.com/bliki/Yagni.html) - I know we talked about this, but don't add this until we need to use it. You won't forget the concept by the time we need to clear out the quartz identity

src/identity/selectors/index.ts
```diff
@@ -0,0 +1,10 @@
+// Types
+import {AppState} from 'src/types'
+import {CurrentIdentity} from 'src/identity/apis/auth'
+
+export const selectQuartzIdentity = (state: AppState): AppState['identity'] => {
+  return state.identity
+}
+
+export const getQuartzIdentity = (state: AppState): CurrentIdentity =>
```
hoorayimhelping, line 9: should this be `selectQuartzIdentity`?

src/shared/containers/GetOrganizations.tsx
```diff
@@ -45,28 +46,51 @@ const canAccessCheckout = (me: Me): boolean => {
...
     }
-  }, [dispatch, status])
+  }, [dispatch, status]) // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
     if (
       isFlagEnabled('uiUnificationFlag') &&
       quartzMeStatus === RemoteDataState.NotStarted
+      // For now, just check whether quartzMeStatus is not set, because quartzMe is what is
+      // currently being used by the application.
```
hoorayimhelping, line 87: This is another one of those comments that has a tendency to get out of date and not be removed which will cause confusion down the road

src/billing/components/BillingPageContents.tsx
```diff
@@ -9,14 +12,34 @@ import MarketplaceBilling from 'src/billing/components/marketplace/MarketplaceBi
...
+    if (
+      CLOUD &&
+      isFlagEnabled('uiUnificationFlag') && // Need this check to avoid having quartz endpoints hit in tools.
+      isFlagEnabled('quartzIdentity') &&
+      !quartzMe.billingProvider
+    ) {
+      // billingProviderThunk populates billingProvider into 'identity' and (for now) 'me' state.
+      dispatch(getBillingProviderThunk())
+    }
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
wdoconnell: Looks like it - the linter isn't happy with us when the comment is removed.

src/identity/actions/creators/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const clearQuartzIdentity = () =>
+  ({
+    type: CLEAR_QUARTZ_IDENTITY,
+  } as const)
+
+export const setCurrentBillingProvider = (
+  billingProvider: 'zuora' | 'aws' | 'gcm' | 'azure'
```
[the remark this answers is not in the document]
wdoconnell: Good point! Added an export to `src/types` that generates this with an indexed access type from `Account[billingProvider]` in unity routes.

src/checkout/context/checkout.tsx
```diff
@@ -160,10 +162,10 @@ export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
   useEffect(() => {
     getBillingSettings()
     return () => {
-      // Call the `quartz/me` endpoint to update the existing user metadata
-      dispatch(getQuartzMeThunk())
+      // Call the 'quartz/me' or 'quartz/identity' endpoint as applicable (checks for presence of feature flag)
```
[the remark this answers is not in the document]
wdoconnell: Agree, done.

src/billing/components/BillingPageContents.tsx
```diff
@@ -9,14 +12,34 @@ import MarketplaceBilling from 'src/billing/components/marketplace/MarketplaceBi
...
+
+// Thunks
+import {getBillingProviderThunk} from 'src/identity/actions/thunks'
 
 const BillingPageContents: FC = () => {
+  const dispatch = useDispatch()
   const quartzMe = useSelector(getQuartzMe)
 
+  useEffect(() => {
+    // After isFlagEnabled is removed, keep other condition. billingProvider isn't delivered by /quartz/identity.
```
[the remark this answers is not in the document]
wdoconnell: Agree, done.

src/identity/apis/auth.ts
```diff
@@ -0,0 +1,213 @@
...
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+
+  const orgDetails = response.data
+  return orgDetails
+}
+
+// Transform a response from `/identity` into a response that Me machinery can use
+export const convertIdentityToMe = (
```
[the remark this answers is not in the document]
wdoconnell: Done.

src/identity/actions/thunks/index.ts
wdoconnell: Helpful, thanks. I'm removing comments in the PR to the extent they're just describing something that can pretty easily be figured out by just reading the function.

src/identity/actions/thunks/index.ts
wdoconnell: Makes sense, I'll remove. Happy to avoid being on the wrong end of a `git blame` in the event something changes in the UI before we deprecate `/quartz/me`.

src/identity/actions/thunks/index.ts
wdoconnell: For both the `billing` (i.e. `account`) and `org` thunks, you're right we shouldn't need this. For `account` the `accountId` is guaranteed to be a `number`  and for `org` the type is guaranteed to be a string. (And it's set to those types by default in the reducer too.)

src/identity/apis/auth.ts
wdoconnell: Gotcha - I think that's from me pulling the file into my feature branch before your PR made it to `master`.

src/identity/reducers/index.test.ts
wdoconnell: Agree, implemented. Not sure why I wasn't doing that in the test file, since there's no need to exit the loop early.

src/identity/actions/creators/index.ts
wdoconnell: We need `SET_QUARTZ_IDENTITY`, which is what gets used by the `quartzIdentity` thunk, but I'll remove `CLEAR_QUARTZ_IDENTITY`, which I added later, and which you're right we're not using at all yet.

src/identity/selectors/index.ts
wdoconnell: This selector is an unused dupe, so can be safely deleted.

src/shared/containers/GetOrganizations.tsx
wdoconnell: Removed this and some other comments that could end up being out of date.

src/identity/utils/getIdentityThunk.ts
```diff
@@ -0,0 +1,10 @@
+import {CLOUD} from 'src/shared/constants'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {getQuartzMeThunk} from 'src/me/actions/thunks'
+import {getQuartzIdentityThunk} from '../actions/thunks'
+
+// For now, check for both uiUnificationFlag and quartzIdentity to ensure this isn't used in tools.
+export const getIdentityThunk = () =>
+  CLOUD && isFlagEnabled('uiUnificationFlag') && isFlagEnabled('quartzIdentity')
+    ? getQuartzIdentityThunk()
+    : getQuartzMeThunk()
```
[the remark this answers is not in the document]
wdoconnell: 100% right that `CLOUD` is a precondition for both, so I don't think the utility was working as I expected. As far as moving it into the thunk itself, that makes sense to me, here's what I've changed:

Everything that previously used the `getIdentityThunk` utility to decide between `getQuartzMe` and `getQuartzIdentity` thunks now just does one thing - calls `getQuartzIdentity`, which has this logic:

Return / do nothing if we're not in an environment with quartz. Still confirming this behaves as expected since we'd be returning out of the thunk without dispatching any actions.
`if (!CLOUD || !isFlagEnabled('uiUnificationFlag')) {
    return
 }`

If the `quartzIdentity` feature flag is off, just use the legacy thunk.
`
  if (!isFlagEnabled('quartzIdentity')) {
    return getQuartzMeThunk()
 }`

The rest of the logic is the same `getQuartzIdentityThunk`. I like the improvement - requires less cleanup after `quartz/me` is deprecated.

Also implemented the `shouldUseQuartzIdentity` idea, which is great and saves us some lines throughout.

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Done))
+
+    const updatedState = getState()
+    const legacyMe = convertIdentityToMe(updatedState.identity)
+    dispatch(setQuartzMe(legacyMe, RemoteDataState.Done))
+    dispatch(setQuartzMeStatus(RemoteDataState.Done))
+  } catch (error) {
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Error))
+    dispatch(setQuartzMeStatus(RemoteDataState.Error))
+    console.error(error)
```
hoorayimhelping, line 79: Same question here regarding `console.error` vs re-throwing the error

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,107 @@
...
+  dispatch: Dispatch<IdentityActions | MeActions>,
+  getState: GetState
+) => {
+  try {
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Loading))
+
+    const state = getState()
+    const orgId = state?.identity?.org?.id
+
+    const orgDetails = await fetchOrgDetails(orgId)
```
hoorayimhelping, line 93: Same questions here about defensiveness and what happens if `orgId` comes back `undefined`.

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    dispatch(setCurrentOrgDetails(orgDetails))
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Done))
+
+    const updatedState = getState()
+    const legacyMe = convertIdentityToMe(updatedState.identity)
```
hoorayimhelping, line 99: Same here as above regarding converting to identity if this is already done in the `getQuartzIdentityThunk`

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    dispatch(setQuartzMe(legacyMe, RemoteDataState.Done))
+    dispatch(setQuartzMeStatus(RemoteDataState.Done))
+  } catch (error) {
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Error))
+    dispatch(setQuartzMeStatus(RemoteDataState.Error))
+    console.error(error)
```
hoorayimhelping, line 105: Same question here about error handling

src/identity/reducers/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    operatorRole: null,
+    orgCount: 0,
+  },
+  org: {
+    clusterHost: '',
+    id: '',
+    name: '',
+  },
+  account: {
+    id: 0,
```
hoorayimhelping, line 32: Just making sure, the other ids are strings, this is a number. Is that correct?

src/identity/utils/convertIdentityToMe.ts
```diff
@@ -0,0 +1,30 @@
...
+export const convertIdentityToMe = (
+  quartzIdentity: CurrentIdentity
+): MeQuartz => {
+  const {account, org, user} = quartzIdentity
+
+  return {
+    // User Data
+    email: user.email,
+    id: user.id,
+    // isOperator is inferred from presence of a string value to operatorRole (read-write or read-only).
```
hoorayimhelping, line 14: :+1: helpful comment!

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -24,13 +24,30 @@ import {CLOUD} from 'src/shared/constants'
...
+  const dispatch = useDispatch()
 
   const expectQuartzData = CLOUD && isFlagEnabled('uiUnificationFlag')
 
+  useEffect(() => {
+    if (CLOUD && shouldUseQuartzIdentity()) {
+      if (!me.quartzMe.billingProvider) {
+        dispatch(getBillingProviderThunk())
+      }
+      if (CLOUD && !me.quartzMe.regionCode) {
```
hoorayimhelping, line 45: No need for this `CLOUD` check here, since the outside `if` statement already handles that.

In response to the question I posed in the identity thunks regarding whether the conversion from `identity` to `quartzMe` needs to happen - it doesn't seem like it, since if this code gets run, then `shouldUseQuartzIdentity()` is true, and the fetching of `identity` has already happened, since that's one of the first actions the app performs. Is that a correct or incorrect assessment?

hoorayimhelping (commented):
Looks good! Just a couple of general error handling thoughts and questions. A lot of the comments I made are repeated, since they apply to the entire structure

src/identity/utils/shouldUseQuartzIdentity.ts
```diff
@@ -0,0 +1,6 @@
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
+export const shouldUseQuartzIdentity = (): boolean =>
+  isFlagEnabled('uiUnificationFlag') && isFlagEnabled('quartzIdentity')
+    ? true
+    : false
```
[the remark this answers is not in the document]
wdoconnell: Good point! Simplified this to:

`export const shouldUseQuartzIdentity = (): boolean => isFlagEnabled('uiUnificationFlag') && isFlagEnabled('quartzIdentity')`

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,107 @@
...
+  try {
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Loading))
+
+    const currentState = getState()
+    const accountId = currentState?.identity?.account?.id
+
+    const accountDetails = await fetchAccountDetails(accountId)
+
+    dispatch(setCurrentBillingProvider(accountDetails.billingProvider))
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Done))
```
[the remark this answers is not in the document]
wdoconnell: The use of the optional chaining operator is really due to paranoia here, and may be more defensive than necessary. `currentState.identity.account.id` will exist, and should always be a number, since the default value in `initialState` for `accountId` is `0`, and the only values assigned to `accountId` are primitives of type `number` returned by the `identity` endpoint.

If we wanted to be extra safe, we could throw an error to be caught by the `catch` block:

  `if (typeof accountId !== 'number') {
      throw new Error(
        'Unable to retrieve billing provider. Account ID value is not a number.'
      )
    }`

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const updatedState = getState()
+    const legacyMe = convertIdentityToMe(updatedState.identity)
+    dispatch(setQuartzMe(legacyMe, RemoteDataState.Done))
+    dispatch(setQuartzMeStatus(RemoteDataState.Done))
```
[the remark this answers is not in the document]
wdoconnell: Right now, `convertIdentityToMe` needs to run any time `identity` is updated. 

The selectors in the app all read data from the `quartzMe` state, not the `identity` state (even if the flag is enabled) to minimize risk of breaking anything and amount of code that needs to be changed or guarded by an `if` block. So the flow when the `quartzIdentity` flag is on is always (1) retrieve data from endpoint; (2) update `identity` state; (3) convert our current identity state to the shape of `quartzMe` -- even if we're only changing a couple properties (e.g., the `org` data) or even one property (e.g. `billingProvider`)

Some alternatives:
- Add `if (isFlagEnabled('quartzIdentity'))` throughout the app so we don't need to convert to quartzMe at all when the flag is on.
- Add some new reducers to `quartzMe` so that we can update say, **just** the billingProvider property in the `quartzMe` state, rather than running `convertIdentityToMe` on the whole state any time we fetch a new set of org information or billing provider.

I'd advise against those alternatives - they'd have us adding even more code to support an interface that we're retiring. Running `convertIdentityToMe` isn't computationally expensive, so I think it's okay to have that re-run for now when the identity flag is on.

src/identity/reducers/index.ts
wdoconnell: Correct, `accountId` is the odd man out - it comes back as a number
Screen Shot 2022-06-21 at 10 45 59 AM
.

src/organizations/components/OrgProfileTab/index.tsx
wdoconnell: Good point - removed the `CLOUD` check that is already happening.

Re: other question, it's correct that if this code is running, we're using `quartzIdentity`. But see earlier discussion re: need to convert to legacy `me` state any time `identity` is updated, since they need to stay in sync, and since the selectors in the app pull from `quartz/me` whether the flag is on or not.

src/identity/actions/thunks/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    dispatch(setQuartzIdentity(quartzIdentity))
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Done))
+
+    const legacyMe = convertIdentityToMe(quartzIdentity)
+    dispatch(setQuartzMe(legacyMe, RemoteDataState.Done))
+    dispatch(setQuartzMeStatus(RemoteDataState.Done))
+  } catch (error) {
+    dispatch(setQuartzIdentityStatus(RemoteDataState.Error))
+    dispatch(setQuartzMeStatus(RemoteDataState.Error))
+    console.error(error)
```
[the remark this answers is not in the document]
wdoconnell: Right now, when quartz (quartz me or identity) is pinged by the UI, the app doesn't make use of `RemoteDataState.Error` or do anything with the error (other than logging it). The app can't meaningfully carry on in a cloud environment if it can't fetch the user's information from quartz.

An improvement here (and probably elsewhere in the UI) would be to convert these `console.log` or `console.error` calls to toasts that notify the user there's an auth issue.

src/identity/actions/thunks/index.ts
wdoconnell: Similar to above - we could add a type check to see whether it's a `string`, but the endpoint and default state values for this are both strings.

src/billing/components/BillingPageContents.tsx
```diff
@@ -10,13 +13,30 @@ import MarketplaceBilling from 'src/billing/components/marketplace/MarketplaceBi
...
+    }
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
+
   if (
     (quartzMe?.accountType === 'pay_as_you_go' ||
       quartzMe?.accountType === 'contract') &&
-    quartzMe?.billingProvider !== 'zuora'
+    // This additional check is needed because, on page load, billingProvider (esp. with /identity on) may be 'null', which !== 'zuora. A 'null' billingProvider corresponds to free accounts.
+    quartzMe.billingProvider !== null &&
+    quartzMe.billingProvider !== 'zuora'
```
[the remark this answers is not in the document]
wdoconnell: Don't think we need the additional case since `billingProvider` is only `null` [when the account type is free.](https://influxdata.slack.com/archives/CHZ011F7E/p1655303863607759)

PR #4826, opened 2022-06-15
chore: remove not found flag

asalem1:
This PR cleans up the NotFound page to remove the old 404 page

### DIFF
--- a/src/shared/components/NotFound.tsx
+++ b/src/shared/components/NotFound.tsx
@@ -29,7 +29,7 @@ import GetInfluxButton from 'src/shared/components/GetInfluxButton'
 import {Organization} from 'src/types'
 
 const NotFoundNew: FC = () => (
-  <AppWrapper type="funnel" className="page-not-found" data-testid="not-found">
+  <AppWrapper type="funnel" className="page-not-found" testID="not-found">
     <FunnelPage enableGraphic={true} className="page-not-found-funnel">
       <FlexBox
         direction={FlexDirection.Row}


src/shared/components/NotFound.tsx
```diff
@@ -134,63 +124,41 @@ const NotFoundNew: FC = () => (
...
 
-  render() {
-    if (this.state.isFetchingOrg) {
-      // don't render anything if this component is actively fetching org id
-      // this prevents popping in a 404 page then redirecting
-      return null
+    const deepLinkingMap = buildDeepLinkingMap(org)
+
+    if (deepLinkingMap.hasOwnProperty(location.pathname)) {
+      event('deeplink')
```
appletreeisyellow, line 142: Do we want to include the pathname in event?

```suggestion
      event('deeplink', {from: location.pathname})
```

appletreeisyellow (approved):
Thanks for cleaning it up!

src/shared/components/NotFound.tsx
```diff
@@ -134,63 +124,41 @@ const NotFoundNew: FC = () => (
...
-      if (deepLinkingMap.hasOwnProperty(this.props.location.pathname)) {
-        event('deeplink', {from: this.props.location.pathname})
-        this.props.history.replace(deepLinkingMap[this.props.location.pathname])
-        return
-      }
-      this.setState({isFetchingOrg: false})
+  const handleDeepLink = useCallback(async () => {
+    if (!org) {
+      setIsFetchingOrg(true)
+      await fetchOrg()
```
hoorayimhelping, line 136: What's the reason not to use the `org` value returned from `fetchOrg`? The idea was when a logged out user hit this page and then logs in, they would be immediately redirected to the link they were trying to go to. It looks like this will cause unecessary re-rendering since the component will have to wait for redux to be populated with `org`

hoorayimhelping (changes requested):
This introduces changes beyond removing a feature flag - it doesn't just change a class component to a function component, it changes the flow of how the component is rendered and how it uses data. Some of the code you changed was put in place to handle issues that only crop up in production environments that are very hard to test locally due to how we fake authentication. It also changes what analytics data is sent to the backend by this feature.

If you want to make these changes, I think they should at the very least be done in separate PRs - the intent of this is to remove a feature flag. I'd also expect that you'd let the Dumplings team know that you're changing the code of a high visibility feature we maintain.

src/shared/components/NotFound.tsx
hoorayimhelping: Yeah, you're right it's kind of odd. I think I meant to add that in afterwards but never got around to it. I think it's related to the fact that we have a `GetOrganizations` that dispatches the action after `getOrganizations` is called.

src/shared/components/NotFound.tsx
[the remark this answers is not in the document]
hoorayimhelping: @asalem1 Ah, I see where your head's at with this, but at the point this event is fired, `location.pathname` will be something like `/me/billing` or `/me/about` - it won't include the org id. It'll be one of the 30ish defined deeplinks which will produce a finite list of tags. Do you think this would be better as a measurement rather than a tag?

You might be right that it may not be necessary or even helpful. But either way, I'd appreciate a heads up on this kind of thing in the future so we can be aware that any analytics that we're using will be changing.

PR #4834, opened 2022-06-17
feat: requery with a tighter window upon zoom-in

TCL735:
Closes #14800
Closes #14801

Zooming in on a graph re-issues the query with a tighter window period

https://user-images.githubusercontent.com/10736577/174201733-688f3d11-101e-4ea3-86eb-264afc513603.mp4

### DIFF
--- a/src/variables/selectors/index.tsx
+++ b/src/variables/selectors/index.tsx
@@ -5,7 +5,10 @@ import {get} from 'lodash'
 import {getActiveQuery} from 'src/timeMachine/selectors'
 import {getRangeVariable} from 'src/variables/utils/getTimeRangeVars'
 import {getTimeRange, getTimeRangeWithTimezone} from 'src/dashboards/selectors'
-import {getWindowPeriodVariableFromVariables} from 'src/variables/utils/getWindowVars'
+import {
+  getWindowPeriodVariableForZoomRequery,
+  getWindowPeriodVariableFromVariables,
+} from 'src/variables/utils/getWindowVars'
 import {
   TIME_RANGE_START,
   TIME_RANGE_STOP,
@@ -131,6 +134,25 @@ export const getAllVariables = (
       return prev
     }, [])
     .filter(v => !!v)
+  return vars
+}
+
+export const getAllVariablesForZoomRequery = (
+  state: AppState,
+  contextID?: string
+): Variable[] => {
+  const vars = getUserVariableNames(state, contextID || currentContext(state))
+    .concat([TIME_RANGE_START, TIME_RANGE_STOP, WINDOW_PERIOD])
+    .reduce((result, variableID) => {
+      if (variableID === WINDOW_PERIOD) {
+        result.push(getWindowPeriodVariableForZoomRequery())
+      } else {
+        result.push(getVariable(state, variableID))
+      }
+      return result
+    }, [])
+    .filter(v => !!v)
+
   return vars
 }
 
--- a/src/visualization/types/Graph/view.tsx
+++ b/src/visualization/types/Graph/view.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FC, useMemo, useContext} from 'react'
+import React, {FC, useMemo, useContext, useState, useEffect} from 'react'
 import {useDispatch, useSelector} from 'react-redux'
 import {
   Config,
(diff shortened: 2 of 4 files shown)


TCL735:
Very likely more changes coming soon. Just putting this up to get the review process started.

src/visualization/types/Graph/view.tsx
```diff
@@ -70,6 +78,15 @@ const XYPlot: FC<Props> = ({
...
   cellID,
 }) => {
+  const [resultState, setResultState] = useState(result)
+  const [preZoomResult, setPreZoomResult] = useState<InternalFromFluxResult>(
+    null
+  )
+
+  useEffect(() => {
+    setResultState(result)
+  }, [result])
```
hoorayimhelping, line 88: what's the difference between `resultState` and `result` here? I'm having trouble understanding the purpose of using `resultState`, when it seems like `result` is being changed to `resultState` lower down.

hoorayimhelping (approved):
This is some nice-to-read code, nice job! I just had a couple of small comments, let me know what you think

src/variables/selectors/index.tsx
```diff
@@ -134,6 +137,25 @@ export const getAllVariables = (
...
+  const vars = getUserVariableNames(state, contextID || currentContext(state))
+    .concat([TIME_RANGE_START, TIME_RANGE_STOP, WINDOW_PERIOD])
+    .reduce((result, variableID) => {
+      if (variableID === WINDOW_PERIOD) {
+        result.push(getWindowPeriodVariableForZoomRequery())
+      } else {
+        result.push(getVariable(state, variableID))
+      }
+      return result
+    }, [])
```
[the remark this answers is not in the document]
TCL735: Yes, this is right. I will make the change to `map`

src/visualization/types/Graph/view.tsx
TCL735: `result` comes from the props and is responsible for the "normal" flow of the data. As the graph is updated (manually hitting the refresh button, for example), we want the graph's current state to match that and have the normal results.

`resultState` can also be affected by zooming in, which temporarily hijacks the graph and shows more fine grained detail. Hence, it is a local state for the view. If the user un-zooms, then we want to go back to the normal flow of data. Or if the user never zooms at all, we don't want to interfere with the `result`

TCL735:
> I might be missing something, but why aren’t we reissuing the query based on the zoomed in time range and narrower window period? Seems like we’re setting state in multiple places and causing divergent sources of truth in this PR

There is only one source of truth: `resultState` which was previously called `result`. The reason for this change is because instead of just being from props, it now needs to consider local state as well. Two different places can modify it, instead of one. I will explain below why that is.

First, the props can still modify the `resultState` just like it was when it was called `result`. This is straightforward. Props modify the results of the graph. No changes here. We of course would like to keep this going forward. Hence, whenever the prop `result` is updated, we update `resultState` to match it.

Second, `resultState` can modified by a user action --- zooming in. We use local state to keep track of this. The user can zoom in any number of times, so it makes sense to keep it as a local state. The user can also decide to revert back to the original state before zooming. Which we also keep track of with a different state variable.

The second part is the whole feature. This makes the feature work: zooming-in re-queries with a tighter window period. We have to make this change. It's a lot better than re-creating an entire view and swapping views in and out. By doing it with local state, we take advantage of Giraffe's built-in zoom and un-zoom callbacks. The architecture is in place to handle this type of feature. The alternative would be to ignore this and make something more complex an unwieldy.

TCL735:
> why aren’t we reissuing the query based on the zoomed in time range

This is a good question. The feature epic's requirements do no ask for a new time range. That doesn't mean we can't change the requirements. But this may need some thought and weighing of pros and cons.

Re-issuing the same query with only the window period change _may_ have some performance advantages on the backend? I emphasize _may_. I don't actually know. It seems like it would make sense for that to be the case, because it's basically the same query with less aggregating.

On the other hand, shrinking the time range makes the payload smaller. We are only displaying a portion of the payload anyways (the zoomed-on domain), so it makes a lot of sense to not ask for extra lines of csv that aren't even going to be rendered.

Good point and definitely something to think about.

I am going to go ahead and dismiss this review. But please understand that your suggestion is definitely being considered. The immediate goal is to get this mvp out the door. This feature has a lot of room for iteration and improvement.
t

TCL735:
Update: @asalem1 we will be adjusting the time range as a separate issue with a separate PR. Thanks for catching this and your feedback!

TCL735:
> Not sure why my previous review was dismissed but I guess it doesn't matter 🤷🏽‍♂️

See my responses above.

PR #2662, opened 2021-09-20
chore: notebooksd changes integration

genehynson:
Adds integration with new notebooksd routes.

✅  Merge this first: https://github.com/influxdata/ui/pull/2664

PR #4838, opened 2022-06-17
chore: normalize function browsing

drdelambre:
we've been spinning out new function browsers for every interaction. this has run into issues in production where certain iterations haven't been developed to include logic connected to services or feature flags correctly. some would handle them fine, others wouldn't show up, or have the desired effect when a flag was toggled. there were even weird artifacts like us rendering the entire list of functions for a user, but not allowing them to scroll through it. or a feature flag that could only inject a function if the monaco editor had rendered first. or an instance that didn't support flags at all or use our standardized design language. instead of asking the developer to know about all of these instances and their nuances, considering that they're all the same interaction expectation from product, lets iterate on one component.

Peek 2022-06-17 10-20


src/shared/components/FilterList/InjectionOption.tsx
```diff
@@ -30,25 +31,39 @@ interface Props {
   setHoverdFunction?: (string: string) => void
 }
 
-const defaultProps = {
```
drdelambre, line 33: weird artifact of us learning functional components from a bit ago

src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -63,6 +65,7 @@ const FluxEditorMonaco: FC<Props> = ({
   const editorDidMount = (editor: EditorType) => {
     prelude.current = setupForReactMonacoEditor(editor)
     prelude.current.updatePreludeModel(variables)
+    setEditor(editor)
```
drdelambre, line 68: this reduces a bunch of prop drilling and has a graceful failure case if the provider is not defined

src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -44,16 +41,6 @@ const TMFluxEditor: FC<{variables: Variable[]}> = props => {
     dispatch(saveAndExecuteQueries())
   }
 
-  const handleInsertVariable = (variableName: string): void => {
```
drdelambre, line 47: this was moved into the toolbar itself

src/shared/components/FilterList/InjectionOption.tsx
```diff
@@ -30,25 +31,39 @@ interface Props {
...
   onClick,
-  testID,
+  testID = 'flux-injection-option',
   ToolTipContent,
   searchTerm,
   setToolTipPopup,
   setHoverdFunction,
 }) => {
+  // TODO: move this to a flag once design wants to branch it out further
+  const {fluxQueryBuilder} = useContext(AppSettingContext)
```
drdelambre, line 45: this triggers the styling changes that are only applied to the function browser in the new data explorer

src/timeMachine/components/FluxToolbar.tsx
```diff
@@ -30,17 +29,26 @@ const FluxToolbar: FC<Props> = ({onInsertFluxFunction, onInsertVariable}) => {
 
   let activeToolbar
 
+  const onInsertFluxFunction = (fn: FluxFunction | FluxToolbarFunction) => {
```
drdelambre, line 32: moved in from parent component since we might as well reference the context where we use it

src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -74,37 +61,31 @@ const TMFluxEditor: FC<{variables: Variable[]}> = props => {
...
-        <div className="flux-editor--right-panel">
-          <FluxToolbar
-            onInsertFluxFunction={handleInsertFluxFunction}
-            onInsertVariable={handleInsertVariable}
+          }
+        >
+          <FluxEditor
+            script={activeQueryText}
+            variables={props.variables}
+            onChangeScript={handleActiveQuery}
```
wiedld, line 78: For other reviewers (e.g. from Dumplings):
* the FluxToolbar now handles the injection of variables and functions.
  * mostly same code, moved into the FluxToolbar
  * but it no longer has a callback to handleSetActiveQueryText() --> set redux state
* But we don't need this callback, because of the onChangeScript
  *  onChange in editor text -->  handleActiveQuery() --> handleSetActiveQueryText() --> redux

src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -74,37 +61,31 @@ const TMFluxEditor: FC<{variables: Variable[]}> = props => {
     [activeQueryText, handleSetActiveQueryText]
   )
 
-  return useMemo(() => {
```
wiedld, line 77: Memoization was put in to reduce re-rendering when filtering/searching on flux functions.
Note to other reviewers --> confirmed that it's no longer needed. All clear.

src/shared/components/FluxMonacoEditor.tsx
wiedld: Like this change.

Note that you missed one in the `src/flows/pipes/RawFluxEditor`.
Once you update that, then the prop `setEditorInstance` can be deleted.

src/dataExplorer/components/SidePane.tsx
```diff
@@ -1,15 +1,48 @@
...
+
+  if (CLOUD && isFlagEnabled('fluxDynamicDocs')) {
+    browser = <DynamicFunctions onSelect={inject} />
+  }
+
   return (
     <div className="container-right-side-bar">
-      <FluxBrowser />
+      <SelectorTitle title="Flux library" info={TOOLTIP} />
+      {browser}
```
wiedld, line 45: Please make the view and logic into a subcomponent.

Since you have another PR currently open, where the menu in the SidePane/SideBar is a subcomponent.

src/shared/components/FilterList/InjectionOption.tsx
```diff
@@ -30,25 +31,39 @@ interface Props {
...
+  }
+
+  const classer = [
+    ['flux-toolbar--list-item', true],
+    ['flux-toolbar--function', true],
+    ['flux-toolbar--new-style', fluxQueryBuilder],
+  ]
+    .filter(c => c[1])
+    .map(c => c[0])
+    .join(' ')
```
wiedld, line 65: `classer` is a nice way to reduce a lot of code, but we need signon (for this new pattern) from dumplings. 

Since this is shared code, and we are effectively collapsing several shared components into single DRY components --> but with different styling.

src/shared/components/FilterList/InjectionOption.tsx
drdelambre: The new view only shows up on the new data explorer and not in the old data explorer. Design is unclear about the roll out plan, so it's not just a flag that is rolled out everywhere

wiedld: Using a globally set value `fluxQueryBuilder`, which toggles on in the new QX FluxBuilder, means that everywhere else in the codebase needs to reset this global value back to false --> whenever we navigate away from the new QX FluxBuilder.

Otherwise this new style change starts popping up in other parts of the application:

https://user-images.githubusercontent.com/10232835/174398386-a3f10288-0cb4-48e6-8fb2-7c0331735505.mov

src/shared/components/FilterList/InjectionOption.tsx
```diff
@@ -74,22 +89,23 @@ const FluxInjectionOption: FC<Props> = ({
...
+        {!fluxQueryBuilder && (
+          <Button
+            testID={`flux--${testID}--inject`}
+            text="Inject"
+            onClick={handleClick}
+            size={ComponentSize.ExtraSmall}
+            className="flux-toolbar--injector"
+            color={ComponentColor.Primary}
+          />
+        )}
```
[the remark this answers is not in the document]
drdelambre: No. Design is working on a new design language that mirrors the interaction in the schema browser, and would like to roll it out after testing it, but didn't have a rollout plan

src/dataExplorer/components/SidePane.tsx
```diff
@@ -1,15 +1,48 @@
...
+      })
+    },
+    [injectFunction]
+  )
+
+  let browser = <Functions onSelect={inject} />
+
+  if (CLOUD && isFlagEnabled('fluxDynamicDocs')) {
+    browser = <DynamicFunctions onSelect={inject} />
+  }
```
wiedld, line 40: Thanks for bringing back in support for both with, and without, the fluxDynamicDocs.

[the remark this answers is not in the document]
drdelambre: If the docs service hasn't been deployed in the cluster, and the flag is on for that cluster, this wont have a list of functions to display. A comment here seems like it'll get lost when the info matters (deploying clusters or toggling the flag). Any ideas where else we capture this? We also want to sequence that with plans to unify all these means to pull in docs

wiedld (changes requested):
I really like the reduction in code, but we have a few changes required. And we need Dumplings approval.

Additionally, the SecretsInject is no longer working. (Yes, this reminds me to write tests -- doing today).
Secrets are no longer listed, with this PR:
Screen Shot 2022-06-17 at 1 33 22 PM
I confirmed that secrets are still listed on master, so it's a specific change here.

src/shared/components/DynamicFunctionsList.tsx
```diff
@@ -53,10 +53,12 @@ const DynamicFunctionsList: FC<Props> = ({onSelect}) => {
     }
   }, [hoveredFunction, tooltipPopup, eventSearchTerm])
 
-  const handleSelectItem = useCallback((func: FluxFunction) => {
-    onSelect(getFluxExample(func))
-    event('flux.function.injected', {name: `${func.package}.${func.name}`})
```
wiedld, line 58: Also...please make sure Dumplings is on board with removing this event, as they placed it in.

src/shared/components/FilterList/InjectionOption.tsx
drdelambre: this pattern has been in the codebase for over a year, the one that comes to mind is from this PR:
https://github.com/influxdata/ui/pull/2662/files#diff-0823902a1cc6d2e5d230a5665bb76c1ab24842cabb58ebe322346eaa4290a167R106

but in it's essence it's the same pattern as 'classnames' without the library import

src/dataExplorer/components/SidePane.tsx
drdelambre: i dont understand the distinction you're trying to make here

src/shared/components/DynamicFunctionsList.tsx
drdelambre: the person who might maybe want to use this sometime for something is gone, so feature matched on other side of the event move

wiedld: @blegesse-git added this event metric, as part of a PR specifically for adding metrics. Think we should at least ping and make sure it's ok to remove. 🤷🏼‍♀️

wiedld (commented):
The QA blockers to merge are now resolved (SecretsList is back, eventing is equivalent to previous, no longer have styles bleeding into other parts of the application). Thank you!

Still waiting on some feedback from Dumplings, to make sure we are good-to-go. (This touches a lot of code they have recently updated.)

wiedld (approved):
Dumplings gave the go ahead to merge.

PR #4840, opened 2022-06-17
feat: add back button on edit subscription page

subirjolly:
# BEFORE
BEFORE

# AFTER
AFTER


helenosheaa:
is there a related issue for this?

I know on the create page where the form is broken up into three stages still there is a back button already and then for the edit page we are only showing users the single page view I thought so a back button isn't needed?

Screen Shot 2022-06-17 at 2 32 29 PM


do you have `subscriptionsSinglePage | true` if you look at influx.list in the console? We should have that set to true everywhere as I don't think we are supporting the old view.

Screen Shot 2022-06-17 at 2 37 48 PM

Issue #4849, opened 2022-06-19
Error context is lost when following link to create an issue


steven-solomon:
## About the bug

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. Cause any error that displays `src/shared/components/DefaultErrorMessage.tsx`.
2. Click the link to create an issue.

**Expected behavior:**
As a user I want to keep open the page containing the error context so that I can report the details of the issue.

**Actual behavior:**
We are routed to GitHub within the same tab as the error, and navigating back shows that the scenario that caused is not maintained in browser history.

**Visual Proof:**

## About your environment

**Environment info:**

N/A

**Config:**

**Logs:**

**Performance:**
Generate profiles with the following commands for bugs related to performance, locking, out of memory (OOM), etc.

```sh
# Commands should be run when the bug is actively.
# Note: This command will run for at least 30 seconds.
curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all?cpu=true"
curl -o vars.txt "http://localhost:8086/debug/vars"
iostat -xd 1 30 > iostat.txt
# Attach the `profiles.tar.gz`, `vars.txt`, and `iostat.txt` output files.
```

PR #4850, opened 2022-06-19
fix: Open issue link in new tab

steven-solomon:
Closes https://github.com/influxdata/ui/issues/4849

I added the `target` attribute to the issue anchor tag.

### DIFF
--- a/src/shared/components/DefaultErrorMessage.tsx
+++ b/src/shared/components/DefaultErrorMessage.tsx
@@ -11,7 +11,7 @@ const DefaultErrorMessage: ErrorMessageComponent = () => {
       style={{display: 'flex', placeContent: 'center'}}
     >
       An InfluxDB error has occurred. Please report the issue&nbsp;
-      <a href="https://github.com/influxdata/ui/issues/new?template=bug_report.md">
+      <a href="https://github.com/influxdata/ui/issues/new?template=bug_report.md" target="_blank">
         here
       </a>
       .


hoorayimhelping:
Hi @steven-solomon,

Thank you for the PR and the corresponding issue. I think this is a good usability change.

We have a requirement (that is incorrectly undocumented on our PR template) that contributors sign our [Community License Agreement.](https://www.influxdata.com/legal/cla/). If you haven't signed it, it needs to be signed before we can accept this PR.

We also have an internal convention to add a `rel` of `noopener noreferrer` to links with a `target` of `'_blank'`, even though modern browsers protect against reverse tabnabbing attacks. We have an [internal component](https://github.com/influxdata/ui/blob/master/src/utils/SafeBlankLink.tsx) to handle this.

In consideration of your time, I've created an [internal PR](https://github.com/influxdata/ui/pull/4852) with you listed as a co-author that recreates this functionality within our conventions.

We appreciate community contributions, but we've recently changed some of our security policies and it looks like we need to update our process to accept contributions from non team-members to take these security policies into account. Rather than force you to wait on us for that, I'll add in this functionality  and give you credit in the commit message.

hoorayimhelping:
This functionality was added by commit [`c2015ba8e70e7231d096b8b80cc25cb41c72a305`](https://github.com/influxdata/ui/commit/c2015ba8e70e7231d096b8b80cc25cb41c72a305).

PR #4852, opened 2022-06-21
chore(DefaultErrorMessage): open new ui issue link in new tab/window

### LINKED ISSUE #4849: Error context is lost when following link to create an issue

hoorayimhelping:
Co-authored by: Steven Solomon <steven-solomon@github.com>

Closes #4849

Copied from: https://github.com/influxdata/ui/pull/4850, which was from a community contributor.

### DIFF
--- a/src/shared/components/DefaultErrorMessage.tsx
+++ b/src/shared/components/DefaultErrorMessage.tsx
@@ -4,6 +4,8 @@ import React from 'react'
 // Types
 import {ErrorMessageComponent} from 'src/types'
 
+import {SafeBlankLink} from 'src/utils/SafeBlankLink'
+
 const DefaultErrorMessage: ErrorMessageComponent = () => {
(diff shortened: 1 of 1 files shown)


wdoconnell (approved):
Looks good to me.

PR #4861, opened 2022-06-21
chore: add events to close gaps in eventing

### LINKED ISSUE #4807: Review instrumentation design guide for new query builder

asalem1:
Closes #4807 

This PR adds a few key events that are missing for us to make wholesale metric analysis on user behavior based on the product spec. More specifically, this PR adds:

1. An event for when the user leaves the Data Explorer page
2. An event for when the user types in the Flux Editor in the New Query Builder
3. An event for when a user selects a bucket in the New Query Builder
4. An event for when a user selects a measurement in the New Query Builder
5. An event for when a user searches in the New Query Builder
6. An event for when a user toggles between the New and Old Query Builder
7. An event for when users selects a tag / value / searches tag keys / search tag values in the Old Query Builder
8. An event for when a user selects a bucket in the Old Query Builder

### DIFF
--- a/src/dataExplorer/components/DataExplorerPage.tsx
+++ b/src/dataExplorer/components/DataExplorerPage.tsx
@@ -57,7 +57,7 @@ const DataExplorerPage: FC = () => {
     return () => {
       event('Exited Data Explorer')
     }
-  })
+  }, [])
 
   return (
     <Page titleTag={pageTitleSuffixer(['Data Explorer'])}>


src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -53,6 +53,12 @@ const DataExplorerPage: FC = () => {
     event('Data Explorer Page - Clicked Notebooks CTA')
   }
 
+  useEffect(() => {
+    return () => {
+      event('Exited Data Explorer')
```
wiedld, line 58: useEffect gets triggered much more often, including:
* page load
* toggling between old and new data explorer.

wiedld (changes requested):
Please change the "exit dataExplorer" event, since that's not when it's triggered.

wiedld (approved):
LGTM

PR #4864, opened 2022-06-21
feat: update Quartz api client SHA

eatondustin1:
Update the SHA for the generated Quartz api client.


wdoconnell (approved):
Looks good to me. Thanks for adding `isUpgradeable`. 👍

PR #4879, opened 2022-06-23
fix: csv not respecting time range

drdelambre:
found this bug where the download csv button wasn't respecting the user's selected time range, should be fixed before launch


appletreeisyellow (approved):
I like how you rearranged the import files into different categories!

PR #4881, opened 2022-06-23
feat: attempt to normalize the bucket language

drdelambre:
the addition of sample data psuedo buckets into the selectable bucket list was a big win for new users in the notebook's query builder, so the state was adopted for the flux query builder. but we forgot to bring over the design language! this left a couple of buckets at the end of the list with no information about where they came from or why they were added. here is an attempt to add the section headers to bring this context back from the query builder and into the bucket selector dropdown.

@juliajames12 look ok?

Peek 2022-06-23 12-40

### DIFF
--- a/src/dataExplorer/components/BucketSelector.tsx
+++ b/src/dataExplorer/components/BucketSelector.tsx
@@ -58,7 +58,7 @@ const BucketSelector: FC = () => {
     <Dropdown.Button
       onClick={onClick}
       active={active}
-      testID="bucket-selector--dropdown"
+      testID="bucket-selector--dropdown-button"
     >
       {buttonText}
     </Dropdown.Button>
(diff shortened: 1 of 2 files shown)


drdelambre:
heres the change julia suggested in back channels

Screenshot from 2022-06-23 13-10-39

src/dataExplorer/components/BucketSelector.tsx
```diff
@@ -1,51 +1,174 @@
...
+    selectBucket(buck)
+  }
+
+  const handleChange = (e: ChangeEvent<HTMLInputElement>): void => {
+    setSearchTerm(e.target.value)
+  }
+
+  let buttonText = 'Loading buckets...'
+  if (loading === RemoteDataState.Done && !selectedBucket?.name) {
+    buttonText = 'Choose a bucket'
```
appletreeisyellow, line 52: do we want to keep the button text consistent with the [design](https://www.figma.com/proto/0qAntPk5LVangAHguWT74X/Query-Experience-Project?page-id=191%3A13710&node-id=567%3A21115&viewport=241%2C48%2C0.15&scaling=min-zoom&starting-point-node-id=567%3A21115) using "Select bucket..."?

appletreeisyellow (approved):
Nice! The bucket dropdown looks better with dividers! Not a blocker, just a comment on placeholder

PR #4885, opened 2022-06-24
fix: use one-way state to capture the initial, pre-zoom domain

TCL735:
Part of idpe 14803

Fixes a bug where the initial, pre-zoom domain is not equal to the selected time range.
Here is a video of the bug:

https://user-images.githubusercontent.com/10736577/175443825-e236284d-7939-499b-ac1f-aba1a2fe09d7.mp4

  
  
With the fix, we use one-way state to lock in the pre-zoom domain which is equal to the selected time range.
Here is a video of the fix:

https://user-images.githubusercontent.com/10736577/175444116-44d0a682-1d74-4604-9971-6dcec66c5d56.mp4


wdoconnell (approved):
Looks good to me. Nice fix.

PR #4901, opened 2022-06-27
feat: users with no configs redirect to overlay

### LINKED ISSUE #4760: firstMile: Bug - The Telegraf option on the homepage should navigate users to the "Create a Telegraf Configuration" Modal
## Current State / Problem
- As a new user, when I click the Telegraf option on the new Home Page, I'm navigated to the Telegraf page showing existing configurations. However, as a new user, I don't have any configurations. 

## Idea
- As a new user, Help me get started with Telegraf faster by automatically displaying the "Create a Telegraf Configuration" Modal. 

## ACs
- Upon clicking on the Telegraf option on the new homepage, users should be navigated to the "Create a Telegraf Configuration" Modal on the Telegraf page. 

## Screenshot of "Create a Telegraf Configuration" Modal
Image

david-rusnak:
Closes #4760 

Redirects users with no telegraf configs to the config overlay when clicking on the homepage telegraf link

https://user-images.githubusercontent.com/106361125/175976049-f61ae7f2-0579-4371-9a96-82ca7763551b.mov

### DIFF
--- a/src/homepageExperience/containers/HomepageContainer.tsx
+++ b/src/homepageExperience/containers/HomepageContainer.tsx
@@ -64,14 +64,13 @@ export const HomepageContainer: FC = () => {
 
   useEffect(() => {
     dispatch(getTelegrafs)
-  }, [telegrafs])
+  }, [telegrafs]) // eslint-disable-line react-hooks/exhaustive-deps
 
   const telegrafLink = () => {
-    if (telegrafs.length != 0) {
-      return telegrafPageLink
-    } else {
+    if (!telegrafs.length) {
       return newTelegrafPageLink
     }
+    return telegrafPageLink
   }
 
   // events handling
--- a/src/homepageExperience/containers/HomepageContainer.tsx
+++ b/src/homepageExperience/containers/HomepageContainer.tsx
@@ -1,5 +1,5 @@
-import React, {FC} from 'react'
-import {useSelector} from 'react-redux'
+import React, {FC, useEffect} from 'react'
+import {useSelector, useDispatch} from 'react-redux'
 import {Link} from 'react-router-dom'
 
 import {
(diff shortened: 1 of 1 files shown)


ChitlangeSahas (approved):
Awesome!

src/homepageExperience/containers/HomepageContainer.tsx
```diff
@@ -53,6 +62,18 @@ export const HomepageContainer: FC = () => {
 
   const squareGridCardSize = '200px'
 
+  useEffect(() => {
+    dispatch(getTelegrafs)
+  }, [telegrafs])
+
+  const telegrafLink = () => {
+    if (telegrafs.length != 0) {
```
hoorayimhelping, line 70: It's generally better to use strict equality with JavaScript, so like

```ts
telegrafs.length !== 0
```
because in JavaScript `"0" == 0`

It's not a huge deal with `.length` since it'll be a number, but it's a good habit to get into as a default, so that in the cases where you actually need `"0"` to equal `0`, the comparison `myValue == 0` can be appropriately called out as intentional.

Also, for this specific check, it's probably better, because it's more explicit, to say:

```ts
if (telegrafs.length > 0)
```
since we want to make sure that something comes back from the telegrafs api request, and technically if `telegrafs.length === -1`, it will pass `!= 0`.

It's perfectdly idiomatic JavaScript to just shorten it to

```ts
if (telegrafs.length)
```
however. Either one is acceptable.

Also, there's no need for an extra `else` since both these statements `return`. It's good to determine what the default/happy path is, and code around that.

```ts
if (telegrafs.length) {
  return telegrafPageLink
}

return newTelegrafPageLink
```

The difference here is subtle, but it's there. The `if/else` form gives equal weight to both choices, and doesn't provide a default or backup case. The `if some case, return early` form subtlely implies that the case the `if` statement is checking for is the exception, because the default case just returns. So think about which case is most likely what the default/happy case is, and make the other case the `if case -> return`

hoorayimhelping (commented):
Have a small code-style suggestion - it's a lot of explanation for a simple line, but read over it and let me know what you think.


Screen Shot 2022-06-27 at 12 21 12 PM

hoorayimhelping (approved):

PR #4908, opened 2022-06-27
fix: fix incorrect provider information

### LINKED ISSUE #4597: Bug: PAYG Org Settings - Stop displaying Billing Provider under "Provider"

david-rusnak:
Closes #4597 

When the `quartzIdentity` feature flag is true, it displays the correct cloud provider, shown here:

Screen Shot 2022-06-27 at 11 29 33 AM

When the `quartzIdentity` feature flag is false, it doesn't display a provider, shown here:

Screen Shot 2022-06-27 at 11 29 54 AM

Edit (After styling update):
Screen Shot 2022-06-27 at 4 54 28 PM

Completed during pair programming with @alexjw3

### DIFF
--- a/src/organizations/components/OrgProfileTab/index.tsx
+++ b/src/organizations/components/OrgProfileTab/index.tsx
@@ -55,7 +55,7 @@ const OrgProfileTab: FC = () => {
   const hasSomeQuartzOrgData =
     identity.org?.provider || me.quartzMe?.regionCode || me.quartzMe?.regionName
 
-  const orgProviderExists = identity.org?.provider
+  const orgProviderExists = !!identity.org?.provider
 
   const OrgProfile = () => (
     <FlexBox.Child
--- a/src/organizations/components/OrgProfileTab/index.tsx
+++ b/src/organizations/components/OrgProfileTab/index.tsx
@@ -53,9 +53,9 @@ const OrgProfileTab: FC = () => {
   const expectQuartzData = CLOUD && isFlagEnabled('uiUnificationFlag')
 
   const hasSomeQuartzOrgData =
-    identity.org.provider || me.quartzMe?.regionCode || me.quartzMe?.regionName
+    identity.org?.provider || me.quartzMe?.regionCode || me.quartzMe?.regionName
 
-  const orgProviderExists = identity?.org?.provider
+  const orgProviderExists = identity.org?.provider
 
   const OrgProfile = () => (
     <FlexBox.Child
--- a/src/organizations/components/OrgProfileTab/index.tsx
+++ b/src/organizations/components/OrgProfileTab/index.tsx
@@ -16,17 +16,18 @@ import CopyableLabeledData from 'src/organizations/components/OrgProfileTab/Copy
 import DeletePanel from 'src/organizations/components/OrgProfileTab/DeletePanel'
 
 // Utils
-import {getOrg} from 'src/organizations/selectors'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {CLOUD} from 'src/shared/constants'
-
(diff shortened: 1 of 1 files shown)


src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -16,42 +16,42 @@ import CopyableLabeledData from 'src/organizations/components/OrgProfileTab/Copy
 import DeletePanel from 'src/organizations/components/OrgProfileTab/DeletePanel'
 
 // Utils
-import {getOrg} from 'src/organizations/selectors'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {CLOUD} from 'src/shared/constants'
+import {shouldUseQuartzIdentity} from 'src/identity/utils/shouldUseQuartzIdentity'
+import 'src/organizations/components/OrgProfileTab/style.scss'
```
wdoconnell, line 22: To keep this clear, let's consider separating these imports out with different comments:

`// Utils` for for the feature flag in utils; `// Constants` for the constants, `// Stylesheets` or equivalent for `scss.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -74,8 +74,8 @@ const OrgProfileTab: FC = () => {
             stretchToFitWidth={true}
             style={{width: '85%'}}
           >
-            {me.quartzMe?.billingProvider && (
-              <LabeledData label="Provider" src={me.quartzMe.billingProvider} />
+            {identity?.org?.provider && (
```
wdoconnell, line 77: Design says we can left-align in the case where only 2 out of 3 boxes exist. 

Could you come up with some logic that would change the style whenever `identity.org.provider` isn't available? (to `$cf-space-2x1`)

Thank you!

david-rusnak: Fixed with adjusting `FlexBox` width depending if provider is available.

wdoconnell: Spacing looks good now. Thanks for handling!

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -16,42 +16,46 @@ import CopyableLabeledData from 'src/organizations/components/OrgProfileTab/Copy
...
     }
   }, []) // eslint-disable-line react-hooks/exhaustive-deps
 
+  const expectQuartzData = CLOUD && isFlagEnabled('uiUnificationFlag')
+
   const hasSomeQuartzOrgData =
-    me.quartzMe?.billingProvider ||
-    me.quartzMe?.regionCode ||
-    me.quartzMe?.regionName
+    identity.org.provider || me.quartzMe?.regionCode || me.quartzMe?.regionName
```
wdoconnell, line 56: Super minor, but let's make the use of optional chaining here consistent for both references to `identity.org` (so that it behaves the same as the references to `me.quartzMe`, and we don't have an uncaught error if `identity` does not have the org property for some reason).

`identity.org?.provider`

wdoconnell (approved):
Thanks for the great work on this, @david-rusnak !

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -16,42 +16,46 @@ import CopyableLabeledData from 'src/organizations/components/OrgProfileTab/Copy
...
+
+  const orgProviderExists = identity?.org?.provider
```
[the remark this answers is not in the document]
wdoconnell: `orgProviderExists` should probably be a boolean so maybe we make this:

`const orgProviderExists = !!identity.org?.provider`

PR #4910, opened 2022-06-27
feat: add schema injection

### LINKED ISSUE #4591: Injection of Schema Selections into Flux Editor

appletreeisyellow:
Closes #4591 

This PR adds the front-end part of the schema injection in MVP version. 

## After

https://user-images.githubusercontent.com/14298407/176258689-06ff42e8-4a4e-4048-b347-b0cbfc030542.mov


src/dataExplorer/context/fluxQueryBuilder.tsx
```diff
@@ -95,6 +124,39 @@ export const FluxQueryBuilderProvider: FC = ({children}) => {
     getTagKeys(selection.bucket, measurement)
   }
 
+  const handleSelectField = (field: string): void => {
+    // Inject field
+    injectViaLsp(ExecuteCommand.InjectField, {
+      bucket:
```
drdelambre, line 130: why do we need the bucket here? isn't that just used on measurement injection for the current iteration?

src/dataExplorer/context/fluxQueryBuilder.tsx
```diff
@@ -95,6 +124,39 @@ export const FluxQueryBuilderProvider: FC = ({children}) => {
...
+    event('flux.schema.injected', {
+      command: ExecuteCommand.InjectField,
+      context: 'flux query builder',
+    })
+  }
+
+  const handleSelectTagValue = (tagKey: string, tagValue: string): void => {
+    // Inject tag value
+    injectViaLsp(ExecuteCommand.InjectTagValue, {
+      bucket:
```
drdelambre, line 146: same question about buckets here

src/dataExplorer/context/fluxQueryBuilder.tsx
appletreeisyellow: It's required in LSP https://github.com/influxdata/flux-lsp/blob/master/src/server/mod.rs#L1333-L1336

Here is my understanding:

No matter the user is injecting measurement, field, or tag value in UI, LSP treats them as injecting `filter()` function. In order to make sure the injection line works, LSP asks for the bucket that filter function is applying to. 

Or there are other intentions? Correct me if I understand it wrong @rockstar

wiedld (approved):

src/dataExplorer/context/fluxQueryBuilder.tsx
wiedld: Append only, to the last flux pipeline in the query.
Confirms that the last flux transformation pipeline has the same bucket as we are seeking.

This is the first iteration of the injection in the LSP.
Plans are already in the works for the next one.

drdelambre (approved):

PR #4911, opened 2022-06-27
chore: update clockface to 6.4.0

appletreeisyellow:
This PR updates clockface version to 4.6.0 (influxdata/clockface#784) on Accordion Header Caret. This component is used in schema browser.


appletreeisyellow:
Hmm, I saw there is a new release in clockface (6.4.1) yesterday. Let me check with dumplings team what's their plan

PR #4912, opened 2022-06-27
chore: update code font and hover color to match Monaco editor

### LINKED ISSUE #4905: Schema browser and function browser design clean up
We'd like to visually connect the injectable items in both the Schema Browser and Functions Browser with the script in the Monaco editor. For this, let's update the following:

- Update the font of the field and tag values to "Roboto Mono"
- Let's double check that the green color is 7CE490 for the fields and tag values on hover
- Update the function list item hover color to 9394FF to match the purple that's in the Monaco editor

appletreeisyellow:
Closes #4905 

This PR updates the font of the field and tag values to "Roboto Mono", and update the hover color of function injectable elements to match with the color in Monaco editor

Note that this PR doesn't change the hover color of fields and tag values in schema browser to match with the green in Monaco editor because of the reason mentioned by Julia [below](https://github.com/influxdata/ui/pull/4912#issuecomment-1168959947)

## Before 

Screen Shot 2022-06-27 at 5 27 32 PM

## After

Screen Shot 2022-06-27 at 5 22 32 PM

Screen Shot 2022-06-28 at 2 17 11 PM

### DIFF
--- a/src/dataExplorer/components/Schema.scss
+++ b/src/dataExplorer/components/Schema.scss
@@ -86,7 +86,7 @@
 .field-selector--list-item--selectable:hover,
 .tag-selector-value--list-item--selectable:hover {
   cursor: pointer;
-  color: $c-krypton;
+  color: #7CE490;
   background-color: $cf-grey-15;
   border-radius: 2px;
   width: fit-content;
--- a/src/dataExplorer/components/Schema.scss
+++ b/src/dataExplorer/components/Schema.scss
@@ -85,7 +85,7 @@
 
     &:hover {
       cursor: pointer;
-      color: #7ce490; // color suggested in https://github.com/influxdata/ui/issues/4905
+      color: $c-krypton;
       background: $cf-grey-15;
       border-radius: 2px;
 
--- a/src/dataExplorer/components/Schema.scss
+++ b/src/dataExplorer/components/Schema.scss
@@ -85,7 +85,7 @@
 
     &:hover {
       cursor: pointer;
-      color: #7CE490; // color suggested in https://github.com/influxdata/ui/issues/4905
+      color: #7ce490; // color suggested in https://github.com/influxdata/ui/issues/4905
       background: $cf-grey-15;
       border-radius: 2px;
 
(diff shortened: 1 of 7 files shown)


src/dataExplorer/components/Schema.scss
```diff
@@ -62,39 +62,45 @@
...
+    width: fit-content;
+    font-family: $cf-code-font;
+    font-weight: 400;
+    background-color: inherit;
+    color: $cf-white;
+    transition-property: color;
+
+    &:hover {
+      cursor: pointer;
+      color: #7ce490; // color suggested in https://github.com/influxdata/ui/issues/4905
```
drdelambre, line 88: why are we using a color not defined in clockface and not changing the color in monaco to match our theming instead? all for the standardization, just seems like this is the wrong direction of that equation?

src/shared/components/FilterList/MinimalistInjectionOption.scss
```diff
@@ -8,7 +8,7 @@
   .cf-text-block {
     width: fit-content;
     font-size: $cf-form-sm-font;
-    font-family: 'Roboto Mono';
+    font-family: $cf-code-font;
```
drdelambre, line 11: nice

src/dataExplorer/components/Schema.scss
appletreeisyellow: @drdelambre It's a good point! I thought of it and not sure if it's too big of a change if I change the color setting in Monaco config for entire UI. I'd prefer to change the color in Monaco instead of the other way around to avoid too many hard coded hex code in our code base. What do you think @juliajames12?

Here are the comparison of colors FYI

## Using suggested color #7ce490
Screen Shot 2022-06-28 at 10 37 37 AM

## Using clockface color #9bf445 ($c-krypton)
Screen Shot 2022-06-28 at 10 36 54 AM

juliajames12:
@appletreeisyellow @drdelambre  It's a fair point and to make matters even more complicated, we're actually rehauling the colors to match the rebranding from marketing (I believe we'll have that transitioned within the next month). Knowing we'll be doing that soon, for now we can keep the clockface colors you had and then when we transition to the new brand colors, we'll update it in both clockface and monaco. There will just be a few weeks or so where they don't match as intended. Does that work?

drdelambre (approved):

PR #4917, opened 2022-06-28
fix: force recalc of child component size

### LINKED ISSUE #4904: flux query builder: fix resizing
right now the monaco editor isn't playing nice in a structure of flex boxes with relative dimensions. the editor seems to only be able to grow in size, leading to a frustrating experience when resizing the panels.

you should be able to use the resize handles to set the flux editor to an arbitrary size.

drdelambre:
Closes #4904

the flux display property has some cool abilities to resize it's children based on their content, but when a child gets a fixed dimension (width: px or height: px) then it prevents the flex parent from shrinking to consume the extra space. this caused our resizing component to collide with our monaco component creating a condition where the editor could only become bigger. at first this just looked like bizarre resizing behavior, but soon induced a state where the editor was the only component on the screen.

turns out the logic was fine for the resizer as long as there weren't these fixed children, we just needed a way to force the child component to be decoupled from the sizing calculations being made for the parent component during rending the IE box model hierarchy. the technique used ended up being having a wrapper component that set up a condition between a flex child (the parent in this situation) which gets resized automatically by it's parent to fill the space desired with the flex-grow property and a position: relative to anchor a new coordinate space for the child in relation to the parent. the wrapper then sets that new coordinate space with a `position: absolute', and sizes that space to fill the parent container with width: 100%, height: 100%

css errors suck

### DIFF
--- a/src/dataExplorer/components/Results.scss
+++ b/src/dataExplorer/components/Results.scss
@@ -4,6 +4,8 @@ $cf-radius-lg: $cf-radius + 4px;
 
 .data-explorer-results {
   height: 100%;
+  width: 100%;
+  position: absolute;
   border-radius: $cf-radius-lg;
   border: 2px solid transparent;
   background-color: $cf-grey-15;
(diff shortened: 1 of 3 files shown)


src/dataExplorer/components/Results.scss
```diff
@@ -39,6 +41,18 @@ $cf-radius-lg: $cf-radius + 4px;
...
+.data-explorer--monaco-outer {
+  flex-grow: 1;
+  position: relative;
+  width: 100%;
+}
+
+.data-explorer--monaco-wrap {
+  position: absolute;
+  width: 100%;
+  height: 100%;
```
appletreeisyellow, line 53: Nice!

PR #4922, opened 2022-06-28
fix(AIM): Fix operator search field to keep cursor location

alejom99:
When using the search box in the operator page, the cursor jumps to the end of the line whenever the user tries to edit the text. See the before video. This PR fixes that by decoupling the component state from the provider's, which keeps he cursor in place. See the after video.

Before:

https://user-images.githubusercontent.com/4347903/176316008-23a50fba-2556-4825-9f89-e1fe56e02880.mov

After:

https://user-images.githubusercontent.com/4347903/176316073-b1e66c76-a90a-4215-97dc-9e7ad493306c.mov

### DIFF
--- a/cypress/e2e/cloud/operatorRO.test.ts
+++ b/cypress/e2e/cloud/operatorRO.test.ts
@@ -87,7 +87,7 @@ describe('Operator Page', () => {
     cy.getByTestID('orgTab').click()
     cy.getByTestID('operator-resource--searchbar').type('678', {
       force: true,
-      delay: 50,
+      delay: 300,
     })
 
     cy.getByTestID('table-body').within(() => {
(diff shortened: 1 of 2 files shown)


cypress/e2e/cloud/operatorRO.test.ts
```diff
@@ -87,7 +87,7 @@ describe('Operator Page', () => {
     cy.getByTestID('orgTab').click()
     cy.getByTestID('operator-resource--searchbar').type('678', {
       force: true,
-      delay: 50,
+      delay: 300,
```
TCL735, line 90: So, I think both the original test and this update are wrong. The test should pass without needing to purposely delay typing. What if I am fast typist?

While, I am not asking for changes here, I do think this could be a source of flakiness in the future.

alejom99: The original code had a 50 ms timeout in the debounce call and it was changed to 300, that's why. 50 ms is too fast for a debounce. If you're gonna have it that low, then there's no point on having it at all.

TCL735: Yes, agreed. I mentioned that the original test was wrong.

I understand why the change was made here--to attempt to match what was happening in the original test.

TCL735 (approved):

PR #4924, opened 2022-06-29
feat: add search

### LINKED ISSUE #4687: Basic Results View
## User Story

<p></p>

As a user, when I do a query, I want to see at least some of the data that resulted in that query so that I can visualization, troubleshoot my script, or simply inspect the data.

<p></p>

## Acceptance Criteria

<p></p>

- The results view will have a columnar format, much like the existing DE or Notebook table view

- ~~Sortable Columns~~

  - ~~The column headers will have a function to ascend or descend sort - the user can click on that option in the result window~~

  - ~~When user chooses to sort by one of the columns, the appropriate line of flux will be added to the flux script, and the query will be re-ran~~

- Search/Filter of Results

  - The user will have an option to search/filter the results of the query

  - When selected, this will simply filter the lines that appear in the results view to only include the lines of the result that have text that is in the search box

  - Highly desirable to have type ahead filtering and not required to type in a whole string and click search; but not an absolute requirement to pass A/C

  - This filtering is only of the result set; the query will not have to be re-ran

  - Display of X number of X lines, after the filter (discuss with Julia for design)

- Switch to Visualization

  - There is an option in the results view to switch to a visualization view

  - The visualization view will re-use the existing DE or Notebooks visualizations; however (note:  table or simple table view may need to be suppressed - up for discussion); it is not intended that we do any other retrofitting of the visualizations as part of this effort

<p></p>

## Notes

<p></p>

- Sortable columns will have a dependency on the flux injection work; so it may be dropped for MVP if a solution is not determined for this

- Actionable Results will come in a later EPIC/Stories - however, the UI design should have this in mind for retrofitting for the future

<p></p>

<p></p>

drdelambre:
Closes #4687 

adds the ability to search across rows of data in the results

Peek 2022-06-29 11-02

### DIFF
--- a/src/dataExplorer/components/Results.tsx
+++ b/src/dataExplorer/components/Results.tsx
@@ -1,4 +1,4 @@
-import React, {FC, useState, useContext} from 'react'
+import React, {FC, useState, useContext, useMemo} from 'react'
 import {
   FlexBox,
   FlexDirection,
@@ -21,6 +21,7 @@ import {
   ViewOptions,
   SUPPORTED_VISUALIZATIONS,
 } from 'src/visualization'
+import {FluxResult} from 'src/types/flows'
 
 import './Results.scss'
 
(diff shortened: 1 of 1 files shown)


src/dataExplorer/components/Results.tsx
```diff
@@ -82,6 +83,55 @@ const Results: FC = () => {
...
+          },
+          {}
+        ),
+      },
+    }
+
+    const len = result.parsed.table.length
+    const keys = Object.keys(result.parsed.table.columns)
+    let newLen = 0,
+      ni = 0
```
appletreeisyellow, line 109: Just to check if I understand it correctly:

`newLen` is to keep track of the row number in `newCols`?
`ni` is to keep track of the row number in `oldCols`?

appletreeisyellow (approved):

src/dataExplorer/components/Results.tsx
appletreeisyellow: > like spliting out a reference to the result.parsed.table.columns object into the oldCols variable outside of the for loop as all of those object member accessors (the .) actually have a noticeable performance hit 

Thanks for explaining it. Good to know!

PR #4931, opened 2022-06-29
chore: add checkmark  based on completeness of the subway nav sections

subirjolly:
Closes https://github.com/influxdata/data-acquisition/issues/399

We currently don't have the ability to have checkbox based on a custom criteria. Instead of adding a criteria function, I am adding an optional isCompleted boolean per step. If the boolean exists for a step then that boolean will take precedence. If it doesn't exist in the navigation step, then the regular logic will be applied.

## BEFORE
BEFORE

## AFTER
AFTER

### Checklist


helenosheaa (commented):
looking good! can you add a screen recording of the behavior for the edit and details view? As I think for the details view we don't want checkmarks.

you also may need to add an empty commit to get your ci to run again it failed on remocal-deploy.

src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
```diff
@@ -80,6 +99,60 @@ const CreateSubscriptionPage: FC = () => {
...
+        formContent.bucket !== '<BUCKET>'
+          ? 'true'
+          : 'false',
+      parsingStepCompleted:
+        formContent.dataFormat &&
+        checkRequiredJsonFields(formContent) &&
+        checkRequiredStringFields(formContent)
+          ? 'true'
+          : 'false',
+      dataFormat: formContent.dataFormat ?? 'not chosen yet',
```
genehynson, line 148: thoughts on moving this business logic to the `utils/form.ts` file? I think we could add functions like `isBrokerStepCompleted()` and just invoke them here. That way all the logic around form validation is in the same file.

subirjolly: i was thinking of moving this to context with the refactoring. but this is a good idea. let me move it.

subirjolly:
@helenosheaa Thoughts on this?

AFTER

helenosheaa (approved):
lgtm!

genehynson (approved):

PR #4932, opened 2022-06-30
feat: Convert to Contract button in operator UI

eatondustin1:
This adds a Convert to Contract button to the operator UI's account page. This button is only visible to read-write operators and is only enabled for free accounts. When clicked, an overlay is shown that allows the user to select a contract start date and confirm the convert to account action.

The setVisible and visible props for `DeleteAccountOverlay` were refactored to be less generic so they could be distinguished from the props used to show `ConvertAccountToContractOverlay`.

There were some auto-fixes from eslint and prettier as well.

This also adds a cypress test for the Convert to Contract button that also tests whether the Delete Account button is enabled when the account is deleteable.

Finally, the datepicker was modeled after the one in the `TimeTickInput`. Because of this, there is an option to select the time as well. Ideally we can update clockface to include a working datepicker without time selection, but until then, this is probably the best we can do. Because the time is not relevant for this component though, I've set it up to not actually do anything when a time is selected. Because of this, the only way to have a time other than 00:00:00 is to enter it manually in the input in the convert to contract overlay.

image

### Checklist

### DIFF
--- a/cypress/e2e/cloud/operator.test.ts
+++ b/cypress/e2e/cloud/operator.test.ts
@@ -9,7 +9,7 @@ describe('Operator Page', () => {
           }).then(() => {
             // This call initializes the operator data for quartz mock so the data is reset before each test.
             // This allows tests to mutate the quartz-mock operator data without breaking subsequent tests.
-            cy.initializeQuartzMockOperator()
+            // cy.initializeQuartzMockOperator()
             cy.quartzProvision({
               isOperator: true,
               operatorRole: 'read-write',
--- a/cypress/e2e/cloud/operator.test.ts
+++ b/cypress/e2e/cloud/operator.test.ts
@@ -7,9 +7,6 @@ describe('Operator Page', () => {
           cy.setFeatureFlags({
             uiUnificationFlag: true,
           }).then(() => {
-            // This call initializes the operator data for quartz mock so the data is reset before each test.
-            // This allows tests to mutate the quartz-mock operator data without breaking subsequent tests.
-            cy.initializeQuartzMockOperator()
             cy.quartzProvision({
               isOperator: true,
               operatorRole: 'read-write',
@@ -205,17 +202,9 @@ describe('Operator Page', () => {
     // should be able to delete deletable accounts
     cy.getByTestID('account-delete--button').should('not.be.disabled')
     // should be able to convert free accounts to contract
-    cy.getByTestID('account-type--body').contains('free')
     cy.getByTestID('account-convert-to-contract--button').should(
       'not.be.disabled'
     )
-    cy.getByTestID('account-convert-to-contract--button').click()
-    cy.intercept('GET', '/api/v2/quartz/operator/accounts/678').as(
-      'quartzReloadAccount'
-    )
-    cy.getByTestID('convert-account-to-contract--confirmation-button').click()
-    cy.wait('@quartzReloadAccount')
-    cy.getByTestID('account-type--body').contains('contract')
 
     cy.getByTestID('account-view--back-button').click()
 
--- a/cypress/e2e/cloud/operator.test.ts
+++ b/cypress/e2e/cloud/operator.test.ts
@@ -7,6 +7,7 @@ describe('Operator Page', () => {
           cy.setFeatureFlags({
             uiUnificationFlag: true,
           }).then(() => {
+            cy.initializeQuartzMockOperator()
             cy.quartzProvision({
               isOperator: true,
               operatorRole: 'read-write',
@@ -145,6 +146,8 @@ describe('Operator Page', () => {
     cy.getByTestID('account-view--header').contains('operator1 (1)')
     // should not be able to delete undeletable accounts
     cy.getByTestID('account-delete--button').should('be.disabled')
+    // should not be able to convert non-free accounts to contract
+    cy.getByTestID('account-convert-to-contract--button').should('be.disabled')
 
     // Associated Users Section
     cy.getByTestID('associated-users--title').contains('Associated Users')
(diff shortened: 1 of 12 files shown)


cypress/e2e/cloud/operator.test.ts
```diff
@@ -7,6 +7,7 @@ describe('Operator Page', () => {
           cy.setFeatureFlags({
             uiUnificationFlag: true,
           }).then(() => {
+            cy.initializeQuartzMockOperator()
```
wdoconnell, line 10: I might add a comment in this test as well to flag that `quartzMockOperator` is using an endpoint that doesn't exist for the UI in prod.

src/operator/account/ConvertAccountToContractOverlay.tsx
```diff
@@ -0,0 +1,266 @@
...
+  ComponentStatus,
+  Columns,
+  Grid,
+  Input,
+} from '@influxdata/clockface'
+import DatePicker from 'src/shared/components/dateRangePicker/DatePicker'
+import {AccountContext} from 'src/operator/context/account'
+import moment from 'moment'
+
+const noOp = () => {}
```
wdoconnell, line 36: Do we need some user notification or feedback here - I see noOp being invoked only when there's an invalid user input. Should we be sending a notification (e.g. `dispatch(notify([notification]))) when that happens?

src/operator/account/ConvertAccountToContractOverlay.tsx
```diff
@@ -0,0 +1,266 @@
...
+  const onClickOutside = () => {
+    if (isOnClickOutsideHandlerActive) {
+      hideDatePicker()
+    }
+  }
+
+  const allowOnClickOutside = () => setIsOnClickOutsideHandlerActive(true)
+  const suppressOnClickOutside = () => setIsOnClickOutsideHandlerActive(false)
+
+  const styles: CSSProperties = isDatePickerOpen
```
wdoconnell, line 137: Not that much css here, but I wonder whether this is more readable if we do something like `className={x ? x : y}`, and make the styles live in a separate .scss file.

src/operator/account/ConvertAccountToContractOverlay.tsx
eatondustin1: We do actually have some user feedback because we set `ComponentStatus` to `Error`, but I do agree that we should have an error reason. However, `dispatch(notify([notification]))` will only show a temporary error, which isn't great for form-related errors from a UX perspective, since it's easy for the user to miss the error and it leaves them without context. I'm okay with doing that if it's standard practice for us, but I think it might be worth reviewing Material Design for how they recommend error handling because there's some good information there. My proposal would be to include an error below our form fields similar to the ones in the Text Field reference below so that the error exists in a useful context within the UI and is persistent until the error is resolved. Ideally this would be built into the `DatePicker` component, but I don't mind adding it in for this component specifically in the meantime. What are your thoughts on that?

Material Design links:
[Text Fields](https://material.io/components/text-fields#anatomy)
[Snackbars](https://material.io/components/snackbars#usage)
[Error Patterns (Archived)](https://material.io/archive/guidelines/patterns/errors.html)

src/operator/account/ConvertAccountToContractOverlay.tsx
eatondustin1: I'm going to open an issue/feature request in [clockface](https://github.com/influxdata/clockface) around the datepicker, which will suggest a number of improvements to the built-in datepicker that I'm hoping will allow most of this code to go away. The main reason I kept a number of these things the way they are is to keep it more consistent with `TimeTickInput`, which will allow for easier refactoring with a new and improved `DatePicker` component that ideally eliminates the need for so much of this extra code, like error handling, manually toggling the datepicker overlay, and these styles.

But realistically, even if I end up spending some time working on clockface's datepicker to get a PR open, it likely won't be done for a little while at best, so I'm okay with whatever aligns this code with our ui development patterns, so if that means adding a separate styles file, we can do that. I definitely prefer that approach personally.

cypress/e2e/cloud/operator.test.ts
eatondustin1: Good thinking, I'll add the comment!

wdoconnell:
Note for standup: this won't pass `e2e` tests until [quartz-mock #24](https://github.com/influxdata/quartz-mock/pull/24/files) is resolved, as that adds mock backend functionality for the new convert-to-contract button.

src/operator/account/ConvertAccountToContractOverlay.tsx
wdoconnell: It's not a lot of css, so I think it looks good as is without moving into a scss file. Keeping this as is for consistency with `TimeTickInput` makes a lot of sense to me.

src/operator/context/account.tsx
```diff
@@ -87,6 +102,28 @@ export const AccountProvider: FC<Props> = React.memo(({children}) => {
...
   }, [handleGetAccount, accountID])
 
+  const handleConvertAccountToContract = useCallback(
+    async contractStartDate => {
+      try {
+        setConvertStatus(RemoteDataState.Loading)
+        const resp = await patchOperatorAccountsConvert({
+          accountId: accountID,
+          data: {contractStartDate},
+        })
```
hoorayimhelping, line 112: @wdoconnell I think we should consider moving this into an API layer and out of the context layer. What do you think? Should it happen in this PR (which is already doing a lot of work and has been in progress for over a week) or is that something we should follow up on?

hoorayimhelping (changes requested):

eatondustin1:
@hoorayimhelping I think I resolved all of the issues you called out, let me know if you see any more, and thanks!

src/operator/account/AccountViewHeader.tsx
```diff
@@ -33,11 +39,28 @@ const AccountViewHeader: FC = () => {
           Back to Account List
         </Link>
       </FlexBox.Child>
+      {hasWritePermissions && (
+        <ButtonBase
+          color={ComponentColor.Primary}
+          shape={ButtonShape.Default}
+          onClick={_e =>
```
hoorayimhelping, line 46:
```suggestion
          onClick={() =>
```

src/operator/account/AccountViewHeader.tsx
```diff
@@ -33,11 +39,28 @@ const AccountViewHeader: FC = () => {
...
+        >
+          Convert to Contract
+        </ButtonBase>
+      )}
       {hasWritePermissions && (
         <ButtonBase
           color={ComponentColor.Danger}
           shape={ButtonShape.Default}
-          onClick={_e => setVisible(!visible)}
+          onClick={_e => setDeleteOverlayVisible(!deleteOverlayVisible)}
```
hoorayimhelping, line 63:
```suggestion
          onClick={() => setDeleteOverlayVisible(!deleteOverlayVisible)}
```

hoorayimhelping (approved):
LGTM, just a couple of minor nits about not using unused events

PR #4933, opened 2022-06-30
fix: subscriptions header

helenosheaa:
Closes https://github.com/influxdata/data-acquisition/issues/398

Lowers z-index so errs are shown above the fixed header.
Fixes width to account for whether the side nav is toggled or not.
Adds padding so we aren't losing the header / text at the top of the page.

https://user-images.githubusercontent.com/38860767/176711677-d950a466-c021-4af3-9df6-529a581dc274.mov

Screen Shot 2022-06-30 at 10 56 05 AM

Screen Shot 2022-06-30 at 10 57 31 AM

### Checklist

### DIFF
--- a/src/writeData/subscriptions/components/BrokerDetails.tsx
+++ b/src/writeData/subscriptions/components/BrokerDetails.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FC} from 'react'
+import React, {FC, useContext} from 'react'
 import {useHistory} from 'react-router-dom'
 import {useSelector} from 'react-redux'
 
@@ -27,6 +27,7 @@ import BrokerFormContent from 'src/writeData/subscriptions/components/BrokerForm
 // Utils
 import {getOrg} from 'src/organizations/selectors'
 import {event} from 'src/cloud/utils/reporting'
+import {AppSettingContext} from 'src/shared/contexts/app'
 
 // Types
 import {SUBSCRIPTIONS, LOAD_DATA} from 'src/shared/constants/routes'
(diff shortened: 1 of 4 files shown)


src/writeData/subscriptions/components/BrokerDetails.tsx
```diff
@@ -56,11 +57,18 @@ const BrokerDetails: FC<Props> = ({
...
+  const navbarOpen = navbarMode === 'expanded'
   return (
     <div className="update-broker-form" id="broker">
       <SpinnerContainer spinnerComponent={<TechnoSpinner />} loading={loading}>
         <Form onSubmit={() => {}} testID="update-broker-form-overlay">
-          <div className="update-broker-form__fixed">
+          <div
+            className="update-broker-form__fixed"
+            style={{
+              width: navbarOpen ? '61%' : '69%',
```
subirjolly, line 69: I was wondering why don't we make the body portion a flexbox so that the width is automatically handled?

src/writeData/subscriptions/components/BrokerForm.tsx
```diff
@@ -61,11 +62,18 @@ const BrokerForm: FC<Props> = ({formContent, updateForm, saveForm}) => {
...
   const isCredit250ExperienceActive = useSelector(shouldGetCredit250Experience)
   const requiredFields = checkRequiredFields(formContent)
+  const {navbarMode} = useContext(AppSettingContext)
+  const navbarOpen = navbarMode === 'expanded'
   return (
     formContent && (
       <div className="create-broker-form" id="broker">
         <Form onSubmit={() => {}} testID="create-broker-form-overlay">
-          <div className="create-broker-form__fixed">
+          <div
```
subirjolly, line 71: Same as above. Why not use flexbox?

src/writeData/subscriptions/components/BrokerDetails.tsx
helenosheaa: I've tried a few different ways but with the position fixed this was the only way I could get it to line up properly to the form. Open to suggestions though if you wanna jump on a zoom!

subirjolly (approved):

src/writeData/subscriptions/components/BrokerDetails.tsx
helenosheaa: As discussed in zoom the page would need a refactor to implement flex here so we are parking it for now as this is fixing the immediate bugs :)

PR #4935, opened 2022-06-30
fix: stopgap error messages

drdelambre:
we gotta show people error messages in the ui when their query fails

here's a quick stab while we wait for design to get a few free cycles to decouple the presentation from the layout shifting. @juliajames12 

Peek 2022-06-30 09-39

### DIFF
--- a/src/dataExplorer/components/QueryTime.scss
+++ b/src/dataExplorer/components/QueryTime.scss
@@ -2,6 +2,7 @@
 
 .query-status {
   flex-grow: 1;
+  margin-top: 14px;
 
   .status {
     width: 12px;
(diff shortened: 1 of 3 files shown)


src/dataExplorer/components/QueryTime.scss
```diff
@@ -2,6 +2,7 @@
 
 .query-status {
   flex-grow: 1;
+  margin-top: 14px;
```
drdelambre, line 5: hoping to remove this soon to be more inline with clockface styling, but had to shoehorn it in

aanthony1243 (commented):
looks good, thanks for jumping on this

PR #4937, opened 2022-06-30
test: fix flaky tests for schema browser

### LINKED ISSUE #4907: Flaky test in Flux Query Builder
The cypress test for Flux Query Builder (`fluxQueryBuilder.test.ts`) is flaky. The tests passes on local machines, but fail on CI. 

It may related to local storage and/or cypress parallelization.

Here is one of the failed monitor CI tests:
https://app.circleci.com/pipelines/github/influxdata/monitor-ci/12801

appletreeisyellow:
Closes #4907 

This PR enhanced the tests for schema browser in flux query builder in following areas:
- Add conditional check on feature flag and toggle to switch to flux query builder page
- Use test IDs (`getByTestID`) instead of class names (`get`) to select DOM elements
- Add `wait` on API calls
- Add support for testing debouncer

### Checklist


cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -11,10 +11,17 @@ describe('FluxQueryBuilder', () => {
...
-              .click()
+            cy.getByTestID('flux-query-builder-toggle').then($toggle => {
+              cy.wrap($toggle).should('be.visible')
+              if ($toggle.hasClass('active')) {
+                // hasClass is a jQuery function
+                // Already switched to Flux Query Builder
+                // do nothing
+              } else {
+                // Switch to Flux Query Builder
+                $toggle.click()
```
TCL735, line 22: This is a strange way to write an `if` statement. You can keep the comments and write a different `if` statement.

TCL735 (changes requested):

src/shared/contexts/editor/index.tsx
```diff
@@ -65,7 +65,11 @@ export const EditorProvider: FC = ({children}) => {
 
   const injectViaLsp = useCallback(
     (cmd, data: Omit<ExecuteCommandArgument, 'textDocument'>) => {
-      connection.current.inject(cmd, data)
+      try {
+        connection.current.inject(cmd, data)
+      } catch (e) {
+        console.error(e)
+      }
```
appletreeisyellow, line 72: The monitor-ci-test kept failing saying `Cannot read properties of null (reading 'current')`. Here is an example of a failed test https://app.circleci.com/pipelines/github/influxdata/monitor-ci/12979

My guess why it fails is that cypress runs test commands so quick that before the monaco editor is mounted, `injectViaLsp` is called by selecting a measurement. Adding an assertion on monaco editor (with test ID `flux-editor`) as well as a try catch logic here resolve the problem

wiedld (commented):
Just a general comment. Nothing blocking this PR.

Once we make changes outside of the cypress folder, then the CI pipeline runs the normal `build` test workflow. Meaning, it does not do the parallel 20x firefox tests and 20x chrome tests.

These changes may make the tests run less flaky. But it does not mean that any future changes to just this test file only (which runs the 20x and 20x again) will not hit issues again.

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -80,76 +87,106 @@ describe('FluxQueryBuilder', () => {
...
-    it.skip('search bar can search fields and tag keys dynamically', () => {
+    it('search bar can search fields and tag keys dynamically', () => {
+      cy.clock(new Date(), ['Date']) // needed by .tick()
+      cy.intercept('POST', '/api/v2/query*').as('query')
+
       // select a bucket
       cy.getByTestID('bucket-selector--dropdown-button').click()
-      cy.getByTestID(`searchable-dropdown--item ${bucketName}`).click()
+      cy.getByTestID(`bucket-selector--dropdown--${bucketName}`).click()
+      cy.wait('@query')
```
drdelambre, line 97: for all of these waits, we should be watching for the ui to change somehow so that we're decoupling intent from implementation. by leveraging the intercepts, we make the tests more fragile by removing the ability for schema look ups to be done by another rest api, or pulling from a local cache for example. is there any visual dom changes we can be looking at, like a dropdown being populated or a list element showing up?

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -80,76 +87,106 @@ describe('FluxQueryBuilder', () => {
...
       cy.getByTestID('measurement-selector--dropdown-button').click()
       cy.getByTestID(`searchable-dropdown--item ${measurement}`).click()
+      cy.wait('@query')
 
       // search a feild, should contain only the feild, no tag keys
       cy.getByTestID('field-tag-key-search-bar').type(searchField)
-      cy.get('.field-selector--list-item--selectable').should(
+      cy.wait('@query')
+
+      cy.getByTestID('field-selector--list-item--selectable').should(
```
drdelambre, line 111: nice

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -80,76 +87,106 @@ describe('FluxQueryBuilder', () => {
...
       cy.getByTestID('dismiss-button').click()
 
-      // search a tag key, should contain only that tag key, no fields
-      cy.getByTestID('field-tag-key-search-bar').type(searchTagKey)
-      cy.get('.field-selector--list-item').should('contain', 'No Fields Found')
-      cy.get('.tag-selector-key--list-item').should('contain', searchTagKey)
+      // search a tag key, should not contain any fields
+      cy.getByTestID('field-tag-key-search-bar')
+        .type(searchTagKey)
+        .tick(600)
```
drdelambre, line 122: how did you chose this number?

appletreeisyellow: The time out for debouncer is `500` from [here](https://github.com/influxdata/ui/blob/master/src/dataExplorer/context/fluxQueryBuilder.tsx#L37), so I choose `600` to make sure the debouncer has time to react. Is it necessary to choose a smaller number, like `501` to shorten the test run time?

cypress/e2e/shared/fluxQueryBuilder.test.ts
appletreeisyellow: @drdelambre do you mean, instead of waiting on intercepts, we use assertions on visual dom for the changes?

cypress/e2e/shared/fluxQueryBuilder.test.ts
drdelambre: it more of a general suspicion of all consts and timers in tests.

consts like this mean that if we change that debouncer time in the code you linked to a year from now and forget about this test, the tests either get flakey or start taking longer than they need to. death by a thousand cuts if we dont keep an eye on it.

timers are normally a red flag for me as they're a sign that we aren't testing the ux's reaction to internal state changes correctly (by looking at changes a user can see) which leads to flakiness. this manifests itself in all sorts of ways, but the most notorious in a CI environment where we're always pushing to run these tests on cheapest nodes we can lease, is that increased load over time introduces variability into our app's responsiveness on async tasks (they take longer to run, but time stays the same), doubly so if we have some noisy neighbors on that cluster outside of our control.

sometimes you gotta use timers, but we should be using them sparingly and seeing if we can instead test the outcome of that debounced action

appletreeisyellow: thanks for the explanation! i may find a way to test the result of the search bar then

drdelambre (approved):
great work on increasing the specificity, and keeping the tests looking at the dom! :crossed_fingers:

PR #4938, opened 2022-06-30
WIP: test: guard with attach assertion rather than visible

### LINKED ISSUE #4936: Flaky test: flowsSidebarExportToClipboard.test.ts
Investigate: flowsSidebarExportToClipboard appears to be flaky sometimes. The problem happens when Cypress clicks on the monaco editor to type. Sometimes there will be a test failure for reason `DOM element is detached` when Cypress tries to type inside the `textarea`

TCL735:
Closes #4936  

Q: Run the tests. Is the flakiness confirmed?
A: Yes, unfortunately we reached flakiness at **119/120** test runs with the existing code

Proposal: since the flakiness is due to interacted elements being detached, let's assert to ensure our element is attached before interacting with it. We can also try going back to typing into the `.view-line`


TCL735:
monitor-ci pipeline 12999

120/120 ✅

PR #794, opened 2021-03-08
fix: annotation toggle button message fixed

ChitlangeSahas:
Closes #793 

The Annotation toggle button on-hover message was wrong as described in #793. This PR fixes it. 

Screen shots:

Screen Shot 2021-03-08 at 7 13 25 AM

Screen Shot 2021-03-08 at 7 13 36 AM

- [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [x] Rebased/mergeable
- [x] Tests pass

PR #4939, opened 2022-06-30
feat: multiorg Global Header

ChitlangeSahas:
Closes #

### Checklist


yarn.lock
```diff
@@ -10027,6 +10044,14 @@ react-virtualized@^9.18.5:
     prop-types "^15.7.2"
     react-lifecycles-compat "^3.0.4"
 
+react-window@^1.8.7:
```
ChitlangeSahas, line 10047: we can delete this from the UI

yarn.lock
```diff
@@ -2075,6 +2085,13 @@
     "@types/prop-types" "*"
     "@types/react" "*"
 
+"@types/react-window@^1.8.5":
```
ChitlangeSahas, line 2088: this can go as well. 👋

yarn.lock
ChitlangeSahas: looks like it's not in package json, i'm not sure why clockface dependency is being added to UI's yarn lock.

package.json
```diff
@@ -164,7 +164,7 @@
   "dependencies": {
     "@codingame/monaco-jsonrpc": "^0.3.1",
     "@docsearch/react": "^3.0.0-alpha.37",
-    "@influxdata/clockface": "^4.6.0",
+    "@influxdata/clockface": "^4.7.3",
```
wdoconnell, line 167: Before this is merged in, we should also get [clockface #794](https://github.com/influxdata/clockface/pull/794) merged in, so that the current item always shows first in the search bar.

src/App.tsx
```diff
@@ -87,10 +90,15 @@ const App: FC = () => {
       <EngagementLink />
       <TreeNav />
       <Suspense fallback={<PageSpinner />}>
-        <Switch>
-          <Route path="/orgs/new" component={CreateOrgOverlay} />
-          <Route path="/orgs/:orgID" component={SetOrg} />
-        </Switch>
+        <div style={{height: '100%', width: '100%'}}>
```
wdoconnell, line 93: Might want to use a custom component like a Flexbox here, but all we really need is to make sure the globalheader is filling 100% of the available window, so that it's not artificially cut off - the UserProfile will be on the right side.

src/identity/components/GlobalHeader/AccountDropdown.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const accountMainMenu = [
+    {
+      name: 'Settings',
+      iconFont: IconFont.CogOutline,
+      href: `/orgs/${activeOrg.id}/accounts/settings`,
+    },
+    // 'Members' is listed on the Figma, but discussed with design, and this may not exist yet.
+    // {
+    //   name: 'Members',
```
wdoconnell, line 33: Flagging this because it's on the figma, but I don't believe the page exists presently. 

@juliajames12 - Did I get it right from our conversation that we don't need to redirect to an accounts 'Members' page at this time?

src/identity/components/GlobalHeader/DefaultEntities.tsx
```diff
@@ -0,0 +1,17 @@
+// Types
+import {UserAccount} from 'src/client/unityRoutes'
+import {OrganizationSummaries} from 'src/client/unityRoutes'
+
+export const emptyAccount: UserAccount = {
```
wdoconnell, line 5: Flagging for a code review re: what the appropriate default state should be here. I think we should set the `setState` default values to an object (or array with one object as applicable) with the same shape, that's empty with everything set to false. 

Using `null` or an empty object is kind of anti-pattern in my opinion, and it causes type issues down the line unless we add conditionals to handle those cases.

src/identity/components/GlobalHeader/GlobalHeader.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [activeOrg, setActiveOrg] = useState(emptyOrg)
+  const [activeAccount, setActiveAccount] = useState(emptyAccount)
+
+  useEffect(() => {
+    dispatch(getQuartzOrganizationsThunk())
+  }, [dispatch])
+
+  useEffect(() => {
+    if (orgsList) {
+      setActiveOrg(orgsList?.find(org => org.isActive === true))
```
wdoconnell, line 59: In a perfect world we'd have one loop here that performs both the find and the sort, but that logic would be unnecessarily complicated, and this really shouldn't impact performance. `find` also exits early, so for that loop we only need to get to the `isActive` account before the loop ends.

src/identity/components/GlobalHeader/GlobalHeaderStyle.ts
```diff
@@ -0,0 +1,12 @@
+export const globalHeaderStyle = {
```
wdoconnell, line 1: Tried using `scss` and setting className, but it looks like the classes aren't overriding in the clockface component, so reverted to a file that holds the variables for setting `style`.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -43,7 +43,7 @@ const OrgProfileTab: FC = () => {
       if (
         !me.quartzMe.regionCode ||
         !me.quartzMe.regionName ||
-        !identity.org.provider
+        !identity.currentIdentity.org.provider
```
wdoconnell, line 46: FYI - these are just updates resulting from a structural refactor to the one state in the app that currently uses `identity` directly, rather than through the conversion to the legacy `quartzMe` state.

wdoconnell:
On Tuesday, let's circle up w/ Bucky re whether he's on board with adding the random org/account generator. It's useful for demonstration purposes so that we don't have duplicate org/account names when we're testing the component for large account/org values, but maybe we don't need that many lines - that file is twice the size of the other globalheader files.

src/shared/copy/notifications/categories/accounts-users-orgs.ts
```diff
@@ -138,3 +138,9 @@ export const updateOrgFailed = (): Notification => ({
   message:
     'Error retrieving new organization information. Please refresh this page.',
 })
+
+export const updateQuartzOrganizationsFailed = (): Notification => ({
+  ...defaultErrorNotification,
+  message:
+    'Error retrieving organizations available to this account. Please refresh this page.',
```
hoorayimhelping, line 145: I think we can make this a little bit less mechanical sounding.

src/identity/components/GlobalHeader/AccountDropdown.tsx
wdoconnell: Agree. Based on the conversation with design, this link isn't needed at this time.

src/identity/utils/alphaSortSelectedFirst.ts
```diff
@@ -0,0 +1,11 @@
...
+  return [...itemArray].sort((a, b) => {
+    if (a.id === selectedItem.id) {
+      return -1
+    } else if (b.id === selectedItem.id) {
+      return 1
+    } else {
+      return a.name.toLowerCase() < b.name.toLowerCase() ? -1 : 1
+    }
+  })
+}
```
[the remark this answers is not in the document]
ChitlangeSahas: ✅

src/identity/components/GlobalHeader/AccountDropdown.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  return (
+    <MenuDropdown
+      selectedOption={selectedAccount}
+      options={accountMainMenu}
+      subMenuOptions={accountsList}
+      menuHeaderIcon={IconFont.Switch_New}
+      menuHeaderText="Switch Account"
+      style={{width: 'auto'}}
+      menuStyle={{width: '250px'}}
```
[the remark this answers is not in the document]
wdoconnell: Got it. I'll think on it, but for now, I'm thinking we use `useMemo` here. 

Maybe `useMemo` is overkill when we're only setting a property, but unfortunately we can't use css for both `style` and `menuStyle`. 

`menuStyle` gets passed as the `style` prop to an underlying `Dropdown` component in clockface, which I don't think we're otherwise able to style without edits to clockface. So rather than using `useMemo` for one and css for the other, let's use `useMemo` for both.

src/identity/components/GlobalHeader/DefaultEntities.tsx
wdoconnell: 👍 Also imported the same default empty object into the reducer.

src/identity/components/GlobalHeader/GlobalHeader.tsx
```diff
@@ -0,0 +1,103 @@
...
+import {alphaSortSelectedFirst} from 'src/identity/utils/alphaSortSelectedFirst'
+
+export const GlobalHeader: FC = () => {
+  const dispatch = useDispatch()
+
+  const identity = useSelector(selectQuartzIdentity)
+
+  // Lines can be uncommented for purposes of mocking/testing with arbitrarily large account and org lists.
+  // const [identity] = useState(randomEntityGenerator('org', 2000))
+  const orgsList = identity.quartzOrganizations?.orgs
```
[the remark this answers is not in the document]
wdoconnell: Confirmed that the `orgs` array will already exist as part of the default state in redux.

src/identity/quartzOrganizations/reducers/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  SET_QUARTZ_ORGANIZATIONS,
+  SET_QUARTZ_ORGANIZATIONS_STATUS,
+} from 'src/identity/quartzOrganizations/actions/creators'
+import produce from 'immer'
+
+import {OrganizationSummaries} from 'src/client/unityRoutes'
+
+const initialState = {
+  orgs: [
+    {id: '', name: '', isActive: false, isDefault: false},
```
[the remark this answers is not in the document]
wdoconnell: Done - works well.

src/identity/utils/alphaSortSelectedFirst.ts
```diff
@@ -0,0 +1,11 @@
+export const alphaSortSelectedFirst = (itemArray, selectedItem) => {
+  return [...itemArray].sort((a, b) => {
+    if (a.id === selectedItem.id) {
+      return -1
+    } else if (b.id === selectedItem.id) {
+      return 1
+    } else {
+      return a.name.toLowerCase() < b.name.toLowerCase() ? -1 : 1
```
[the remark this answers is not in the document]
wdoconnell: Done. 👍

src/shared/copy/notifications/categories/accounts-users-orgs.ts
wdoconnell: Done.

yarn.lock
wdoconnell: @ChitlangeSahas Looks like we can't remove it - Updated with `yarn install` and ran a `yarn why`, and it looks like it has to stay as a dependency of clockface:

Screen Shot 2022-07-07 at 11 09 54 AM

src/identity/utils/alphaSortSelectedFirst.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      {id: '23443', name: 'Sly'},
+      {id: '234642', name: 'Tuk'},
+      {id: '23453', name: 'zote'},
+    ]
+
+    const result = alphaSortSelectedFirst(inputArray, selectedItem)
+
+    expect(result).toEqual(expectedResult)
+  })
+})
```
hoorayimhelping, line 69: 🎉

hoorayimhelping (approved):
great work!

PR #4947, opened 2022-07-01
feat: nifi timestamp precision

helenosheaa:
helps https://github.com/influxdata/data-acquisition/issues/204
related https://github.com/influxdata/idpe/pull/15037

- adds the ability to provide a timestamp precision
- updates tests

supported timestamp types pulled from PutInfluxDatabaseRecord_2 which is using import java.util.concurrent.TimeUnit;
NS, MS, S, US.

two clarifications:
Currently you can set precision even when you don't provide a path/pattern to a timestamp should I make it disabled when not provided? 
And for line protocol it's always the default of NS.

Changes to Edit/Create:
Screen Shot 2022-06-30 at 2 52 56 PM

Screen Shot 2022-06-30 at 2 54 05 PM

Screen Shot 2022-07-01 at 1 09 49 PM

Details:
Screen Shot 2022-07-01 at 1 10 26 PM

Screen Shot 2022-07-01 at 3 11 45 PM

Screen Shot 2022-07-01 at 3 13 08 PM

### Checklist


src/writeData/subscriptions/utils/form.ts
```diff
@@ -82,6 +82,10 @@ export const sanitizeForm = (form: Subscription): Subscription => {
     delete form.brokerUsername
     delete form.brokerPassword
   }
+
+  if (form.dataFormat === 'lineprotocol') {
+    form.timestampPrecision = 'NS'
+  }
```
genehynson, line 88: I think we'll need to let users sending LP messages set this property too

src/writeData/subscriptions/components/JsonParsingForm.tsx
```diff
@@ -46,6 +46,17 @@ const JsonParsingForm: FC<Props> = ({formContent, updateForm, edit}) => {
   const numberType = 'Number'
   const dataTypeList = [stringType, numberType]
   const [dataTypeM, setDataTypeM] = useState(stringType)
+  const microsecondsType = 'MS'
```
drdelambre, line 49: if you're looking for tighter TS support for these kinds of patterns, i think this works:
``` Javascript
enum Precision {
  Millisecond = 'MS'
  Second = 'S'
  Microsecond = 'US'
  Nanosecond = 'NS'
}

const precisionList = Object.keys(Precision)
```

if not, please disregard

src/writeData/subscriptions/utils/form.ts
helenosheaa: gotcha, adding now :)

src/writeData/subscriptions/components/JsonParsingForm.tsx
genehynson, line 49:
```suggestion
  const millisecondsType = 'MS'
```

src/writeData/subscriptions/components/JsonParsingForm.tsx
```diff
@@ -46,6 +46,17 @@ const JsonParsingForm: FC<Props> = ({formContent, updateForm, edit}) => {
...
+  const secondsType = 'S'
+  const microseconds2Type = 'US'
```
genehynson, line 51:
```suggestion
  const microsecondsType = 'US'
```

src/writeData/subscriptions/components/LineProtocolForm.tsx
```diff
@@ -1,38 +1,79 @@
...
+  const microsecondsType = 'MS'
+  const secondsType = 'S'
+  const microseconds2Type = 'US'
+  const nanosecondsType = 'NS'
+  const precisionList = [
+    nanosecondsType,
+    microsecondsType,
+    secondsType,
+    microseconds2Type,
+  ]
```
genehynson, line 31: I see we're defining this repeatedly - could it be abstracted somewhere? Maybe stored in a type within a context?

genehynson (changes requested):
I think there's a typo in the precisions, left a comment

src/writeData/subscriptions/components/LineProtocolForm.tsx
helenosheaa: I'll pull it out!

src/writeData/subscriptions/components/JsonParsingForm.tsx
helenosheaa: opps thanks haha

genehynson:
> Currently you can set precision even when you don't provide a path/pattern to a timestamp should I make it disabled when not provided?

I don't think this is necessary. It won't change the nifi processor's behavior at all to have this value set different when no timestmap is provided. And I think some people might be inclined to set the precisions first before typing in their jsonpath/regex on how to find it. If it was disabled, it could be more confusing.

genehynson:
@helenosheaa could you add a screenshot of what this looks like in the "read only" state in the update screen?

helenosheaa:
yes added more screenshots and pulled out the enum type. also realized I didn't need the setState as I can rely on the form funcs for managing the dropdown.

genehynson (approved):
Awesome!

Issue #4952, opened 2022-07-01
Add ability to subscribe to annotation stream[s] for an entire dashboard


glinton:
**Proposal:**
Add the ability to "subscribe" to annotation stream[s] for the entire dashboard

**Definition:**
Annotations are notes on a point in or range of time on a graph, that aid in the ability to draw connections between data (assuming that note can be shown on multiple graphs). People and dashboards have different concerns, so annotations were given the ability to be constrained to what is referred to as a "stream". Streams allow for an easily query-able way to get related annotations. One example of a stream is an incident stream. 

**Current Behavior:**
Currently, any time an incident occurs, an annotation can be created in the "incidents" stream. What can not happen today is to overlay that stream across any dashboard of one's choosing (cell granularity may be the technical solution, but dashboards typically are grouped logically and only giving the option to show a stream or not across an entire dashboard would suffice ([and gain parity with chronograf v1 annotations](https://docs.influxdata.com/chronograf/v1.9/guides/annotations/) (although the time scale is different across the cells in the image, the annotation is the same))).

Annotation's, in my opinion, are not functionally valuable in their current form since the UI restricts them to being related to only on an individual dashboard "cell" (or graph). These cannot be shared across dashboard cells, so even if you have the perfect dashboard to go to during an incident, you would have to create an annotation on each of the dashboard cells to be able to correlate data anomalies with the time of that incident. You also cannot have visit another dashboard, enable the "incident" stream, and view other correlations, symptoms, or side effects related to the annotated incident.

**Desired Behavior:**
 - ability to "subscribe" to annotation stream[s] for the entire dashboard
 - ability to specify which stream an annotation belongs to in the ui (currently the only way to do this is via the api)

**Use case:**
Migrating to cloud (porting annotations from cloud 1/chronograf)
Usability/data correlation (especially during incident response)

PR #4953, opened 2022-07-01
feat(annotations): create and use more broad scoped annotation

glinton:
This changes the annotation stream used from an individual cell-id stream to a dashboard wide-stream.

This is in no way release-ready, cells will still have to request annotations from streams named with the cell's id for backwards compatibility. It would also be more ideal to specify which stream to show annotations from (inclusive list) so they can be shared across dashboards. This is, imo, better than the current implementation where an annotation is restricted to a single cell.

A non-react developer's attempt at #4952 to get the ball finally rolling on it.

image

### Checklist

### DIFF
--- a/src/annotations/components/AddAnnotationOverlay.tsx
+++ b/src/annotations/components/AddAnnotationOverlay.tsx
@@ -10,8 +10,6 @@ import {OverlayContext} from 'src/overlays/components/OverlayController'
 
 // Selectors
 import {getOverlayParams} from 'src/overlays/selectors'
-import {getTimeFormatForView} from 'src/views/selectors'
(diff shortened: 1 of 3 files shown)


wiedld:
What's the status on this draft PR @glinton? 
It needs to be rebased and the tests debugged -- wanted to check if this is still under progress?

glinton:
@wiedld I very much would love to see this progress, however I won't be pushing it forward anymore. This PR was more a proof of concept to show a (seemingly) simple way to get the annotations functionality in the UI closer to using the full features available in the api. (And more inline with how grafana's or chronograf's annotation functionality worked).

wdoconnell:
Closing as unused

PR #4962, opened 2022-07-05
chore(4646/4836): helper function to calculate a windowPeriod from a duration.

wiedld:
Precondition for #4646 and #4836 .

Will need to be able to calculate a windowPeriod, when given a duration.


wiedld:

PR #4963, opened 2022-07-05
chore(4646/4836): construct a windowPeriod AST node, from a range or scopeMap nodes.

wiedld:
Precondition for #4646 and #4836 

Construct a windowPeriod AST node, when given either:
* the startNode and stopNode of a range
* or when given a subtree/node (e.g. could be aliased in scope map).
   * the scopeMap will store AST subtree nodes.
   * this way we can recursively lookup in the scopeMap (aliases of aliases).


wiedld:

PR #4964, opened 2022-07-05
chore(4646/4836): calculate window period from UI type TimeRange

wiedld:
Precondition for #4646 and #4836 

When given a UI type TimeRange, be able to provide the windowPeriod (as either a milliseconds number, or a string duration).

These methods will be used by the visitor, and later on by the UI view layer.


wiedld:

PR #4968, opened 2022-07-05
feat(4646/4836): scoped visitor to find the proper node and return th…

wiedld:
Scoped visitor to find the proper node and return the accumulated scope, at that node.

Part of #4646 
Part of #4836


wiedld:

PR #4969, opened 2022-07-06
feat(4646/4836): from a given flux AST, extract the proper scoped windowPeriod.

wiedld:
Part of  #4646 
Part of  #4836 

Using the scoped node visitor, (which returns the namespace scope captured at the node we are seeking) -- determine the windowPeriod for a given flux AST.

## Details
* line count is actually `+402`. The rest is all a giant test fixture.
* Has 2 major functions:
  * return the AST node for windowPeriod:
    * to be used for the extern, send to the backend during flux queries.
  * return a single numeric value. always.
    * to be used for areas of the UI which requires a numeric windowPeriod (in seconds) be available.

## Not in this PR, but in the next PR:
* the next PR will include updates to `src/variables/utils/getWindowVars.ts`
  * these changes will include:
    * whenever cannot parse query to AST --> return windowPeriod based on outerContext UI (a.k.a. context outside the flux query itself).

### DIFF
--- a/src/shared/utils/windowPeriod/fromAST.test.ts
+++ b/src/shared/utils/windowPeriod/fromAST.test.ts
@@ -237,6 +237,26 @@ describe('Getting a windowPeriod from a flux query AST', () => {
     })
   })
 
+  describe('can handle inner scope of functions bound to variables', () => {
+    /*
+        RAW = () => {
+          from(bucket: "sample_data")
+            |> range(start: -12h, stop: now())
+            |> filter(fn: (r) => r["_measurement"] == "airSensors")
+            |> filter(fn: (r) => r["_field"] == "co")
+            |> group()
+            |> sort(columns:["_time"])
+            |> aggregateWindow(fn: mean, every:v.windowPeriod)
+            |> drop(columns:["_measurement","_field","_start","_stop"])
+            |> yield(name:"raw ${v.windowPeriod}")
+          return true
+        }
+        RAW()
+        */
+    const ast = textFixtures['variable_bound_function']
+    runTest(ast, true, 120000)
+  })
+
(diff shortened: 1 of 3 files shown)


src/shared/utils/windowPeriod/fromAST.test.ts
```diff
@@ -0,0 +1,285 @@
...
+import {Package} from 'src/types/ast'
+import {TimeRange} from 'src/types'
+import {asAssignment} from 'src/variables/selectors'
+import {buildVarsOption} from 'src/variables/utils/buildVarsOption'
+import {TIME_RANGE_START, TIME_RANGE_STOP} from 'src/variables/constants'
+import {FALLBACK_WINDOW_PERIOD} from 'src/shared/constants/timeRanges'
+
+import textFixtures from './ast_fixtures.json'
+
+describe('Getting a windowPeriod from a flux query AST', () => {
```
wiedld, line 15: Test cases used in this file are an exact copy from the scoping issue ticket:
https://github.com/influxdata/ui/issues/4836

src/shared/utils/windowPeriod/fromAST.test.ts
```diff
@@ -0,0 +1,285 @@
...
+          v = {windowPeriod: chunk}
+          array.from(rows: [{x: v.windowPeriod}])
+          */
+      const ast = textFixtures['alias_window_variable']
+      runTest(ast, false, 2444)
+    })
+  })
+
+  // TODO: https://github.com/influxdata/ui/issues/4695
+  describe.skip('Can handle complex inner expressions:', () => {
```
wiedld, line 241: Explicitly stating the limitations of this submodule contract. What it cannot do, as defined in these skipped tests, and the issue ticket for the missing functionality.

PR #4977, opened 2022-07-06
fix(TasksPage): remove showInactive setState from constructor

### LINKED ISSUE #4897: "show inactive" in task menu doesn't save state when I return to main task menu
## About the bug

I was de-activating a bunch of tasks in an environment I shall not name, so I used that very useful toggle to not show the inactive tasks. If I'd click on a task to look at it, when I'd go back the toggle would no longer be set how I had it previously.

**Steps to reproduce:**

1. go to the tasks menu
2. click the toggle to make it so it does not show you the inactive tasks
3. click on a task
4. hit the back button or hit the task page
5. the toggle will be reset.

Maybe it's working as expected, but I felt a little sad each time I hit the back button and it wasn't toggled as I had it set before, as when I filter on a word and go back it's still filtered on that word.  Since I just spent too much time in those task menus thought i'd pass the feedback back to you all.

**Expected behavior:**

When I went back to the main task page, I thought the toggle would still be set to not show the inactive tasks.

**Actual behavior:**

when I went back, it was reset to the default (show the inactive tasks)

**Visual Proof:**

Here's a picture of the toggle set to not show the inactive tasks. I can provide more if you like but this is what i saved.

keeptoggle

## About your environment

internal cloud 2 env.

blegesse-git:
Closes #4897 

Bug: show inactive toggle inside Tasks page resets back to true when user leaves and returns to page. 
fix: removed setState responsible for re-setting `showInactive` to true.  
https://user-images.githubusercontent.com/66275100/177883487-ea85201c-4ba3-41bb-8ecb-14fe25bb63e5.mov

### Checklist

### DIFF
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -125,7 +125,7 @@ export const getAllTasks = (name?: string) => async (
     const resp = await fetchTasks(query)
 
     let nonNormalizedTasks = resp.data.tasks
-    let next = resp.data.links?.next
+    let next = resp.data?.links?.next
     while (next && next.includes('after=')) {
       const afterAndExtras = next.split('after=')
       if (afterAndExtras.length < 2) {
--- a/src/tasks/actions/thunks.ts
+++ b/src/tasks/actions/thunks.ts
@@ -125,7 +125,7 @@ export const getAllTasks = (name?: string) => async (
     const resp = await fetchTasks(query)
 
     let nonNormalizedTasks = resp.data.tasks
-    let next = resp.data.links.next
+    let next = resp.data.links?.next
     while (next && next.includes('after=')) {
       const afterAndExtras = next.split('after=')
       if (afterAndExtras.length < 2) {
--- a/src/tasks/containers/TasksPage.test.tsx
+++ b/src/tasks/containers/TasksPage.test.tsx
@@ -14,6 +14,7 @@ import TasksPage from './TasksPage'
 import {deleteTask, patchTask, postTask, getTask} from 'src/client'
 import {parse} from 'src/languageSupport/languages/flux/parser'
 import {mocked} from 'ts-jest/utils'
+import {initialState} from 'src/tasks/reducers/helpers'
 
 const sampleScript =
   'option task = {\n  name: "beetle",\n  every: 1h,\n}\n' +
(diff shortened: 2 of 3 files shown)


src/tasks/containers/TasksPage.tsx
```diff
@@ -67,10 +67,6 @@ class TasksPage extends PureComponent<Props, State> {
   constructor(props: Props) {
     super(props)
 
-    if (!props.showInactive) {
-      props.setShowInactive()
-    }
-
```
blegesse-git, line 73: this code runs every time TasksPage mounts. Which is exactly what's re-setting inactive tasks to show after user has explicitly turned off the `show inactive` toggle. 
I think the purpose here was to show inactive tasks on initial load. But since we have a default state (set to true) saved in redux, there is no need to set it to true again here.

src/tasks/containers/TasksPage.test.tsx
```diff
@@ -151,6 +152,7 @@ const setup = (override = {}) => {
         status: RemoteDataState.Done,
       },
       tasks: {
+        ...initialState(),
```
blegesse-git, line 155: needed to add `showInactive` state here so `expect(ui.store.getState().resources.tasks.showInactive).toBe(true)` can pass; line 188

src/tasks/actions/thunks.ts
```diff
@@ -125,7 +125,7 @@ export const getAllTasks = (name?: string) => async (
     const resp = await fetchTasks(query)
 
     let nonNormalizedTasks = resp.data.tasks
-    let next = resp.data.links.next
+    let next = resp.data.links?.next
```
hoorayimhelping, line 128: I think you'll probably also want to add a `?.` to `resp.data` in case links doesn't come back so like

```ts
resp.data?.links?.next
```

hoorayimhelping (approved):
Looks good - one defensive coding suggestion.

Also a commit name suggestion - semantic commits allow for adding additional context to the type of commit it is. Typically we do it by saying which section of the app we're making the change in, which I think would be helpful here. The commit would probably be more communicative if it mentioned it was on the task page, so something like:

```
fix(TasksPage): remove showInactive setState from constructor
```

It's a small thing, but it's valuable in that it adds context to someone doing a high-level scan of the git log - say in the future when we're trying to pinpoint when a bug was introduced. Knowing that this touches the tasks page makes it so we don't have to look at the changes to see where `showInactive` was removed - we'd know it's in the tasks page section of the app, which would allow someone doing a high level scan of commits to know which section of the app this touched and would help them know to either move on, or dig in deeper, depending on what their goal was.

src/tasks/actions/thunks.ts
blegesse-git: good call

Issue #4882, opened 2022-06-23
Handle the most highly frequent honeybadger-logged errors.


wiedld:
## Goal:
To convert honeybadger into a maintenance tool, for handling incoming errors in prod.
* Current challenges:
  * too many errors
  * unclear which are errors versus noise. 
* action item:
  * See the top errors. Handle them.

## Bugs to handle:
Screenshot of bugs. (Taken on 2022-07-06.)
Filtering criteria:
* last occurred was since June 1st
* sort order by occurrance count

<img width="757" alt="Screen Shot 2022-07-06 at 3 04 09 PM" src="https://user-images.githubusercontent.com/10232835/177651275-9299b9a7-7bef-4c44-9898-7ec3ee060781.png">
<img width="757" alt="Screen Shot 2022-07-06 at 4 21 56 PM" src="https://user-images.githubusercontent.com/10232835/177658956-33ec4cfe-2bd5-49d1-8bdb-e8dc2f5837ad.png">

## Questions for consideration:
* is it an error?
* can we resolve now?
* should we add/update logging, such that we get more information on this error?

wiedld:
Bug 2:
* general `window.onerror` triggered via the ResizeObserver
* This is expected, and highly reported elsewhere:
  * https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded
  * `https://github.com/WICG/resize-observer/issues/38` 
* we are already doing 2 (out of 3) code patterns to reduce the triggering occurances. We can also do the 3rd one too (cancel animation frame).
* actions suggested:
  *  cancel animation frame
  * setup honeybadger to ignore this error (in the UI honeybadger platform)

wiedld:
Did action items for top 16 bugs. Unsure of any followups @hralexander

PR #4982, opened 2022-07-07
fix(4882): reduce ResizerObserver looping even further

wiedld:
Part of #4882 

Part of bug squashing mission.

## Bug:
* general window.onerror triggered via the ResizeObserver
* What this is:
  * Excessive recursion upon resize. But is a know issue because of how this window API should be used, and is chrome-specific behavior (matching what we see in honeybadger for the browsers).
  * This is expected, and highly reported elsewhere:
    * https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded
    * `https://github.com/WICG/resize-observer/issues/38`
* Previously done:
  * we are already doing 2 (out of 3) code patterns to reduce the triggering occurrences. We can also do the 3rd one too (cancel animation frame).

In addition to this PR, going to be requesting permission to start filtering out (ignoring) this error in honeybadger.

### DIFF
--- a/src/visualization/types/SimpleTable/PagedTable.tsx
+++ b/src/visualization/types/SimpleTable/PagedTable.tsx
@@ -239,13 +239,14 @@ const PagedTable: FC<Props> = ({result, properties}) => {
     }
 
     let timeout
+    let animationFrameID
     const resizer = new ResizeObserver(entries => {
       if (timeout) {
         clearTimeout(timeout)
       }
 
       timeout = setTimeout(() => {
-        requestAnimationFrame(() => {
+        animationFrameID = requestAnimationFrame(() => {
           setHeight(entries[0].contentRect.height)
         })
       }, 200)
@@ -263,6 +264,7 @@ const PagedTable: FC<Props> = ({result, properties}) => {
 
     return () => {
       resizer.disconnect()
+      cancelAnimationFrame(animationFrameID)
       if (timeout) {
         clearTimeout(timeout)
       }


src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -239,13 +239,14 @@ const PagedTable: FC<Props> = ({result, properties}) => {
...
       if (timeout) {
         clearTimeout(timeout)
       }
 
       timeout = setTimeout(() => {
-        requestAnimationFrame(() => {
+        animationFrameID = requestAnimationFrame(() => {
           setHeight(entries[0].contentRect.height)
         })
       }, 200)
```
hoorayimhelping, line 252: do you know why this RAF is in a `setTimeout`?

wiedld: There is a general pattern when using ResizeObserver --> to then trigger a callback (e.g. set height of inner component) which then may recursively trigger ResizeObserver again. This is a known problem, which is why browsers do not surface these errors in the console...but it does still gets report in `window.onerror()` --> which we pipe to honeybadger.

Here is an issue ticket from WICG, which includes this^^ conclusion, as well as many code samples where a community of devs have listed all the code patterns which they use to decrease the number of recursive triggers (although is impossible to eliminate). One of the patterns is to have the callback action be encapsulated in a `setTimeout()` (`https://github.com/WICG/resize-observer/issues/38#issuecomment-422085926`), which queues it up at the back of the javascript runtime macrotask queue.

hoorayimhelping: >One of the patterns is to have the callback action be encapsulated in a `setTimeout()` which queues it up at the back of the javascript runtime macrotask queue.

Thanks for the explanation. Is the 200 millisecond delay significant? Could this could this be achieved with a delay of 1 or 0 milliseconds? The example doesn't set the delay value explicitly, which defaults it to a value of 0

wiedld: The timeout was put in previously. I can only guess the 200ms is arbitrary. The 0 is sufficient for insertion at the back of the macrotask queue.

PR #4994, opened 2022-07-07
feat: add ability to retry-with-backoff for failed auth attempts

hoorayimhelping:
Connects #4798 

Adds the function `retryFetchIdentity` and associated tests. It's not hooked up into the application, that will come in a future PR.

This function will retry `500` level errors after requesting data from the identity endpoint (either `/me`, `/quartz/me`, or `/quartz/identity` depending on the state of the application flags). It will retry up to 4 times, for a total of five attempts to authenticate, with an increasing backoff time between retries.

Descriptions of the jest tests added:
```
retrying failed authentication attempts
    ✓ doesn't retry when the first call is successful
    ✓ throws the error in the event of a 401
    ✓ retries once
    ✓ retries multiple times
    ✓ retries up to the retry limit
    ✓ retries until a non 500 error is encountered
    the timing of auth retrying
      ✓ retries once after the default of 30000 milliseconds
      ✓ retries multiple times, waiting longer each time
      ✓ retries multiple times, waiting longer each time, until it fails
```

### Checklist

### DIFF
--- a/src/identity/apis/auth.test.ts
+++ b/src/identity/apis/auth.test.ts
@@ -5,7 +5,6 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {
   getIdentity,
   Error as UnityError,
-  GetIdentityParams,
   Identity,
 } from 'src/client/unityRoutes'
 
--- a/src/identity/apis/auth.test.ts
+++ b/src/identity/apis/auth.test.ts
@@ -1,0 +1,282 @@
+import {mocked} from 'ts-jest/utils'
+
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
+import {
+  getIdentity,
+  Error as UnityError,
+  GetIdentityParams,
+  Identity,
+} from 'src/client/unityRoutes'
+
+import {
+  retryFetchIdentity,
+  UnauthorizedError,
+  ServerError,
+} from 'src/identity/apis/auth'
+
+jest.mock('src/shared/constants', () => ({CLOUD: true}))
+
+jest.mock('src/shared/utils/featureFlag.ts', () => ({
+  isFlagEnabled: jest.fn(),
+}))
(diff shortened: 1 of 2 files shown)


src/identity/apis/auth.ts
```diff
@@ -96,6 +103,42 @@ export const fetchIdentity = async () => {
   return fetchQuartzMe()
 }
 
+const identityRetryDelay = 30000 // 30 seconds
```
wdoconnell, line 106: Should we flag for services and e-commerce that these are the delay timings we're using so they can weigh in? (starting at thirty, incrementing delay by 30s each time)?

wdoconnell (approved):
Looks good to me. Left a couple quick comments/questions about the strategy.

ChitlangeSahas (approved):
Looks good to me!

src/identity/apis/auth.ts
hoorayimhelping: I think that's a good idea, but I'll flag them when it comes time to integrate this change into the application

src/identity/apis/auth.test.ts
```diff
@@ -0,0 +1,281 @@
...
+  } as any)
+}
+
+const flushPromises = () => {
+  return new Promise(jest.requireActual('timers').setImmediate)
+}
+
+describe('retrying failed authentication attempts', () => {
+  beforeEach(() => {
+    // return true so that `uiUnificationFlag` and `quartzIdentity` both return true
```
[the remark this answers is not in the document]
hoorayimhelping: @wdoconnell it's more that since the call is being mocked in this test, it doesn't really matter if it's making the call to `/me` or `/quartz/me` or `/quartz/identity`. The logic being tested is how the response is handled.

Although now that I think about it `/quartz/me` can return a 404, which isn't handled here 🤔

PR #4996, opened 2022-07-08
chore: remove feature flag

asalem1:
This PR removes the UI feature flag for the `Save to Version History` button so that the button appears at the menu header and not the dropdown button. After leaving the flag on for a month, we've seen user engagement double with the publish feature in Notebooks after moving the button outside the dropdown list. See the analytics [here](https://analytics.amplitude.com/influxdb/chart/83f46to?source=redirect%3A%20chart%20saved?source=copy+url)


appletreeisyellow (approved):
🎉

Issue #4049, opened 2022-03-09
Multi Org Phase 1 - User Profile Page


Luckette:
**Multi Org - User Profile Page**

**User Story: User Profile Page**
As a user, I want a user profile page separate from account and org pages so that I can set user-specific settings such as my default org.

    **ACs**
    - The user profile page displays the user's first and last name, email address, and Default Settings including default account selection and default org selection for users' active account. 
    - Hide the account and org name dropdowns in the nav bar
    - Continue to display the user profile icon in the nav bar
    - Allow the user to update their first name and last name
    - Allow the user to set a default account via a default account dropdown
    - Allow the user to set a default org for the account they're currently in (display their current active account)
    - User clicks 'save changes' to save everything on the page. 1 save button for the whole page. 

   **Dev Notes**
        - See Figma or the image below in the comments.

Luckette:
On hold, until we receive the new design from @juliajames12.

juliajames12:
Here's the final design until we get user feedback.

<img width="1512" alt="User Profile" src="https://user-images.githubusercontent.com/81173353/180860149-403fb834-2b73-4d62-8b5c-3d0b5f343d4e.png">

https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1598%3A10855

wdoconnell:
Just discussed at multi-org checkin: 

1. user profile page will not hide the globalHeader, as we are now only displaying one account at a time.
2. ability to change user first name and last name is phase 2 rather than phase 1 - this requires a new endpoint.
3. simplify notifications to just say something more basic like 'Saved user preferences.' or 'Unable to save user preferences.'

@juliajames12 @Luckette @eatondustin1

eatondustin1:
> 2. ability to change user first name and last name is phase 2 rather than phase 1 - this requires a new endpoint.

@wdoconnell Same with the email if I'm not mistaken, but that all sounds right to me.

wdoconnell:
@eatondustin1 Right -  I sensed concern though with bundling email into the same ticket, both because it requires more confirmation steps than just changing name (e.g., need a confirmation email), and also security considerations (e.g., maybe user shouldn't be allowed to switch to an email on a different domain; maybe it needs confirmation from another approval, etc.).

eatondustin1:
Ah that makes sense, I just didn't know if you left it out of your comment because you were thinking it was still part of Phase 1. So just a misunderstanding, all good!

PR #5023, opened 2022-07-11
feat: Add redux and API layer support for User Profile

wdoconnell:
This PR provides a framework that will be used to support the user profile page: UI #4049. Specifically, it

1. Adds a thunk and a reducer to handle changing the user's default organization.
2. Adds a method in the API layer to clean up error handling for changing the user's default account
     * (switch-account functionality is handled by context, not redux)
3. Adds error notifications if the request to change the user's default organization fails. 

Also cleaned up error notifications for failure to change the default account.

### Checklist

- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed) - N/A

### DIFF
--- a/src/identity/apis/auth.ts
+++ b/src/identity/apis/auth.ts
@@ -66,6 +66,13 @@ export interface CurrentIdentity {
   status?: RemoteDataState
 }
 
+export enum NetworkErrorTypes {
+  UnauthorizedError = 'UnauthorizedError',
+  NotFoundError = 'NotFoundError',
+  ServerError = 'ServerError',
+  GenericError = 'GenericError',
+}
+
 // 401 error
 export class UnauthorizedError extends Error {
   constructor(message) {
@@ -108,6 +115,42 @@ export const fetchIdentity = async () => {
   }
 
   return fetchQuartzMe()
+}
+
+const identityRetryDelay = 30000 // 30 seconds
+const retryLimit = 5
+
+export const retryFetchIdentity = async (
+  retryAttempts = 1,
+  retryDelay = identityRetryDelay
+) => {
+  try {
+    return await fetchIdentity()
+  } catch (error) {
+    if (
+      error.name === NetworkErrorTypes.UnauthorizedError ||
+      error.name === NetworkErrorTypes.GenericError
+    ) {
+      throw error
+    }
+
+    if (error.name === NetworkErrorTypes.ServerError) {
+      if (retryAttempts >= retryLimit) {
+        throw error
+      }
+      return new Promise((resolve, reject) => {
+        setTimeout(() => {
+          retryFetchIdentity(retryAttempts + 1, retryDelay)
+            .then(user => {
+              resolve(user)
+            })
+            .catch(error => {
+              reject(error)
+            })
+        }, retryAttempts * retryDelay)
+      })
+    }
+  }
 }
 
 // fetch user identity from /quartz/identity.
@@ -246,5 +289,6 @@ export const putDefaultQuartzOrg = async (orgId: string) => {
   if (response.status !== 204) {
     throw new ServerError(response.data.message)
   }
+
   return response.data
 }
--- a/src/identity/apis/auth.ts
+++ b/src/identity/apis/auth.ts
@@ -229,9 +229,9 @@ export const fetchAccountDetails = async (
 }
 
 // change the user's default account
-export const putDefaultQuartzAccount = async (
+export const updateDefaultQuartzAccount = async (
   accountId: number
-): Promise<Organization> => {
+): Promise<void> => {
   const response = await putAccountsDefault({
     data: {
       id: accountId,
@@ -242,9 +242,7 @@ export const putDefaultQuartzAccount = async (
     throw new ServerError(response.data.message)
   }
 
-  // success status code is 204; no response in body is expected.
-  const responseEmpty = response.data
-  return responseEmpty
+  // success status code is 204; no data in response.body is expected.
 }
 
 // fetch details about user's current organization
--- a/src/identity/apis/auth.ts
+++ b/src/identity/apis/auth.ts
@@ -277,7 +277,7 @@ export const fetchQuartzOrgs = async (): Promise<OrganizationSummaries> => {
 }
 
 // change default organization for a given account
-export const putDefaultQuartzOrg = async (orgId: string) => {
+export const updateDefaultQuartzOrg = async (orgId: string) => {
   const response = await putOrgsDefault({
     data: {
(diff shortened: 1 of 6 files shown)


src/identity/apis/auth.ts
```diff
@@ -227,6 +228,25 @@ export const fetchAccountDetails = async (
   return accountDetails
 }
 
+// change the user's default account
+export const putDefaultQuartzAccount = async (
```
hoorayimhelping, line 232: Not a huge deal, but we might consider calling this `updateDefaultQuartzAccount` since `PUT` is an obvious HTTP verb, and `update` is a little more descriptive, and this is the layer where we should feel comfortable translating HTTP concepts into app-domain concepts. In other words PUT is an HTTP-level concept, `update` is an app-domain concept.

For example, if we decide that we actually don't want to use PUT, and instead the implementation changes to POST, this method will either make no sense, or need to be changed in lockstep

src/identity/apis/auth.ts
```diff
@@ -227,6 +228,25 @@ export const fetchAccountDetails = async (
...
+    },
+  })
+
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+
+  // success status code is 204; no response in body is expected.
+  const responseEmpty = response.data
+  return responseEmpty
```
hoorayimhelping, line 247: I think this should be removed, since it doesn't return anything and this is kind of confusing, and it looks like the return value isn't used by anything anyway. The return value should probably be `Promise<void>`. JavaScript functions implicity return `undefined` if there's no return value, which it looks like will be the same as this code. The comment should probably stay, just to keep someone who is reading this block of code from having to look at the definition of `putAccountsDefault` in the `quartzRoutes`

src/shared/copy/notifications/categories/accounts-users-orgs.ts
```diff
@@ -61,6 +71,29 @@ export const orgRenameFailed = (orgName): Notification => ({
...
+  message:
+    'Error retrieving new organization information. Please refresh this page.',
+})
+
+export const updateQuartzOrganizationsFailed = (): Notification => ({
+  ...defaultErrorNotification,
+  message:
+    'We were unable to retrieve the list of your InfluxData organizations. Please refresh this page.',
+})
+
```
hoorayimhelping, line 96: Absent any other organization scheme, I'd like us to default to alphabetization, since that will probably be the easiest way to find functions in a file like this.

src/identity/quartzOrganizations/actions/thunks/index.ts
```diff
@@ -33,3 +41,20 @@ export const getQuartzOrganizationsThunk = () => async (
...
+  try {
+    dispatch(setQuartzOrganizationsStatus(RemoteDataState.Loading))
+    await putDefaultQuartzOrg(newDefaultOrg.id)
+    dispatch(setQuartzDefaultOrg(oldDefaultOrg.id, newDefaultOrg.id))
+
+    dispatch(setQuartzOrganizationsStatus(RemoteDataState.Done))
+    dispatch(notify(accountDefaultSettingSuccess(newDefaultOrg.name)))
+  } catch (err) {
+    dispatch(setQuartzOrganizationsStatus(RemoteDataState.Error))
+    dispatch(notify(accountDefaultSettingError(newDefaultOrg.name)))
```
hoorayimhelping, line 58: This doesn't need to be changed right now, but I think we should consider not notifying from the thunks layer if we're going to be setting the `RemoteDataState` to `Error`. Dispatching actions for notifications in the thunk layer will make re-usability of the thunks a little bit harder because it tightly couples the error handling with the messaging. It seems like it should be the Component layer's responsibility to notify the user based on whether the `RemoteDataState` is set to `Done` or `Error`.

Give it a think and let me know how you feel

hoorayimhelping (approved):

src/identity/apis/auth.ts
wdoconnell: Makes sense, and agree. I've changed this to `updateDefaultQuartzAccount` throughout.

src/shared/copy/notifications/categories/accounts-users-orgs.ts
wdoconnell: SGTM. 

There was a comment in there, `/* USERS NOTIFICATIONS */` that seems to separate out notifications relating to user invites. But there's not really a logical reason why those should be distinct in the file, so I'll impose alpha order here.

src/identity/quartzOrganizations/actions/thunks/index.ts
wdoconnell: @hoorayimhelping Thanks - completely agree on placing the notifications in the Component layer instead. I'll save the notification for the UserProfile component itself. Some thoughts here:

On using `RemoteDataState`, I'm sort of divided. Checking `if (status === 'done')` versus `if (status === 'error')` as the condition for deciding which notification we dispatch doesn't seem ideal in my view, because either we either:

1. have to create a separate 'status' in redux for every new API request or thunk we create (updateDefaultOrgStatus, getOrgsStatus, getBillingInfoStatus) - which is annoying and cluttered, or 
2. have to uncross the wires from multiple thunks potentially updating the same redux `status` (e.g., API requests that are getting orgs or setting new org properties both set the `status` for `organizations` to 'error' if they go wrong, so our `if` block can't distinguish the error associated with one failed api request from another). That seems worse to me because then we know the app is in an error state but we can't trace why, and there's a risk the user gets an error notification for the wrong reason.

Maybe we throw an error in the thunk if the API request fails; and in the UI, we keep the existing `try/catch` blocks, using them to dispatch the notification so that it's tied directly to the error, instead of indirectly to the change in the redux state. We can also update the `RemoteDataState`, but maybe this is a bigger question about what those data state updates are being used as an input for.

hoorayimhelping: >Maybe we throw an error in the thunk if the API request fails; and in the UI, we keep the existing try/catch blocks, using them to dispatch the notification so that it's tied directly to the error, instead of indirectly to the change in the redux state.

hahaha, we've come full-circle on the conversation about this we had a couple of weeks ago when you asked if we should return the error from the thunk, or dispatch an action. I think you make good points in both cases, but I tend to think we should handle the returned error directly, for the reasons you outlined here.

wdoconnell: Hah, agreed. There's probably a bigger conversation to be had about the use of the redux status - it's possible I'm missing the utility of doing it that way, but we can always reconsider if this behavior doesn't seem optimal.

PR #5027, opened 2022-07-12
fix(4882): change docs link used for csv.from(), to get closer to what works.

wiedld:
Part of #4882 

## Error message in honeybadger:
* `uploadCsv function: error in csv.from()...`
* There is a whole set of these errors coming from when the csv uploaded (via flux query to backend) is invalid.
  * Expected error. This information is already surfaced to the user.
* However, there is ALSO a big spike in these errors starting April 12th.
  * Screen Shot 2022-07-11 at 6 24 26 PM
  * the vast majority of these are `error in csv.from(): failed to read metadata: missing expected annotation datatype. consider using the mode: "raw" for csv`

## Observation from use:
* the example provided at the link does not work.
  * is both out of date (older flux version). And points out a bunch of annotated csv examples which does not works. (Because we need ALL header and annotation rows.)
  * Will get the "raw" error seen in honeybadger.
* This example does work.
  * https://docs.influxdata.com/influxdb/cloud/reference/syntax/annotated-csv/#annotated-csv-in-flux
    * this is actually what we use to upload the csv anyways. We are using annotated-csv-in-flux (in our hardcoded flux). 
  * NOTE: the user will still have to add in the `_field`. Since adding to a bucket. (This explicit message will be surfaced to the user.)

## Followup:
- [x] file ticket with docs, to get the `_field` added to that example too. So it will work on copy/paste.
    * https://github.com/influxdata/docs-v2/pull/4206

### DIFF
--- a/src/writeData/components/fileUploads/AnnotatedCSV.md
+++ b/src/writeData/components/fileUploads/AnnotatedCSV.md
@@ -1,4 +1,4 @@
-For more detailed and up to date information check out the [Annotated CSV Documentation](https://docs.influxdata.com/influxdb/v2.3/reference/syntax/annotated-csv/#annotated-csv-in-flux)
+For more detailed and up to date information check out the [Annotated CSV Documentation](https://docs.influxdata.com/influxdb/cloud/reference/syntax/annotated-csv/#annotated-csv-in-flux)
 
 ##### Getting Started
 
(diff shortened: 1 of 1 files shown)


wiedld:
> Do we have an example of the failures? Like, is the CSV that users are trying to upload a supported CSV and so we should change the mechanism with which we're uploading the CSVs? Or is it just flat out incorrect?

The honeybadger error does not record the csv used in the attempt. Only the error message. What I found was that by clicking on the provided link, and copy/paste the example --> I got the exact same error which kept being reported. Then I went looking for a docs example which I could copy/paste over, and get it to work. That's the updated link.

To be clear -- this will not remove the honeybadger error permanently. A certain percentage of those errors will always occur, when users upload a bad formatted file. So the error is expected, within reason. But since the incidents of this error spiked so much...I wanted to offset with the link to proper documentation.

wiedld:
Updated the link used:
* old link went to OSS (not cloud), and version 2.0
* updated the link to be cloud --> which in docs-v2 still refers version 2.0 (not ver 2.3)

PR #5039, opened 2022-07-12
feat(sub): edit form field validation

genehynson:
Closes https://github.com/influxdata/data-acquisition/issues/415

Edit form field validation:
- same validation as create form - "save changes" button is disabled if form is not valid
- tweaked edit button - Edit -> Cancel
- editing broker hostname clears password
- added note that changing broker hostname will clear password

https://user-images.githubusercontent.com/6411855/178611046-835affb7-5f69-41b3-91bf-b932ead0e702.mov

### Checklist

### DIFF
--- a/src/writeData/subscriptions/components/BrokerFormContent.tsx
+++ b/src/writeData/subscriptions/components/BrokerFormContent.tsx
@@ -184,6 +184,11 @@ const BrokerFormContent: FC<Props> = ({
               validationFunc={() =>
                 handleValidation('Broker Host', formContent.brokerHost)
               }
+              helpText={
+                edit
+                  ? "Changing the broker's hostname will require your broker's password to be provided again."
+                  : ''
+              }
             >
               {status => (
                 <Input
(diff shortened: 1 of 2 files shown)


cannonpalms (approved):

src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -196,6 +201,11 @@ const BrokerFormContent: FC<Props> = ({
                     updateForm({
                       ...formContent,
                       brokerHost: e.target.value,
+                      // clear the password field if broker host is edited
+                      brokerPassword:
+                        className === 'create'
+                          ? formContent.brokerPassword
```
subirjolly, line 207: i think you need this: `formContent?.brokerPassword`

subirjolly:
`added note that changing broker hostname will clear password`
Also for this, I think we should have a patch so that we don't force the customers to add pass again, once done.

subirjolly (approved):
other than just 1 comment, looks good

cannonpalms:
> `added note that changing broker hostname will clear password` Also for this, I think we should have a patch so that we don't force the customers to add pass again, once done.

We have to require the password anytime the host is changed, because otherwise, changing the host will result in their existing password being sent somewhere else. 

Imagine I get into your account and want the password to your broker. I change the hostname to any public proxy/MITM service and read your password out of the logs in cleartext.

cannonpalms (approved):

subirjolly:
> 

I totally understand and agree with you. I meant to divide up a Subscription into sections so that we can update one field or a small section of the subscription without affecting the other fields. Completely agree that host info/creds needs to be bundled as a section. Sorry for not being clear earlier.

PR #5048, opened 2022-07-13
fix: sub user password

helenosheaa:
Relates https://github.com/influxdata/data-acquisition/issues/412

- [x] turns password field into a password input obscuring the entered text
- [x] returns the username
- [x] correctly selects the username/password toggle for the edit view
- [x] clears the username and password if they change the toggle back to none

https://user-images.githubusercontent.com/38860767/178832834-88258ba0-0d87-4255-8abd-c1a8f8401b7f.mov

### Checklist

### DIFF
--- a/src/writeData/subscriptions/components/BrokerFormContent.tsx
+++ b/src/writeData/subscriptions/components/BrokerFormContent.tsx
@@ -297,8 +297,8 @@ const BrokerFormContent: FC<Props> = ({
                   {feature: 'subscriptions'}
                 )
                 setSecurity('none')
-                formContent.brokerUsername = undefined
-                formContent.brokerPassword = undefined
+                formContent.brokerUsername = null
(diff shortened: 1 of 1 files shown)


src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -291,6 +297,8 @@ const BrokerFormContent: FC<Props> = ({
                   {feature: 'subscriptions'}
                 )
                 setSecurity('none')
+                formContent.brokerUsername = undefined
+                formContent.brokerPassword = undefined
```
genehynson, line 301: can we change this to `null`? I know we're handling both on the backend but I think it would be nice for this to have this set the values correctly in case that `if block` in nifid is ever removed

genehynson (approved):

PR #5050, opened 2022-07-13
fix: normalize windowPeriod when null for zoom re-query feature

TCL735:
Part of EAR 2422

- allows the zoom re-query feature to work properly on complex Flux functions or whenever the `windowPeriod` is `null`
- apply a fallback method for calculating `windowPeriod` when `null` by using the user selected domain (zoom domain) and the original data set to estimate the `windowPeriod` over the optimal number of graph points

Here is a video of the zoom re-query working with a complex Flux script:

https://user-images.githubusercontent.com/10736577/178856359-a4909ccb-1bdf-4c68-a6ad-f196c8848916.mp4

### DIFF
--- a/src/variables/utils/getWindowVars.ts
+++ b/src/variables/utils/getWindowVars.ts
@@ -1,3 +1,6 @@
+// Libraries
+import {NumericColumnData} from '@influxdata/giraffe'
+
 // APIs
 import {parse} from 'src/languageSupport/languages/flux/parser'
 
@@ -17,11 +20,12 @@ import {
 
 // Types
 import {VariableAssignment, Package} from 'src/types/ast'
-import {RemoteDataState, Variable} from 'src/types'
+import {RemoteDataState, TimeRange, Variable} from 'src/types'
 import {SELECTABLE_TIME_RANGES} from 'src/shared/constants/timeRanges'
 
 const DESIRED_POINTS_PER_GRAPH = 360
 const FALLBACK_WINDOW_PERIOD = 15000
+const MINIMUM_WINDOW_PERIOD = 1
 
 /*
   Compute the `v.windowPeriod` variable assignment for a query.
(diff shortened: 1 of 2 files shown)


src/variables/utils/getWindowVars.ts
```diff
@@ -227,3 +231,83 @@ export const getVariableForZoomRequery = (
...
+ *     to the original data set and divide by the optimal number of graph points
+ *     to get the estimated windowPeriod
+ */
+export const normalizeWindowPeriodForZoomRequery = (
+  windowPeriod: number | null,
+  timeRange: TimeRange,
+  domain: Array<number>,
+  column: NumericColumnData | string[]
+): number => {
+  if (windowPeriod || !timeRange || !column) {
```
hoorayimhelping, line 248: what does `!timeRange` and `column` actually mean? The [type for TimeRange](https://github.com/influxdata/ui/blob/master/src/types/queries.ts#L18) makes it seem like `!timeRange` is something that can't happen? And it looks like `NumericColumnData` is going to be an array, so should it instead be `!column.length`?

src/variables/utils/getWindowVars.ts
```diff
@@ -227,3 +231,83 @@ export const getVariableForZoomRequery = (
...
+      endIndex = counter
+    }
+    counter += 1
+  }
+
+  const duration =
+    (Number(column[endIndex]) - Number(column[startIndex])) /
+    DESIRED_POINTS_PER_GRAPH
+
+  if (duration !== duration) {
```
hoorayimhelping, line 274: is this to guard against `NaN`? If so, could [`isNaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) be used instead?

hoorayimhelping (approved):
I'm not super familiar with this area of the code, so I'm afraid I can't offer too great of a review here

src/variables/utils/getWindowVars.ts
TCL735: I always get confused whether `isNaN` or `Number.isNaN` is better and usually end up with this equality check as the quicker way of guarding against `NaN`.

But you are right, we should use a built-in checker for this.

The correct one is... `Number.isNaN` so I will go with that.

PR #5071, opened 2022-07-15
chore: remove unused code, when team decided to take different route with windowPeriod

wiedld:
Team decided to not fix the current windowPeriod calculation. Removing what code has been merged.


appletreeisyellow (approved):
Thanks for cleaning the unused code up!

PR #5076, opened 2022-07-17
feat: remove the webworker parser since it's unused

asalem1:
This PR consolidates the changes made to the fromFlux parser with the webworker parser. The changes being applied here are the changes that have occurred to the fromFlux parser in Giraffe that have been tested and iterated upon for over a month now, with the final changes (for now) being ported over and updated in the Giraffe library.


drdelambre (approved):
the web worker didn't buy as much of a perf boost as we had all imagined going into this call, and this implementation has a few issues elsewhere in the flag where the queries sometimes aren't run, preventing the flag from being turned on. as there's issues and no one to fix them, we're depreciating the flag and this is a good first step on that path

PR #5101, opened 2022-07-19
chore: clean up the cell context to reduce the noise;

asalem1:
This PR cleans up the CellContext to make things a bit clearer w/r/t component-based concerns. In doing some investigative work on the dashboard slowness, I came across this component where we were doing things in two different fashions. This normalizes the pattern to be more in line with functional components. In doing so, we were able to remove the `variables` from the component since it was never being used


appletreeisyellow (approved):
Thanks for cleaning it up!

PR #5108, opened 2022-07-19
fix: improve quartzOrganizations reducer compatibility with quartz

### LINKED ISSUE #5103: Improve quartzOrganizations reducer compatibility with quartz backend
Quartz allows for the possibility of an account without a default organization. The quartzOrganizations reducer should accept this as a valid option.

'Existing default org' should also not be an argument that is part of the action dispatched to the redux store. The current default org can be gleaned in the reducer itself from the list of orgs, to avoid the UI getting out of sync with the redux store.

wdoconnell:
Closes #5103 

This PR resolves two issues with the current `quartzOrganizations` reducer, which is to be used by GlobalHeader.

1. quartz allows for the possibility that there is no current 'default' organization. The reducer needs to not `return state` or `throw Error` in that case.

2. the _current_ default org shouldn't need to be part of the action dispatched to the reducer. The reducer already has that information available to it in the list of organizations. Having this property in the action just leaves room for error.

This fix resolves both issues.

### Checklist

- [X] Feature flagged, if applicable [YES - MULTIORG]

### DIFF
--- a/src/identity/quartzOrganizations/actions/thunks/index.ts
+++ b/src/identity/quartzOrganizations/actions/thunks/index.ts
@@ -42,16 +42,16 @@ export const getQuartzOrganizationsThunk = () => async (
   }
 }
 
-export const updateDefaultOrgThunk = (
-  oldDefaultOrg: DefaultOrg,
-  newDefaultOrg: DefaultOrg
-) => async (dispatch: Dispatch<Actions>, getState: GetState) => {
+export const updateDefaultOrgThunk = (newDefaultOrg: DefaultOrg) => async (
+  dispatch: Dispatch<Actions>,
+  getState: GetState
+) => {
   try {
     dispatch(setQuartzOrganizationsStatus(RemoteDataState.Loading))
 
     await updateDefaultQuartzOrg(newDefaultOrg.id)
 
-    dispatch(setQuartzDefaultOrg(oldDefaultOrg.id, newDefaultOrg.id))
+    dispatch(setQuartzDefaultOrg(newDefaultOrg.id))
 
     dispatch(setQuartzOrganizationsStatus(RemoteDataState.Done))
   } catch (err) {
--- a/src/identity/quartzOrganizations/reducers/index.ts
+++ b/src/identity/quartzOrganizations/reducers/index.ts
@@ -8,11 +8,19 @@ import {
 import {emptyOrg} from 'src/identity/components/GlobalHeader/DefaultEntities'
 import produce from 'immer'
 
-import {OrganizationSummaries} from 'src/client/unityRoutes'
+// Types
 
-const initialState = {
+import {OrganizationSummaries} from 'src/client/unityRoutes'
+export const initialState = {
   orgs: [emptyOrg] as OrganizationSummaries,
 } as QuartzOrganizations
+
+export class OrgValidationError extends Error {
+  constructor(message) {
+    super(message)
+    this.name = 'OrgValidationError'
+  }
+}
 
 export default (state = initialState, action: Actions): QuartzOrganizations =>
   produce(state, draftState => {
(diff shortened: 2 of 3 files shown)


src/identity/quartzOrganizations/reducers/index.ts
```diff
@@ -27,19 +35,35 @@ export default (state = initialState, action: Actions): QuartzOrganizations =>
...
+            org.isDefault = true
+            newIdCount++
+          }
+        })
+
+        if (newIdCount !== 1) {
+          throw new OrgValidationError(
+            'Error: Failed to identify new default org.'
+          )
+        }
```
hoorayimhelping, line 65: We almost certainly don't want to throw an error from a reducer, since they need to be pure, and throwing an error is a side effect (functional programming, amirite?) I think instead the reducer will need to set some kind of error state

hoorayimhelping (changes requested):

src/identity/quartzOrganizations/reducers/index.ts
wdoconnell: 💯 I'll rewrite this so that it sets the status of this portion of the redux state to `RemoteDataState.Error`.

src/identity/quartzOrganizations/reducers/index.ts
```diff
@@ -19,27 +21,46 @@ export default (state = initialState, action: Actions): QuartzOrganizations =>
...
-            if (org.id === oldDefaultOrgId) {
-              org.isDefault = false
-            }
-
-            if (org.id === newDefaultOrgId) {
-              org.isDefault = true
-            }
-          })
+        // No existing default org is acceptable; oldDefaultOrg may be undefined.
+        const oldDefaultOrg = draftState.orgs.find(
```
wdoconnell, line 36: A totally valid criticism of this approach is that we could probably ditch both the `find()` and the `forEach()` loop if we just normalized the array of orgs we get from the API into an object with the `orgIds` as strings, and stored the current default `orgId` as a property so that we could do direct lookup on specific orgs. I could see that being a useful way to have state organized for some purposes, but for now I wouldn't implement the reducer that way for these reasons:

1. The `find()` and `forEach()` aren't expensive in a meaningful way for the user. They're not happening in a render, and the user only clicks 'change default' once in a while. This is getting run in response to a specific click, not page load.
2. Orgs are expected to be in an array format in the UI and in the clockface components, so we'd need to de-normalize the object anyway, which probably erases the performance benefit.
3. I think it'd be overengineering this reducer. This isn't doing that much work, so it doesn't need 2x as much code.

src/identity/quartzOrganizations/actions/thunks/index.ts
```diff
@@ -31,7 +31,6 @@ export const getQuartzOrganizationsThunk = () => async (
     const quartzOrganizations = await fetchQuartzOrgs()
 
     dispatch(setQuartzOrganizations(quartzOrganizations))
-    dispatch(setQuartzOrganizationsStatus(RemoteDataState.Done))
```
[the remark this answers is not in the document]
hoorayimhelping: makes sense; good change

src/identity/quartzOrganizations/reducers/index.ts
```diff
@@ -19,27 +21,46 @@ export default (state = initialState, action: Actions): QuartzOrganizations =>
...
+            newIdCount++
+          }
+        })
+
+        if (newIdCount !== 1) {
+          draftState.status === RemoteDataState.Error
+          return
+        }
+
+        draftState.status === RemoteDataState.Done
```
hoorayimhelping, line 63: should these comparisons be assignments?

hoorayimhelping (approved):

PR #5112, opened 2022-07-19
feat: filter out sample buckets from the bucket dropdown

### LINKED ISSUE #5082: Filter Sample Data from new Data Explorer
## **Task Description**

<p></p>

Sample data doesn't work with the new data explorer, and since sample data uses special flux handling of CSVs vs. Buckets that would now be visibile and confusing the user, we want to filter sample data in the new data explorer until we can revisit how sample data works/behaves so it can be handled just like any other bucket.

<p></p>

## Acceptance Criteria

<p></p>

Sample Data no longer shows in new Data Explorer

<p></p>

<p></p>

asalem1:
Closes #5082 

This PR filters the `sample` type buckets from the Bucket Selector. 

### Background

The purpose of the new query builder is to empower the user to learn flux. Since the `sample` bucket types aren't actually buckets, we are removing the options from the dropdown so that new flux users don't get confused by the semantic difference since the way that the UI presents the data, it's unclear what the difference is or how those differences impact a query

### DIFF
--- a/cypress/e2e/shared/fluxQueryBuilder.test.ts
+++ b/cypress/e2e/shared/fluxQueryBuilder.test.ts
@@ -33,8 +33,8 @@ describe('FluxQueryBuilder', () => {
   describe('Schema browser', () => {
     const bucketName = 'defbuck'
     const measurement = 'ndbc'
-    const searchField = 'air_temp_degc'
-    const searchTagKey = 'station_id'
+    const searchTagKey = 'air_temp_degc'
+    const searchField = 'station_id'
 
     it('can search buckets, measurements, fields and tag keys dynamically and loads more data when truncated', () => {
       // open the bucket list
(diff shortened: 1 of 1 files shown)


appletreeisyellow (approved):

src/dataExplorer/components/BucketSelector.tsx
```diff
@@ -25,7 +25,6 @@ is stored. You can think of a bucket like you would a database in SQL systems.`
 const REMAP_BUCKET_TYPES = {
   user: 'My Data',
   system: 'System Data',
-  sample: 'Sample Data',
```
drdelambre, line 28: keep this line

drdelambre:
@appletreeisyellow i talked to ari after you approved it and we worked to move the filtering logic from the view layer to the data layer and apply sample data buckets only on the output of the cache (and keep api buckets on the inside of the cache) and only if a prop is sent down. mind re-reviewing?

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -10,6 +10,11 @@ describe('FluxQueryBuilder', () => {
           cy.setFeatureFlags({
             newDataExplorer: true,
           }).then(() => {
+            const writeData = []
+            for (let i = 0; i < 30; i++) {
+              writeData.push(`ndbc,air_temp_degc=${i}_degrees station_id=${i}`)
```
appletreeisyellow, line 15:
```suggestion
              writeData.push(`ndbc,station_id=${i} air_temp_degc=${i}_degrees`)
```

Or change the line protocol to keep `air_temp_degc` as field and `station_id` as tag key

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -26,21 +31,12 @@ describe('FluxQueryBuilder', () => {
   })
 
   describe('Schema browser', () => {
-    const bucketName = 'NOAA National Buoy Data'
+    const bucketName = 'defbuck'
     const measurement = 'ndbc'
     const searchField = 'air_temp_degc'
     const searchTagKey = 'station_id'
```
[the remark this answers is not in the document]
drdelambre: i chose this one!

cypress/e2e/shared/fluxQueryBuilder.test.ts
appletreeisyellow, line 15:
```suggestion
              writeData.push(`ndbc,air_temp_degc=${i}_degrees station_id_${i}=${i}`)
```

Might want to create 30 different measurements so that "Load more" button can be tested properly later in line 151 and 160

drdelambre: op. i see the confusion. ari removed the testing of the load more functionality for fields, but also added it for tag values?

appletreeisyellow: Got it! Then disregard my suggestion 😅

appletreeisyellow (approved):
Thanks for updating the test cases!

PR #5116, opened 2022-07-19
fix: only add visualization for active query

drdelambre:
this pr addresses some of the product weirdness around exporting notebooks from data explorer that i found while verifying that an EAR was indeed closed. there is one strange artifact from the EAR that i cannot reproduce (though it's totally in the user's data), where the query defined in draftQuery between the two pipes.push references back to back was different. that i can only assume came from a reference error cause by some external mutation so i dropped in a deep clone there to isolate it just in case.

### DIFF
--- a/src/dataExplorer/components/SaveAsNotebookForm.tsx
+++ b/src/dataExplorer/components/SaveAsNotebookForm.tsx
@@ -41,7 +41,7 @@ interface Props {
 const SaveAsNotebookForm: FC<Props> = ({dismiss}) => {
   const [notebookName, setNotebookName] = useState('')
   const orgID = useSelector(getOrg).id
-  const {draftQueries, autoRefresh, timeRange} = useSelector(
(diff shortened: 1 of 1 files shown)


src/dataExplorer/components/SaveAsNotebookForm.tsx
```diff
@@ -41,7 +41,7 @@ interface Props {
 const SaveAsNotebookForm: FC<Props> = ({dismiss}) => {
   const [notebookName, setNotebookName] = useState('')
   const orgID = useSelector(getOrg).id
-  const {draftQueries, autoRefresh, timeRange} = useSelector(
+  const {draftQueries, autoRefresh, timeRange, activeQueryIndex} = useSelector(
```
appletreeisyellow, line 44: Just curious, what does `activeQueryIndex` do?  My understanding is that there can be many queries but only one is active when the user actively working on it, and in this case, the user is saving a specific query from data explorer to notebook

appletreeisyellow (approved):
Hope this can resolve the weirdness!

Issue #5093, opened 2022-07-18
Feature Flag Cleanup: firstMile


hoorayimhelping:
### Acceptance Criteria
- remove feature flag `firstMile`

PR #5125, opened 2022-07-20
chore: remove feature flag firstMile

hoorayimhelping:
Connects https://github.com/influxdata/ui/issues/5093

- Removes the feature flag `firstMile` and the component, `MePage.tsx` that the firstMile experience replaces.
- Moves the styles that were linked to `MePage.tsx` into the `HomepageExperience.scss` stylesheet
- Removes the MePage index file that isn't in use
- Removes the `home.test.ts` since they target the old home page, and the firstMile tests handle the new functionality.
- Skips the pinned items tests, since the feature is off in cloud and the tests were also targeting the home page

### Checklist


src/homepageExperience/components/HomepageExperience.scss
```diff
@@ -110,3 +111,43 @@ h1#first-mile--header {
 .homepage-wizard--code-highlight {
   color: #b7b8ff;
 }
+
+/*
+ * Styles copied from the removed MePage.scss
```
wdoconnell, line 116: Is the reason we're keeping the styles that they get imported via the chronograf file - and they're used elsewhere in the app? I spot checked a couple of these and that seemed to be the case.

wdoconnell (approved):
LGTM.

src/homepageExperience/components/HomepageExperience.scss
hoorayimhelping: correct. Other parts of the application depend on these styles, but the concept of a `MePage.scss` doesn't make sense anymore since `MePage.tsx` was removed

src/shared/containers/SetOrg.tsx
```diff
@@ -296,12 +295,7 @@ const SetOrg: FC = () => {
...
 
           {/* Getting Started */}
-          {isFlagEnabled('firstMile') ? (
-            <Route exact path="/orgs/:orgID" component={HomepageContainer} />
-          ) : (
-            <Route exact path="/orgs/:orgID" component={MePage} />
-          )}
-
+          <Route exact path="/orgs/:orgID" component={HomepageContainer} />
           {isFlagEnabled('firstMile') && [
```
blegesse-git, line 299: If `HomepageContainer` is what will be shown on the home page (despite firstMile ff being on or off), should we also remove this ff here?

blegesse-git (approved):

src/shared/containers/SetOrg.tsx
hoorayimhelping: oh wow, nice find. I totally missed this.

blegesse-git: noooice!

cypress/e2e/cloud/home.test.ts
```diff
@@ -1,63 +0,0 @@
-import {Organization} from '../../../src/types'
```
wdoconnell, line 1: I don't know anything about the `alertsActivity` flag, but it does still appear to exist (it's set to `false`) in prod. Do we need to move these tests over to new firstMile page, rather than removing them?

cypress/e2e/cloud/pinned.test.ts
```diff
@@ -1,6 +1,11 @@
 import {Organization} from '../../../src/types'
 
-describe('Pinned Items', () => {
+// This functionality is not currently on in cloud. These tests target
```
wdoconnell, line 3: Do we need to cross-reference this on any PR so that anyone working on the pinned items feature knows these tests are getting turned off?

wdoconnell (commented):
LGTM, but had two questions first re: the test disabling/removal.

cypress/e2e/cloud/home.test.ts
hoorayimhelping: No, since the `AlertsActivity` component isn't part of first mile right now

cypress/e2e/cloud/pinned.test.ts
hoorayimhelping: I don't believe there are any issues or PRs related to pinned items at the moment

blegesse-git (approved):

cypress/e2e/cloud/home.test.ts
wdoconnell: Great! 👍

Issue #4767, opened 2022-06-07
firstMile: InfluxCLI Onboarding - Install Dependencies - LINUX


Luckette:
ACs
    -Install Dependencies page should include three tabs underneath the "Install Dependencies" text specific to the user's operating system (MAC OS, WINDOWS, LINUX). See Figma
    - Depending on which OS tab the user clicks, the install information below will change accordingly. The selected Tab should highlight blue and the text should be white. 
    - For LINUX, the first sub-step should be "Download from the command line." Copy/Code snippet should match what's in the Figma. (If the URL containing "lastest is available for use by the CLI team, display it in the code snippet)
    - Text underneath the copy/code snippet should say "If you prefer to manually download the CLI package, see our documentation. The word "documentation" should hyperlink to: https://docs.influxdata.com/influxdb/cloud/tools/influx-cli/?t=Linux        
    - The next sub-step should be "Place the executable in your $PATH (optional)"
    - Text underneath "Place the executable in your $PATH (optional)" should read "If you do not move the influx binary into your $PATH, prefix the executable ./ to run it in place." Then display the Copy/Code snippet of how to place the executable in a path (See Figma)
    - The next sub-step should be "Useful InfluxCLI Commands"
    - Text underneath "Useful InfluxCLI Commands" should be "To invoke a command, use the following format in the command line: influx <command>" (See Figma)
    - In the "Useful InfluxCLI Commands" sub-step/section should be a table of useful commands and their definitions. (bucket, export, help, help <command>, query, and write)
    - Underneath the table of useful commands should be the following text and link "A full list of InfluxCLI Commands is available in our documentation." The word documentation should be hyperlinked to:(https://docs.influxdata.com/influxdb/cloud/reference/cli/influx/)              
    
   Dev Notes
    - If URL  to the InfluxCLI is used, use the URL with "latest" in the title - assuming that the edge/cli team finished their ticket (https://github.com/influxdata/influx-cli/issues/394)  
    - Figma: https://www.figma.com/file/JfGYCOb7ZyQ8JHSqeu4fma/Onboarding-Exploration?node-id=1137%3A5331

Issue #4766, opened 2022-06-07
firstMile: InfluxCLI Onboarding - Install Dependencies - WINDOWS


Luckette:
ACs
    - Install Dependencies page should include three tabs underneath the "Install Dependencies" text specific to the user's operating system (MAC OS, WINDOWS, LINUX). See Figma
    - Depending on which OS tab the user clicks, the install information below will change accordingly. The selected Tab should highlight blue and the text should be white.         
    - For WINDOWS, the next sub-step on the Install Dependencies page should be "Download the CLI package"
    - Text underneath the "Download the CLI package" should be "Download via our documentation" The entire text should be hyperlinked to: https://docs.influxdata.com/influxdb/cloud/tools/influx-cli/?t=Windows
    - The next sub-step should be "Expand the downloaded archive"
    - The text underneath "Expand the downloaded archive" should be "Expand the downloaded archive into C:\Program Files\InfluxData\ and rename if desired"
    - Display the copy/code snippet to expand the downloaded archive (See Figma)
    - The next sub-step should be "Grant network access (optional)"
    - Text underneath "Grant network access (optional)" should be: "To grant network access to the InfluxCLI, do the following: 1. Select Private networks, such as my home or work network. 2. Click Allow access.
    - The next and last sub-step in the WINDOWS flow should be "Useful CLI Commands."
    - Text underneath "Useful InfluxCLI Commands" should be "To invoke a command, use the following format in the command line: influx <command>" (See Figma)
    - In the "Useful InfluxCLI Commands" sub-step/section should be a table of useful commands and their definitions. (bucket, export, help, help <command>, query, and write)
    - Underneath the table of useful commands should be the following text and link "A full list of InfluxCLI Commands is available in our documentation." The word documentation should be hyperlinked to:(https://docs.influxdata.com/influxdb/cloud/reference/cli/influx/)         

   Dev Notes
    - If URL to the InfluxCLI is used, use the URL with "latest" in the title - assuming that the edge/cli team finished their ticket (https://github.com/influxdata/influx-cli/issues/394)    
    - Figma: https://www.figma.com/file/JfGYCOb7ZyQ8JHSqeu4fma/Onboarding-Exploration?node-id=1137%3A5331

PR #5131, opened 2022-07-20
feat: install dependencies page for CLI

### LINKED ISSUE #4765: firstMile: InfluxCLI Onboarding - Install Dependencies - MAC OS
ACs
    - Install Dependencies page should include three tabs underneath the "Install Dependencies" text specific to the user's operating system (MAC OS, WINDOWS, LINUX). See Figma
    - Depending on which OS tab the user clicks, the install information below will change accordingly. The selected Tab should highlight blue and the text should be white. 
    - For MAC OS, the next sub-step on the Install Dependencies page should be "Use Homebrew" with the copy code snippet "brew install influxdb-cli"
    - Text underneath the Homebrew code snippet should be "If you prefer to manually download and install the CLI package, follow our documentation."
    - The word "documentation" should be hyperlinked (https://docs.influxdata.com/influxdb/cloud/tools/influx-cli/)
    - The next sub-step in the MAC OS flow for Instal Dependencies should be "Useful InfluxCLI Commands"
    - Text underneath "Useful InfluxCLI Commands" should be "To invoke a command, use the following format in the command line: influx <command>" (See Figma)
    - In the "Useful InfluxCLI Commands" sub-step/section should be a table of useful commands and their definitions. (bucket, export, help, help <command>, query, and write)
    - Underneath the table of useful commands should be the following text and link "A full list of InfluxCLI Commands is available in our documentation." The word documentation should be hyperlinked to:(https://docs.influxdata.com/influxdb/cloud/reference/cli/influx/)        
   
   Dev Notes
   -  If URL to the InfluxCLI is used, use the URL with "latest" in the title - assuming that the edge/cli team finished their ticket (https://github.com/influxdata/influx-cli/issues/394)  
    -  Figma: https://www.figma.com/file/JfGYCOb7ZyQ8JHSqeu4fma/Onboarding-Exploration?node-id=1137%3A5331

alexjw3:
Closes #4765, #4766, #4767

Adds the install dependencies page on the new CLI walkthrough.

### Checklist

### DIFF
--- a/src/homepageExperience/components/steps/cli/InstallDependencies.tsx
+++ b/src/homepageExperience/components/steps/cli/InstallDependencies.tsx
@@ -56,8 +56,9 @@ sudo cp influxdb2-client-latest-linux-arm64/influx /usr/local/bin/
   `
 
   type CurrentOSSelection = 'Linux' | 'Mac' | 'Windows'
-  const [currentSelection, setCurrentSelection] = useState<CurrentOSSelection>('Mac')
-
+  const [currentSelection, setCurrentSelection] = useState<CurrentOSSelection>(
+    'Mac'
+  )
 
   return (
     <>
@@ -65,21 +66,33 @@ sudo cp influxdb2-client-latest-linux-arm64/influx /usr/local/bin/
       <ButtonGroup orientation={Orientation.Horizontal}>
         <Button
           text="mac OS"
-          color={currentSelection === 'Mac' ? ComponentColor.Primary : ComponentColor.Default}
+          color={
+            currentSelection === 'Mac'
+              ? ComponentColor.Primary
+              : ComponentColor.Default
+          }
           onClick={() => {
             setCurrentSelection('Mac')
           }}
         />
         <Button
           text="windows"
-          color={currentSelection === 'Windows' ? ComponentColor.Primary : ComponentColor.Default}
+          color={
+            currentSelection === 'Windows'
+              ? ComponentColor.Primary
+              : ComponentColor.Default
+          }
           onClick={() => {
             setCurrentSelection('Windows')
           }}
         />
         <Button
           text="linux"
-          color={currentSelection === 'Linux' ? ComponentColor.Primary : ComponentColor.Default}
+          color={
+            currentSelection === 'Linux'
+              ? ComponentColor.Primary
+              : ComponentColor.Default
+          }
           onClick={() => {
             setCurrentSelection('Linux')
           }}
@@ -100,9 +113,7 @@ sudo cp influxdb2-client-latest-linux-arm64/influx /usr/local/bin/
               documentation.
             </SafeBlankLink>{' '}
           </p>
-          <h2 style={headingWithMargin}>
-            Useful InfluxCLI commands
-          </h2>
+          <h2 style={headingWithMargin}>Useful InfluxCLI commands</h2>
           <p>
             To invoke a command, use the following format in the command line:
           </p>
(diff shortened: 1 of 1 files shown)


src/homepageExperience/components/steps/cli/InstallDependencies.tsx
```diff
@@ -0,0 +1,312 @@
...
+    <>
+      <h1>Install Dependencies</h1>
+      <ButtonGroup orientation={Orientation.Horizontal}>
+        <Button
+          text="mac OS"
+          color={macSelected ? ComponentColor.Primary : ComponentColor.Default}
+          onClick={() => {
+            setMacSelected(true)
+            setWindowsSelected(false)
+            setLinuxSelected(false)
```
hoorayimhelping, line 70: So this wil work, but it's a lot of state to accomplish the goal - there are three different flags that need to be set, and the value of one of them depends on the value of the other two - if `macSelected` is true, then `windowsSelected` and `linuxSelected` must both be false, and it's up to the coder to make sure this happens. These things tend to be hard for most people to keep in their head at once, and they end up being really easy to cause bugs in accidentally.

Rather than having a `isSelected` type flag for each option, let's instead have a single variable that tracks the currently selected item. There are three different options for which OS is selected: Mac, Linux, and Windows. If we called the variable `currentSelection` (or something similar, it doesn't have to be called this), and give it a limited range of values, we could accomplish the same thing with a single variable.

TypeScript has a feature called [string literal types](https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types) that fit this nicely. What we'd do is create a TypeScript type that only allows the narrow range of values we care about, and then set the value of our variable to one of those Typed values.

```ts
type CurrentOSSelection = 'Linux' | 'Mac' | 'Windows'
```
then when using the `useState` hook, we could do something like:

```ts
const [currentSelection, setCurrentSelection] = useState<CurrentOSSelection>('Mac')
```
Then, the `Buttons` become something like:

```ts
color={currentSelection === 'Mac' ? ComponentColor.Primay : ComponentColor.Default}
onClick={() => {
  setCurrentSelection('Mac')
}}
```
and the check for which part to render becomes:
```ts
{currentSelection === 'Mac' && (
...
```

src/homepageExperience/components/steps/cli/InstallDependencies.tsx
```diff
@@ -0,0 +1,312 @@
...
+  Table,
+  BorderType,
+} from '@influxdata/clockface'
+
+import React, {useEffect, useState} from 'react'
+import {useDispatch} from 'react-redux'
+import {getBuckets} from 'src/buckets/actions/thunks'
+import {event} from 'src/cloud/utils/reporting'
+
+export const InstallDependencies = () => {
```
hoorayimhelping, line 17:
```suggestion
export const InstallDependencies: FC = () => {
```
`FC` is a type exported by React. For example here: https://github.com/influxdata/ui/blob/master/src/App.tsx#L48

src/homepageExperience/components/steps/cli/InstallDependencies.tsx
```diff
@@ -0,0 +1,312 @@
...
+          onClick={() => {
+            setMacSelected(false)
+            setWindowsSelected(false)
+            setLinuxSelected(true)
+          }}
+        />
+      </ButtonGroup>
+      {macSelected && (
+        <>
+          <h2 style={{marginTop: '48px', marginBottom: '0px'}}>Use Homebrew</h2>
```
hoorayimhelping, line 98: since this style object literal is used in multiple places, it's probably best to pull it out into a variable that is defined *outside* the function component, so it's not redefined on every render call.

e.g.
```ts
const headingWithMargin = {marginTop: '48px', marginBottom: '0px'}
...
<h2 style={headingWithMargin}
```
the variable name is just an example. It doesn't need to be named `headingWithMargin`

hoorayimhelping (changes requested):
Component looks good, just a suggestion about simplifying the internal component state

src/homepageExperience/components/steps/cli/InstallDependencies.tsx
```diff
@@ -0,0 +1,301 @@
...
+import {event} from 'src/cloud/utils/reporting'
+
+export const InstallDependencies: FC = () => {
+  const dispatch = useDispatch()
+
+  useEffect(() => {
+    dispatch(getBuckets())
+  }, [])
+
+  const headingWithMargin = {marginTop: '48px', marginBottom: '0px'}
```
hoorayimhelping, line 24: let's move this outside the function component, so like:

```ts
const headingWithMargin = {marginTop: '48px', marginBottom: '0px'}

```

If it stays inside the component, it'll get re-created on each render and react won't be able to understand that it's the same object.

hoorayimhelping (approved):
LGTM!

One minor change, but it's not worth holding up the PR on that

PR #5132, opened 2022-07-20
refactor: don't use parse to check for valid query

rockstar:
This patch removes the use of `parse` from `flux-lsp` as a way to
determine if the query is valid. Instead, it opts for checking to see if
its length is zero. The issue with using `parse` here is that there is
no validity check; the queries `a = 0`, `b =` and `from(bucket: ` will
all return an ast with a non-zero body length.

### DIFF
--- a/src/flows/pipes/Visualization/index.ts
+++ b/src/flows/pipes/Visualization/index.ts
@@ -12,12 +12,8 @@ export default register => {
     initial: {
       properties: SUPPORTED_VISUALIZATIONS['xy'].initial,
     },
-    visual: (data, query) => {
+    visual: (data, query: String) => {
       if (!query) {
-        return ''
-      }
-
-      if (query.length === 0) {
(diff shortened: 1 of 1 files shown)


drdelambre:
seems like you're missing the case where the query is all comments (default case for raw flux editor panels)

original change was prompted by this: https://github.com/influxdata/ui/pull/2585

rockstar:
> If this is the direction we want to go down, we should have a discussion on what it is we want out of our UI with regards to allowing users to submit queries that might not be valid, and how we would like to deliver a good user experience without sacrificing performance

Exactly. When I first approached you about this patch, this is the kind of conversation I was hoping to start. This is less of a technical conversation and more of a philosophical/architectural conversation. I'm in no hurry to merge this patch, and think it better to have consensus across the team and with Product™ before pushing too hard ahead.[

PR #5146, opened 2022-07-21
feat: add orgname to salesforce requests. futureproof for quartzident…

### LINKED ISSUE #4585: Multi Org - Update help Bar logic to pass in the currently selected Org name and Org ID
Update Help Bar SFDC Ticket creation logic to pass in the currently selected Org Name and Org ID when paid users choose to submit a support ticket directly through the UI.

When paid users choose to submit a support ticket via the UI, our system generates an sfdc ticket. We want to pass the org name and the org ID in the ticket too. If sfdc already has an org name and org ID field, we can send these peices of meta-data to the correct sfdc field. Otherwise we can just auto-populate org name and org ID in the ticket body.

_Note: Bucky worked the original sfdc ticket generation capability in the help bar._

wdoconnell:
Closes #4585 

EDIT: Adjusted in response to @hoorayimhelping comments so that we're appending, not prepending, the `orgName` and `orgId`. 

Product requests that Help bar's SFDC ticket creation logic send the user's currently selected `orgName` and `orgId` to salesforce. 

This PR passes the user's `orgName` and `orgId` to salesforce as part of the description through `createSfdcSupportCase`. Related tweaks/inline refactors:

1. Pull `orgName` and `orgID` from quartz, not IDPE, since the IDPE endpoints are being deprecated. This is safe because this component is exclusively `CLOUD`.
2. Check feature flag status to decide whether to pull the data from `quartzMe` or `quartzIdentity`.

New format for the description portion of the ticket looks like this (edited - now appending org data at end):
Screen Shot 2022-07-22 at 11 15 04 AM

### Checklist

- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed). N / A
- [x] Feature flagged, if applicable  YES - `quartzIdentity` decides whether to pull user data from `/quartz/me` or `/quartz/identity`

### DIFF
--- a/src/support/components/PayGSupportOverlay.tsx
+++ b/src/support/components/PayGSupportOverlay.tsx
@@ -27,10 +27,12 @@ import {OverlayContext} from 'src/overlays/components/OverlayController'
 
 // Selectors
 import {getOrg} from 'src/organizations/selectors'
-import {getMe, getQuartzMe} from 'src/me/selectors'
+import {getQuartzMe} from 'src/me/selectors'
+import {selectQuartzIdentity} from 'src/identity/selectors'
 
 // Utils
 import {event} from 'src/cloud/utils/reporting'
+import {shouldUseQuartzIdentity} from 'src/identity/utils/shouldUseQuartzIdentity'
 
 // API
 import {createSfdcSupportCase} from 'src/shared/apis/sfdc'
@@ -63,9 +65,11 @@ interface OwnProps {
 }
 
 const PayGSupportOverlay: FC<OwnProps> = () => {
-  const {id: orgID} = useSelector(getOrg)
+  const quartzIdentity = useSelector(selectQuartzIdentity)
+  const {user: identityUser, org: identityOrg} = quartzIdentity.currentIdentity
+
   const quartzMe = useSelector(getQuartzMe)
-  const me = useSelector(getMe)
+  const quartzOrg = useSelector(getOrg)
 
   const [subject, setSubject] = useState('')
   const [severity, setSeverity] = useState('3 - Standard')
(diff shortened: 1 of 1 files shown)


src/support/components/PayGSupportOverlay.tsx
```diff
@@ -96,20 +100,34 @@ const PayGSupportOverlay: FC<OwnProps> = () => {
...
+      orgID = identityOrg.id
+    } else {
+      // Optional chaining operator needed because quartzMe's initial reducer state is null.
+      userID = quartzMe?.id
+      userEmail = quartzMe?.email
+      orgName = quartzOrg?.name
+      orgID = quartzOrg?.id
+    }
+
+    const descriptionWithOrgId = `[Org Name: ${orgName}] [Org Id: ${orgID}] \n\n ${description}`
```
wdoconnell, line 118: Added two newlines between the org name/id and description to make the ticket description read more cleanly.

hoorayimhelping: Vereesh asked me to put the org id at the end of the description. Has that changed?

src/support/components/PayGSupportOverlay.tsx
```diff
@@ -96,20 +100,34 @@ const PayGSupportOverlay: FC<OwnProps> = () => {
...
+      userEmail = identityUser.email
+      orgName = identityOrg.name
+      orgID = identityOrg.id
+    } else {
+      // Optional chaining operator needed because quartzMe's initial reducer state is null.
+      userID = quartzMe?.id
+      userEmail = quartzMe?.email
+      orgName = quartzOrg?.name
+      orgID = quartzOrg?.id
+    }
```
hoorayimhelping, line 116: Can you think of a way that this logic can be encapsulated so that we don't need to sprinkle it all over the codebase? Or is this maybe just a one-off situation?

src/support/components/PayGSupportOverlay.tsx
```diff
@@ -96,20 +100,34 @@ const PayGSupportOverlay: FC<OwnProps> = () => {
...
-        email,
+        userEmail,
         translatedSeverity,
         subject
       )
       event(
         'helpBar.paygSupportRequest.submitted',
         {},
-        {userID: me.id, orgID: orgID}
+        {userID: userID, orgName: orgName, orgID: orgID}
```
hoorayimhelping, line 130: minor style nit, but just use that good ole es6 object property shorthand

```suggestion
        {userID, orgName, orgID}
```

hoorayimhelping (approved):
Some minor stuff, but let me know if things need to be re-reviewed after a change

src/support/components/PayGSupportOverlay.tsx
wdoconnell: The clean way to do would be to just ignore identity and use the `quartzMe` state, since even when `identity` is on, `convertIdentityToMe` converts the identity shape into the shape of `quartzMe` so that it can be consumed throughout the app without lots of `if` blocks.

Unfortunately, we can't do that here because the _org id_ isn't returned by the `quartzMe` endpoint. So I think this is a brief 'one-off', especially since `quartzIdentity` is in prod. So this might be one of the last times we have to do it before the ff is off.

wdoconnell: Alternatively - we could just leave it using `quartzMe` and the idpe `org` endpoint for now. And worry about using identity later. That's less code.

hoorayimhelping (approved):

src/support/components/PayGSupportOverlay.tsx
wdoconnell: As discussed - I've reverted this to the existing location in the description until told otherwise.

PR #5152, opened 2022-07-22
fix: use optional chaining and fix dependency array for SimpleTable

TCL735:
Part of https://github.com/influxdata/giraffe/issues/781

- Use optional chaining in a few places to prevent a console error when the results are an empty data set (blank csv)
- Remove `ref.current` as a dependency because `ref` is mutable and has no effect on the hook, see https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom

NOTE: the above fixes will be implemented in Giraffe also.

### DIFF
--- a/src/visualization/types/SimpleTable/PagedTable.tsx
+++ b/src/visualization/types/SimpleTable/PagedTable.tsx
@@ -43,7 +43,7 @@ const measurePage = (
   let signature
 
   while (rowIdx < result.table.length) {
-    if (result.table.columns.table.data[rowIdx] !== currentTable) {
+    if (result.table.columns?.table?.data?.[rowIdx] !== currentTable) {
       signature = Object.values(result.table.columns)
         .map(
           c =>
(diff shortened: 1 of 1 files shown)


src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -43,7 +43,7 @@ const measurePage = (
   let signature
 
   while (rowIdx < result.table.length) {
-    if (result.table.columns.table.data[rowIdx] !== currentTable) {
+    if (result.table.columns?.table?.data?.[rowIdx] !== currentTable) {
```
TCL735, line 46: `currentTable` starts as `undefined` which we do not change. The left side can now be `undefined` due to optional chaining. This will cause the entire `if` block to be skipped, which is fine because `result.table` has no `columns to work with.

src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -110,26 +110,28 @@ const subsetResult = (
...
 
   const tables: SubsetTable[] = []
-  let lastTable
+  let lastTable = ''
 
   // group by table id (series)
   for (let ni = 0; ni < size; ni++) {
     if (
-      `y${subset['result'][0].data[ni]}:t${subset['table'][0].data[ni]}` ===
+      `y${subset['result']?.[0]?.data?.[ni]}:t${subset['table']?.[0]?.data?.[ni]}` ===
```
TCL735, line 118: `lastTable` is set to an empty string to make it clear that it's different than `undefined`.

The interpolated string `y${subset['result']?.[0]?.data?.[ni]}:t${subset['table']?.[0]?.data?.[ni]}` may be `yundefined:tundefined`. This looks bizarre and could be incorrectly interpreted in the future as intended to be `undefined`. This is the reason why we set an initial empty string value for `lastTable` above.

src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -269,7 +271,7 @@ const PagedTable: FC<Props> = ({result, properties}) => {
         clearTimeout(timeout)
       }
     }
-  }, [ref?.current])
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
```
TCL735, line 274: `ref.current` should not be a dependency. See https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom

Removing it leaves an empty dependency array, which means this hook runs only once. But he hook also has a cleanup function, so rest assured that the hook is cleaning up after itself when the SimpleTable is unmounted.

hoorayimhelping (approved):

PR #5158, opened 2022-07-22
feat: add types

helenosheaa:
Closes https://github.com/influxdata/data-acquisition/issues/454

We were only offering String or Number. Nifi's Avro Schema supports: string, boolean, int, float:

Screen Shot 2022-07-22 at 10 39 15 AM

new options:
Screen Shot 2022-07-22 at 11 01 44 AM

form** it does display Integer not int to the user this was just a bad screenshot:
Screen Shot 2022-07-22 at 11 12 30 AM

user json input:
Screen Shot 2022-07-22 at 10 31 46 AM

output:
Screen Shot 2022-07-22 at 11 22 01 AM

Screen Shot 2022-07-22 at 11 22 33 AM

subs that exist previously will be of type Number which under the hood is double... updated utils sanitizeType to catch this. Nifi supports both float or double as a value.

### Checklist


src/writeData/subscriptions/components/JsonParsingForm.tsx
```diff
@@ -45,8 +45,10 @@ interface Props {
 
 const JsonParsingForm: FC<Props> = ({formContent, updateForm, edit}) => {
   const stringType = 'String'
-  const numberType = 'Number'
-  const dataTypeList = [stringType, numberType]
+  const floatType = 'Float'
+  const intType = 'Integer'
```
genehynson, line 49: I'm trying to think through the value of adding `Integer` 🤔 Since JSON only supports "number", do we really need to specify whether the number in the message has a decimal or not? Or could we get away with using the avro "double" type that we've been using already? 

InfluxDB actually assumes that all numerical values are floats unless you append an `i` to the value: https://docs.influxdata.com/influxdb/v1.8/write_protocols/line_protocol_tutorial/#data-types

So I think even if we let the user choose between Float and Integer in the UI, all of those values would still be written as floats to InfluxDB at the end of the day

src/writeData/subscriptions/components/JsonPathInput.tsx
```diff
@@ -48,7 +48,11 @@ const JsonPathInput: FC<Props> = ({
   itemNum,
   edit,
 }) => {
-  const dataTypeList = ['String', 'Number']
+  const stringType = 'String'
+  const floatType = 'Float'
+  const intType = 'Integer'
+  const booleanType = 'Boolean'
```
genehynson, line 54: Could we put these in `utils` so we only declare them once? Defining them here means that each time the component is re-rendered they are redefined which uses more browser memory

src/writeData/subscriptions/utils/form.ts
```diff
@@ -99,8 +99,8 @@ export const sanitizeForm = (form: Subscription): Subscription => {
       if (newVal) {
         f.path = newVal
       }
-      if (f.type === 'number') {
-        f.type = 'double'
+      if (f.type === 'integer') {
+        f.type = 'int'
```
genehynson, line 103: totally a future PR but I'm thinking we could refactor all of these `if` statements a bit. We could create an object that has the display value and the avro value as properties and have the user select one of those objects in the dropdown. Then instead of all these if statements we could just have the onChange event of the dropdown set `type` to the avro value or something. But anyways, a future PR :)

helenosheaa: 100%

src/writeData/subscriptions/components/JsonParsingForm.tsx
genehynson: yeah I thinkkkkk the way it works is that the values are all _written_ as floats but a type is inferred when you query it back. Could you repeat the experiment of sending `1` and `1.2` using just `double` as the arvo type and see if notebooks shows the same result?

helenosheaa: based on [this](https://influxdata.slack.com/archives/C023R2B5KD3/p1658935448594759) conversation i think we should add in the int/float support to be inline with telegraf json v2 parsing

genehynson (approved):

PR #5149, opened 2022-07-21
fix(pinned): access pinned card from within container

blegesse-git:
This is an attempt to fix one of the flaky test in cloud's pinned dashboards test 

cypress seems to have a hard time finding `[data-testid="pinneditems--card"]` after navigating to the home page. 
Pr tried to solve problem by guiding cypress to find `pinneditems-card` through the `pinneditems--container`. 

Authors and Reviewer(s), please verify the following:

- [x] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [x] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [ ] Documentation updated or issue created (provide link to issue/PR)
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [ ] Feature flagged, if applicable

PR #5159, opened 2022-07-22
test(pinned): wait for element to be visible before accessing

blegesse-git:
related to #5149 

PR adds code to all individual tests in `Pinned.test.ts` to wait on UI elements to be visible before accessing said element. 

I ran each test at least 10 times to check for flaky behavior. No issues observed. 

### DIFF
--- a/cypress/e2e/cloud/pinned.test.ts
+++ b/cypress/e2e/cloud/pinned.test.ts
@@ -42,9 +42,11 @@ describe('Pinned Items', () => {
 
       cy.visit('/')
       cy.getByTestID('tree-nav')
-      cy.getByTestID('pinneditems--container').within(() => {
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):
You've already done great work looking at these tests, it pains me to say this: But if these keep causing problems, let's just disable these tests, since the service isn't on in production at the moment.

blegesse-git:
> You've already done great work looking at these tests, it pains me to say this: But if these keep causing problems, let's just disable these tests, since the service isn't on in production at the moment.

I haven't seen any more flakiness from pinned. Let me just merge this one and if the flakiness persists, we can disable like you said. But looks like there is another flaky test coming from `fluxQueryBuilder.test` 🥴 it's not letting my pr pass

PR #5172, opened 2022-07-25
feat: moving the errors up a bit

### LINKED ISSUE #5014: Error and help message display
When a user runs a query that results in an error, we need to display the error message in an easy-to-read location so it can be quickly read and resolved. 

Right now we've placed the error message below the query controls. Sometimes this message pushes the controls to be outside of the viewport which causes the user to have to reload the page to get those controls back.

Image

We'd like to move the error messaging up to right below the Monaco editor since the error is directly related to the script they've written. The error message should wrap so is doesn't close panels or move controls off the page. 

Also, we'd like to introduce a help message for scripts that produce no results because the metrics chosen don't included values within the chosen time range. 

Figma: https://www.figma.com/file/0qAntPk5LVangAHguWT74X/Query-Experience-Project?node-id=1509%3A57098

Design:
https://user-images.githubusercontent.com/81173353/178779257-57774a31-059a-4d56-9de8-76f1ab7f8690.mov

drdelambre:
closes #5014 

Screen Shot 2022-07-25 at 10 18 20 AM

Screen Shot 2022-07-25 at 10 23 27 AM

### DIFF
--- a/src/dataExplorer/components/QueryTime.scss
+++ b/src/dataExplorer/components/QueryTime.scss
@@ -27,10 +27,4 @@
       background: $c-rainforest;
     }
   }
-
-  pre {
-    color: $c-fire;
-    font-family: $cf-code-font;
-    margin: $cf-space-2xs;
-  }
 }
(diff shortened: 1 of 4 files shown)


appletreeisyellow (approved):
Nice!

Issue #4966, opened 2022-07-05
Bug: Typeahead Dropdown for Variables in Dashboards is unresponsive with large lists


ChitlangeSahas:
The typeahead dropdown is unresponsive for long lists. We need to use the updated clockface component to pull in the fix that handles the long lists better. 

https://influxdata.slack.com/files/UQZURJM96/F03NB4X400M/copy-paste-var.mp4

ChitlangeSahas:
@abalone23 Reported this. 👍

ChitlangeSahas:
My PR here has the fix, we just need to use that in the UI. 

https://github.com/influxdata/clockface/pull/795

ChitlangeSahas:
This is fixed.

PR #5194, opened 2022-07-26
refactor: use Clockface component for TypeAheadVariableDropdown

ChitlangeSahas:
Part of this: [#4966](https://github.com/influxdata/ui/issues/4966)

This is part of the ticket attached above. It uses the Clockface component instead of the ui code that is out of date. 

Video of it working: 

https://user-images.githubusercontent.com/18511823/181113779-d2ab6aa2-2b49-4950-8d83-8cdcb959adac.mov


src/variables/components/TypeAheadVariableDropdown.test.tsx
```diff
@@ -10,12 +10,12 @@ import {renderWithRedux} from 'src/mockState'
 import {AppState, RemoteDataState} from 'src/types'
 
 const values = {
+  always: 'always',
   def: 'defbuck',
   def2: 'defbuck2',
   foo: 'foobuck',
   goo: 'goobuck',
   new: 'newBuck',
```
ChitlangeSahas, line 18: Needs to be alphabetic.

src/variables/components/TypeAheadVariableDropdown.test.tsx
```diff
@@ -71,7 +71,7 @@ const setInitialState = (state: AppState): AppState => {
             values: {
               '03cbdc8a53a63000': {
                 values,
-                selected: ['defbuck'],
+                selected: [''],
```
ChitlangeSahas, line 74: The new behavior is to show only the selected item in the list, the whole list is shown after the input is cleared. I think this should be changed in the future as that isn't how dropdown's work.

hoorayimhelping: I agree with this take and I think that would be a positive UX addition. Adding to that, we might consider changing the behavior shown at 0:17 in the video - rather than closing the dropdown when removing the selected item, it might be better UX to keep it open and show the newly non-filtered list. This isn't something to block this PR on, mind. Just a suggestion for future work.

src/variables/components/TypeAheadVariableDropdown.tsx
```diff
@@ -19,403 +12,58 @@ import {selectValue} from 'src/variables/actions/thunks'
...
+    const selectedVariableValue = {
+      id: selectedValue,
+      name: selectedValue,
+    } as SelectableItem
     return (
-      <Dropdown
-        style={{width: '192px'}}
-        className="variable-dropdown--dropdown"
+      <TypeAheadDropDown
+        style={{width: '150px'}}
```
hoorayimhelping, line 42: If this style needs to be inline, rather than part of a stylesheet, can you move the definition of `{width: '150px'}` to outside of the component and use the variable here please?

hoorayimhelping (approved):

src/variables/components/TypeAheadVariableDropdown.test.tsx
ChitlangeSahas: I agree! I can add that in a future clockface PR 👍

cypress/e2e/shared/dashboardsVariables.test.ts
```diff
@@ -176,13 +176,13 @@ describe('Dashboard - variable interactions', () => {
 
           // Make sure typeAhead input box is rendered and is visible
           cy.getByTestID(
-            `variable-dropdown-input-typeAhead--${bucketVarName}`
+            `variable-dropdown--${bucketVarName}--typeAhead-input`
```
ChitlangeSahas, line 179: These are just testID changes, seem like a lot but basically it's just changing the testID order.

ChitlangeSahas:
Just need to wait for this to merge in, https://github.com/influxdata/clockface/pull/820

and then should be good to go 🎉

src/variables/components/TypeAheadVariableDropdown.test.tsx
```diff
@@ -87,50 +93,53 @@ describe('Dashboards.Components.VariablesControlBar.TypeAheadVariableDropdown',
...
 
     // clear input, should see everything:
-    checkDropdown('', values)
+    checkDropdown('', xkcd_name_to_url_map)
 
     // filter again by text:
-    checkDropdown('e', evalues)
-    checkDropdown('al', alvalues)
+    checkDropdown('b', values_with_b)
+    checkDropdown('c', values_with_c)
```
hoorayimhelping, line 136: Since this refactors the existing test this is fine. But I think we'll want to add some cases for searching more than a single letter to test that that things correctly show up (e.g. searching for `ao` to make sure that `Bad Code` shows up) and that things don't show up (e.g. searching for `azzzzzz` to make sure nothing shows up)

hoorayimhelping (approved):

src/variables/components/TypeAheadVariableDropdown.test.tsx
ChitlangeSahas: These suggestions have been added to the Clockface!

src/variables/components/TypeAheadVariableDropdown.test.tsx
ChitlangeSahas: Got it! I'll add it to the test!

src/variables/components/TypeAheadVariableDropdown.test.tsx
```diff
@@ -9,37 +9,43 @@ import TypeAheadVariableDropdown from 'src/variables/components/TypeAheadVariabl
...
+This is necessary because the TypeAheadDropDown component now uses react-window to render the list and,
+the list renders limited items at a time to optimize performance.
+
+Limiting the number of items rendered at a time helps performance,
+but it also means that the test can not target all items at once with getAllByTestId when testing whether the dropdown items are rendered correctly.
+*/
+
+// map variable values
+const xkcd_name_to_url_map = {
+  'Bad Code': 'https://xkcd.com/1926/',
```
wdoconnell, line 25: Nice.

src/variables/components/TypeAheadVariableDropdown.test.tsx
```diff
@@ -9,37 +9,43 @@ import TypeAheadVariableDropdown from 'src/variables/components/TypeAheadVariabl
...
-  foo: 'foobuck',
-  goo: 'goobuck',
-  new: 'newBuck',
-  always: 'always',
-  really: 'reallyYes',
-  REALLy2: 'anotherReally',
+// map variable name
+const variable_name = 'xkcd_meme'
+
+/* Note(sahas): Note that there are only 5 values in the variable for the purposes of this unit test.
```
wdoconnell, line 15: Thanks again - I think this is a useful comment because folks (who might not realize this is using react-window) may otherwise be confused later about why all dropdown items aren't being rendered.

wdoconnell (approved):

PR #5199, opened 2022-07-27
chore: checking and fixing eslint warnings on the onboarding wizard pages

### LINKED ISSUE #4900: eslint warnings relating to useEffect hook dependencies in onboarding wizard pages
## About the bug
This is a subset of issue #4815, which is being chunked into smaller, logical units so that it can be analyzed in greater detail by a few members of the team.

**Steps to reproduce:**

1. run `yarn eslint` in `ui` root directory
2. receive warnings from React hooks with missing dependencies in `homepageExperience` directories

**Expected behavior:**
Eslint warnings for missing dependencies either do not exist, or (where appropriate) are suppressed.

**Actual behavior:**
Eslint warnings like the following, as shown in the attached file. `warning  React Hook useEffect has a missing dependency: 'dispatch'. Either include it or remove the dependency array`

**Visual Proof:**
Attached sample `eslint` log.
[eslintlog.txt](https://github.com/influxdata/ui/files/8992674/eslintlog.txt)

## About your environment

**Environment info:**
- System info: Darwin 21.5.0 x86_64

alexjw3:
Investigated and fixed multiple eslint warnings on the onboarding pages. After investigating these, I come out with 4 questions that could possibly lead to more tickets being made:
1. How many suppressed eslint messages are there that actually could just have an unchanging variable in the array? Could this help with debugging later if something actually were wrong, as at least this warning wouldn't be suppressed? Said differently, what is the harm of adding the unchanging functions or variables into the hook just as default? (Refer here: https://github.com/facebook/react/issues/14920#issuecomment-471070149)
2. It seems as though that all the onboarding wizard pages reuse a lot of the same code. Is there a way that we can streamline it slightly better or pare down the code?
3. Specifically within the overview section, due to the nature of Bill's fix (read fix #4640) we event every time the video is unfocused and refocused, not just the first time it is clicked (unfocused and refocused in this context is clicking outside the video then back into the video). Should we only allow this to happen once?
4. On the initializeClient and Tokens page, we are using await when it seems to not be necessary. A warning is also being thrown, saying that await has no effect on this type of expression (dispatch). Should they be removed?

Closes #4900.

### Checklist


hoorayimhelping:
>Said differently, what is the harm of adding the unchanging functions or variables into the hook just as default?

I think we probably add functions to the dependency arrays. I think the ignoring it might be a case of misunderstanding the impelementation details how how hooks work (which is completely understandable - to use hooks effectively, you have to understand how react implements them, unfortunately).

>It seems as though that all the onboarding wizard pages reuse a lot of the same code. Is there a way that we can streamline it slightly better or pare down the code?

Yep, the way to do it is to refactor the code now that it's been in use for a while and has been reused and pushed beyond what it was originally designed to do.

>we event every time the video is unfocused and refocused, not just the first time it is clicked (unfocused and refocused in this context is clicking outside the video then back into the video). Should we only allow this to happen once?

I'm not sure what the expected behavior is or what is useful for Amy and the analytics team. But we should figure out what we need this to do and make it do that. Good find.

>On the initializeClient and Tokens page, we are using await when it seems to not be necessary. A warning is also being thrown, saying that await has no effect on this type of expression (dispatch). Should they be removed?

Seems like the `await`s should be removed if they have no affect

src/homepageExperience/components/steps/Finish.tsx
```diff
@@ -79,7 +79,7 @@ export const Finish = (props: OwnProps) => {
       props.markStepAsCompleted()
       fireConfetti()
     }
-  }, [])
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
```
wdoconnell, line 82: I would avoid this if possible - it should be okay to put the props and wizard information in the dependency array, since none of that should change. Try removing the comment, and then testing the page with all of the dependencies React flags for you.

src/homepageExperience/components/steps/Overview.tsx
```diff
@@ -23,7 +23,7 @@ export const Overview: FC<Props> = ({wizard}) => {
       }
     }
     window.addEventListener('blur', checkVidClick)
-  }, [])
+  }, [wizard])
```
wdoconnell, line 26: Good. This is exactly what we should be doing here.

src/homepageExperience/components/steps/Tokens.tsx
```diff
@@ -79,14 +79,14 @@ export const Tokens: FC<OwnProps> = ({
       dispatch(createAuthorization(authorization))
       event(`firstMile.${wizardEventName}.tokens.tokenCreated`)
     }
-  }, [sortedPermissionTypes.length])
+  }, [sortedPermissionTypes.length, tokenValue]) // eslint-disable-line react-hooks/exhaustive-deps
```
wdoconnell, line 82: If you add `tokenValue`, you should need to add `// eslint-disable-line react-hooks/exhaustive-deps`. Is that to resolve another missing dependency - like `wizardEventName`? It's totally fine to include those in the dep array, they won't change.

src/homepageExperience/components/steps/cli/InitializeClient.tsx
```diff
@@ -104,14 +104,14 @@ export const InitializeClient: FC<OwnProps> = ({
       dispatch(createAuthorization(authorization))
       event(`firstMile.${wizardEventName}.tokens.tokenCreated`)
     }
-  }, [sortedPermissionTypes.length])
+  }, [sortedPermissionTypes.length, tokenValue]) // eslint-disable-line react-hooks/exhaustive-deps
```
wdoconnell, line 107: Same comment here. Unless there's a reason I would add all missing dependencies to the array, and not do `// eslint-disable-line`

src/homepageExperience/components/steps/cli/InstallDependencies.tsx
```diff
@@ -9,18 +9,10 @@ import {
...
 } from '@influxdata/clockface'
 
-import React, {FC, useEffect, useState} from 'react'
-import {useDispatch} from 'react-redux'
-import {getBuckets} from 'src/buckets/actions/thunks'
+import React, {FC, useState} from 'react'
 import {event} from 'src/cloud/utils/reporting'
 
 export const InstallDependencies: FC = () => {
-  const dispatch = useDispatch()
```
wdoconnell, line 18: What's the reason for proposing removing this useEffect hook?

src/homepageExperience/components/steps/python/WriteData.tsx
```diff
@@ -46,7 +46,7 @@ export const WriteDataComponent = (props: OwnProps) => {
   useEffect(() => {
     setSelectedBucket(bucket)
     onSelectBucket(bucket.name)
-  }, [bucket])
+  }, [bucket, onSelectBucket])
```
wdoconnell, line 49: Final Q - for all of these, double check to make sure we're not still missing a dependency on any. (It would be annoying if you spent the time fixing these issues, and then there was one remaining on the same page!)

wdoconnell (changes requested):
Good progress - I would definitely err on the side of removing the linter comments (// ) and instead including dependencies in the array, unless it results in a bug on the page or some other issue with the logic.

If you run into issues when testing these without the comment removed, let me know!

wdoconnell:
> > Said differently, what is the harm of adding the unchanging functions or variables into the hook just as default?
> 
> I think we probably add functions to the dependency arrays. I think the ignoring it might be a case of misunderstanding the impelementation details how how hooks work (which is completely understandable - to use hooks effectively, you have to understand how react implements them, unfortunately).
> 
> > It seems as though that all the onboarding wizard pages reuse a lot of the same code. Is there a way that we can streamline it slightly better or pare down the code?
> 
> Yep, the way to do it is to refactor the code now that it's been in use for a while and has been reused and pushed beyond what it was originally designed to do.
> 

@alexjw3 Would you be able to take a look at the ones that use recycled code, and create an issue for it?  That's a great refactor to clean up repetitive code.

> > we event every time the video is unfocused and refocused, not just the first time it is clicked (unfocused and refocused in this context is clicking outside the video then back into the video). Should we only allow this to happen once?
> 
> I'm not sure what the expected behavior is or what is useful for Amy and the analytics team. But we should figure out what we need this to do and make it do that. Good find.
> 

@alexjw3 That's a totally good point, but I ended up resolving this one with Amy. Per @Luckette , the event should be generated "every time - Data analysts will be able to filter for unique clicks". I should have added Amy's response to my PR for document why we did it that way though. Let me do that now so it doesn't come up again.

> > On the initializeClient and Tokens page, we are using await when it seems to not be necessary. A warning is also being thrown, saying that await has no effect on this type of expression (dispatch). Should they be removed?
> 
> Seems like the `await`s should be removed if they have no affect

Yup - the `await` @alexjw3 showed me is just `await dispatch`, and since that's synchronous, the `await` wasn't doing anything. So safe to propose removing from that file.

src/homepageExperience/components/steps/Finish.tsx
```diff
@@ -69,30 +69,38 @@ const handleNextStepEvent = (wizardEventName: string, nextStepName: string) => {
...
-    if (!props.finishStepCompleted) {
-      event(`firstMile.${props.wizardEventName}.finished`)
-      props.markStepAsCompleted()
+    if (!finishStepCompleted) {
+      event(`firstMile.${wizardEventName}.finished`)
+      markStepAsCompleted()
       fireConfetti()
     }
-  }, [])
+  }, [finishStepCompleted, markStepAsCompleted, wizardEventName])
```
wdoconnell, line 92: This is kind of an interesting case that it's worth discussing, so let's pair up on it briefly. I believe this is the right result, but it's worth walking through the code a bit to talk about why it's all good to add `finishStepCompleted` here.

src/homepageExperience/components/steps/Tokens.tsx
```diff
@@ -62,11 +62,11 @@ export const Tokens: FC<OwnProps> = ({
   )
 
   useEffect(() => {
-    const fetchResources = async () => {
-      await dispatch(getAllResources())
+    const fetchResources = () => {
```
wdoconnell, line 65: This one is also interesting - and the issues with this `useEffect` hook have led us to a bug we need to fix in the wizards that generate tokens.

1. The `await` wasn't doing anything here, because there was no code following it in the hook.
2. Really, the bigger issue here is that dispatching the thunk middleware needs to have a `try/catch` block around it, or invoke the `.then()` and `.catch()` methods from the returned promise. `getAllResources` throws an error if it doesn't receive a `200` status code from the endpoint. Right now, that error is uncaught, so the UI will likely bug out on this page when the token request fails if we don't add error handling.

src/homepageExperience/components/steps/cli/InitializeClient.tsx
```diff
@@ -80,18 +80,18 @@ export const InitializeClient: FC<OwnProps> = ({
...
+  }, [dispatch])
 
   useEffect(() => {
     onSelectBucket(bucket.name)
   }, [bucket, onSelectBucket])
 
   useEffect(() => {
-    const fetchResources = async () => {
-      await dispatch(getAllResources())
+    const fetchResources = () => {
```
wdoconnell, line 90: Same as above - this led us to a nice bug we can squash in both instances where `getAllResources()` gets invoked.

src/homepageExperience/components/steps/cli/InitializeClient.tsx
```diff
@@ -104,14 +104,21 @@ export const InitializeClient: FC<OwnProps> = ({
       dispatch(createAuthorization(authorization))
       event(`firstMile.${wizardEventName}.tokens.tokenCreated`)
     }
-  }, [sortedPermissionTypes.length])
+  }, [
+    dispatch,
+    me.id,
+    org.id,
+    sortedPermissionTypes,
```
wdoconnell, line 111: This is likely good since `sortedPermissionTypes` is memoized, but let's talk about it.

wdoconnell (changes requested):
Looks good. This has helped identified some bugs in the wizards that we can track down and solve in this ticket.

src/homepageExperience/components/steps/Tokens.tsx
wdoconnell: @alexjw3 and I talked about this - this is a great one to talk about at code review.

src/homepageExperience/components/steps/cli/InitializeClient.tsx
wdoconnell: Talked about this in pairing - sticking with `sortedPermissionTypes.length` here may be sufficiently safe, since what we care about is whether or not the array has `length`.

wdoconnell:
@alexjw3 I'm going to re-run CI here before the midday CI/CD 👻 s start throwing us `502 bad gateway` errors 😺

PR #5200, opened 2022-07-27
fix(deepLinks): fix broken check against org reference

### LINKED ISSUE #5130: Flaky Test: UncaughtException in DeepLink code
Null check should be added to deepLinks.ts code for the buildDeepLinkingMap(org) function call.

If the org is null, have calling code re-login?  Attempt to retrieve org again?

At the moment, if org == null, then exception is thrown for org.id (unable to retrieve property id of null).
see https://app.circleci.com/pipelines/github/influxdata/monitor-ci/13616/workflows/6bf8e29d-1948-47f7-b4d6-99b3224f212e/jobs/836236 for example of failure.

hoorayimhelping:
Closes #5130

https://github.com/influxdata/ui/pull/4826 Changed `NotFound.tsx` from a class component to a function component, and in doing so, chaged what used to be an instance variable into a React `ref`. Unfortunately, it didn't update the check if `org` existed, and so the code that handles org not being present (in the case of logged-out users) never runs, causing OSS tests to fail and the feature to be broken for logged-out users.

### Checklist

### DIFF
--- a/cypress/e2e/shared/deepLinks.test.ts
+++ b/cypress/e2e/shared/deepLinks.test.ts
@@ -12,7 +12,6 @@ describe('Deep linking', () => {
   // so you'll probably need to follow-up with the docs and/or marketing teams.
   it('should be redirected to the approprate page from a shortened link', () => {
     cy.get('@org').then((org: Organization) => {
-      cy.wait(1000)
       cy.visit('/me/about')
       cy.location('pathname').should('eq', `/orgs/${org.id}/about`)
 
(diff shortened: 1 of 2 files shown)


cypress/e2e/shared/deepLinks.test.ts
```diff
@@ -12,7 +12,6 @@ describe('Deep linking', () => {
   // so you'll probably need to follow-up with the docs and/or marketing teams.
   it('should be redirected to the approprate page from a shortened link', () => {
     cy.get('@org').then((org: Organization) => {
-      cy.wait(1000)
```
hoorayimhelping, line 15: this `wait` was part of the feature flagging code, which waited after flipping the feature flag, but (erroneously) wasn't removed when the feature flag was removed

wdoconnell (approved):
Yassss 🎉

PR #5204, opened 2022-07-27
chore(billing): Remove Region from Billing Info

distortia:
In a multi-org/account world, instead of displaying the current organization's region. We have opted to remove the region altogether.

I have to update the OpenAPI repo to remove the `region` from `BillingInfo`

### Checklist

### DIFF
--- a/cypress/e2e/cloud/billing.test.ts
+++ b/cypress/e2e/cloud/billing.test.ts
@@ -88,8 +88,6 @@ describe('Billing Page PAYG Users', () => {
 
     // PAYG section
     cy.getByTestID('payg-plan--header').contains('Pay As You Go')
-    cy.getByTestID('payg-plan--region-header').contains('Region')
-    cy.getByTestID('payg-plan--region-body').contains('aws')
 
     cy.getByTestID('payg-plan--balance-header').contains('Account Balance')
     cy.getByTestID('payg-plan--balance-body').contains('10.00')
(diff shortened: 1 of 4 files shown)


distortia:
Updated the SHA from https://github.com/influxdata/openapi/commit/fae7b1adb67f961e5818ea62e2c02becaaca30cb and rebased

abshierjoel (approved):

wdoconnell (approved):
Looks good to me - I just gave it a shot in my remocal.

Note that on the organization settings page in the UI, we still have a field for the _cloud_ provider (as distinct from billing provider). Looks like this. And would also only reflect the region for the currently active org. 

Do we want to remove that too? If so LMK, and we'll open a UI ticket. 

Screen Shot 2022-07-29 at 12 40 35 PM

distortia:
We want to keep the provider and region on the Org settings page. We will have to update the SHA for OpenAPI changes and retrigger CI.

rockstar (commented):
To be frank, I don't know enough about the UI to check this for technical soundness. I would assume that if no tests break as a result of this patch, it's good to go.

distortia:
@rockstar I think @hoorayimhelping wanted us to make sure that https://github.com/influxdata/openapi/commit/3fdd5849c90ef9b7b062d8449616f1770b1a7b6b would not impact you and the team.

wdoconnell:
> We want to keep the provider and region on the Org settings page. We will have to update the SHA for OpenAPI changes and retrigger CI.

@distortia Thanks - we'll leave the org page as-is then.

wdoconnell:
@distortia Does this need to be merged?

wdoconnell:
FYI, seeing no disagreement, we'll plan to merge this in. Other UI tickets need access to newer openAPI SHAs that cannot be used until this is merged in.

PR #5230, opened 2022-07-28
fix: triming text

drdelambre:
looks like we had some updates to the api logic in idpe around labels and now our label tests are failing. only thing i can see out of place is that the trimming logic changed slightly

new validation rule created here:
https://github.com/influxdata/idpe/pull/15252/files#diff-b587b520be75ba4cd19694bbe99c89ad0af5d8c8dc9b53a3ced1423252c2a396R275


genehynson:
Looks like it's still somehow looking for a label that's not trimmed?
```
'(Δ4) Per attributum intelligo id quod intellectus de substantia percipit tanquam ejusdem essentiam constituens. ' 
```

drdelambre:
ooo.. you were writing the comment as a pushed up the change

TCL735 (approved):

PR #5232, opened 2022-07-28
test: fix flaky test in flux query builder

### LINKED ISSUE #5171: Flaky Test: fluxQueryBuilder detached DOM element
Reference pipeline:
https://app.circleci.com/pipelines/github/influxdata/monitor-ci/13760/workflows/cf6740d0-8fad-4e73-af1c-7573f17a2e9a/jobs/850970

Error:
> CypressError: Timed out retrying after 10000ms: "cy.trigger()" failed because this element is detached from the DOM.
> "<button class="tag-selector-value--load-more-button" data-testid="tag-selector-value--load-more-button">+ Load ...</button>"
> Cypress requires elements be attached in the DOM to interact with them.
> The previous command that ran was:
>   "cy.should()"
> This DOM element likely became detached somewhere between the previous and current command.

Fix:
move the commands on lines 138-160 into a .then() so the functions on those lines will wait for click() to finish rendering.

appletreeisyellow:
Closes #5171 

This PR fixes a flaky `oss-e2e` test in flux query builder. 

~~The flakiness happens when the test failed to find "Load more" button DOM. Moving that testing block into a `then()` function ensures the test on "Load more" button run after the tag value list is loaded~~

The flakiness happens when cypress tries to click on "Load more" button to show more tag values, but the tag value list becomes un-expanded, so "Load more" button DOM disappeared. It's super wired and I cannot find why cypress behaves in this way.

To resolve the flakiness, I updated the line protocol so that we can test "Load more" button for fields instead of for tag values. 

### Checklist

~- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)~
~- [ ] Feature flagged, if applicable~


drdelambre (approved):
good luck! i think you need to remove this skip before merging though
https://github.com/influxdata/ui/pull/5232/files#diff-0cf13a24a819e43dc4ac9ed3657ea59098aab2a08b44d386862aa07333071387R3

appletreeisyellow:
good catch 🤦‍♀️

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -12,7 +12,9 @@ describe.skip('FluxQueryBuilder', () => {
           }).then(() => {
             const writeData = []
             for (let i = 0; i < 30; i++) {
-              writeData.push(`ndbc,air_temp_degc=${i}_degrees station_id=${i}`)
+              writeData.push(
+                `ndbc,air_temp_degc=70_degrees station_id_${i}=${i}`
```
appletreeisyellow, line 16: I updated the line protocol here to have *one* tag value and *30* fields

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -82,36 +87,30 @@ describe.skip('FluxQueryBuilder', () => {
...
+      // search a field, should contain only the field, no tag keys
       cy.getByTestID('field-tag-key-search-bar')
         .should('be.visible')
         .type(searchField)
       cy.getByTestID('field-selector')
         .should('be.visible')
         .should('not.contain', 'Loading')
         .should('not.contain', 'No Fields Found')
-      cy.getByTestID('tag-selector--key')
+      cy.getByTestID('tag-selector')
```
drdelambre, line 102: i was having issues here with it not being able to find the Loading text because the tags that show up had a Loading div somewhere in their markup that was "hidden" but cypress was finding them as visible.

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -82,36 +87,30 @@ describe.skip('FluxQueryBuilder', () => {
...
-
+      cy.getByTestID('field-tag-key-search-bar').clear()
       cy.getByTestID('field-selector')
         .should('be.visible')
         .should('not.contain', 'Loading')
         .should('not.contain', 'No Fields Found')
-      cy.getByTestID('tag-selector--key')
-        .should('be.visible')
-        .should('not.contain', 'Loading')
-      cy.getByTestID('tag-selector-key')
```
drdelambre, line 114: this is different than `--key`

src/dataExplorer/components/TagSelector.tsx
```diff
@@ -115,13 +115,16 @@ const TagValues: FC<Prop> = ({loading, tagKey, tagValues}) => {
       <Accordion className="tag-selector-value">
         <div
           onClick={() => handleSelectTagKey(tagKey)}
-          data-testid="tag-selector--key"
+          data-testid="tag-selector-value--header-wrapper"
```
drdelambre, line 118: def a better name, but the tests should key off of this div and not the `tag-selector-value` iirc

cypress/e2e/shared/fluxQueryBuilder.test.ts
drdelambre: how come?

appletreeisyellow: @drdelambre the old line protocol was generating *30* tag values and *one* field, so we were testing "Load more" button under tag values. Cypress was super wired when it 'click()' on "Load more" button, the tag value list got un-expanded, so "Load more" button DOM disappeared and the test failed at this point.

Then I thought to test the "Load more" button that is under fields to walk around the wired behavior. That's why I updated the line protocol

src/dataExplorer/components/TagSelector.tsx
appletreeisyellow: @drdelambre i think the test does key off `tag-selector-value--header-wrapper` and not the `tag-selector-value`. is this what you mean?

drdelambre (approved):
I don't understand how the change here fixes the test, and it goes against some of the stuff I was seeing around waiting for loading states to appear then resolve and hidden divs being picked up by cypress when I was poking around in here, but you've been showing extreme ownership over this, so i trust your call here and my lack of imagination shouldn't be a blocker

PR #5238, opened 2022-07-28
feat(operator): eliminate the Remove User button

randycoulman:
**CLOUD ONLY**

The Remove User button on the operator Accounts page actually removes the user from an organization, not an account. When there was a 1:1 mapping between accounts and orgs, this was fine, but we are now moving into a multi-org world where this no longer makes sense.

We need to do more significant work to the operator UI to support multi-org, and this work will happen in a later phase of the project.

For now, given that there is a feature in the main part of the UI where users can remove themselves or other users from an organzation, we've decided to remove the `Remove User` button.

This is the only client of `ResourcesTableRow` that passed additional `children`, and I believe that ability was added as a bit of hack to support this one use case. Therefore, I also removed the `children` prop from `ResourcesTableRow` as dead code. Should we need it again in the future, it should be relatively easy to add back.

PR to remove the corresponding endpoint is at https://github.com/influxdata/openapi/pull/455.

### Checklist


hoorayimhelping (approved):
Fantastic description. I really appreciate that the explanation of the PR is almost as long as the PR itself!

>This is the only client of ResourcesTableRow that passed additional children, and I believe that ability was added as a bit of hack to support this one use case. Therefore, I also removed the children prop from ResourcesTableRow as dead code.

very nice!

PR #5240, opened 2022-07-29
fix: calculate the correct heights for the header and rows in SimpleTable

### LINKED ISSUE #5239: SimpleTable: rows of data are hidden in overflow and cannot be viewed
https://user-images.githubusercontent.com/10736577/181654313-2fc4c16d-60db-4279-8aa4-f975f378da5c.mp4

TCL735:
Closes #5239  

- Instead of using old and incorrect values for the header and row heights of the SimpleTable, calculate them.
- This allows SimpleTable to make all rows visible without relying on vertical scrolling.
- **_There must be enough vertical space to fit at least 25% of the height of the next row for that row to render._**

https://user-images.githubusercontent.com/10736577/181658050-05cc838a-e5a1-4823-95e4-89a8c8eaeb32.mp4

### DIFF
--- a/src/visualization/types/SimpleTable/PagedTable.tsx
+++ b/src/visualization/types/SimpleTable/PagedTable.tsx
@@ -6,6 +6,7 @@ import React, {
   useEffect,
   useState,
 } from 'react'
+import {nanoid} from 'nanoid'
 import {DapperScrollbars} from '@influxdata/clockface'
 import {FluxDataType} from '@influxdata/giraffe'
 import {
@@ -221,9 +222,6 @@ interface Props {
   result: FluxResult['parsed']
 }
 
-const pagedTableHeaderId = 'pagedTableRowId'
-const pagedTableBodyId = 'pagedTabledBodyId'
-
 const PagedTable: FC<Props> = ({result, properties}) => {
   const {
     offset,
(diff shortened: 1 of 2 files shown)


src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -232,8 +234,31 @@ const PagedTable: FC<Props> = ({result, properties}) => {
...
   } = useContext(PaginationContext)
   const [height, setHeight] = useState(0)
+  const [headerHeight, setHeaderHeight] = useState(0)
+  const [rowHeight, setRowHeight] = useState(0)
   const ref = useRef()
 
+  if (headerHeight === 0) {
+    const calculatedHeaderHeight =
+      document.querySelector<HTMLElement>(`#${pagedTableHeaderId}`)
+        ?.offsetHeight ?? 0
```
hoorayimhelping, line 244: Something to consider when using `querySelector` vs `ref`s: https://stackoverflow.com/a/69260293/92446. Not sure how feasible it is to use refs here, or if one is better than the other.

src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -274,8 +299,8 @@ const PagedTable: FC<Props> = ({result, properties}) => {
   }, []) // eslint-disable-line react-hooks/exhaustive-deps
 
   const size = useMemo(() => {
-    return measurePage(result, offset, height)
-  }, [result, offset, height])
+    return measurePage(result, offset, height, headerHeight, rowHeight)
+  }, [result, offset, height, headerHeight, rowHeight])
```
hoorayimhelping, line 303: should `measurePage` be included in the dependency array?

src/visualization/types/SimpleTable/PagedTable.tsx
```diff
@@ -285,8 +310,8 @@ const PagedTable: FC<Props> = ({result, properties}) => {
   }, [size]) // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
-    setMaxSize(measurePage(result, 0, height))
-  }, [height, result]) // eslint-disable-line react-hooks/exhaustive-deps
+    setMaxSize(measurePage(result, 0, height, headerHeight, rowHeight))
+  }, [result, height, headerHeight, rowHeight]) // eslint-disable-line react-hooks/exhaustive-deps
```
hoorayimhelping, line 314: should `setMaxSize` and `measurePage` be included in the dependency array?

hoorayimhelping (approved):
Nice find; Works for me, but a couple of comments about implementation details

src/visualization/types/SimpleTable/PagedTable.tsx
TCL735: Yeah, `useRef` is easier. i wouldn't need `ids` at all even though this calculation is done only once. It's a minor improvement, but still an improvement!

src/visualization/types/SimpleTable/PagedTable.tsx
TCL735: No, the linter doesn't complain about `measurePage`. There is only one `measurePage` in the whole module, as it is a constant. That's probably the reason why the linter doesn't complain about it.

src/visualization/types/SimpleTable/PagedTable.tsx
TCL735: See above for `measurePage`.

For `setMaxSize`, it's not needed since `setMaxSize` is the setter from a `useState`. As far as I am aware, the setter doesn't change and would make no difference in a dependency array.

PR #5241, opened 2022-07-29
test: e2e testing for CLI and firstMile

### LINKED ISSUE #4776: firstMile: InfluxCLI Onboarding - End to End Testing

david-rusnak:
Closes #4776

 - Added e2e tests for the CLI onboarding flow and updated the firstMile e2e tests, now both use regex for navigation to avoid flakiness. 
 - Added test ids to the next/previous buttons on the python flow. 
 - Also updated the spacing on the names for the subnav on the onboarding wizards.

### Checklist

### DIFF
--- a/cypress/e2e/shared/firstmile.tests.ts
+++ b/cypress/e2e/shared/firstmile.tests.ts
@@ -1,85 +1,86 @@
-describe('First mile experience', () => {
-  beforeEach(() => {
-    cy.flush()
-    cy.signin()
-    cy.setFeatureFlags({firstMile: true})
-  })
-
-  it('should load the main home-page that has the language tiles', function() {
-    cy.getByTestID('home-page--header').should('exist')
-    cy.getByTestID('language-tiles--scrollbox').should('exist')
-  })
-
-  describe('Python', () => {
+for (let i = 0; i < 20; i++) {
+  describe('First mile experience', () => {
     beforeEach(() => {
-      cy.getByTestID('homepage-wizard-language-tile--python').click()
-    })
-    it("runs through the python's wizard pages", function() {
-      cy.getByTestID('subway-nav').should('exist')
-
-      // first page is overview
-      cy.contains('Hello, Time-Series World!')
-
-      // second page is install dependencies
-      cy.contains('Next').click()
-      cy.contains('Install Dependencies')
-
-      // third page is create token
-      cy.contains('Next').click()
-      cy.contains('Create a Token')
-
-      // fourth page is initalize client
-      cy.contains('Next').click()
-      cy.contains('Initalize Client')
-
-      // fifth page is write data
-      cy.contains('Next').click()
-      cy.contains('Write Data')
-
-      // sixth page is execute query
-      cy.contains('Next').click()
-      cy.contains('Execute a Flux Query')
-
-      // seventh page is execute aggregate query
-      cy.contains('Next').click()
-      cy.contains('Execute an Aggregate Query')
-
-      // eight page is finish
-      cy.contains('Next').click()
-      cy.contains('Congrats!')
+      cy.flush()
+      cy.signin()
+      cy.setFeatureFlags({firstMile: true})
     })
 
-    describe('Subway Nav Bar', () => {
-      it('navigates to the correct page when the respective navigation button is clicked', function() {
-        cy.get('h1').contains('Hello, Time-Series World!')
+    it('should load the main home-page that has the language tiles', function() {
+      cy.getByTestID('home-page--header').should('exist')
+    })
 
-        cy.contains('Previous').should('be.disabled')
+    describe('Python', () => {
+      beforeEach(() => {
+        cy.getByTestID('homepage-wizard-language-tile--python').click()
+      })
+      it("runs through the python's wizard pages", function() {
+        cy.getByTestID('subway-nav').should('exist')
 
-        cy.contains('Install Dependencies').click()
-        cy.get('h1').contains('Install Dependencies')
+        // first page is overview
+        cy.contains('Hello, Time-Series World!')
 
-        cy.contains('Previous').should('not.be.disabled')
+        // second page is install dependencies
+        cy.getByTestID('python-next-button').click()
+        cy.contains('Install Dependencies')
 
-        cy.contains('Create a Token').click()
-        cy.get('h1').contains('Create a Token')
+        // third page is create token
+        cy.getByTestID('python-next-button').click()
+        cy.contains('Tokens')
 
-        cy.contains('Initialize Client').click()
-        cy.get('h1').contains('Initialize Client')
+        // fourth page is initalize client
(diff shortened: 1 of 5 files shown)


cypress/e2e/shared/firstmile.tests.ts
```diff
@@ -7,7 +7,6 @@ describe('First mile experience', () => {
 
   it('should load the main home-page that has the language tiles', function() {
     cy.getByTestID('home-page--header').should('exist')
```
blegesse-git, line 9: we will need to add `cy.getByTestID('language-tiles--scrollbox').should('exist')` back. Since it applies to the first mile flow, it makes sense to include it in the test.

blegesse-git: @ChitlangeSahas where is this test id coming from? David had to remove it because the test fails locally because cypress can't find it

cypress/e2e/shared/firstmile.tests.ts
```diff
@@ -21,64 +20,64 @@ describe('First mile experience', () => {
       cy.contains('Hello, Time-Series World!')
 
       // second page is install dependencies
-      cy.contains('Next').click()
```
hoorayimhelping, line 24: what's the intent behind changing this from `contains('Next')` to `getByTestID('python-next-button')`?

- The `contains('Next')` was deliberate - it's more akin to a user's behavior than a machine's behavior, and it tests behavior more than implementation. The user will click the Next button. A machine will find an element in the DOM by test id and click the button matching that. The implementation of the feature is to have a specific test id, whereas the behavior of the feature is to advance the page when a user clicks Next. By targeting test ids when we could target the Next button, we more tightly couple the test to the implementation of the DOM.

cypress/e2e/shared/firstmile.tests.ts
```diff
@@ -21,64 +20,64 @@ describe('First mile experience', () => {
...
 
     describe('Subway Nav Bar', () => {
       it('navigates to the correct page when the respective navigation button is clicked', function() {
         cy.get('h1').contains('Hello, Time-Series World!')
 
-        cy.contains('Previous').should('be.disabled')
+        cy.getByTestID('python-prev-button').should('be.disabled')
 
-        cy.contains('Install Dependencies').click()
+        cy.contains(/^Install Dependencies$/).click()
```
hoorayimhelping, line 57: Do you know why searching for the string 'Install Dependencies' is flaky but searching for the regex `^Install Dependencies$`?

Putting `^Some Text$` into https://regex101.com/ (a great tool if you're not familiar) tells us:

```
^ asserts position at start of a line
Some Text
 matches the characters Some Text literally (case sensitive)
$ asserts position at the end of a line
```

Do you know why the text without `^` and `$` flakes out?

src/homepageExperience/utils.ts
```diff
@@ -6,27 +6,27 @@ export const HOMEPAGE_NAVIGATION_STEPS = [
     glyph: IconFont.BookOutline,
   },
   {
-    name: 'Install\nDependencies',
+    name: `Install\n Dependencies`,
```
hoorayimhelping, line 9: since there's no templating happening in this string (i.e. no `${variable}`) it's better to use single or double quotes (single quotes in our case since that's our standard for strings). Reserving backticks `` ` `` for string interpolation means that when you see a string using `` ` ``, you can be pretty sure it's going to have a variable being output in it.

src/homepageExperience/utils.ts
```diff
@@ -6,27 +6,27 @@ export const HOMEPAGE_NAVIGATION_STEPS = [
...
     glyph: IconFont.Install,
   },
   {
     name: 'Tokens',
     glyph: IconFont.CopperCoin,
   },
   {
-    name: 'Initialize\nClient',
+    name: 'Initialize\n Client',
```
hoorayimhelping, line 17: Why the space? I think we decided that not having a space was preferred here, since these are separated by a new line. Having a trailing or leading space, while not the end of the world, might cause issues for screen readers and will show up when copy/pasting highighted words. It might also affect string literals when searching or in e2e tests

hoorayimhelping (changes requested):
Thanks for getting new e2e tests in!

I have a couple of questions about this. I might be lacking context here, so let me know when you answer them!

cypress/e2e/shared/firstmile.tests.ts
ChitlangeSahas: from what I remember this was for :

Screen Shot 2022-07-29 at 8 52 27 AM

does it give you anything if you global search for `home-page--header`?

cypress/e2e/shared/firstmile.tests.ts
david-rusnak: So what ended up happening was there is one page in the CLI onboarding that has text somewhere along the lines of `Next, we will write data` or similar, so the test tried clicking on that instead of the `Next` button, causing it to fail every time on that page, so I changed it on every section where it clicks on that button.

cypress/e2e/shared/firstmile.tests.ts
blegesse-git: ooh sorry i mean `language-tiles--scrollbox` <- this test id

cypress/e2e/shared/firstmile.tests.ts
david-rusnak: Again something similar happened to above. When clicking on the `Tokens` step in the python wizard, for some reason, using `cy.contains('Tokens')` wanted to click on `API Tokens` in the other nav bar. So changing it to regex fixed that one, but I changed that on all of them so that anything similar to that wouldn't happen again.

src/homepageExperience/utils.ts
david-rusnak: Good catch, I forgot to change that after I tested something when creating the tests!

src/homepageExperience/utils.ts
david-rusnak: So when copying and pasting without the space, it pasted as `InstallDependencies` with no space. But if that space will mess with screen readers, it probably is better to remove the space since I doubt users will actually copy and paste those nav items often if at all.

cypress/e2e/shared/firstmile.tests.ts
hoorayimhelping: Big sad

cypress/e2e/shared/firstmile.tests.ts
hoorayimhelping: Good thinking, ole boy!

src/homepageExperience/utils.ts
hoorayimhelping: This is a very minor nitpick, and if it copy/pasted wrong, then it's fine to leave it as it is :)

hoorayimhelping (approved):
thanks for the solid responses

PR #5246, opened 2022-07-29
feat(DocSearch): add poc link under Useful Links

### LINKED ISSUE #5120: Request a POC - Useful Links
ACs
    - Add an option to the Useful Links list called "Request a Proof of Concept"
    - If a user clicks "Request a Proof of Concept" in the Useful Links section, send them here: https://www.influxdata.com/proof-of-concept/
    - Add eventing to track when users click the 'request a POC' option via the Useful Links
    - Manage with a feature flag
    - Keep behind a feature flag until PM or Engineering Management gives the green light to flip on in prod
    - Keep the feature flag available since this is a growth campaign

blegesse-git:
Closes #5120 

Adds `Request Proof of Concept` under Useful Links of UI homepage. 
behind `requestpoc` feature flag. 
Screen Shot 2022-07-29 at 11 56 12 AM

### Checklist

### DIFF
--- a/src/me/components/DocSearchWidget.tsx
+++ b/src/me/components/DocSearchWidget.tsx
@@ -34,6 +34,11 @@ const supportLinks = [
       'https://github.com/influxdata/influxdb/issues/new?template=feature_request.md',
     title: 'Feature Requests',
   },
+  {
+    link:
+      'https://www.influxdata.com/proof-of-concept/',
+    title: 'Request Proof of Concept',
+  },
 ]
 
 const DocSearchWidget: FC = () => {
--- a/src/me/components/DocSearchWidget.tsx
+++ b/src/me/components/DocSearchWidget.tsx
@@ -13,7 +13,7 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 import 'src/me/components/DocSearchWidget.scss'
 
-let supportLinks = [
+const supportLinks = [
   {
     link: `https://docs.influxdata.com/influxdb/${DOCS_URL_VERSION}/query-data/get-started/`,
     title: 'Get Started with Flux',
--- a/src/me/components/DocSearchWidget.tsx
+++ b/src/me/components/DocSearchWidget.tsx
@@ -4,12 +4,12 @@ import React, {FC} from 'react'
 // Components
 import DocSearch, {DocSearchType} from 'src/shared/search/DocSearch'
 
-// Utils
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
-
 // Constants
 import {CLOUD} from 'src/shared/constants'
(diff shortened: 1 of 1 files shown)


src/me/components/DocSearchWidget.tsx
```diff
@@ -34,8 +40,18 @@ const supportLinks = [
...
+    title: 'Request Proof of Concept',
+  },
 ]
 
+if (CLOUD && !isFlagEnabled('requestPoc')) {
+  supportLinks = supportLinks.filter(
+    item => item.title !== 'Request Proof of Concept'
+  )
+}
+
```
hoorayimhelping, line 54: I think it might be a little bit clearer to remove the `Request Proof of Concept` object from the `supportLinks` array and then do something like this:

```ts
if (CLOUD && !isFlagEnabled('requestPoc')) {
  supportLinks.push({
    link: 'https://www.influxdata.com/proof-of-concept/',
    title: 'Request Proof of Concept',
  })
}
```

What do you think? It feels a little more natural to me to add something under certain conditions rather than adding something then immediately removing them under certain conditions.

hoorayimhelping (commented):
Thoughts on this suggestion?

src/me/components/DocSearchWidget.tsx
blegesse-git: yeah, that's a fair point. I think I opted out for filtering out poc object after it's added to make it easier for clean up. Whenever we decide to remove the `requestPoc` feature flag, all we will have to do is remove line 49-53 and leave the `supportLinks` object as is.

blegesse-git: But considering how small of a task this is in general, let's go with your suggestion.

hoorayimhelping: >I think I opted out for filtering out poc object after it's added to make it easier for clean up. Whenever we decide to remove the `requestPoc` feature flag, all we will have to do is remove line 49-53 and leave the `supportLinks` object as is. 

That's fair reasoning. And I know you said let's go with my suggestion, but I feel an explanation still might help. I think this falls under YAGNI - You Ain't Gonna Need It. We should write the code for the problem we're solving now, not the one in the future. In general, we'll know more in the future than we know now, and things in the future will change - "always in motion is the future". Writing code to solve future problems that may never arrive results in code that can end up confusing and hard to follow. It can result in code that tends to give the impression that the author knew something in the past more than we know now, which makes the code scary to change. Now like you said, this is a small change, but I think the reasoning is applicable to a lot of code that we write :)

hoorayimhelping (approved):

src/me/components/DocSearchWidget.tsx
blegesse-git: wow, that's definitely something I didn't consider. To summarize - writing code with preconceived notions can potentially lead to unexpected problems. Makes perfect sense.

Issue #2068, opened 2021-07-19
Bug: SimpleTable - Remove `useGiraffeGraphs` flag


hoorayimhelping:
It's been on in Prod with no complaints. We should remove it.

### Steps
1. Remove all branching code referencing the flag
1. Remove flag from flags.yml in OSS and Cloud
1. Remove flag from configcat

TCL735:
This is pending the resolution of SimpleTable's implementation being re-ported back to Giraffe.

TCL735:
First step is to upgrade Giraffe to the version that has the latest SimpleTable. The PR is here: https://github.com/influxdata/ui/pull/5249

TCL735:
More updates have been made and Giraffe has been updated accordingly. Flag is currently ON in production. We will continue to monitor for issues.

Flag is slated for removal during the week of August 15 - 19.

PR #5249, opened 2022-07-30
chore: update Giraffe in preparation for removing feature flag

TCL735:
The first step of #2068 

identical

Old Data Explorer:

https://user-images.githubusercontent.com/10736577/182497586-443ef8e8-b498-4c6d-9c21-17307922c7e4.mp4

New Data Explorer:

https://user-images.githubusercontent.com/10736577/182497142-0511cd00-4bf6-4450-85a3-cd60691f9e90.mp4


src/visualization/types/SimpleTable/view.tsx
```diff
@@ -8,27 +8,23 @@ import {FluxResult} from 'src/types/flows'
...
-
-import './style.scss'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
+// Components
 import PageControl from 'src/visualization/types/SimpleTable/PageControl'
 import PagedTable from 'src/visualization/types/SimpleTable/PagedTable'
 import {PaginationProvider} from 'src/visualization/context/pagination'
 
+import './style.scss'
```
hoorayimhelping, line 18: We might consider renaming this from `style.scss` (search the codebase for `style.scss` and see how many there are) to something like `SimpleTableStyle.scss`, and making its path absolute rather than relative. It'll be much easier to work with if we can search for import path, rather than just `style.scss`.

hoorayimhelping (approved):

TCL735:
> Is the parseFromFluxResults still necessary in the other table?

We will address this as a separate issue since the other table is a separate visualization.

TCL735:
e2e test fixes are still pending. Currently there is another issue that is more urgent.

src/visualization/types/SimpleTable/view.tsx
TCL735: Done.

PR #5251, opened 2022-08-01
feat: sync quartzOrganizations thunks to new API layer for new quartz organizations routes

wdoconnell:
Part of [#4049](https://github.com/influxdata/ui/issues/4049)

UI PR [#5197](https://github.com/influxdata/ui/pull/5197) added API layer support for new backend routes that can (1) retrieve the orgs associated with any given account id and (2) set the default org for any given account. 

This PR makes corresponding updates to the quartzOrganizations thunks to use these new routes, as the old ones will be deprecated. The new user profile page is based on these updates.

This PR also improves error handling in the thunk for setting a new default org, distinguishing for reporting purposes whether the failure resulted from a state management problem or a network error. 

### Checklist

- [X] Feature flagged, if applicable - FLAGGED BEHIND `quartzIdentity` and `multiOrg`

### DIFF
--- a/src/identity/quartzOrganizations/actions/thunks/index.ts
+++ b/src/identity/quartzOrganizations/actions/thunks/index.ts
@@ -29,7 +29,7 @@ interface UpdateOrgParams {
 // Utils
 import {reportErrorThroughHoneyBadger} from 'src/shared/utils/errors'
 
-export enum ThunkErrorNames {
+export enum OrganizationThunkErrors {
   DefaultOrgStateError = 'DefaultOrgStateError',
   DefaultOrgNetworkError = 'DefaultOrgNetworkError',
 }
@@ -37,14 +37,14 @@ export enum ThunkErrorNames {
 export class DefaultOrgStateError extends Error {
   constructor(message) {
     super(message)
-    this.name = ThunkErrorNames.DefaultOrgStateError
+    this.name = OrganizationThunkErrors.DefaultOrgStateError
   }
 }
 
 export class DefaultOrgNetworkError extends Error {
   constructor(message) {
     super(message)
-    this.name = ThunkErrorNames.DefaultOrgNetworkError
+    this.name = OrganizationThunkErrors.DefaultOrgNetworkError
   }
 }
 
@@ -88,7 +88,9 @@ export const updateDefaultOrgThunk = ({
     const orgStatus = state.identity.currentIdentity.status
 
     if (orgStatus === RemoteDataState.Error) {
-      throw new DefaultOrgStateError(ThunkErrorNames.DefaultOrgStateError)
+      throw new DefaultOrgStateError(
+        OrganizationThunkErrors.DefaultOrgStateError
+      )
     }
   } catch (err) {
     reportErrorThroughHoneyBadger(err, {
@@ -100,11 +102,6 @@ export const updateDefaultOrgThunk = ({
       },
     })
 
-    switch (err.name) {
-      case ThunkErrorNames.DefaultOrgStateError:
-        throw new DefaultOrgStateError(err)
-      default:
-        throw new DefaultOrgNetworkError(err)
-    }
+    throw Error(err)
   }
 }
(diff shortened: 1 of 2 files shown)


src/identity/quartzOrganizations/actions/thunks/index.ts
```diff
@@ -13,73 +16,95 @@ import {
...
+      orgId: newDefaultOrg.id,
+    })
 
     dispatch(setQuartzDefaultOrg(newDefaultOrg.id))
 
     const state = getState()
     const orgStatus = state.identity.currentIdentity.status
 
     if (orgStatus === RemoteDataState.Error) {
-      const defaultOrgErrMsg =
```
wdoconnell, line 66: Removed this, because by throwing an error in the case of a state management problem, we can make a single call to `reportErrorThroughHoneyBadger` in the catch block whether the error is state-based or network-based, and `err.name` will report the correct error.

src/identity/quartzOrganizations/actions/thunks/index.ts
```diff
@@ -13,73 +16,95 @@ import {
...
-        name: defaultOrgErrMsg,
-        context: {
-          org: newDefaultOrg,
-          state: getState(),
-        },
-      })
+      throw new DefaultOrgStateError(ThunkErrorNames.DefaultOrgStateError)
     }
   } catch (err) {
-    dispatch(setQuartzOrganizationsStatus(RemoteDataState.Error))
```
wdoconnell, line 78: This call to `dispatch` isn't necessary anymore because the error state is set in the reducer itself if there is a failure to update the user's default organization.

src/identity/quartzOrganizations/actions/thunks/index.ts
```diff
@@ -13,73 +16,95 @@ import {
...
-
     reportErrorThroughHoneyBadger(err, {
-      name: 'Failed to update /quartz/orgs/default',
-      context: {state: getState()},
+      name: err.name,
+      context: {
+        message: err.message,
+        org: newDefaultOrg,
```
wdoconnell, line 98: The error reporting needs to pass the `newDefaultOrg` as a property in the `context` object so that we can trace errors here. Sending the app state alone won't cut it, because the state doesn't yet include the new default org or its id.

wdoconnell (commented):
Couple of comments from me explaining choices made here.

src/identity/quartzOrganizations/actions/thunks/index.ts
```diff
@@ -13,73 +16,95 @@ import {
...
-class OrgNotFoundError extends Error {
+interface UpdateOrgParams {
+  accountId: number
+  newDefaultOrg: DefaultOrg
+}
+
+// Utils
+import {reportErrorThroughHoneyBadger} from 'src/shared/utils/errors'
+
+export enum ThunkErrorNames {
```
hoorayimhelping, line 32: `ThunkErrorNames` is fairly generic. Can you think of anything more specific and descriptive for this? I was thinking something like `OrganizationNetworkErrors`, or something along those lines, since it describes more specifically what these are.

src/identity/quartzOrganizations/actions/thunks/index.ts
```diff
@@ -13,73 +16,95 @@ import {
...
+        state: getState(),
+      },
     })
+
+    switch (err.name) {
+      case ThunkErrorNames.DefaultOrgStateError:
+        throw new DefaultOrgStateError(err)
+      default:
+        throw new DefaultOrgNetworkError(err)
+    }
```
hoorayimhelping, line 108: Will the UI properly catch and handle these errors?

hoorayimhelping (approved):

src/identity/quartzOrganizations/actions/thunks/index.ts
wdoconnell: Good point - I'll rename the export something more descriptive that's related to `Organizations` specifically.

src/identity/quartzOrganizations/actions/thunks/index.ts
wdoconnell: This could be a YAGNI situation since the UI doesn't care which error is thrown (we only need this for error reporting). I think for now, we can just throw a generic `Error` to the UI, which is all that's needed.

PR #5255, opened 2022-08-01
feat: added copy logging when copy from keyboard

### LINKED ISSUE #5187: firstMile: Eventing - Fire the same code snippet copied events when users copy how they want
**ACs**
- Where ever code snippets are copied in all firstMile wizards, make sure we fire the sample code copied events if users copy via a keyboard shortcut or right-click copy. 
- See all code copied events in our UI Event Dictionary here:https://influxdb.atlassian.net/wiki/spaces/IPO/pages/edit-v2/2530574337

david-rusnak:
Closes #5187

Now logs when users copy code snippets with their keyboard instead of using the `Copy to Clipboard` button. Added only to CLI, but if this proves useful, can add to the rest of the wizards.

### Checklist


src/utils/keyboardCopy.ts
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+
+enum OperatingSystems {
+  Mac = 'MacOS',
+  Windows = 'Windows',
+  Linux = 'Linux',
+}
+
+export const keyboardCopyTriggered = (
```
hoorayimhelping, line 9: this can just stay in the `crossPlatform` file I think, since they both share the `OperatingSystems` enum and they're both focused on cross platform operations.

src/utils/keyboardCopy.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  if (window?.navigator?.userAgent.includes('Windows')) {
+    OS = OperatingSystems.Windows
+  }
+
+  if (window?.navigator?.userAgent.includes('Linux')) {
+    OS = OperatingSystems.Linux
+  }
+
+  if (OS === OperatingSystems.Mac && event.metaKey && event.keyCode == 67) {
```
hoorayimhelping, line 22: can you add a comment that `keyCode` 67 is `c` please?

src/homepageExperience/components/steps/cli/ExecuteAggregateQuery.tsx
```diff
@@ -27,6 +32,24 @@ export const ExecuteAggregateQuery = (props: OwnProps) => {
...
+
+  const fireKeyboardCopyEvent = event => {
+    if (
+      keyboardCopyTriggered(event) &&
+      userSelection().includes('influx query')
+    ) {
+      logCopyCodeSnippet()
+    }
+  }
+
```
hoorayimhelping, line 52: I think in all these components, this code should probably just be part of the `useEffect` function as well, since I think these are side effects

hoorayimhelping (commented):
Looks good, I have a couple of minor comments. Let me know what you think

src/utils/crossPlatform.test.ts
```diff
@@ -68,5 +117,25 @@ describe('checking whether a link should be opened in a new tab', () => {
...
+      const keyEvent = new KeyboardEvent('keydown', {
+        ctrlKey: true,
+        key: 'KeyD',
+      })
+      expect(keyboardCopyTriggered(keyEvent)).toBeFalsy()
+    })
+
+    it('gets the correct selection', () => {
+      expect(userSelection()).toBe('test')
+    })
```
hoorayimhelping, line 139: these test changes are solid, but they only target 1/4 the cases - when the modifier key is pressed but the `C` key isn't pressed. Can you also add cases for the other three cases?

- modifier key is pressed && `C` key is pressed
- modifier key is not pressed && `C` key is pressed
- modifier key is not pressed && `C` key is not pressed

hoorayimhelping (approved):
very solid work!

PR #5257, opened 2022-08-02
fix: wraps page elements in a flexbox to give SimpleTable correct height

### LINKED ISSUE #5256: SimpleTable: pagination controls are hidden in New Data Explorer
SimpleTable has had updates recently to fix bugs and other issues.

However, new Data Explorer's rendering of SimpleTable no longer shows the pagination controls.

Steps to reproduce:
- Ensure the `newDataExplorer` feature flag is on
- Navigate to Data Explorer
- Toggle the "Try new Data Explorer" button on
- Create a query, select Table, Run
- Ensure that the query created in the previous step has enough data to require pagination
- Notice that SimpleTable does not render the pagination controls

Screen Shot 2022-08-01 at 6 25 29 AM

TCL735:
Closes #5256    

This guy:
Screen Shot 2022-08-01 at 7 14 30 PM

And this guy:
Screen Shot 2022-08-01 at 7 14 36 PM

should be wrapped in a flexbox. This fixes the issue of the hidden navigation in SimpleTable for New Data Explorer.

BEFORE: hidden navigation

Screen Shot 2022-08-01 at 6 25 29 AM

AFTER: all fixed

https://user-images.githubusercontent.com/10736577/182278279-7ace3b11-fdbd-482a-9e67-eac017f9195b.mp4

### DIFF
--- a/src/dataExplorer/components/FluxQueryBuilder.tsx
+++ b/src/dataExplorer/components/FluxQueryBuilder.tsx
@@ -4,9 +4,13 @@ import {RemoteDataState} from 'src/types'
 // Components
 import {
   DraggableResizer,
+  FlexBox,
+  FlexDirection,
   Orientation,
   Button,
   IconFont,
+  AlignItems,
+  JustifyContent,
 } from '@influxdata/clockface'
 import {QueryProvider, QueryContext} from 'src/shared/contexts/query'
 import {EditorProvider} from 'src/shared/contexts/editor'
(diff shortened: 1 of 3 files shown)


src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -44,27 +48,38 @@ const FluxQueryBuilder: FC = () => {
...
+          alignItems={AlignItems.Stretch}
+          style={{height: '100%'}}
         >
-          <DraggableResizer.Panel>
-            <Schema />
-          </DraggableResizer.Panel>
-          <DraggableResizer.Panel
-            testID="flux-query-builder-middle-panel"
-            className="new-data-explorer-rightside"
+          <div className="flux-query-builder--menu" style={{width: '100%'}}>
```
genehynson, line 57: nit: could `width: 100%` move to the `flux-query-builder--menu` class?

genehynson (approved):

drdelambre (approved):
ooo i hate `height: 100%`

src/dataExplorer/components/FluxQueryBuilder.tsx
TCL735: Yes.

And `width: 100%` is actually not needed. Originally was there before i played around with the flexbox options. The latest flexbox options takes care of the width for this element.

I will remove this entirely.

drdelambre:
oh no i mean the concept in general. it's what caused this bug when the header got dropped in. if you look at like... all of the components going all the way down that chain from the root to the visualization, most of them _have_ to have a `height:100%` on  them. and the pixel value of 100% is chosen by some weird logic of it's parent height and not just something like the `object-fit: contain` mechanism that these are trying to replicate, so we get to weird cases in the code where we have `calc(100% - superFragileConst)` because we have to. i'm upset with the ie box model, not your changes.

PR #5258, opened 2022-08-02
chore: allow globalheader to serve typeahead independently

wdoconnell:
Part of [UI #4049](https://github.com/influxdata/ui/issues/4049)

The multi-org figma uses the same design for the globalheader (which has a menu menu, and typeahead sub-menu) as for the dropdown typeahead on the user profile page (which is just the typeahead submenu). This tweaks the globalheader by adding an optional property `typeAheadOnly`, which will cause only the submenu to render.

With `typeAheadOnly` not set (or `false`), the main menu renders. Here's an example from GlobalHeader.
Screen Shot 2022-08-02 at 11 01 00 AM

With `typeAheadOnly` set to `true`, main menu doesn't render, and goes straight to typeahead, to be used on the user profile page.
Screen Shot 2022-08-02 at 11 00 49 AM

### Checklist

### DIFF
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
@@ -40,7 +40,7 @@ export interface Props extends StandardFunctionProps {
   mainMenuHeaderText?: string
   mainMenuHeaderIcon?: IconFont
   mainMenuOptions: MainMenuItem[]
-  typeAheadOnly?: boolean
+  onlyRenderSubmenu?: boolean
   typeAheadSelectedOption?: TypeAheadMenuItem
   typeAheadMenuOptions: TypeAheadMenuItem[]
   typeAheadInputPlaceholder?: string
@@ -56,7 +56,7 @@ export class GlobalHeaderDropdown extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props)
     this.state = {
-      showTypeAheadMenu: this.props.typeAheadOnly ?? false,
+      showTypeAheadMenu: this.props.onlyRenderSubmenu ?? false,
       selectedItem: this.props.typeAheadSelectedOption || null,
     }
   }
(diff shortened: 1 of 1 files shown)


src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
```diff
@@ -40,6 +40,7 @@ export interface Props extends StandardFunctionProps {
   mainMenuHeaderText?: string
   mainMenuHeaderIcon?: IconFont
   mainMenuOptions: MainMenuItem[]
+  typeAheadOnly?: boolean
```
hoorayimhelping, line 43: `typeAheadOnly` is a bit arcane and requires some knowledge of the implementation of this component, and couples this property to the existing implementation. If we change the submenu to be something that isn't a `TypeAheadDropdown` component, this property will be hella confusing. I think we'd be better served by naming this in a way that describes what it's doing, rather than how it's doing it. For example `subMenuOnly` or `onlyRenderSubmenu`, etc.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
```diff
@@ -153,7 +155,7 @@ export class GlobalHeaderDropdown extends React.Component<Props, State> {
     return (
       <Dropdown.Menu theme={dropdownMenuTheme} style={dropdownMenuStyle}>
         {/* Multi-org UI tickets #4051 and #4047, when user only has 1 account or 1 org, switch button is disabled */}
-        {typeAheadMenuOptions.length > 1 && (
+        {typeAheadOnly === false && typeAheadMenuOptions.length > 1 && (
```
hoorayimhelping, line 158: this is subtle, but `typeAheadOnly === false` rather than `!typeAheadOnly` implies that `typeAheadOnly` can be a falsy value (like '', or 0) but we only care if it's false. Is that the intent here?

hoorayimhelping (commented):
https://twitter.com/iamdevloper/status/397664295875805184?s=20&t=pdGpS8FRy8Q4KTxQX4qMxw

src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
wdoconnell: Ack, and makes sense. I'll go with `onlyRenderSubmenu`.

wdoconnell:
> https://twitter.com/iamdevloper/status/397664295875805184?s=20&t=pdGpS8FRy8Q4KTxQX4qMxw

Welp, buckle up for the user profile PR then...

hoorayimhelping (approved):

PR #5260, opened 2022-08-02
feat: add user profile page

### LINKED ISSUE #4049: Multi Org Phase 1 - User Profile Page

### LINKED ISSUE #4050: Multi Org Phase 1 - Default Org Experience
**Multi Org - Default Org Experience**

**User Story: Default Org Experience**
As a user, I want to be able to view and change my default organization upon account creation.

    **ACs**
    - The user can view their default org upon account creation on the User Profile Page
    - The user can change their default account via the user profile page
    - The user can change their default org via the user profile page 

    **Dev Notes**
    - We cannot save both account AND org changes simultaneously because org lists depend on the account selected. 
    (Implementation Detail)
    - When does it make most sense for a user to create their orgs within the user journey? 
    - The system automatically sets the default organization for the user upon account creation (Quartz)
    - Assumption: eCommerce will automatically set the default org upon account creation

wdoconnell:
Closes #4049 
Closes #4050 

Multi-org Phase 1 includes a user profile page that:
1. displays information about the user (first name, last name, email address); and 
2. provides ability to change (a) the default account the user hits on logging in and (b) the default org for the _currently active_ account.

These revisions were made to the ACs after the multi-org check-in this week:
1. Account and Org globalheader 'dropdown' at top of the page will _not_ be hidden. No longer confusing, since only one active account will be shown on the user profile page.
2. User cannot update their first name and last name yet. That will be part of multi-org phase 2.
3. 'Save' button notification is more concise, and simply confirms whether the attempt to update the selected preferences was successful or not. If _unsuccessful_, the 'save' button will not change status to disabled, since the change in preferences did not succeed.

Changes persist if the preference update succeeds:

https://user-images.githubusercontent.com/91283923/182427248-d95fd056-0aee-442b-97c2-7d43f2f24162.mov

Changes don't persist if the preference update fails:

https://user-images.githubusercontent.com/91283923/182429370-2a546104-a9fc-4e2f-a030-3c2bb300fc60.mov

### Checklist


src/accounts/context/userAccount.tsx
```diff
@@ -31,10 +31,18 @@ import {updateDefaultQuartzAccount} from 'src/identity/apis/auth'
 export type Props = {
   children: JSX.Element
 }
+
+interface SetDefaultAccountOptions {
```
wdoconnell, line 35: Users currently switch accounts from the switch-account page. We are re-using the switch-account context provider in order to change the default account in User Profile. What we _don't_ want are the wordy notifications generated by the context provider - we want to use the concise ones generated by UI on the User Profile page.

Fix I propose here is: add an `options` object to the handler for setting a default account, that lets us disable the dispatching of notifications from the context provider. When that handler is called from User Profile, this disables the dispatching, so that we don't get the context provider's notifications.

Other ways to fix this behavior could be:
1. Delete the notifications from the switch-account context-provider, and have them be dispatched from the UI in the switch-account page. That's better - the provider probably shouldn't be dispatching the notifications - but there's no need to introduce new code for that when the switch-account page is about to get replaced by user profile.

3. Change the notifications being dispatched from the context provider so that they yield the same message that's expected on the user profile page. I wouldn't do that, both because it changes what appears on the Switch-Account page from what the ACs for that page outline, and because we want a _single_ notification on the User Profile page regardless of whether the user changes their default account, default org-within-an-account, or both.

src/accounts/context/userAccount.tsx
```diff
@@ -103,15 +111,25 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
-      dispatch(notify(accountDefaultSettingSuccess(accountName)))
+
+      if (setDefaultAccountOptions?.disablePopUps !== true) {
+        dispatch(notify(accountDefaultSettingSuccess(accountName)))
+      }
     } catch (error) {
-      dispatch(notify(accountDefaultSettingError(accountName)))
+      if (setDefaultAccountOptions?.disablePopUps !== true) {
+        dispatch(notify(accountDefaultSettingError(accountName)))
+      } else {
```
wdoconnell, line 130: When disabling popups from `handleSetDefaultAccount`, the desired behavior in the case of an error is for an error to be thrown to the User Profile page, so that the user profile page can catch it, and dispatch a single error notification.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderDropdown.scss
```diff
@@ -16,6 +16,29 @@
   z-index: 0;
 }
 
+.global-header--list {
```
wdoconnell, line 19: These are fixes with many thanks to @ChitlangeSahas to sleekify the change-default dropdowns, and make them consistent with the figma.

src/identity/components/userprofile/DefaultDropdown.tsx
```diff
@@ -0,0 +1,62 @@
+// Libraries
```
wdoconnell, line 1: This is a wrapper component for the change-default-account and change-default-org dropdowns.

src/identity/components/userprofile/UserDefaults.tsx
```diff
@@ -0,0 +1,140 @@
...
+          await dispatch(
+            updateDefaultOrgThunk({
+              accountId: loggedInAccount.id,
+              newDefaultOrg: selectedOrg,
+            })
+          )
+        }
+
+        dispatch(notify(userProfileSaveSuccess()))
+      } catch (err) {
```
wdoconnell, line 105: Since the user will get the error message from either function failing, there's sort of an atomicity problem: the user could successfully change their account, but then the `orgs` API goes down, and the user would see the error notification. I think that's fine:

1. the new default account will still show correctly - so the user has the correct information. `handleSetDefaultAccount` will update context; which will trigger an update of the `accounts`, which triggers an update of `defaultAccount`, which triggers an update of `setDefaultAccount`.
2. in that case, the user was notified that not everything changed - so they know they have to retry to update their preferences.
3. the atomicity issue would need to be resolved at the quartz level - we can't really do anything about it in the UI, unless we want to _reverse_ the update that was correct (and then what if that reversal fails).

src/shared/copy/notifications/categories/accounts-users-orgs.ts
```diff
@@ -91,16 +91,6 @@ export const orgCreateSuccess = (): Notification => ({
   message: 'Organization was successfully created',
 })
 
-export const orgDefaultSettingError = (orgName: string): Notification => ({
```
wdoconnell, line 94: We don't need these notifications, since we decided to go with a single 'user successfully updated preferences' notification.

src/identity/components/userprofile/UserDefaults.tsx
```diff
@@ -0,0 +1,140 @@
...
+        dispatch(notify(userProfileSaveError()))
+      }
```
wdoconnell, line 107: With this behavior, the user will see the update didn't work - the same unsaved preference appears in the dropdown - and the 'save' button is still active.

Another behavior here could be, when the update doesn't succeed - reset the failed dropdown to the current default, which will also reset the 'save' button to inactive.

src/identity/components/userprofile/UserDefaults.tsx
```diff
@@ -0,0 +1,140 @@
...
+    setSelectedOrg(defaultOrg)
+  }, [orgs, defaultOrg])
+
+  const userPickedNewAccount =
+    defaultAccount?.id !== selectedAccount?.id && selectedAccount !== null
+
+  const userPickedNewOrg =
+    defaultOrg?.id !== selectedOrg?.id && selectedOrg !== null
+
+  const userChangedPrefs = userPickedNewAccount || userPickedNewOrg
```
hoorayimhelping, line 79: nice, very clean

src/identity/components/userprofile/UserDefaults.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const saveButtonStatus =
+    userPickedNewAccount || userPickedNewOrg
```
hoorayimhelping, line 82:
```suggestion
    userChangedPrefs
```

src/identity/components/userprofile/UserDefaults.tsx
```diff
@@ -0,0 +1,140 @@
...
+          })
+        }
+
+        if (userPickedNewOrg) {
+          await dispatch(
+            updateDefaultOrgThunk({
+              accountId: loggedInAccount.id,
+              newDefaultOrg: selectedOrg,
+            })
+          )
```
hoorayimhelping, line 101: after these two network requests, do the `setSelectedAccount` and `setSelectedOrg` state methods need to be called?

src/identity/components/userprofile/UserProfileInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+  status: ComponentStatus
+  header: string
+  text: string
+}
+
+const inputStyle = {
+  width: '368px',
+  backgroundColor: 'black',
+  color: InfluxColors.White,
+}
```
hoorayimhelping, line 20: :+1:

src/shared/copy/notifications/categories/accounts-users-orgs.ts
```diff
@@ -130,3 +120,14 @@ export const removeUserSuccessful = (): Notification => ({
...
+
+export const userProfileSaveSuccess = (): Notification => ({
+  ...defaultSuccessNotification,
+  message: 'Saved changes to your profile.',
+})
+
+export const userProfileSaveError = (): Notification => ({
+  ...defaultErrorNotification,
+  message:
+    'Failed to save all changes to your profile. Please try again later.',
```
hoorayimhelping, line 132: We might consider dropping `later`. That always makes me think I should come back in a couple of hours or days.

src/utils/deepLinks.ts
```diff
@@ -33,4 +33,5 @@ export const buildDeepLinkingMap = (org: Organization) => ({
   '/me/usage': `/orgs/${org.id}/usage`,
   '/me/users': `/orgs/${org.id}/users`,
   '/me/variables': `/orgs/${org.id}/settings/variables`,
+  '/user/profile': `/orgs/${org.id}/user/profile`,
```
hoorayimhelping, line 36: - this should be `/me/profile` not `/user/profile`
 - can you alphabetize this please?
 - like we mentioned offline, be sure to add a case for this new link to the e2e test
 - don't forget to add it to the deep link internal docs page either

hoorayimhelping (commented):
Good work! A few comments and questions.

>Changes persist if the preference update succeeds:

When you change your default account from the dropdown, I'd expect the account property to change in the Default Organization section below it. Is it not supposed to do that?

wdoconnell:
@juliajames12 has a couple adjustments, putting this back in draft for a moment.

wdoconnell:
> Good work! A few comments and questions.
> 
> > Changes persist if the preference update succeeds:
> 
> When you change your default account from the dropdown, I'd expect the account property to change in the Default Organization section below it. Is it not supposed to do that?

@hoorayimhelping Correct, it's not supposed to do that. The design folks ended up deciding on was: 

1. the top dropdown reflects the user's _current default_ account (whether they're currently logged into that account or not).
2. the bottom dropdown reflects the user's current default org in the _currently logged in account_.

Might be confusing because we added endpoints that let you do more than this - it's now _possible_ on the backend to set a default org in an account you're not in. But pending user feedback, the idea is that functionality might be too confusing for the page at this point.

src/accounts/context/userAccount.tsx
```diff
@@ -103,15 +111,25 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
+    setDefaultAccountOptions?: SetDefaultAccountOptions
+  ) {
     const accountName = getAccountNameById(newDefaultAcctId)
 
     try {
       await updateDefaultQuartzAccount(newDefaultAcctId)
       setDefaultAccountId(newDefaultAcctId)
-      dispatch(notify(accountDefaultSettingSuccess(accountName)))
+
+      if (setDefaultAccountOptions?.disablePopUps !== true) {
```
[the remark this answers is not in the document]
wdoconnell: Good point - changing this in next commit.

src/identity/components/userprofile/DefaultDropdown.tsx
```diff
@@ -0,0 +1,62 @@
...
+    >
+      <Form.Element
+        label={`Default ${entityLabel}`}
+        className="user-profile-page--form-element"
+      >
+        <GlobalHeaderDropdown
+          mainMenuOptions={[]}
+          onlyRenderSubmenu={true}
+          typeAheadMenuOptions={entityList}
+          typeAheadInputPlaceholder={`Search ${entityLabel}s ...`}
```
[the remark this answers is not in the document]
wdoconnell: It won't, but in this case the component is used for only two purposes identified in the `enum EntityLabel` - `Account` or `Organization`, so I think it's safe.

src/identity/components/userprofile/UserDefaults.tsx
wdoconnell: fortunately not - `handleSetDefaultAccount` and `updateDefaultOrgThunk` update context and the selector and thus `accounts` and `orgs`. when those variables change, the `useEffect` hooks call `setSelectedAccount` and `setSelectedOrg`.

src/utils/deepLinks.ts
wdoconnell: Is there a reason we want this to be `me` instead of `/user`? I see that's where the other deep links go, but wondered whether we should go with something different here since we're moving away from the concept of `me` elsewhere.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderDropdown.scss
```diff
@@ -16,6 +16,29 @@
...
+  /* Works on Firefox */
+  scrollbar-width: thin !important;
```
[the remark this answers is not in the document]
wdoconnell: The use of `!important!` actually comes up in two stylesheets in the PR, so here's how I resolved it:

1. This particular class didn't actually need the `!important`. It can be safely removed.

2. Elsewhere, it was necessary to force the width property on the portion of the `GlobalHeader` component relying on `Dropdown` from clockface. After playing around it with, the width property _is_ getting correctly populated to the `Dropdown` via the classname, but that actually causes a separate issue, which is that both the parent component and the `Dropdown` have the same style, so the margin and padding would have to be separately adjusted for each with different class names in order for us to get the right width.

Simpler solution was to add this into a style outside the component, since there are only two properties being adjusted, so that's what I ended up proposing in my new commit.

wdoconnell:
Adding additional ACs that you will see in the user profile page per conversation with @juliajames12 today. FYI @Luckette @kristinarobinson  

1. We should not display the 'change default accounts/orgs' panel on this page at all if the user doesn't have more than one account, and doesn't have more than one org in the current account (therefore, the user has nothing to change). In this case, only the 'user details' panel will be shown.

2. If the user has only one _account_, do not show the 'change default account' fieldset. 

3. If the user has only _one org in the current account_, do not show the 'change default org' fieldset.

src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
```diff
@@ -60,7 +60,7 @@ class IdentityUserAvatar extends React.Component<Props, State> {
           <hr />
         </div>
         <div className="user-popover-footer">
-          <Link className="user-popover-footer--button" to="/">
+          <Link className="user-popover-footer--button" to="/me/profile">
```
hoorayimhelping, line 63: This should link directly to the page, rather than going through the deep link, since using a deep link will actually route to the `NotFound` page first, which will also fire an event

hoorayimhelping (approved):
LGTM barring using a deep link inside a ReactRouter `Link` component

src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
wdoconnell: Got it - thanks. I'll change this to route directly to the page, which is: `/orgs/${org.id}/user/profile`

wdoconnell:
@ChitlangeSahas To make deep linking work, I needed to give the User Avatar component access to the `org.id` property. I also made some tweaks to the props, so that we're only passing primitive values directly to the child component. Please let me know if you have any edits. 😺

hoorayimhelping (approved):
LGTM, but one question about `shouldUseQuartzIdentity`

src/shared/containers/SetOrg.tsx
```diff
@@ -314,6 +316,14 @@ const SetOrg: FC = () => {
               component={CliWizard}
             />
           )}
+          {/* User Profile Page */}
+          {CLOUD && shouldUseQuartzIdentity() && isFlagEnabled('multiOrg') && (
```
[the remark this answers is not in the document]
wdoconnell: So: the globalheader relies on states generated by the identity reducer, so if the user tries to flip on multi-org but isn't using quartz identity, it won't work.

PR #5262, opened 2022-08-02
feat: adding function path ui

### LINKED ISSUE #5141: Bug: Duplicate flux functions
In the flux function list, several flux functions are duplicated (ie. array.concat, array.filter, array.from). 

Image

Flux functions should only show once in the list.

drdelambre:
Closes #5141 

there aren't dupes, just information loss about the function's path between the function subsystem and the user. this adds a little header so the user doesn't get confused

Kapture 2022-08-02 at 10 47 42

### DIFF
--- a/src/shared/components/DynamicFunctionsList.tsx
+++ b/src/shared/components/DynamicFunctionsList.tsx
@@ -1,7 +1,9 @@
 import React, {FC, useCallback, useEffect, useState} from 'react'
 import {useDispatch, useSelector} from 'react-redux'
+import {DapperScrollbars} from '@influxdata/clockface'
 
 // Components
+import SearchWidget from 'src/shared/components/search_widget/SearchWidget'
 import {FluxFunction} from 'src/types/shared'
 import Fn from 'src/shared/components/FilterList/InjectionOption'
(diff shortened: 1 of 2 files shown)


src/shared/components/DynamicFunctionsList.tsx
```diff
@@ -60,6 +64,74 @@ const DynamicFunctionsList: FC<Props> = ({onSelect}) => {
...
+        a.toLocaleLowerCase().localeCompare(b.toLocaleLowerCase())
+      )
+      .reduce((acc, curr) => {
+        acc.push(
+          <div className="flux-toolbar--path" key={`fn-head-${curr}`}>
+            {curr}
+          </div>
+        )
+
+        return acc.concat(
```
appletreeisyellow, line 94: Why you use `push` in line 88 but use `concat` here? Is it because you want to make it clear to return a new set of array?

appletreeisyellow (approved):
Nice fix! I left one comment as a question. Not a blocker

src/shared/components/DynamicFunctionsList.tsx
drdelambre: the push is for adding just one item and concat is for adding a bunch of items

Issue #5263, opened 2022-08-02
"Successfully deleted pinned task" alert displays when deleting a dashboard


taramk:
## About the bug

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. Go to an account in production
2. Delete any dashboard from the dashboard card view (org/dashboards-list)

**Expected behavior:**
I don't see an alert about a pinned task

**Actual behavior:**
I see an alert about a pinned task

**Visual Proof:**
![image](https://user-images.githubusercontent.com/1813419/182452965-c22f43ae-885a-4128-8c19-04c90cba1e95.png)

## About your environment
* Using my Cloud 2 account, org id 78c7ba79046ae6fa
* Tested deleting several random dashboards and saw the alert each time

PR #5264, opened 2022-08-02
fix(dashboards): only notify when a pinned item is actually deleted

hoorayimhelping:
Closes https://github.com/influxdata/ui/issues/5263

- When deleting a dashboard, only notify that a pinned dashboard has been deleted if that dashboard was actually pinned
- Fix notification text to say dashboard, not task

Screen Shot 2022-08-02 at 3 29 42 PM

### Checklist

- ~[] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)~

### DIFF
--- a/src/dashboards/components/dashboard_index/DashboardCard.tsx
+++ b/src/dashboards/components/dashboard_index/DashboardCard.tsx
@@ -231,11 +231,13 @@ class DashboardCard extends PureComponent<Props> {
   private handleDeleteDashboard = async () => {
     const {id, name, onDeleteDashboard} = this.props
     onDeleteDashboard(id, name)
-    try {
-      await deletePinnedItemByParam(id)
-      this.props.sendNotification(pinnedItemSuccess('task', 'deleted'))
-    } catch (error) {
-      this.props.sendNotification(pinnedItemFailure(error.message, 'delete'))
+    if (isFlagEnabled('pinnedItems') && CLOUD && this.props.isPinned) {
+      try {
+        await deletePinnedItemByParam(id)
+        this.props.sendNotification(pinnedItemSuccess('dashboard', 'deleted'))
+      } catch (error) {
+        this.props.sendNotification(pinnedItemFailure(error.message, 'delete'))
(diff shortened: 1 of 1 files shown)


TCL735 (approved):

src/dashboards/components/dashboard_index/DashboardCard.tsx
```diff
@@ -231,11 +231,13 @@ class DashboardCard extends PureComponent<Props> {
   private handleDeleteDashboard = async () => {
     const {id, name, onDeleteDashboard} = this.props
```
wdoconnell, line 232: since these are getting destructured anyway, to save some space should we destructure `isPinned` and `sendNotification` here too?

src/dashboards/components/dashboard_index/DashboardCard.tsx
```diff
@@ -231,11 +231,13 @@ class DashboardCard extends PureComponent<Props> {
...
     onDeleteDashboard(id, name)
-    try {
-      await deletePinnedItemByParam(id)
-      this.props.sendNotification(pinnedItemSuccess('task', 'deleted'))
-    } catch (error) {
-      this.props.sendNotification(pinnedItemFailure(error.message, 'delete'))
+    if (isFlagEnabled('pinnedItems') && CLOUD && this.props.isPinned) {
```
wdoconnell, line 234: Why `CLOUD` only? Doesn't exist for OSS?

hoorayimhelping: correct, it's a cloud-only feature because it uses a backend nodejs service that isn't available in OSS installs

src/dashboards/components/dashboard_index/DashboardCard.tsx
hoorayimhelping: For consistency that might be a good idea, but I'd rather get the bugfix in rather than go through another round of CI (which itself is an indication that our CI takes too long to run).

Issue #5169, opened 2022-07-25
Thought experiment: What if `flux-lsp-*/{parse|format_from_js_file}` went away?


rockstar:
The title says it all. What if, for whatever reason, we went to import `flux-lsp-browser/parse` and it wasn't there. What would break? What could be re-worked to work in a different way? What features would we lose? What bugs would go away?

Below is a (comprehensive as possible) list of places where `parse` and `format_from_js_file` are being used, and an attempt to speak to how this might need to change in order to keep feature sets.

Please consider this issue a place to add historical context and clarifying information.

- [ ] `flows/pipes/Visualization/index.ts` - `parse` is being used here to check to see if the flux has any statements it, and if not, return an empty string. The `ast` is then thrown away, and then the raw query string could have `|> limit(n: 100)` and/or `|> last()` tacked onto the end. The original intent was to check if there were only comments in the file before trying to append to the query. A potential fix is in #5132 .
- [ ] `flows/pipes/Notification/ExportTask.ts`
  - `parse` is used to extract `option v` values and `option task` values
  -  `parse` is used to generate a series of imports from raw text
  - `format_from_js_file` is used to insert a mutated query into another query as a string
- [ ] `src/flows/pipes/Notification/view.tsx`
  - `parse` is used to find `option task` statements
- [ ] `src/flows/pipes/RawFluxEditor/index.ts`
  - `parse` and `format_from_js_file` are used to merge two queries together  
- [ ] `src/flows/pipes/Schedule/view.tsx`
- [ ] `src/flows/pipes/Table/index.ts`
- [ ] `src/flows/templates/types/task.ts`
- [ ] `src/languageSupport/languages/flux/parser/index.ts`
- [ ] `src/shared/contexts/query.tsx`
- [ ] `src/writeData/components/ClientCodeQueryHelper.tsx`

PR #5270, opened 2022-08-02
feat: toggle extern

### LINKED ISSUE #5057: Error messages different between old/new data explorer
in the old data explorer, a certain erroring query results in a meaningful error message.  in the new, the error message seems unrelated to the query problem.  

Image

Image

drdelambre:
Closes #5057 

this one is just to delay the widowPeriod calc until the end. meaning it can't happen in the beginning (like notebooks). Found that we can actually have this query properly use the fallback window period (even if it can't calculate it currently) by moving the try catch to a different place, but the ask is to make the error the same as data explorer.

Kapture 2022-08-02 at 16 39 34


src/shared/contexts/query.tsx
```diff
@@ -183,9 +219,50 @@ const _addWindowPeriod = (ast, optionAST): void => {
...
+  defaults: Record<string, string> = {}
+) => {
+  // remove and join duplicate options declared in the query
+  const joinedOption = remove(
+    ast,
+    node =>
+      node.type === 'OptionStatement' && node.assignment.id.name === optionName
+  ).reduce((acc, curr) => {
+    // eslint-disable-next-line no-extra-semi
+    ;(curr.assignment?.init?.properties || []).reduce((_acc, _curr) => {
```
appletreeisyellow, line 234: i know this was from previous code, just curious why `;` at the front?

src/flows/context/flow.query.tsx
```diff
@@ -46,7 +51,10 @@ export const DEFAULT_CONTEXT: FlowQueryContextType = {
   queryDependents: () => {},
   getPanelQueries: _ => ({
     id: '',
-    scope: {},
+    scope: {
+      region: '',
+      org: '',
+    },
```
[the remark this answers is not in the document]
drdelambre: i think it's an artifact of me slowly coming to terms that [this line](https://github.com/influxdata/ui/blob/master/src/shared/contexts/query.tsx#L305)  means that all scopes should be optional. that includes all of the hardcoded ones in the bucket and fluxquerybuilder contexts you added. i'll tackle that as it's own PR, to just clean that all up

appletreeisyellow (approved):
It's cool to have a new feature while fixing a bug! 🎉

src/shared/contexts/query.tsx
```diff
@@ -29,15 +29,50 @@ interface CancelMap {
...
+  Params,
+  Extern,
+  Injection,
+}
+
+export interface QueryOptions {
+  overrideMechanism: OverrideMechanism
+}
+
+export interface QueryScope {
```
wiedld, line 42: TY for defining this more explicitly!

src/shared/contexts/query.tsx
```diff
@@ -316,11 +361,105 @@ export const QueryProvider: FC = ({children}) => {
...
+
+      const optionTexts = Object.entries(options)
+        .map(([k, v]) => {
+          const vals = Object.entries(v).map(([_k, _v]) => `  ${_k}: ${_v}`)
+          return `option ${k} =  {${vals.join(',\n')}}`
+        })
+        .join('\n\n')
+
+      const queryAST = parse(query)
+      let optionAST = parse(optionTexts)
```
wiedld, line 390: An observation. Not something to change now. 

Because we parse the `option v = {...}` into an AST, and then modify/unify it...we are technically including AST metadata (e.g. location, source) in the extern option sent with the network payload:
Screen Shot 2022-08-03 at 12 49 13 PM

This is different from the old DataExplorer and Dashboards, which construct all of the these AST nodes without parsing them. And therefore the metadata is not included in the extern.
```
const variables: Variable[] = []  // typed objects which shared QueryContext does not use 
const extern = buildUsedVarsOption(text, variables)
```

wiedld: Also a note to @rockstar . This is an additional point at which the code will need to update for this issue to remove parse() (https://github.com/influxdata/ui/issues/5169). In order to remove the use of parse here, we have to construct our extern object.

In old DataExplorer and Dashboards, they have utils methods which utilize the Variable objects in order to construct an AST. Since shared QueryContext does not utilize the Variable types (for reasons that @drdelambre is most familiar about)...we just have to create our own helper methods here.

wiedld (commented):
Only requesting the following changes:
* please have the branching `if (mechanism !== OverrideMechanism.Extern` be a separate helper
* would about the questions regarding `option params`?

src/shared/contexts/query.tsx
```diff
@@ -316,11 +361,105 @@ export const QueryProvider: FC = ({children}) => {
...
 
+    if (mechanism !== OverrideMechanism.Injection) {
+      const options: Record<string, any> = {}
+
+      if (Object.keys(override?.vars ?? {}).length) {
+        options.v = override.vars
+      }
+      if (Object.keys(override?.params ?? {}).length) {
+        options.params = override.params
+      }
```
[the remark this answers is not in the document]
drdelambre: totes is (https://github.com/influxdata/idpe/commit/3546d58b42cb6f909b168750514ab02d52eae05d). reading through idpe, i have no idea what the difference is between `options.params`,  `params`, or `options.v` are, or why any of those distinctions are being mades. extern vs ast also looking suss, but i haven't looked at that one at all. which to choose is all over the place in our docs and examples too. generically supporting these reserved keywords for this iteration allows us to at least run queries without breaking as we flesh out the ability / choice, and if we have to move the params up a level and out of the options down the line for some reason, it's pretty trivial to do that.

src/shared/contexts/query.tsx
```diff
@@ -316,11 +361,105 @@ export const QueryProvider: FC = ({children}) => {
       headers['Authorization'] = `Token ${override.token}`
     }
 
-    const body = {
+    const body: RequestBody = {
       query,
       dialect: {annotations: ['group', 'datatype', 'default']},
     }
 
+    if (mechanism !== OverrideMechanism.Injection) {
```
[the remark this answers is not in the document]
drdelambre: i think you're confused on the mechanism enum choices. how about i rename them to:
injection => inline
extern => ast
param => json

src/shared/contexts/query.tsx
```diff
@@ -316,11 +361,105 @@ export const QueryProvider: FC = ({children}) => {
...
+
+            node.assignment.init.properties.push({
+              type: 'Property',
+              key: {
+                type: 'Identifier',
+                name: 'windowPeriod',
+              },
+              value: {
+                type: 'DurationLiteral',
+                values: [{magnitude: FALLBACK_WINDOW_PERIOD, unit: 'ms'}],
```
wiedld, line 428: On fallback (a.k.a. cannot modify the AST...such as the complex queries), can we try a tiered fallback?
1. using the windowPeriod from the presets (`SELECTABLE_TIME_RANGES`)
2. then the `FALLBACK_WINDOW_PERIOD`

wiedld (changes requested):
This one is a requested change please!

src/shared/contexts/query.tsx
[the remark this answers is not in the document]
wiedld: The Code in the branch for the `mechanism !== OverrideMechanism.Injection` does the extern AST. So it should be a branch which is `mechanism == OverrideMechanism.Extern`.

Then a separate branch would exist for the json. `mechanism == OverrideMechanism.Param`

src/shared/contexts/query.tsx
drdelambre: i do  not know what you mean

drdelambre: are you talking about this that's already in there? https://github.com/influxdata/ui/blob/master/src/shared/contexts/query.tsx#L158

that wont work for the bug i'm working on as it fails here https://github.com/influxdata/ui/blob/master/src/shared/contexts/query.tsx#L117, meaning we aren't getting to any valid ranges to look up against the `SELECTABLE_TIME_RANGES`,  and i can't find any equivalent functionality in the old data explorer code path that utilizes this fallback differently

maybe i'm missing something?

drdelambre: offlined: this was a feature request and out of scope

wiedld (approved):
Thanks for making the resquested changes!

Note that this is handling AST and json as the same. Where they are not (json is not an option AST). But that's something to fix in the future, once we implement parameterized queries.

PR #5289, opened 2022-08-03
test: demonstrate PoC, for server events triggering UI events.

wiedld:
**This is a PoC only. Not intended to merge.**

## Challenge

We need a structured way, within the LSP spec, to listen to incoming message from the LSP. And act to surface that message in the UI (not monaco-editor).

### Discarded approaches:
- Using spec method `workspace/applyEdit`
    - This posts to the browser console.
    - Screen Shot 2022-08-03 at 3 51 19 PM

    - Instead…can we listen to this traffic?
- monaco-languageclient:
    - Listening to the connection between the LSP and the monaco-editor, using the existing interfaces provided by the monaco-languageclient.
    - options:
        - Middleware
            - Are interceptors of provider features. Such as intercepting didOpen, didChange, etc. There is no intercept provider for the `window/onMessage`.
            - Also provides overrides of the workspace and window objects. This is too heavy for our use. (We are not wrapping the Monaco-editor in our own typed LSP window target.)
        - onResponse
            - This lib is an extension of the vscode-languageserver BaseLanguageClient. Neither provides the proper interface to intercept the server response. (There is no onResponse(). This has been commented upon in other github issues too.)
            - interface: https://github.com/microsoft/vscode-languageserver-node/blob/64156c6aa08be17a1475a126b0bb2d9b6c8ba1bc/client/src/common/client.ts#L429 

### Successful approach:
- Using telemetry.
- parts:
    - Server sends `telemetry/event`
    - monaco-languageclient listener interface for onTelemetry().
- This feels like the right approach:
    - https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#telemetry_event
    - Found at least one similar use case in terraform LSP.
- Did PoC, surfaced as a simple UI notification....but we are actually going to do different things (based on design).
    - Screen Shot 2022-08-03 at 4 25 05 PM

### DIFF
--- a/src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
+++ b/src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
@@ -26,6 +26,11 @@ import {EditorType} from 'src/types'
 // error notification
 import {reportErrorThroughHoneyBadger} from 'src/shared/utils/errors'
 
+// UI interaction, outside the editor
+import {notify} from 'src/shared/actions/notifications'
+import {defaultErrorNotification} from 'src/shared/copy/notifications'
+import {storeSingleton} from 'src/store/configureStore'
+
 // install Monaco language client services
 MonacoServices.install(monaco)
 
@@ -85,6 +90,14 @@ export function initLspWorker() {
   messageWriter = new BrowserMessageWriter(worker)
   const connection = createMessageConnection(messageReader, messageWriter)
   const languageClient = createLanguageClient(connection)
+  languageClient.onTelemetry((data: unknown) => {
+    storeSingleton.dispatch(
+      notify({
+        ...defaultErrorNotification,
+        message: (data as {message: string}).message,
+      })
+    )
+  })
   const disposable = languageClient.start()
   connection.onError(e => handleConnectionError(e))
   connection.onClose(() => {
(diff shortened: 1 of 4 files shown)


src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
```diff
@@ -85,6 +90,14 @@ export function initLspWorker() {
   messageWriter = new BrowserMessageWriter(worker)
   const connection = createMessageConnection(messageReader, messageWriter)
   const languageClient = createLanguageClient(connection)
+  languageClient.onTelemetry((data: unknown) => {
+    storeSingleton.dispatch(
```
wiedld, line 94: Note: not going to merge this.
Don't plan to use this pattern of a direct store dispatch. Only for fast PoC.

wiedld:
After a fantastic brainstorming convo with @rockstar and @appletreeisyellow , we are going in a different direction. The general need outlined here (e.g. detection of outgoing LSP traffic, by the UI for event and error handling), will be handled at a middleware level between the LSP and the monaco-editor. This is NOT the same as the monaco-languageclient middleware configuration (which we already ruled out in this PoC)....rather, it will be our own middleware between the monaco-languageclient and the worker thread.

Issue #5290, opened 2022-08-04
GetOrganizations: /quartz/identity is being called repeatedly


hoorayimhelping:
`/quartz/identity` is being called multiple times unnecessarily. The subsequent calls are returning a 304 status, meaning the browser has cached the response and is returning the value from the cache.

![Image](https://user-images.githubusercontent.com/146112/182848164-533c2941-401d-437f-9f4a-c6abdad0ec1e.png)

### Acceptance Criteria
- Only call `quartz/identity` from the `GetOrganizations` component when necessary

PR #5291, opened 2022-08-04
fix(GetOrganizations): check quartzIdentityStatus before calling getQuartzIdentityThunk

hoorayimhelping:
Closes https://github.com/influxdata/ui/issues/5290

`/quartz/identity` was being called multiple times by the `GetOrganizations` component. The HTTP response on subsequent calls was a 304, meaning the browser has cached the response and is working from the cache. While this is good because we're not using a ton of HTTP resources, the response from the API still updated redux state, which causes re-renders.

The issue was that the call to `getIdentityThunk` was only checking that `quartzMe.status` isn't `NotStarted`. The fix is to make it also check that `me.identity.status` is `NotStarted`. When the `quartzIdentity` flag is true, `quartzMe.status` isn't updated from `NotStarted` to `Done` until after several re-renders of `GetOrganizations`. Within that time, `getIdentityThunk` is called each time, needlessly.

When the `quartzIdentity` flag is false, `/quartz/me` is correctly only called once, even with the aditional check.

### Before
before

### After
after

### Checklist

### DIFF
--- a/src/shared/containers/GetOrganizations.tsx
+++ b/src/shared/containers/GetOrganizations.tsx
@@ -76,7 +76,8 @@ const GetOrganizations: FunctionComponent = () => {
   useEffect(() => {
     if (
       isFlagEnabled('uiUnificationFlag') &&
-      quartzMeStatus === RemoteDataState.NotStarted
+      quartzMeStatus === RemoteDataState.NotStarted &&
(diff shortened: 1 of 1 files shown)


wdoconnell (approved):
Amaze - nice fix!

PR #5300, opened 2022-08-04
fix: correct globalheader switch-account URL.

### LINKED ISSUE #5293: fix switch-account URL for globalheader
GlobalHeader 'switch-account' is not currently working as intended. The `switchAccount` function routes the user to orgs/orgId/accounts/accountId, but should be routing the user simply to /accounts/accountId.

**Steps to reproduce:**

1. enable multiOrg and quartzIdentity feature flags.
2. refresh page
3. switch account from globalHeader

**Expected behavior:**
Refresh of app; switch into new account

**Actual behavior:**
`404`, as URL for that account is not visitable

**Visual Proof:**
https://user-images.githubusercontent.com/91283923/182898197-c064a3f9-3a52-41b9-a146-0dc942727292.mov

wdoconnell:
Closes #5293

Switch-account in GlobalHeader was not working because the URL the user needs to visit to change accounts should be `${CLOUD_URL}/accounts/:accountId` (not the current URL that included the org id - like `/orgs/:orgId/accounts/:accountId`. We can be confident this works effectively, since this exact URL is used on the switch account page (which works).   

The switch orgs URL should be similarly updated to `$CLOUD_URL/orgs/:orgId`.

### Checklist

- [X] Feature flagged, if applicable - YES, `multiOrg`

### DIFF
--- a/src/identity/components/GlobalHeader/AccountDropdown.tsx
+++ b/src/identity/components/GlobalHeader/AccountDropdown.tsx
@@ -39,12 +39,12 @@ export const AccountDropdown: FC<Props> = ({
     {
       name: 'Settings',
       iconFont: IconFont.CogSolid_New,
-      href: `${CLOUD_URL}/orgs/${activeOrg.id}/accounts/settings`,
+      href: `/orgs/${activeOrg.id}/accounts/settings`,
     },
     {
       name: 'Billing',
       iconFont: IconFont.Bill,
-      href: `${CLOUD_URL}/orgs/${activeOrg.id}/billing`,
+      href: `/orgs/${activeOrg.id}/billing`,
     },
   ]
 
--- a/src/identity/components/GlobalHeader/AccountDropdown.tsx
+++ b/src/identity/components/GlobalHeader/AccountDropdown.tsx
@@ -1,12 +1,10 @@
+// Libraries
 import React, {FC} from 'react'
 import {IconFont} from '@influxdata/clockface'
-import {OrganizationSummaries, UserAccount} from 'src/client/unityRoutes'
-import {
-  GlobalHeaderDropdown,
-  TypeAheadMenuItem,
-} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown'
 
+// Types
 type OrgSummaryItem = OrganizationSummaries[number]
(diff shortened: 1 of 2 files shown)


src/identity/components/GlobalHeader/AccountDropdown.tsx
```diff
@@ -1,22 +1,30 @@
+// Libraries
 import React, {FC} from 'react'
 import {IconFont} from '@influxdata/clockface'
-import {OrganizationSummaries, UserAccount} from 'src/client/unityRoutes'
-import {
```
wdoconnell, line 4: This is just a refactor/reorganization of imports.

src/identity/components/GlobalHeader/AccountDropdown.tsx
```diff
@@ -1,22 +1,30 @@
...
 const menuStyle = {width: '250px'}
 
+// Components
+import {
+  GlobalHeaderDropdown,
+  TypeAheadMenuItem,
+} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown'
+
+// Constants
+import {CLOUD_URL} from 'src/shared/constants'
```
wdoconnell, line 26: CLOUD_URL points to /auth in cloud environments, so the URL for switching account ends up being: /auth/accounts/:accountId. 

Tested this in remocal, verified that it's the same code as is used for switch-account; and verified in prod that this is the URL that is actually hit by the switch-account page.

TCL735 (approved):

src/identity/components/GlobalHeader/OrgDropdown.tsx
```diff
@@ -1,13 +1,21 @@
...
+
+// Types
+import {OrganizationSummaries} from 'src/client/unityRoutes'
+
+// Constants
+import {CLOUD_URL} from 'src/shared/constants'
 
 const switchOrg = (org: TypeAheadMenuItem) => {
-  window.location.href = `/orgs/${org.id}`
+  window.location.href = `${CLOUD_URL}/orgs/${org.id}`
```
wdoconnell, line 18: I've confirmed both of these URLs with the AIM team.

TCL735 (approved):

PR #5303, opened 2022-08-04
chore(identity): tweak getCurrentOrgDetailsThunk signature and when it's called

hoorayimhelping:
While working on reducing re-renders, it became apparent that the call to `getCurrentOrgDetailsThunk` in `GetOrganizations` wasn't currently doing anything and was called later in the component tree. This PR removes the later call and tweaks the call in `GetOrganizations` so that it relies on `org.id`. This has the effect of making call slightly more obvious, and reducing re-renders, since `GetOrganizations` doesn't have to wait on deeper component to populate data.

### Checklist

### DIFF
--- a/src/organizations/components/OrgProfileTab/index.tsx
+++ b/src/organizations/components/OrgProfileTab/index.tsx
@@ -38,14 +38,16 @@ const OrgProfileTab: FC = () => {
 
   const dispatch = useDispatch()
 
+  const identityOrgId = identity.currentIdentity.org.id
+
   useEffect(() => {
-    if (CLOUD && shouldUseQuartzIdentity()) {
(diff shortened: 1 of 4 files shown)


TCL735 (approved):

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -38,14 +38,16 @@ const OrgProfileTab: FC = () => {
 
   const dispatch = useDispatch()
 
+  const identityOrgId = identity.currentIdentity.org.id
```
hoorayimhelping, line 41: identity.currentIdentity.org.id all have proper default values, so there is no need for optional chaining here, since the default value of `identity.currentIdentity.org.id` is `''`

wdoconnell (approved):

PR #5318, opened 2022-08-05
fix: cap the estimated height to the screen height

### LINKED ISSUE #5317: RawDataView: toggling this on with a large data set can cause the browser to freeze
When a user with a large data set tries to use the RawDataView, which is backed by the SimpleTable visualization, by toggling the "View Raw Data" option, the browser will freeze.

**One probable cause** is the ResizeObserver in one of the subcomponents of SimpleTable. The ResizeObserver tries to observe the element with the class "visualization--simple-table--results", and estimate the initial height of this element for use by other code. This element has potentially a very large height proportional to the size of the data set. Thus, further calculations based on a large height make take a very long time and cause the browser to freeze.

**Proposal:** cap the estimated height to the maximum height of the screen in pixels. This will prevent runaway calculations that aren't necessary.

Data set example pending.

TCL735:
Closes #5317  

Cap the estimated initial height. This prevents further calculations on ridiculously large heights that wouldn't fit everything on the screen anyways.

### DIFF
--- a/src/visualization/types/SimpleTable/PagedTable.tsx
+++ b/src/visualization/types/SimpleTable/PagedTable.tsx
@@ -278,7 +278,9 @@ const PagedTable: FC<Props> = ({result, properties}) => {
 
       timeout = setTimeout(() => {
         animationFrameID = requestAnimationFrame(() => {
-          setHeight(entries[0].contentRect.height)
+          setHeight(
+            Math.min(entries[0].contentRect.height, window.screen.height)
+          )
         })
       }, 200)
     })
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):
if this doesn't fix the browser freezing on large datasets, will this change stay in or be reverted?

TCL735:
> if this doesn't fix the browser freezing on large datasets, will this change stay in or be reverted?

I think it should stay regardless of whether it fixes the browser freezing problem. No element that should be rendered should be taller than the screen height without a special reason. I can't think of a reason why we want this element to be potentially taller than the screen height.

PR #5322, opened 2022-08-08
chore: remove fastFromFlux flag

asalem1:
This PR removes the fastFromFlux feature flag. The purpose of this feature flag was to allow ongoing performance work to the fromFlux CSV parser without breaking UI functionality. This work gave us the opportunity to iterate upon the fromFlux parser incrementally without introducing new breaking changes, while also enabling us to have a failsafe in case something did go wrong with the new implementation. Since the performance gains have been significant, have been ported over to the fromFlux parser, AND since there are no plans in the foreseeable future to iterate on the fromFlux parser after the latest changes, this PR simply removes the conditional logic that was put in place to support the stability of the UI

### DIFF
--- a/src/alerting/utils/activity.ts
+++ b/src/alerting/utils/activity.ts
@@ -1,8 +1,7 @@
 // Utils
 import {runQuery} from 'src/shared/apis/query'
-import {fromFlux, fastFromFlux} from '@influxdata/giraffe'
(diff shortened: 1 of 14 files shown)


drdelambre (approved):
never got an answer on why parsing is so expensive, but we should address that in the parser and not by caching the parsed results. pr looks good.

Issue #5234, opened 2022-07-28
Dashboard Performance: memoize selectors


hoorayimhelping:
Common selectors are re-running frequently, which cause unnecessary re-renders, which cause the selectors to re-run unecessarily.

hoorayimhelping:
Have done a few small pieces of work which should clean up the main components of the application and make future work more effective.

PR #5326, opened 2022-08-08
perf(GetOrganizations): move selector creation to outside of component

hoorayimhelping:
Connects #5234

Moves selectors created in `GetOrganizations.tsx` into their own selector files. There probably isn't any performance to be gained from this, but it makes things easier and simpler in general, especially for future work when trying to memoize selectors.

From the react-redux docs:

>When using `useSelector` with an inline selector as shown above, a new instance of the selector is created whenever the component is rendered. This works as long as the selector does not maintain any state.

https://react-redux.js.org/api/hooks

### Checklist

### DIFF
--- a/src/identity/selectors/index.ts
+++ b/src/identity/selectors/index.ts
@@ -1,9 +1,13 @@
 // Types
 import {AppState} from 'src/types'
+import {RemoteDataState} from '@influxdata/clockface'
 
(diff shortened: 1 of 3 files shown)


ChitlangeSahas (approved):
This makes sense to me and looks good 👍

PR #5333, opened 2022-08-08
chore: remove the authSessionCookie flag

asalem1:
Related to https://github.com/influxdata/ui/pull/5330, removing the `authSessionCookie` flag in this PR since the partitioned UI logic is no longer relevant

### DIFF
--- a/src/Logout.tsx
+++ b/src/Logout.tsx
@@ -7,13 +7,7 @@ import {useHistory} from 'react-router-dom'
 import {postSignout} from 'src/client'
 
 // Constants
-import {
-  CLOUD,
-  CLOUD_URL,
-  CLOUD_LOGOUT_PATH,
-  CLOUD_SIGNOUT_PATHNAME,
-} from 'src/shared/constants'
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {CLOUD, CLOUD_SIGNOUT_PATHNAME} from 'src/shared/constants'
 
 // Components
 import {reset} from 'src/shared/actions/flags'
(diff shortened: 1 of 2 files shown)


src/Logout.tsx
```diff
@@ -23,31 +17,15 @@ const Logout: FC = () => {
...
-          handleReset()
-        }
+        const url = new URL(
+          `${window.location.origin}${CLOUD_SIGNOUT_PATHNAME}`
+        ).href
 
         window.location.href = url
         return
       } else {
-        if (isFlagEnabled('authSessionCookieOn')) {
```
wdoconnell, line 48: What was this conditional doing before? Since already `!CLOUD`, is this an OSS flag? I don't see it there, but if it is, I think we need to check with edge before we remove this code.

src/Logout.tsx
```diff
@@ -23,31 +17,15 @@ const Logout: FC = () => {
   const dispatch = useDispatch()
 
   useEffect(() => {
-    const handleReset = () => {
-      dispatch(reset())
-      dispatch({type: 'USER_LOGGED_OUT'})
```
wdoconnell, line 28: I think this action should be explicitly declared somewhere. We have `reset()` declared (and are importing it from the flag action creator), but aren't importing the `USER_LOGGED_OUT` action, which makes it confusing to trace if you're trying to follow dispatched actions through the reducer logic.

wdoconnell (changes requested):
Cloud portion looks good to me, but I think it's still needed to check on the code block I flagged re: whether removing the `authSessionCookieOn` flag would impact OSS.

src/Logout.tsx
wdoconnell: Looks good to me, understanding that this is a `CLOUD`-only flag.

wdoconnell (approved):

PR #5344, opened 2022-08-09
feat(deepLinks): make deeplinking work with multi org

### LINKED ISSUE #5138: Multi-Org Phase 1 - Fix Deep Linking to work with multi-org
When a user uses the deep linking code, and is NOT logged in, then update the current code to retrieve the default account's  default org. Then redirect the user to the link within that account/org.

Current behavior: Uses IDPE getOrg(), which will be deprecated.

blegesse-git:
Closes #5138 

When a user uses the deep linking code, and is NOT logged in, pr code retrieves the default account's default org. Then redirects user to the link within that account/org.

Attached video uses quartz remocal to show how logged out user gets redirected to page associated with default account's default org id using deeplinking. 

https://user-images.githubusercontent.com/66275100/183775501-29622b68-7848-455b-9704-a7d556841f8b.mov

### Checklist


src/identity/apis/auth.ts
```diff
@@ -283,3 +287,20 @@ export const updateDefaultOrgByAccountID = async ({
...
+export const getDefaultAccount = async () => {
+  try {
+    const resp = await getAccounts({})
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
+    }
+    const {data} = resp
+
+    if (Array.isArray(data) && data.length) {
+      return data.find(line => line.isDefault)
```
wdoconnell, line 301: Thanks for adding this function to the API layer!  This was going to be saved for the switch-account refactor, but I appreciate you doing it now.

Two other minor suggestions - I might pick some more descriptive var than `line` (like `acct` or `account`). And I might leave eventing out, since how we event might depend on how this function is called. At least for the new quartz APIs, we don't do that in other parts of the API layer right now - instead, we handle eventing in a thunk, or at the component level.

src/shared/components/NotFound.tsx
```diff
@@ -136,10 +140,18 @@ const NotFound: FC = () => {
...
-      setIsFetchingOrg(true)
-      org.current = await fetchOrg()
-    }
+      if (shouldUseQuartzIdentity()) {
+        setIsFetchingOrg(true)
+        const defaultAccount = await getDefaultAccount()
+        const quartzOrgs = await fetchOrgsByAccountID(defaultAccount.id)
 
+        const deafultQuartzOrg = quartzOrgs.find(org => org.isDefault)
+        org.current = deafultQuartzOrg
```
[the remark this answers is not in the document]
wdoconnell: Shouldn't, but I think we should probably handle it anyway, especially if quartz behavior could change. My recommendation would be that if find doesn't find anything that `isDefault` (returning `undefined`), we returned the first account in the array.

src/shared/components/NotFound.tsx
```diff
@@ -136,10 +140,18 @@ const NotFound: FC = () => {
 
   const handleDeepLink = useCallback(async () => {
     if (!org.current) {
-      setIsFetchingOrg(true)
-      org.current = await fetchOrg()
-    }
+      if (shouldUseQuartzIdentity()) {
+        setIsFetchingOrg(true)
+        const defaultAccount = await getDefaultAccount()
```
wdoconnell, line 145: Once you implement the change Bucky's suggesting to the function in the API layer, this will have the potential to throw an error if there's a problem. At that point, I think you'll need to add a `try/catch` block here, and then any eventing or other error handling can happen in the `catch`.

src/identity/apis/auth.ts
```diff
@@ -283,3 +287,20 @@ export const updateDefaultOrgByAccountID = async ({
 
   // success status code is 204; no response body expected.
 }
+
+// fetch user's default account
+export const getDefaultAccount = async () => {
+  try {
+    const resp = await getAccounts({})
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
```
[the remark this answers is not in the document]
blegesse-git: done and done :)

src/shared/components/NotFound.tsx
blegesse-git: okay that a really good idea. Is there some other way we can handle error other than `console.error()` here? 🤔

src/identity/apis/auth.ts
blegesse-git: yup totally meant to change the naming to something semantic. Dang it! I've updated it to `account`

src/shared/components/NotFound.tsx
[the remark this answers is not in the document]
blegesse-git: Based on my limited research, one of the org objects will always have a `isDefault` value of `true`. But totally agree with Bill, we should have a fall back here and use one of the other orgs.

src/identity/apis/auth.ts
```diff
@@ -283,3 +285,21 @@ export const updateDefaultOrgByAccountID = async ({
...
+  }
+
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+  const {data} = response
+
+  if (Array.isArray(data) && data.length) {
+    return data.find(account => account.isDefault)
+  }
```
hoorayimhelping, line 304: This is a much nicer implementation!

However, there are still two cases that will result in `undefined` being returned. `Array.isArray(data) && data.length` returning `false` (which will reach the end of the function without returning anything, which will implicitly return `undefined`), and `data.find(account => account.isDefault)` finding no `account` with `isDefault` set to true. Both of these are very unlikely, but they're still possible, so they should probably be accounted for.

I think doing something like this is probably the easiest way to handle it:

```ts
if (Array.isArray(data) && data.length) {
  const defaultAccount = data.find(account => account.isDefault)
  if (defaultAccount) {
    return defaultAccount
  }
}

throw new GenericError('No default account found') // just a suggestion on the error message. If you can think of a better one, by all means, use that!
```

This way, if everything works, we return the default account, and if not, we fall through and throw the `GenericError`

src/shared/components/NotFound.tsx
```diff
@@ -136,10 +140,23 @@ const NotFound: FC = () => {
...
-      org.current = await fetchOrg()
-    }
+      if (shouldUseQuartzIdentity()) {
+        try {
+          setIsFetchingOrg(true)
+          const defaultAccount = await getDefaultAccount()
+          const quartzOrgs = await fetchOrgsByAccountID(defaultAccount.id)
 
+          const deafultQuartzOrg =
+            quartzOrgs.find(org => org.isDefault) || quartzOrgs[0]
```
hoorayimhelping, line 150: Very nice work and very clean code! 🎉 

It might make for cleaner code at this level to combine these two calls into a single function inside the `identity/api/auth` layer, but I'll leave it to your judgement.

src/shared/components/NotFound.tsx
```diff
@@ -136,10 +140,23 @@ const NotFound: FC = () => {
...
+          org.current = deafultQuartzOrg
+        } catch (error) {
+          console.error(error)
```
hoorayimhelping, line 153: We're trying to move away from using `console.error` anywhere in code directly.

I'm not 100% certain on this, but I think if we reach this point, we might want to redirect the user to the no orgs page. Doing something like [this](https://github.com/influxdata/ui/blob/master/src/shared/containers/SetOrg.tsx#L116) is probably what we want to do here.

src/utils/deepLinks.ts
```diff
@@ -1,7 +1,10 @@
 import {Organization} from 'src/client'
+import {OrganizationSummaries} from 'src/client/unityRoutes'
 import {PROJECT_NAME_PLURAL} from 'src/flows'
 
-export const buildDeepLinkingMap = (org: Organization) => ({
+export const buildDeepLinkingMap = (
+  org: Organization | OrganizationSummaries[number]
```
hoorayimhelping, line 6: 👍

You don't have to worry about this, but just for your own information when it comes to software design: I think that I'm going to refactor `buildDeepLinkingMap` to take in an `orgId` rather than an `org`. Right now, the function needs to know the structure of the `Organization` type to function (it needs to know that it has an `id` and how to access it). It shouldn't need to worry about that, since every line in this function only operates on the org id. It would be a cleaner function from a design perspective to pass in `orgId`, since that's all the function needs. This is better because if the structure of `Organization` needs to change, this function won't have to change with it, since it won't care about where org id comes from.

hoorayimhelping (commented):
This is solid work! I have a couple more minor suggestions! Almost there!!

src/shared/components/NotFound.tsx
blegesse-git: great idea. Yeah, routing user to `/no-orgs` page makes sense if for some reason our api fails to fetch org id.

src/utils/deepLinks.ts
blegesse-git: That's definitely the right move. It will make the function reusable with other types of structures if we were to limit the argument to `orgId` instead of the org object in its entirety.

src/identity/apis/auth.ts
```diff
@@ -283,3 +284,31 @@ export const updateDefaultOrgByAccountID = async ({
...
+    // fetch default org
+    if (defaultAccount) {
+      const quartzOrg = await fetchOrgsByAccountID(defaultAccount.id)
+      const defaultQuartzOrg =
+        quartzOrg.find(org => org.isDefault) || quartzOrg[0]
+      return defaultQuartzOrg
+    }
+  }
+
+  throw new GenericError('No default account found')
```
hoorayimhelping, line 313: I think I told you to write this so it's my mistake (listening to my adivce was your mistake 😉), but I think this should say `No default organization found`

hoorayimhelping (approved):
Great work! I love seeing you taking on more challenging problem like this and solving them with gusto and aplomb!

src/identity/apis/auth.ts
blegesse-git: haha, I did copy paste your suggestion ...BUT i dont think you were wrong. If `if (Array.isArray(data) && data.length)` returns false we want to return an error that says something like `no account found` because it means we were able to fetch a response but the response had relevant data in it? Taking a second look at this is making me think maybe we should also wrap `const quartzOrg = await fetchOrgsByAccountID(defaultAccount.id)` in a `try{}catch{}` ... maybe something like this? 
Screen Shot 2022-08-10 at 11 28 46 AM

hoorayimhelping: `fetchOrgsByAccountID` will throw an error on its own if it fails, which will bubble up to the code calling this function if it's not caught here, which means this function doesn't **need** to catch the error, but it can if it wants.

The `try/catch` here will catch an error thrown by `fetchOrgsByAccountID`, but it will still return `undefined` if `defaultQuartzOrg` comes back as `undefined` by the `find` function not finding anything, so this code will still need to account for that.

hoorayimhelping (approved):

PR #5349, opened 2022-08-09
feat(5342): schema composition coordination between LSP, monaco-editor, and UI session store

### LINKED ISSUE #5342: Schema Composition: coordination between the LSP vs monaco-editor vs UI
This is the Q3 PoC solution.
(Not the middleware addition, to be done in Q4.)

wiedld:
Closes #5342

## Video of PoC

Last minute change not in vid: once we diverge (type in the composition block), the entire blue block gets removed.
https://user-images.githubusercontent.com/10232835/185288812-c36ede2a-7feb-4846-88a7-484f336e304a.mp4

## Feature requirements
* coordinate communication btwn: LSP, UI session state, and monaco-editor style (for composiiton block).
* do this^^, while keeping in mind the division of responsibilities:
   * LSP = creation of valid flux (the injection itself)
   * session state = UI stateful src of truth, for schema builder (outside of editor)
   * monaco-editor = holds the styles applied (visually) to the code block.
        * And is src of truth for current query string.
* everything behind feature flag `schemaComposition`

## Not done:
* minor style updates for the sync affordance (within the monaco-editor block).
   * right now, it's a solid blue block


src/dataExplorer/context/persistance.tsx
```diff
@@ -4,40 +4,48 @@ import {DEFAULT_TIME_RANGE} from 'src/shared/constants/timeRanges'
...
 
   setHorizontal: (val: number[]) => void
   setVertical: (val: number[]) => void
   setRange: (val: TimeRange) => void
   setQuery: (val: string) => void
-  setSelection: (val: MeasurementSelection) => void
+  setSelection: (val: SchemaSelection) => void
 }
 
+export const DEFAULT_SCHEMA = Object.freeze({
```
wiedld, line 31: Object.freeze() means that it cannot be updated. So no one accidentally passing in the same pointer object to be used again and again. Instead, must clone into a mutable object.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -91,6 +107,161 @@ class LspConnectionManager {
...
+      false,
+      schema.composition?.startLine || 1,
+      schema.composition?.startLine || 1
+    )
+  }
+
+  // hack for now. To trigger for example video.
+  private _turnOnWithFirstCall = true
+
+  onSchemaSessionChange(schema: SchemaSelection, sessionCb) {
```
wiedld, line 197: This is the method, which subscribes to all session state changes.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -91,6 +107,161 @@ class LspConnectionManager {
...
+    if (this._turnOnWithFirstCall) {
+      this._turnOnWithFirstCall = false
+      setTimeout(() => {
+        console.error('MOCK SCHEMA INJECTION FROM USER')
+        this._toggleSchemaSync()
+      }, 5000)
+    }
+  }
+
+  onLspCompositionChange(_jsonrpcMiddlewareResponse: unknown) {
```
wiedld, line 221: this is the method, which will be subscribing to all LSP outbound communications (after middleware is in place).

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -91,6 +107,165 @@ class LspConnectionManager {
...
+            range: new MonacoTypes.Range(startLine, 1, endLine, 1),
+            options: {
+              isWholeLine: true,
+              inlineClassName: 'composition-block',
+            },
+          },
+        ]
+      : []
+
+    this._compositionStyle = this._editor.deltaDecorations(
```
wiedld, line 263: I have a whole writeup on why `editor.deltaDecorations()` was chosen for composition block styling. Please feel to ask for details.

FyI: assessed, tested, and rejected options were:
* LSP `textDocument/documentColor` and `textDocument/colorPresentation`
* LSP `textDocument/documentHighlight`
* `monaco.editor.colorizeModelLine()`
* `monaco.editor.colorizeElement()`
* `monaco.editor.colorize()`

src/shared/components/FluxMonacoEditor.scss
```diff
@@ -42,4 +42,27 @@
   .current-line ~ .line-numbers {
     color: #bec2cc;
   }
+
+  // TODO: proper style
```
wiedld, line 46: All the styles in this file will change, based on feedback from Julia & co.

src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -65,9 +74,23 @@ const FluxEditorMonaco: FC<Props> = ({
     connection.current.updatePreludeModel(variables)
   }, [variables])
 
+  useEffect(() => {
+    if (
+      connection.current &&
+      fluxQueryBuilder &&
+      isFlagEnabled('schemaComposition')
```
wiedld, line 81: This is the feature flagged turning on of schema composition. This is where the `LspConnectionManager` subscribes to the session store.

Note to @appletreeisyellow , we'll also need to feature flag (using this flag) the turning off of the old style of schema injection.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -18,11 +25,22 @@ import {
...
 class LspConnectionManager {
   private _worker: Worker
+  private _editor: EditorType
   private _model: MonacoTypes.editor.IModel
   private _preludeModel: MonacoTypes.editor.IModel
   private _variables: Variable[] = []
+  private _canInitComposition: boolean = true // cannot re-init, if raw edits
+  private _compositionStyle: string[] = []
+  private _session: SchemaSelection
+  private _callbackSetSession: (schema: SchemaSelection) => void = () => null
```
wiedld, line 43: Reasons for state saved. (Feel free to suggest ways to avoid.)
* `_canInitComposition` == right now, the irreversiblbility of composition init is on the UI side
* `_compositionStyle` == required by monaco-editor
   * must provide both previous style, and new style, to the method. (It calc's the diff.)
* `_session`  and `_callbackSetSession`
  * session is our src of truth
  * diff'ing previous vs current
  * callback to set state on session (if need to change anything)

src/shared/components/FluxMonacoEditor.tsx
wiedld: What we **will need to do** is: 
1. have selected fields, tags, and tag_values be stored in the session state
2. update `LspConnectionManager.onSchemaSessionChange()` to detect the diff.
3. upon diff --> we call `LspConnectionManager.inject()`

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -91,6 +111,193 @@ class LspConnectionManager {
...
+    this._callbackSetSession({
+      ...this._session,
+      composition: {...this._session.composition, synced: true},
+    })
+  }
+
+  _setCompositionSyncToggle() {
+    setTimeout(() => {
+      const syncIcon = document.getElementsByClassName(ICON_SYNC_ON)[0]
+      const clickableInvisibleDiv = document.getElementById(ICON_SYNC_ID)
```
wiedld, line 127: Right now, the icon is placed within the monaco-editor by passing classnames to the editor, and asking the classname be applied (per line in the code block). The icon is added to line 1. But this icon is not clickable.

In order to create a clickable element with a handler, we place an invisible div over the icon. In this way we get the best of both worlds:
* the monaco-editor itself places the icon on the right line, so it always gets the proper top+left position regardless of responsive re-sizing.
* Then we use that exact top+left position --> to then set the invisible clickable div exactly correct.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -91,6 +111,193 @@ class LspConnectionManager {
...
+    }
+
+    // msg to LSP
+    const {bucket, measurement} = schema
+    const payload = {bucket: bucket?.name}
+    if (measurement) {
+      payload['measurement'] = measurement
+    }
+    // TODO: finish LSP update first
+    // this.inject(ExecuteCommand.CompositionInit, payload)
```
wiedld, line 195: After the LSP schema composition service is online, then we call `this.inject()` with the proper payload.

src/shared/contexts/editor/index.tsx
```diff
@@ -66,7 +66,9 @@ export const EditorProvider: FC = ({children}) => {
   const injectViaLsp = useCallback(
     (cmd, data: Omit<ExecuteCommandArgument, 'textDocument'>) => {
       try {
-        connection.current.inject(cmd, data)
+        if (connection.current && !isFlagEnabled('schemaComposition')) {
+          connection.current.inject(cmd, data)
+        }
```
wiedld, line 71: We do not need to featureFlag the existing use of `injectViaLsp()` in the schema browser. Since it's all feature flagged in this single place.

src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -102,6 +125,7 @@ const FluxEditorMonaco: FC<Props> = ({
 
   return (
     <ErrorBoundary>
+      <div id={ICON_SYNC_ID} style={{zIndex: 999, position: 'absolute'}} />
```
wiedld, line 128: This is the invisible div, which will be re-positioned in `_setCompositionSyncToggle()`

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -91,6 +111,193 @@ class LspConnectionManager {
...
+      setTimeout(() => {
+        console.error('MOCK SCHEMA INJECTION FROM USER')
+        this._toggleSchemaSync()
+      }, 5000)
+    }
+  }
+
+  onLspCompositionChange(_jsonrpcMiddlewareResponse: unknown) {
+    // TODO: middleware to read LSP response => call this method
+    // TODO: wait for successful LSP roundtrip, then use response for:
```
wiedld, line 254: `onLspCompositionChange()` = this is the method, which will be subscribing to all LSP outbound communications (after the middleware is in place --> Q4 goal).

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -18,11 +22,23 @@ import {
...
+
 class LspConnectionManager {
   private _worker: Worker
+  private _editor: EditorType
   private _model: MonacoTypes.editor.IModel
   private _preludeModel: MonacoTypes.editor.IModel
   private _variables: Variable[] = []
+  private _compositionStyle: string[] = []
+  private _session: SchemaSelection
+  private _callbackSetSession: (schema: SchemaSelection) => void = () => null
```
wiedld, line 41: Reasons for state saved. (Feel free to suggest ways to avoid.)
* `_compositionStyle` == required by monaco-editor
   * must provide both previous style, and new style, to the method. (It calc's the diff.)
* `_session`  and `_callbackSetSession`
  * session is our src of truth
  * diff'ing previous vs current
  * callback to set state on session (if need to change anything)

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -91,6 +108,202 @@ class LspConnectionManager {
...
+      ...this._session,
+      composition: {...this._session.composition, synced},
+    })
+  }
+
+  _setCompositionSyncToggle() {
+    setTimeout(() => {
+      const syncIconOn = document.getElementsByClassName(ICON_SYNC_ON)[0]
+      const syncIconOff = document.getElementsByClassName(ICON_SYNC_OFF)[0]
+      const syncIcon = syncIconOn || syncIconOff
```
wiedld, line 122: Right now, the icon is placed within the monaco-editor by passing classnames to the editor, and asking the classname be applied (per line in the code block). The icon is added to line 1. But this icon is not clickable.

In order to create a clickable element with a handler, we place an invisible div over the icon. In this way we get the best of both worlds:
* the monaco-editor itself places the icon on the right line, so it always gets the proper top+left position regardless of responsive re-sizing.
* Then we use that exact top+left position --> to then set the invisible clickable div exactly correct.

rockstar:
Removing my review request so that Slack doesn't send me reminders for a PR that isn't ready yet. Please re-add me when it's ready.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -91,6 +109,203 @@ class LspConnectionManager {
...
+        options: {
+          linesDecorationsClassName: ICON_SYNC_CLASSNAME,
+        },
+      },
+    ]
+
+    this._compositionStyle = this._editor.deltaDecorations(
+      this._compositionStyle,
+      startLineStyle.concat(endLineStyle)
+    )
```
wiedld, line 211: I have a whole writeup on why `editor.deltaDecorations()` was chosen for composition block styling. Please feel to ask for details.

FyI: assessed, tested, and rejected options were:
* LSP `textDocument/documentColor` and `textDocument/colorPresentation`
* LSP `textDocument/documentHighlight`
* `monaco.editor.colorizeModelLine()`
* `monaco.editor.colorizeElement()`
* `monaco.editor.colorize()`

src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -102,6 +130,10 @@ const FluxEditorMonaco: FC<Props> = ({
 
   return (
     <ErrorBoundary>
+      <div
+        id={ICON_SYNC_ID}
+        style={{zIndex: 999, position: 'absolute', opacity: 0.6}}
+      />
```
wiedld, line 136: This is the invisible div, which will be re-positioned in `_setEditorSyncToggle()`

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -18,11 +22,24 @@ import {
...
   private _worker: Worker
+  private _editor: EditorType
   private _model: MonacoTypes.editor.IModel
   private _preludeModel: MonacoTypes.editor.IModel
   private _variables: Variable[] = []
+  private _compositionStyle: string[] = []
+  private _session: SchemaSelection
+  private _callbackSetSession: (
+    schema: RecursivePartial<SchemaSelection>
+  ) => void = () => null
```
wiedld, line 42: Reasons for state saved. (Feel free to suggest ways to avoid.)
* `_compositionStyle` == required by monaco-editor
   * must provide both previous style, and new style, to the method. (It calc's the diff.)
* `_session`  and `_callbackSetSession`
  * session is our src of truth
  * diff'ing previous vs current
  * callback to set state on session (if need to change anything)

src/languageSupport/languages/flux/lsp/utils.ts
```diff
@@ -147,4 +153,5 @@ export enum ExecuteCommand {
   InjectField = 'injectFieldFilter',
   InjectTag = 'injectTagFilter',
   InjectTagValue = 'injectTagValueFilter',
+  CompositionInit = 'composition/initialize',
```
wiedld, line 156: We'll need to support both the old and new style of schema injection. (Feature flagged for new schema composition.)

This means that the execute_command() method in the LSP server (receiving the `workspace/executeCommand` payloads) will need to be extended to handle other methods.

Here I placed an example `'composition/initialize'` method. But we can make it be whatever you suggest @rockstar .

src/dataExplorer/context/persistance.tsx
```diff
@@ -1,43 +1,57 @@
...
   query: string
-  selection: MeasurementSelection
+  selection: SchemaSelection
 
   setHorizontal: (val: number[]) => void
   setVertical: (val: number[]) => void
   setRange: (val: TimeRange) => void
   setQuery: (val: string) => void
-  setSelection: (val: MeasurementSelection) => void
+  setSelection: (val: RecursivePartial<SchemaSelection>) => void
```
appletreeisyellow, line 30: Why `RecursivePartial<SchemaSelection>` instead of just `SchemaSelection`? From my understanding, `RecursivePartial` is for recursive data structure, but `SchemaSelection` has a set data structure instead of recursive.

src/languageSupport/languages/flux/lsp/connection.ts
appletreeisyellow: Thanks for digging into the Monaco Editor APIs!

appletreeisyellow (approved):
Thank you for setting up the bridge between LSP, Monaco editor, and UI! 🙌 I can see much of the work came out from the digging and research in Monaco editor APIs. 

No blockers, just some comments and questions for my understanding

src/languageSupport/languages/flux/lsp/utils.ts
wiedld: He already has that whole LSP spec, which states the methods to use.

Followup PR will use those methods. That PR will simply be:
* in the `LspConnectionManager` methods, write a few lines to:
(1) construct the json blob, with the method (per spec)
(2) use existing `LspConnectionManager.inject()` to send to LSP.
(3) and that's it folks.

src/dataExplorer/context/persistance.tsx
```diff
@@ -1,43 +1,57 @@
...
 import {DEFAULT_TIME_RANGE} from 'src/shared/constants/timeRanges'
 import {useSessionStorage} from 'src/dataExplorer/shared/utils'
 import {Bucket} from 'src/types'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
-interface MeasurementSelection {
+interface SchemaComposition {
+  synced: boolean
+  diverged: boolean
+}
```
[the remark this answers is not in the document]
wiedld: Great idea!

The synced thing is weird. Since I have trouble imagining the use case for sync'ing as it stands now. This is what I take it to be:
```
interface SchemaComposition {
  synced: boolean // defined == cannot modify session's schema
  diverged: boolean // defined == cannot re-sync. (e.g. user has typed in the composition block)
}
```

After convo with @juliajames12 tmrw, I'll push comments into this code.

src/dataExplorer/context/persistance.tsx
wiedld: The `setSchemaSelection()` method in the persistence context takes in this recursive partial. Meaning:
* we have a nested object
* we want to only update part of the nested object
* therefore, the method is updated such that we pass in only the nested part...and the `setSchemaSelection()` updates only that nested prop.

What we have been doing is:
* Passing in an object which is: `setSelection({...old, ...new})`, at the level of many different components.
* In order to make that^^ work, we have a code pattern where we pass the old/previous session object around to these components.
* This works...until you hit possible points where we are not memoizing correctly.

Since we plan to eventually use the new ScriptBuilder everywhere, (which will include many other devs using memorization), it felt prudent to defensively encode against this exposure to memoziation. That is why I updated the `PersistenceContext.setSchemaSelection()` to take it's own state as the `...old`, and we only pass in the partial recursive object for the `...new`.

Does this make sense?

appletreeisyellow: Got it. Thanks for explaining!

PR #5352, opened 2022-08-09
chore: cleaning up linter warnings in cliWizard

alexjw3:
Subset of https://github.com/influxdata/ui/pull/5199.

### Checklist


src/homepageExperience/components/steps/cli/InitializeClient.tsx
```diff
@@ -79,18 +79,18 @@ export const InitializeClient: FC<OwnProps> = ({
...
-  }, []) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [dispatch])
 
   useEffect(() => {
     onSelectBucket(bucket.name)
   }, [bucket, onSelectBucket])
 
   useEffect(() => {
-    const fetchResources = async () => {
-      await dispatch(getAllResources())
```
wdoconnell, line 90: Same here - safe to remove `await` because it isn't being used to hold up any other code.

src/homepageExperience/components/steps/cli/InstallDependencies.tsx
```diff
@@ -61,34 +52,6 @@ sudo cp influxdb2-client-latest-linux-arm64/influx /usr/local/bin/
     'Mac'
   )
 
-  useEffect(() => {
```
wdoconnell, line 64: What's the rationale for proposing removal of these hooks? Are they not used?

wdoconnell (changes requested):
One question re: proposal to remove hooks on InstallDependencies.

src/homepageExperience/components/steps/cli/InstallDependencies.tsx
```diff
@@ -10,18 +10,10 @@ import {
 } from '@influxdata/clockface'
 
 import React, {FC, useEffect, useState} from 'react'
-import {useDispatch} from 'react-redux'
-import {getBuckets} from 'src/buckets/actions/thunks'
 import {event} from 'src/cloud/utils/reporting'
 import {keyboardCopyTriggered, userSelection} from 'src/utils/crossPlatform'
 
 export const InstallDependencies: FC = () => {
-  const dispatch = useDispatch()
```
wdoconnell, line 19: @alexjw3 @david-rusnak Is it correct that any state retrieved from `getBuckets()` isn't necessary here?

src/homepageExperience/components/steps/cli/InitializeClient.tsx
hoorayimhelping: follow up - if the `async/await` is removed, then there's no point to call `dispatch(getAllResrouces())` inside the `fetchAllResources` function. Remove `fetchAllResources` and just call `dispatch(getAllResrouces())` in the body of the hook's anonymous function

wdoconnell: @alexjw3 In this PR, could you also make the same cleanup to `Tokens.tsx`, which has equivalent code (no need for a helper function for `dispatch`, since we're not `await`ing it anymore? Thanks!

src/homepageExperience/components/steps/cli/InstallDependencies.tsx
david-rusnak: Yes, that is unnecessary!

src/homepageExperience/components/steps/cli/InitializeClient.tsx
alexjw3: Makes sense.

wdoconnell (approved):
Looks good - thanks for updating!

PR #5357, opened 2022-08-10
fix: disable form if not pay as you go or free flag customer

helenosheaa:
Closes #https://github.com/influxdata/data-acquisition/issues/494 && https://github.com/influxdata/data-acquisition/issues/478

Interim change to disable subscription form for free users to alleviate some confusion.

- makes upgrade button same color as the rest of the app
- disables form elements
- adds optional variables to the shared create bucket component so that we can make it disabled but not require changes elsewhere in the codebase
- fixes placeholder text bug

Before:

https://user-images.githubusercontent.com/38860767/183917553-f72ae38e-5bd2-45ca-a2cb-80991d4b9c47.mov

After:

https://user-images.githubusercontent.com/38860767/183917606-48e0d276-bba4-4d12-9f66-f47ae656b920.mov

changed color of text and fontweight since screenrecording:
Screen Shot 2022-08-10 at 12 52 37 PM

Placeholders:
Screen Shot 2022-08-10 at 10 25 50 AM
Screen Shot 2022-08-10 at 10 26 08 AM

### Checklist

### DIFF
--- a/src/writeData/components/WriteDataHelperBuckets.tsx
+++ b/src/writeData/components/WriteDataHelperBuckets.tsx
@@ -22,11 +22,13 @@ import {
 interface Props {
   className?: string
   useSimplifiedBucketForm?: boolean
+  disabled?: boolean
 }
 
 const WriteDataHelperBuckets: FC<Props> = ({
   className = 'write-data--details-widget-title',
   useSimplifiedBucketForm = false,
+  disabled = false,
 }) => {
   const {bucket, buckets, changeBucket} = useContext(WriteDataDetailsContext)
   const isSelected = (bucketID: string): boolean => {
@@ -68,7 +70,19 @@ const WriteDataHelperBuckets: FC<Props> = ({
     body = (
       <List
         backgroundColor={InfluxColors.Grey5}
-        style={{height: '200px'}}
+        style={
+          disabled
+            ? {
+                pointerEvents: 'none',
+                height: '200px',
+                cursor: 'not-allowed',
+                borderColor: '#1a1a2a',
+                backgroundColor: '#1a1a2a',
+                opacity: '0.5',
+                color: '#828294',
+              }
+            : {height: '200px'}
+        }
         maxHeight="200px"
         testID="buckets--list"
       >
@@ -76,9 +90,9 @@ const WriteDataHelperBuckets: FC<Props> = ({
           <List.Item
             size={ComponentSize.Small}
             key={b.id}
-            selected={isSelected(b.id)}
+            selected={disabled ? null : isSelected(b.id)}
             value={b}
-            onClick={changeBucket}
+            onClick={disabled ? () => {} : changeBucket}
             wrapText={true}
             gradient={Gradients.GundamPilot}
           >
@@ -96,6 +110,7 @@ const WriteDataHelperBuckets: FC<Props> = ({
         <CreateBucketButton
           useSimplifiedBucketForm={useSimplifiedBucketForm}
           callbackAfterBucketCreation={changeBucket}
+          disabled={disabled}
         />
       </Heading>
       {body}
--- a/src/writeData/components/WriteDataHelperBuckets.tsx
+++ b/src/writeData/components/WriteDataHelperBuckets.tsx
@@ -70,29 +70,18 @@ const WriteDataHelperBuckets: FC<Props> = ({
     body = (
       <List
         backgroundColor={InfluxColors.Grey5}
-        style={
-          disabled
-            ? {
-                pointerEvents: 'none',
-                height: '200px',
-                cursor: 'not-allowed',
-                borderColor: '#1a1a2a',
-                backgroundColor: '#1a1a2a',
-                opacity: '0.5',
-                color: '#828294',
-              }
-            : {height: '200px'}
-        }
+        style={{height: '200px'}}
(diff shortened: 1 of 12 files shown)


sjwang90:
Nice work!! Is there any way we can bold and color the "Upgrade Now" text under Connect to a Broker? For the short term, I would like to see if we could emphasize a tad more why all of this is disabled for free users. It won't be super obvious until the new design but I think highlighting that text could make it a little more obvious.

helenosheaa:
@sjwang90 yes good idea, done -->

Screen Shot 2022-08-10 at 11 44 52 AM



I think it makes it more obvious to users to look for that upgrade button. I know in the first iteration of new designs that upgrade button is actually going to be within the form instead but this is a good interim.

sjwang90 (approved):

src/writeData/components/WriteDataHelperBuckets.tsx
```diff
@@ -68,17 +70,29 @@ const WriteDataHelperBuckets: FC<Props> = ({
...
+                pointerEvents: 'none',
+                height: '200px',
+                cursor: 'not-allowed',
+                borderColor: '#1a1a2a',
+                backgroundColor: '#1a1a2a',
+                opacity: '0.5',
+                color: '#828294',
+              }
+            : {height: '200px'}
+        }
```
hoorayimhelping, line 85: It looks like the [List](https://influxdata.github.io/clockface/?path=/story/components-list-family--list) component will handle this styling for you if you pass `disabled={disabled}` into it

src/writeData/components/WriteDataHelperBuckets.tsx
```diff
@@ -68,17 +70,29 @@ const WriteDataHelperBuckets: FC<Props> = ({
...
         backgroundColor={InfluxColors.Grey5}
-        style={{height: '200px'}}
+        style={
+          disabled
+            ? {
+                pointerEvents: 'none',
+                height: '200px',
+                cursor: 'not-allowed',
+                borderColor: '#1a1a2a',
+                backgroundColor: '#1a1a2a',
```
hoorayimhelping, line 80: In general, we shouldn't use hex color literals in ui code, because these colors are defined by Clockface.

In a component, you'll want to import `InfluxColors` from clockface and use `InfluxColors.Grey15`, and in a stylesheet, you'd want to `@import '@influxdata/clockface/dist/variables.scss';` and use `$cf-grey-15`.

src/writeData/components/WriteDataHelperBuckets.tsx
ChitlangeSahas, line 85: I think we can move this into a variable or the stylesheet. I learned that it helps performance. 

something like : 
```
const listStyle = disabled
            ? {
                pointerEvents: 'none',
                height: '200px',
                cursor: 'not-allowed',
                borderColor: '#1a1a2a',
                backgroundColor: '#1a1a2a',
                opacity: '0.5',
                color: '#828294',
              }
            : {height: '200px'}
```      

and then using the variable in the component props.

src/writeData/subscriptions/components/BrokerForm.scss
```diff
@@ -76,6 +72,12 @@
   &__text {
     color: $cf-grey-95;
   }
+  &__upgrade-text {
+    background: -webkit-linear-gradient(45deg, #34bb55 0%, #00a3ff 100%);
```
hoorayimhelping, line 76:
```suggestion
    background: -webkit-linear-gradient(45deg, #34bb55 0%, $c-pool 100%);
```

it looks like `#34bb55` is defined in Clockface as `Rainforest` but it doesn't look like there's an scss variable for it.

src/writeData/components/WriteDataHelperBuckets.tsx
ChitlangeSahas: For reference : https://www.aaronbuxbaum.com/how-object-literals-hurt-performance/

hoorayimhelping (commented):
Some suggestions about making full use of Clockface things

src/writeData/components/WriteDataHelperBuckets.tsx
helenosheaa: ah - thanks! I'll definitely use that I tried handing it into List and that wasn't a prop that exists but handing it into List.Item does the trick.

src/writeData/components/WriteDataHelperBuckets.tsx
helenosheaa: gotcha, I'll remember that going forward, I'm going to remove this styling anyway and rely on the disabled styling that's already there.

src/writeData/components/WriteDataHelperBuckets.tsx
ChitlangeSahas: Yes I saw that, that was a good catch by Bucky!

helenosheaa:
> looks good as a temporary fix! longer term i'd like to find a different way to emphasize the "Upgrade Now" link as we don't typically do gradients on links.

it's actually not a link just emphasized text. I think it's just a short term quick improvement as we'll probably move that Upgrade Button to that location per your new designs in the next round.

ChitlangeSahas (approved):
Looks good to me, pending our conversation about the upgrade button font weight, good to merge 👍

PR #5369, opened 2022-08-11
feat: add certificate support for subscriptions

subirjolly:
Closes # https://github.com/influxdata/data-acquisition/issues/441
This feature is behind a Feature Flag `enableCertificateSupport`.

**Switching to Certificate Security**
image

**Form Valid With `Certificate authority` Field With `Private Key` and `Certificate` Fields Empty**
image

**Form Invalid When Either Private Key or Certificate Missing**
image

**Form Valid With All Certificate Fields Filled In**
image

### Details Page
**When on Details Page in Read-Only Mode**
image

**When on Details Page in Edit Mode**
image

**Modal on `Replace Certificate` Button Clicked**
image

**Editing When Clicked on `Yes, Continue` Button**
image

### Checklist


src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -68,11 +68,6 @@ const BrokerFormContent: FC<Props> = ({
     })
   }, [])
 
-  useEffect(() => {
-    if (formContent.brokerUsername) {
-      setSecurity('user')
-    }
```
genehynson, line 74: In order to avoid creating a dependency cycle between this PR and the IDPE one, could we temporarily bring this change back and add the following conditional to handle the certificate scenario? That way this could be merged without `authType` existing. Don't lose the commit because I like what you've changed here with the `authType` integration but I think we could make that change in a followup PR.

```
    if (formContent.brokerUsername) {
      setSecurity('user')
    }
    if (formContent.brokerClientCert) {
      setSecurity('certificate')
    }
```

genehynson: nevermind, Subir was able to merge the authType property to nifid. So this no longer applies

src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
```diff
@@ -52,6 +52,8 @@ import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 // Styles
 import 'src/writeData/subscriptions/components/CreateSubscriptionPage.scss'
 
+// Contexts
+
```
genehynson, line 56: nit: I think you could remove this change to reduce the PR impact

genehynson (approved):
looks good to me. Could you add a screenshot of what the UI looks like with the `enableCertificateSupport` flag off?

subirjolly:
> looks good to me. Could you add a screenshot of what the UI looks like with the `enableCertificateSupport` flag off?

**With Flag Off**
image

**Flag On/Off**
Kapture 2022-08-23 at 16 21 38

src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
subirjolly: done

Issue #5307, opened 2022-08-05
Multi Org Phase 1 - Adjust global header style per design input


wdoconnell:
Design has reviewed the implemented global header behind the feature flag, and would like the following adjustments to the current style:

1. 'Switch Account' button should be adjusted so that it is no longer in a hover state while the typeahead dropdown is expanded.
2. Dropdown should 'hug' items if there are below a given number (e.g., no extra space in the dropdown if there are only two items, see below).
<img width="271" alt="Screen Shot 2022-08-04 at 5 30 59 PM" src="https://user-images.githubusercontent.com/91283923/183082268-a8575b37-2fc9-4571-aac8-06b08d48eee7.png">

3. Test changes to search bar background, to see if a color aside from black (e.g., matching gray style) looks better but is still legible.

@Luckette @juliajames12

ChitlangeSahas:
For #2 , the behavior seen is consistent across the Button Component in all over the App. I can fix it, but I think it's worth noting that it's not global header specific:

The hover state stays after click and needs a click outside to get rid of it.

https://user-images.githubusercontent.com/18511823/184192083-1777c1fa-54a6-4061-8d15-2124fc112682.mov

juliajames12:
I spoke with @ChitlangeSahas about this over slack, but my recommendation is that we treat the "Switch" buttons differently than the other buttons because this is a special use case where we are allowing users to drill down to a submenu vs. go back to the main menu. Since those are two separate actions I don't think the button should stay active.

PR #5376, opened 2022-08-11
fix(TypeAheadDropdown): dropdown hugs the item's list

ChitlangeSahas:
Connects #5307 

Added some logic to dynamically calculate the height of the dropdown. 

https://user-images.githubusercontent.com/18511823/184192436-90fef23a-ee85-42cb-80f3-709618fa5e24.mov

### Checklist

### DIFF
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
@@ -19,6 +19,7 @@ type State = {
 }
 
 export class GlobalHeaderTypeAheadMenu extends React.Component<Props, State> {
+  private listItemHeight = 50
   constructor(props: Props) {
     super(props)
     this.state = {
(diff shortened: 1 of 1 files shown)


src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
```diff
@@ -98,14 +99,18 @@ export class GlobalHeaderTypeAheadMenu extends React.Component<Props, State> {
       />
     )
 
+    // max height should be 150, so we take the min of calculated and 150
+    const calculateDropdownHeight = () =>
+      Math.min(queryResults.length * this.listItemHeight, 150)
```
hoorayimhelping, line 104: since this is a class componet, it's probably better to make this a private instance method on the class that takes the `queryResults.length` as an argument, rather than re-creating the function on every render call. I think it's also a good idea to make a variable to store the value `150` into so it provides context and isn't a magic number (even though it's probably fine since the comment explains what 150 means.

hoorayimhelping (approved):
lgtm, but a couple of minor suggestions

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
ChitlangeSahas: I agree, I fixed those things 👍

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
```diff
@@ -103,9 +108,11 @@ export class GlobalHeaderTypeAheadMenu extends React.Component<Props, State> {
         {filterSearchInput}
         {queryResults && queryResults.length > 0 ? (
           <List
-            height={style?.height ?? 150}
+            height={
+              style?.height ?? this.calculateDropdownHeight(queryResults.length)
+            }
             itemCount={queryResults.length}
-            itemSize={50}
+            itemSize={this.listItemHeight}
```
wdoconnell, line 115: Thank you for handling this!

PR #5379, opened 2022-08-11
refactor(deepLinks): deep links deal with org id instead of org

### LINKED ISSUE #5378: Deep Links: refactor so that the deep linking function accepts org id, not org
Currently the function to build the deep links based on the current organization id requires the entire Organization object, when the deep linking code only relies on `org.id`. This should be refactored so that the function takes in simpler values, rather than a complex object.  This is better because if the structure of `Organization` needs to change, the function won't have to change with it, since it won't care about where org id comes from.

### Acceptance Criteria
- make `buildDeepLinkingMap` only deal with `org.id`, rather than `org`

### LINKED ISSUE #5387: firstMile: CLI - Write Data Downloaded CSV --file path for MAC should be "downloads/"
Update code snippet for the downloaded csv on the write data page to be: 

influx write --bucket cli_test_2 --file downloads/air-sensor-data-annotated.csv

context (mac os)

hoorayimhelping:
Closes #5378 

Refactors `buildDeepLinkingMap` to only depend on `orgId`, rather than the entire `org` object.

### Checklist


src/shared/components/NotFound.tsx
```diff
@@ -154,7 +154,7 @@ const NotFound: FC = () => {
         org.current = await fetchOrg()
       }
     }
-    const deepLinkingMap = buildDeepLinkingMap(org.current)
+    const deepLinkingMap = buildDeepLinkingMap(org.current?.id)
```
hoorayimhelping, line 157: if this comes back as `undefined`, it will pass `undefined` into `buildDeepLinkingMap` which will cause a redirect to e.g. `/orgs/undefined/about` which will redirect to `/orgs`

blegesse-git (approved):

PR #5380, opened 2022-08-11
feat: arduino init client page

### LINKED ISSUE #5214: firstMile: Arduino - Initialize Client
**ACs**
    - When the user is on the Initialize Client step, highlight that portion of the wizard
    - Display the same text that's on the other wizards - unless advised otherwise by Sahas; note: the text may contain links
    - The user can use a copy code component to copy/paste code snippets
    -  When the user clicks the "Next" button, navigate the user to the next step in the wizard and checkmark the overview step as being completed (blue checkmark)
    - When the user clicks the "PREVIOUS" button, navigate the user to the previous step in the wizard

david-rusnak:
Closes #5214

Adds `Initialize Client` page to the arduino onboarding wizard. This page allows user to select or create a bucket to use for the tutorial. If the user selects or creates one, the code autopopulates with the bucket name. If the user does not select a bucket, `YOUR_BUCKET` is placed as a place holder. The code also autopopulates with a token, org-id, and host url.

https://user-images.githubusercontent.com/106361125/184224600-15be6717-9ebe-477f-bf81-8dcb7be797bb.mov

### Checklist

### DIFF
--- a/src/homepageExperience/components/steps/arduino/InitializeClient.tsx
+++ b/src/homepageExperience/components/steps/arduino/InitializeClient.tsx
@@ -7,17 +7,24 @@ import {
   createAuthorization,
   getAllResources,
 } from 'src/authorizations/actions/thunks'
+import {getBuckets} from 'src/buckets/actions/thunks'
+
+// Components
+import CodeSnippet from 'src/shared/components/CodeSnippet'
+import {
+  Columns,
+  ComponentSize,
+  Grid,
+  InfluxColors,
+  Panel,
+} from '@influxdata/clockface'
+import WriteDataHelperBuckets from 'src/writeData/components/WriteDataHelperBuckets'
+import {WriteDataDetailsContext} from 'src/writeData/components/WriteDataDetailsContext'
 
 // Selectors
 import {getOrg} from 'src/organizations/selectors'
 import {getMe} from 'src/me/selectors'
 import {getAllTokensResources} from 'src/resources/selectors'
-
-// Thunks
-import {getBuckets} from 'src/buckets/actions/thunks'
-
-// Helper Components
-import {WriteDataDetailsContext} from 'src/writeData/components/WriteDataDetailsContext'
 
 // Utils
 import {allAccessPermissions} from 'src/authorizations/utils/permissions'
@@ -26,9 +33,6 @@ import {keyboardCopyTriggered, userSelection} from 'src/utils/crossPlatform'
 
 // Types
 import {AppState, Authorization} from 'src/types'
-
-// Styles
-import './ArduinoSteps.scss'
 
 type OwnProps = {
   setTokenValue: (tokenValue: string) => void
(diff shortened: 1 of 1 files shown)


src/homepageExperience/components/steps/arduino/InitializeClient.tsx
```diff
@@ -47,31 +51,28 @@ export const InitializeClient: FC<OwnProps> = ({
...
   const sortedPermissionTypes = useMemo(
     () => allPermissionTypes.sort((a, b) => collator.compare(a, b)),
     [allPermissionTypes]
   )
+
   const {bucket} = useContext(WriteDataDetailsContext)
 
   useEffect(() => {
     dispatch(getBuckets())
+    dispatch(getAllResources())
```
hoorayimhelping, line 70: do we need to call `getAllResources()` here? @ChitlangeSahas do you remember why this was called in the initial wizards?

src/homepageExperience/components/steps/arduino/InitializeClient.tsx
```diff
@@ -95,22 +96,110 @@ export const InitializeClient: FC<OwnProps> = ({
...
+  
+  #define INFLUXDB_URL "${url}"
+  #define INFLUXDB_TOKEN "${token}"
+  #define INFLUXDB_ORG "${org.id}"
+  #define INFLUXDB_BUCKET "${
+    bucket.name === '<BUCKET>' ? 'YOUR_BUCKET' : bucket.name
+  }"
+  
+  // Time zone info
+  #define TZ_INFO "UTC−07:00"
```
hoorayimhelping, line 133: should this be hardcoded to -7 UTC (which is Sahas' timezone)? Or should we do something like this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset ?

Edit: seems like it, since it says accurate timezone information is necessary.

src/homepageExperience/components/steps/arduino/InitializeClient.tsx
```diff
@@ -95,22 +96,110 @@ export const InitializeClient: FC<OwnProps> = ({
...
+      <p className="small-margins">
+        Paste the following snippet into a blank Arduino sketch file.
+      </p>
+      <CodeSnippet
+        text={codeSnippet}
+        onCopy={logCopyCodeSnippet}
+        language="arduino"
+      />
+      <p style={{fontSize: '14px', marginTop: '8px', marginBottom: '48px'}}>
+        Note: you will need to set the WIFI_SSID and WIFI_PASSWORD variables to
```
hoorayimhelping, line 200: I think we'll want to wrap `WIF_SSID` and `WIFI_PASSWORD` in `<code className="homepage-wizard--code-highlight">` blocks, like [here](https://github.com/influxdata/ui/blob/master/src/homepageExperience/components/steps/python/InstallDependencies.tsx#L17)

hoorayimhelping (approved):
Some minor comments

src/homepageExperience/components/steps/arduino/InitializeClient.tsx
ChitlangeSahas: Yes, this needs to be the User's TZ info. 👍

src/homepageExperience/components/steps/arduino/InitializeClient.tsx
```diff
@@ -95,22 +96,110 @@ export const InitializeClient: FC<OwnProps> = ({
...
   const logCopyCodeSnippet = () => {
-    event(`firstMile.arduinoWizard.buckets.code.copied`)
+    event(`firstMile.arduinoWizard.initializeClient.code.copied`)
   }
 
-  return <h1>Initialize Client</h1>
+  return (
+    <>
+      <h1>Initialize Client</h1>
+      <p>Select or create a bucket to write data to.</p>
```
ChitlangeSahas, line 179: I think instead of saying "to write data to" we should say, "initialize the arduino client with"

src/homepageExperience/components/steps/arduino/InitializeClient.tsx
ChitlangeSahas: We dont have those two dispatch calls in any of the initialize wizards, but because for arduino, you need to select the bucket in initializing step, that's why david might have added it from write data page. I think.

david-rusnak: So for some reason I thought getAllResources() would get the authorization for the tokens as well as the buckets, but for some reason it only gets the authorization, so I needed to also add in getBuckets() for it to display the buckets and make a token

src/homepageExperience/components/steps/arduino/InitializeClient.tsx
david-rusnak: Fixed!

src/homepageExperience/components/steps/arduino/InitializeClient.tsx
david-rusnak: Good point, fixed!

src/homepageExperience/components/steps/arduino/InitializeClient.tsx
```diff
@@ -47,31 +51,28 @@ export const InitializeClient: FC<OwnProps> = ({
...
     onSelectBucket(bucket.name)
-  }, [bucket.name, onSelectBucket])
-
-  useEffect(() => {
-    const fetchResources = async () => {
-      await dispatch(getAllResources())
-    }
-    fetchResources()
-  }, [])
+  }, [bucket, onSelectBucket])
```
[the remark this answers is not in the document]
david-rusnak: Ahh yeah good catch!

PR #5383, opened 2022-08-12
feat: notify users with M chip to install rosetta

### LINKED ISSUE #5377: firstMile: CLI - Note what prereqs are needed for homebrew to effectively download
For some users, in order to download homebrew, they may need to upgrade their Mac. We should note what other prereqs are necessary in order for the download to be successful. Even if it's a general statement to make sure you have the latest upgrades to your homebrew and Mac OS. 

Note: This feedback came from Gary Fowler, we can contact him for more information about his experience if needed. 

Thoughts? Concerns? Feedback?

david-rusnak:
Closes #5377

Adds a note telling users that if they have an M chip, they will need Rosetta before continuing.

Screen Shot 2022-08-12 at 10 10 21 AM

### Checklist

### DIFF
--- a/src/homepageExperience/components/steps/cli/InstallDependencies.tsx
+++ b/src/homepageExperience/components/steps/cli/InstallDependencies.tsx
@@ -120,6 +120,15 @@ sudo cp influxdb2-client-latest-linux-arm64/influx /usr/local/bin/
       {currentSelection === 'Mac' && (
         <>
           <h2 style={headingWithMargin}>Use Homebrew</h2>
+          <p style={{fontSize: '14px'}}>
+            Note: if your Mac is equipped with an{' '}
+            <code className="homepage-wizard--code-highlight">M</code> chip,
+            ensure you have{' '}
+            <SafeBlankLink href="https://support.apple.com/en-us/HT211861">
+              Rosetta
+            </SafeBlankLink>{' '}
+            downloaded and installed before continuing.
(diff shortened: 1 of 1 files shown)


src/homepageExperience/components/steps/cli/InstallDependencies.tsx
```diff
@@ -120,6 +120,15 @@ sudo cp influxdb2-client-latest-linux-arm64/influx /usr/local/bin/
       {currentSelection === 'Mac' && (
         <>
           <h2 style={headingWithMargin}>Use Homebrew</h2>
+          <p style={{fontSize: '14px'}}>
```
hoorayimhelping, line 123: Not necessary to hold up this PR for this, but I'd like to see you follow up on this: we already style paragraph tags with a font size of 14px in the [this file](https://github.com/influxdata/ui/blob/master/src/homepageExperience/components/HomepageExperience.scss#L81) - I think we should make use of that scss file rather than having an inline style here.

hoorayimhelping (approved):

src/homepageExperience/components/steps/cli/InstallDependencies.tsx
david-rusnak: Hmm, tried that and for some reason it really messes with the sentence, breaks it into 3 parts that are all over the place

PR #5389, opened 2022-08-12
fix(tokens): fetch user id before creating all access token

### LINKED ISSUE #5277: [bug] sometimes api token generation fails with a length error
## About the bug

When generating an all access token, i get an error saying the length of of the id must be 16 bytes. I have to refresh the page to get past the error.

**Steps to reproduce:**

1. unknown
2. generate api toke
3. observe error

**Expected behavior:**
no errors when making an api token or at least errors should be recoverable with steps on what to do.

**Actual behavior:**
api toke json id length error - and no token!

**Visual Proof:**

Screen Shot 2022-08-02 at 9 15 50 PM

Screen Shot 2022-08-02 at 9 15 59 PM
Screen Shot 2022-08-02 at 9 16 05 PM

## About your environment

**Environment info:**
cloud2

related to #4992

blegesse-git:
Closes #5277 

According to linked ticket, sometimes error `Failed to create API token: invalid json structure: id must have 16 bytes` occurs when attempting to create an all access token. To get past this problem, user has to hard refreshes the page. 
When testing the error locally, I was able to generate that same error by passing in an empty string in place of `meID`. 

fix: 
Inside `handleSave` Before constructing the tokens object, check if the value of meID is empty and dispatch an action to fetch its value. 

### Checklist

### DIFF
--- a/src/authorizations/components/AllAccessTokenOverlay.tsx
+++ b/src/authorizations/components/AllAccessTokenOverlay.tsx
@@ -21,6 +21,7 @@ import {
 // Actions
 import {createAuthorization} from 'src/authorizations/actions/thunks'
 import {showOverlay, dismissOverlay} from 'src/overlays/actions/overlays'
+import {getIdpeMeThunk} from 'src/me/actions/thunks/index'
 
 // Contexts
 import {OverlayContext} from 'src/overlays/components/OverlayController'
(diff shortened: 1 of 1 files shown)


src/authorizations/components/AllAccessTokenOverlay.tsx
```diff
@@ -58,6 +59,10 @@ const AllAccessTokenOverlay: FC<OwnProps> = props => {
   )
 
   const handleSave = () => {
+    if (!meID) {
```
blegesse-git, line 62: meID being empty is what's causing the error mentioned in the bug ticket to occur. Checking for its value, then dispatching an action to store result in state will make sure we are including `meID` when making a post request to create a token.

PR #5391, opened 2022-08-12
fix: fixes most of the windows commands

### LINKED ISSUE #5390: firstMile: InfluxCLI Onboarding - Fix windows commands
We need to update the code snippets based on which operating system the users are using. Mac and linux should be the same but windows needs its own commands.

david-rusnak:
Closes #5390

Displays commands based on which OS the user is using. Now windows commands are correct and the directions are more clear. More work needs to be done to figure out why the queries are not identifying the buckets in windows.

### Checklist

### DIFF
--- a/src/homepageExperience/components/steps/cli/ExecuteAggregateQuery.tsx
+++ b/src/homepageExperience/components/steps/cli/ExecuteAggregateQuery.tsx
@@ -73,9 +73,9 @@ export const ExecuteAggregateQuery = (props: OwnProps) => {
       <p>In the InfluxDB CLI, run the following:</p>
       <CodeSnippet
         text={
-          window?.navigator?.userAgent.includes('Mac')
-            ? codeSnippetMac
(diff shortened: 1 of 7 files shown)


src/homepageExperience/components/steps/cli/ExecuteAggregateQuery.tsx
```diff
@@ -71,7 +72,11 @@ export const ExecuteAggregateQuery = (props: OwnProps) => {
       />
       <p>In the InfluxDB CLI, run the following:</p>
       <CodeSnippet
-        text={codeSnippet}
+        text={
+          window?.navigator?.userAgent.includes('Windows')
```
ChitlangeSahas, line 76: should we put this into a utility function? i remember you made a utility to detect ctrl-c keyboard, seems like a good candidate to put in there.

ChitlangeSahas (approved):
Rest looks great!

PR #5386, opened 2022-08-12
feat: create e2e tests for user profile

wdoconnell:
Connects #4053 

Adds e2e tests for the user profile page:

All Users
--
- displays the user's name and email

Multi-Account / Multi-Org Users
--
- allows the user to change their default account
- allows the user to change their default org
- allows the user to change both their default account and org
- notifies if the user's preferences were not saved

Single-Account and/or Single-Org Users
--
- doesn't show the 'change defaults' form if there is only one account and one org
- doesn't show the 'switch default account' fieldset if there is only one account
- doesn't show the 'switch default org' fieldset if there is only one org in the current account

https://user-images.githubusercontent.com/91283923/184644729-b48d7386-e3a6-4a2f-a394-aacbf4c1e291.mov

### Checklist

Authors and Reviewer(s), please verify the following:

- [X] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [X] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [X] Documentation updated or issue created (provide link to issue/PR)
- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [X] Feature flagged, if applicable - `multiOrg`

PR #5429, opened 2022-08-16
feat: add e2e tests for change-account and change-org dropdowns

### LINKED ISSUE #4053: Multi Org Phase 1 - End to End Testing

wdoconnell:
Closes #4053 

Adds e2e tests for the 'change account' and 'change org' dropdowns in the multi-org phase 1 global header. The remaining e2e tests for multi-org were completed in #5386 and #5381. 

1. Don't render the header if the server fails to provide the user's account/org data.

3. Account Dropdown
    - Navigate to the Settings page
    - Navigate to the Billing page
    - Change the current account
4. Organization Dropdown
    - Navigate to the Settings page
    - Navigate to the Members page
    - Navigate to the Usage page
    - Change the current organization

The hurdle in testing global header is that the URLs targeted by the dropdown include IDPE-generated org IDs. In prod, the IDPE org id and the quartz org ID are the same. In remocal and CI, they are not, because quartz-mock serves static org ids.

This PR makes those links visitable during e2e testing by making one API call to retrieve the IDPE org ID, then intercepting calls to the quartz API and replacing the org id served by quartz mock with the IDPE org id.

https://user-images.githubusercontent.com/91283923/184989072-c55a7f67-c95e-4d29-adef-05ffca2ee756.mov

The `404` at the end of the video is intended behavior. Logging out requires a quartz redirect that we can't accomplish during testing, so the test focuses on the target URL, rather than the expected content of the login page.

### Checklist

- [X] Feature flagged, if applicable - `multi-Org`

### DIFF
--- a/src/identity/components/GlobalHeader/GlobalHeader.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeader.tsx
@@ -1,13 +1,13 @@
 // Library imports
-import React, {useContext, useEffect, useState, FC} from 'react'
-import {useSelector, useDispatch} from 'react-redux'
+import React, {FC, useContext, useEffect, useState} from 'react'
+import {useDispatch, useSelector} from 'react-redux'
 import {
   ComponentSize,
   FlexBox,
-  IconFont,
   Icon,
-  JustifyContent,
+  IconFont,
   InfluxColors,
+  JustifyContent,
 } from '@influxdata/clockface'
 
 // Selectors and Context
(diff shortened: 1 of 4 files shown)


src/identity/components/GlobalHeader/GlobalHeader.tsx
```diff
@@ -77,33 +77,42 @@ export const GlobalHeader: FC = () => {
     }
   }, [orgsList])
```
wdoconnell, line 79: These conditionals ensure that we don't render a non-functional globalheader to the user in the case where quartz endpoints fail. Rather than having an unpopulated dropdown or profile avatar, the global header will simply not render if quartz doesn't load. 

The figma doesn't specify one way or another, but my preference is not to have a toast notification for the user in this case. If quartz goes down, the user won't be able to change accounts or orgs, but they may well be able to use other functionality like data explorer until quartz comes back online, and won't want to be pestered with notifications about functionality that is only used once in a while.

hoorayimhelping (approved):
very nice! good job thinking over what happens when things aren't ideal!

PR #5438, opened 2022-08-17
fix(inLineLabels): handle null prop in selector

### LINKED ISSUE #5434: Flaky Test(buckets): Element is detached from DOM
```
CypressError: Timed out retrying after 10050ms: `cy.click()` failed because this element is detached from the DOM.

`<button class="cf-button cf-button-xs cf-button-default" title="Settings" tabindex="0" type="button" data-testid="bucket-settings">...</button>`

Cypress requires elements be attached in the DOM to interact with them.
```

Buckets e2e test consistently fails on monitor ci with the error above. 
Getting this error means you've tried to interact with a "dead" DOM element - meaning it's been detached or completely removed from the DOM. So by the time, cypress is about to click the element eq() was either detached or removed from the DOM.

Possible solution: 

- Guard Cypress from running commands until a specific condition is met

### LINKED ISSUE #5439: bug(buckets): editing and saving buckets with no labels causes UI to crash
## About the bug

**Steps to reproduce:**

1. Create a bucket 
2. remove/avoid adding labels to the bucket
3. Go to its settings 
4. click `save changes`

**Expected behavior:**
Bucket should update without ui crashing 

**Actual behavior:**
ui crashes after clicking save

**Visual Proof:**

https://user-images.githubusercontent.com/66275100/185193482-8ac3be2f-9cb5-4f52-8b70-3082f88a8405.mov

## About your environment

**Environment info:**

**Config:**

**Logs:**

**Performance:**

```sh
iostat -xd 1 30 > iostat.txt
```

blegesse-git:
Closes #5439 

Pr adds a check in `mstp` in order to prevent null value `props.selectedLabelIDs` from being used in selector. 

Before fix: 

https://user-images.githubusercontent.com/66275100/185191639-e0407a4d-130b-499f-96d2-21a7d66e412d.mov

After fix: 

https://user-images.githubusercontent.com/66275100/185191722-7b2d2ece-2504-4edf-bdc1-b98ad12c8b93.mov

### Checklist

### DIFF
--- a/src/shared/components/inlineLabels/InlineLabels.tsx
+++ b/src/shared/components/inlineLabels/InlineLabels.tsx
@@ -114,7 +114,11 @@ class InlineLabels extends Component<Props> {
 
 const mstp = (state: AppState, props: OwnProps) => {
   const labels = getAll<Label>(state, ResourceType.Labels)
-  const selectedLabels = getLabels(state, props.selectedLabelIDs)
+
+  let selectedLabels = []
+  if (props.selectedLabelIDs) {
+    selectedLabels = getLabels(state, props.selectedLabelIDs)
+  }
 
   return {labels, selectedLabels}
 }


drdelambre (approved):

src/shared/components/inlineLabels/InlineLabels.tsx
```diff
@@ -114,7 +114,11 @@ class InlineLabels extends Component<Props> {
 
 const mstp = (state: AppState, props: OwnProps) => {
   const labels = getAll<Label>(state, ResourceType.Labels)
-  const selectedLabels = getLabels(state, props.selectedLabelIDs)
+
+  let selectedLabels = []
+  if (props.selectedLabelIDs) {
+    selectedLabels = getLabels(state, props.selectedLabelIDs)
+  }
```
blegesse-git, line 121: instead of this, we can also just add null check in the `getLabels` selector. If we go that route, we just have to make sure this component can handle undefined as `selectedLabels` value

drdelambre: can also run a `selectedLabels = getLabels(state, props.selectedLabelIDs??[])??[]` so it'll always have something

blegesse-git: noted!

hoorayimhelping (approved):
This will address the symptom of the problem, but will leave the problem itself intact. The underlying problem as I understand it right now is that after the bucket schema is updated, the label property associated with the bucket is `null`, which gets passed to `InlineLabels.tsx`, which causes it to crash because it's expecting an array.

This is a reasonable measure to unblock the CI pipeline, but we should continue looking at this to understand two things:

1. Why the `label` property is null for a single render cycle
2. Why the tests seemed to work until yesterday with this problem in the code, which has been there since no later than [this commit from July 18th](https://github.com/influxdata/ui/commit/a7753cb0f1b0f4bf650e2463ca674ecbbff96442), which was the commit before the most recent change to `BucketList.tsx`. I checked out that commit to see if it was a recent change that caused this (it wasn't).

PR #5447, opened 2022-08-17
fix(updateBucket): set labels to [] if reponse is null

blegesse-git:
see #5439 

Due to recent changes made to the labels API, `getBucketsLabels` returns  `null` instead of `[]` for buckets with no labels. This response is incongruent with what the UI expects and causes the crash mentioned in the ticket linked. 

Pr adds a guard forcing `labels` to be an empty array when the api responds with a `null`. 

after fix: 

https://user-images.githubusercontent.com/66275100/185214225-7ccec46f-3b59-40bc-81b0-d6ca574ccdd4.mov

### Checklist

### DIFF
--- a/src/buckets/actions/thunks.ts
+++ b/src/buckets/actions/thunks.ts
@@ -225,7 +225,7 @@ export const updateBucket = (bucket: OwnBucket) => async (
     }
 
     const newBucket = normalize<Bucket, BucketEntities, string>(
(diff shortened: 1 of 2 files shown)


hoorayimhelping (approved):

hoorayimhelping:
Thanks for the detailed PR description explaining the context of this!

PR #5461, opened 2022-08-18
chore: fixing the next button in arduino wizard

### LINKED ISSUE #5449: firstMile: Arduino - Add capability to hit next on second to last page
On the Execute aggregate page there is no capability to hit the next button.

Image

alexjw3:
Closes #5449

Allowing the next button to be used in the aggregate query page.

### Checklist

### DIFF
--- a/src/homepageExperience/containers/ArduinoWizard.tsx
+++ b/src/homepageExperience/containers/ArduinoWizard.tsx
@@ -217,7 +217,8 @@ export class ArduinoWizard extends PureComponent<{}, State> {
                   size={ComponentSize.Large}
                   color={ComponentColor.Primary}
                   status={
-                    this.state.currentStep < 8
+                    this.state.currentStep <
+                    HOMEPAGE_NAVIGATION_STEPS_ARDUINO.length
                       ? ComponentStatus.Default
                       : ComponentStatus.Disabled
                   }
--- a/src/homepageExperience/containers/ArduinoWizard.tsx
+++ b/src/homepageExperience/containers/ArduinoWizard.tsx
@@ -64,12 +64,6 @@ export class ArduinoWizard extends PureComponent<{}, State> {
   }
 
   handleNextClick = () => {
-    this.setState({
-      currentStep: Math.min(
-        this.state.currentStep + 1,
-        HOMEPAGE_NAVIGATION_STEPS_ARDUINO.length
-      ),
-    })
     event(
       'firstMile.arduinoWizard.next.clicked',
       {},
@@ -82,10 +76,15 @@ export class ArduinoWizard extends PureComponent<{}, State> {
         ),
       }
     )
+    this.setState({
+      currentStep: Math.min(
+        this.state.currentStep + 1,
+        HOMEPAGE_NAVIGATION_STEPS_ARDUINO.length
+      ),
+    })
   }
 
   handlePreviousClick = () => {
-    this.setState({currentStep: Math.max(this.state.currentStep - 1, 1)})
     event(
       'firstMile.arduinoWizard.previous.clicked',
       {},
@@ -98,6 +97,7 @@ export class ArduinoWizard extends PureComponent<{}, State> {
         ),
       }
     )
+    this.setState({currentStep: Math.max(this.state.currentStep - 1, 1)})
   }
 
   handleNavClick = (clickedStep: number) => {
--- a/src/homepageExperience/containers/ArduinoWizard.tsx
+++ b/src/homepageExperience/containers/ArduinoWizard.tsx
@@ -64,38 +64,38 @@ export class ArduinoWizard extends PureComponent<{}, State> {
   }
 
   handleNextClick = () => {
-    this.setState(
+    this.setState({
+      currentStep: Math.min(
+        this.state.currentStep + 1,
+        HOMEPAGE_NAVIGATION_STEPS_ARDUINO.length
+      ),
+    })
+    event(
+      'firstMile.arduinoWizard.next.clicked',
+      {},
       {
-        currentStep: Math.min(
-          this.state.currentStep + 1,
-          HOMEPAGE_NAVIGATION_STEPS_ARDUINO.length
+        clickedButtonAtStep: normalizeEventName(
+          HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep - 1].name
         ),
-      },
-      () => {
-        event(
-          'firstMile.arduinoWizard.next.clicked',
-          {},
-          {
-            clickedButtonAtStep: this.state.currentStep - 1,
-            currentStep: this.state.currentStep,
-          }
-        )
+        currentStep: normalizeEventName(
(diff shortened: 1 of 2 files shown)


src/homepageExperience/containers/ArduinoWizard.tsx
```diff
@@ -64,46 +64,38 @@ export class ArduinoWizard extends PureComponent<{}, State> {
...
         ),
-      },
-      () => {
-        event(
-          'firstMile.arduinoWizard.next.clicked',
-          {},
-          {
-            clickedButtonAtStep: normalizeEventName(
-              HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep - 1].name
-            ),
```
ChitlangeSahas, line 81: Why do we take it out of the setState callback?

src/homepageExperience/containers/ArduinoWizard.tsx
```diff
@@ -225,7 +217,7 @@ export class ArduinoWizard extends PureComponent<{}, State> {
                   size={ComponentSize.Large}
                   color={ComponentColor.Primary}
                   status={
-                    this.state.currentStep < 7
+                    this.state.currentStep < 8
```
ChitlangeSahas, line 220: to future proof this, we can do `HOMEPAGE_NAVIGATION_STEPS_ARDUINO.length` (?)

src/homepageExperience/containers/ArduinoWizard.tsx
alexjw3: We take it out because the event is expecting the previous state rather than the updated state.

ChitlangeSahas: Then i think we should put that before the setState, to make that explicitly clear.

hoorayimhelping: I agree, it's probably better to put it before the call to `setState` so we don't have to do wonky index math when firing the event

alexjw3: I decided instead to not use indexing at all to avoid messiness. See the new commit where we only report the current step rather than the index. This is what we used for previous wizards so could be better for consistency.

ChitlangeSahas: Hmm.. i am a little concerned about replacing the normalizeEventName function. It seems like a good addition for consistency.

Plus, we're going from step names to step numbers, which makes it more confusing to people who are going to read the events. like, 

```
clickedButtonAtStep: Finish
clickedButtonAtStep: 9
```

the first event name is better to follow.

ChitlangeSahas: Or am I misunderstanding the change?

src/homepageExperience/containers/ArduinoWizard.tsx
```diff
@@ -64,46 +64,38 @@ export class ArduinoWizard extends PureComponent<{}, State> {
...
-            clickedButtonAtStep: normalizeEventName(
-              HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep - 1].name
-            ),
-            currentStep: normalizeEventName(
-              HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep].name
-            ),
-          }
-        )
+    this.setState({currentStep: Math.max(this.state.currentStep - 1, 1)})
+    event(
```
ChitlangeSahas, line 89: Now that the event function call is out of the callback, its not guranteed what the state of `currentStep` will be. This might get called before/after the state is updated. 

we should move the event call before the setState for that reason. 

Does that make sense? 

https://stackoverflow.com/a/42038724

src/homepageExperience/containers/ArduinoWizard.tsx
```diff
@@ -64,48 +64,40 @@ export class ArduinoWizard extends PureComponent<{}, State> {
...
-            currentStep: normalizeEventName(
-              HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep].name
-            ),
-          }
-        )
+    event(
+      'firstMile.arduinoWizard.previous.clicked',
+      {},
+      {
+        clickedButtonAtStep: normalizeEventName(
```
ChitlangeSahas, line 92: I think this should be `this.state.currentStep` , `clickedButtonAtStep ===> currentStep`

src/homepageExperience/containers/ArduinoWizard.tsx
```diff
@@ -64,48 +64,40 @@ export class ArduinoWizard extends PureComponent<{}, State> {
...
+          HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep - 1].name
+        ),
+        currentStep: normalizeEventName(
+          HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep - 2].name
```
ChitlangeSahas, line 96: I think this should be `this.state.currentStep` instead of `this.state.currentStep - 2`.

src/homepageExperience/containers/ArduinoWizard.tsx
```diff
@@ -64,48 +64,40 @@ export class ArduinoWizard extends PureComponent<{}, State> {
...
-              HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep].name
-            ),
-          }
-        )
+    event(
+      'firstMile.arduinoWizard.previous.clicked',
+      {},
+      {
+        clickedButtonAtStep: normalizeEventName(
+          HOMEPAGE_NAVIGATION_STEPS_ARDUINO[this.state.currentStep - 1].name
```
ChitlangeSahas, line 93: And this one looks good.

src/homepageExperience/containers/CliWizard.tsx
```diff
@@ -63,13 +63,6 @@ export class CliWizard extends PureComponent<{}, State> {
...
-    this.setState({
-      currentStep: Math.min(
-        this.state.currentStep + 1,
-        HOMEPAGE_NAVIGATION_STEPS_SHORT.length
-      ),
-    })
-
     event(
       'firstMile.cliWizard.next.clicked',
       {},
```
ChitlangeSahas, line 68: seems like we're missing the metadata here? (the data about what step the user is on, and where was the next clicked?

ChitlangeSahas (changes requested):
Let's pair up on this tomorrow if needed. I am having some trouble understanding the intent behind changing the eventing logic. And seems like the eventing logic in CLI onboarding needs some work too. 

it might help for you to print the events to console when you click next/previous so you can understand what the values are at certain steps.

src/homepageExperience/containers/CliWizard.tsx
ChitlangeSahas: This was github minimizing the next code lines, sorry for the false alarm.

ChitlangeSahas (approved):

Issue #5473, opened 2022-08-19
Skipping labels tests temporarily.


ChitlangeSahas:
Labels tests are failing because of the labels service being down. We should unblock the UI pipeline and then re-enable the tests later.

PR #5474, opened 2022-08-19
test(labels): skip failing labels tests temporarily until labels service is back

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/5473

### Checklist

### DIFF
--- a/cypress/e2e/shared/labels.test.ts
+++ b/cypress/e2e/shared/labels.test.ts
@@ -23,7 +23,7 @@ describe('labels', () => {
     return `background-color: rgb(${red}, ${green}, ${blue});`
   }
 
-  it('can create a label', () => {
+  it.skip('can create a label', () => {
     const newLabelName = 'Substantia (サブスタンス)'
     const newLabelDescription =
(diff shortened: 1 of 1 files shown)


cypress/e2e/shared/labels.test.ts
```diff
@@ -23,7 +23,7 @@ describe('labels', () => {
     return `background-color: rgb(${red}, ${green}, ${blue});`
   }
 
-  it('can create a label', () => {
+  it.skip('can create a label', () => {
```
blegesse-git, line 26: should we skip the entire test instead of just this one? I would assume the remaining tests in this file will also fail

ChitlangeSahas: As far as I can see it was only the `can create a label` test,  let me check if others are failing too.

blegesse-git: the flaky test fails with `Expected to find element: `[data-testid="label-card"]`, but never found it.`  I don't see others failing, you might be right

blegesse-git (approved):

cypress/e2e/shared/labels.test.ts
ChitlangeSahas: We're going to let this CI run finish and see if it passes, if not I can skip the whole suite.

Issue #5472, opened 2022-08-19
Black screen on homepage


hoorayimhelping:
Two members of our support organization, Jason Lytle and Aaron Koerner reported that after logging in, they were seeing a black screen on the homepage. Bucky and Bill were able to reproduce the error and were able to get a log pointing to this line:

https://github.com/influxdata/ui/blob/master/src/me/actions/thunks/index.ts#L75

Asking Jason and Aaron to toggle the `rudderstackReporting` flag locally caused the error to go away.

![Screen Shot 2022-08-19 at 9 13 28 AM](https://user-images.githubusercontent.com/146112/185663838-1abeb39f-e9ce-4f46-9cf8-263b2a6b995d.png)

PR #5476, opened 2022-08-19
fix(rudderstack): move identify call later where org.id will exist

hoorayimhelping:
Closes https://github.com/influxdata/ui/issues/5472

Rudderstack calls `identify` with user id, user email address, and org id. The call to `identify` was using an `org` variable that didn't exist under certain circumstances, and was crashing the application. Where the call is in the flow of the application depends on `org.id` existing in local state. If it doesn't it casues the black screen shown in the referenced issue.

### Checklist

### DIFF
--- a/src/me/actions/thunks/index.ts
+++ b/src/me/actions/thunks/index.ts
@@ -1,6 +1,5 @@
 // Libraries
 import HoneyBadger from 'honeybadger-js'
-import {identify} from 'rudder-sdk-js'
 import {Dispatch} from 'react'
 
 // Functions making API calls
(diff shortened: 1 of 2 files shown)


src/me/actions/thunks/index.ts
```diff
@@ -61,20 +55,10 @@ export const getIdpeMeThunk = () => async (
       },
     })
 
-    updateReportingContext({
-      userID: user.id,
-    })
```
hoorayimhelping, line 66: this is already being called a couple of lines above this line

wdoconnell (approved):
Lgtm.

PR #5484, opened 2022-08-19
fix: centralize the getter for the query needed by adaptive zoom

### LINKED ISSUE #5319: Query Zoom (GA) - Bug: Notebooks & New Data Explorer: zooming in does not issue a query properly for more data
- [ ] In Notebooks, zooming does not actually issue a proper query, but it is supposed to. The probable cause is that the query is buried somewhere in context rather than redux state.

- [ ] In the New Data Explorer, zooming also does not issue a proper query.

TCL735:
Closes #5319  

- allows adaptive zoom to be used in Notebooks by getting the right context to access the query
- allows adaptive zoom to be used in New Data Explorer by getting the right context to access the query

Here are the 3 different primary ways to visualize a graph and the adaptive zoom working in all of them:

https://user-images.githubusercontent.com/10736577/185716360-d560f815-2ed9-4424-95a5-ed37081135f9.mp4

### DIFF
--- a/src/visualization/types/Graph/view.tsx
+++ b/src/visualization/types/Graph/view.tsx
@@ -45,6 +45,7 @@ import {useAxisTicksGenerator} from 'src/visualization/utils/useAxisTicksGenerat
 import {getFormatter} from 'src/visualization/utils/getFormatter'
 import {useLegendOpacity} from 'src/visualization/utils/useLegendOpacity'
 import {useStaticLegend} from 'src/visualization/utils/useStaticLegend'
+import {useZoomQuery} from 'src/visualization/utils/useZoomQuery'
 import {
   useVisXDomainSettings,
   useVisYDomainSettings,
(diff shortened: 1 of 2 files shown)


src/visualization/types/Graph/view.tsx
```diff
@@ -181,13 +182,14 @@ const XYPlot: FC<Props> = ({
     return useVisYDomainSettings(storedDomain, memoizedYColumnData)
   }
 
+  const zoomQuery = useZoomQuery(properties)
```
hoorayimhelping, line 185: should this call be moved down so that it's inside the 

```ts
if (isFlagEnabled('zoomRequery') {
```
check?

hoorayimhelping (approved):

src/visualization/types/Graph/view.tsx
TCL735: no, it's a hook and cannot be called conditionally.

PR #5486, opened 2022-08-22
fix: add eventing to multi-org

### LINKED ISSUE #4046: Multi Org Phase 1 - Header Global Component
**Multi Org - Header Global Component**

**User Story: Header Global Component**
As a user, I want to be able to see which account and org I'm on at all times and always be able to navigate to my user profile regardless of what screen I'm on.

  **ACs**
     - The header section (global component) displays the current account, org, and user profile icon. 
     - The header section and account/org dropdowns are global components that can be viewed on any screen
     - Display additional icons near switch org and switch account as shown in the Figma. For example, the breadcrumb icon that shows org being lower in hierarchy than account. 
     - The "Upgrade Now" Button should be to the left of the User Profile icon in the header component
     - The header component should be either black or transparent
     - If the user is in light mode - which is allowed on the dashboard page - then the icons and text should be Dark Grey (consult design - Julia James for exact color)
     - If the user is in regular standard dark mode, then the icons and text should be light grey and white respectively. (consult design - Julia James for exact color)
     - In support of mobile, ensure the header component is responsive and fits mobile screens. Refer to the Figma file. 

  Dev Notes
     - We need to figure out the implementation of how to fetch the account and organization information.
        - Fetch on demand? 
        - Fetch at load? 
        - Or some combination?

### LINKED ISSUE #4056: Multi Org Phase 1 - Measure Success - Setup Eventing
**Why: Assess how users utilize the Phase 1 multiOrg features**

- Track if a user completes switching to a new account
    (How frequently do users switch accounts?)

- Track if a user completes switching to a new org   
    (How frequently do users switch organizations?)

- Track if a user navigates to the user profile page   
    (How frequently do users utilize the user profile page?)

- Track how often users change their default account on the user profile page
    (How frequently do users change their default account?)

- Track how often users change their default org on the user profile page
    (How frequently do users change their default org?)

**ACs**

-  Add UI eventing to multiOrg Phase 1 features in alignment with our new naming conventions described in our UI Eventing Dictionary: https://influxdb.atlassian.net/wiki/spaces/IPO/pages/2530574337/UI+Eventing+Dictionary

**Dev Notes**

- Send events to Snowflake. 

- Eventually, this will end up in Domo for PM and Analytics to create visuals to assess the impact of the capabilities.

wdoconnell:
Closes #4046
Closes #4056 

This PR adds 22 events to track the use of multi-org phase 1 features. It also refactors (mostly re-alphabetizes) the components these changes impact. The event names and associated tags and fields can be found in the eventing dictionary in confluence.

The components generating these events are shared by multiple parts of the multi-org implementation (e.g., same component is used in the global header and the user profile page; same component is used for changing 'accounts' versus changing 'orgs'). As a result, and to organize the event names into a single file, the event names are handled by enums in [src/identity/events/multiOrgEvents.ts](https://github.com/influxdata/ui/blob/5deae96cb6a46c2d97811e6a332eb5f10944e7ca/src/identity/events/multiOrgEvents.ts) that, in many cases, are passed as props.

Per discussion with @Luckette:
1. We will pass account name and account ID through the rudderstack `identify` call. That information is not currently passed to rudderstack, but is likely necessary to analyze the eventing data.
2. We are not adding the event mentioned in the ACs for 'amount of time spent on the profile page.'

Testing with `streamEvents` flag enabled (`console.log`s events):

User Profile
--
https://user-images.githubusercontent.com/91283923/185948749-9f9773f7-13f5-4abd-b166-0c40e967590c.mov

Global Header
--
https://user-images.githubusercontent.com/91283923/185948541-7d9fb121-da82-487d-9fe7-9d0af21e797b.mov

### Checklist


hoorayimhelping:
you can test these locally by deploying your remocal instance with the following command:

```sh
make deploy-remocal FRONTEND_SERVICES_TO_INFLUXDB=true
```
then creating a bucket called `frontendservices` locally. This will write the metrics your app generates to your local `frontendservices` bucket.

You can also check the network requests sent by the `app-metrics` service, which is the name of the ui service running frontend services by filtering network requests on `app-metrics` and checking the request body.

hoorayimhelping (commented):
Suggestion on just using `event` and a question if we need more defensiveness when ussing `account` properties

src/identity/events/multiOrgEvents.ts
```diff
@@ -0,0 +1,48 @@
+import {event} from 'src/cloud/utils/reporting'
+import {PointFields, PointTags} from 'src/cloud/apis/reporting'
+
+export const multiOrgEvent = (eventName: string, fields?: PointFields) => {
+  const tags: PointTags = {
+    initiative: 'multiOrg',
+  }
+  event(eventName, tags, fields)
+}
```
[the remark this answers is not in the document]
wdoconnell: Thanks! This will be complete in newest commit that is pushed up.

src/shared/containers/GetOrganizations.tsx
```diff
@@ -80,9 +81,14 @@ const GetOrganizations: FunctionComponent = () => {
 
   useEffect(() => {
     if (CLOUD && org?.id && isFlagEnabled('rudderstackReporting')) {
-      identify(meId, {email, orgID: org.id})
+      identify(meId, {
+        email,
+        orgID: org.id,
+        accountID: account.id,
+        accountName: account.name,
```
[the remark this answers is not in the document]
wdoconnell: Agreed, good call, and added in next commit. 

On the one hand, I do think we can be assured these properties will at least exist, since they are part of the default redux state. On the other hand, they're set to empty strings - and there's no point in calling `identify` until this data is retrieved from quartz. The useEffect hook will run again, and then make a single `identify` call once all properties are retrieved, which is better.

wdoconnell:
FYI: force-push is a rebase to current master.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
```diff
@@ -25,15 +34,15 @@ export class GlobalHeaderTypeAheadMenu extends React.Component<Props, State> {
...
       selectedItem: this.props.defaultSelectedItem,
     }
   }
 
-  private selectAllTextInInput = (event?: ChangeEvent<HTMLInputElement>) => {
-    if (event) {
-      event.target.select()
+  private selectAllTextInInput = (e?: ChangeEvent<HTMLInputElement>) => {
+    if (e) {
+      e.target.select()
```
hoorayimhelping, line 45: why the change from `event` to `e`?

I'm not a fan of using `e` for events, since in JavaScript, `e` can be `event` or `error` and is often used interchangeably. Try searching the codebase for just `e` and finding an event handler or an error handler. (Regex searching for `\(e` yielded 1896 matches across 585 files for me). There's almost no benefit to using `e` over `event`, other than it's slightly faster to type, and it's faster to read, at the expense of taking more work to comprehend and map to a value when holding this block of code in your head.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
```diff
@@ -85,6 +94,15 @@ export class GlobalHeaderTypeAheadMenu extends React.Component<Props, State> {
     }
   }
 
+  private sendTypeAheadSearchEvent = (e: ChangeEvent<HTMLInputElement>) => {
+    const {typeAheadEventPrefix} = this.props
+
+    // No event should be sent if the input field is empty, or just contains whitespace.
+    if (e.target.value.trim().length) {
```
hoorayimhelping, line 101: apart from not using `e` to represent events mentioned previously, does this need to use optional chaining? If no event is sent if the input field is empty, will trying to `trim` its value end up casing a null pointer exception?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
```diff
@@ -57,8 +66,8 @@ export class GlobalHeaderTypeAheadMenu extends React.Component<Props, State> {
     }
   }
 
-  private handleInputChange = (event: ChangeEvent<HTMLInputElement>) => {
-    const filterString = event?.target?.value
+  private handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
```
hoorayimhelping, line 69: same thing here for changing `event` to `e`

hoorayimhelping (approved):
LGTM, mostly

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
wdoconnell: Agreed - but in this file, we're using `event` as a variable, so my thinking was that it's confusing to have a variable (even one in another scope) that is called `event`.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
wdoconnell: per discussion - will change these to `evt`

src/identity/components/GlobalHeader/GlobalHeaderDropdown/GlobalHeaderTypeAheadMenu.tsx
wdoconnell: Tested this with empty fields (which seem to send empty string), but no harm in adding optional chaining here, in case a null could be sent.

PR #5489, opened 2022-08-22
chore: fixing linter warnings in arduino

### LINKED ISSUE #5488: Tech Debt: redoing eslint warnings for homepageExperience
Its also important to note with this issue, especially having it have been done previously, that we have to be careful while reusing/copy and pasting code from other components that they are completely bug free. This is an extremely easy way to multiply small bugs into bigger problems.

alexjw3:
Closes #5488 

Fixes eslint warnings on Arduino page. Its also important to note with this PR, especially having it have been done previously, that we have to be careful while reusing/copy and pasting code from other components that they are completely bug free. This is an extremely easy way to multiply small bugs into bigger problems. All homepage lint warnings had previously been fixed but now came back up because of code reuse.

Showing that the behaviour is still as expected:

https://user-images.githubusercontent.com/13280708/186968526-c5dcb386-604d-4f2a-aa4a-5fcf3c67080c.mov

### Checklist

### DIFF
--- a/src/homepageExperience/components/steps/arduino/InitializeClient.tsx
+++ b/src/homepageExperience/components/steps/arduino/InitializeClient.tsx
@@ -86,7 +86,7 @@ export const InitializeClient: FC<OwnProps> = ({
       dispatch(createAuthorization(authorization))
       event(`firstMile.arduinoWizard.tokens.tokenCreated`)
     }
-  }, [dispatch, me.id, org.id, sortedPermissionTypes, tokenValue])
+  }, [dispatch, me?.id, org?.id, sortedPermissionTypes, tokenValue])
 
   // when token generated, save it to the parent component
   useEffect(() => {
--- a/src/homepageExperience/components/steps/arduino/InitializeClient.tsx
+++ b/src/homepageExperience/components/steps/arduino/InitializeClient.tsx
@@ -69,7 +69,7 @@ export const InitializeClient: FC<OwnProps> = ({
   useEffect(() => {
     dispatch(getBuckets())
     dispatch(getAllResources())
-  }, []) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [dispatch])
 
   useEffect(() => {
     onSelectBucket(bucket.name)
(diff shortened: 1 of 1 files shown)


src/homepageExperience/components/steps/arduino/InitializeClient.tsx
```diff
@@ -86,14 +86,14 @@ export const InitializeClient: FC<OwnProps> = ({
       dispatch(createAuthorization(authorization))
       event(`firstMile.arduinoWizard.tokens.tokenCreated`)
     }
-  }, [sortedPermissionTypes.length])
+  }, [dispatch, me.id, org.id, sortedPermissionTypes, tokenValue])
```
hoorayimhelping, line 89: do we need to use optional chaining (`?.`) for `me.id` and `org.id`?

alexjw3: I don't believe so, I think we confirm these are non-null earlier on.

wdoconnell: I don't see confirmation these aren't `null` or `undefined` earlier, so I'd add the optional chaining. There's two things to think about here. 

1. If this is coming from state (e.g., Redux, Context), what's the default value, and is it the type we expect?

For example, in `/me/reducers`, we can see `me.id` should be a `string` and exist.  Great.

2. However, even if it's coming from state, and there's a default state, is it going to get reassigned after the user data is retrieved to something we don't expect? 

In `/me/reducers`, `me.id` gets set (action name is `SET_ME`) to whatever the backend sends us for that property - there isn't any type validation logic in the reducer. So while IDPE _shouldn't_ be sending us a null or undefined id, there's nothing on our side of the house to prevent that from happening.

wdoconnell (changes requested):

wdoconnell (commented):
Same here - edits look good to me, but a picture is a thousand words, so a short video showing that adding these to the dependency array don't cause unintended behavior would be awesome.

wdoconnell:
I guess that means a video is . . . 2000 words? 😺

wdoconnell (approved):
Looks good -- thanks for adding the video, and the optional chaining.

PR #5493, opened 2022-08-22
feat: save a query as a notebook

asalem1:
Closes #

This work is the initial work being done to integrate "save / load" functionality into the new query builder, where a user should be able to save their editor text as an external resource. More specifically, this PR integrates the existing infrastructure and iterates upon that existing design until we can get new designs for the `Save As` functionality. This PR will be one in a long-line of PRs to introduce these changes incrementally so we can test them out as and when without needing to worry about introducing too large a body of work and making sure that the scope is well-defined for each leg of the work

### Checklist

### DIFF
--- a/src/dataExplorer/components/FluxQueryBuilder.tsx
+++ b/src/dataExplorer/components/FluxQueryBuilder.tsx
@@ -1,5 +1,6 @@
 import React, {FC, useContext, useCallback} from 'react'
 import {RemoteDataState} from 'src/types'
+import {useHistory, useLocation} from 'react-router-dom'
 
 // Components
 import {
@@ -11,6 +12,7 @@ import {
   IconFont,
   AlignItems,
   JustifyContent,
+  ComponentColor,
 } from '@influxdata/clockface'
 import {QueryProvider, QueryContext} from 'src/shared/contexts/query'
 import {EditorProvider} from 'src/shared/contexts/editor'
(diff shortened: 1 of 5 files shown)


src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -45,6 +50,10 @@ const FluxQueryBuilder: FC = () => {
     setSelection({bucket: null, measurement: ''})
   }, [setQuery, setStatus, setResult, setSelection, cancel])
 
+  const handleShowOverlay = () => {
+    history.push(`${pathname}/save`)
```
drdelambre, line 54: using pathname instead of declaring it from selector state is gross and makes a weird coupling between router state and internal state. but its a pattern used elsewhere and i think this is temporary, so whateves.

drdelambre (approved):
conceptually, the notebook save/load cycle is confusing to me. i know we can export just fine, as 
 you illustrated here, but what does loading look like? does it load the last generated query?

src/dataExplorer/components/FluxQueryBuilder.tsx
drdelambre: Again, not too invested in anything changing here, but it's that org I'd part that should look to the "currently selected org" single source of truth in the reducer for that info, not the route

rockstar (commented):
Hm, this seems confusing to me, based on my understanding of what our goals for Q3 and Q4 are. I didn't think we wanted to continue to invest deeply in notebooksd. I'd like to get some clarity before we build the foundation of the next few quarters on it.

rockstar:
> it is by no means out of scope from my understanding of the eventual outcome of our goal to save the editor text as a resource

I'm not sure I agree. I don't think we should prioritize any resource but Invocable Scripts, as that's the one we're most interested about.

> since we didn't currently have openAPI integrated with the UI

I feel like this is the first step we should take. That we have to do this work manually is a sign of some tech debt, sure, but I think we start with Invocable Scripts.

Issue #5444, opened 2022-08-17
Make user profile avatar support users without firstname, lastname


wdoconnell:
User `email` is always populated, but historical users may not have the `firstName` and `lastName` properties populated. Prospectively, certain users that use social sign-on may not have `firstName` and `lastName` populated. 

The multi-org user profile avatar should be enabled and render a default avatar image (rather than the dynamically generated image with first initial, last initial) when `firstName` and `lastName` are not set.

![Screen Shot 2022-08-17 at 1 05 46 PM](https://user-images.githubusercontent.com/91283923/185200496-06019513-cdb5-47ff-bce1-631e77561baa.png)

PR #5494, opened 2022-08-22
feat(multi-org): Show user icon when FirstName and LastName are absent 

ChitlangeSahas:
Closes [#5444](https://github.com/influxdata/ui/issues/5444)

https://user-images.githubusercontent.com/18511823/186028648-baae61df-a100-44d8-925b-00faf47f4cc1.mov

Still works normally: 

Screen Shot 2022-08-22 at 3 23 43 PM

### Checklist

### DIFF
--- a/src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
+++ b/src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
@@ -5,7 +5,6 @@ import {
   ButtonShape,
   ClickOutside,
   ComponentColor,
-  ComponentSize,
   FlexBox,
   FlexDirection,
   Icon,
--- a/src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
+++ b/src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
@@ -5,6 +5,7 @@ import {
   ButtonShape,
   ClickOutside,
   ComponentColor,
+  ComponentSize,
   FlexBox,
   FlexDirection,
   Icon,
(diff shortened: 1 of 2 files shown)


src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
```diff
@@ -100,18 +100,17 @@ class IdentityUserAvatar extends React.Component<Props, State> {
...
+          <Button
+            text={this.getInitials()}
+            shape={ButtonShape.Square}
+            color={
+              isPopoverOpen ? ComponentColor.Default : ComponentColor.Tertiary
+            }
+            onClick={this.togglePopoverState}
+            className={userAvatarButtonClassName}
+            testID="global-header--user-avatar"
+            icon={!this.getInitials() ? IconFont.User : null}
```
wdoconnell, line 112: Looks good to me.

For context for others here: the issue here is that when a user doesn't have a first name or last name, we don't want the user avatar to blow up. In Clockface, `<Button />` throws an error unless at least one of the `text` and `icon` properties is truthy.

```
if (!icon && !text) {
      throw new Error('Button requires either "text" or "icon" props')
}
```
No such error is possible here. In the usual case where a user has a first name, `text` will be truthy because `getInitials()` will be truthy. 

In the edge case where the user doesn't have a name - i.e., if `getInitials()` evaluates to empty string (the default result from redux state) or another falsy value, `text` will be `falsy`, making `icon` truthy and setting `icon` to the default profile picture.

wdoconnell (approved):

Issue #5506, opened 2022-08-23
Hide horizontal divider when switch account or org is hidden


ChitlangeSahas:
When one account or org is present, we hide the switch account/org button. with that we should also hide the horizontal divider. 

![image](https://user-images.githubusercontent.com/18511823/186203554-944eecd1-0e95-46c3-91b4-c081426d5a3d.png)

PR #5509, opened 2022-08-23
fix(multi-org): no HR line when switch button hidden

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/5506

When the org count is one, we hide the `hr` line

Screen Shot 2022-08-23 at 11 46 28 AM

### Checklist

### DIFF
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
@@ -107,7 +107,6 @@ export class GlobalHeaderDropdown extends React.Component<Props, State> {
     const {mainMenuOptions} = this.props
     return (
       <div>
-        <hr className="line-break" />
         {mainMenuOptions.map(value => {
           const iconEl = <Icon glyph={value.iconFont} className="button-icon" />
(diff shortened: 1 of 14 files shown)


src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
```diff
@@ -172,27 +171,30 @@ export class GlobalHeaderDropdown extends React.Component<Props, State> {
...
+                {showTypeAheadMenu === false && (
+                  <Icon
+                    glyph={IconFont.CaretRight_New}
+                    className="button-icon"
+                    style={{marginRight: 0}}
+                  />
+                )}
+              </FlexBox>
+            </Dropdown.Item>
+            <hr className="line-break" />
```
ChitlangeSahas, line 196: make it part of the switch button, so it only renders when switch button is supposed to.

wdoconnell: Looks good - nice fix! This is def cleaner than what I was proposing.

wdoconnell (approved):

PR #5521, opened 2022-08-24
chore: add guardrails to clearing out a script

asalem1:
Closes https://github.com/influxdata/ui/issues/5527

This PR adds the basic infrastructure around clearing out previously written queries and lays the foundation for saving editor text as a Script. The next step will be to connect the confirmation and connect the Save Script button

guardrails

### DIFF
--- a/src/dataExplorer/components/FluxQueryBuilder.tsx
+++ b/src/dataExplorer/components/FluxQueryBuilder.tsx
@@ -1,5 +1,4 @@
-import React, {FC, useContext, useCallback} from 'react'
-import {RemoteDataState} from 'src/types'
+import React, {FC, useState, useContext} from 'react'
 
 // Components
 import {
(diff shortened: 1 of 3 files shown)


appletreeisyellow (approved):
LGTM! I like the way you are doing this feature incrementally!

src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -60,10 +51,24 @@ const FluxQueryBuilder: FC = () => {
             data-testid="flux-query-builder--menu"
           >
             <Button
-              onClick={clear}
+              onClick={() => setIsPromptVisible(true)}
```
wiedld, line 54: Ah...I believe we need to still run `clear()`, because that also resets the state used for schema composition.

PR #5522, opened 2022-08-24
fix: user avatar renders a default avatar for users with no name

### LINKED ISSUE #5520: user without firstName, lastName should have user avatar visible
In [UI 5494](https://github.com/influxdata/ui/pull/5494), we allowed for a 'default' user profile avatar to be visible, even if the user does not have a `user.firstName` and `user.lastName` property set in quartz. In that case, they will see an avatar, but it will have no initials.

The user avatar isn't showing up for users without names, as there is still a guard in Global Header that won't load the avatar if the user has no first and last name.  It is now safe for us to remove that check, as it's intended behavior that the avatar should load for users without names.

Example from c2usage:
Screen Shot 2022-08-24 at 9 44 52 AM

wdoconnell:
Closes #5520 
Connects #5444, #5494 

If a user doesn't have a `firstName` or `lastName` (`null` or `undefined`), guards around the profile avatar prevent it from loading. We want the avatar to load and render a default image if the user doesn't have a name.

This PR removes the requirement that `firstName` and `lastName` be set before the user avatar renders. If `firstName` or `lastName` are `null` or `undefined`, the values will be set to empty `string`. Here is the new behavior:

- Both names: image using the user's initials is rendered
- Only one name: image using the one initial is rendered
- No names: default image renders.

User Has First and Last Name
--
Screen Shot 2022-08-25 at 9 22 15 AM

One Name Null
--
Screen Shot 2022-08-25 at 9 23 27 AM

One Name Undefined
--
Screen Shot 2022-08-25 at 9 24 19 AM

Both Names Null
--
Screen Shot 2022-08-25 at 9 25 36 AM

Both Names Undefined
--

Screen Shot 2022-08-25 at 9 25 07 AM

### Checklist

### DIFF
--- a/src/identity/components/GlobalHeader/GlobalHeader.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeader.tsx
@@ -80,8 +80,10 @@ export const GlobalHeader: FC = () => {
   const shouldLoadDropdowns = Boolean(activeOrg?.id && activeAccount?.id)
 
   const shouldLoadAvatar = Boolean(user?.email && org?.id)
-  const avatarFirstName = user?.firstName ?? ''
-  const avatarLastName = user?.lastName ?? ''
+  const avatarFirstName =
+    typeof user?.firstName === 'string' ? user.firstName.trim() : ''
+  const avatarLastName =
+    typeof user?.lastName === 'string' ? user.lastName.trim() : ''
 
   const shouldLoadGlobalHeader = Boolean(
     shouldLoadDropdowns || shouldLoadAvatar
--- a/src/identity/components/GlobalHeader/GlobalHeader.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeader.tsx
@@ -78,9 +78,11 @@ export const GlobalHeader: FC = () => {
   }, [orgsList])
 
   const shouldLoadDropdowns = Boolean(activeOrg?.id && activeAccount?.id)
-  const shouldLoadAvatar = Boolean(
-    user?.firstName && user?.lastName && user?.email && org?.id
-  )
+
+  const shouldLoadAvatar = Boolean(user?.email && org?.id)
+  const avatarFirstName = user?.firstName ?? ''
+  const avatarLastName = user?.lastName ?? ''
+
   const shouldLoadGlobalHeader = Boolean(
     shouldLoadDropdowns || shouldLoadAvatar
   )
@@ -109,8 +111,8 @@ export const GlobalHeader: FC = () => {
       {shouldLoadAvatar && (
         <IdentityUserAvatar
           email={user.email}
-          firstName={user.firstName}
-          lastName={user.lastName}
+          firstName={avatarFirstName}
+          lastName={avatarLastName}
           orgId={org.id}
         />
       )}
--- a/src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
+++ b/src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
@@ -55,7 +55,7 @@ class IdentityUserAvatar extends React.Component<Props, State> {
       <>
         <div className="user-popover-header">
           <div className="user-popover-header-name">
-            {firstName} {lastName}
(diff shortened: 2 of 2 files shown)


src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
```diff
@@ -55,7 +55,7 @@ class IdentityUserAvatar extends React.Component<Props, State> {
       <>
         <div className="user-popover-header">
           <div className="user-popover-header-name">
-            {firstName} {lastName}
+            {firstName && firstName} {lastName && lastName}
```
wdoconnell, line 58: Probably better not to render `firstName` unless `firstName` is truthy, as the empty string will still end up as a DOM node that we don't need.

https://stackoverflow.com/questions/60206732/reactjs-return-null-vs-empty-string-from-component

src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
```diff
@@ -55,7 +55,7 @@ class IdentityUserAvatar extends React.Component<Props, State> {
       <>
         <div className="user-popover-header">
           <div className="user-popover-header-name">
-            {firstName} {lastName}
+            {firstName ?? null} {lastName ?? null}
```
wdoconnell, line 58: Edit is to avoid [rendering unnecessary DOM nodes](https://stackoverflow.com/questions/30097091/correct-way-to-define-an-empty-dom-element-in-react#:~:text=If%20you%20use%20an%20empty,so%20it's%20not%20the%20same.&text=Bump%3A%20per%20react%20v.) when `firstName` and `lastName` are empty strings.

src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
```diff
@@ -94,22 +94,24 @@ class IdentityUserAvatar extends React.Component<Props, State> {
       'user-popover--open': this.state.isPopoverOpen,
     })
 
+    const avatarInitials = this.getInitials()
```
wdoconnell, line 97: We only need this function invoked once, so moving it outside the return.

src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
```diff
@@ -94,22 +94,24 @@ class IdentityUserAvatar extends React.Component<Props, State> {
...
+            text={avatarInitials}
             shape={ButtonShape.Square}
             color={
               isPopoverOpen ? ComponentColor.Default : ComponentColor.Tertiary
             }
             onClick={this.togglePopoverState}
             className={userAvatarButtonClassName}
             testID="global-header--user-avatar"
-            icon={!this.getInitials() ? IconFont.User : null}
+            icon={avatarInitials === '' ? IconFont.User : null}
```
wdoconnell, line 114: This is to be a little more explicit about what we expect here. Since `firstName` and `lastName` are forced to be strings, the only case where we render `IconFont.User` - the default profile image that doesn't have user initials - is when both names are empty strings.

src/identity/components/GlobalHeader/GlobalHeader.tsx
```diff
@@ -78,9 +78,13 @@ export const GlobalHeader: FC = () => {
   }, [orgsList])
 
   const shouldLoadDropdowns = Boolean(activeOrg?.id && activeAccount?.id)
-  const shouldLoadAvatar = Boolean(
-    user?.firstName && user?.lastName && user?.email && org?.id
-  )
+
+  const shouldLoadAvatar = Boolean(user?.email && org?.id)
+  const avatarFirstName =
+    typeof user?.firstName === 'string' ? user.firstName.trim() : ''
```
wdoconnell, line 84: This is to be really explicit that the user's names will be strings. If they're not, coerce an empty string. Added `trim()` because, otherwise, the user avatar won't render both initials if a user added whitespace at the beginning or end of their name fields.

hoorayimhelping (approved):

PR #5528, opened 2022-08-24
chore: add DEFAULT_PORT to test and port placeholder

subirjolly:
Closes # https://github.com/influxdata/data-acquisition/issues/489

## BEFORE
Kapture 2022-08-24 at 14 15 53

## AFTER
Kapture 2022-08-24 at 14 12 47

## UPDATE

Update per team experience session.
Kapture 2022-08-26 at 15 35 29

### Checklist

### DIFF
--- a/src/writeData/subscriptions/components/BrokerFormContent.tsx
+++ b/src/writeData/subscriptions/components/BrokerFormContent.tsx
@@ -55,9 +55,6 @@ const BrokerFormContent: FC<Props> = ({
   const mqttProtocol = 'MQTT'
   const protocolList = [mqttProtocol]
   const [protocol, setProtocol] = useState(mqttProtocol)
-  const DEFAULT_PORT = isFlagEnabled('subscriptionsCertificateSupport')
-    ? 8883
-    : 1883
 
   useEffect(() => {
     updateForm({...formContent, protocol: protocol.toLowerCase()})
@@ -72,6 +69,9 @@ const BrokerFormContent: FC<Props> = ({
     })
   }, [])
 
+  const showHostPortExampleText =
+    !!formContent.brokerHost ||
+    (!!formContent.brokerPort && !isNaN(formContent.brokerPort))
   return (
     <Grid>
       <Grid.Row>
@@ -189,6 +189,7 @@ const BrokerFormContent: FC<Props> = ({
                 )}
               />
             </div>
+            {/* show this iff brokerHost is populated */}
             <Form.ValidationElement
               label="Hostname or IP Address"
               value={formContent.brokerHost}
@@ -205,7 +206,7 @@ const BrokerFormContent: FC<Props> = ({
               {status => (
                 <Input
                   type={InputType.Text}
-                  placeholder="0.0.0.0"
+                  placeholder="e.g: broker.example.com"
                   name="host"
                   autoFocus={false}
                   value={formContent.brokerHost}
@@ -248,11 +249,7 @@ const BrokerFormContent: FC<Props> = ({
               {status => (
                 <Input
                   type={InputType.Number}
-                  placeholder={`${
-                    !formContent.brokerPort || isNaN(formContent.brokerPort)
-                      ? DEFAULT_PORT
-                      : formContent.brokerPort
-                  }`}
+                  placeholder="Between 1025 - 65535"
                   name="port"
                   autoFocus={false}
                   value={formContent.brokerPort}
@@ -277,21 +274,24 @@ const BrokerFormContent: FC<Props> = ({
               )}
             </Form.ValidationElement>
           </FlexBox>
-          <Heading
-            element={HeadingElement.H5}
-            weight={FontWeight.Regular}
-            className={`${className}-broker-form__example-text`}
-          >
-            {getSchemaFromProtocol(
-              formContent.protocol,
-              isFlagEnabled('subscriptionsCertificateSupport') &&
-                formContent.authType === BrokerAuthTypes.Certificate
-            )}
-            {formContent.brokerHost ? formContent.brokerHost : '0.0.0.0'}:
-            {formContent.brokerPort
-              ? formContent.brokerPort
-              : `${DEFAULT_PORT}`}
-          </Heading>
+          {showHostPortExampleText && (
+            <Heading
+              element={HeadingElement.H5}
+              weight={FontWeight.Regular}
+              className={`${className}-broker-form__example-text`}
+            >
+              {getSchemaFromProtocol(
+                formContent.protocol,
+                isFlagEnabled('subscriptionsCertificateSupport') &&
+                  formContent.authType === BrokerAuthTypes.Certificate
+              )}
+              {`${formContent.brokerHost}:${
+                !!formContent.brokerPort && !isNaN(formContent.brokerPort)
+                  ? formContent.brokerPort
+                  : ''
+              }`}
+            </Heading>
+          )}
         </Grid.Column>
         <Grid.Column widthXS={Columns.Twelve}>
           <Heading
--- a/src/writeData/subscriptions/components/BrokerFormContent.tsx
+++ b/src/writeData/subscriptions/components/BrokerFormContent.tsx
@@ -236,7 +236,7 @@ const BrokerFormContent: FC<Props> = ({
             </Form.ValidationElement>
             <Form.ValidationElement
               label="Port"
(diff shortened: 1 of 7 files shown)


src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -54,6 +54,7 @@ const BrokerFormContent: FC<Props> = ({
   const mqttProtocol = 'MQTT'
   const protocolList = [mqttProtocol]
   const [protocol, setProtocol] = useState(mqttProtocol)
+  const DEFAULT_PORT = formContent.authType === 'certificate' ? 8883 : 1883
```
genehynson, line 57: I think having the value change depending on the authentication selected might be confusing. What I proposed in the issue was actually having the value change depending on if the certificate feature flag is enabled. If certificate auth's feature flag is enabled, use 8883 as the default always. Otherwise use 1883 as the default always.

src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -244,10 +245,14 @@ const BrokerFormContent: FC<Props> = ({
               {status => (
                 <Input
                   type={InputType.Number}
-                  placeholder="1883"
+                  placeholder={`${
+                    !formContent || isNaN(formContent.brokerPort)
+                      ? DEFAULT_PORT
+                      : formContent.brokerPort
+                  }`}
```
genehynson, line 252: I think we can probably just remove the placeholder altogether now that the value is defaulted. I think it can be confusing for the user to tell if the placeholder is the default value or not. So it's probably better to just default to the default port value and if the user erases it, let the box be empty.

src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -244,10 +245,14 @@ const BrokerFormContent: FC<Props> = ({
...
                   name="port"
                   autoFocus={false}
-                  value={formContent.brokerPort}
+                  value={formContent.brokerPort ?? DEFAULT_PORT}
```
genehynson, line 255: this only works because `brokerPort` defaults to `undefined` instead of empty string. Can we change it to the following in case we ever change the default brokerPort value?

```suggestion
                  value={formContent.brokerPort ? formContent.brokerPort : DEFAULT_PORT}
```

src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -277,7 +282,9 @@ const BrokerFormContent: FC<Props> = ({
             {/* TODO: update `false` after cert support */}
             {getSchemaFromProtocol(formContent.protocol, false)}
```
genehynson, line 283: not necessarily part of the scope of this PR - but I saw this comment and remembered that its now applicable for us to address it. Feel free to move this to a new PR if you'd like, or include it here. 

```suggestion
            {getSchemaFromProtocol(formContent.protocol, isFlagEnabled('subscriptionsCertificateSupport') && formContent.authType === BrokerAuthTypes.Certificate)}
```

genehynson (commented):
could/should we just set the default port [here](https://github.com/influxdata/ui/blob/master/src/writeData/subscriptions/context/subscription.create.tsx#L35)?

src/writeData/subscriptions/components/BrokerFormContent.tsx
subirjolly: ah gotcha. i thought as people use 1883 for non-secure connections and 8883 for secure ones, we should do this. just like 80 vs 443. But let me change this.

genehynson: I hear ya. I'm just worried since the people fill out the form from top to bottom they are going to make sure the port is correct and then move on to select their authentication. If toggling the authentication method changes the port I can see people not noticing the change and getting frustrated why the connection isn't working.

src/writeData/subscriptions/components/BrokerFormContent.tsx
cannonpalms: Same as 
```ts
formContent.brokerPort || DEFAULT_PORT
```

src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -274,10 +282,15 @@ const BrokerFormContent: FC<Props> = ({
...
+            {getSchemaFromProtocol(
+              formContent.protocol,
+              isFlagEnabled('subscriptionsCertificateSupport') &&
+                formContent.authType === BrokerAuthTypes.Certificate
+            )}
             {formContent.brokerHost ? formContent.brokerHost : '0.0.0.0'}:
-            {formContent.brokerPort ? formContent.brokerPort : '1883'}
+            {formContent.brokerPort
+              ? formContent.brokerPort
+              : `${DEFAULT_PORT}`}
```
genehynson, line 293: Do we need this ternary now that brokerPort has a default value in the context?

src/writeData/subscriptions/components/BrokerFormContent.tsx
genehynson: Hmm I'm torn on this one. I feel like this box should become inconsistent from the other ones because it's the only one that has a default value. But I'm fine with leaving the placeholder for now and getting feedback from others before changing it

src/writeData/subscriptions/components/BrokerFormContent.tsx
genehynson: That's fair @subirjolly I think we should just remove this ternary altogether

src/writeData/subscriptions/components/BrokerFormContent.tsx
subirjolly: yes, because if you delete the port then you will see something like `0.0.0.0:` which is supposed to show what the url should look like.

src/writeData/subscriptions/components/BrokerFormContent.tsx
subirjolly: @cannonpalms Tie-Breaker please 😆

src/writeData/subscriptions/components/BrokerFormContent.tsx
cannonpalms: Ship it

src/writeData/subscriptions/components/BrokerFormContent.tsx
cannonpalms: Not a fan of the placeholder, personally, as it leads the user to believe the form has already been filled in. My vote is to nuke it.

src/writeData/subscriptions/components/BrokerFormContent.tsx
genehynson: Right but that placeholder would be accurate, right? Because if you delete the port, the address is invalid. So we should show the user the invalid address so they understand what needs to be fixed. Or am I missing something?

helenosheaa: i agree with gene here if we have a default port we no longer need a ternary

src/writeData/subscriptions/components/BrokerFormContent.tsx
subirjolly: ## BEFORE
BEFORE

## AFTER
AFTER

src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -185,6 +189,7 @@ const BrokerFormContent: FC<Props> = ({
                 )}
               />
             </div>
+            {/* show this iff brokerHost is populated */}
```
helenosheaa, line 192: do we need this comment?

subirjolly: nope.. were just a note. thanks for catching it

helenosheaa (approved):
lgtm

src/writeData/subscriptions/components/BrokerFormContent.tsx
```diff
@@ -244,7 +248,7 @@ const BrokerFormContent: FC<Props> = ({
               {status => (
                 <Input
                   type={InputType.Number}
-                  placeholder="1883"
+                  placeholder="Between 1025 - 65535"
```
genehynson, line 251: can we use the variables in `utils/form.ts` `MAX_PORT` & `MIN_PORT` here

genehynson (approved):

Issue #152, opened 2020-10-07
Quartz local development


drdelambre:
The ecommerce team already has a basic provisioned elixir machine that they're using for automation, but it's not so good for local development of quartz itself. They're using a framework called phoenix, which has the capability to do hot reloading whenever a source file gets changed. This seems to be a pretty similar need to what the monitoring team is using for their local developement.

We need to:
 1. map out what folders they need to watch
 1. figure out how to launch a phoenix server in dev_mode
 1. give them their own set of yml files and `.env` entries in monitor-ci

What this ticket isn't asking for:
 1. integration with idpe
 1. any vault work
 1. any work around provisioning state

This is a requirement for:
influxdata/idpe#8263
influxdata/quartz#3405

hercules-influx:
Metric | Score | Level
------------ | ------------- | ------------- 
performance | 21 | :red_circle:
accessibility | 65 | :yellow_circle:
best-practices | 80 | :yellow_circle:
seo | 89 | :green_circle:
pwa | null | :question:

subirjolly:
Seems like this is what we are asking for:
https://github.com/influxdata/quartz/blob/master/README.md

This setup is pretty easy and takes a few hours but it does work.

Asalem1:
Gonna close this since it was resolved by the e-commerce team

PR #5534, opened 2022-08-24
fix: force null return from global header until all values loaded

### LINKED ISSUE #5526: minified react error #152
Staging environments using Electron (likely cypress) are throwing [minified react error #152](https://reactjs.org/docs/error-decoder.html/?invariant=152&args[]=ce), which is caused when an invocation of `render` does not return anything (including `null`). Appears to be correlated with `multiOrg` flag. 

## About the bug

**Steps to reproduce:**

1. Run tests with multiOrg flag enabled in Electron (e.g. cypress).

**Expected behavior:**
No error

**Actual behavior:**
Minified react error #152

**Visual Proof:**

## About your environment

**Environment info:**
Electron

wdoconnell:
Closes #5526

This fixes a bug where globalheader could throw minified react error #152 - `Nothing was returned from render.` This was identified in Electron environments in staging, but I'm able to reproduce it elsewhere by manually setting values associated with the user's identity to `undefined`.

The cause of the issue is that the expressions used for checking whether the data associated with the header and user avatar weren't type cast to `boolean`.

When at least one element in _both_ the guards against loading the avatar and dropdown was `undefined` - for example, when the optional chaining operator cast the value to `undefined` because the property didn't exist yet - `shouldLoadGlobalHeader` also evaluated to undefined, a value that the functional component can't return, throwing an error before the user-associated data loaded.

The fix type casts to `boolean` each of the guards against loading the dropdowns; avatar; and header, and explicitly forces a `null` return if the header should not yet load.

Pre-Fix: Error thrown before header loads
--
Screen Shot 2022-08-24 at 4 18 38 PM

Post-Fix: Header only loads after all values present
--
Pre-load - some values still `undefined`
Screen Shot 2022-08-24 at 4 47 04 PM

Post-load
Screen Shot 2022-08-24 at 4 47 49 PM

### Checklist

### DIFF
--- a/src/identity/components/GlobalHeader/GlobalHeader.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeader.tsx
@@ -77,42 +77,40 @@ export const GlobalHeader: FC = () => {
     }
   }, [orgsList])
 
-  const shouldLoadDropdowns = activeOrg?.id && activeAccount?.id
+  const shouldLoadDropdowns = !!activeOrg?.id && !!activeAccount?.id
   const shouldLoadAvatar =
-    user?.firstName && user?.lastName && user?.email && org?.id
-  const shouldLoadGlobalHeader = shouldLoadDropdowns || shouldLoadAvatar
+    !!user?.firstName && !!user?.lastName && !!user?.email && !!org?.id
+  const shouldLoadGlobalHeader = !!shouldLoadDropdowns || !!shouldLoadAvatar
 
   const caretStyle = {fontSize: '18px', color: InfluxColors.Grey65}
 
-  return (
-    shouldLoadGlobalHeader && (
-      <FlexBox
-        className="multiaccountorg--header"
-        justifyContent={JustifyContent.SpaceBetween}
-        margin={ComponentSize.Large}
-        testID="global-header--container"
-      >
-        {shouldLoadDropdowns && (
-          <FlexBox margin={ComponentSize.Medium}>
-            <AccountDropdown
-              activeAccount={activeAccount}
-              activeOrg={activeOrg}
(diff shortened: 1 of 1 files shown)


ChitlangeSahas (approved):

src/identity/components/GlobalHeader/GlobalHeader.tsx
```diff
@@ -77,42 +77,44 @@ export const GlobalHeader: FC = () => {
     }
   }, [orgsList])
 
-  const shouldLoadDropdowns = activeOrg?.id && activeAccount?.id
-  const shouldLoadAvatar =
-    user?.firstName && user?.lastName && user?.email && org?.id
-  const shouldLoadGlobalHeader = shouldLoadDropdowns || shouldLoadAvatar
+  const shouldLoadDropdowns = !!(activeOrg?.id && activeAccount?.id)
+  const shouldLoadAvatar = !!(
```
hoorayimhelping, line 81: if we're already wrapping this boolean cast in parens, might as well do something like

```ts
Boolean(activeOrg?.id && activeAccount?.id)
```

which does the same thing as

```ts
!!(activeOrg?.id && activeAccount?.id)
```

but is generally more explicit and clear to people who aren't super familiar with JavaScript

hoorayimhelping (approved):
Have a minor nitpick/suggestion. But you don't have to hold up this PR for it.

src/identity/components/GlobalHeader/GlobalHeader.tsx
wdoconnell: No, I like your change - there's a mix of both `!!` and `Boolean` in the codebase, but `Boolean` is definitely clearer. Will change.

PR #5541, opened 2022-08-24
fix(5342): with flux-LSP composition released, test and debug coordination (UI vs LSP vs editor)

### LINKED ISSUE #5539: Edit Schema Browser such that when "Flux Sync" is disabled the schema browser enters a read only state (nothing can be injected) but the buckets and measurements can still be changed so the user can browse their schema.

wiedld:
Fix for #5342 
Closes #5539 

## Vid:

https://user-images.githubusercontent.com/10232835/186798228-77e5bf25-d45b-4d2a-817b-f0c8ecc76711.mp4

## Tasks Done:
- [x] found and fixed race conditions:
  * on page reload with existing schema composition:
      1. LspConnectionManager detects the schema, and sends jsonrpc to init the composition
      2. editor has not yet sent the `didOpen` for the file
      3. the schema composition init will fail, on no File being found yet.
  * on re-syncing, (or reload) with existing schema composition:
      1. all the executeCommands get sent to the Lsp too fast.
      2. Lsp spec assumes atomic changes, without a reliance on order.
           *composition looks at most recent text....and then make the applyEdit.
          * so our next read (of the current text), must occur after the previous edit has been applied
- [x] fix bugs with previous code:
  * need to resize the `clickableInvisibleDiv` each time we update the block size:
      1. change in editor text (or other event) triggers `this._setEditorBlockStyle()`
      5. it recalcs the size of the current composition block
      6. we then need to resize the `clickableInvisibleDiv` --> `this._alignInvisibleDivToEditorBlock()`
  * when no composition block is present:
      * `this._getCompositionBlockLines()` should return null. NOT a block size of 1 (on the first line).
      * Have downstream consumers of this^^ method, handle the null case.
- [x] make change per request of design:
  * when we have diverged and can no longer sync --> remove all block styles

### DIFF
--- a/src/languageSupport/languages/flux/lsp/connection.ts
+++ b/src/languageSupport/languages/flux/lsp/connection.ts
@@ -1,3 +1,4 @@
+import isEqual from 'lodash/isEqual'
 import * as MonacoTypes from 'monaco-editor/esm/vs/editor/editor.api'
 import {format_from_js_file} from '@influxdata/flux-lsp-browser'
 
@@ -7,7 +8,10 @@ import {buildUsedVarsOption} from 'src/variables/utils/buildVarsOption'
 
 // handling schema composition
 import {RecursivePartial} from 'src/types'
-import {SchemaSelection} from 'src/dataExplorer/context/persistance'
+import {
+  DEFAULT_SCHEMA,
+  SchemaSelection,
+} from 'src/dataExplorer/context/persistance'
 
 // LSP methods
 import {
@@ -36,18 +40,13 @@ class LspConnectionManager {
   private _preludeModel: MonacoTypes.editor.IModel
   private _variables: Variable[] = []
   private _compositionStyle: string[] = []
-  private _session: SchemaSelection
+  private _session: SchemaSelection = JSON.parse(JSON.stringify(DEFAULT_SCHEMA))
   private _callbackSetSession: (
     schema: RecursivePartial<SchemaSelection>
   ) => void = () => null
 
-  // handle schema composition, on reload.
-  private _firstCompositionInit = false
-  private _insertIntoBuffer = false
-  private _initBufferComposition: [
-    ExecuteCommand,
-    Omit<ExecuteCommandArgument, 'textDocument'>
-  ][] = []
+  // only add handlers on first page load.
+  private _compositionHandlersSet = false
 
   constructor(worker: Worker) {
     this._worker = worker
@@ -98,10 +97,6 @@ class LspConnectionManager {
     command: ExecuteCommand,
     data: Omit<ExecuteCommandArgument, 'textDocument'>
   ) {
-    if (this._insertIntoBuffer) {
-      return this._initBufferComposition.push([command, data])
-    }
-
     let msg
     try {
       msg = executeCommand([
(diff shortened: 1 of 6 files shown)


src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -223,27 +261,99 @@ class LspConnectionManager {
...
+      msg = this._initBufferComposition[this._o % 100]
     }
+    return
+  }
 
-    this._initLsp(schema)
+  _initComposition() {
+    // handle race condition, on page reloads
+    this._insertIntoBuffer = true
+    setTimeout(() => this._consumeInitBuffer(), 2000)
```
wiedld, line 356: This is the race condition, where we are sending executeCommands before the `didOpen` is received.
Screen Shot 2022-08-24 at 11 42 15 AM

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -186,33 +181,76 @@ class LspConnectionManager {
     })
   }
 
+  _alignInvisibleDivToEditorBlock() {
+    // elements in monaco-editor. positioned by editor.
```
wiedld, line 185: this method is code previously in `_setEditorSyncToggle()`, now in a helper method which can be used in the callback `_setEditorBlockStyle()`.

src/languageSupport/languages/flux/lsp/connection.ts
wiedld: note: I did a survey of all the event handlers in the monaco-editor. (On the editor object, and on the model.) There is nothing which fires after the didOpen was initiated.

Hence, the horrible timeout.

wiedld: once we have the middleware (Q4), then we can do this properly by listening to traffic.

src/languageSupport/languages/flux/lsp/utils.ts
```diff
@@ -71,23 +71,38 @@ export interface ExecuteCommandInjectTagValue extends ExecuteCommandInjectTag {
 
 export type ExecuteCommandInjectField = ExecuteCommandInjectMeasurement
 
-export interface ExecuteCommandCompositionInit {
+export interface CompositionInitParams {
   bucket: string
   measurement?: string
 }
 
+interface CompositionValueParams extends ExecuteCommandParams {
```
rockstar, line 79: I realize it's not really a habit we're in right now, and that these specific interfaces aren't really that complex, but it'd be _really_ helpful if these interfaces (and everything, really) came with some information (in the form of doc comments or whatever) about how they are used and _why_ they are used.

rockstar (approved):
Thanks for banging this out. I feel a lot better that this gives us a demo-able feature, with the technical debt properly on the balance sheet.

src/languageSupport/languages/flux/lsp/utils.ts
wiedld: absolutely agreed. I can add this in the code cleanup PR after the demo tmrw.
(If I make changes, then I cannot merge. So since the only changes are for code comments, I'll do that tmrw in another PR.)

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -222,55 +259,149 @@ class LspConnectionManager {
...
     // handlers to trigger end composition
     this._setEditorSyncToggle()
     this._setEditorIrreversibleExit()
 
-    // handlers for composition block size
-    // eventually, this could be from the LSP response. onLspMessage()
-    this._model.onDidChangeContent(
-      () => this._session.composition?.synced && this._setEditorBlockStyle()
-    )
+    // XXX: wiedld (25 Aug 2022) - eventually, this could be from the LSP response.
```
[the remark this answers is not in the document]
wiedld: will update in comments-only changes PR, tmrw.

PR #5542, opened 2022-08-25
feat: adds the adaptive zoom view option

### LINKED ISSUE #4988: Query Zoom (GA) - On/Off capability for adaptive zoom
ACs
- User should be able to toggle on/off the adaptive zoom per cell
- Their toggle choice should persist for that cell
- The decisions that users make at the cell level will supersede any default settings
- All users with a Solona email address should automatically have the toggle turned ON for any NEW cell that a Solona client creates. (New cells for Solona should default to adaptive zoom being turned ON.)

_Implementation will depend on whether or not Solona is an org or an account._

The following are not requirements, but rather, they're potential implementation paths depending on if Solana is an Account or an Org:
- Ability to set an org-level default for adaptive zoom or digital zoom.
- Ability to set an account-level default for adaptive zoom or digital zoom.

TCL735:
Closes #4988 

Adds the view option for the Adaptive Zoom feature.

https://user-images.githubusercontent.com/10736577/186544602-9245954e-75f6-4713-98b8-5a562a4b0c5b.mp4

### DIFF
--- a/src/visualization/components/internal/AdaptiveZoomOption.tsx
+++ b/src/visualization/components/internal/AdaptiveZoomOption.tsx
@@ -1,0 +1,72 @@
+// Libraries
+import React, {CSSProperties, FC} from 'react'
+
+// Components
+import {
+  AlignItems,
+  ComponentSize,
+  FlexBox,
+  FlexDirection,
+  InfluxColors,
+  InputLabel,
+  QuestionMarkTooltip,
+  SlideToggle,
+} from '@influxdata/clockface'
+
+// Metrics
+import {event} from 'src/cloud/utils/reporting'
+
+interface AdaptiveZoomToggleProps {
+  adaptiveZoomHide: boolean
+  update: (obj: any) => void
+  testID?: string
+  type: string
+}
+
+const getToggleColor = (toggle: boolean): CSSProperties => {
+  if (toggle) {
+    return {color: InfluxColors.Grey95}
+  }
+  return {color: InfluxColors.Grey65}
+}
+
+export const AdaptiveZoomToggle: FC<AdaptiveZoomToggleProps> = ({
+  adaptiveZoomHide,
+  update,
+  testID = 'adaptive-zoom-toggle',
+  type,
+}) => {
+  const handleSetAdaptiveZoom = () => {
+    update({
+      adaptiveZoomHide: !adaptiveZoomHide,
+    })
+    event(`visualization.customize.adaptiveZoom.${!adaptiveZoomHide}`, {
+      type,
+    })
+  }
+
+  return (
+    <FlexBox
+      direction={FlexDirection.Row}
+      alignItems={AlignItems.Center}
+      margin={ComponentSize.Medium}
+      stretchToFitWidth={true}
+      className="adaptive-zoom-toggle"
+      testID={testID}
+    >
+      <SlideToggle
+        active={!adaptiveZoomHide}
+        size={ComponentSize.ExtraSmall}
+        onChange={handleSetAdaptiveZoom}
+      />
+      <InputLabel style={getToggleColor(adaptiveZoomHide)}>
+        Adaptive Zoom
+      </InputLabel>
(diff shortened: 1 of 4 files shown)


src/visualization/components/internal/AdaptiveZoomOption.tsx
```diff
@@ -0,0 +1,72 @@
...
+  adaptiveZoomHide,
+  update,
+  testID = 'adaptive-zoom-toggle',
+  type,
+}) => {
+  const handleSetAdaptiveZoom = () => {
+    update({
+      adaptiveZoomHide: !adaptiveZoomHide,
+    })
+    event(`visualization.customize.adaptiveZoom.${!adaptiveZoomHide}`, {
```
hoorayimhelping, line 43: Does using `!adaptiveZoomHide` invert the meaning of what happened? Or does it describe what the value is changing to?

We generally want to move away from variable event names, but since this is a boolean value, it's not a huge deal. One stylistic change I'd suggestion is consider making it more human readable, by saying something like `on` or `off` instead of `true` or `false`, so it's more clear to people analazying the data what happened

```ts
event(`visualization.customize.adaptiveZoom.${adaptiveZoomHide ? 'toggleOn' : 'toggleOff'}`
```

or something to that effect.

src/visualization/components/internal/AdaptiveZoomOption.tsx
```diff
@@ -0,0 +1,72 @@
...
+  if (toggle) {
+    return {color: InfluxColors.Grey95}
+  }
+  return {color: InfluxColors.Grey65}
+}
+
+export const AdaptiveZoomToggle: FC<AdaptiveZoomToggleProps> = ({
+  adaptiveZoomHide,
+  update,
+  testID = 'adaptive-zoom-toggle',
```
hoorayimhelping, line 36: does this need to be changeable? It looks like the only place this is used is here, and it's not changing the default test id. YAGNI

src/visualization/components/internal/AdaptiveZoomOption.tsx
```diff
@@ -0,0 +1,72 @@
...
+        size={ComponentSize.ExtraSmall}
+        onChange={handleSetAdaptiveZoom}
+      />
+      <InputLabel style={getToggleColor(adaptiveZoomHide)}>
+        Adaptive Zoom
+      </InputLabel>
+      <QuestionMarkTooltip
+        diameter={16}
+        tooltipContents="When enabled, will re-query on the selected domain when zooming in on the graph. Requires window period set to Auto."
+        tooltipStyle={{fontSize: '13px', padding: '8px'}}
```
hoorayimhelping, line 68: Recommend pulling this `{fontSize: '13px', padding: '8px'}` out into a variable defined outside the function component so it doesn't cause re-renders of the options panel when the graph is being rendered

hoorayimhelping (approved):

src/visualization/components/internal/AdaptiveZoomOption.tsx
TCL735: Yup. I moved this out along with creating the new tooltip.

src/visualization/components/internal/AdaptiveZoomOption.tsx
TCL735: `adaptiveZoomHide` is the current setting. The toggle flips the current setting to the other setting. So, `!adaptiveZoomHide` would be the opposite of the current setting, or flipping the current setting.

TCL735: i would argue that 

```
event(`visualization.customize.adaptiveZoom.${adaptiveZoomHide ? 'toggleOn' : 'toggleOff'}
```

is more confusing that necessary. Because we are not firing an event on the current state of `adaptiveZoomHide`. We are firing an event when the user flips the toggle.

ChitlangeSahas (approved):

PR #5543, opened 2022-08-25
feat: save scripts

### LINKED ISSUE #5531: Connect the Save Script button from the modal
As a user, when I am prompted to a Save a Script, I should be able to save the editor text as a Script.

Screen Shot 2022-08-24 at 11 24 22 AM

asalem1:
Closes #5531

The purpose of this PR is to integrate the save functionality for saving a Script from the flux query builder. This PR is strictly a means of resource generation and is not currently integrated with the work that's being done in parallel to sync the currently generated resource into the lifecycle of the user's experience with the product. More specifically, this manifests itself with no resource persistence awareness in the current implementation - this is something that is being developed in concert and can be found at a high level here:

https://github.com/influxdata/ui/pull/5540/files

save-script

### DIFF
--- a/src/dataExplorer/context/scripts.tsx
+++ b/src/dataExplorer/context/scripts.tsx
@@ -1,12 +1,12 @@
 import React, {FC, createContext, useContext, useCallback} from 'react'
-import {useDispatch} from 'react-redux'
-import * as api from 'src/client/scriptsRoutes'
-import {notify} from 'src/shared/actions/notifications'
-import {
-  scriptSaveFail,
-  scriptSaveSuccess,
-} from 'src/shared/copy/notifications/categories/scripts'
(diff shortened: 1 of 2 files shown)


src/dataExplorer/context/scripts.tsx
```diff
@@ -0,0 +1,97 @@
...
+  name: `Untitled Script: ${new Date().toISOString()}`,
+
+  handleSave: () => {},
+  updateDescription: (_description: string) => {},
+  updateName: (_name: string) => {},
+}
+
+export const ScriptContext = createContext<ContextType>(DEFAULT_CONTEXT)
+
+export const ScriptProvider: FC = ({children}) => {
```
drdelambre, line 37: i figure imma be replacing most of this, but before i blindly do so, what is important to keep in this provider?

drdelambre (approved):

PR #5552, opened 2022-08-25
feat: adding generic resource interface

drdelambre:
part of the split up of the giant save/load pr (found here: https://github.com/influxdata/ui/pull/5540/files).

In this chapter, we add a generic `/from` templating system for the new data explorer that'll allow us to load any registered resource into the new flux editor and save it out again.

### DIFF
--- a/src/dataExplorer/components/resources/TemplatePage.tsx
+++ b/src/dataExplorer/components/resources/TemplatePage.tsx
@@ -1,0 +1,65 @@
+import React, {FC, useEffect, useState, useContext} from 'react'
+import {useSelector} from 'react-redux'
+import {Switch, Route, useHistory, useParams} from 'react-router-dom'
+import {RemoteDataState} from 'src/types'
+
+import {getOrg} from 'src/organizations/selectors'
+
+import {RESOURCES} from 'src/dataExplorer/components/resources'
+import {
+  PersistanceContext,
+  PersistanceProvider,
+} from 'src/dataExplorer/context/persistance'
+
+const Template: FC = () => {
+  const {setQuery, setResource} = useContext(PersistanceContext)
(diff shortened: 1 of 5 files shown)


src/dataExplorer/components/resources/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  ) => Promise<ResourceConnectedQuery<any>>
+}
+
+interface Resources {
+  [type: string]: ResourceRegistration
+}
+
+export const RESOURCES: Resources = {}
+  // eslint-disable-next-line no-extra-semi
+;[].forEach(mod => {
```
appletreeisyellow, line 26: Is the `forEach` function intentionally applied on an empty array as a placeholder for the future code?

appletreeisyellow (approved):
LGTM! Just one question below

src/dataExplorer/components/resources/TemplatePage.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const org = useSelector(getOrg)
+
+  return (
+    <PersistanceProvider>
+      <Switch>
+        <Route
+          path={`/orgs/${org.id}/data-explorer/from/*`}
+          component={Template}
+        />
+        <Route component={() => <div />} />
```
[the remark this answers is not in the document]
drdelambre: Oh gotcha. I misunderstood your last comment. I think we want the user to know they have a bad link and that it didn't just succeed in the background yeah? There's def a better way of doing that than this, bit i think succeeding in an error is wrong

src/dataExplorer/components/resources/index.ts
drdelambre: YES

PR #5555, opened 2022-08-25
feat: add generic script loader

drdelambre:
this is part of splitting up the big PR at https://github.com/influxdata/ui/pull/5540/files in order to make the review cycle more digestible. this piece adds the generic interface for loading and saving scripts into the persistance layer of the new data explorer

### DIFF
--- a/src/dataExplorer/components/resources/index.ts
+++ b/src/dataExplorer/components/resources/index.ts
@@ -23,7 +23,7 @@ interface Resources {
 
 export const RESOURCES: Resources = {}
   // eslint-disable-next-line no-extra-semi
-;[].forEach(mod => {
+;[require('./types/script')].forEach(mod => {
   mod.default((def: ResourceRegistration) => {
     if (RESOURCES.hasOwnProperty(def.type)) {
       throw new Error(
--- a/src/dataExplorer/components/resources/types/script/index.ts
+++ b/src/dataExplorer/components/resources/types/script/index.ts
@@ -1,6 +1,6 @@
 import {ResourceType} from 'src/types/resources'
 import editor from './editor'
-import {getScript, patchScript, postScript} from 'src/client/scriptsRoutes.ts'
+import {getScript, patchScript, postScript} from 'src/client/scriptsRoutes'
 
 export default function script(register) {
   register({
--- a/src/dataExplorer/components/resources/types/script/index.ts
+++ b/src/dataExplorer/components/resources/types/script/index.ts
@@ -1,0 +1,61 @@
+import {ResourceType} from 'src/types/resources'
+import editor from './editor'
+import {getScript, patchScript, postScript} from 'src/client/scriptsRoutes.ts'
+
+export default function script(register) {
+  register({
+    type: ResourceType.Scripts,
+    editor,
+    init: id => {
+      if (!id) {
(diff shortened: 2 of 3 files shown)


src/dataExplorer/components/resources/index.ts
```diff
@@ -23,7 +23,7 @@ interface Resources {
 
 export const RESOURCES: Resources = {}
   // eslint-disable-next-line no-extra-semi
-;[].forEach(mod => {
+;[require('./types/script')].forEach(mod => {
```
rockstar, line 26: Why do we add a dynamic import here?

drdelambre: this isn't dynamic, it's resolved at build time

src/dataExplorer/components/resources/types/script/index.ts
```diff
@@ -0,0 +1,61 @@
...
+          },
+        })
+      }
+
+      return getScript({
+        scriptID: id,
+      }).then(resp => {
+        return {
+          type: ResourceType.Scripts,
+          flux: resp.script,
```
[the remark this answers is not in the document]
drdelambre: i'm hearing that. but i think the idea would be to add a `sql` field after we implement the interface to support it on the other side and write the translator once we cross that bridge. all that to say.. i think iox means we support flux AND sql at the same time.

src/dataExplorer/components/resources/index.ts
rockstar: I don't want to split hairs. It's not resolved (easily) at code read time, especially from the consumer of `RESOURCES`, which is why I'm asking. As someone new to this codebase, I've seen this pattern used a couple of times now, and it takes me 5x the time to track down what's actually happening. It breaks IDEs and tools that would help reason about the code.

My concern is that it's clever, but not clear, and I tend to prefer clear. If there's a reasoning for it, I'd like to understand it.

wiedld: Let's define terms:
1. static check time
2. webpack build time
3. runtime

Code navigation and our static analysis tools all occur at step 1. Which is before the generation of this interface map. That is why our IDE navigation tools fail.

drdelambre: not splitting hairs. i gotta be honest, i'm not quite following where the confusion is. i'm not seeing how calling a default export to update an object is particularly clever, and encapsulating transformation by resource type is something that most of our reducers do across the codebase (some do it by feature, but that's gotten us in trouble). what's the missing piece i'm not seeing here?

I sorry i can't help you with why is your ide is not able to follow the `require` statement outlined above, maybe updating the lsp might help? i've been using this plugin with neovim for a couple months and i haven't seen a lot of gaps from my expectations: https://github.com/neoclide/coc-tsserver. but i'd expect you to be having a lot more issues around the code base than just here as we use the `require('src/client/generatedRoutes')` reference a lot to get around linting for OSS where OSS doesn't have the same endpoints as cloud.

drdelambre:
this is blocking other prs and progress on save / load,  and i'm not getting responses, so i'm gonna merge it. lets offline this discussion and see if we can close the gap and turn it into something productive next week when everyone is together at the same time.

PR #5559, opened 2022-08-26
feat: add generic task loader

drdelambre:
this is part of splitting up the big PR at https://github.com/influxdata/ui/pull/5540/files in order to make the review cycle more digestible. this piece adds the generic interface for loading and saving tasks into the persistence layer of the new data explorer


wiedld (changes requested):
Can we have a conversation first, about how we can use a different way of building the interface map -- such that our static tools (IDE navigation, entry point analysis) can follow?

I believe the updates required would be very quick..so the main point would be a convo first please!

drdelambre:
i do not understand your concern

PR #5584, opened 2022-08-29
feat: update naming conventions and saving restrictions

### LINKED ISSUE #5577: Disable the Save button for the Script Editor in specific usages
The `Save` button should be disabled if the user hasn't made any changes on a new Script, or if they haven't made any additional changes on a previously saved Script

asalem1:
Closes #5577

This PR resolves a couple of issues around the saving state. While the issue linked does reference renaming Scripts as an actionable item to allow users to know they are in a savable state, it turns out that the API doesn't support those types of transactions as per the [following thread](https://influxdata.slack.com/archives/C03G35C0NTE/p1661804577248349). As such, this PR simplifies the scope of saving to make the button disabled based on whether any action has been taken on the resource based on the previously clean state. More specifically, a user should NOT be able to rename the Script once it's been created. In addition, they shouldn't be able to Save the Script after making no modifications on a loaded script, or if they have just saved the script

### DIFF
--- a/cypress/e2e/shared/fluxQueryBuilder.test.ts
+++ b/cypress/e2e/shared/fluxQueryBuilder.test.ts
@@ -24,9 +24,7 @@ describe('FluxQueryBuilder', () => {
               if (!$toggle.hasClass('active')) {
                 // hasClass is a jQuery function
                 $toggle.click()
-                cy.getByTestID('flux-query-builder--menu').contains(
-                  'New Script'
-                )
+                cy.getByTestID('flux-query-builder--menu').contains('Clear')
               }
             })
           })


src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -44,33 +43,17 @@ import {PROJECT_NAME, PROJECT_NAME_PLURAL} from 'src/flows'
...
-        ...resource.data,
-        name,
-      },
-    })
-  }
+  let pageTitle = 'Data Explorer'
 
-  let pageTitle = <Page.Title title="Data Explorer" />
   if (fluxQueryBuilder && resource?.data?.hasOwnProperty('name')) {
-    pageTitle = (
```
drdelambre, line 66: can you limit this down to ResourceType.Scripts instead of removing it?

drdelambre (dismissed):
i'm wary of dirty flags like some people are wary of while loops, but this looks good

drdelambre (approved):

PR #5606, opened 2022-08-31
fix: handle OSS instance resource

jeffreyssmith2nd:
OSS recently added a new resource type, `instance`. This resource is currently not intended to be included in any allAccess/operator tokens. This PR should address that (the e2e tests are currently failing against OSS) and also help move us forward towards unpinning the UI.

### Checklist

### DIFF
--- a/src/authorizations/utils/permissions.ts
+++ b/src/authorizations/utils/permissions.ts
@@ -49,7 +49,9 @@ export const allAccessPermissions = (
   userID: string
 ): Permission[] => {
   const withOrgID = ensureT(orgID, userID)
-  return allPermissionTypes.filter(perm => String(perm) !== "instance").flatMap(withOrgID)
+  return allPermissionTypes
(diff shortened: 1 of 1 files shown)


src/authorizations/utils/permissions.ts
```diff
@@ -49,7 +49,9 @@ export const allAccessPermissions = (
   userID: string
 ): Permission[] => {
   const withOrgID = ensureT(orgID, userID)
-  return allPermissionTypes.flatMap(withOrgID)
+  return allPermissionTypes
+    .filter(perm => String(perm) !== 'instance')
```
jeffreyssmith2nd, line 53: The reason for converting the type to a string is that the instance type does not exist at build time in a cloud build, so you get the `this line will always be true` linting error.

If that would be preferred over this, happy to change it or if there is a better way to do this in general.

hoorayimhelping (approved):

Issue #5607, opened 2022-08-31
dashboard can have bottom of screen clipped with globalheader enabled


wdoconnell:
When viewing dashboards with multi-org on, the bottom of the screen can be clipped at the bottom, obscuring the bottom-most part of the dashboard.

## About the bug

**Steps to reproduce:**

1. Turn multi-org on
2. Navigate to a dashboard that touches the bottom of the page.

**Expected behavior:**
Same visibility as with multi-org on

**Actual behavior:**
Bottom-most item can be cut off

**Visual Proof:**
![image (2)](https://user-images.githubusercontent.com/91283923/187705592-0dfec83f-a917-4f0c-b2c7-6b167271a880.png)

PR #5620, opened 2022-09-01
fix: adjust page-specific styles for globalheader

### LINKED ISSUE #5603: New Script Editor does not show pagination numbers
## About the bug

There is a bug [reported](https://influxdata.slack.com/archives/C030G5783M3/p1661885938946989) by @aanthony1243 in new script editor, after run a query, the result table doesn't have pagination numbers to click on or no vertical scroll bar to see all the results. 

So far this bug is only in new tools, which may related to the new multi-org heading.

**Steps to reproduce:**

1. Use new tools
2. In new script editor
3. Run a basic query

**Expected behavior:**

Should have pagination numbers for users to click on in the table result

**Actual behavior:**

Cannot find pagination numbers to click on

**Visual Proof:**

Screen Shot 2022-08-30 at 3 25 13 PM

wdoconnell:
Connects #5607  
Closes #5603 

When multi-org is on, the bottom of certain pages is being clipped. The primary cause is that our existing `Page` components and `DapperScrollBars` presume that 100% of the viewport is available for rendering the page. The `GlobalHeader` component is out of `Page`, so we end up with a slight overflow on certain pages beyond the viewport (equal in size to the height of `GlobalHeader`).

Adjusting `Page` size conditionally does not work consistently on all pages due to minor formatting differences, so some custom page-specific adjustments are needed.

This PR resolves the cutoff on all pages on which I've located this problem except for pages using `MonacoEditor`. I'll make that the subject of a separate PR, as it requires more investigation, and need not hold up this fix.

Classic UI (#5607)
--
https://user-images.githubusercontent.com/91283923/187976452-e2af7790-1806-425a-b1d1-47f62190e4a0.mov

New Data Explorer & Script Editor (#5603)
--
https://user-images.githubusercontent.com/91283923/187992538-d8a19ae2-fb65-46e8-828d-a57918578241.mov

### Checklist

### DIFF
--- a/src/MultiOrgOverrideStyles.scss
+++ b/src/MultiOrgOverrideStyles.scss
@@ -12,7 +12,12 @@
   height: calc(100% - gh.$globalheader-height);
 }
 
-// Page-specific fix for Data Explorer
+// Page-specific fix for Data Explorer - Classic
 .multi-org .data-explorer > .time-machine > .cf-draggable-resizer {
+  height: calc(100% - gh.$globalheader-height);
+}
+
+// Page-specific fix for Data Explorer - New
+.multi-org .flux-query-builder--container {
   height: calc(100% - gh.$globalheader-height);
 }
--- a/src/MultiOrgOverrideStyles.scss
+++ b/src/MultiOrgOverrideStyles.scss
@@ -1,5 +1,18 @@
+@use 'src/identity/components/GlobalHeader/GlobalHeaderStyle.scss' as gh;
+
+// Reduce size of page headers
 .multi-org .cf-page-header {
-  height: 45px !important;
-  flex: 0 0 45px !important;
-  margin-bottom: 12px !important;
+  height: 46px !important;
+  flex: 0 0 46px !important;
+  margin-bottom: 14px !important;
+}
+
+// Adjust most pages for multi-org by changing the size of the Dapper Scroll Bar wrapper.
+.multi-org .cf-page-contents > .ScrollbarsCustom-Wrapper {
+  height: calc(100% - gh.$globalheader-height);
+}
+
+// Page-specific fix for Data Explorer
+.multi-org .data-explorer > .time-machine > .cf-draggable-resizer {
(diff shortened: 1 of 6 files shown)


src/identity/components/GlobalHeader/GlobalHeaderStyle.scss
```diff
@@ -1,10 +1,13 @@
+// If this height is changed also change GLOBAL_HEADER_HEIGHT constant.
```
ChitlangeSahas, line 1: 👍

src/MultiOrgOverrideStyles.scss
```diff
@@ -1,5 +1,23 @@
...
+}
+
+// Page-specific fix for Data Explorer - Classic
+.multi-org .data-explorer > .time-machine > .cf-draggable-resizer {
+  height: calc(100% - gh.$globalheader-height);
+}
+
+// Page-specific fix for Data Explorer - New
+.multi-org .flux-query-builder--container {
+  height: calc(100% - gh.$globalheader-height);
```
appletreeisyellow, line 22: Nice!

appletreeisyellow (approved):
Looks good to me! I like you @use the variable `$globalheader-height` and make the code easy to read and maintain

src/identity/components/GlobalHeader/constants/index.ts
```diff
@@ -0,0 +1 @@
+export const GLOBAL_HEADER_HEIGHT = 60
```
hoorayimhelping, line 1: I think a similar comment added saying change `$globalheader-height` in `GlobalHeaderStyles.scss` would be useful

hoorayimhelping (dismissed):

src/identity/components/GlobalHeader/constants/index.ts
wdoconnell: 👍 Good idea.

hoorayimhelping (approved):

PR #5625, opened 2022-09-01
fix: bump the bytes number for cloud

### LINKED ISSUE #5608: Script Editor quickly goes unresponsive with any large dataset
When trying to query using the new query editor, if you choose any large dataset the tab will become unresponsive.

asalem1:
Closes #5608 

This PR adds a feature flagged value for product to up the limits on the number of bytes a CSV should be truncated at, with a baseline of the 27 MiB limit that's currently set for CLOUD in the original data explorer. Having said that, doing so doesn't seem to resolve slowness around the flux query builder. Namely, there seems to be two issues:

1. app-metrics is taking a LONG time to resolve (could be a red-herring and a symptom of an issue)
2. state management looks like it's going haywire for some reason in the burndown chart from current performance reviews. Those seem unrelated to these changes and should be addressed moving forward without complicating these changes

### DIFF
--- a/src/shared/contexts/query.tsx
+++ b/src/shared/contexts/query.tsx
@@ -593,9 +593,6 @@ export const QueryProvider: FC = ({children}) => {
               getFlagValue('increaseCsvLimit') ?? FLUX_RESPONSE_BYTES_LIMIT
 
             while (!read.done) {
-              if (!pending.current[id]) {
-                throw new CancellationError()
-              }
               const text = decoder.decode(read.value)
 
               bytesRead += read.value.byteLength
@@ -611,9 +608,6 @@ export const QueryProvider: FC = ({children}) => {
             }
 
             reader.cancel()
-            if (pending.current[id]) {
-              delete pending.current[id]
-            }
 
             return {
               type: 'SUCCESS',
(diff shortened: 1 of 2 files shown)


src/shared/constants/index.ts
```diff
@@ -109,7 +109,7 @@ export const CLOUD_USAGE_PATH = '/usage'
 export const CLOUD_LOGOUT_PATH = '/logout'
 
 export const FLUX_RESPONSE_BYTES_LIMIT = CLOUD
-  ? 27 * 1024 * 1024 // 27 MiB  (desa: this was determined by looking at queries responses in the cloud app)
```
drdelambre, line 112: This context might be important to share with product, that something like 99% of queries done at that time fell below this previous threshold. Maybe they can find a new "best" number?

src/shared/contexts/query.tsx
```diff
@@ -57,6 +61,35 @@ interface RequestBody {
...
+  we want to trim the last incomplete line, so that the result is
+
+            r,baz,3
+      foo,bar,baz,2
+
+*/
+const trimPartialLines = (partialResp: string): string => {
+  let i = partialResp.length - 1
+
+  while (partialResp[i] !== '\n') {
```
drdelambre, line 83: What's the added overhead of this? It's reparsing the CSV right?

drdelambre (approved):

drdelambre (approved):
nice flag addition

PR #5629, opened 2022-09-01
fix: update the query results for adaptive zoom only for a non-empty response

### LINKED ISSUE #5627: Query Zoom (GA) - Bug: zooming in over an empty time period causes an error
Adaptive Zoom does not interpret a time period with no data correctly. It will return an error and indicate the data cannot be visualized. However, with this feature turned off, the graph never returns an error. Hence, Adaptive Zoom is not correctly handling a time period with no data.

With Adaptive Zoom on empty time period, returns an error:

https://user-images.githubusercontent.com/10736577/188018197-8dfb983a-7290-4fca-bbbe-d8ce6eca1066.mp4

Without Adaptive Zoom on empty time period, graph remains:

https://user-images.githubusercontent.com/10736577/188018426-a9b9a63c-068a-4a3a-bcdc-ed50debeea7b.mp4

TCL735:
Closes #5627 

Updates the query results for an adaptive zoom only when the response is non-empty. This prevents the graph from erroring out.

Adaptive Zoom over empty time period:

https://user-images.githubusercontent.com/10736577/188025067-0ff43a23-48fa-4f81-a2ea-4ae16f28f23d.mp4

### DIFF
--- a/src/visualization/utils/useVisDomainSettings.ts
+++ b/src/visualization/utils/useVisDomainSettings.ts
@@ -242,10 +242,11 @@ export const useZoomRequeryXDomainSettings = (args: ZoomRequeryArgs) => {
         setRequeryStatus(RemoteDataState.Loading)
         runQuery(orgId, query, extern).promise.then(
           (result: RunQueryResult) => {
-            if (result.type === 'SUCCESS') {
+            if (result?.type === 'SUCCESS') {
               setRequeryStatus(RemoteDataState.Done)
-              const parsed = fromFlux(result.csv)
-              setResult(parsed)
+              if (result?.csv?.trim().length > 0) {
+                setResult(fromFlux(result.csv))
+              }
             } else {
               setRequeryStatus(RemoteDataState.Error)
             }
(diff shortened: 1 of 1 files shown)


src/visualization/utils/useVisDomainSettings.ts
```diff
@@ -242,10 +242,11 @@ export const useZoomRequeryXDomainSettings = (args: ZoomRequeryArgs) => {
         setRequeryStatus(RemoteDataState.Loading)
         runQuery(orgId, query, extern).promise.then(
           (result: RunQueryResult) => {
-            if (result.type === 'SUCCESS') {
+            if (result?.type === 'SUCCESS') {
               setRequeryStatus(RemoteDataState.Done)
-              const parsed = fromFlux(result.csv)
-              setResult(parsed)
+              if (result?.csv?.trim().length > 0) {
```
hoorayimhelping, line 247: I think this line is fine, but I want to make sure it doesn't open this expression up to something like `undefined.trim()`.

hoorayimhelping (approved):

src/visualization/utils/useVisDomainSettings.ts
TCL735: No, if any part of the chain is `undefined` then it becomes `if (undefined > 0)` which is false

TCL735: Here is a quick proof from the console

Screen Shot 2022-09-02 at 9 24 13 AM

Issue #5575, opened 2022-08-29
Schema Composition: cleanup on initial PoC


wiedld:
Reconsider/address decisions made with initial PoC. Decide which is feasible, and the timeline:
- [ ] We need to add increased integration between the LSP layer and the UI code so that errors that occur can be communicated bi-directionally.
- [ ] First time schema composition is turned on after page load occurs there is a hardcoded 3 second delay to prevent race conditions.
- [ ] Race conditions can occur when a user clicks too many elements within Schema Browser too quickly.
- [ ] When a composition block is large and then the user changes the measurement the composition block must be entirely rebuilt and is occasionally generating errors that desynchronize the LSP from the UI.
- [ ] Any general error occurring within the LSP will desynchronize it from the UI and the user will be required to reload the page to reinitialize the composition block.
- [ ] “Flux Sync’ is a One Time One way process.
   - [ ] We want to add bi-directional capability. I.e. the ability to recreate the schema browser state from an existing query state.
   - [ ] We want to add support for multiple schema composition blocks with the caveat that only one block may be active in the sync functionality at one time.
- [ ] We did not yet implement the default query text comment on page load which was included in design documentation. It is a small addition but it may interfere with default schema composition state so we wanted to be careful around testing it.

Small items that we can quickly resolve:
- [ ] Flux Sync should be turned on by default.

rockstar:
> First time schema composition is turned on after page load occurs there is a hardcoded 3 second delay to prevent race conditions.

This hard-coded three seconds is mostly a guess, and on systems with slower cpus or slower connections, this is still likely to break. The solution could be to raise that limit to 5 seconds, but where does it end? 10 seconds?  30 seconds?

One of the things react et al gives us that is otherwise a more straightforward API for managing state, and allowing components to update and change based on state (via javascript's event-based model). In this situation, what we're missing is a state that says "we're ready for composition to be turned on." Composition shouldn't be "turned on" until it can be turned on in the place where it matters.

> Race conditions can occur when a user clicks too many elements within Schema Browser too quickly.

In this case, it's because we're again assuming that we don't have state/change events that allow us to see when changes are applied. It is the UI component's responsibility to handle input too quick (we debounce in text boxes, for instance). There are a number of ways we can hack around this, but the most straightforward way is likely to add a new composition criteria and not allow anymore composition criteria to be added until the query itself changes (likely by checking that the change gets updated in `QueryContext`, but I'm still trying to get my head around all the places where the query is stored.

[...]

PR #5630, opened 2022-09-01
feat(5622): reinit full composition block at once

### LINKED ISSUE #5622: Schema Composition: re-initalize a full schema composition block
### Problem:
* on re-sync with an entire change in the schema composition block, with new measurements + fields + tagValues, right now that's a series of request/response loops (executeCommand and applyEdits). This has a ramification of: (1) increased chance of getting out of sync, and (2) takes more time. 
* instead, have a single composition block initialize command be able to handle fields and tagValues too. So is a single request/response.

### Solution:
- [x] changes in flux-lsp
         * https://github.com/influxdata/flux-lsp/pull/575
- [x] changes in the UI to use it

wiedld:
Closes #5622 

## Done:
* updated resyncing (and page reloads):
    * **Problem:** on re-sync with an entire change in the schema composition block, with new measurements + fields + tagValues, right now that's a series of request/response loops (executeCommand and applyEdits). This has a ramification of: (1) increased chance of getting out of sync, and (2) takes more time.
    * **Solution:**  instead, have a single composition block initialize command be able to handle fields and tagValues too. So is a single request/response.
* fixed heuristic used in `_editorChangeIsFromLsp()`, to be accurate (based on LSP implementation).

## Vid:

https://user-images.githubusercontent.com/10232835/188017441-4c9a602e-62da-476e-9bcd-fa655dc55e1d.mov

## Future:
We replace the composition block entirely each time, in the applyEdit from the LSP. Arguably, we could start just doing a full reinit with every session change, and send over the latest session (instead of this add versus remove 1 element). 

The only downside of this approach is that we don't have a clean failure case. Meaning, if an `addField` fails in the LSP --> the middleware could detect and simply sync the local UI session (remove the field from the session). This simple failure action would not be possible if we always do a re-init.

### DIFF
--- a/src/languageSupport/languages/flux/lsp/connection.ts
+++ b/src/languageSupport/languages/flux/lsp/connection.ts
@@ -155,8 +155,7 @@ class LspConnectionManager {
   }
 
   _editorChangeIsFromLsp(change) {
-    // heuristic. Not 100% accurate.
-    return !!change.forceMoveMarkers
+    return change.text?.includes('|> yield(name: "_editor_composition")')
   }
 
   _editorChangeIsWithinComposition(change) {
--- a/src/languageSupport/languages/flux/lsp/connection.ts
+++ b/src/languageSupport/languages/flux/lsp/connection.ts
@@ -311,13 +311,34 @@ class LspConnectionManager {
                     * buffer of executeCommands, send to Lsp at a throttled pace (hack timeouts)  
                   * longterm: middleware? changes in Lsp?
     */
-    if (toAdd.bucket || toAdd.measurement) {
-      const payload = {bucket: toAdd.bucket?.name || this._session.bucket?.name}
-      if (toAdd.measurement) {
-        payload['measurement'] = toAdd.measurement
+    const numFieldChanges =
+      (toAdd.fields?.length || 0) + (toRemove.fields?.length || 0)
+    const numTagValueChanges =
+      (toAdd.tagValues?.length || 0) + (toRemove.tagValues?.length || 0)
+    const reInitBlock =
+      toAdd.bucket ||
+      toAdd.measurement ||
+      numFieldChanges + numTagValueChanges > 1
+
+    if (reInitBlock) {
+      const payload = {
+        bucket: toAdd.bucket?.name || this._session.bucket?.name,
+      }
+      if (toAdd.measurement || this._session.measurement) {
+        payload['measurement'] = toAdd.measurement || this._session.measurement
+      }
+      if (toAdd.fields || this._session.fields) {
+        payload['fields'] = toAdd.fields || this._session.fields
+      }
+      if (toAdd.tagValues || this._session.tagValues) {
+        payload['tagValues'] = (
+          toAdd.tagValues || this._session.tagValues
+        ).map(({key, value}) => [key, value])
       }
       this._insertBuffer([ExecuteCommand.CompositionInit, payload])
+      return // re-initialize full block. no more requests needed.
     }
+
     if (toRemove.fields?.length) {
       toRemove.fields.forEach(value =>
         this._insertBuffer([ExecuteCommand.CompositionRemoveField, {value}])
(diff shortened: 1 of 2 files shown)


src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -155,8 +155,7 @@ class LspConnectionManager {
   }
 
   _editorChangeIsFromLsp(change) {
-    // heuristic. Not 100% accurate.
-    return !!change.forceMoveMarkers
```
wiedld, line 159: Made this change because of heuristic failing:
(see the 3rd change object).

Screen Shot 2022-09-01 at 1 57 32 PM

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -311,13 +310,34 @@ class LspConnectionManager {
                     * buffer of executeCommands, send to Lsp at a throttled pace (hack timeouts)  
                   * longterm: middleware? changes in Lsp?
     */
```
appletreeisyellow, line 312: Should this comment also be updated since each time the the buffer updates, the entire composition is reinitiated?

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -311,13 +310,34 @@ class LspConnectionManager {
...
+      if (toAdd.tagValues || this._session.tagValues) {
+        payload['tagValues'] = (
+          toAdd.tagValues || this._session.tagValues
+        ).map(({key, value}) => [key, value])
       }
       this._insertBuffer([ExecuteCommand.CompositionInit, payload])
+      return // re-initialize full block. no more requests needed.
     }
+
     if (toRemove.fields?.length) {
```
appletreeisyellow, line 341: Is there any case the rest of the code in this function will be hit? If not, then we may want to remove them unless you intentionally leave them there

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -311,13 +310,34 @@ class LspConnectionManager {
...
+      if (toAdd.measurement || this._session.measurement) {
+        payload['measurement'] = toAdd.measurement || this._session.measurement
+      }
+      if (toAdd.fields || this._session.fields) {
+        payload['fields'] = toAdd.fields || this._session.fields
+      }
+      if (toAdd.tagValues || this._session.tagValues) {
+        payload['tagValues'] = (
+          toAdd.tagValues || this._session.tagValues
+        ).map(({key, value}) => [key, value])
```
appletreeisyellow, line 335: Just wondering, why you convert each tag-value object into an array? Is that because LSP expects array instead of object?

src/languageSupport/languages/flux/lsp/connection.ts
wiedld: we still use the buffer, because it does:
* re-makes the entire composition on re-sync (the change here)
* still buffers user click-by-click for all other actions.
   * these actions are not a re-init.

in most cases, messages will be inserted into the buffer and then immediately incremented. Exceptions being:
* the user is clicking faster than the roundtrip request/response with the LSP
* or we still in the page (re-)load process


I updated the checklist under the description here, to make evident what still remains unfixed. https://github.com/influxdata/ui/issues/5575 The question is how frequently we'll see users getting out-of-sync; thereby setting the urgency for the middleware.

src/languageSupport/languages/flux/lsp/connection.ts
wiedld: Yes. Anytime the user is not reselected a bucket or measurement, or not re-syncing.

In the other cases, when the user clicks a single field or tagValue, then this single change request (non-init) is called.

src/languageSupport/languages/flux/lsp/connection.ts
wiedld: exactly.
https://github.com/influxdata/flux-lsp/blob/master/src/server/commands.rs#L163

src/languageSupport/languages/flux/lsp/connection.ts
appletreeisyellow: Ah! Got it

appletreeisyellow (approved):
🙌

Issue #3970, opened 2022-02-26
Tech Debt: Telegraf UI: refactor overlays to use overlay controller


TCL735:
The Telegraf UI Refresh created new components that use overlays. The overlays are triggered with routing. This is not the best way to handle the overlays when the overlays need to appear at different routes, because it will mean changing the url every time the user triggers the overlay. Instead, use the overlay controller which does not change the route (url).

- [ ] refactor the TelegrafsPage to use the OverlayController to trigger the overlay (rather than a route)
- [ ] refactor the BucketCard to use the OverlayController to trigger the overlay (re-using the above shared

TCL735:
Split from https://github.com/influxdata/ui/issues/3747

PR #5640, opened 2022-09-03
feat(telegrafPage): use OverlayController to open instructions overlay

blegesse-git:
Part of #3970 

This is a part of a bigger pr that refactors `TelegrafsPage` to use overlay controller instead of React Router.
This pr removes the `/instructions` route for `TelegrafInstructionsOverlay` and connects it to the overlay controller so it can be triggered without route change.

https://user-images.githubusercontent.com/66275100/188704485-238274d8-cf39-4273-bbfd-103590364092.mov

### Checklist

### DIFF
--- a/src/telegrafs/components/CollectorCard.tsx
+++ b/src/telegrafs/components/CollectorCard.tsx
@@ -24,6 +24,7 @@ import {
   addTelegrafLabelAsync,
   removeTelegrafLabelAsync,
 } from 'src/telegrafs/actions/thunks'
+import {showOverlay, dismissOverlay} from 'src/overlays/actions/overlays'
 
 import {cloneTelegraf} from 'src/telegrafs/actions/thunks'
 // Selectors
@@ -52,7 +53,7 @@ class CollectorRow extends PureComponent<
   Props & RouteComponentProps<{orgID: string}>
 > {
   public render() {
-    const {collector, org} = this.props
+    const {collector} = this.props
 
     return (
       <ResourceCard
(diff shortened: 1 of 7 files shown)


src/telegrafs/components/CollectorCard.tsx
```diff
@@ -76,8 +77,9 @@ class CollectorRow extends PureComponent<
         />
         <ResourceCard.Meta>
           <Link
-            to={`/orgs/${org.id}/load-data/telegrafs/${collector.id}/instructions`}
+            to=""
```
TCL735, line 80: Does this do anything unintended or is it harmless to have a blank string for the `to` prop?

blegesse-git: I looked into this and didn't find anything that discourages passing an empty string to the `to` prop. The overlay opens as expected when the link is clicked with no errors logged in the console. An alternative is to use a button and style it as a link.

hoorayimhelping: @blegesse-git I think the proper move is to remove this `Link` component and replace it with something that doesn't have an `href` (`<Link to="http://example.com">` gets converted to something like `<a href="http://example.com">`), maybe like you suggested with a button. Previously, the behavior of this `Link` relied on

```ts
<Route
  path={`${telegrafsPath}/:id/instructions`}
  component={TelegrafInstructionsOverlay}
/>
```

existing, and rendering an overlay when the url matched. Since this PR triggers the overlay via redux, I think leaving this `Link` might have strange behavior

blegesse-git: okay no problem! It's probably better to take the safe approach. I'll replace it with a button and style it accordingly.

TCL735 (approved):

hoorayimhelping (approved):

Issue #5645, opened 2022-09-06
Styling/Clockface - When the app's width is below 800px, content disappears


hoorayimhelping:
When the width of the app wrapper is greater than 800px, the content displays fine. When the content breaks below 800px, the flex direction (among other things) changes which causes the content to disappear

### Above 800px
![Image](https://user-images.githubusercontent.com/146112/188654800-0ded4072-26f4-416e-b2b7-bc528c8a058d.png)

### Below 800px
![Image](https://user-images.githubusercontent.com/146112/188654799-6cbd67f1-b5ac-45d3-8f61-c6df0f7f86c8.png)

PR #5651, opened 2022-09-06
fix: page content now visible on mobile devices

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/5645

App works on mobile now as well. 

Clockface was behaving weird with the `div` in the App.tsx, so I changed it to a Page wrapper and everything seems to work again. 

Video: 

https://user-images.githubusercontent.com/18511823/188725099-32838f1e-88d2-4679-bb33-d2849aaf8315.mov

### Checklist


wdoconnell:
@ChitlangeSahas Could you please test with a few of the pages involving multi-org content before we commit? I suspect the multi-org adjustment styles should be okay since it's only targeting children of the element with the class `multiOrg`, but could you double-check before we do? Thanks!

hoorayimhelping (approved):

wdoconnell (changes requested):
Let's hold up for a moment, as this is impacting the multi-org style fixes. Posting a screenshot in a moment.

wdoconnell:
Right now, the way we fix styles when the multiOrg header is on is to override clockface and reduce the total Page size by the current size of the global header `(calc 100% - gh.$globalheader-height)`. Right now, the fixes presume we don't have a 'page within a page', so a more general css selector like this is sufficient:

MultiOrgOverrideStyles.css
```
.multi-org .cf-page-contents > .ScrollbarsCustom-Wrapper {
  height: calc(100% - gh.$globalheader-height);
}
```

If we have a `<Page>` element outside of the existing `<Page>` element, we need to increase the selector specificity for the multi-org fixes, otherwise I think we're limiting the height 2x - first, by adjusting the window size for the first-level Page, then again for the second-level Page. For this to work with multi-Org I think we need to do something more like this:

MultiOrgOverrideStyles.css
```
.multi-org > [more specific selectors, so that we only adjust the second-level Page] > .cf-page-contents > .ScrollbarsCustom-Wrapper {
  height: calc(100% - gh.$globalheader-height);
}
```


Before changing the `div` to `Page`
--
Screen Shot 2022-09-06 at 4 00 57 PM

If we change the `div` to `Page` (bottom part cut off)
--
Screen Shot 2022-09-06 at 4 02 17 PM

wdoconnell:
Re: interaction with multi-org styles, based on my testing it looks like we would still need the adjustment to page headers and notebooks. BUT it looks like we no longer need the custom adjustment to Page DapperScrollbars, Data Explorer, and Tasks if everything is wrapped in `<Page>`

hoorayimhelping (approved):

wdoconnell (approved):

ALuckette:
nice work guys, good call making sure this works with multi-org.

PR #5659, opened 2022-09-06
test: fix undefined variable in annotation test setup

TCL735:
Just housekeeping to clean up something i noticed in my code editor.

While there have not been any recent issues with the annotation e2e tests, i noticed that the function call `cy.createBucket(orgID, name, 'devbucket')` is actually referencing an undefined variable called `name`. A quick search in the file will confirm this.

This looks like the code was originally copied over without the `name` variable being destructured from an outer function call. We fix this by doing that now.


hoorayimhelping (approved):
good find. it looks like `undefined` is a valid org name when creating a bucket. very odd

PR #5667, opened 2022-09-07
build(lsp): upgrade flux-lsp to 0.8.34

wiedld:
Bug patch. Removed off-by-1 error in the schema composition applyEdit messages.
ASAP patch, before internal release.


wiedld:
> I'm curious why the urgency and what this dependency bump brings?

Good point. Description updated.

PR #5673, opened 2022-09-07
feat: add multiselect for measurements

### LINKED ISSUE #5648: Data Explorer - Multi-Measurement Support
## User Story

<p></p>

As a user, I want to be able to search fields and tags within all measurements so that I can find things even if I don't know what measurement they are in and I want to be able to pick multiple measurements so that I can find things in more than 1 individual measurement.

<p></p>

## Acceptance Criteria

<p></p>

- A new item will be available at the bottom of the measurement list labeled "select all" (at the bottom so that it doesn't just get checked out of ease)

- When select all is chosen, then no measurement filter will be added to the flux script (it will be like it worked previously when you could select a field or a tag first).

- A user will be able to select multiple measurements, not just one.

- When choosing select all, all other measurement boxes will be grayed out and then cannot be selected unless the user unselects the select all button

- When a user has chosen an individual measurement(s) and then chooses "select all" - the previous selections will be lost

<p></p>

## Notes

<p></p>

This is to answer the backlash of feedback from customers upset when we put forced measurement selection in place, and will buy us more time to add multi-measurement selection to the new script editor.

<p></p>

drdelambre:
Closes #5648

Peek 2022-09-07 10-41

this adds an `_all` tag value constant for the frontend specific query builder state so that people can select all measurements if they want. the underscore ensures there's no collision with user data and adding a special case ensures that we dont end up in the "all access token" problem where, by saving all measurements at point of creation, leads to a condition where adding a measurement to your data means your meta data disappears and your query fails.

behind the `measurementMultiselect` flag for product and waiting on design sign off.


mavarius:
@drdelambre Looks good so far but the behavior is really that of a "select none" rather than a "select all" which leads to some ambiguity. Could we have it instead be a checkbox for "All Measurements" which keeps the same behavior of disabling the others?

**Clicking "All Measurements" sets all items to unchecked and appears disabled**
image

**Clicking it again returns to normal selection behavior**
image

**Multi-select enabled**
image

src/timeMachine/components/TagSelector.tsx
```diff
@@ -232,13 +266,38 @@ class TagSelector extends PureComponent<Props> {
...
+        multiSelect={
+          !this.props.isInCheckOverlay ||
+          (this.isCompliant && isFlagEnabled('measurementMultiselect'))
+        }
+        disabled={selectedAll}
+      >
+        {this.isCompliant && isFlagEnabled('measurementMultiselect') && (
+          <List.Item
+            className="selector-list--item select-all"
+            testID="selector-list --select-all"
```
appletreeisyellow, line 288:
```suggestion
            testID="selector-list--item select-all"
```
Is it missing `item`?

src/timeMachine/components/TagSelector.tsx
```diff
@@ -350,6 +409,9 @@ const mstp = (state: AppState, ownProps: OwnProps) => {
     values = [...ADDITIONAL_GROUP_BY_COLUMNS, ...tags.map(tag => tag.key)]
   }
   const isInCheckOverlay = getIsInCheckOverlay(state)
+  const isLast = ownProps.index === tags.length - 1
+  const shouldClearLast =
+    ownProps.index === tags.length - 2 && !tags[tags.length - 1].values.length
```
appletreeisyellow, line 414: Just wondering, what is the purpose of `shouldClearLast`?

src/timeMachine/components/TagSelector.tsx
drdelambre: oh no i wanted to keep this separate from the user generated list items for tests that are like "push 10 items in, expect there to be 10 selector-list--item test ids"

src/timeMachine/components/TagSelector.tsx
drdelambre: there's this weird behavior where if you deselect the last selected list item in a card, and there's another card after the current one, and the next card doesn't have any selected items in it, we should delete that extra card.

appletreeisyellow (approved):

src/timeMachine/components/TagSelector.tsx
wiedld: I believe you mean this, which is a great catch....but we still have it occurring. Since it's an unrelated, found bug -- can we file an issue please?


https://user-images.githubusercontent.com/10232835/189423421-2a1f6ed6-ec47-4d2f-b7fa-cd23be70160a.mov

wiedld (approved):
LGTM.

Could we just make sure we have a issue filed for that extra card bug?

drdelambre:
there's also another bug i had to try REAL hard not to address in this PR where deselecting the last item of the last card causes an error to be thrown. but it was preexisting and it has been unclear to me how much investment product wants to make in this component. iirc product is gone on monday, but lets get a call from them on standup tuesday on the value add for these low hanging fruit.

src/timeMachine/components/TagSelector.tsx
drdelambre: def

PR #5688, opened 2022-09-07
test: add new test for adpative zoom

### LINKED ISSUE #5492: Query Zoom (GA) - e2e test
Add integration testing for the query zoom (adaptive zoom) feature.

- [ ] Test the behavior with the flag turned on by zooming in on a line graph and ensuring a request made for a query is made.

- [ ] Test the behavior with the flag turned off by zooming in on a line graph and ensuring no request for a query is made.

TCL735:
Closes #5492 

Test for a query being made when zooming on with the feature on.

With the feature off, we don't need to test, mainly due to the difficulty of proving that an api call does not happen in Cypress rather than happening. There is no clean way to do this without arbitrary wait times.

### DIFF
--- a/cypress/e2e/shared/adaptiveZoom.test.ts
+++ b/cypress/e2e/shared/adaptiveZoom.test.ts
@@ -1,0 +1,105 @@
+import {Organization} from '../../../src/types'
+import {points} from '../../support/commands'
+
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):

TCL735:
Any additional tests will be added in follow-up PRs.

Issue #5523, opened 2022-08-24
Resolve pushing into frozen array error in src/variables/selectors/index.tsx


wdoconnell:
## About the bug

`variables/selectors/index.tsx` can throw an error related to trying to push into a frozen array intended to contain the currently selected variable (`vari.selected`). We're seeing this issue specifically in monitor301.

The variables (`vari`) object is taken directly from redux state, which means that the variables object is frozen via `immer`. The variables object (`vari`) is then cloned, but using the spread operator, not a recursive deep cloning method. As a result, any attempts to push into arrays that are properties of the originally frozen object fail.

A fix would be to deep clone `vari` instead of cloning it using the spread operator, so that arrays that are properties of the variables object are extensible. But also need to investigate root cause to see why this block of code is even hit. 

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. Log in to `monitor301`
2. Browse to dashboard that uses variables

**Expected behavior:**
Dashboard loads

**Actual behavior:**
Sometimes throws error `can't define array index property past the end of an array with non-writable length` or `Cannot add property 0, object is not extensible`.

Error sometimes fixed by clearing the cache.

**Visual Proof:**
![frozen array](https://user-images.githubusercontent.com/91283923/186459906-0376704a-7442-4499-ad7d-30338ed2be65.png)

cannonpalms:
When I experienced this last night, I saw that the exception was being thrown when a variable named `node_id` was being processed. Not sure why it wasn't an issue for other variables, but I'm guessing it has something to do with the predicate that encapsulates the offending line of code. 

There is additional context on the exact `state` that was resulting in the error in [this slack thread](https://influxdata.slack.com/archives/CLHATC50E/p1661299308854909?thread_ts=1660833756.695399&cid=CLHATC50E)

PR #5689, opened 2022-09-07
fix: add specific error reporting around default-selected-variable code

wdoconnell:
Connects #5523 

I'm unable to reproduce the `object not extensible` error caused by the app trying to push into a frozen array (`vari.selected`) when there is no currently selected variable, either in remocal or in the environment in which we actually see the error (New Tools). 

This PR adds specific error reporting around this code block, so that we can actually see which variable is causing the issue, and what the state of the app is at that time. The current top-level error reporting doesn't provide that level of detail.

### Checklist

### DIFF
--- a/src/variables/selectors/index.tsx
+++ b/src/variables/selectors/index.tsx
@@ -218,12 +218,13 @@ export const getVariable = (state: AppState, variableID: string): Variable => {
       reportErrorThroughHoneyBadger(err, {
         name: 'Failed to set selected variable to default, zero-indexed value',
         context: {
+          identityState: state.identity,
+          resourceState: state.resources,
           variable: vari,
           variableType: vari.arguments.type,
           normalizedVariable: vals,
           variableLenth: vari.selected.length,
           normalizedVariableLength: vals.length,
-          appState: state,
         },
       })
     }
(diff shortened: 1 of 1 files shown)


src/variables/selectors/index.tsx
```diff
@@ -210,7 +211,22 @@ export const getVariable = (state: AppState, variableID: string): Variable => {
...
+      // Temporary measure to resolve errors relating to pushing into non-extensible object.
+      reportErrorThroughHoneyBadger(err, {
+        name: 'Failed to set selected variable to default, zero-indexed value',
+        context: {
+          variable: vari,
+          variableType: vari.arguments.type,
+          normalizedVariable: vals,
+          variableLenth: vari.selected.length,
+          normalizedVariableLength: vals.length,
+          appState: state,
```
hoorayimhelping, line 226: I think maybe instead of sending the whole app state, we should start with just the variables state.

wdoconnell: Good point. What if we send something slightly broader than `variables`, just to make sure we have all of the relevant information?

1. `state.resources` - One level higher than `variables` in state, so that we can also see `buckets` `dashboards` and `cells`
2. `state.identity` - so that we know which user @ influxdata hit the error, so we can ask for any context.

hoorayimhelping (approved):

hoorayimhelping (approved):

PR #5701, opened 2022-09-08
fix(dataLoaders): add styling to telegraf conf overlay

### LINKED ISSUE #5700: Telegraf: fix the styling of telegraf config overlays
There are 3 places in the UI where this telegraf config overlay shows; 

- Inside Telegraf page when creating a new telegraf configuration
- Inside Source page when adding a plugin to an existing configuration 
- Inside Source page when adding a plugin to a new configuration

The padding of these specific overlay (see screenshots) need some tweaking. Add styling to the overlay body to correctly position the content of the page.

Image

Image

Image

blegesse-git:
Closes #5700 

Adds styling to three telegraf overlays found 
- Inside Telegraf page when creating a new telegraf configuration
- Inside Source page when adding a plugin to an existing configuration
- Inside Source page when adding a plugin to a new configuration

Depending on where user is at in the configuration step (`currentStepIndex` `substepIndex`), pr adds padding to the overlay body. 

Before: 
https://user-images.githubusercontent.com/66275100/189175626-8664e8b2-999b-4767-bb7a-51c19710fd1f.mov

After: 
https://user-images.githubusercontent.com/66275100/189175604-e9bef583-c130-4dcd-a647-952b5e4dac41.mov

### Checklist

### DIFF
--- a/src/dataLoaders/components/DataLoadersOverlay.scss
+++ b/src/dataLoaders/components/DataLoadersOverlay.scss
@@ -43,6 +43,12 @@
   padding: $ix-marg-d;
 }
 
+.configuration-overlay--body {
+  padding: 45px !important;
+  padding-top: 0 !important;
+  padding-bottom: 0 !important;
+}
+
 /*
   Telegraf Instructions Overlay
   ------------------------------------------------------------------------------
--- a/src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
+++ b/src/dataLoaders/components/collectorsWizard/TelegrafUIRefreshWizard.tsx
@@ -45,6 +45,7 @@ import {isSystemBucket} from 'src/buckets/constants'
 
 // Constants
 import {getBucketOverlayWidth} from 'src/buckets/constants'
+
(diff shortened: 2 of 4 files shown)


ChitlangeSahas (approved):
Looks good to me, I am just curious why the Overlay isn't handling this by default, we shouldn't need to override those styles. 

The Overlay structure looks correct to me, like defined [here](https://influxdata.github.io/clockface/?path=/story/components-overlays-family--overlay) but for some reason it's not applying the overlay styles by default. weird.

blegesse-git:
> Looks good to me, I am just curious why the Overlay isn't handling this by default, we shouldn't need to override those styles.
> 
> The Overlay structure looks correct to me, like defined [here](https://influxdata.github.io/clockface/?path=/story/components-overlays-family--overlay) but for some reason it's not applying the overlay styles by default. weird.

I looked through the entire folder trying to figure that out. The overlay this pr touches doesn't have any styles applied to it, which means we shouldn't have to provide additional styling for the content to display properly. But that's not what's happening. My guess is the `DataLoadersOverlay.scss` is interfering with the `Overlay.Body` style. Because some of the telegraf configuration steps are rendered using this scss, we probably needed to add custom style adjustments to the other overlays.

PR #5716, opened 2022-09-09
fix: add a loading check prior to setting the buckets

### LINKED ISSUE #5702: Error when attempting to open a notebook
## About the bug

**Steps to reproduce:**

1. Go to Buckets and select any bucket OR
2. Go to Notebook and select New Notebook or one from the list
3. Page attempts to load then displays the error message "An InfluxDB error has occurred".

**Expected behavior:**
I expect the notebook page to load.

**Actual behavior:**
Screenshot 2022-09-08 094202
Screenshot 2022-09-08 094448

**Visual Proof:**
Attached

## About your environment

**Environment info:**
Standard cloud environment, in trial phase.

**Config:**

**Logs:**

**Performance:**

```sh
iostat -xd 1 30 > iostat.txt
```

asalem1:
Closes #5702 

This PR addresses an issue that was causing Notebooks with a `queryBuilder`, or `metricSelector` to crash when the panel is initialized. More specifically, something with the Multi-org feature being enabled was causing an infinite callstack, the cause of which has still TBD. Adding this check seems to fix the crashing bug. I've verified that this issue does not break the flux query builder (the other usage of the BucketProvider) with the feature flags for multi-org on AND off. I have also verified that the check does not break notebooks with the feature flags on AND off

### Checklist

### DIFF
--- a/src/shared/contexts/buckets.tsx
+++ b/src/shared/contexts/buckets.tsx
@@ -91,16 +91,18 @@ export const BucketProvider: FC<Props> = ({
 
   // keep the redux store in sync
   useEffect(() => {
-    dispatch(
-      setBuckets(
-        RemoteDataState.Done,
-        normalize<Bucket, BucketEntities, string[]>(
-          buckets.filter(b => b.type !== 'sample'),
-          arrayOfBuckets
+    if (loading === RemoteDataState.Done) {
+      dispatch(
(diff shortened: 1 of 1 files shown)


hoorayimhelping (approved):
shakeee

drdelambre (approved):
k

PR #5721, opened 2022-09-09
feat: display window period on cells

### LINKED ISSUE #4989: Query Zoom (GA) - Display Window Period on Graph per Cell
**ACs**
- If the period is the same per yield, show it at the cell level. 
- @taramk has a Figma for this scenarios (Show upon hover)

**Design**
Image

TCL735:
Closes #4989


taramk:
@TCL735 - can we make it the same text size as the cell title if it isn't already? it looks slightly larger but that might just be because it isn't bold. otherwise looks good, thank you!

hoorayimhelping (approved):

ChitlangeSahas (approved):

PR #5723, opened 2022-09-12
fix: centralize state and reduce render cycles to derisk component

### LINKED ISSUE #5722: Investigate notebooks failure to resolve with multi-org enabled
Follow up issue to https://github.com/influxdata/ui/issues/5702 to determine the root cause of notebooks error when multi-org is enabled.

Steps to reproduce:

1. Open post-production in an incognito window
2. Visit notebooks, and create a new notebook
3. Watch the crash

Steps to verify that multi-org is related:

1. Open post-production in an incognito window
2. turn the multiOrg flag OFF
3. Visit notebooks and create a new notebook
4. No crash

@asalem1's says:

> The only weird thing that I’ve isolated is that the `SET_QUARTZ_ORGANIZATIONS_STATUS` action gets dispatched when entering a notebook, which was unexpected since I’m not seeing that action fired off when entering other parts of the app, but again, not sure how that’s related.

asalem1:
Closes #5722 

The purpose of this PR is to derisk the `GlobalHeaderDropdown` component by consolidating the state to the parent component, thereby preventing drift between state and the potential callstack exceeded issues moving forward. 

### Investigative Conclusions

Prior to the `multiOrg` and `quartzIdentity` flags being turned on, Notebooks that use the `context/bucket.tsx` file were loading correctly. While this accurate behavior did have some problems, those problems only came to light when the flags were turned on and an infinite render cycle was produced. The cause of that infinite render cycle has been pinpointed to two things:

1. The number of times the `setBuckets` was dispatched to update the reducer state
2. The number of times the `GlobalHeaderDropdown` componentDidUpdate was called.

It appears as though these two state-setting operations were feeding back to one another, so that whenever the `setBuckets` was called, the `GlobalHeaderDropdown` would be updated. That update would then trigger the `context/bucket` to be updated, and so on and so on. 

### Solution

While the [previous PR](https://github.com/influxdata/ui/pull/5716) worked to tighten up when the `setBuckets` call should be dispatched (thereby limiting the render cycles), this PR aims to provide guardrails around the render cycle of the `GlobalHeaderDropdown`. More specifically, it seems as though the main purpose of the `selectedItem` state setting was simply a means of setting a default for the state based on an initial selection. However, in doing so, the potential for state drift between two sources of truth (the parent props and the component state) were manually being managed by a lifecycle method (componentDidUpdate), which would be triggered any time the component updated. What this PR does is centralize the state to the parent as the single source of truth, thereby removing the lifecycle method that was previously being used to centralize updates.

### Checklist

### DIFF
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/index.tsx
@@ -63,7 +63,6 @@ export interface Props extends StandardFunctionProps {
 
 type State = {
   showTypeAheadMenu: boolean
-  selectedItem?: TypeAheadMenuItem
 }
 
 export class GlobalHeaderDropdown extends React.Component<Props, State> {
@@ -71,18 +70,8 @@ export class GlobalHeaderDropdown extends React.Component<Props, State> {
     super(props)
     this.state = {
       showTypeAheadMenu: this.props.onlyRenderSubmenu ?? false,
-      selectedItem: this.props.typeAheadSelectedOption || null,
     }
   }
-
-  componentDidUpdate() {
-    if (this.props.typeAheadSelectedOption !== this.state.selectedItem) {
-      this.setState({
-        selectedItem: this.props.typeAheadSelectedOption,
-      })
-    }
-  }
-
   private dropdownButton = (
     active: boolean,
     onClick: (e: MouseEvent<HTMLElement>) => void
@@ -92,8 +81,8 @@ export class GlobalHeaderDropdown extends React.Component<Props, State> {
       defaultTestID,
       dropdownButtonIcon,
       dropdownButtonSize,
+      typeAheadSelectedOption,
     } = this.props
-    const {selectedItem} = this.state
     return (
       <Dropdown.Button
         active={active}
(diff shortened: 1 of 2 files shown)


wdoconnell (approved):
Looks good, and thanks again for handling. Here are two videos showing that multi-org functionality still works, and that we see no re-render issue on other pages.

https://user-images.githubusercontent.com/91283923/189732643-152910f9-57c6-4fcb-9f61-0888b37eb9ac.mov

https://user-images.githubusercontent.com/91283923/189732663-1413ce10-dcb5-4894-8ec1-88d273d39239.mov

wdoconnell:
> > These edits look good, and thanks for the great work tracing this. I'd like to do some additional local testing before we merge this because:
> > 
> > 1. I think we should double check the notebooks buckets context before we flip the flag back on, especially if that context is used on other pages. Checking the `loading` status in Notebooks is a good guard, but that fix leaves me wondering why that useEffect hook is running so much - `buckets` should not be changing on every render.
> > 2. The flag is still on in prod for three organizations, and everywhere in staging and tools, so I'd like to make sure the removal of this state doesn't detrimentally affect that functionality.
> 
> Totally. So:
> 
> 1. The `buckets` context issue was resolved in the PR linked in the description, but it can be found [here](https://github.com/influxdata/ui/pull/5716) as well.
> 2. Not sure how I'd go about testing out how this change affects customers with the feature flag currently on, unless we want to feature flag a new component and switch between the two? Would love to hear suggestions @wdoconnell

On 2: I've tested the multi-org features, and also pages generally in the app to ensure this doesn't cause any unanticipated issue (which it shouldn't - we're just taking out an unnecessary `setState`).

On 1: It could be that I'm missing context, but the `buckets` context issue was resolved by adding a check where, we don't dispatch the `setBuckets` action until `loading === RemoteDataState.Done`. That's great and seems to have resolved the immediate issue, but I don't fully understand why that `useEffect` hook was running more than once unless `buckets` is changing. But maybe that's a question for another time.

PR #5740, opened 2022-09-12
chore: toggle quartzIdentity and multiOrg on in cloud e2e tests

### LINKED ISSUE #5737: Set multiorg feature flag to on in all Cloud E2E tests
Since multi-org will be left on for all users, we should have the flag enabled in all Cloud Cypress tests.

This includes ensuring that multi-org is toggled on for the Cypress PDTs.

wdoconnell:
Closes #5737

This PR toggles the `quartzIdentity` and `multiOrg` flags on for all cloud2 e2e tests, as these flags are going to be on in prod for all users. Where feasible, I've refactored to make sure any flag adjustments happen in a single `beforeEach` rather than in multiple individual tests (`it`). There are a few places this isn't possible - e.g., where cypress needs to toggle on one flag for a specific block to test a new feature.

### Checklist

- [X] Feature flagged, if applicable - tests toggle on `quartzIdentity` and `multiOrg`

### DIFF
--- a/cypress/e2e/cloud/about.test.ts
+++ b/cypress/e2e/cloud/about.test.ts
@@ -5,13 +5,18 @@ describe.skip('About Page for free users with only 1 user', () => {
   beforeEach(() =>
     cy.flush().then(() =>
       cy.signin().then(() => {
-        cy.get('@org').then(({id}: Organization) => {
-          cy.quartzProvision({
-            accountType: 'free',
-            hasUsers: false,
-          }).then(() => {
-            cy.visit(`/orgs/${id}/org-settings`)
-            cy.getByTestID('about-page--header').should('be.visible')
+        cy.setFeatureFlags({
+          quartzIdentity: true,
+          multiOrg: true,
+        }).then(() => {
+          cy.get('@org').then(({id}: Organization) => {
+            cy.quartzProvision({
+              accountType: 'free',
+              hasUsers: false,
+            }).then(() => {
+              cy.visit(`/orgs/${id}/org-settings`)
+              cy.getByTestID('about-page--header').should('be.visible')
+            })
           })
         })
       })
@@ -49,13 +54,18 @@ describe('About Page for free users with multiple users', () => {
   beforeEach(() =>
     cy.flush().then(() =>
       cy.signin().then(() => {
-        cy.get('@org').then(({id}: Organization) => {
-          cy.quartzProvision({
-            accountType: 'free',
-            hasUsers: true,
-          }).then(() => {
-            cy.visit(`/orgs/${id}/org-settings`)
-            cy.getByTestID('about-page--header').should('be.visible')
+        cy.setFeatureFlags({
+          quartzIdentity: true,
+          multiOrg: true,
+        }).then(() => {
+          cy.get('@org').then(({id}: Organization) => {
+            cy.quartzProvision({
+              accountType: 'free',
+              hasUsers: true,
+            }).then(() => {
+              cy.visit(`/orgs/${id}/org-settings`)
+              cy.getByTestID('about-page--header').should('be.visible')
+            })
           })
         })
       })
@@ -80,12 +90,17 @@ describe('About Page for PAYG users', () => {
   beforeEach(() =>
     cy.flush().then(() =>
       cy.signin().then(() => {
-        cy.get('@org').then(({id}: Organization) => {
-          cy.quartzProvision({
-            accountType: 'pay_as_you_go',
-          }).then(() => {
-            cy.visit(`/orgs/${id}/org-settings`)
-            cy.getByTestID('about-page--header').should('be.visible')
+        cy.setFeatureFlags({
+          quartzIdentity: true,
+          multiOrg: true,
+        }).then(() => {
+          cy.get('@org').then(({id}: Organization) => {
+            cy.quartzProvision({
+              accountType: 'pay_as_you_go',
+            }).then(() => {
+              cy.visit(`/orgs/${id}/org-settings`)
+              cy.getByTestID('about-page--header').should('be.visible')
+            })
           })
         })
       })
--- a/cypress/e2e/cloud/billing.test.ts
+++ b/cypress/e2e/cloud/billing.test.ts
@@ -2,53 +2,53 @@ import {Organization} from '../../../src/types'
 
 describe('Billing Page Free Users', () => {
   beforeEach(() =>
-    cy.flush().then(() =>
+    cy.flush().then(() => {
       cy.signin().then(() => {
-        cy.setFeatureFlags({
-          quartzIdentity: true,
-          multiOrg: true,
-        }).then(() => {
-          cy.get('@org').then(({id}: Organization) => {
-            cy.quartzProvision({
-              accountType: 'free',
-            }).then(() => {
-              cy.visit(`/orgs/${id}/billing`)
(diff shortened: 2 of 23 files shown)


wdoconnell:
This PR is best reviewed with `?w=1`.

cypress/e2e/cloud/about.test.ts
```diff
@@ -80,12 +90,17 @@ describe('About Page for PAYG users', () => {
   beforeEach(() =>
     cy.flush().then(() =>
       cy.signin().then(() => {
-        cy.get('@org').then(({id}: Organization) => {
-          cy.quartzProvision({
-            accountType: 'pay_as_you_go',
-          }).then(() => {
-            cy.visit(`/orgs/${id}/org-settings`)
-            cy.getByTestID('about-page--header').should('be.visible')
+        cy.setFeatureFlags({
```
wdoconnell, line 93: Need this separately in each `beforeEach` because they are part of separate `describe` blocks.

cypress/e2e/cloud/billing.test.ts
```diff
@@ -52,154 +57,172 @@ describe('Billing Page PAYG Users', () => {
...
+      cy.getByTestID('billing-contact--header').contains('Contact Information')
+      cy.getByTestID('form-label--First Name').contains('First Name')
+      cy.getByTestID('contact-info--Test').contains('Test')
+      cy.getByTestID('form-label--Last Name').contains('Last Name')
+      cy.getByTestID('contact-info--PAYG').contains('PAYG')
+      cy.getByTestID('form-label--Company Name').contains('Company Name')
+      cy.getByTestID('contact-info--InfluxData').contains('InfluxData')
+      cy.getByTestID('form-label--Country').contains('Country')
+      cy.getByTestID('contact-info--USA').contains('USA')
+      cy.getByTestID('form-label--Physical Address').contains(
```
wdoconnell, line 157: A few of these linter adjustments cropped up during saving.

cypress/e2e/cloud/about.test.ts
wdoconnell: Note that I've had to adjust the position of where we call `setFeatureFlags`. Certain pages we're navigating to can be 'flaky' with the flag reset, so in each case I've tested for the best location to reset flags (which isn't always identical between tests).

cypress/e2e/cloud/helpBar.test.ts
```diff
@@ -60,10 +64,16 @@ describe('Help bar support for PAYG users', () => {
     cy.getByTestID('payg-support-overlay-header').should('exist')
 
     cy.getByTestID('contact-support-subject-input').clear().type(subject)
-    cy.getByTestID('dropdown--button').click()
-    cy.getByTitle('1 - Critical').click()
-
-    cy.getByTestID('support-description--textarea').clear().type(description)
-    cy.getByTestID('payg-contact-support--submit').should('not.be.disabled')
+    cy.getByTestID('severity-level-dropdown')
+      .within(() => {
```
wdoconnell, line 68: This is necessary because, with multi-org on, there's more than one dropdown on the page, so clicking on something with a testID of 'dropdown--button' isn't sufficiently specific anymore.

cypress/e2e/cloud/usage.test.ts
```diff
@@ -26,7 +27,7 @@ describe('Usage Page Free User No Data', () => {
 
   it('should display the usage page common features', () => {
     // Display the upgrade button when the user is a free user
-    cy.get('.cf-page-header--fluid').children().should('have.length', 2)
+    cy.get('.cf-page-header--fluid').children().should('have.length', 1)
```
wdoconnell, line 30: This change is necessary because the location of the upgrade button has moved. The fluid page-header will always only have one item present in that row.

cypress/e2e/cloud/usage.test.ts
```diff
@@ -108,16 +109,14 @@ describe('Usage Page PAYG With Data', () => {
...
+            cy.setFeatureFlags({quartzIdentity: true, multiOrg: true})
           })
         })
       })
     )
   })
 
   it('should display the usage page with data for a PAYG user', () => {
-    // The implication here is that there is no Upgrade Now button
-    cy.get('.cf-page-header--fluid').children().should('have.length', 1)
```
wdoconnell, line 119: This check is now useless. It's true that this page-header will have length one, but that's because the upgrade now button is moved into the header.

The better way to test this is by checking that the upgrade button _actually_ isn't present, rather than implying it's not present based on the size of the header.

cypress/e2e/cloud/userAccounts.test.ts
```diff
@@ -4,158 +4,87 @@ const doSetup = (cy, numAccounts: number) => {
   cy.flush().then(() => {
     cy.signin().then(() => {
       cy.get('@org').then(({id}: Organization) => {
+        cy.visit(`/orgs/${id}/accounts/settings`)
```
wdoconnell, line 7: UserAccounts required significant refactoring.

First, there's no longer any point in checking for the 'switch-account' button or doing tests around that. The switch-account functionality lives in globalheader now, so those tests are moot (and will fail when multiOrg is on, since the account page doesn't have a switch-account button in that case).

Second, userAccounts has been causing some pipeline issues for us because the tests do not have a clean interaction with our test environment. They modify the account state in quartz-mock and don't reset it, so when multiple tests are being run in parallel (or another test relying on account information happens to be run afterwards), it contaminates the environment, causing test failures down the line.

The better solution here is to dispense with using 'quartzProvision,' and simply use `cy.intercept` to mock the quartz get-account and rename-account APIs.

cypress/e2e/cloud/userProfile.test.ts
```diff
@@ -1,28 +1,3 @@
-// Utility Functions
```
wdoconnell, line 1: This change just moves this block lower in the file to make reviewing the variables it relies on more intuitive.

cypress/e2e/cloud/explorer.test.ts
```diff
@@ -33,61 +34,75 @@ describe('DataExplorer', () => {
...
-      })
-      cy.getByTestID('flux-toolbar-search--input').type('{backspace}')
+      cy.setFeatureFlags({
+        fluxDynamicDocs: true,
+        quartzIdentity: true,
+        multiOrg: true,
+      }).then(() => {
+        cy.get('.view-line').should('be.visible')
+
+        cy.getByTestID('flux-toolbar-search--input')
```
wdoconnell, line 44: Again, linter adjustment.

cypress/support/commands.ts
```diff
@@ -1048,6 +1048,14 @@ export const setFeatureFlags = (flags: FlagMap): Cypress.Chainable => {
   })
 }
 
+export const setFeatureFlagsNoNav = (flags: FlagMap): Cypress.Chainable => {
```
wdoconnell, line 1051: The reason `setFeatureFlags` wasn't working on the checkout page is that `TreeNav` doesn't display there; so we were failing the 'wait for tree nav' check that is implicit in `setFeatureFlags`. 

This adds a separate cypress command we can use to override feature flags on pages that don't have `TreeNav`.

hoorayimhelping (approved):

cypress/e2e/cloud/billing.test.ts
wdoconnell: Did not explicitly run yarn prettier, but prettier runs on a hard save in my editor

PR #5741, opened 2022-09-12
feat(5340): use default editorText in script editor

### LINKED ISSUE #5340: Schema Composition: explanation/instructions to user?
Instructions via comments, in flux script?
* On new session, make placeholder query:
    * Schema builder on left.
    * Function injection on right.
* On first session composition:
    * Does some intro into what the composition does.
    * Explains the sync button.
    * States when we’ll turn off sync for you. (e.g. once you type in the composition block.)

wiedld:
Closes #5340 

## DONE:
* in new QX script editor:
  * Use default text for:
    * default session query string
    * clear() methods to reset the session
    * default query string in script resource
  * Could have tried doing the minimum (change only the default script resource). But decided to make all reference the proper default, to avoid any current/future races.
  * Is not feature flagged since the description is accurate with, or without, the new features in the QX script editor.
* Default message does NOT appear in:
  * notebooks
  * old data explorer. 

## Video:

https://user-images.githubusercontent.com/10232835/189738763-27194a58-4a62-4537-8306-1ec5955bbeed.mov

### DIFF
--- a/src/dataExplorer/components/SaveAsScript.tsx
+++ b/src/dataExplorer/components/SaveAsScript.tsx
@@ -12,7 +12,10 @@ import {
 import {useHistory} from 'react-router-dom'
 import {QueryContext} from 'src/shared/contexts/query'
 import {ResultsContext} from 'src/dataExplorer/components/ResultsContext'
-import {PersistanceContext} from 'src/dataExplorer/context/persistance'
+import {
+  DEFAULT_EDITOR_TEXT,
+  PersistanceContext,
+} from 'src/dataExplorer/context/persistance'
 import {RemoteDataState} from 'src/types'
 import './SaveAsScript.scss'
 import {CLOUD} from 'src/shared/constants'
@@ -34,8 +37,14 @@ interface Props {
 const SaveAsScript: FC<Props> = ({onClose, type}) => {
   const dispatch = useDispatch()
   const history = useHistory()
-  const {hasChanged, resource, setResource, save} =
-    useContext(PersistanceContext)
+  const {
+    hasChanged,
+    resource,
+    setResource,
+    setQuery,
+    clearSchemaSelection,
+    save,
+  } = useContext(PersistanceContext)
   const {cancel} = useContext(QueryContext)
   const {setStatus, setResult} = useContext(ResultsContext)
   const org = useSelector(getOrg)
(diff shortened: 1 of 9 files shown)


src/dataExplorer/components/resources/types/script/index.ts
```diff
@@ -1,6 +1,7 @@
 import {ResourceType} from 'src/types/resources'
 import editor from './editor'
 import {CLOUD} from 'src/shared/constants'
+import {DEFAULT_EDITOR_TEXT} from 'src/dataExplorer/context/persistance'
```
drdelambre, line 4: The changes in this file should be the only change required to enact this change. The other places where you set the query then navigate to call this code

drdelambre (changes requested):
Consolidate the places where the query is set to reduce how many times it's getting set

wiedld:
> Consolidate the places where the query is set to reduce how many times it's getting set

Made and pushed changes, but now we get a missing init message:

https://user-images.githubusercontent.com/10232835/190283314-2bbd04dc-3df9-44e3-9579-bc70732fdafc.mp4

Going to add back the default message into the session store. even if it's not set again elsewhere.

src/dataExplorer/components/resources/types/script/index.ts
wiedld: If I only make the changes in the script resource then we have a missing message on first load of the page.

see that I demonstrate this with the commit (`test: demonstrate that we need the default query set on both the session, and in the script resource. In order to have on-load provide the default message.`), and then made a video of the behavior (refer to the PR convo).

src/dataExplorer/components/SaveAsScript.tsx
```diff
@@ -79,6 +79,7 @@ const SaveAsScript: FC<Props> = ({onClose, type}) => {
   const clear = useCallback(() => {
     cancel()
     setStatus(RemoteDataState.NotStarted)
+    clearSchemaSelection()
```
wiedld, line 82: This was missing, based removed on an earlier Save/Load PR.

On clear, we also need to clear out the session for the schema.
Otherwise the schema composition will try to re-create the editor text based on the session.

src/dataExplorer/components/resources/types/script/index.ts
drdelambre: makes sense

drdelambre (approved):
thanks for reducing that down

Issue #5591, opened 2022-08-30
Schema Composition: e2e tests for integration with other UI bits. (session store, Load/Save, page reload, etc)


wiedld:
After finding this bug:
https://github.com/influxdata/ui/pull/5586

Decided we needed tests asap. Specifically, for parts where the other UI parts hit the schema composition (e.g. the session, the resetting of the script, page load).

PR #5742, opened 2022-09-12
chore(5591): tests to ensure composition does not accidentally appear where it shouldn't

wiedld:
Part of e2e tests #5591 

## Done:
* composition block styling, should not exist in:
   * notebooks
   * old data explorer
   * new QX builder, when feature flag is off
* should only exist:
  * new QX builder, when feature flag is on

## Already Done (not added here):
* normal injection routes, outside of composition, are already tested. Did not need to add those tests.
* injection via schema composition

### DIFF
--- a/cypress/e2e/shared/editor.test.ts
+++ b/cypress/e2e/shared/editor.test.ts
@@ -50,6 +50,13 @@ describe('Editor+LSP communication', () => {
           cy.visit(`${orgs}/${id}`)
         })
       )
+      // Double check that the new schemaComposition flag does not interfere.
+      cy.setFeatureFlags({
+        schemaComposition: true,
+      })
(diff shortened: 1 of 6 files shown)


wiedld:
> Can we also set the feature flag to `true` for the `dataExplorer` and all other places where we might author a query? I don't anticipate things breaking, but it adds some confidence knowing that applying these changes to our cypress tests everywhere there's a flux editor should accurately reflect the reality of our application when the feature flag is turned on

Excellent point!
I've turned the flag on (`true`) for any test where it may come into play. This means:
1. the shared/editor tests, where we make sure the composition style does not appear in other editors.
2. in any other tests where we are actively using the editor injection:
    * explorer
    * variables
    * flows

PR #5769, opened 2022-09-15
feat: use quartz, not idpe, for user sessions

### LINKED ISSUE #5022: Multi Org Phase 1 - switch to using quartz in lieu of IDPE for auth once orphaned user/org backfill completes
In [UI #4896](https://github.com/influxdata/ui/pull/4896/) we are reverting to using `idpe/me` for periodic polling to check session validity, since there are still some cloud users that cannot hit quartz endpoints. 

Once [quartz #6362](https://github.com/influxdata/quartz/issues/6362) is complete, we should go return to using quartz endpoints for those users, so that we can complete the move away from `idpe/me` in the cloud.

**Proposal:**
Return to polling `quartz` endpoints for session validity in lieu of `idpe` endpoints.

**Current behavior:**
Use `idpe/me` for auth polling after login.

**Desired behavior:**
Use `quartz` for auth polling after login.

**Use case:**
Part of multi-org epic.

wdoconnell:
Closes #5022 

This PR changes login behavior so that, in Cloud, Quartz (not IDPE) is polled to check for the validity of the user's session. 

New code block is behind the flag `quartzSession` and will be tested in other environments before being enabled in prod.

Old Behavior
--
At login, and every 60 seconds . . . 

Make a `GET` request to `/api/v2/me` (IDPE /me).

New Behavior
--
At login, and every 60 seconds . . . 

OSS
1. No change.

CLOUD
1. `quartzIdentity` enabled? Make `GET` request to `/api/v2/quartz/identity`
2. Otherwise, make `GET` request to `/api/v2/quartz/me`

Notes
- The second option in cloud will be removed soon, as `/quartz/identity` is on in all environments and `/quartz/me` is deprecated, and is slated to be removed from the codebase shortly.
- Once the flag is removed we'll only need to make one call - `fetchIdentity` - as that function contains the logic for distinguishing between CLOUD and non-CLOUD environments.

### Checklist

- [X] Feature flagged, if applicable - `quartzSession`

### DIFF
--- a/src/Signin.tsx
+++ b/src/Signin.tsx
@@ -32,7 +32,7 @@ import {
 import {RemoteDataState} from 'src/types'
 
 // APIs
-import {fetchLegacyIdentity} from './identity/apis/auth'
+import {fetchIdentity, fetchLegacyIdentity} from './identity/apis/auth'
 
 interface State {
   loading: RemoteDataState
@@ -96,7 +96,12 @@ export class Signin extends PureComponent<Props, State> {
 
   private checkForLogin = async () => {
     try {
-      await fetchLegacyIdentity()
+      if (isFlagEnabled('quartzSession')) {
+        await fetchIdentity()
+      } else {
+        await fetchLegacyIdentity()
+      }
+
       this.setState({auth: true})
       const redirectIsSet = !!getFromLocalStorage('redirectTo')
       if (redirectIsSet) {
--- a/src/identity/apis/auth.ts
+++ b/src/identity/apis/auth.ts
@@ -118,7 +118,7 @@ export const fetchIdentity = async () => {
   if (!CLOUD) {
     return fetchLegacyIdentity()
   }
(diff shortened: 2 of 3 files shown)


src/Signin.tsx
```diff
@@ -96,7 +96,12 @@ export class Signin extends PureComponent<Props, State> {
 
   private checkForLogin = async () => {
     try {
-      await fetchLegacyIdentity()
+      if (isFlagEnabled('quartzSession')) {
```
hoorayimhelping, line 99: do we also want a `CLOUD &&` check here and in the other call?

wdoconnell: I thought about that and am open to it, but I ended up deciding against it because `fetchIdentity()` already includes a CLOUD check:
Screen Shot 2022-09-15 at 10 41 20 AM

I should remove that comment in `fetchIdentity` (RIP `uiUnificationFlag`)

hoorayimhelping (approved):

Issue #5029, opened 2022-07-12
Multi Org Phase 1 - Refactor userAccount to use API layer


wdoconnell:
The accounts page (UserAccount.tsx) should be refactored so that `getAccounts` and `patchAccount` from unityroutes are also using the API layer. Noticed this when working on #5023, but this should be a separate ticket since it would affect existing code that's already being used in production. 

**Expected behavior:**
Accounts page uses API layer for maintainability and consistent error handling.

**Actual behavior:**
New method is using API layer, old methods aren't.

PR #5773, opened 2022-09-15
chore: refactor useraccounts to use api layer

subirjolly:
Related: #5029

Refactor user accounts page (UserAccount.tsx) so that `getAccounts` and `patchAccount` from unityroutes are being used by the auth API layer.

### Checklist

### DIFF
--- a/src/accounts/context/userAccount.tsx
+++ b/src/accounts/context/userAccount.tsx
@@ -12,7 +12,6 @@ import {
   accountDefaultSettingSuccess,
   accountRenameError,
   accountRenameSuccess,
-  defaultAccountNotFoundError,
 } from 'src/shared/copy/notifications'
 
 // Utils
--- a/src/accounts/context/userAccount.tsx
+++ b/src/accounts/context/userAccount.tsx
@@ -84,14 +84,14 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
     try {
       const accounts = await getUserAccounts()
       setUserAccounts(accounts)
-      const defaultAcctArray = accounts.find(acct => acct.isDefault)
+      const defaultAcctArray = accounts.find(acct => acct.isDefault === true)
       if (defaultAcctArray) {
         const defaultId = defaultAcctArray[0].id
         setDefaultAccountId(defaultId)
       }
 
       // isActive: true is for the currently logged in/active account
-      const activeAcctArray = accounts.find(acct => acct.isActive)
+      const activeAcctArray = accounts.find(acct => acct.isActive === true)
       if (activeAcctArray) {
         const activeId = activeAcctArray[0].id
         setActiveAccountId(activeId)
@@ -127,7 +127,7 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
 
   const handleRenameActiveAccount = useCallback(
     async (accountId, newName) => {
-      const activeAccount = userAccounts.find(acct => acct.isActive)
+      const activeAccount = userAccounts.find(acct => acct.isActive === true)
       try {
         const accountData = await updateUserAccount(accountId, newName)
         event('multiAccount.renameAccount')
--- a/src/accounts/context/userAccount.tsx
+++ b/src/accounts/context/userAccount.tsx
@@ -87,7 +87,6 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
   const handleGetAccounts = useCallback(async () => {
     try {
       const accounts = await getUserAccounts()
-
       setUserAccounts(accounts)
       const defaultAcctArray = accounts.find(acct => acct.isDefault)
       if (defaultAcctArray) {
@@ -141,7 +140,14 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
         dispatch(notify(accountRenameSuccess(activeAccount.name, newName)))
 
         activeAccount.name = newName
-        setUserAccounts(userAccounts)
+        const updatedAccounts = userAccounts.map(acct => {
+          if (acct.id === activeAccount.id) {
+            return {...acct, name: accountData.name}
+          }
+
+          return acct
+        })
+        setUserAccounts(updatedAccounts)
 
         if (isFlagEnabled('avatarWidgetMultiAccountInfo')) {
           const name = accountData.name
--- a/src/accounts/context/userAccount.tsx
+++ b/src/accounts/context/userAccount.tsx
@@ -89,15 +89,15 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
       const accounts = await getUserAccounts()
 
       setUserAccounts(accounts)
-      const defaultAcctArray = accounts.filter(line => line.isDefault)
-      if (defaultAcctArray && defaultAcctArray.length === 1) {
+      const defaultAcctArray = accounts.find(line => line.isDefault)
+      if (defaultAcctArray) {
         const defaultId = defaultAcctArray[0].id
         setDefaultAccountId(defaultId)
       }
 
       // isActive: true is for the currently logged in/active account
-      const activeAcctArray = accounts.filter(line => line.isActive)
-      if (activeAcctArray && activeAcctArray.length === 1) {
+      const activeAcctArray = accounts.find(line => line.isActive)
+      if (activeAcctArray) {
         const activeId = activeAcctArray[0].id
         setActiveAccountId(activeId)
       }
--- a/src/accounts/context/userAccount.tsx
+++ b/src/accounts/context/userAccount.tsx
@@ -15,7 +15,6 @@ import {
 } from 'src/shared/copy/notifications'
 
 // Utils
-import {getAccounts, patchAccount} from 'src/client/unityRoutes'
 
 // Metrics
 import {event} from 'src/cloud/utils/reporting'
--- a/src/accounts/context/userAccount.tsx
+++ b/src/accounts/context/userAccount.tsx
@@ -140,16 +140,8 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
         event('multiAccount.renameAccount')
         dispatch(notify(accountRenameSuccess(activeAccount.name, newName)))
 
-        // change the name, and reset the active accts:
-        const updatedAccounts = userAccounts.map(acct => {
-          const account = {...acct}
-          if (acct.id === activeAccount.id) {
-            account.name = accountData.name
-          }
-
-          return account
-        })
-        setUserAccounts(updatedAccounts)
+        activeAccount.name = newName
+        setUserAccounts(userAccounts)
 
         if (isFlagEnabled('avatarWidgetMultiAccountInfo')) {
           const name = accountData.name
--- a/src/accounts/context/userAccount.tsx
+++ b/src/accounts/context/userAccount.tsx
@@ -26,7 +26,11 @@ import {setMe} from 'src/me/actions/creators'
 import {MeState} from 'src/me/reducers'
 
 // API
-import {updateDefaultQuartzAccount} from 'src/identity/apis/auth'
+import {
+  getUserAccounts,
+  updateDefaultQuartzAccount,
+  updateQuartzAccount,
+} from 'src/identity/apis/auth'
 
 export type Props = {
   children: JSX.Element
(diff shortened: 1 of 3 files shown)


src/accounts/context/userAccount.tsx
```diff
@@ -83,32 +86,25 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
-        const activeAcctArray = data.filter(line => line.isActive)
-        if (activeAcctArray && activeAcctArray.length === 1) {
-          const activeId = activeAcctArray[0].id
-          setActiveAccountId(activeId)
-        }
+      // isActive: true is for the currently logged in/active account
+      const activeAcctArray = accounts.filter(line => line.isActive)
+      if (activeAcctArray && activeAcctArray.length === 1) {
+        const activeId = activeAcctArray[0].id
+        setActiveAccountId(activeId)
```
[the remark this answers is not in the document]
subirjolly: good catch!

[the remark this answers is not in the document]
subirjolly: Double checked that only 1 account can be `isActive` at any time. So, I will make this change.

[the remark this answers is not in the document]
subirjolly: updated!

src/accounts/context/userAccount.tsx
```diff
@@ -83,32 +86,25 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
   const handleGetAccounts = useCallback(async () => {
     try {
-      const resp = await getAccounts({})
-      if (resp.status !== 200) {
-        // set user account status to error;...TODO
-        throw new Error(resp.data.message)
+      const accounts = await getUserAccounts()
+
+      setUserAccounts(accounts)
+      const defaultAcctArray = accounts.find(line => line.isDefault)
```
wdoconnell, line 92: This isn't absolutely necessary, but one suggestion I have is that `line` could also be more semantic - it might be clearer if the element were `acct` or `account`.

src/accounts/context/userAccount.tsx
```diff
@@ -132,35 +128,39 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
   }
 
-  async function handleRenameActiveAccount(accountId, newName) {
-    const isActiveAcct = acct => acct.isActive
-    const activeIndex = userAccounts.findIndex(isActiveAcct)
-    const oldName = userAccounts[activeIndex].name
+  const handleRenameActiveAccount = useCallback(
+    async (accountId, newName) => {
+      const isActiveAcct = acct => acct.isActive
+      const activeIndex = userAccounts.findIndex(isActiveAcct)
```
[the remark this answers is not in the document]
subirjolly: updated

src/accounts/context/userAccount.tsx
```diff
@@ -132,35 +128,39 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
+      const oldName = userAccounts[activeIndex].name
```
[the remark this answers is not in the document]
subirjolly: updated

src/accounts/context/userAccount.tsx
```diff
@@ -132,35 +128,39 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
-        userAccounts[activeIndex].name = newName
-        setUserAccounts(userAccounts)
+        const updatedAccounts = userAccounts.map((account, idx) => {
+          const acct = {...account}
+          if (idx === activeIndex) {
+            acct.name = accountData.name
+          }
+
+          return acct
+        })
```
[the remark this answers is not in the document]
subirjolly: deliberate usage of `map` here so that I am not modifying the state variable directly. refactored other parts.

src/identity/apis/auth.ts
```diff
@@ -313,3 +315,29 @@ export const getDefaultAccountDefaultOrg = async (): Promise<
...
+
+  return response.data
+}
+
+export const updateUserAccount = async (accountId, name) => {
+  const resp = await patchAccount({accountId, data: {name}})
+
+  if (resp.status !== 200) {
+    throw new Error(`Account rename update failed`)
+  }
```
hoorayimhelping, line 340: this can throw `401`, `422` and `500`, so let's handle those errors similar to how the other functions in this file handle errors

src/accounts/context/userAccount.tsx
```diff
@@ -83,32 +86,25 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
-        }
 
-        // isActive: true is for the currently logged in/active account
-        const activeAcctArray = data.filter(line => line.isActive)
-        if (activeAcctArray && activeAcctArray.length === 1) {
-          const activeId = activeAcctArray[0].id
-          setActiveAccountId(activeId)
-        }
+      // isActive: true is for the currently logged in/active account
+      const activeAcctArray = accounts.find(line => line.isActive)
```
wdoconnell, line 99: Same here re: `line`

src/accounts/context/userAccount.tsx
```diff
@@ -83,32 +86,25 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
+      if (activeAcctArray) {
+        const activeId = activeAcctArray[0].id
+        setActiveAccountId(activeId)
       }
     } catch (error) {
       event('multiAccount.retrieveAccounts.error', {error})
     }
-  }, [dispatch, defaultAccountId])
```
wdoconnell, line 111: I don't see `dispatch` or `defaultAccountId` in here - was this a mistake, or were these present before?

Is there anything else missing from the useEffect dependency array that would create a new warning?

src/accounts/context/userAccount.tsx
```diff
@@ -132,35 +128,40 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
   }
 
-  async function handleRenameActiveAccount(accountId, newName) {
-    const isActiveAcct = acct => acct.isActive
-    const activeIndex = userAccounts.findIndex(isActiveAcct)
-    const oldName = userAccounts[activeIndex].name
-
-    try {
-      const resp = await patchAccount({accountId, data: {name: newName}})
+  const handleRenameActiveAccount = useCallback(
```
wdoconnell, line 131: What's the rationale for adding a `useCallback` here?

src/accounts/context/userAccount.tsx
subirjolly: i was just moving the existing code. but good catch. i will rename it.

src/accounts/context/userAccount.tsx
subirjolly: they were extraneous. my changes seem to be working fine but i will test my changes again.

src/identity/apis/auth.ts
subirjolly: updated

src/accounts/context/userAccount.tsx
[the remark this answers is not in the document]
wdoconnell: This is probably okay either way, as the old code set the `.name` property directly. But I'm also a little leery of directly manipulating context even if we're about to call setUserAccounts anyway. The O(n) from `map` isn't going to be expensive in a meaningful way, and following the usual pattern of not manipulating state directly keeps us out of trouble if we have to change this later. 

For example, we'll probably move the account state into redux eventually. And in redux we'll be foreclosed from making direct changes to state because the state will be frozen by immer.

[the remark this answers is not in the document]
wdoconnell: @hoorayimhelping Do you have a view on this one?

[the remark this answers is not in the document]
hoorayimhelping: The code here has changed since these comments were made, so I'm trying to get the context of the original discussion, but even without a ton of context, there are two things to consider here:

1. If we're thinking about the performance of this, keep in mind that using the spread operator is not a free operation. It likely has the same or slightly higher cost as a `O(n)` `map` operation, since it needs to iterate through all the elements of a collection and copy them (the copying operation accounting for the slightly higher cost)
2. All things (like performance and security) being equal, the best solution is the one that finds the balance between clarity and simplicity. If mapping over an array and changing a single item is necessary to call attention to, do it. If we can do the same thing without a loop, that is probably better

wdoconnell:
FYI - I'll be able to review first thing tomorrow.

src/accounts/context/userAccount.tsx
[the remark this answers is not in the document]
subirjolly: updated!

src/identity/apis/auth.ts
```diff
@@ -313,3 +315,38 @@ export const getDefaultAccountDefaultOrg = async (): Promise<
...
+  if (response.status === 401) {
+    throw new UnauthorizedError(response.data.message)
+  }
+  if (response.status === 422) {
+    throw new UnprocessableEntityError(response.data.message)
+  }
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+  if (response.status !== 200) {
```
wdoconnell, line 347: What's the rationale for this one - do we expect anything other than a `200`, `422`, or `500`? This does look more cautious, but I only see `200`, `422`, and the `default` server error in the openapi spec.

wdoconnell (changes requested):
I had a few more questions, but otherwise this looks good to me - I might add the video showing this working to the PR too.

src/identity/apis/auth.ts
subirjolly: yea, being extra cautious. makes sense to remove it. will do.

subirjolly:
WORKS

src/accounts/context/userAccount.tsx
```diff
@@ -132,35 +127,38 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
+          }
 
-        // change the name, and reset the active accts:
-        userAccounts[activeIndex].name = newName
-        setUserAccounts(userAccounts)
+          return acct
+        })
+        setUserAccounts(updatedAccounts)
 
         if (isFlagEnabled('avatarWidgetMultiAccountInfo')) {
```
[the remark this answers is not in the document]
subirjolly: will clean this up as part of https://github.com/influxdata/ui/issues/5802

subirjolly:
> I had a few more questions, but otherwise this looks good to me - I might add the video showing this working to the PR too.

updated!

src/identity/apis/auth.ts
wdoconnell: Paranoia is good in my book - I think it's okay if you want to leave it in.

src/accounts/context/userAccount.tsx
```diff
@@ -132,35 +127,38 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
-    const oldName = userAccounts[activeIndex].name
-
-    try {
-      const resp = await patchAccount({accountId, data: {name: newName}})
+  const handleRenameActiveAccount = useCallback(
+    async (accountId, newName) => {
+      const activeAccount = userAccounts.find(acct => acct.isActive)
+      if (!activeAccount) {
+        return
+      }
```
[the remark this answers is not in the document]
subirjolly: absolutely good point. let me check if ever a user can who has a single account can have `inActive` status for that account. if so, then i will add this: `dispatch(notify(defaultAccountNotFoundError()))`. If not then, i will remove this check. Sounds good?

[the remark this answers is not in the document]
subirjolly: I confirmed that it's not possible to have 0 account(s) which have `isActive` as `false`. So, removing this check.

You guys ok with this? ☝️  @asalem1 @wdoconnell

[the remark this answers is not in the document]
subirjolly: done!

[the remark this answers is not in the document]
wdoconnell: Sounds good to me too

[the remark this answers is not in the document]
wdoconnell: Since it doesn't have a chance of occurring, I don't think we need the new notification code

subirjolly:
Resolved merge commits. Require approval again 🙏

src/accounts/context/userAccount.tsx
```diff
@@ -78,32 +79,24 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
-        }
+
+      // isActive: true is for the currently logged in/active account
+      const activeAcctArray = accounts.find(acct => acct.isActive)
+      if (activeAcctArray) {
+        const activeId = activeAcctArray[0].id
+        setActiveAccountId(activeId)
       }
     } catch (error) {
       event('multiAccount.retrieveAccounts.error', {error})
```
wdoconnell, line 97: I don't think hold up is necessary here, but do we want this to go to honeybadger where our other error reporting goes?  

Eventing will just go to snowflake, and I don't think the analytics folks are going to do anything with that.

src/shared/copy/notifications/categories/accounts-users-orgs.ts
```diff
@@ -4,6 +4,11 @@ import {
   defaultSuccessNotification,
 } from 'src/shared/copy/notifications'
 
+export const defaultAccountNotFoundError = (): Notification => ({
```
wdoconnell, line 7: I don't think this notification that's been added is being used anywhere, but please LMK if that's wrong.

wdoconnell (changes requested):
Two more questions on this below

src/accounts/context/userAccount.tsx
subirjolly: i will remove it in the new PR: https://github.com/influxdata/ui/pull/5810

subirjolly: actually, where else should we report? i think we should report it to somewhere.

src/shared/copy/notifications/categories/accounts-users-orgs.ts
subirjolly: ah yea, i added this and then decided to go against it. will add it in the above mentioned PR.

subirjolly: removed in the other PR.

src/accounts/context/userAccount.tsx
subirjolly: Let me use this: `reportErrorThroughHoneyBadger`

src/accounts/context/userAccount.tsx
```diff
@@ -127,28 +128,34 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
...
+          }
+
+          return acct
+        })
+        setUserAccounts(updatedAccounts)
+      } catch (error) {
+        dispatch(notify(accountRenameError(activeAccount.name)))
+        reportErrorThroughHoneyBadger(error, {
+          name: 'UserAccount',
+          context: {
```
wdoconnell, line 151: Pending the tweak to this `reportErrorThroughHoneyBadger` call - adding a more descriptive name, and adding relevant information about the account to context (account name, account id, user name, or id), this looks good to me.

hoorayimhelping (approved):

wdoconnell (approved):

PR #5782, opened 2022-09-15
chore(5780): track usage of fallback worker

### LINKED ISSUE #5780: Add metrics to understand how often the Fallback worker gets used rather than the Worker api
[Web workers have had support in all the major browsers since 2012](https://caniuse.com/webworkers). The browsers that seem to have partial support for other workers (shared web workers, service workers) are all mobile browsers, which Monaco editor explicitly does not support. While [the Web Worker Mozilla docs specify checking for support](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#worker_feature_detection), I'm not sure the `Fallback` shim is all that necessary. As web workers have their own event loops, i.e. they have to be their own thread/process, I worry that trying to do the same in the web browser itself is going to lead to async event loop blocking on sufficiently complex flux; using the `Fallback` might be creating bugs.

wiedld:
Closes #5780 

For architecture design decisions, want more metrics on editor-Lsp usage.

### Checklist

### DIFF
--- a/src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
+++ b/src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
@@ -23,8 +23,9 @@ import fluxWorkerUrl from 'worker-plugin/loader!./worker/flux.worker'
 import Fallback from 'src/languageSupport/languages/flux/lsp/worker/flux.fallback'
 import {EditorType} from 'src/types'
 
-// error notification
+// utils
 import {reportErrorThroughHoneyBadger} from 'src/shared/utils/errors'
+import {event} from 'src/cloud/utils/reporting'
 
 // install Monaco language client services
 MonacoServices.install(monaco)
@@ -71,8 +72,10 @@ export function initLspWorker() {
     return
   }
   if (window.Worker) {
+    event('Used LSP worker.')
     worker = new Worker(fluxWorkerUrl)
   } else {
+    event('Used LSP fallback worker.')
     worker = new Fallback()
   }
   worker.onerror = (err: ErrorEvent) => {


src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
```diff
@@ -71,8 +72,10 @@ export function initLspWorker() {
     return
   }
   if (window.Worker) {
+    event('Used LSP worker.')
```
rockstar, line 75: I don't see this function in scope. Where does this go? How can we check up on it?

Additionally, I'm not sure we care when we use the native worker. That's the happy path. I just want to know if we have users that are out there using browsers from 2011.

src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
```diff
@@ -71,8 +72,10 @@ export function initLspWorker() {
...
     worker = new Worker(fluxWorkerUrl)
   } else {
+    event('Used LSP fallback worker.')
```
rockstar, line 78: Can we send any other additional information? I'd like to know what browser it is that they are using. Maybe also the org id or something? I think there's value in understanding that user's use case, so maybe Product can learn more about the user.

rockstar (commented):
Have some questions I'd like to sort out before going too much further.

src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
wiedld: Yes, it's in scope. see:
```
```

It gets reported with our amplitude metrics.

wiedld: Reporting both numbers, web worker versus fallback worker, so we can examine as a ratio. Then we can easily justify with "X ratio of users are...".

src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
wiedld: Metadata is collected with the event, including platform (e.g. web) and browser.

src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
rockstar: Oh shoot, I must've just scanned over it in this patch. I saw it, wondered what it did, so I went to my vscode (where I already had this file open) and tried to walk it back. Sorry about that!

rockstar: > Reporting both numbers, web worker versus fallback worker, so we can examine as a ratio.

I'm not sure we care about a ratio. I'm mostly interested in "does this code get used _at all_." I have a suspicion that it probably doesn't, but if we need to figure out how to move on from it, that's a completely different task. If we look at this in two weeks and see no traffic, we don't even have to care about the happy path.

I'm happy to leave it in, but I just want to be clear that it is a little more than we need right now.

rockstar (approved):

src/languageSupport/languages/flux/lsp/monaco.flux.lsp.ts
wiedld: Yeah. I'm just hoping that if it's use at all...the ratio can be so tiny that we decide not to support. 🤷🏼  Just a thought.

PR #5791, opened 2022-09-15
chore: add guards to pushing into selected variable array, remove extra reporting

wdoconnell:
We added more error [reporting to honeybadger](https://github.com/influxdata/ui/pull/5689/) around a code block in `variables` that throws an `object not extensible` error to resolve this [issue](https://github.com/influxdata/ui/issues/5523). We're now getting more detailed error reports, but when the monitor301 dashboard that experiences this error hits it, it hits it 300+ times. We're getting reports of sluggishness on that monitor301 dashboard and the following error, which suggests the network reporting of the error might be overkill. Also, we don't really need it anymore, since we have logs of the error in honeybadger.

This PR does two things. 
- First, it removes the honeybadger reporting added to this block. 
- Second, it attempts to resolve the immediate error by _not_ pushing a default value into the 'selected' variable array if the object isn't extensible.

This is safe to do because this _only_ prevents the error-generating push in the case where the object isn't extensible, which (before we added error handling) would throw an uncaught error.

This is a shallow fix. The root cause of `vari.selected` not being extensible is probably from the assumptions this selector makes about object cloning. Trying to clone an extensible copy of a frozen piece of state, vari, like this selector does: `vari = {...vari}` doesn't work for properties of that object that are themselves objects - those are still going to point to the same reference in memory. Those should probably be _deep_ clones. But I'll save that for another PR since that requires a more substantial rework and probably a feature flag.

### Checklist

### DIFF
--- a/src/variables/selectors/index.tsx
+++ b/src/variables/selectors/index.tsx
@@ -28,7 +28,6 @@ import {AppState, VariableArgumentType, Variable} from 'src/types'
 
 // Utils
 import {filterUnusedVars} from 'src/shared/utils/filterUnusedVars'
-import {reportErrorThroughHoneyBadger} from 'src/shared/utils/errors'
 
 export const extractVariableEditorName = (state: AppState): string => {
   return state.variableEditor.name
--- a/src/variables/selectors/index.tsx
+++ b/src/variables/selectors/index.tsx
@@ -215,17 +215,11 @@ export const getVariable = (state: AppState, variableID: string): Variable => {
       vari.selected.push(vals[0])
     } catch (err) {
       // Temporary measure to resolve errors relating to pushing into non-extensible object.
-      reportErrorThroughHoneyBadger(err, {
+      console.error({
         name: 'Failed to set selected variable to default, zero-indexed value',
-        context: {
-          identityState: state.identity,
-          resourceState: state.resources,
-          variable: vari,
-          variableType: vari.arguments.type,
-          normalizedVariable: vals,
-          variableLength: vari.selected.length,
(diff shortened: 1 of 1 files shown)


philjb (approved):

wdoconnell:
Following up on this - @philjb  confirmed that blocking honeybadger sped up one of the dashboards, so I think this is a worthwhile fix. We can navigate to the monitor301 dashboard and look at the console.errors to debug the underlying issue.

TCL735 (approved):

src/variables/selectors/index.tsx
```diff
@@ -211,22 +210,8 @@ export const getVariable = (state: AppState, variableID: string): Variable => {
   }
 
   if (!vari.selected.length && vals.length) {
-    try {
+    if (Object.isExtensible(vari.selected)) {
```
wdoconnell, line 213: Let's use `isExtensible` here. Per the error, we know `vari.selected` isn't extensible, but not the precise way it was made not extensible - e.g., `Object.freeze`, `Object.preventExtensions`, `Object.seal`. It's possible for the object not to be _frozen_ but still be unextensible if properties can be deleted.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isExtensible
https://www.oreilly.com/library/view/javascript-the-definitive/9781449393854/rn01re159.html

hoorayimhelping (approved):

PR #5810, opened 2022-09-16
feat(multi-org): update account name in multiorg header and profile page

### LINKED ISSUE #5787: Multi Org Phase 1 - changing account name should update multiorg header and profile page
Related Issue: https://github.com/influxdata/ui/issues/5029

Changing the name of the account on `account/settings` page doesn't reflect on the multi-org header and nor does it reflect on the profile page.

**Proposal:**
We should make `UserAccountProvider` wrap `App` so that we have 1 instance of userAccount data and whereever we need to use that data, we can just use `UserAccountContext`.

**Current behavior:**
Account name change only happens in 1 of 3 places above mentioned.

**Desired behavior:**
Account name change happens in all 3 places above mentioned.

**Alternatives considered:**
Describe other solutions or features you considered.

Update: All 3 places should be updated after this issue is done as well: https://github.com/influxdata/ui/issues/5788

### LINKED ISSUE #5788: Multi Org Phase 1 - update account in redux when changing account name(profile page) in useraccountcontext
Related Issue: https://github.com/influxdata/ui/issues/5787

Discovered that the account info on profile page is being fetched from a separate endpoint and is stored in redux. when we update account name in `UserAccountContext`, that does not update the name on profile page. So, we need to dispatch the account name update to redux so that the account names are in sync.

subirjolly:
Closes #5787 
Closes #5788 

Changing the name of the account on account/settings page doesn't reflect on the multi-org header and nor does it reflect on the profile page.

Proposal:
We should make UserAccountProvider wrap App so that we have 1 instance of userAccount data and whereever we need to use that data, we can just use UserAccountContext.

Current behavior:
Account name change only happens in 1 of 3 places above mentioned.

Desired behavior:
Account name change happens in all 3 places above mentioned.

Alternatives considered:
Describe other solutions or features you considered.

## BEFORE
BEFORE

## AFTER
AFTER

### Checklist


src/App.tsx
```diff
@@ -137,6 +138,8 @@ const App: FC = () => {
 
 export default () => (
   <AppSettingProvider>
-    <App />
+    <UserAccountProvider>
+      <App />
+    </UserAccountProvider>
```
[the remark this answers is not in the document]
subirjolly: I updated it to have this check.
```
{CLOUD && isFlagEnabled('multiOrg') && shouldUseQuartzIdentity() ?
...
```

Please feel free to suggest a better way.

src/App.tsx
```diff
@@ -137,6 +138,12 @@ const App: FC = () => {
...
 export default () => (
   <AppSettingProvider>
-    <App />
+    {CLOUD && isFlagEnabled('multiOrg') && shouldUseQuartzIdentity() ? (
+      <UserAccountProvider>
+        <App />
+      </UserAccountProvider>
+    ) : (
+      <App />
+    )}
```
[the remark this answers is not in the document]
wdoconnell: My recommendation is to wait to figure this out until the other PR, which refactors the account context provider, is tested, merged, and working. I do think it's a good idea to move the Account Provider higher up, but we don't want to subordinate to feature flags. Otherwise, we'll break the account information for users if we have to flip off one of these flags.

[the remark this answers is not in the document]
wdoconnell: Let's take this one slow and think about it. But my initial thinking is that the appropriate check would be something like `if (CLOUD)` rather than `if (CLOUD)` plus feature flag logic.

subirjolly:
@asalem1 @wdoconnell Please let me know if I need to make any other change.

src/App.tsx
```diff
@@ -137,6 +139,12 @@ const App: FC = () => {
 
 export default () => (
   <AppSettingProvider>
-    <App />
+    {CLOUD ? (
```
wdoconnell, line 142: Before we restructure it like this - could you map out what the relationship between the User Account Provider and the app is?  Is this provider exclusively cloud-only or do parts of it need to be available in OSS?

For example, it looks to me like `AccountPage` contains the `UserAccountProvider` even though that's a shared route (at least according to SetOrg). I think we need to be 100% confident what is and isn't using the context before we move it conditionally.

src/accounts/AccountPage.tsx
```diff
@@ -210,9 +207,7 @@ const AccountPage: FC = () => {
     <Page titleTag={pageTitleSuffixer(['Account Settings Page'])}>
       <AccountHeader testID="account-page--header" />
       <UsersProvider>
-        <UserAccountProvider>
-          <AccountAboutPage />
-        </UserAccountProvider>
+        <AccountAboutPage />
```
wdoconnell, line 210: Same comment here. AFAIK, this is a shared component. If we delete this provider, and aren't in the cloud, this component will no longer have access to the `UserAccountProvider`. So we need to determine whether or not that component needs access to the provider in CLOUD v. non-CLOUD.

src/identity/actions/creators/index.ts
```diff
@@ -35,3 +42,9 @@ export const setQuartzIdentityStatus = (status: RemoteDataState) =>
     type: SET_QUARTZ_IDENTITY_STATUS,
     status,
   } as const)
+
+export const setQuartzIdentityAccountName = (account: CurrentAccount) =>
```
wdoconnell, line 46: To avoid confusion we might want the function and the action type to be the same (set current identity account name versus set quartz identity account name).

wdoconnell (changes requested):

src/App.tsx
```diff
@@ -44,7 +44,9 @@ import {executeVWO} from 'src/utils/vwo'
 
 // Styles
 import './MultiOrgOverrideStyles.scss'
-import {Page} from '@influxdata/clockface'
+
+// Providers
+import {UserAccountProvider} from './accounts/context/userAccount'
```
hoorayimhelping, line 49:
```suggestion
import {UserAccountProvider} from 'src/accounts/context/userAccount'
```

subirjolly: good catch

src/App.tsx
subirjolly: Upon checking and discussing with @wdoconnell , we have determined that this check is correct and wouldn't impact OSS negatively as the affected components/pages shouldn't be available to OSS without a copied/bookmarked link.

src/App.tsx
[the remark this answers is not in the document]
subirjolly: I spun up OSS instance and tested it with multiOrg on/off. Worked fine. Noticed one thing with Billing Page is that in OSS, it's redirected to 404 which wasn't happening for `accounts/settings` page. So, moved this route behind cloud flag as well so we get a consistent response.

subirjolly:
## UPDATED VIDEO(GIF)
WORKS

src/identity/apis/auth.ts
```diff
@@ -316,5 +316,9 @@ export const updateUserAccount = async (accountId, name) => {
     throw new ServerError(response.data.message)
   }
 
-  return response.data
+  // TODO: Remove the following line of code when this PR(https://github.com/influxdata/quartz/issues/6846) is done
+  //       Instead return response.data.
+  const data = {...response.data, id: parseInt(`${response.data.id}`, 10)}
```
wdoconnell, line 321: Would it be easier to just set `id` to `Number(response.data.id)` --  instead of using `parseInt`, stringifying the value, and passing passing base 10 as a param?

src/App.tsx
wdoconnell: This is a great solution to the problem of consistency between providers. I'm not worried about re-render impact of putting the provider above app since the `useEffect` hook there will only run when the default or current account changes (default account will rarely change, and when current account changes, the app is going to re-render anyway).  But we should watch for any unexpected rendering change until the account context is converted over to redux.

wdoconnell (commented):
I might consider rebasing to current master since we have a lot of commits now, but this looks good to me. I'll approve pending these last couple comments.

src/identity/apis/auth.ts
subirjolly: ah I hadn't used this function before. this is a neato. Thanks!

wdoconnell (approved):

PR #5817, opened 2022-09-19
feat: add new time range component for the UI for the new data explorer

### LINKED ISSUE #4828: Improved Time Range Selection
## User Story

<p></p>

As a user, I want an improved time range selector that makes it easy for me to both choose a relative time range (including manually entering a range) and a specific custom range with a calendar/date picker so that it is easy for me to define my range visually.

<p></p>

## Acceptance Criteria

<p></p>

- Julia has created a new UI design for this; this ticket is for the implementation of that design (see attached)

<p></p>

<https://www.figma.com/file/0qAntPk5LVangAHguWT74X/Query-Experience-Project?node-id=1125%3A44592>

<p></p>

<p></p>

<p></p>

asalem1:
Closes #4828

The purpose of this PR is to add the new timerange dropdown for the new data explorer. As part of this work, some updates to the existing stylesheet were made in order to flatten out the way that styles were being applied / overwritten. These changes allow us to more easily navigate styles and overwrites without sacrificing the existing styles. 

As part of this work, the existing placement of the timezone dropdown has been moved into the context of the dropdown and has been removed from its previous placement in the Results component. Finally, some shared validation logic has been moved into a more central utility file for both the existing and new date picker to make use of.

This PR is still a WIP since we are waiting on final designs for informing the user of all the specific input types available

### DIFF
--- a/src/shared/components/dateRangePicker/NewDatePicker.tsx
+++ b/src/shared/components/dateRangePicker/NewDatePicker.tsx
@@ -1,5 +1,6 @@
 import React, {FC, useState, useContext, useEffect, useCallback} from 'react'
 import {
+  AlignItems,
   Button,
   ComponentColor,
   ComponentSize,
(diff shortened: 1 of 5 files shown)


src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -0,0 +1,512 @@
...
+        >
+          <FlexBox
+            direction={FlexDirection.Row}
+            alignItems={AlignItems.Stretch}
+          >
+            {isDatePickerOpen && (
+              <Dropdown.Item
+                className="react-datepicker-ignore-onclickoutside"
+                selected={false}
+                style={{
```
drdelambre, line 358: move this into the sass

src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -0,0 +1,512 @@
+import React, {FC, useState, useContext, useEffect, useCallback} from 'react'
```
drdelambre, line 1: nit: this component is pretty big and pretty dense. nothing specifically bad about how it is currently, just historically / stylistically larger components tend to mix dependencies within the business logic more than they may need to. if there's a way that breaking up the component also breaks up the `handle` logic, i'd float towards that.

src/shared/components/dateRangePicker/utils.ts
```diff
@@ -0,0 +1,13 @@
+import {isISODate} from 'src/shared/utils/dateTimeUtils'
+
+import {isValidStrictly} from 'src/utils/datetime/validator'
+
+export const isValidDatepickerFormat = (d: string): boolean => {
```
drdelambre, line 5: i wish there was a way to use the wasm to validate these things as the validation is defined in there and is already on the page

drdelambre (approved):
looks good. validation observation was a thought outside the scope of this PR

PR #5831, opened 2022-09-19
feat: add user feedback button

### LINKED ISSUE #5790: Add "Provide Feedback" link to the new script editor
In an effort to gather feedback after we've launched the new script editor, we would like to add a more permanent option for users to opt-in to provide feedback instead of a time-based or event-triggered survey. This will allow users to provide feedback more than once and will be less distracting while they are in the middle of building a query.

Image

In order to trigger the UserPilot survey from this link, we'll need to add the following code: 
`userpilot.trigger("1663269889fDfn2554");`

appletreeisyellow:
Closes #5790 

This PR adds a button in new script editor for users to provide feedback through third-party Userpilot. 

This button is behind the feature flag `userFeedback` created in influxdata/idpe#15847 which is already merged

## After
Screen Shot 2022-09-19 at 5 30 31 PM

https://user-images.githubusercontent.com/14298407/191354788-17aad2f4-e40c-4f92-a3c4-71dd5d8a6e39.mov

### Checklist


garylfowler:
I actually like the link a bit better than a button (less real estate and seems less intrusive).  I don't feel that the user will get confused, even though I do understand, from a purist standpoint, the feedback.

Issue #5628, opened 2022-09-01
Bug - Resolve stray id error - checkout page


wdoconnell:
Occasionally we see a honeybadger alert on the checkout page: `Cannot read properties of null (reading 'id')`. Investigate what the cause of this error is, and whether we're missing an optional chaining operator or some other guard against trying to access the `id` property on an object that may not be retrieved in time for a component to load.

PR #5848, opened 2022-09-21
fix(reporting): guards to prevent accessing null values from useSelector

ChitlangeSahas:
Closes [#5628](https://github.com/influxdata/ui/issues/5628)

My hypothesis of what was happening: 

`useSelector` is an async operation, and the current config of useEffect hook it runs on `mount` as well as `update`.

Perhaps we don't need to run this on mount, i.e. when the org info isn't retrieved yet. 

We can see these errors in honey-badger occasionally. Seems like a race condition. 

Screen Shot 2022-09-21 at 12 00 24 PM

Screen Shot 2022-09-21 at 12 00 12 PM

`

### Checklist

### DIFF
--- a/src/cloud/components/experiments/GoogleOptimizeExperiment.tsx
+++ b/src/cloud/components/experiments/GoogleOptimizeExperiment.tsx
@@ -36,7 +36,7 @@ export const GoogleOptimizeExperiment: FC<Props> = ({
       getExperimentVariantId(experimentID, activationEvent)
     )
 
-    if (variantID) {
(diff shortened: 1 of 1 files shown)


TCL735 (approved):

hoorayimhelping (approved):
where `GoogleOptimizeExperiment` is being called might also have an affect on when this error shows up. I wonder if it's a specific place where this component is rendered that this error happens

ChitlangeSahas:
@hoorayimhelping  yes, this only happens on the Checkout Page.

PR #5865, opened 2022-09-22
fix: tags and fields are pointer objects, so stop resetting with the same pointer to a stateful object.

wiedld:
## Bug:
* fields was holding state
* session itself was correctly updating

https://user-images.githubusercontent.com/10232835/191843865-473eea69-58bf-48eb-91e9-2410d82ca0fa.mov

* error disappeared on page reload -- so is the non-session store.
* we were using the same pointer object when resetting fields, and tags. Fixed that code.


appletreeisyellow (dismissed):
Good call!

appletreeisyellow (approved):

rockstar (approved):

Issue #5818, opened 2022-09-19
Epic: Library Upgrades/Maintenance


kristinarobinson:
Upgrade all NPM library dependencies.

PR #5866, opened 2022-09-22
feat: replace remark-external-links with rehype-external-links and upgrade react-markdown

TCL735:
Part of #5818  

When attempting to upgrade `remark-external-links`, we encountered a deprecation notice:
Screen Shot 2022-09-22 at 1 06 46 PM

Instead of using a deprecated dependency, let's switch over to `rehype-external-links`. While we are at it, upgrade `react-markdown` as well, to ensure compatibility.


TCL735:
The failures look legitimate. We can't switch to `rehype-external-links` without even more changes.

PR #5867, opened 2022-09-22
fix: flaky about test

wiedld:
Fix the flaky about test.


wiedld:
Tried simple fix per suggestion. Didn't work -- all firefox tests are failing. Will leave for someone else to pickup tmrw.

wdoconnell:
Thanks very much for jumping on this - but I think now resolved by UI #5872

PR #5872, opened 2022-09-23
fix: resolve flaky about test 

wdoconnell:
The shared `about` test in the Cloud CI run has gone flaky in firefox. In firefox, the org Rename button is deemed visible, but on click, it's deemed detached from the DOM. 

Underlying cause is the `useEffect` hook in this component. This page requires a separate network call in cloud to retrieve additional information about the user's org. The cypress test verifies that the rename button is visible, but the panel re-renders on response from the new network call, which causes cypress to find the previously identified element detached from the DOM.

Credit to @TCL735 for the more comprehensive fix here, which explicitly waits for the response from that network call in cloud before checking for the rename button.

Repro:
--

https://user-images.githubusercontent.com/91283923/191995187-d050890b-e0b1-48d4-8c06-44f089856a7a.mov

### Checklist

### DIFF
--- a/cypress/e2e/shared/about.test.ts
+++ b/cypress/e2e/shared/about.test.ts
@@ -1,10 +1,18 @@
 import {Organization} from '../../../src/types'
 
 describe('About Page', () => {
+  const apiOrgPath =
+    Cypress.env('dexUrl') === 'OSS' ? '/api/v2/orgs/*' : '/api/v2/quartz/orgs/*'
+  const CLOUD = Cypress.env('dexUrl') === 'OSS' ? false : true
+
   beforeEach(() => {
     cy.flush()
     cy.signin()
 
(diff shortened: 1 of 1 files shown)


wiedld (dismissed):
Thanks for debugging quickly.
Yes, we do have an issue with excessive re-renders.

cypress/e2e/shared/about.test.ts
```diff
@@ -1,28 +1,39 @@
 import {Organization} from '../../../src/types'
 
 describe('About Page', () => {
+  const apiOrgPath =
```
subirjolly, line 4: Can we also have IS_OSS const so that you don't have to do url check again in a couple of other places?

wdoconnell: Converted to a constant

subirjolly (approved):

subirjolly (approved):

PR #5881, opened 2022-09-25
refactor: simplify worker and enable "ready" event

### LINKED ISSUE #5832: Remove the Fallback lsp interface
The intention for the `Fallback` interface was to provide the lsp support when web workers aren't available. However, if, using content-security-policy, we disable all web workers (browsers don't have a mechanism to do it (anymore?)), monaco itself doesn't work right. Given that monaco requires web workers, it's totally okay for the ui to require web workers as well. Small thing, but it will simplify our code here.

### LINKED ISSUE #5859: Remove lsp buffering
Once the code in #5844 no longer needs the buffer code, it should be removed from source.

DOD:
  - the code in `src/languangeSupport/languages/flux/lsp/worker/buffer.ts` is deleted

rockstar:
This patch dramatically reduces the complexity for communicating with
the lsp worker. The worker can now signal its ready state, allowing
dependent resources to not start until the lsp itself has loaded and is
ready.

Additionally, that initialization all happens within the flux editor.
There is a single place where a guard is added to not try to act on the
resources initialized by the lsp initialization until that
initialization is complete, but there is a note to address that as part
of this on-going work.

Fixes #5832
Fixes #5859


src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -61,7 +80,54 @@ const FluxEditorMonaco: FC<Props> = ({
   wrapLines,
   variables,
 }) => {
-  const connection = useRef<ConnectionManager>(null)
```
wiedld, line 64: with this change, make sure your QA includes when there are multiple editors in the same page. Example, in a flow/notebook there can be >1 editor at once.

src/languageSupport/languages/flux/lsp/worker.ts
```diff
@@ -0,0 +1,31 @@
+// LSP worker code.
+//
+// IMPORTANT! All the code here will run in a web worker. This has ramifications in
+// the environment, e.g. there is no console, no dom, etc.
```
wiedld, line 4: no dom, but there is a console we can access. fyi.

rockstar: This seems to have iffy implementations across browsers, so I just noted it. It definitely works in chrome, but Firefox, even though it claims to support it.

rockstar:
Superseded by #6272

PR #5884, opened 2022-09-26
chore: remove quartzMe references from org pages

wdoconnell:
Connects #4821 
For other PRs removing this code see #5847, #5876, #5878, #5883

Removes references to the quartzMe data shape on the org profile tab and related pages.

### Checklist

Authors and Reviewer(s), please verify the following:

- [X] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [X] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [X] Documentation updated or issue created (provide link to issue/PR)
- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [X] Feature flagged, if applicable

PR #5883, opened 2022-09-26
chore: remove quartzMe references from operator pages

wdoconnell:
Connects #4821 
For other PRs removing this code see #5847, #5876, #5878

Removes references to the quartzMe data shape on operator-related pages.

### Checklist

Authors and Reviewer(s), please verify the following:

- [X] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [X] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [X] Documentation updated or issue created (provide link to issue/PR)
- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [X] Feature flagged, if applicable

PR #5876, opened 2022-09-23
chore: remove refs to quartzMe data shape in onboarding experience

wdoconnell:
Connects issue [#4821](https://github.com/influxdata/ui/issues/4821) and PR #5847 

Removes references to the quartzMe data shape from the homepage experience portion of the app.

This is one of several PRs that are going to remove references to the `quartzMe` data shape. The app no longer uses `quartzMe`, but is still referencing properties using the old `quartzMe` selectors (which are populated using a conversion function from the `identity` api). 

All references using the old selector will be removed; then the conversion function that populates the `quartzMe` reducer using the identity data will be removed.

### Checklist

Authors and Reviewer(s), please verify the following:

- [X] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [X] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [X] Documentation updated or issue created (provide link to issue/PR)
- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [X] Feature flagged, if applicable - NO

PR #5885, opened 2022-09-26
chore: remove refs to quartzMe data shape in homepage and nav

wdoconnell:
Connects #4821 

For other PRs removing this code see #5847, #5876, #5878, #5883, #5884

Removes references to the quartzMe data shape in the nav bar and home page.

### Checklist

### DIFF
--- a/src/pageLayout/constants/navigationHierarchy.ts
+++ b/src/pageLayout/constants/navigationHierarchy.ts
@@ -36,7 +36,7 @@ export interface NavItem {
 export const generateNavItems = (): NavItem[] => {
   const state: AppState = getStore().getState()
   const orgID = state?.resources?.orgs?.org?.id ?? ''
-  const quartzMe = state?.me?.quartzMe ?? null
+  const identity = state?.identity?.currentIdentity ?? null
 
   const orgPrefix = `/orgs/${orgID}`
 
@@ -194,7 +194,10 @@ export const generateNavItems = (): NavItem[] => {
     },
     {
       id: 'operator',
-      enabled: () => CLOUD && quartzMe?.isOperator === true,
+      enabled: () =>
+        CLOUD &&
+        (identity?.user?.operatorRole === 'read-write' ||
+          identity?.user?.operatorRole === 'read-only'),
       testID: 'nav-item--operator',
       icon: IconFont.Shield,
       label: 'Operator',
--- a/src/shared/containers/GetOrganizations.tsx
+++ b/src/shared/containers/GetOrganizations.tsx
@@ -20,9 +20,9 @@ import RouteToOrg from 'src/shared/containers/RouteToOrg'
 
 // Selectors
 import {getAllOrgs} from 'src/resources/selectors'
-import {getMe, getQuartzMe, getQuartzMeStatus} from 'src/me/selectors'
+import {getMe} from 'src/me/selectors'
 import {
-  selectQuartzIdentity,
+  selectCurrentIdentity,
   selectQuartzIdentityStatus,
 } from 'src/identity/selectors'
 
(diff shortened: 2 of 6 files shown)


src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -36,7 +36,7 @@ export interface NavItem {
 export const generateNavItems = (): NavItem[] => {
   const state: AppState = getStore().getState()
   const orgID = state?.resources?.orgs?.org?.id ?? ''
-  const quartzMe = state?.me?.quartzMe ?? null
+  const identity = state?.identity?.currentIdentity ?? null
```
wdoconnell, line 39: Erring on the safe side of including a guard here: we aren't using a selector to grab this, as this is a pure function that takes in a direct reference to state. 

It should be the case that the default state exists in the store already, but not worth the risk of removing the guards imho.

src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -194,7 +194,10 @@ export const generateNavItems = (): NavItem[] => {
     },
     {
       id: 'operator',
-      enabled: () => CLOUD && quartzMe?.isOperator === true,
+      enabled: () =>
+        CLOUD &&
+        (identity.user.operatorRole === 'read-write' ||
+          identity.user.operatorRole === 'read-only'),
       testID: 'nav-item--operator',
```
wdoconnell, line 201: In identity land, there is no `isOperator`: there is only ~~Zuul~~ `operatorRole`.

`OperatorRole` two potential values - `read-write` or `read-only`. Since this check is what gives access to the operator route, I'd rather an extra degree of safety here to check not just that `operatorRole` is populated, but that it's exactly what we expect it to be - read-write or read-only.

src/shared/containers/GetOrganizations.tsx
```diff
@@ -35,108 +35,89 @@ import {convertStringToEpoch} from 'src/shared/utils/dateTimeUtils'
...
 // Thunks
 import {getQuartzIdentityThunk} from 'src/identity/actions/thunks'
 
-const canAccessCheckout = (me: Me): boolean => {
-  if (Boolean(me?.isRegionBeta)) {
-    return false
-  }
-  return me?.accountType !== 'pay_as_you_go' && me?.accountType !== 'contract'
-}
-
```
wdoconnell, line 51: The purpose of this function was to convert the combination of the `isRegionBeta` and `accountType` properties into a `boolean` of whether the user could access checkout. We don't need that anymore, as `isUpgradeable` is a `boolean` that does just that.

src/shared/containers/GetOrganizations.tsx
```diff
@@ -35,108 +35,89 @@ import {convertStringToEpoch} from 'src/shared/utils/dateTimeUtils'
...
           email,
           organization_id: orgId,
         },
       })
     }
-  }, [quartzMeStatus, status]) // eslint-disable-line react-hooks/exhaustive-deps
-
-  const isRegionBeta = quartzMe?.isRegionBeta ?? false
-  const accountType = quartzMe?.accountType ?? 'free'
+  }, [identityStatus, orgLoadingStatus]) // eslint-disable-line react-hooks/exhaustive-deps
```
wdoconnell, line 103: We could add `account.type`, `meId`, etc. to the dependency array here, but this looks _super_ intentional given that the `useEffect` hook is keyed off of the redux loading state and not directly off of the properties being sent to google analytics, so I'd rather not mess with the array at this time.

src/shared/containers/GetOrganizations.tsx
```diff
@@ -35,108 +35,95 @@ import {convertStringToEpoch} from 'src/shared/utils/dateTimeUtils'
...
-  return me?.accountType !== 'pay_as_you_go' && me?.accountType !== 'contract'
-}
-
 const GetOrganizations: FunctionComponent = () => {
-  const {status, org} = useSelector(getAllOrgs)
-  const identity = useSelector(selectQuartzIdentity)
-  const quartzMeStatus = useSelector(getQuartzMeStatus)
-  const quartzMe = useSelector(getQuartzMe)
-  const quartzIdentityStatus = useSelector(selectQuartzIdentityStatus)
   const {id: meId = '', name: email = ''} = useSelector(getMe)
```
hoorayimhelping, line 45: are `meId` or `email` used anymore? they seem to have been replaced with `account.id` and `account.email`

src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -194,7 +194,10 @@ export const generateNavItems = (): NavItem[] => {
     },
     {
       id: 'operator',
-      enabled: () => CLOUD && quartzMe?.isOperator === true,
+      enabled: () =>
+        CLOUD &&
+        (identity?.user?.operatorRole === 'read-write' ||
+          identity?.user?.operatorRole === 'read-only'),
```
hoorayimhelping, line 200: In the future, especially if we have more calls to this expression, we might consider making this a helper function

src/shared/containers/GetOrganizations.tsx
```diff
@@ -35,108 +35,95 @@ import {convertStringToEpoch} from 'src/shared/utils/dateTimeUtils'
...
-  const {account} = identity.currentIdentity
+  const {account, user, org: quartzOrg} = useSelector(selectCurrentIdentity)
+  const identityLoadingStatus = useSelector(selectQuartzIdentityStatus)
+  const {status: orgLoadingStatus, org} = useSelector(getAllOrgs)
```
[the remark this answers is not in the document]
wdoconnell: `quartzOrg` is the user's current organization as retrieved from quartz (api/v2/quartz/identity) - see the identity reducer and src/identity/actions/thunks.

`org` - from selector `getAllOrgs` - is the user's current organization as retrieved from idpe (/api/v2/orgs) - see the organizations (and resources) reducer and src/organizations/actions/thunks.

In `CLOUD` the id and name of `quartzOrg` and `org` are the same and refer to the same org. In OSS, only `org` is populated (because quartz doesn't exist, and its endpoints aren't called), and `quartzOrg` is an object with default values (empty string). 

Depending on whether/how the app splits: in CLOUD, continuing to use the IDPE org id is fine and avoids us adding `if (cloud)` everywhere in the app that the org id is currently used. But there will no longer be any need going forward to refer to the IDPE org in CLOUD, as that endpoint isn't giving us anything that the new quartz endpoints don't provide. Whereas in OSS we will continue to need the IDPE org endpoint and reducer.

src/shared/containers/GetOrganizations.tsx
wdoconnell: I think that's correct that we can replace these. These are used in `useEffect` hooks that won't run in OSS. I believe the replacement should be `user.id` and `user.email` from quartz.

src/shared/containers/GetOrganizations.tsx
```diff
@@ -35,108 +35,95 @@ import {convertStringToEpoch} from 'src/shared/utils/dateTimeUtils'
...
-        email,
-        orgID: org.id,
+        email: user.email,
+        orgID: quartzOrg.id,
         accountID: account.id,
         accountName: account.name,
       })
     }
-  }, [meId, email, org?.id, account?.id, account?.name])
+  }, [meId, user.email, quartzOrg.id, account.id, account.name])
```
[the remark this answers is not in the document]
wdoconnell: Same - we should be able to use `user.id` for reporting. Let me double check with AIM that those ids are also the same.

src/shared/containers/GetOrganizations.tsx
```diff
@@ -35,108 +35,95 @@ import {convertStringToEpoch} from 'src/shared/utils/dateTimeUtils'
...
+  const {account, user, org: quartzOrg} = useSelector(selectCurrentIdentity)
+  const identityLoadingStatus = useSelector(selectQuartzIdentityStatus)
+  const {status: orgLoadingStatus, org} = useSelector(getAllOrgs)
+
+  const shouldLoadIdentity =
+    identityLoadingStatus === RemoteDataState.NotStarted
+  const identityFinishedLoading = identityLoadingStatus === RemoteDataState.Done
+
+  const shouldLoadOrg = orgLoadingStatus === RemoteDataState.NotStarted
+  const orgFinishedLoading = orgLoadingStatus === RemoteDataState.Done
```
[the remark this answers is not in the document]
wdoconnell: Maybe it makes sense, in that case, to just remove my changes here. The old code just checks whether the `status === RemoteDataState.NotStarted` or `RemoteDataState.Done`, which while repetitive in one or two cases is still perfectly clear, and probably doesn't require any abstraction.

src/shared/containers/GetOrganizations.tsx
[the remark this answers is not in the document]
hoorayimhelping: >In `CLOUD` the id and name of `quartzOrg` and `org` are the same and refer to the same org. In OSS, only `org` is populated (because quartz doesn't exist, and its endpoints aren't called), and `quartzOrg` is an object with default values (empty string).

so in both cloud and oss environments, we're going to call an unnecessary selector? I think we should have a single `org` selector called `getOrg` that encapsulates and abstracts the logic of whether we're in cloud or oss environments so that at the component level, we only have to call `useSelector(getOrg)` and not worry about it. Because there will surely be similar questions about why we're selecting both `org` and `quartzOrg` from other people.

What do you think?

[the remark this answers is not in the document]
wdoconnell: I like that idea - and let me play with it in the current files here and see what can be improved.

Replacing this logic elsewhere is a bigger overhaul though - the idpe org selector is called all over the application, so trying to unify the app to use a single org selector is its own project.

subirjolly (dismissed):

wdoconnell:
FYI: I am going to split this into two PRs since these are both key components.

wdoconnell:
Closing this PR in favor of #5910 . Will open another one for the nav changes.

PR #5896, opened 2022-09-26
chore(firstMile/CLI): remove onboardCLI feature flag

ChitlangeSahas:
Part of #4779 

Removes that flag from the UI.

Next after deleting here, i will delete it from idpe, and then config cat. 

### Checklist

### DIFF
--- a/src/shared/containers/SetOrg.tsx
+++ b/src/shared/containers/SetOrg.tsx
@@ -311,13 +311,11 @@ const SetOrg: FC = () => {
             />
           )}
           ,
-          {isFlagEnabled('onboardCLI') && (
-            <Route
-              exact
-              path="/orgs/:orgID/new-user-setup/cli"
-              component={CliWizard}
-            />
-          )}
+          <Route
+            exact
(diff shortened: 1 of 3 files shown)


hoorayimhelping (approved):

TCL735 (approved):

src/shared/containers/SetOrg.tsx
```diff
@@ -311,13 +311,11 @@ const SetOrg: FC = () => {
             />
           )}
           ,
-          {isFlagEnabled('onboardCLI') && (
```
wdoconnell, line 314: This is getting released to OSS as well, correct?

ChitlangeSahas: Just like other flows, I think so!

wdoconnell: Looks good!  I don't think this needs any changes then - just wanted to make sure this doesn't need a `CLOUD` check.

wdoconnell (approved):

PR #5905, opened 2022-09-27
feat(onboarding): scroll next page into view if previous page was scrolled

### LINKED ISSUE #5452: Bug - All Onboarding Wizards - Navigation - Scroll to top after clicking Next
Currently when navigating through the wizards, clicking `Next` takes the user to the midpoint of the next page. Users should be put at the top of the next page so they do not have to scroll.

hoorayimhelping:
Closes #5452 

Adds a function that is called when `Next`, `Previous` or the Subway Nav buttons are clicked, which scrolls the `h1` element at the top of each page into view.

Since setting state and re-rendering is an asycnrhonous operation, the function sets a timeout of 0 seconds so that the function call is added to the next run of the event queue, to let react render the next component. **This solution is a bit hacky and might end up causing weird race conditions.** This is why it has been hidden behind the feature flag `firstMileScrollTop`

https://user-images.githubusercontent.com/146112/192382849-03734be8-0346-4838-b0a1-f7ffcdcf82b4.mov

### Checklist

### DIFF
--- a/src/homepageExperience/utils.ts
+++ b/src/homepageExperience/utils.ts
@@ -1,5 +1,7 @@
 import {IconFont} from '@influxdata/clockface'
 
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
 export const HOMEPAGE_NAVIGATION_STEPS = [
   {
     name: 'Overview',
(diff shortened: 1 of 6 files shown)


src/homepageExperience/utils.ts
```diff
@@ -100,3 +102,18 @@ export const HOMEPAGE_NAVIGATION_STEPS_ARDUINO = [
...
   },
 ]
+
+// Each onboarding page in the wizard has a single h1 at the top of it. Attempt to scroll it into view smoothly
+// Set a timeout of 0 so that this function call gets run after react has a had a chance to update state and
+// re-render on the main thread.
+export const scrollNextPageIntoView = () => {
+  if (isFlagEnabled('firstMileScrollTop')) {
+    setTimeout(() => {
+      document.querySelector('h1').scrollIntoView({
```
wdoconnell, line 112: I was going to suggest a classname, but it looks like these `h1s` don't have 'em.

wdoconnell (approved):

src/homepageExperience/utils.ts
hoorayimhelping: probably less risky, but it would require changing every page in every first mile flow to add a class to the `h1`s

PR #5915, opened 2022-09-27
feat(universalLogin): redirect to QuartzLoginURL on invalid session

ChitlangeSahas:
Part of https://github.com/influxdata/idpe/issues/15477

### Checklist


src/onboarding/containers/LoginPage.tsx
```diff
@@ -56,7 +57,10 @@ export const LoginPage: FC = () => {
   } else {
     if (isFlagEnabled('universalLogin')) {
       if (CLOUD) {
-        const url = new URL(`${CLOUD_QUARTZ_URL}`).href
+        let url
+        getUiproxyQuartzLoginUrl({}).then(response => {
+          url = response.data
```
hoorayimhelping, line 62: the redirect will also need to go inside the `then`'s callback function. the `then` isn't like an `await` where the code after it gets run after it resolves, this code will run in this order:

1. `getUiproxyQuartzLoginUrl` makes a request
1. `console.warn` is fired
1. `window.location.replace(url)` is called
1. `url gets set to response.data`

hoorayimhelping (changes requested):

src/onboarding/containers/LoginPage.tsx
ChitlangeSahas: Ahhh I see! Okay i'll fix that!

hoorayimhelping (dismissed):
this is fine for now and for testing purposes, but we'll want to clean up the call to `getUiproxyQuartzLoginUrl` - maybe put it in its own api layer and handle any failures better

hoorayimhelping (approved):

hoorayimhelping (approved):

hoorayimhelping (approved):

PR #5916, opened 2022-09-27
chore: update the timerange

### LINKED ISSUE #5864: Updates to time range dropdown
* Can we make only the “Time range options” section scrollable and keep the rest of the dropdown a static?
* Since the timezone dropdown gets cut off I think we should make the bottom margin of the Time Zone dropdown 24px instead of 8px
* I spoke with Gary about this, but for the time range button we'd like to see what it would look like to make it dynamic and expand the button to show the full time range when a user has made that selection. Otherwise the button should remain the same width it is today. 
* Can we sync the calendar to match changes to the dates in the custom time range input fields?

asalem1:
Closes #5864

This PR updates some of the UX around the new time range dropdown. Most of the diff here is making the menu its own component. One of the reasons for this is that we were previously mounting the entire UI together, and so we didn't get reactivity benefits from the useEffect hook that we do know. 

### Checklist

### DIFF
--- a/src/shared/components/dateRangePicker/NewDatePicker.tsx
+++ b/src/shared/components/dateRangePicker/NewDatePicker.tsx
@@ -14,6 +14,7 @@ import {
   InputLabel,
   InputType,
   QuestionMarkTooltip,
+  Icon,
 } from '@influxdata/clockface'
 import ReactDatePicker from 'react-datepicker'
 import {PersistanceContext} from 'src/dataExplorer/context/persistance'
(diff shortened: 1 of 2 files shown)


drdelambre (dismissed):

src/timeMachine/components/Vis.tsx
```diff
@@ -95,7 +95,7 @@ const TimeMachineVis: FC<Props> = ({
       !!giraffeResult.table.length)
 
   useEffect(() => {
-    if (viewProperties.hasOwnProperty('colors')) {
+    if (viewProperties.hasOwnProperty('colors') && !isViewingRawData) {
```
appletreeisyellow, line 98: Just curious, how is this related to time range?

src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -220,9 +229,187 @@ const DatePicker: FC = () => {
...
+              <Input
+                className="date-picker__input"
+                onChange={handleSetStartDate}
+                status={
+                  inputStartErrorMessage === NBSP
+                    ? ComponentStatus.Default
+                    : ComponentStatus.Error
+                }
+                type={InputType.Text}
+                value={inputStartDate}
```
appletreeisyellow, line 298:
```suggestion
                value={inputStartDate || ''}
```
Just an edge case, there is a dev warning when click on the time range picker before selecting any bucket. Adding an empty string should remove the dev warning

src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -220,9 +229,187 @@ const DatePicker: FC = () => {
...
+              <Input
+                className="date-picker__input"
+                onChange={handleSetEndDate}
+                status={
+                  inputEndErrorMessage === NBSP
+                    ? ComponentStatus.Default
+                    : ComponentStatus.Error
+                }
+                type={InputType.Text}
+                value={inputEndDate}
```
appletreeisyellow, line 318:
```suggestion
                value={inputEndDate || ''}
```
Same as above

src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -138,14 +144,17 @@ const DatePicker: FC = () => {
       startInput = `${dateParts} ${timeParts}`
     }
     if (endInput instanceof Date) {
+      const endDate = new Date(endInput)
+      // this sets the time to the end of the selected end day
+      endDate.setMinutes(endDate.getMinutes() - 1)
```
appletreeisyellow, line 149: Nice!

appletreeisyellow (commented):
I left two comments related to dev warnings in an edge case. Not a big deal. It would be good to fix it in this PR so that the dev console is clean for future development

appletreeisyellow (approved):

PR #5924, opened 2022-09-28
chore: actually resolve the UI issue

asalem1:
Turns out that remocal disconnected on me without me noticing, and so everything I was seeing in local development was a false positive that didn't address the styling issue we're seeing with the timerange. This actually resolves the issue since (upon further inspection of production), it appears as if there was remaining styling that was in place that snuck in when I tried to update the UI to be more on point with the designs.

Here's how it looks:

Screen Shot 2022-09-28 at 1 57 43 PM


drdelambre (approved):

rockstar (commented):
Is there a way we can test this?

PR #5917, opened 2022-09-27
revert: commit removing quartzMe references from getOrganizations

wdoconnell:
This reverts commit b838e0729387157d7b0de533ca86d5fe9c305c93, which would have, had it hit prod, caused a re-render loop for free users on the account/billing page.

Screen Shot 2022-09-27 at 5 45 32 PM

https://user-images.githubusercontent.com/91283923/192641591-851ae4da-645a-4b78-848d-89430a554d44.mov

https://user-images.githubusercontent.com/91283923/192642741-fe150057-7986-47a3-a063-3d697e5d0f8f.mov

### Checklist

Authors and Reviewer(s), please verify the following:

- [X] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [X] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [X] Documentation updated or issue created (provide link to issue/PR)
- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [X] Feature flagged, if applicable

PR #5938, opened 2022-09-29
fix: split up quartz-identity redux loading state into three states

### LINKED ISSUE #5936: Multi Org Phase 1 - break quartz identity redux loading states into multiple pieces
quartzIdentity's redux loading `status` needs to be broken into three separate statuses - one for identity, one for account billing status, and one for organization details. We can't complete the transition from having the `PageSpinner` component in GetOrganizations rely on quartzIdentity loading `status` until subsidiary components have their _own_ status that is not interfering with the status of the parent component.

See [UI 5917](https://github.com/influxdata/ui/pull/5917) for more complete analysis of the need for this, and why this blocks removal of the remaining quartzMe code (even though we're no longer pulling data from that endpoint).

wdoconnell:
Closes #5936 

Splits the single redux loading status for `quartzIdentity` into three separate statuses. 

The motivation for this is more fully explained in UI #5917. But basically, [having a single loading status ](https://stackoverflow.com/questions/65849464/react-and-redux-proper-way-to-store-loading-state-in-redux) for all identity-related data increases the risk of a render loop in child components that update aspects of the identity state through calls to other endpoints.

The loading state shouldn't be focused on whether we're in the process of changing the state maintained for the user. Instead, it should answer the question: Did we already get a good response from the endpoint we needed to hit _to load this specific part of the app_ - or do we still need to make another call? 

This allows for a more explicit check in the account/billing and organization settings pages on whether we got the response we expected, rather than an _inferential_ check based on the current value of a property used by the page, which can be intentionally `null`, `undefined`, or `''`.

- identityStatus: User can view `<App />` after quartz/identity loads.
- billingStatus: User can view the Account/Billing tab, after the billing provider loads.
- orgDetailsStatus: User can view the Organization settings page, after the user's organization details (e.g., `regionName`, cloud `provider`) load.

App States after Change
--
https://user-images.githubusercontent.com/91283923/193295803-c1252f3c-e1fe-4370-933d-9c3e27f19807.mov

### Checklist

### DIFF
--- a/src/identity/reducers/index.ts
+++ b/src/identity/reducers/index.ts
@@ -1,7 +1,7 @@
 // Libraries
 import produce from 'immer'
 
-import {CurrentIdentity} from '../apis/auth'
+import {CurrentIdentity} from 'src/identity/apis/auth'
 import {RemoteDataState} from 'src/types'
 
 // Actions
@@ -10,7 +10,9 @@ import {
   SET_QUARTZ_IDENTITY,
   SET_QUARTZ_IDENTITY_STATUS,
   SET_CURRENT_BILLING_PROVIDER,
+  SET_CURRENT_BILLING_PROVIDER_STATUS,
   SET_CURRENT_ORG_DETAILS,
+  SET_CURRENT_ORG_DETAILS_STATUS,
   SET_CURRENT_IDENTITY_ACCOUNT_NAME,
 } from 'src/identity/actions/creators'
 
@@ -37,7 +39,11 @@ export const initialState: CurrentIdentity = {
     paygCreditStartDate: '',
     isUpgradeable: false,
   },
-  status: RemoteDataState.NotStarted,
+  loadingStatus: {
+    identityStatus: RemoteDataState.NotStarted,
+    billingStatus: RemoteDataState.NotStarted,
+    orgDetailsStatus: RemoteDataState.NotStarted,
+  },
 }
 
 export default (state = initialState, action: Actions): CurrentIdentity =>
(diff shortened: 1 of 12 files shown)


src/identity/reducers/index.ts
```diff
@@ -37,7 +39,11 @@ export const initialState: CurrentIdentity = {
     paygCreditStartDate: '',
     isUpgradeable: false,
   },
-  status: RemoteDataState.NotStarted,
+  loadingStatus: {
+    identityStatus: RemoteDataState.NotStarted,
+    billingStatus: RemoteDataState.NotStarted,
+    orgDetailsStatus: RemoteDataState.NotStarted,
+  },
```
hoorayimhelping, line 46: how do these properties map to the existing state? `identityStatus` seems to be related to `currentIdentity.user`, `orgDetailsStatus` seems to map to `currentIdentity.org`, but what is the relationship between `billingStatus` and `account`? Should these statuses should be part of the object they're meant to be a status for, rather than a separate object?

The thing I'm thinking about is that these statuses aren't actually part of the domain we're modelling. The loading status is a tool we use for mechanical reasons, not domain-modelling reasons; it's not related to the actual concept of an account or billing details.

hoorayimhelping (commented):
Looks good to me; just thinking about the shape of the data and how it should be organized

src/identity/reducers/index.ts
wdoconnell: - `identityStatus` maps to the user's `currentIdentity`, retrieved from `/identity. 
- `billingStatus` maps to the `billingProvider` property, which is retrieved from /quartz/accounts/:accountId.  There's other stuff that comes back from this endpoint, but we don't use it at this point. So arguably this should be 'accountDetailsStatus' or something like that.
- `orgDetailsStatus` maps to the organization details retrieved from /quartz/accounts

I think it's right that this isn't related to an account or billing necessarily - the loading state imho is really a check for 'is it safe to load this page yet, or is there something else I need to do?' Maybe there's a better name here related to which pages we can load.

As far as keeping the status in the object - that's one possibility. However, I think it's safer to keep this in a separate object. Otherwise, changes to the loading status are going to trigger the selectors that rely on the retrieved data itself. 

I.e. - when you change the state to `LOADING` it'll trigger an unnecessary re-render of any page that's _using_ that data before the new data is retrieved.

hoorayimhelping (approved):

Issue #5946, opened 2022-09-30
Data explorer / Customize panel has some inconsistent spacing


ChitlangeSahas:
We can fix the inconsistent font and spacing issues in the data explorer / customize.

Examples: 

1. First heading "Data" has a large upper margin. Makes it feel like some content failed loading in that space. 

<img width="490" alt="Screen Shot 2022-09-29 at 11 27 31 PM" src="https://user-images.githubusercontent.com/18511823/193204892-5b8089a0-ebb7-4ce0-abfb-a9697c4a12ee.png">

2. The toggle text changes color on toggling. This makes it feel like the option is disabled. 

<img width="391" alt="Screen Shot 2022-09-29 at 11 28 48 PM" src="https://user-images.githubusercontent.com/18511823/193204960-cb73f83a-98c6-4f3a-9f3e-b4d7a07078b3.png">

3. Shade area below graph is alien looking.

<img width="433" alt="Screen Shot 2022-09-29 at 11 29 43 PM" src="https://user-images.githubusercontent.com/18511823/193205123-f0735cd8-b50d-4b64-b613-199b7bff9dd2.png">

4. Orientation text needs to match other sub-headings

<img width="322" alt="Screen Shot 2022-09-29 at 11 30 26 PM" src="https://user-images.githubusercontent.com/18511823/193205158-4ba50b14-d331-4695-b80f-8232756f7c9f.png">

PR #5947, opened 2022-09-30
fix(dataExplorer): inconsistent spacing and component usage

ChitlangeSahas:
Closes https://github.com/influxdata/ui/issues/5946

Fixes all the UI issues noted in the above issue. 

Video: 

https://user-images.githubusercontent.com/18511823/193205405-c8ca38ee-f6d2-4665-a24f-5698a5c0ade7.mov

### Checklist


src/visualization/types/Band/options.tsx
```diff
@@ -174,7 +174,9 @@ const BandViewOptions: FC<Props> = ({properties, results, update}) => {
           widthMD={Columns.Six}
           widthLG={Columns.Four}
         >
-          <h5 className="view-options--header">Data</h5>
+          <h5 className="view-options--header" style={{marginTop: 0}}>
```
hoorayimhelping, line 177: this element has a classname, let's apply the style to that classname, rather than as an inline style. This component will be rendered over a graph, so we should do everything we can to minimize performance issues like inline style re-renders. This is doubly true since this style looks like it's being applied to every `h5` optiosn header

hoorayimhelping (changes requested):

src/timeMachine/components/ViewOptions.scss
```diff
@@ -20,6 +20,12 @@
   }
 }
 
+.view-options-container {
+  h5:first-child {
+    margin-top: 0;
+  }
+}
+
```
ChitlangeSahas, line 28: @hoorayimhelping , let me know if this works!

PR #5948, opened 2022-09-30
fix: set a default when there is a row

asalem1:
This PR adds a default condition for the number of tables in a given data set based on whether there is data and the calculated table number is 0. Besides that, it cleans up the original implementation to have those calculations happen after the hook and initial return so that we reduce the overhead of those calculations until after the table has rendered. Finally, the Region pipe has been updated to reflect the new tools cluster and the post-production url to allow for querying across different regions for those particular cases


wiedld (commented):
I have an alternative approach, which gets the full table numbers without storing in memory). Still need to performance test. Can we hold on this until the end of Monday, so we can compare these two approaches?

wiedld:
Still have a bit more to do on this one. But the draft (and the direction taken) is here:
https://github.com/influxdata/ui/pull/5979

PR #5967, opened 2022-10-03
chore: transition nav components to identity data shape

wdoconnell:
Connects #4821 

Removes references to the old quartzMe data shape from the left nav bar, and replaces them with references to the quartzIdentity data shape (which comes from the same source).

Nav After Changes - User Not An Operator
--
https://user-images.githubusercontent.com/91283923/193899096-7efae302-8cbd-4b97-a739-f63078b263f3.mov

Nav After Changes - User Is An Operator
--
https://user-images.githubusercontent.com/91283923/193899119-3dbfab38-12cd-4947-b5f5-c31916c0079d.mov

### Checklist


src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -36,7 +40,8 @@ export interface NavItem {
 export const generateNavItems = (): NavItem[] => {
   const state: AppState = getStore().getState()
   const orgID = state?.resources?.orgs?.org?.id ?? ''
```
wdoconnell, line 42: We can't remove references to the IDPE org id from NavItem, since those are necessary to build the nav bar in OSS.

src/pageLayout/containers/TreeNav.Helpbar.test.tsx
```diff
@@ -2,17 +2,46 @@ import React from 'react'
 import {screen, fireEvent, cleanup, waitFor} from '@testing-library/react'
 import {jest} from '@jest/globals'
 
+// Mocks
```
wdoconnell, line 5: The mock being used in this test was outdated because it required references to `quartzMe`.

src/pageLayout/containers/TreeNav.tsx
```diff
@@ -1,8 +1,7 @@
 // Libraries
 import React, {FC, useContext, useEffect, MouseEvent} from 'react'
 import {Link, useLocation} from 'react-router-dom'
-import {useSelector} from 'react-redux'
-import {connect, ConnectedProps} from 'react-redux'
+import {useDispatch, useSelector} from 'react-redux'
```
wdoconnell, line 5: Right now `TreeNav` is using a mix of the `connect` and hooks apis from react-redux. 

Since we're already using `useSelector` here we might as well use `useDispatch` instead of the old map-dispatch-to-props and map-state-to-props paradigm.

src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -33,11 +36,10 @@ export interface NavItem {
   activeKeywords: string[]
 }
 
-export const generateNavItems = (): NavItem[] => {
-  const state: AppState = getStore().getState()
-  const orgID = state?.resources?.orgs?.org?.id ?? ''
-  const quartzMe = state?.me?.quartzMe ?? null
-
```
wdoconnell, line 40: Ideally we should not [be calling getStore()](https://stackoverflow.com/questions/34455996/is-using-getstate-in-redux-an-anti-pattern) in a utility function if we can help it (which we can here).

src/pageLayout/containers/TreeNav.tsx
```diff
@@ -262,15 +267,4 @@ const TreeSidebar: FC<ReduxProps> = ({
...
-
-const mdtp = {
-  showOverlay,
-  dismissOverlay,
-}
-
-const connector = connect(mstp, mdtp)
-
-export default connector(TreeSidebar)
+export default TreeSidebar
```
wdoconnell, line 270: I'll kill this `export default` alongside any PR comments.

hoorayimhelping (dismissed):

hoorayimhelping (approved):

PR #5969, opened 2022-10-03
fix(universalLogin): return null instead of undefined to prevent render errors

ChitlangeSahas:
Part of https://github.com/influxdata/idpe/issues/15477

React needs us to return something, instead of `undefined`. `null` is a valid return. 

### Checklist


TCL735 (approved):

hoorayimhelping (approved):
big if true

PR #5971, opened 2022-10-03
chore: remove quartzMe references from writedata components

wdoconnell:
Connects #4821 

Replaces references to the quartzMe state in the `writeData` components with quartzIdentity state.

Video
--
https://user-images.githubusercontent.com/91283923/193696850-659dca39-5c8d-4346-bd74-3e021a4ce40b.mov

https://user-images.githubusercontent.com/91283923/193697329-ce65a6b5-433f-49f6-a1d9-5de838a6c49d.mov

### Checklist

### DIFF
--- a/src/writeData/components/PluginCreateConfiguration/Footer.tsx
+++ b/src/writeData/components/PluginCreateConfiguration/Footer.tsx
@@ -24,6 +24,7 @@ import {PluginConfigurationStepProps} from 'src/writeData/components/AddPluginTo
 
 // Selectors
 import {getAll} from 'src/resources/selectors'
+import {selectCurrentAccountType} from 'src/identity/selectors'
 
 // Utils
 import {CLOUD} from 'src/shared/constants'
@@ -162,7 +163,8 @@ const mstp = (state: AppState) => {
       steps: {orgID},
     },
   } = state
-  const accountType = state?.identity?.currentIdentity?.account?.type ?? 'free'
+  const accountType = selectCurrentAccountType(state)
+
   let telegrafConfig = null
   if (telegrafConfigID) {
     const telegrafs = getAll<Telegraf>(state, ResourceType.Telegrafs)
--- a/src/writeData/components/PluginCreateConfiguration/Footer.tsx
+++ b/src/writeData/components/PluginCreateConfiguration/Footer.tsx
@@ -24,7 +24,6 @@ import {PluginConfigurationStepProps} from 'src/writeData/components/AddPluginTo
 
 // Selectors
 import {getAll} from 'src/resources/selectors'
-import {getQuartzMe} from 'src/me/selectors'
 
 // Utils
 import {CLOUD} from 'src/shared/constants'
@@ -163,7 +162,7 @@ const mstp = (state: AppState) => {
       steps: {orgID},
     },
   } = state
-  const accountType = getQuartzMe(state)?.accountType ?? 'free'
+  const accountType = state?.identity?.currentIdentity?.account?.type ?? 'free'
   let telegrafConfig = null
   if (telegrafConfigID) {
     const telegrafs = getAll<Telegraf>(state, ResourceType.Telegrafs)
--- a/src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
+++ b/src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
@@ -36,9 +36,12 @@ import {
   StepsStatus,
 } from 'src/types'
 
-// Utils
+// Selectors
 import {getAll} from 'src/resources/selectors'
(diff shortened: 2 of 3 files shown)


src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
```diff
@@ -57,7 +60,7 @@ const CreateSubscriptionPage: FC = () => {
   const {formContent, saveForm, updateForm, loading} = useContext(
     SubscriptionCreateContext
   )
-  const quartzMe = useSelector(getQuartzMe)
+  const identity = useSelector(selectCurrentIdentity)
   const buckets = useSelector((state: AppState) =>
```
wdoconnell, line 64: This is a cloud-only component, so there is no need for `CLOUD` guards.

The path is: 
- `SetOrg` contains a `CLOUD` and `isFlagEnabled('subscriptionsUI')` check for whether the `CreateSubScriptionForm` can be loaded.
- `CreateSubscriptionForm` is actually a lazy import of `CreateSubscriptionPage`
- `CreateSubscriptionPage` is a default export parent component consisting of several components from `CreateSubcriptionPage`

There are no other references in the UI to `CreateSubscriptionPage`

src/writeData/components/PluginCreateConfiguration/Footer.tsx
```diff
@@ -163,7 +162,7 @@ const mstp = (state: AppState) => {
       steps: {orgID},
     },
   } = state
-  const accountType = getQuartzMe(state)?.accountType ?? 'free'
+  const accountType = state?.identity?.currentIdentity?.account?.type ?? 'free'
```
wdoconnell, line 165: I've changed the logic here because I do not think there should be a need to directly call the selector function.

Either we call the selector function using `useSelector` in the React component, or we directly access the `state` object passed into `matchStateToProps` as an argument.

wdoconnell: This is not a cloud-only component. `accountType` will be empty string by default. 

As before, there is no need for an additional `CLOUD` guard because `accountType` is already guarded by, and only used, within a `CLOUD` check on line 69.

```
  if (CLOUD && accountType === 'free') {
        config = config.replace(
          agentSettingPatternDefault,
          agentSettingPatternFreeTier
        )
      }
```

TCL735 (dismissed):

genehynson (approved):

src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
```diff
@@ -88,10 +91,10 @@ const CreateSubscriptionPage: FC = () => {
   useEffect(() => {
     event(
       'visited creation page',
-      {userAccountType: quartzMe?.accountType ?? 'unknown'},
+      {userAccountType: identity.account.type || 'unknown'},
```
wdoconnell, line 94: Due to differences in the reducers we need to use logical OR here. The nullish coalescing (??) operator was used because `quartzMe` had no default reducer state, so the fallback was triggered when `quartzMe` was undefined. 

quartzIdentity has a legit reducer state, and the default value for account.type is `''`. So now, we need the fallback to happen when `account.type` is falsy in any form, not `undefined`.

helenosheaa (dismissed):

src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
hoorayimhelping: >quartzIdentity has a legit reducer state, and the default value for account.type is `''`

It looks like it's [`free`](https://github.com/influxdata/ui/blob/master/src/identity/reducers/index.ts#L37). Seems like the whole fallback isn't necessary? Also, I think we shouldn't have different fallbacks in different parts of the app. In the above component, the fallback was `'free'` but here it's `'unknown'`. That seems like it has the potential to be really confusing to both people maintaining the codebase and to users

hoorayimhelping (commented):
Let me know if the thoughts I had made any sense

src/writeData/components/PluginCreateConfiguration/Footer.tsx
wdoconnell: You're right on both counts - let me revise this and I'll rebase and push up again.

src/writeData/subscriptions/components/CreateSubscriptionPage.tsx
wdoconnell: Yes, you're right - and I think this can be resolved by your other comment. Since `free` is the default, sending this in the event name is just going to create a mismatch between the eventing data and what's in state.

hoorayimhelping (approved):

genehynson (approved):

PR #5979, opened 2022-10-04
feat(5901): perform full table counts, even when truncated data.

### LINKED ISSUE #5901: Incorrect # of Rows Displayed
When querying large datasets using the new Script Editor, a situation occurred where the number of tables that were shown in the display was zero, even though there always has to be at least 1 and the actual data did have tables included.  See slack thread:  https://influxdata.slack.com/archives/C030G5783M3/p1664228908454359

The counts should match what is in the actual payload.

wiedld:
Closes #5901

When viewing larger datasets, we are truncating the data. This means that:
* the total number of tables are shown in the UI to decrease, even if it's increased in the network payload
* we are also sometimes showing `0 tables`, when we clearly have >1 table.

Desired:
* show the accurate total # of tables, even if not parsing all of the rows into view data
* give a message to indicate that the # of rows are truncated
* split up the yielding of results:
  * allow the yielding of truncated, parsed table results
  * then later, the yielding of the full calculation of query metadata 

Scope of code impact:
* it only impacts the shared query context:
  * notebooks
  * new Qx Builder 

## Vid

In QxBuilder (with 30 days of data):

https://user-images.githubusercontent.com/10232835/193963885-8a4c8a03-cdab-49c6-be15-180dee506a74.mov

In notebooks (with 2 days of data):

https://user-images.githubusercontent.com/10232835/193963894-4eeb9fae-6653-47fc-83dc-2fcf82b40547.mov

## Checklist
- [x] ~Feature flagged, if applicable~. N/A

### DIFF
--- a/cypress/e2e/shared/fluxQueryBuilder.test.ts
+++ b/cypress/e2e/shared/fluxQueryBuilder.test.ts
@@ -1,4 +1,5 @@
 import {Organization} from '../../../src/types'
+const path = require('path')
 
 const DEFAULT_SCHEMA = {
   bucket: null,
@@ -169,6 +170,14 @@ describe('Script Builder', () => {
             .should('not.contain', 'Truncated')
             .contains(`${rowCnt} rows`)
         }
+
+        cy.log('can download csv data')
+        cy.getByTestID('data-explorer--csv-download').should('exist').click()
+        const downloadsFolder = Cypress.config('downloadsFolder')
+        const filename = path.join(downloadsFolder, 'influx.data.csv')
+        cy.readFile(filename, {timeout: 15000}).should('have.length.gt', 500)
+        /// TODO: wiedld (5 Oct 2022) -- check for entire dataset is download.
+        /// after fix this: https://github.com/influxdata/ui/issues/5994
       }
 
       it('will return the complete dataset for smaller payloads', () => {
--- a/cypress/e2e/shared/fluxQueryBuilder.test.ts
+++ b/cypress/e2e/shared/fluxQueryBuilder.test.ts
@@ -104,9 +104,9 @@ describe('Script Builder', () => {
         cy.wait(1200).then(() => {
           cy.reload()
           cy.getByTestID('flux-sync--toggle')
+          cy.getByTestID('flux-editor', {timeout: 30000})
         })
       })
-
       cy.get('@org').then(({id}: Organization) => {
         cy.intercept('POST', `/api/v2/query?orgID=${id}`, req => {
           const {extern} = req.body
@@ -126,7 +126,6 @@ describe('Script Builder', () => {
     })
 
     it('will allow querying of different data ranges', () => {
-      cy.getByTestID('flux-editor', {timeout: 30000})
       selectSchema()
       confirmSchemaComposition()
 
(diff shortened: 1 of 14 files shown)


wiedld:
Did a performance assessment. We are now updating the code to fork the readable stream into 2, in order to separately return the truncated table data while still calculating the table metadata totals.

Here is the summary outcome:
* used two data sets, 3 time ranges, n=10 at each datapoint. Avg & Stdev. Checked heap usage diff and time diffs.
* example of dataset A:
    * has 24 tables. Always, even when only 1 hr of data.
    * has total of 68785 rows over 30 days.
* BEFORE:
   * truncated data stays at constant time (ofc), based on truncated read. 
   * but table count gets more off, because not reading all the data. 
   * e.g. dataset # 1:
      * table count dropped as timeRange increased (because more data truncated).
      * time to read stayed constant. 17-20ms 
* AFTER:
   * all data must be read -- (even if not stored) since can have tables out of order.
       * this means linear time.
   * dataset 1:
      * correctly reports 24 tables, always.
      * but time to read increase based on more rows read.
      * time increased from 201ms (with 1 hr of data), to 2.6 seconds (with 30 days of data).

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -104,9 +104,9 @@ describe('Script Builder', () => {
         cy.wait(1200).then(() => {
           cy.reload()
           cy.getByTestID('flux-sync--toggle')
+          cy.getByTestID('flux-editor', {timeout: 30000})
```
wiedld, line 107: Pulled this flux-editor check into the `beforeEach()`, within each separate test context.
e.g. Results display tests (added here), schema browser tests (added below), and schema composition test (already added).

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -438,10 +523,9 @@ describe('Script Builder', () => {
       })
 
       it('should clear the editor text and schema browser, with a new script', () => {
-        cy.getByTestID('flux-editor', {timeout: 30000})
-
         cy.log('modify schema browser')
         selectSchema()
+        confirmSchemaComposition()
```
wiedld, line 528: reduce some flaky

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -453,7 +537,9 @@ describe('Script Builder', () => {
         cy.getByTestID('overlay--container')
           .should('be.visible')
           .within(() => {
-            cy.getByTestID('flux-query-builder--no-save').click()
+            cy.getByTestID('flux-query-builder--no-save')
+              .should('be.visible')
+              .click()
```
wiedld, line 542: reduce some flaky

src/dataExplorer/components/ResultsContext.tsx
```diff
@@ -58,6 +65,16 @@ export const ResultsProvider: FC = ({children}) => {
     } as SimpleTableViewProperties,
   })
 
+  const handleSetResult = useCallback(
+    (newResult: Partial<FluxResult>) => {
+      setResult({
```
wiedld, line 70: `setResult()` is called throughout codebase with the fromFlux parse function outcome. Don't allow it to overwrite the metadata.

src/shared/utils/fluxData.ts
```diff
@@ -0,0 +1,135 @@
...
+        nextLine()
+        break
+    }
+  }
+  return maxTableIdx
+}
+
+export async function* fluxDataReader(
+  responseBody: ReadableStream<Uint8Array>
+): AsyncGenerator<FluxResult, FluxResultComplete> {
```
wiedld, line 84: The return signature is very intentional. Such that the consumers of this function do not need to care if the data was truncated (with 2 yields) or not (with only 1 yield). That is why the yield signatures are identical except for the extended `FluxResultComplete` type includes the metadata.

wiedld:
Per request, we decided to change the product requirements. See the original ticket for final decision. Closing this PR, to re-create new one (is cleaner for merge queue commit message).

PR #5985, opened 2022-10-04
chore: move src/clockface/components -> src/shared/components

### LINKED ISSUE #3492: Tech Debt: Swap out internal Clockface components with @influxdata/clockface
In order to reduce some drift in styling and within our codebase, we should consolidate and replace any lingering clockface components that exist within the UI repo to `@influxdata/clockface`. 

### Why is the current structure a problem?

As it stands, we are using internal clockface components for part of the UI when we could potentially swap that out to our own internal library that's currently maintained. The issues that surround the current setup is potential code drift, styling differences, and a bloated codebase where we maintain similar code in two different places

### How to remediate the issue

We should audit the `src/clockface` components to see what needs to be ported over to the `@influxdata/clockface` library, then slowly deprecate / delete the existing components within `src/clockface` and replace their usages with ones in the `@influxdata/clockface` library.

### Definition of Done

Being able to delete the `src/clockface` components without having any issues in the UI

TCL735:
Closes #3492 

And finally, `src/clockface` can be put to rest.


src/style/chronograf.scss
```diff
@@ -6,143 +6,146 @@
 // Modules
 @import 'modules';
 
-// Clockface UI Kit
-@import 'src/clockface/styles';
-
 // Components
```
TCL735, line 9: This section is alphabetized.

hoorayimhelping: as long as there aren't any implicit dependencies on things being loaded in the proper order (i.e. if one stylesheet depends on another stylesheet being imported before it can work) this will work

hoorayimhelping (approved):
Love the removal of the default exports. Can you make sure to check out the look of pages to make sure styling isn't off to make sure there aren't implicit needs in the loading order of the stylesheets? There shouldn't be, but who knows

src/style/chronograf.scss
TCL735: i'll take a quick visual pass through the pages.

Honestly, if any are relying on the order, it should be considered a bug. And i'm hoping to have them surface so we can fix them. But as i said, i will take a quick visual pass to see if anything is glaringly out of place.

hoorayimhelping (approved):

hoorayimhelping:
>Honestly, if any are relying on the order, it should be considered a bug. And i'm hoping to have them surface so we can fix them. But as i said, i will take a quick visual pass to see if anything is glaringly out of place.

that's actually a fair point. I'll leave you to it since you're on bug duty

PR #5992, opened 2022-10-05
fix: prevent decimal places customization from crashing the visualization

### LINKED ISSUE #5977: Bug: changing decimal places causes visualizations to crash
Changing the decimal places option under "Customize" causes the visualization to temporarily disappear with an error message. This is not ideal, because a user must delete the previous input in order to change it to a new input -- this is the most common way a user would interact with decimal places.

Here is an example:

https://user-images.githubusercontent.com/10736577/193704450-82e2ff4c-b39a-4f81-b91b-0328be75ad53.mp4

**Proposal:**
Changing the decimal places should default to a sane number in order to keep the visualizations stable. Or better yet: keep the decimal places exactly the same until the user has updated the decimal places with valid input.

TCL735:
Closes #5977  

- adds a new component `DecimalPlaces` to handle the customization of decimal places
- prevents the visualization from crashing while changing the input for decimal places
- updates the visualization only upon valid input

Note: `DecimalPlaces` was added to Table's options as a placeholder to replace the customization's previous incarnation. Table does not actually adjust any decimal places at the moment. This visualization is slated for a major design overhaul or possibly deletion (and folded into a different table-like visualization.)

https://user-images.githubusercontent.com/10736577/194127600-96bbf10d-1459-41e0-9275-d10b35a367b2.mp4

### DIFF
--- a/src/visualization/components/internal/DecimalPlaces.tsx
+++ b/src/visualization/components/internal/DecimalPlaces.tsx
@@ -1,0 +1,79 @@
+// Libraries
+import React, {FC, useState} from 'react'
+
+// Components
+import {
+  AutoInput,
+  AutoInputMode,
+  Form,
+  Input,
+  InputType,
+} from '@influxdata/clockface'
+
+// Utils
+import {convertUserInputToNumOrNaN} from 'src/shared/utils/convertUserInput'
+
+// Constants
+import {
+  MIN_DECIMAL_PLACES,
+  MAX_DECIMAL_PLACES,
+} from 'src/visualization/constants'
+
+interface DecimalPlacesProps {
+  isEnforced: boolean
+  digits: number
+  update: (obj: any) => void
+}
+
+export const DecimalPlaces: FC<DecimalPlacesProps> = (
+  props: DecimalPlacesProps
+) => {
+  const {isEnforced, digits, update} = props
+  const [decimalPlaces, setDecimalPlaces] = useState<number | null>(digits)
+
+  const setDigits = (updatedDigits: number | null) => {
+    setDecimalPlaces(updatedDigits)
+    if (!Number.isNaN(updatedDigits)) {
+      update({
+        decimalPlaces: {
+          isEnforced,
+          digits: updatedDigits,
+        },
+      })
+    }
+  }
+
+  const handleChangeMode = (mode: AutoInputMode): void => {
+    if (mode === AutoInputMode.Auto) {
+      setDigits(null)
+    } else {
+      setDigits(2)
+    }
+  }
+
+  return (
+    <Form.Element label="Decimal Places">
+      <AutoInput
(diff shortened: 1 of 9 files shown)


src/visualization/components/internal/DecimalPlaces.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const DecimalPlaces: FC<DecimalPlacesProps> = (
+  props: DecimalPlacesProps
+) => {
+  const {isEnforced, digits, update} = props
+  const [decimalPlaces, setDecimalPlaces] = useState<number | null>(digits)
+
+  const setDigits = (updatedDigits: number | null) => {
+    setDecimalPlaces(updatedDigits)
+    if (Number.isNaN(updatedDigits) === false) {
```
hoorayimhelping, line 36: Just so I'm clear, this is updating the state before the check if it's a valid digit? Why is that the behavior rather than only updating the state if `updatedDigits` isn't `NaN`.

Also, is the explicit check for `false` for readability purposes, or will `if (!Number.isNaN(updatedDigits))` produce different behavior?

src/visualization/types/Gauge/options.tsx
```diff
@@ -1,141 +1,98 @@
...
+                  update({tickSuffix: evt.target.value})
+                }}
+                placeholder="%, MPH, etc."
+              />
+            </Form.Element>
+          </Grid.Column>
+        </Grid.Row>
+        {properties.decimalPlaces && (
+          <DecimalPlaces
+            isEnforced={properties?.decimalPlaces?.isEnforced ?? false}
```
hoorayimhelping, line 75: can this be simplified to

```ts
properties?.decimalPlaces?.isEnforced === true
```

so we don't need to worry about fallback behavior?

hoorayimhelping (commented):
A couple of questions

src/visualization/types/Gauge/options.tsx
TCL735: Oh nice, yeah that is both cool & concise while also very easy to understand!

src/visualization/components/internal/DecimalPlaces.tsx
TCL735: This change prevents the visualization from crashing. Previously, immediately updating the Redux state will have the NaN or null sent back in as a prop which gets sent to Giraffe, which crashes the viz.

By having a local state, we can highlight the input (with an error border) to show the user that the number of decimal places should not be blank. (I don't think they can type anything other than a number in this particular input). And while the input is in this error state, do not update the Redux state just yet. Wait till the input is valid before updating the Redux state.

Local state saves the day by handling the bad input and only filtering the good input to Redux state, in a sense.

hoorayimhelping (approved):

PR #5998, opened 2022-10-05
feat: add basic Result Options

### LINKED ISSUE #5976: Result Options: Add basic structures in UI
To prepare for the implementation of #5367, #5761, and #5759, there are some basic work to do

- [x] Create feature flag `resultOptions` in idpe (influxdata/idpe#16007)
- [ ] Add the basic UI of "Result Options" behind this feature flag (#5998)

appletreeisyellow:
Closes #5976 

This PR adds basic structure of the Result Options in new Script Editor so that the future work of fields as columns, group, and aggregate can be done independently. 

This work is behind the feature flag `resultOptions`

## After

https://user-images.githubusercontent.com/14298407/194126338-a9ed9a52-2349-4291-9c45-3c1af64503c8.mov

### Checklist

### DIFF
--- a/src/dataExplorer/components/ResultOptions.tsx
+++ b/src/dataExplorer/components/ResultOptions.tsx
@@ -1,4 +1,4 @@
-import React, {useState} from 'react'
+import React, {FC, useState} from 'react'
 
 // Components
 import {
@@ -13,17 +13,42 @@ import SelectorTitle from 'src/dataExplorer/components/SelectorTitle'
 import './Sidebar.scss'
 
 const TOOLTIP_CONTENT = {
-  FIELDS_AS_COLUMNS: `Tooltip content for Fields as Columns`,
+  FIELDS_AS_COLUMNS: `test`,
+  GROUP: `test`,
 }
 
-const ResultOptions = () => {
-  const [fieldsAsColumnsOn, setFieldsAsColumnsOn] = useState(false)
+interface ToggleWithLabelToolTipProps {
+  active: boolean
+  onChange: () => void
+  label: string
+  tooltipContents?: string | JSX.Element
+}
+
+const ToggleWithLabelToolTip: FC<ToggleWithLabelToolTipProps> = ({
+  active,
+  onChange,
+  label,
+  tooltipContents = '',
+}) => {
+  return (
+    <FlexBox>
+      <SlideToggle active={active} onChange={onChange} />
+      <InputLabel>
+        <SelectorTitle label={label} tooltipContents={tooltipContents} />
+      </InputLabel>
+    </FlexBox>
+  )
+}
+
+const ResultOptions: FC = () => {
+  const [fieldsAsColumnsActive, setFieldsAsColumnsActive] = useState(false)
+  const [groupActive, setGroupActive] = useState(false)
 
   const fieldsAsColumns = (
     <FlexBox>
       <SlideToggle
-        active={fieldsAsColumnsOn}
-        onChange={() => setFieldsAsColumnsOn(current => !current)}
+        active={fieldsAsColumnsActive}
+        onChange={() => setFieldsAsColumnsActive(current => !current)}
       />
       <InputLabel>
         <SelectorTitle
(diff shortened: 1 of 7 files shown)


src/dataExplorer/components/Sidebar.scss
```diff
@@ -16,6 +33,68 @@
   left: 0;
   right: 0;
 
+  .flux-library-original {
+    height: calc(100% - 30px); // remove the ugly white scroll bar
```
wiedld, line 37: I believe you are talking about this scroll bar?

https://user-images.githubusercontent.com/10232835/194182331-9560cf57-6caf-4bce-a911-625eebc301a9.mov



Assuming yes...then both of the two instances of `height: calc(100% - 30px); // remove the ugly white scroll bar`, refer to the same scroll bar on the outer container. That scroll bar can be removed by hiding the overflow on that outer container. Setting `overflow-y: hidden` on the `container-right-side-bar`.

src/dataExplorer/components/Sidebar.scss
```diff
@@ -16,6 +33,68 @@
...
+    @include transparent-accordion;
+    @include accordion-header-uppercase;
+
+    .selector-title {
+      padding: $cf-space-2xs 0;
+    }
+  }
+
+  .cf-accordion--body-container {
+    height: calc(100% - 40px); // remove the ugly white scroll bar
```
wiedld, line 95: See other comment. This can be removed, I believe.

wiedld (changes requested):
Excellent work. Just 1 change requested, and 1 open ended [curiosity] question.

src/dataExplorer/components/Sidebar.scss
appletreeisyellow: Thanks! Setting the `overflow-y: hidden` on the `container-right-side-bar` does makes the code straightforward. Since there is already `overflow-x: hidden` in `container-right-side-bar`, I just combined them to be `overflow: hidden`. 

I still kept `height: 100%` without the `calc()` in the two instances to show the flux function list

appletreeisyellow: Actually, we still need the `calc()` to be there for the two instances. Otherwise, the items at the bottom of the flux library will be cut off. See video below

https://user-images.githubusercontent.com/14298407/194342694-6795511d-61c2-4ef2-b8a6-8804ce41a165.mov

src/dataExplorer/components/ResultOptions.tsx
```diff
@@ -0,0 +1,25 @@
...
+const ResultOptions: FC = () => {
+  return (
+    <Accordion className="result-options" expanded={true}>
+      <Accordion.AccordionHeader className="result-options--header">
+        Result Options
+      </Accordion.AccordionHeader>
+      <FieldsAsColumns />
+      <GroupBy />
+      <Aggregate />
+    </Accordion>
```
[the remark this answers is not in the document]
appletreeisyellow: Good point!

I thought and tired wrapping `Accordion.AccordionBodyItem` around `FieldsAsColums`, `GroupBy`, and `Aggregate` according to the example given on Clockface website.

However, it brings in extra overwriting the css styles which I think is unnecessary. If we can achieve the same effect by simply adding the body items, I don't see the benefit of wrapping the items in `Accordion.AccordionBodyItem` with extra css styles. 

It will be beneficial to wrap the items with `Accordion.AccordionBodyItem` if we want to keep most of the css styles, such as background color and paddings, see below

Accordion AccordionBodyItem

Since the design for result options items is different than the style provided by the Clockface component, that is why I take the current route. I'm open to use `Accordion.AccordionBodyItem` if you find other benefits of it.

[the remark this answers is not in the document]
wiedld: Good reasoning, since we currently don't make use of structured child elements of a typed parent component. So the tradeoff is not worth it. Thanks for explaining!

src/dataExplorer/components/Sidebar.scss
wiedld: blargh. ok then. Thanks for testing, and checking the details.

wiedld (approved):
Thanks for being so detailed focus, and checking for missing DOM nodes.

PR #6000, opened 2022-10-05
chore(CSVParser): migrate parser code into UI codebase

### LINKED ISSUE #5662: csv upload experience: translate non-annotated csv parsing into JavaScript
### Background
The telegraf codebase has a [CSV parser](https://github.com/influxdata/telegraf/blob/master/plugins/parsers/csv/parser.go) written in go that handles annotated and non-annotated CSVs. Take that code's structure, and translate it from go into JavaScript. This will require following the code, looking up library functions in go like `strings.SplitN()`, and either using the analagous JavaScript function, or writing similar code.

### Acceptance Criteria
- Translate the CSV go parser into JavaScript
   - The parser should handle annotated and un-annotated CSVs
- Test the parser with unit tests to confirm that it parses un-annotated CSVs properly

blegesse-git:
Closes #5662 

As part of DND non-annotated CSV epic, pr introduces CSV parsing code that will be integrated into the UI to allow users the ability to upload non-annotated CSVs into the UI.

`parser.ts` is a direct translation of telegraf's [CSV parser](https://github.com/influxdata/telegraf/blob/master/plugins/parsers/csv/parser.go). 
This javascript parser leverages an external library, Papa Parse to read and parse csv files.

Upcoming work utilizing this parser will be hidden behind a feature flag `uploadCSV`. 

### Checklist


src/utils/csvParser/parser.ts
```diff
@@ -0,0 +1,595 @@
+import dayjs from 'dayjs'
+import dayjsCustomParseFormat from 'dayjs/plugin/customParseFormat'
+import dayjsUTC from 'dayjs/plugin/utc'
+import dayjsTimezone from 'dayjs/plugin/timezone'
+dayjs.extend(dayjsCustomParseFormat)
+dayjs.extend(dayjsUTC)
+dayjs.extend(dayjsTimezone)
```
hoorayimhelping, line 7: before you merge this in, let's pair up and see if we can do this without adding a new dependency

blegesse-git: yup! sorry, forgot we said we'd do that first

src/utils/csvParser/parser.test.ts
```diff
@@ -0,0 +1,1544 @@
+/* eslint-disable @typescript-eslint/no-loss-of-precision */
```
hoorayimhelping, line 1: how many eslint errors are there related to loss of precision?

src/utils/csvParser/parser.test.ts
```diff
@@ -0,0 +1,1544 @@
...
+import {expect, jest} from '@jest/globals'
```
hoorayimhelping, line 2: is this import necessary or can it be done away with? it doesn't look like any of our other tests import `expect` or `jest`

src/utils/csvParser/parser.test.ts
```diff
@@ -0,0 +1,1544 @@
...
+import {CSVParser} from './parser'
+import {EOFError} from './utils'
+
+interface Metric {
```
hoorayimhelping, line 6: rather than re-creating this here and in `parser.ts`, just export it from `parser.ts` and import it here

src/utils/csvParser/parser.ts
```diff
@@ -0,0 +1,611 @@
+import {DateTime} from 'luxon'
+import {EOFError, LocalFile} from './utils'
+import {CSVReader} from './reader'
+
+// Utils
+import {getLuxonFormatString} from 'src/utils/datetime/validator'
+
+interface Metric {
```
hoorayimhelping, line 8: is it possible to reuse [`Point`](https://github.com/influxdata/ui/blob/master/src/cloud/apis/reporting.ts#L6) from the reporting section of the app?

src/utils/csvParser/parser.ts
```diff
@@ -0,0 +1,611 @@
...
+  private gotColumnNames = false
+  private gotInitialColumnNames = false
+  private remainingSkipRows = 0
+  private remainingHeaderRows = 0
+  private remainingMetadataRows = 0
+  public metadataTags: Record<string, string> = {}
+  public metadataSeparatorList: string[] = []
+  config: Config
+
+  // InitFromConfig
```
hoorayimhelping, line 160: what's this comment mean? I think it's okay to remove

src/utils/csvParser/parser.ts
```diff
@@ -0,0 +1,611 @@
...
+        this.remainingSkipRows--
+        throw EOFError
+      }
+      if (this.remainingMetadataRows > 0) {
+        this.remainingMetadataRows--
+        throw EOFError
+      }
+    }
+
+    const metrics = await this.parseCSV(line)
```
hoorayimhelping, line 214: can/should this be something like

```ts
const metrics: Metric[] = await ...
```
?

hoorayimhelping (commented):
looks great! just a couple of minor nagging thoughts - let me know what you think

src/utils/csvParser/parser.test.ts
blegesse-git: Just one error. 
precision_loss_screenshot
Instead of disabling eslint for the entire file, I can just disable it for this specific line.

src/utils/csvParser/parser.test.ts
blegesse-git: you're right. Other unit tests in the code base don't import jest 🤔. But removing the import throws a lot of errors like this: 
jest_error
I can't think of why this test file requires an import while others don't. I'll look more into this

src/utils/csvParser/parser.test.ts
blegesse-git: good call

src/utils/csvParser/parser.ts
blegesse-git: we can. But because `this.parseCSV()` is returning metrics with explicit `Metric` type, typescript is able to make that inference. But i don't see the harm in adding it here.

src/utils/csvParser/parser.ts
blegesse-git: Yes! I ended up importing it and using that instead. Had no clue it already existed in the codebase.

src/utils/csvParser/parser.test.ts
hoorayimhelping: I wonder if this is the first test to be created after we upgraded jest

hoorayimhelping (approved):

PR #6003, opened 2022-10-05
fix: dont show header on 404 page

### LINKED ISSUE #5750: Multi Org Phase 1 - When a page returns 404, we should not show the global header
If a page does not exist and we redirect to 404 page, we shouldn't show the global header.

Image

wdoconnell:
Closes #5750 

In cloud, the multi-org global header displayed on 404 pages, because it was rendered by the parent of the component containing our React Router routes (SetOrg).

This PR adds another condition for rendering the header: it won't display if the current page is the `not found` page. 

Conveniently, we already have a redux state for this (which is used in dashboards). So detecting the 'not found' page simply requires updating (and cleaning up) that state when the user hits the `Not Found` component.

Before
--

https://user-images.githubusercontent.com/91283923/194159129-70806657-b8d7-410f-8601-446805c046d2.mov

After
--

https://user-images.githubusercontent.com/91283923/194159141-8626a047-8675-4ac2-870e-2758438ba804.mov

### Checklist

### DIFF
--- a/src/shared/components/NotFound.tsx
+++ b/src/shared/components/NotFound.tsx
@@ -13,10 +13,13 @@ import {
   Panel,
 } from '@influxdata/clockface'
 import React, {useState, FC, useCallback, useEffect, useRef} from 'react'
-import {useSelector} from 'react-redux'
+import {useDispatch, useSelector} from 'react-redux'
 import {useHistory, useLocation} from 'react-router-dom'
 import {getOrg} from 'src/organizations/selectors'
 import {getOrg as fetchOrg} from 'src/organizations/apis'
+
+// Actions
+import {setCurrentPage} from 'src/shared/reducers/currentPage'
 
 // Utils
 import {buildDeepLinkingMap} from 'src/utils/deepLinks'
@@ -135,6 +138,8 @@ const NotFound: FC = () => {
   const location = useLocation()
   const history = useHistory()
   const reduxOrg = useSelector(getOrg)
+  const dispatch = useDispatch()
+
   const org = useRef<Organization>(reduxOrg)
 
   const handleDeepLink = useCallback(async () => {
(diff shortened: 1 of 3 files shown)


src/shared/components/NotFound.tsx
```diff
@@ -169,6 +174,13 @@ const NotFound: FC = () => {
     }
   }, [handleDeepLink])
 
+  useEffect(() => {
+    dispatch(setCurrentPage('not found'))
+    return () => {
+      dispatch(setCurrentPage('not set'))
```
blegesse-git, line 180: just to clarify, is this return statement here to set the `currentPage` state to `not set` right after the component unmounts?

blegesse-git (approved):

src/shared/components/NotFound.tsx
wdoconnell: Yup! You nailed it @blegesse-git 

This is also the way we currently handle it on the dashboards page.

wdoconnell: Basically - this ensures we get a redux state update whenever the user clicks away from the "Not Found" page - whether it's through navigating to a new url, clicking a link on the navbar, etc.

blegesse-git: thanks for the explanation! makes perfect sense

TCL735:
what about the side nav? it's funny that we still have a side nav. But i guess the side nav is helpful in navigating away from the 404 page.

TCL735 (approved):

wdoconnell:
> what about the side nav? it's funny that we still have a side nav. But i guess the side nav is helpful in navigating away from the 404 page.

Yeah this is existing behavior @TCL735 , which is interesting. Here's how it works (this PR doesn't change it):

If you try to navigate to a URL that does include your org route - /orgs/39024902/BLAH, it will still show the left nav.

If you try to navigate to a URL that doesn't include your org route /BLAH, it won't show the left nav.

PR #6010, opened 2022-10-06
fix: update Giraffe to fix Simple Table with multiple table headers

### LINKED ISSUE #6004: Cropped content in table; table doesn't scroll
## About the bug

**Steps to reproduce:**

* use a small screen (e.g. a laptop)
* make a query that produces a couple of tables (I used the flux profiler)

**Actual behavior:**

I see there is a row of data but I can only see a few pixels of it (see picture below). The table doesn't scroll vertically (tried with scrolling wheel / scrolling gestures)

By moving the splitter up and down I can fully reveal the row or I can fully obscure the row and thus cause it to move to the next page in the pager.

**Expected behavior:**

I expect to see the results without fiddling with the splitter.

**Visual Proof:**

image

**Environment**

Chrome 105.0.5195.125 macos 12.5.1

TCL735:
Closes #6004  

BEFORE - notice the last row on page 2 overlaps with the edge of the table and horizontal scrollbar:

https://user-images.githubusercontent.com/10736577/194380028-1b0be211-c84f-4145-a064-2c0c4cc4fa5d.mp4

AFTER - fixed, no overlapping:

https://user-images.githubusercontent.com/10736577/194380106-57f707af-8cb7-4e40-b99a-84c98f95c8f7.mp4

### DIFF
--- a/cypress/e2e/shared/simpleTable.test.ts
+++ b/cypress/e2e/shared/simpleTable.test.ts
@@ -46,8 +46,8 @@ describe('simple table interactions', () => {
     // click last page
     cy.getByTestID('pagination-item').last().should('be.visible')
     cy.getByTestID('pagination-item').last().click()
-    // verify correct number of pages
-    cy.getByTestID('pagination-item').last().contains('50')
+    // verify correct number of pages: 300 items at 3 items per page
+    cy.getByTestID('pagination-item').last().contains('100')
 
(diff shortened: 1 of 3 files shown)


hoorayimhelping (dismissed):

TCL735:
Will need to update the e2e tests for Simple Table.

ChitlangeSahas (approved):

PR #6014, opened 2022-10-06
perf(FluxEditor): only send to redux when user has finished typing

ChitlangeSahas:
Closes #

Sending to redux was taking 300ms each query, this becomes really expensive for every keystroke.

I made it so that it only sends it to the redux when user has finished typing. 

https://user-images.githubusercontent.com/18511823/194423461-2470bae9-3cf3-4a85-be27-446a3899ba1c.mov

### Checklist

### DIFF
--- a/src/timeMachine/components/TimeMachineFluxEditor.tsx
+++ b/src/timeMachine/components/TimeMachineFluxEditor.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FC, lazy, Suspense} from 'react'
+import React, {FC, lazy, Suspense, useState} from 'react'
 import {useSelector, useDispatch} from 'react-redux'
 import {
   RemoteDataState,
@@ -29,10 +29,18 @@ const FluxEditor = lazy(() => import('src/shared/components/FluxMonacoEditor'))
 const TMFluxEditor: FC<{variables: Variable[]}> = props => {
   const dispatch = useDispatch()
   const activeQueryText = useSelector(getActiveQuery).text
+  const [monacoInput, setMonacoInput] = useState(activeQueryText)
+  let userIdleTimer
 
   const handleSetActiveQueryText = React.useCallback(
     (text: string) => {
-      dispatch(setActiveQueryText(text))
+      setMonacoInput(text)
+
+      // only send to redux when the user hasn't typed for one second.
+      clearTimeout(userIdleTimer)
+      userIdleTimer = setTimeout(() => {
+        dispatch(setActiveQueryText(text))
+      }, 1000)
     },
     [dispatch]
   )
@@ -73,7 +81,7 @@ const TMFluxEditor: FC<{variables: Variable[]}> = props => {
           }
         >
           <FluxEditor
-            script={activeQueryText}
+            script={monacoInput}
             variables={props.variables}
             onChangeScript={handleActiveQuery}
             onSubmitScript={handleSubmitQueries}


src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -73,7 +81,7 @@ const TMFluxEditor: FC<{variables: Variable[]}> = props => {
           }
         >
           <FluxEditor
-            script={activeQueryText}
+            script={monacoInput}
```
ChitlangeSahas, line 84: I'll rename this, just seeing how the tests do.

src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -29,10 +29,18 @@ const FluxEditor = lazy(() => import('src/shared/components/FluxMonacoEditor'))
 const TMFluxEditor: FC<{variables: Variable[]}> = props => {
   const dispatch = useDispatch()
   const activeQueryText = useSelector(getActiveQuery).text
+  const [monacoInput, setMonacoInput] = useState(activeQueryText)
```
ChitlangeSahas, line 32: store this locally, and then send into redux later.

src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -29,10 +29,18 @@ const FluxEditor = lazy(() => import('src/shared/components/FluxMonacoEditor'))
...
 
   const handleSetActiveQueryText = React.useCallback(
     (text: string) => {
-      dispatch(setActiveQueryText(text))
+      setMonacoInput(text)
+
+      // only send to redux when the user hasn't typed for one second.
+      clearTimeout(userIdleTimer)
+      userIdleTimer = setTimeout(() => {
+        dispatch(setActiveQueryText(text))
```
ChitlangeSahas, line 42: This call takes 300ms to finish. Doing it every keystroke is affecting performance

wiedld (commented):
Right now we have the global store [redux] copy of the query, treated as the single src of truth for the timeMachine editor. This means that this src:
* is used to determine what variables are present in several downstream code locations.
* this src is updated in callbacks from the monaco-editor itself, whenever we modify flux code in the editor (e.g. injecting flux functions).

Given that tests related to these^^ are failing, I'm guess the delay is causing race conditions.

Instead, I pushed up a commit off your branch with a few minor tweaks...attempted to keep a local (fast updating) copy, which gets overwritten by the src of truth. [It passes the tests](https://github.com/influxdata/ui/commit/727e2616b02fddd324fdf8199e60bd501b07ed95), but **note that it still would need more QA**. Especially since this alternative approach also has potentially introduced other race conditions.

src/timeMachine/components/TimeMachineFluxEditor.tsx
hoorayimhelping: Check out the concepts of debouncing and throttling: https://blog.openreplay.com/forever-functional-debouncing-and-throttling-for-performance

and the related [`lodash.debounce`](https://lodash.com/docs/4.17.15#debounce) functionality so you don't have to roll your own

hoorayimhelping:
>Given that tests related to these^^ are failing, I'm guess the delay is causing race conditions... Especially since this alternative approach also has potentially introduced other race conditions.

@wiedld let's all three of us get together early this week to talk about these race conditions. It sounds like the downstream code is coupled to this code's implementation which is why things get wonky when the implementation changes. If that's the case, I'm interested in cleaning up that coupling.

wiedld:
We decided that this approach was not the route to take. Undecided on which route to take, but we can circle back to this code if needed.

PR #6015, opened 2022-10-06
chore: improve set feature flags

subirjolly:
WIP: DO NOT REVIEW YET

### Checklist

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -8,7 +8,7 @@ import {
   Secret,
 } from '../../src/types'
 import {Bucket, Organization} from '../../src/client'
-import {setOverrides, FlagMap} from 'src/shared/actions/flags'
+import {FlagMap} from 'src/shared/actions/flags'
 import {addTimestampToRecs, addStaggerTimestampToRecs, parseTime} from './Utils'
 import 'cypress-file-upload'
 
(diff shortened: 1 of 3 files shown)


cypress/support/commands.ts
```diff
@@ -1040,20 +1040,22 @@ export const makeGraphSnapshot = (() => {
...
-  return cy.window().then((win: any) => {
-    // eslint-disable-next-line no-extra-semi
-    win.store.dispatch(setOverrides(flags))
+  const flagsURIs = ['/api/v2private/flags', '/api/v2/flags']
+  flagsURIs.forEach(url => {
+    cy.intercept(url, req => {
+      req.continue(res => {
+        res.send({...res.body, ...flags})
+      })
+    }).as('setFeatureFlagsResponse')
```
genehynson, line 1055: this is really clever, nice!

cypress/support/commands.ts
```diff
@@ -1040,20 +1040,22 @@ export const makeGraphSnapshot = (() => {
...
 
 export const setFeatureFlags = (flags: FlagMap): Cypress.Chainable => {
-  // make sure the app is loaded before dispatching
-  cy.getByTestID('tree-nav')
-  return cy.window().then((win: any) => {
-    // eslint-disable-next-line no-extra-semi
-    win.store.dispatch(setOverrides(flags))
-  })
+  // Need to refresh page so that flags responses can be overwritten
+  return cy.setFeatureFlagsNoNav(flags).reload().getByTestID('tree-nav')
```
genehynson, line 1044: instead of reloading which will slow down each test, could we call `cy.setFeatureFlags` before `cy.signin()` is called in each test? Maybe we could refactor it so folks wouldn't have to remember this order - like adding a `cy.signInAndSetFeatureFlags()` command

genehynson (commented):
love this! just had an idea to make it a little faster. let me know your thoughts

cypress/support/commands.ts
subirjolly: Awesome Idea.

I thought of this but I didn't think of adding another implementation. I was just thinking of modifying the existing `signin`. I think your idea of adding a separate implementation is awesome. I just wanted to keep the changes to a minimum. I really like this idea. So, added a new issue to tackle this separately: https://github.com/influxdata/ui/issues/6026

genehynson (approved):

PR #6020, opened 2022-10-07
feat(Logout): use quartz logout endpoint + refactoring login redirect code

ChitlangeSahas:
We want to use the quartz endpoint to logout instead.

Note: I used `api/env/quartz-login-url` , it might be confusing at a first glance why it's calling a `login` path to logout. This was just a naming thing, we thought the endpoint will only be used for logging in, but now that we're using it for logging out, I will make follow up changes to rename it to `quartz-url`

 

### Checklist

### DIFF
--- a/src/Logout.tsx
+++ b/src/Logout.tsx
@@ -7,10 +7,15 @@ import {useHistory} from 'react-router-dom'
 import {postSignout} from 'src/client'
 
 // Constants
-import {CLOUD, CLOUD_SIGNOUT_PATHNAME} from 'src/shared/constants'
+import {
+  CLOUD,
+  CLOUD_LOGOUT_PATH,
+  CLOUD_SIGNOUT_PATHNAME,
+} from 'src/shared/constants'
 
 // Components
 import {reset} from 'src/shared/actions/flags'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 const Logout: FC = () => {
   const history = useHistory()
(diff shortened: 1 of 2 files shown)


src/shared/constants/index.ts
```diff
@@ -97,7 +97,7 @@ export const RUDDERSTACK_WRITE_KEY = formatConstant(
   process.env.RUDDERSTACK_WRITE_KEY
 )
 
-export const CLOUD = !!process.env.CLOUD_URL
+export const CLOUD = true
```
TCL735, line 100: Is this supposed to be in here?

src/identity/components/GlobalHeader/IdentityUserAvatar.tsx
```diff
@@ -88,7 +88,7 @@ class IdentityUserAvatar extends React.Component<Props, State> {
           <Link
             className="user-popover-footer--button"
             onClick={this.handlePopoverClick(HeaderNavEvent.UserLogoutClick)}
-            to="/logout"
+            to="/logoutyeah"
```
TCL735, line 91: Is this the correct path?

ChitlangeSahas: My bad, all testing changes snuck in as well, fixed now!

TCL735 (approved):

src/Logout.tsx
```diff
@@ -19,6 +24,26 @@ const Logout: FC = () => {
   useEffect(() => {
     const handleSignOut = async () => {
       if (CLOUD) {
+        if (isFlagEnabled('universalLogin')) {
+          fetch('/api/env/quartz-login-url')
+            .then(response => {
+              response
+                .text()
+                .then(response => {
+                  const redirectUrl = `${response}${CLOUD_LOGOUT_PATH}`
```
hoorayimhelping, line 33: instead of `response`, maybe we should call this something like `quartzLoginUrl` or something

ChitlangeSahas: Okay yeah I agree! I'll change that here, and in the LoginPage file 👍

hoorayimhelping (approved):

PR #6025, opened 2022-10-11
chore: update the status code check;

asalem1:
Closes #

Closes the issue where a patch request should return a 200 and not a 201. The bug that arose was a false negative, where the operation to update the script would succeed but show an error toast

### Checklist


appletreeisyellow (approved):
LGTM! Just wondering, is it a recent change for status code from `201` to `200`?

Issue #5960, opened 2022-10-03
Separate Entry Points: Refactor Top of App


hoorayimhelping:
The 'top of the application', which refers to the `index.tsx` file and the first components that are rendered and the data that are fetched by those components, is spread out over several components, a few of which don't even render anything meaningful, and only fetch data.

The following screenshot demonstrates the relationship of the top-of-app components.

![Image](https://user-images.githubusercontent.com/146112/193607022-75359a7c-0523-4490-a016-fe2a3f05e0f9.png)

Most of the foreseeable multiple entry points will need to make use of this top-of-app flow. Refactoring it to be less spread out and more contained will (hopefully) make the process of adding multiple entry points simpler.

### Acceptance Criteria
- Remove components that don't render anything meaningful, and instead encapsulate their functionality into a separate module (like a thunk, a util, an api-model layer)
- Rename components/functions so that their behavior matches their description (e.g. `Setup.tsx` defines login routes and flows, while `Signin.tsx` makes a request to the API for authentication, but doesn't actual handle the flow of signing in)

wdoconnell:
This ticket is being closed due to change in priority.

PR #6027, opened 2022-10-11
refactor(Signin): clean up variables, component names, and constants

hoorayimhelping:
Helps With #5820
Helps With #5960

- Removes constants for `LOGIN` => `'/login'`, `SIGNIN` => `'/signin'`, `LOGOUT` => `'/logout'` in favor of using the literal string values. These constants were only used by this component, and were providing no abstraction. They instead forced a reader to open another file to get information that could be available in the current file.
- Renames component state variable `allowed` to `shouldShowOnboarding`
- Renames `LoginPage` and `SigninPage` to `CloudLoginPage` and `OSSLoginPage` respectively, because I could never remember which page mapped to which experience. Also renames `LoginPage.scss` to `CloudLoginPage.scss`

### Checklist


TCL735 (approved):

src/Setup.tsx
```diff
@@ -8,20 +8,19 @@ import {getSetup} from 'src/client'
...
 // lazy loading the signin component causes wasm issues
 import Signin from 'src/Signin'
 
 const OnboardingWizardPage = lazy(
   () => import('src/onboarding/containers/OnboardingWizardPage')
 )
-const SigninPage = lazy(() => import('src/onboarding/containers/SigninPage'))
+const OSSLoginPage = lazy(
+  () => import('src/onboarding/containers/OSSLoginPage')
+)
```
wdoconnell, line 20: Not for this PR - but what do you think about [this](https://reactjs.org/docs/code-splitting.html#named-exports) as far as resolving the issue with using named imports with React lazy? Adding a separate component for each lazy import is annoying, but maybe less so if we keep them in one folder (e.g. src/lazyimports).

Actually, having a separate wrapper component for anything that is imported using lazy could also give us space to [add retry functions around the components to deal with the chunk loading error issue.](https://medium.com/@botfather/react-loading-chunk-failed-error-88d0bb75b406)

wdoconnell (approved):

src/Setup.tsx
hoorayimhelping: @wdoconnell we [have ways of handling named exports](https://github.com/influxdata/ui/blob/master/src/shared/containers/index.tsx#L102-L136) and react lazy loading. The idea about having a lazy component loader that handles exceptions might be worth looking into 🤔

PR #6031, opened 2022-10-11
chore: remove references to multiOrg flag

### LINKED ISSUE #4055: Multi Org Phase 1 - Remove Feature Flag

wdoconnell:
Connects #4055. I will close #4055 after the flag is removed from IDPE, and then Configcat.

Removes the `multiOrg` feature flag from the UI, as the multiOrg phase 1 features have now been used in all `CLOUD` environments since Sept. 15.

Cloud UI after PR (no change in behavior)
--
https://user-images.githubusercontent.com/91283923/195176298-03275db7-7b3a-4142-8142-0bcc34f510c6.mov

Onboarding Pages (see discussion in comments re: retaining header space)
--
https://user-images.githubusercontent.com/91283923/195184200-e49fe0c8-c471-4204-b7a4-4929b468c9b7.mov

### Checklist


src/App.tsx
```diff
@@ -42,20 +42,19 @@ import {CLOUD} from 'src/shared/constants'
 import {executeVWO} from 'src/utils/vwo'
 
 // Styles
-import './MultiOrgOverrideStyles.scss'
+import 'src/identity/styles/MultiOrgStyles.scss'
```
wdoconnell, line 46: I don't think these styles should live in the root directory. I'd prefer keeping them with the other user-identity-related components.

src/homepageExperience/containers/ArduinoWizard.tsx
```diff
@@ -177,13 +175,7 @@ export class ArduinoWizard extends PureComponent<{}, State> {
...
     return (
       <Page>
-        <Page.Header fullWidth={false}>
-          {/* Need an empty div so the upgrade button aligns to the right. (Because clockface uses space-between to justifyContent)*/}
-          <div />
-          {!isFlagEnabled('multiOrg') && (
-            <RateLimitAlert location="firstMile.arduinoWizard" />
-          )}
-        </Page.Header>
+        <Page.Header fullWidth={false}></Page.Header>
```
wdoconnell, line 178: Although the page header no longer contains any elements, we don't want to remove it, because our page styles expect a header to exist. If we remove the header, the onboarding page looks like this. Way too close to the top.

https://user-images.githubusercontent.com/91283923/195183599-09e906ba-37a4-4de7-8775-4d25825be12e.mov

Instead, we want the page to keep looking like this.

https://user-images.githubusercontent.com/91283923/195183669-8e9cbd3f-4891-4c72-b66b-d0b32e87d7a5.mov

TCL735 (dismissed):

src/App.tsx
```diff
@@ -42,20 +42,19 @@ import {CLOUD} from 'src/shared/constants'
...
```
hoorayimhelping, line 45: i think this actually belongs in `chronograf.scss` now since it's used app-wide

src/homepageExperience/containers/ArduinoWizard.tsx
hoorayimhelping: minor thing but

```suggestion
        <Page.Header fullWidth={false} />
```

hoorayimhelping (commented):
A couple of minor things

src/homepageExperience/containers/ArduinoWizard.tsx
wdoconnell: Good idea. I'll do this for all the onboarding pages.

src/App.tsx
wdoconnell: 💯 . I'll move this into the chronograf scss file.

hoorayimhelping (approved):

src/accounts/AccountHeader.tsx
```diff
@@ -10,6 +11,7 @@ type Props = {
 const AccountHeader: FC<Props> = ({testID = 'member-page--header'}) => (
   <Page.Header fullWidth={true} testID={testID}>
     <Page.Title title="Account" />
+    <LimitChecker />
```
wdoconnell, line 14: I added this component back in to the Account Header and billing page because it's actually doing more than acting as a wrapper for `RateLimitAlert`. 

On mount, the component also runs the `getReadWriteCardinalityLimits` thunk, which retrieves and updates the status of the user's read, write, and cardinality limits; and dispatches a notification to the user if they've exceeded their read limit. (Makes sense that this would happen when checking the status of the account, or the billing page.)

LimitChecker's props presumed child components, so I modified it slightly to make the rendering of child components optional and turn it into a singleton tag. This way, we can keep it on the page, in the same structure, without needing it to contain another component.

src/accounts/AccountPage.tsx
```diff
@@ -170,9 +154,6 @@ const AccountAboutPage: FC = () => {
             </FlexBox>
           </>
         )}
-        <Overlay visible={isSwitchAccountVisible}>
-          <SwitchAccountOverlay onDismissOverlay={closeSwitchAccountDialog} />
-        </Overlay>
         <CancelServiceProvider>
```
wdoconnell, line 157: We don't need the overlay or `closeSwitchAccountDialog` either, because there is no longer a switch-account button on this page.

wdoconnell:
New, improved, and - most importantly - with moar Red Diff&trade;.

hoorayimhelping (approved):

PR #6033, opened 2022-10-11
fix: fix the timerange not accurately reflecting the selected time range

### LINKED ISSUE #5834: Graph Date Range not always Updating when changing Time Range in Query (in new Script Editor)
When running queries with the table/graph toggle set to graph, when changing the date range and re-running the query, the graph doesn't always update to reflect the new data range (see video).  Sometimes it does, but often it does not.

https://user-images.githubusercontent.com/97983559/191316653-13a92852-abf4-48f4-a737-9d3f7130f7e5.mov

asalem1:
Closes #5834 

This PR updates the Results to pass the `range` from the persistence layer down to the view so that we can accurately update the UI based on a user's timerange selection

timerange-fix


wiedld (approved):
lgtm

PR #6035, opened 2022-10-12
chore: breakup shared query context helpers

wiedld:
Cleanup PR.
Done as a precursor to work in this PR: https://github.com/influxdata/ui/pull/6036

## Done:
* breakup the large shared query context file, into a submodule
* show the preprocessing, versus the postprocessing
* NO CODE WAS CHANGED. Only moved code into separate files.
    * except: I did change 2 default exports, into named exports 

## Checklist


wiedld:
> I am curious how this is a precursor to the work moving forward

The editor context file was growing in size to be huge and hard to navigate. And the difference was all in many helper functions, with no side effects, and located outside the context itself. So before adding yet more no-side-effect helper functions, I decided to break up the helpers into other files.

I choose the `preprocessing` versus `postprocessing`, since we had earlier conversations about how much preprocessing we do before running each query. And how much this preprocessing is standardized across different parts of the application. So this breakout was done with that future goal/work in mind.

Issue #5994, opened 2022-10-05
CSV download should be the entire dataset.


wiedld:
## Expectation

We are a DaaS. Therefore, we expect query results to be correct. This includes the data downloaded into csv.

## Bug

The new QxBuilder, and old DataExplorer, both have a csv download. It currently downloads the truncated dataset.
* have it download the complete dataset
* please don't store complete dataset in memory prior to download request

wiedld:
Merged

PR #6036, opened 2022-10-12
feat(5994): csv download entire dataset

wiedld:
Part of #5994 

This is the csv download for new Qx Builder only.
(The old dataExplorer csv dowload, will be a separate PR.)

## Vid:

https://user-images.githubusercontent.com/10232835/195221053-a51dc0f1-65bf-4be2-a30f-05d143842aa3.mp4

## Choices made:
* download into Blob, then provide file download API.
  * blob is a raw File binary object.
  * we convert the response directly into a blob, with no intermediate javascript allocation
  * most memory efficient we can get, with in-memory objects. 
  * limitations:
     * RAM
     * each browser sets it's own max blob size limitation
        * example in chrome: https://source.chromium.org/chromium/chromium/src/+/main:storage/browser/blob/blob_memory_controller.cc?q=CalculateBlobStorageLimitsImpl&ss=chromium 
* alternative not chosen:
  *  piping stream to stream. Output network response directly into fs download.
     * `response.body` (a ReadableStream) --> FileSystem write stream.
     * use the FileSystem web API
        * one of the goals is to enable larger binary files. also includes writing of files.
           * https://fs.spec.whatwg.org/ 
        * for firefox, they have spec'ed the use cases as being a bit more narrow:
           * https://developer.mozilla.org/en-US/docs/Web/API/File_and_Directory_Entries_API/Introduction#usefulness_of_the_api  
  *  not chosen because:
     * the browser FileSystem spec was never fully agreed upon, and firefox only did a partial implementation
          * https://developer.mozilla.org/en-US/docs/Web/API/File_and_Directory_Entries_API/Firefox_support    
          * the result == firefox's web API does not support our use case.

## Checklist


src/shared/contexts/query/context.tsx
```diff
@@ -86,6 +87,95 @@ export const DEFAULT_CONTEXT: QueryContextType = {
 export const QueryContext =
   React.createContext<QueryContextType>(DEFAULT_CONTEXT)
 
+const buildQueryRequest = (
```
appletreeisyellow, line 90: Nice! Moving these logic into a function makes it clear what they are doing

src/shared/contexts/query/context.tsx
```diff
@@ -248,6 +223,67 @@ export const QueryProvider: FC = ({children}) => {
...
+      return {
+        type: 'SUCCESS',
+        csv,
+        bytesRead,
+        didTruncate,
+      }
+    }
+
+    const handleRawBlob = async response => {
+      const data = await response.blob()
```
appletreeisyellow, line 270: Simple!

appletreeisyellow (approved):
LGTM! Only one comment on adding an `event()` that was mentioned in IPM yesterday. Not a blocker, and I think it can be done in another PR

appletreeisyellow (approved):

PR #6038, opened 2022-10-12
refactor(Signin): remove GetMe component; rename and clean up Signin.tsx

hoorayimhelping:
Helps With #5820
Helps With #5960

- Renames `Signin.tsx` to `Authenticate.tsx`
- Previously `Signin.tsx` would authenticate, start a timer to authenticate every 60 seconds, then render `GetMe.tsx` to authenticate again to store the results in redux. Now `Authenticate.tsx` authenticates and stores the results in redux before starting the timer. This allows the removal of `GetMe.tsx`
- Removes `GetMe.tsx`
- Adds a method `getQuartzIdentityNoErrorHandling` (open for better naming suggestions here) that lets the calling site handle errors, for use in `Authenticate.tsx`. This method will store results in redux. The existing `getQuartzIdentity` method now calls this one.

### Old behavior (three calls: one to /identity, one to /me, one to /identity later)
old

### New behavior calling /me (two calls: one to /me, one to /identity later)
new with me

### New behavior calling only /identity (single call to /identity)
new with identity

### Checklist


src/Authenticate.tsx
```diff
@@ -49,28 +51,106 @@ type Props = OwnProps & RouteComponentProps & ReduxProps
...
   public state: State = {
     loading: RemoteDataState.NotStarted,
-    auth: false,
+    isAuthenticated: false,
   }
 
   private hasMounted = false
   private intervalID: NodeJS.Timer
 
+  private handleAuthenticationFailure = error => {
```
hoorayimhelping, line 63: this behavior used to be part of `checkForLogin`. It has been pulled out into a reusable method

src/me/actions/thunks/index.ts
```diff
@@ -18,29 +18,25 @@ import {MeState} from 'src/me/reducers'
...
-    })
-
-    HoneyBadger.setContext({
-      user_id: user.id,
-    })
-
-    dispatch(setMe(user as MeState))
-  } catch (error) {
-    console.error(error)
+  const resp = await getIdpeMe({})
```
hoorayimhelping, line 21: this was previously only called in the `GetMe` component. It's now called only in `Authenticate`, which handles the error coming back, so the error handling was removed.

TCL735 (approved):

wdoconnell (approved):

Issue #5890, opened 2022-09-26
Graph visualization doesn't respect query range defined in flux query


genehynson:
Query:

```
from(bucket: "iot")
  |> range(start: 2022-09-21T15:26:00Z, stop: 2022-09-21T15:30:00Z)
  |> filter(fn: (r) => r["_measurement"] == "obd")
  |> filter(fn: (r) => r["_field"] == "mph")
  |> toFloat()
```

Graph doesn't respect the range:

<img width="1538" alt="image" src="https://user-images.githubusercontent.com/6411855/192343080-922c5202-47ee-469c-8ab7-fc472cb581bb.png">

Change query to use custom time range dropdown:

```
from(bucket: "iot")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "obd")
  |> filter(fn: (r) => r["_field"] == "mph")
  |> toFloat()
```

And now the visualization is correct:

<img width="1546" alt="image" src="https://user-images.githubusercontent.com/6411855/192343291-91d3873f-a9e2-462e-9642-97c3a21cbe47.png">

## About your environment

**Environment info:**
- https://us-west-2-1.aws.cloud2.influxdata.com/

kristinarobinson:
@TCL735, @hralexander -- I know that Denise and Timmy both looked at the query range when the dropdown is used or when hardcoded within the flux.  Can either or both of you see if you can repro, and hand off to the appropriate owner.

cannonpalms:
## Hard-coded time range in flux query

On https://go.influxcorp.co/newtools
```flux
from(bucket: "query_log")
  |> range(start: -24h)
  |> filter(fn: (r) => r["_measurement"] == "query_log")
  |> filter(fn: (r) => r["_field"] == "executeDuration")
  |> filter(fn: (r) => r["env"] == "prod01-eu-central-1")
  |> filter(fn: (r) => r["orgID"] == "2bc31ebe26081a78")
  |> filter(fn: (r) => r["source"] == "influxdb-client-go")
  |> filter(fn: (r) => r["role"] == "queryd-pull-external")
  |> group()
  |> map(fn: (r) => ({ r with _value: float(v: r._value) / float(v: int(v: 1s))}))
  |> aggregateWindow(every: 5m, fn: (tables=<-, column) => tables
      |> quantile(q: 0.99), createEmpty: false)
      |> exponentialMovingAverage(n: 200)
```

[...]

PR #6048, opened 2022-10-12
fix: update syncing behavior for adaptive zoom to account for zoom level

TCL735:
Related to #5890  

With the adaptive zoom feature, on by default, re-running a query without zooming causes the graph to slowly "shrink". This is because the domain was not being properly synced.

- Fixes the syncing to account for zoomed-in or not zoomed-in
- When zoomed-in, domain is controlled by the `setDomain` callback
- When not zoomed-in, domain is synced according to a few variables

BEFORE - the line slowly shrinks as the user re-runs the query without zooming in

https://user-images.githubusercontent.com/10736577/195465838-64d8ea94-20ed-4f8f-9bc3-5cb27677569a.mp4

AFTER - the line does not shrink and the time axis "moves" with the refreshes

https://user-images.githubusercontent.com/10736577/195465892-f6b61f52-3c87-42c0-a23a-97e927e589dc.mp4


hoorayimhelping (approved):
Nice videos!

subirjolly (approved):

PR #6061, opened 2022-10-13
chore: fix checkout test, resolve required property blocking form validation

wdoconnell:
This PR started as an effort to remove a flag toggle from the checkout test. Making edits to that individual test - or even making no changes to the test - resulted in a CI test failure, which didn't make sense. The underlying cause is that the test was actually _already_ broken, but wasn't being run by CI because it was in a separate subfolder that wasn't being run by the test runner.

This PR resolves the issue by (a) moving the checkout tests into the base cloud directory and (b) fixing the broken test, by removing the `required` from an `<Input />` tag. The `required` was supposed to be present only in the parent component, as having `required` in the child component prevents the form `onSubmit` from running, which prevents the checkout validation from working correctly.

Here's the way the test behaves before (broken form).
--
https://user-images.githubusercontent.com/91283923/195872495-77aa44b1-c32c-452e-8ddb-56ab48d4e15f.mov

Here's the way the test behaves now.
--
https://user-images.githubusercontent.com/91283923/195872600-0bc6e789-cda5-4c96-b19a-ea3b016f5430.mov

Component Flow
--
1. Hitting 'Upgrade' triggers the `onSubmit` in CheckoutForm.tsx
2. If `handleFormValidation() > 0`, we `return` without submitting the form to Zuora.
3. `handleFormValidation` runs `getInvalidFields`, which filters the array of inputs and returns an array containing any fields that are set to empty string (`''`). If we get an array of non-zero length, the form still doesn't submit because `handleFormValidation() > 0`.

### Checklist


cypress/e2e/cloud/checkout/checkoutpage.test.ts
```diff
@@ -27,135 +27,127 @@ describe('Checkout Page Works', () => {
...
-      // Click Upgrade
-      cy.getByTestID('checkout-upgrade--button').click()
+    // Click Upgrade
+    cy.getByTestID('checkout-upgrade--button').click()
 
-      // Check all errors are visible
-      cy.getByTestID('balanceThreshold--input').scrollIntoView()
-      cy.getByTestID('balanceThreshold--form-element-error').should(
-        'be.visible'
-      )
```
wdoconnell, line 49: These are changes enforced by the new line width preferences in prettier.

cypress/e2e/cloud/checkout/checkoutpage.test.ts
```diff
@@ -27,135 +27,127 @@ describe('Checkout Page Works', () => {
...
-      cy.getByTestID('shouldNotify--checkbox').click()
-      cy.getByTestID('shouldNotify--checkbox--input').should('be.checked')
-      cy.getByTestID('notifyEmail--input').should('have.value', email)
-      cy.getByTestID('balanceThreshold--input').should('have.value', limit)
-
-      // should render US Billing Address
+    // Check all errors are visible
+    cy.getByTestID('balanceThreshold--input').scrollIntoView()
+    cy.getByTestID('balanceThreshold--form-element-error').should('be.visible')
+    cy.getByTestID('balanceThreshold--form-element-error').contains(
```
wdoconnell, line 45: These are changes enforced by prettier. 

The line width no longer exceeds its 80-character preference since this code block is no longer part of the promise chain from `cy.setFeatureFlagsNoNav`.

hoorayimhelping (dismissed):

wdoconnell:
Another solution here would be to change the test to acknowledge (implicitly) the browser-based prompt for the required form element. Doesn't seem to be causing any issues, because you can't submit the form in that case either. But I assume that's not the intent, since our custom validation logic is presumably there to help guide the user on a more specific path.

wdoconnell:
I don't think it changes the business logic for the following reason. Here are the form inputs marked `required` in the existing logic. 

- email
- balanceThreshold
- Canada and International Contact Forms
  - Country [`required` doesn't do anything here because there is no null option - you have to select a country]
  - City
- US Contact Form
  - Same as Canada + International, plus
    - State (`usSubdivision`)
    -  Zip (`postalCode`)

The effect of having these inputs marked `required` [is that if the input is an empty string for any of these fields, the browser opens a standard tooltip ](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) and prevents submission of the form, and in fact, doesn't run the `onSubmit` function (which means our validation logic isn't used).

Our intended business logic is designed to do the same check `required` does - see if the value of the input is equal to an empty string, and in the case of the US, also require state and zip code values that are not empty strings. There's no validation required for 'country' because that input is a dropdown with no `''` option, so the user can't select an empty country.

Screen Shot 2022-10-14 at 11 51 31 AM

So, having `required` in these inputs does the same thing as our (current) validation logic, but is obscuring the fact that the validation logic isn't actually running in the one case it needs to - when a field is empty. If we later change the validation logic (which we probably should - US zip needs to be a number, the user should need to enter a valid US state, etc.), the presence of `required` will conceal the fact that changes to the validation logic aren't actually doing anything in the form.

wdoconnell:
We've resolved the reason this only comes up when you change this file: this test suite isn't getting run, because it's located in a subfolder of the cloud test directory. This test should not only be fixed, but also _moved_ into the main cloud directory.

We've checked monitor-ci, and this test is indeed currently not working, and moreover isn't run unless changes are made to this specific test.

PR #6062, opened 2022-10-13
fix(5994): timeMachine download complete csv

### LINKED ISSUE #5994: CSV download should be the entire dataset.

wiedld:
Closes #5994 
This is the 2nd half of this ticket. To have timeMachine download the entire network payload in the csv.

## DONE:
* decisions on why using a blob, are documented in the PR description here:
   *  https://github.com/influxdata/ui/pull/6036
* decisions to keep timeMachine maintainable:
   * kept this legacy feature in the legacy code patterns, so easier to maintain by other devs
   * this meant:
      *  centralized state
      * logic in centralized actions
      * stateful components
      * connector pattern
* Goal: keep the query run in both cases: on submit, and on csv download
   * this means the same `runTimeMachineQuery()` and same `timeMachineQueryErrorNotification()`
   * only diff is the post query processing (`responseProcess`)
* new feature in timeMachine:
   * able to cancel long-running download queries 

## Vid:

https://user-images.githubusercontent.com/10232835/195486571-a43001d7-defd-4289-b270-6804d3e92a97.mp4

## Checklist

### DIFF
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -1,8 +1,9 @@
 import {Organization} from '../../../src/types'
 import {points} from '../../support/commands'
-const path = require('path')
 
 describe('DataExplorer', () => {
+  let route: string
+
   beforeEach(() => {
     cy.flush()
     cy.signin()
@@ -16,7 +17,8 @@ describe('DataExplorer', () => {
     cy.get('@org').then(({id}: Organization) => {
       cy.createMapVariable(id)
       cy.fixture('routes').then(({orgs, explorer}) => {
-        cy.visit(`${orgs}/${id}${explorer}`)
+        route = `${orgs}/${id}${explorer}`
+        cy.visit(route)
         cy.getByTestID('tree-nav').should('be.visible')
       })
     })
@@ -912,6 +914,8 @@ describe('DataExplorer', () => {
   })
 
   describe('download csv', () => {
+    // docs for how to test form submission as file download:
+    // https://github.com/cypress-io/cypress-example-recipes/blob/cc13866e55bd28e1d1323ba6d498d85204f292b5/examples/testing-dom__download/cypress/e2e/form-submission-spec.cy.js
     const downloadsDirectory = Cypress.config('downloadsFolder')
 
     const validateCsv = (csv: string, rowCnt: number) => {
@@ -930,6 +934,10 @@ describe('DataExplorer', () => {
     })
 
     it('can download a file', () => {
+      cy.intercept('POST', '/api/v2/query?*', req => {
+        req.redirect(route)
+      }).as('query')
+
       cy.getByTestID('time-machine--bottom').within(() => {
         cy.getByTestID('flux-editor', {timeout: 30000}).should('be.visible')
           .monacoType(`from(bucket: "defbuck")
@@ -939,15 +947,19 @@ describe('DataExplorer', () => {
           .click()
       })
 
-      cy.task('getDownloads', {dirPath: downloadsDirectory}).should(files => {
-        ;(files as string[]).forEach(filename => {
-          expect(filename).contains('influxdata_')
-          const fullPathFile = path.join(downloadsDirectory, filename)
-          cy.readFile(fullPathFile, {timeout: 15000})
-            .should('have.length.gt', 50)
-            .then(csv => validateCsv(csv, 20))
+      cy.wait('@query')
+        .its('request')
+        .then(req => {
+          cy.request(req)
+            .then(({body, headers}) => {
+              expect(headers).to.have.property(
+                'content-type',
+                'text/csv; charset=utf-8'
+              )
+              return Promise.resolve(body)
+            })
+            .then(csv => validateCsv(csv, 1))
         })
-      })
     })
   })
 })
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -914,6 +914,15 @@ describe('DataExplorer', () => {
   describe('download csv', () => {
     const downloadsDirectory = Cypress.config('downloadsFolder')
 
+    const validateCsv = (csv: string, rowCnt: number) => {
+      const numHeaderRows = 4
+      cy.wrap(csv)
+        .then(doc => doc.trim().split('\n'))
+        .then(list => {
+          expect(list.length).to.equal(rowCnt + numHeaderRows)
+        })
+    }
+
     beforeEach(() => {
       cy.writeData(points(20))
       cy.task('deleteDownloads', {dirPath: downloadsDirectory})
@@ -921,8 +930,6 @@ describe('DataExplorer', () => {
     })
 
     it('can download a file', () => {
-      cy.intercept('POST', '/api/v2/query?*').as('query')
-
       cy.getByTestID('time-machine--bottom').within(() => {
         cy.getByTestID('flux-editor', {timeout: 30000}).should('be.visible')
           .monacoType(`from(bucket: "defbuck")
@@ -932,17 +939,13 @@ describe('DataExplorer', () => {
           .click()
       })
 
-      cy.wait('@query')
-      cy.wait(1200)
-
       cy.task('getDownloads', {dirPath: downloadsDirectory}).should(files => {
         ;(files as string[]).forEach(filename => {
-          expect(filename).contains('influxdb_data.csv')
+          expect(filename).contains('influxdata_')
           const fullPathFile = path.join(downloadsDirectory, filename)
-          cy.readFile(fullPathFile, {timeout: 15000}).should(
-            'have.length.gt',
-            50
-          )
+          cy.readFile(fullPathFile, {timeout: 15000})
+            .should('have.length.gt', 50)
+            .then(csv => validateCsv(csv, 20))
         })
       })
     })
--- a/cypress/e2e/shared/explorer.test.ts
+++ b/cypress/e2e/shared/explorer.test.ts
@@ -1,5 +1,6 @@
 import {Organization} from '../../../src/types'
 import {points} from '../../support/commands'
+const path = require('path')
 
 describe('DataExplorer', () => {
   beforeEach(() => {
@@ -905,6 +906,43 @@ describe('DataExplorer', () => {
           cy.getByTestID('variable-name-input').clear().type('bad-name')
           cy.getByTestID('variable-form-save').should('be.disabled')
           cy.get('.cf-overlay--dismiss').click()
+        })
+      })
+    })
+  })
+
+  describe('download csv', () => {
+    const downloadsDirectory = Cypress.config('downloadsFolder')
+
+    beforeEach(() => {
+      cy.writeData(points(20))
+      cy.task('deleteDownloads', {dirPath: downloadsDirectory})
+      cy.getByTestID('switch-to-script-editor').should('be.visible').click()
+    })
+
+    it('can download a file', () => {
+      cy.intercept('POST', '/api/v2/query?*').as('query')
+
+      cy.getByTestID('time-machine--bottom').within(() => {
+        cy.getByTestID('flux-editor', {timeout: 30000}).should('be.visible')
+          .monacoType(`from(bucket: "defbuck")
+  |> range(start: -10h)`)
+        cy.getByTestID('time-machine--download-csv')
+          .should('be.visible')
+          .click()
+      })
+
+      cy.wait('@query')
+      cy.wait(1200)
+
+      cy.task('getDownloads', {dirPath: downloadsDirectory}).should(files => {
+        ;(files as string[]).forEach(filename => {
+          expect(filename).contains('influxdb_data.csv')
+          const fullPathFile = path.join(downloadsDirectory, filename)
+          cy.readFile(fullPathFile, {timeout: 15000}).should(
+            'have.length.gt',
+            50
+          )
         })
       })
     })
--- a/src/shared/apis/query.ts
+++ b/src/shared/apis/query.ts
@@ -41,8 +41,7 @@ export const runQuery = (
   orgID: string,
   query: string,
   extern?: File,
-  abortController?: AbortController,
-  transformResponse = processResponse
+  abortController?: AbortController
 ): CancelBox<RunQueryResult> => {
   const url = `${API_BASE_PATH}api/v2/query?${new URLSearchParams({orgID})}`
 
@@ -67,7 +66,7 @@ export const runQuery = (
   })
 
   const promise = request
-    .then(transformResponse)
+    .then(processResponse)
     .catch(e =>
       e.name === 'AbortError'
         ? Promise.reject(new CancellationError())
@@ -86,25 +85,6 @@ export const processResponse = async (
   switch (response.status) {
     case 200:
       return processSuccessResponse(response)
-    case RATE_LIMIT_ERROR_STATUS:
-      return processRateLimitResponse(response)
-    default:
-      return processErrorResponse(response)
-  }
-}
-
-export const processResponseBlob = async (
-  response: Response
-): Promise<RunQueryResult> => {
-  switch (response.status) {
-    case 200:
-      const blob = await response.blob()
-      return {
-        type: 'SUCCESS',
-        csv: blob as unknown as string,
-        bytesRead: blob.size,
-        didTruncate: false,
-      }
     case RATE_LIMIT_ERROR_STATUS:
       return processRateLimitResponse(response)
     default:
--- a/src/shared/apis/query.ts
+++ b/src/shared/apis/query.ts
@@ -41,7 +41,8 @@ export const runQuery = (
   orgID: string,
   query: string,
   extern?: File,
-  abortController?: AbortController
+  abortController?: AbortController,
+  transformResponse = processResponse
 ): CancelBox<RunQueryResult> => {
   const url = `${API_BASE_PATH}api/v2/query?${new URLSearchParams({orgID})}`
 
@@ -66,7 +67,7 @@ export const runQuery = (
   })
 
   const promise = request
-    .then(processResponse)
+    .then(transformResponse)
     .catch(e =>
       e.name === 'AbortError'
         ? Promise.reject(new CancellationError())
@@ -85,6 +86,25 @@ export const processResponse = async (
   switch (response.status) {
     case 200:
       return processSuccessResponse(response)
+    case RATE_LIMIT_ERROR_STATUS:
+      return processRateLimitResponse(response)
+    default:
+      return processErrorResponse(response)
+  }
+}
+
+export const processResponseBlob = async (
+  response: Response
+): Promise<RunQueryResult> => {
+  switch (response.status) {
+    case 200:
+      const blob = await response.blob()
+      return {
+        type: 'SUCCESS',
+        csv: blob as unknown as string,
+        bytesRead: blob.size,
+        didTruncate: false,
+      }
     case RATE_LIMIT_ERROR_STATUS:
       return processRateLimitResponse(response)
     default:
--- a/src/shared/components/CSVExportButton.tsx
+++ b/src/shared/components/CSVExportButton.tsx
@@ -70,12 +70,6 @@ class CSVExportButton extends PureComponent<Props, State> {
   }
 
   private handleClick = () => {
-    // Known cypress issue with form submit for file download
-    // Cypress automatically expects a redirect for any form submission
-    if ((window as any).Cypress) {
-      setTimeout(() => location.reload(), 3000)
-    }
-
     this.props.download()
   }
 }
--- a/src/shared/components/CSVExportButton.tsx
+++ b/src/shared/components/CSVExportButton.tsx
@@ -1,49 +1,79 @@
 // Libraries
 import React, {PureComponent} from 'react'
-import {connect} from 'react-redux'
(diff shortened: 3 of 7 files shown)


cypress/e2e/shared/explorer.test.ts
```diff
@@ -909,4 +910,41 @@ describe('DataExplorer', () => {
...
+        cy.getByTestID('time-machine--download-csv')
+          .should('be.visible')
+          .click()
+      })
+
+      cy.wait('@query')
+      cy.wait(1200)
+
+      cy.task('getDownloads', {dirPath: downloadsDirectory}).should(files => {
+        ;(files as string[]).forEach(filename => {
```
rockstar, line 939: Does this semi-colon have some colloquial meaning?

src/shared/components/CSVExportButton.tsx
```diff
@@ -1,72 +1,102 @@
...
+    super(props)
+    this._controller = new AbortController()
+    this._controller.signal.addEventListener(DOWNLOAD_EVENT_COMPLETE, () => {
+      this.setState({state: RemoteDataState.Done})
+    })
+    this.state = {state: RemoteDataState.NotStarted}
+  }
+
   public render() {
+    if (this.state.state == RemoteDataState.Loading) {
```
rockstar, line 42: Yeah, this is a good example of why I'm unsure wrapping a single object in a single object makes sense. `this.state.state` has a weird stutter, and has a "yo dawg" feel about state, e.g. "we heard you like state, so I gave your state some state, so you can check the state of your state."

src/shared/components/CSVExportButton.tsx
```diff
@@ -1,72 +1,102 @@
...
 
-interface StateProps {
-  files: string[] | null
+interface State {
+  state: RemoteDataState
 }
 
-class CSVExportButton extends PureComponent<StateProps, {}> {
+class CSVExportButton extends PureComponent<Props, State> {
+  private _controller: AbortController
```
rockstar, line 30: I've seen this pattern a bit, where we preface private variables with `_`. Is it necessary? In languages that don't have private attrs, it makes sense. In languages that do, usually the pattern I've seen is that the `_` prefix means "I'm not used anywhere, and am just trying to satisfy the compiler."

src/shared/apis/query.ts
```diff
@@ -92,6 +93,25 @@ export const processResponse = async (
   }
 }
 
+export const processResponseBlob = async (
```
rockstar, line 96: There's a bit of misdirection in this API, so it's not clear from this patch: does this mean we're using the `fetch` mechanism (or XHR family of apis) in the browser to do the download? It _appears_ that's so, but I want to be sure.

rockstar (commented):
Have a handful of questions here. Curious about the answers before we go too much further.

cypress/e2e/shared/explorer.test.ts
```diff
@@ -909,4 +910,41 @@ describe('DataExplorer', () => {
...
+        cy.getByTestID('flux-editor', {timeout: 30000}).should('be.visible')
+          .monacoType(`from(bucket: "defbuck")
+  |> range(start: -10h)`)
+        cy.getByTestID('time-machine--download-csv')
+          .should('be.visible')
+          .click()
+      })
+
+      cy.wait('@query')
+      cy.wait(1200)
```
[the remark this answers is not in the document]
wiedld: the `cy.wait('@query')` is for the network call to finish.
the `cy.wait(1200)` is to make sure the write to disk (download) is complete. Cypress does not offer a direct event listener for this (at least, that I could find).

cypress/e2e/shared/explorer.test.ts
wiedld: Semi-colon was automatically added by our linter.

There are reasons why a dev may choose to add this semi-colon themselves (e.g. as a sanity-check that post-javascript parsing, any previous 'bad' script has ended). But I didn't add it myself -- the linter automatically added it, and fails if I removed it.

src/shared/components/CSVExportButton.tsx
```diff
@@ -1,72 +1,102 @@
...
 // Types
-import {AppState} from 'src/types'
+import {AppState, RemoteDataState} from 'src/types'
+
+type Props = ConnectedProps<typeof connector>
 
-interface StateProps {
-  files: string[] | null
+interface State {
+  state: RemoteDataState
```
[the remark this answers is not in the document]
wiedld: Yes, we count set the class prop directly for state.

I was electing to follow standardized react `setState()` approach because we (1) use this approach in this legacy part of the codebase, and (2) it always retriggers a re-render per react's listeners.

src/shared/components/CSVExportButton.tsx
wiedld: Fair enough. Will remove.

src/shared/apis/query.ts
wiedld: Yes, we are.
We are specifically using all the same machineary used when running the normal (on submit) query, except that the post-processing (`processResponse`) is different for the csv download. That way we can:
* not allow any drift from the onSubmit query versus the onDownload query
* while still changing the post-request processing to not truncate for the onDownload.

wiedld: Agreed on the misdirection. This is because of how the existing APIs are structured. So instead of re-writing how the APIs are wired in this legacy part of the codebase, I choose to scope changes to providing a different post-network-call processor.

cypress/e2e/shared/explorer.test.ts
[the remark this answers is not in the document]
rockstar: Sounds like that's a great comment to be added.

cypress/e2e/shared/explorer.test.ts
rockstar: Can you file a bug about this? A linter shouldn't be making _less_ readable code.

src/shared/apis/query.ts
rockstar: I don't think this is what we want. We specifically _don't_ want to use the same machinery here. The browser already has capabilities for downloading this data, handling pauses and cancels, etc. The data here is of indeterminate size, and we don't want to be reading it into the browser.

src/shared/components/CSVExportButton.tsx
```diff
@@ -1,72 +1,102 @@
...
+class CSVExportButton extends PureComponent<Props, State> {
+  private controller: AbortController
+
+  constructor(props) {
+    super(props)
+    this.controller = new AbortController()
+    this.controller.signal.addEventListener(DOWNLOAD_EVENT_COMPLETE, () => {
+      this.setState({state: RemoteDataState.Done})
+    })
+    this.state = {state: RemoteDataState.NotStarted}
```
rockstar, line 38: Why doesn't the linter complain when we set state directly here?

src/shared/components/CSVExportButton.tsx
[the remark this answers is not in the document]
rockstar: In the two places where you call `setState`, you're doing it with `setState({state: ...})`. Why not just `setState(...)`, since it's a single enum value?

cypress/e2e/shared/explorer.test.ts
hoorayimhelping: This isn't a bug, this is an example of Automatic Semicolon Insertion (ASI) from the compiler. This happens because the expression.

```ts
(files as string[])
```

could be interpreted as a function call if a line above it was added that looked like this:

```ts
const foo = bar()
```

which the compiler would interpret as:

```ts
const foo = bar()(files as string[])
```

[This doc](https://prettier.io/docs/en/rationale.html#semicolons) explains it better than I can.

I *think* a simple fix for this would be:

```ts
cy.task('getDownloads', {dirPath: downloadsDirectory}).should((files: string[]) => {
  files.forEach(filename =>
```

rockstar: Ah, so there is a spelling that makes that more clear and eliminates the ASI. Could you make those changes, @wiedld

src/shared/components/CSVExportButton.tsx
wiedld: it's in the constructor. That is permitted.

cypress/e2e/shared/explorer.test.ts
wiedld: `files.forEach(` fails typechecking. Hence why I added the explicit casting...which then the linter required the ASI.

src/shared/apis/query.ts
wiedld: The "same machineary" refers to:
1. how we construct the query to send to the network call
2. the same `runQuery()` method, which every place outside of notebooks/qx-Builder uses to send the actual network call
3. how we notify the user if the network call fails.

I simply added another post-call-processor to the `runQuery()`, which could be optionally passed. This new post-call-processor does not read the data in the browser, but simply downloads it to a blob with a pointer reference.

Are you asking for us to create duplicate code for all 3 steps above, for only the csv download?

wiedld: Also...just as an fyi. The code `const blob = await response.blob()` takes the ReadableStream object (from the browser network call API response.body), and takes the pointed at stream and converts directly into a Blob/File object with a reference. The limits on this file object is the amount of memory available.

See the description here for why this was the best choice I could find, as I could not pipe directly from the Browser ReadableStream into a FileSystem writeable stream: https://github.com/influxdata/ui/pull/6036

wiedld: The only other way to completely bypass the browser is a change server-side, to update this response header for downloads:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

Every other frontend side approach is updating the post-processor.

rockstar: Yeah, I was absolutely expecting this to be the answer: we have to change the server side. We should look at what that takes. We had a implementation/feature here that was a partial download, presumably because we were downloading large files and blowing up the ram requirements of the browser. This patch would re-introduce that possible case.

Can we look at what it would take on the server side? I expect we'll have to make changes to Query™️.

src/shared/components/CSVExportButton.tsx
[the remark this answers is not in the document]
wiedld: React API wants an object. And more specifically, if we try using a non-object as the state, it will fails type checking. Specifically because of this:
Screen Shot 2022-10-19 at 1 10 35 PM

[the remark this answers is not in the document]
wiedld: But I can change the state attribute name, to avoid `this.state.state`

src/shared/components/CSVExportButton.tsx
wiedld: Renamed to `loadingState`, to avoid the stutter.

cypress/e2e/shared/explorer.test.ts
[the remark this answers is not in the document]
wiedld: no longer relevant. line removed.

src/shared/apis/query.ts
wiedld: Made it work with direct filesystem download.

wiedld:
Have it working with direct filesystem download:
(note: requires [this PR in idpe](https://github.com/influxdata/idpe/pull/16212) to merge first.)


https://user-images.githubusercontent.com/10232835/197157839-a8c38d55-a3be-4b09-93b5-3455ce64bab4.mov

wiedld:
Since it's taking awhile to get the backend change in, and the backend change is only setting the name of the file downloaded...maybe we should update the tests and merge the UI work now?

The UI tests would then pass, if we update it to only look for for `query.csv` file downloaded.

rockstar:
I've had this tab open for a few days now, waiting for the dependent patches to go through. I'd like to see them merged and this patch be based on what's in the production, as I'm expecting some changes related to the rubber meeting the road.

src/shared/workers/downloadHelper.ts
```diff
@@ -0,0 +1,69 @@
+/*
+    Background:
```
wiedld, line 2: For the new/old batch of reviewers:
*this PR has been open for awhile, since the backend changes took time.
* This PR is ready now, but there is a whole history of why this approach was chosen. This in-code comment describes the why.
* Additionally, spun up a minimal js fullstack app to test other approaches (where I could ctrl the full stack without any server middleware doing header stripping) -- and this was still the only approach found which worked.

Feel free to suggest/discover other approaches. Open to other options, as long as it works!

src/shared/workers/downloadHelper.ts
```diff
@@ -0,0 +1,69 @@
...
+        * have a form submission request:
+            * the XMLhttpRequest does not directly download, even with the appropriate header.
+            * cannot use the `a` tag download feature, as that is only GET requests
+            * a form submit is build into the browser to allow downloads. **this is the only approach**
+
+    To get the form submit to have a http request which looked like our normal query requests:
+        * we used a service worker, which can be an interceptor/listener for form submissions.
+            * forms themselves do not allow headers to be attached.
+        * then modify the headers, and parse the body to the appropriate format
+*/
```
hoorayimhelping, line 24: very helpful comment :+1:

hoorayimhelping (approved):
The comments on the PR are helpful, and the code is well-written 👍 

There's nothing here that's a hard blocker for me. I had some questions and light suggestions though, but since you have deeper understanding than I do, I'll let you decide whether they're worth implementing or skipping

src/shared/components/CSVExportButton.tsx
```diff
@@ -1,72 +1,90 @@
...
+class CSVExportButton extends PureComponent<Props, State> {
+  constructor(props) {
+    super(props)
+    this.state = {isEnabled: false}
+
+    if ('serviceWorker' in navigator) {
+      navigator.serviceWorker.register(downloadWorker).then(
+        () => this.setState({isEnabled: true}),
+        function (err) {
+          console.error('ServiceWorker registration failed: ', err)
```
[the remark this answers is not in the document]
wiedld: so I wasn't sure for this one. Since it doesn't appear as an error to the user; instead, the user will simply not have the option to download (the `this.state.isEnabled` = false).

The purpose of the console.error, is in case we have a user complain that they don't have the download button. Then we can ask them to check their console, and give a reason why.

src/shared/components/CSVExportButton.tsx
```diff
@@ -1,72 +1,90 @@
...
+
     return (
       <Button
         titleText={this.titleText}
         text="CSV"
         icon={IconFont.Download_New}
         onClick={this.handleClick}
-        status={this.buttonStatus}
+        status={
+          this.props.disabled
```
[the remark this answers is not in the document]
wiedld: `props.disabled` is the parent deciding any reason to disable the button. For example, if the user has not created a query yet.

`this.state.isEnabled` is about if the user can use the csv button in general...specifically (for this use case), if their browser has service workers enabled.

Great idea on the better naming. Will choose `this.state.browserSupportsDownload`.

src/shared/workers/downloadHelper.ts
```diff
@@ -0,0 +1,69 @@
...
+
+self.addEventListener('fetch', function (event: any) {
+  const contentType = (event.request.headers as Headers).get('Content-Type')
+  if (
+    new URL(event.request.url).pathname == '/api/v2/query' &&
+    contentType == 'application/x-www-form-urlencoded'
+  ) {
+    const headers = new Headers()
+    for (const pair of event.request.headers) {
```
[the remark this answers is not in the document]
wiedld: excellent idea. done.

src/timeMachine/actions/queries.ts
```diff
@@ -244,15 +250,54 @@ export const setQueryByHashID = (queryID: string, result: any): void => {
...
+  queryText: string,
+  state: AppState,
+  abortController: AbortController
+) => {
+  const allBuckets = getAll<Bucket>(state, ResourceType.Buckets)
+  const allVariables = getAllVariables(state)
+
+  event('executeQueries query', {}, {query: queryText})
+
+  const orgID = getOrgIDFromBuckets(queryText, allBuckets) || getOrg(state).id
```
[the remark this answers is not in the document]
wiedld: This pattern (of precedence of orgId) was existing in the code. I'm unclear as to the history for why (it was added [with this commit from 3 years ago](https://github.com/influxdata/ui/commit/45d1b2d30a5f0481a5db56bb9a31e3d07cbf2e3b))...so I made the call to leave in.

Do you know the history? Shall I just remove `getOrgIDFromBuckets()` method in general? It's used in a total of 4 places.

src/timeMachine/actions/queries.ts
```diff
@@ -370,6 +385,52 @@ export const executeQueries =
...
+        extern,
+        dialect: {annotations: ['group', 'datatype', 'default']},
+      })
+    )
+    hiddenForm.appendChild(input)
+    document.body.appendChild(hiddenForm)
+
+    hiddenForm.submit()
+  } catch (error) {
+    console.error(error)
```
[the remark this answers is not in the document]
wiedld: Is actionable. Did toast.

src/timeMachine/actions/queries.ts
```diff
@@ -370,6 +385,52 @@ export const executeQueries =
     }
   }
 
+export const runDownloadQuery = () => async (dispatch, getState: GetState) => {
+  const state = getState()
+  const queryText = getActiveQuery(state).text
+
+  try {
+    await dispatch(hydrateVariables())
```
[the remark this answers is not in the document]
wiedld: variable hydration occurs in 3 places (as of right now):
* when executing a timeMachine query
* when downloading a csv from timeMachine.
   * (note: this can be done without running a query first)
* when opening the variables toolbar.

Since the csv download can occur seperately (without a precursor) of the other 2 steps, then IMO we do need to hydrate here.

src/timeMachine/actions/queries.ts
```diff
@@ -244,15 +250,54 @@ export const setQueryByHashID = (queryID: string, result: any): void => {
...
+
+export const runTimeMachineQuery = (
+  queryText: string,
+  state: AppState,
+  abortController: AbortController
+) => {
+  const allBuckets = getAll<Bucket>(state, ResourceType.Buckets)
+  const allVariables = getAllVariables(state)
+
+  event('executeQueries query', {}, {query: queryText})
```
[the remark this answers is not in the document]
wiedld: This is moved code ([from here](https://github.com/influxdata/ui/blob/93002cdad8692f470353c6355f284a66ae84e785/src/timeMachine/actions/queries.ts#L278)), and uses the current event naming. Please let me know if there are reasons to change/update.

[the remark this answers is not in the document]
hoorayimhelping: It's probably fine to leave as is. I'm not sure if we're going to move forward with the new convention or not, since that was something Kristina and Fabian were implementing.

For context, when I was building the app-metrics system, I had always assumed that we'd use something like `location.action.outcome` to name events, so in this case an event might be named `'timeMachine.query.execute'`, since a lot of places I've worked have a similar naming convention because it provides context to people analyzing the data. Alex didn't want to do that and insisted that we do it in a more sentence-style way because he preferred it. After Kristina joined, she wanted to use a similar system to the one I described. Alex didn't want to and just ignored that and kept doing them (and kept the QX team doing them) in his preferred way. So over the past few months before the layoff, Kristina was working with Fabian to try to get all our event names consistent, because Fabian preferred the system Kristina and I wanted to use becaue it was easier for him to work with, as I understand it.

src/timeMachine/actions/queries.ts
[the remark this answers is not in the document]
hoorayimhelping: >Do you know the history?

I don't klnow the history. I think we should simplify this in whatever way makes the most sense to you. I also don't think you should up this PR to do this cleanup / refactor. I think the real win here is realizing where this came from and that it's only used in 4 places.

src/timeMachine/actions/queries.ts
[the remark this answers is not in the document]
hoorayimhelping: Makes sense! Thank you for explaining that to me :)

src/shared/components/CSVExportButton.tsx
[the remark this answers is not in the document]
hoorayimhelping: Thanks for the explanation. It's a non-actionable error, so there's not really a good reason to show the user a toast. Makes perfect sense.

In this case, I'm starting to lean more and more on the idea that we should send this error to Honeybadger, rather than asking the user to look at their console. The error is less ephemeral when it's in Honeybadger, so we can reference it later. It's kind of amateurish to ask someone paying us money to use our product and get support to open up their console and tell us what went wrong. It's also possible that they had the error, reported it, and it's gone now and they don't know or don't want to recreate it.

hoorayimhelping (approved):
I think we've deliberated on this enough. Let's light this candle, and if anything still needs to be tweaked (like a honeybadger error vs a console error, or removing a selector) you can follow up on this ticket with it.

wiedld:
Already merged. Removing from queue and manually closing.

See here:
Screen Shot 2022-12-01 at 3 27 30 PM

PR #6075, opened 2022-10-14
chore: fix broken checkout test and related form validation issue

wdoconnell:
Needs to be reviewed with `?w=1` - otherwise makes it look like more lines are changed. Should see this.

Screen Shot 2022-10-14 at 4 53 49 PM

This PR started as an effort to remove a flag toggle from the checkout test. Making edits to that individual test - or even making no changes to the test - resulted in a CI test failure, which didn't make sense. The cause is that the test was already broken, but wasn't being run by our CI pipeline (unless you made edits to this test specifically) because it was in a separate subfolder that wasn't being run by the test runner.

This PR resolves the first part of the issue by fixing the broken test, by removing the `required` from an `<Input />` tag. The `required` was supposed to be present only in the parent component, as having `required` in the child component prevents the form `onSubmit` from running, which prevents the checkout validation from working correctly. I will make a second PR moving the test into the correct folder.

Here's the way the test behaves before (broken form).
--
https://user-images.githubusercontent.com/91283923/195872495-77aa44b1-c32c-452e-8ddb-56ab48d4e15f.mov

Here's the way the test behaves now.
--
https://user-images.githubusercontent.com/91283923/195872600-0bc6e789-cda5-4c96-b19a-ea3b016f5430.mov

### Checklist

### DIFF
--- a/cypress/e2e/cloud/checkout/checkoutpage.test.ts
+++ b/cypress/e2e/cloud/checkout/checkoutpage.test.ts
@@ -27,135 +27,127 @@ describe('Checkout Page Works', () => {
   )
 
   it('should render the checkout page and allow for pointing and clicking', () => {
-    cy.setFeatureFlagsNoNav({
-      multiOrg: true,
-    }).then(() => {
-      const email = 'asalem@influxdata.com'
-      const limit = 10
-      const numberError = 'Please enter a value of 1 or greater'
-      const genericError = 'This is a required field'
+    const email = 'asalem@influxdata.com'
+    const limit = 10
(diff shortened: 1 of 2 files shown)


hoorayimhelping (dismissed):

wdoconnell:
FYI - I'm going to split this into two PRs to preserve history (1 PR fixing the test, and 1 PR moving it into the correct folder). Otherwise, it'll be hard to trace what happened here later because the diff will make it look like a new test.

hoorayimhelping (approved):

Issue #5932, opened 2022-09-29
Multi Org Phase 2 - Add GET /api/v2private/clusters to UI multi org API Layer


wdoconnell:
Add support and error handling for /api/v2private/clusters to the UI to support org creation modal. This route is used to retrieve the list of available clusters in which the org can be created.

See [Quartz 6241](https://github.com/influxdata/quartz/issues/6241) generally for the endpoints created by AIM in connection with this epic.

PR #6081, opened 2022-10-14
chore: add quartz getClusters to API layer

wdoconnell:
Helps with #5932 

Adds UI API layer support for the quartz /clusters endpoint. This retrieves the list of clusters available to the user, so that they can be made available as options when creating a new org.

Not used anywhere yet.

Matching openAPI spec [here](https://github.com/influxdata/openapi/blob/master/src/unity/paths/clusters.yml).

Proof (response from endpoint)
-

### Checklist

### DIFF
--- a/src/identity/apis/createOrg.ts
+++ b/src/identity/apis/createOrg.ts
@@ -1,0 +1,19 @@
+// API Calls
+import {getClusters} from 'src/client/unityRoutes'
+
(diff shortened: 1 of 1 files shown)


wdoconnell:
Closing this in favor of #6082 because they belong and are reviewable in the same PR.

Issue #5894, opened 2022-09-26
Multi Org Phase 2 - Add API layer to support createOrg route


wdoconnell:
Part of #5205. Need API layer to handle responses and errors from createOrg endpoint.

wdoconnell:
Closed by [quartz-mock #36](https://github.com/influxdata/quartz-mock/pull/36)

PR #6082, opened 2022-10-14
chore: add quartz POST /orgs and GET /clusters routes to API layer

wdoconnell:
Helps with #5932, #5894

Adds UI API layer support for the quartz endpoints for creating orgs (`POST` /orgs) and getting the list of available clusters in which an org can be created (`GET` /clusters).

Matching openAPI specs: [create orgs](https://github.com/influxdata/openapi/blob/master/src/unity/paths/orgs.yml), [get clusters](https://github.com/influxdata/openapi/blob/master/src/unity/paths/clusters.yml).

Refactors /identity/apis into three files for readability: (1) auth.ts (apis related to user's identity); (2) account.ts (apis related to user's account); (3) org.ts (apis related to user's org).

Create Orgs Response
--
In quartz-remocal, we get a `401` as expected, as the default free account cannot create more than one org.
Screen Shot 2022-10-14 at 5 11 27 PM

Get Clusters Response
--

### Checklist

### DIFF
--- a/src/identity/apis/account.ts
+++ b/src/identity/apis/account.ts
@@ -1,0 +1,95 @@
+// APIs
+import {
+  getAccount,
+  getAccounts,
+  patchAccount,
+  putAccountsDefault,
+  Account,
+  IdentityAccount,
+  UserAccount,
+} from 'src/client/unityRoutes'
+
+// Types
+import {
+  GenericError,
+  ServerError,
+  UnauthorizedError,
+  UnprocessableEntityError,
+} from 'src/types/error'
+
+export interface CurrentAccount extends IdentityAccount {
+  billingProvider?: 'zuora' | 'aws' | 'gcm' | 'azure'
+}
+
+// get a list of the user's accounts
+export const getUserAccounts = async (): Promise<UserAccount[]> => {
+  const response = await getAccounts({})
+
+  if (response.status === 401) {
+    throw new UnauthorizedError(response.data.message)
+  }
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+  if (!Array.isArray(response.data)) {
+    throw new GenericError('No account found')
+  }
+
+  return response.data
+}
+
+// fetch more details about one of the user's accounts
+export const fetchAccountDetails = async (
+  accountId: string | number
+): Promise<Account> => {
+  const accountIdString = accountId.toString()
+
+  const response = await getAccount({
+    accountId: accountIdString,
+  })
+
+  if (response.status === 401) {
+    throw new UnauthorizedError(response.data.message)
+  }
+
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+
+  const accountDetails = response.data
+  return accountDetails
+}
+
+// change the user's default account
+export const updateDefaultQuartzAccount = async (
+  accountId: number
+): Promise<void> => {
+  const response = await putAccountsDefault({
+    data: {
+      id: accountId,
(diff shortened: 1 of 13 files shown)


src/identity/apis/createOrg.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (response.status === 401) {
+    throw new UnauthorizedError(response.data.message)
+  }
+
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+
+  return response.data
+}
```
hoorayimhelping, line 67: is this where all the org api functionality will be? if so, maybe just call this file `src/identity/apis/org.ts`

src/identity/apis/createOrg.ts
```diff
@@ -0,0 +1,67 @@
...
+  postOrg,
+  OrganizationCreateRequest,
+} from 'src/client/unityRoutes'
+
+// Error Types
+import {
+  ServerError,
+  UnauthorizedError,
+  UnprocessableEntityError,
+} from 'src/identity/apis/auth'
```
hoorayimhelping, line 13: follow up work on this: it's probably time to put these in their own area for error handling since they're starting to be used outside of the identity auth api layer

hoorayimhelping (approved):

src/identity/apis/createOrg.ts
wdoconnell: There are other org related functions in the identity api file (auth.ts).  But at this point that file has grown oversize with everything included in it. 

You'll see a bigger diff here shortly to reflect the fact that I'm going to split auth.ts up into three separate files, plus a separate file for the types.

src/types/error.ts
```diff
@@ -0,0 +1,55 @@
...
+  UnprocessableEntityError = 'UnprocessableEntity', // 422
+  ServerError = 'ServerError', // 500
+  GenericError = 'GenericError',
+}
+
+// 401 error
+export class UnauthorizedError extends Error {
+  constructor(message) {
+    super(message)
+    this.name = 'UnauthorizedError'
```
subirjolly, line 14: I am not sure what this is being used for. I am guessing for response. If not then, I think we should use `xyz.constructor.name`. If this is being used in response, then we can just do `this.constructor.name` instead of hard-coding.

src/identity/apis/account.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  if (response.status === 401) {
+    throw new UnauthorizedError(response.data.message)
+  }
+
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+
+  const accountDetails = response.data
```
subirjolly, line 59: `nit`: feel free to ignore but I think this is kind of unnecessary.

src/types/error.ts
wdoconnell: Sorry - what would you propose changing this to? Won't referencing the `constructor` property return a reference to the constructor function itself, not the primitive value of name?

src/identity/apis/account.ts
wdoconnell: If we edit further, I'll implement. 👍  This is part of the existing account API, which I didn't make changes to.

src/types/error.ts
subirjolly:
```
this.name = this.constructor.name
```

wdoconnell:
Closing this to avoid confusion - splitting this into two PRs, first the refactor, then the new /clusters and /orgs routes.

Issue #6063, opened 2022-10-13
Composition: support import statements


wiedld:
As a precursor to the Pivot Toggle work.
Since the `schema.fieldsAsCols()` inserts an import statement. 

## Bug:
As a general rule, we want to be able to support import statements above the composition block.

Here is the current error, showing that the flux sync fails whenever an import statement is added:

https://user-images.githubusercontent.com/10232835/195709603-53e94c3b-f013-4e48-8031-dea4421e05d8.mov

Track down and include support for composition blocks, with import statements, in the flux-lsp and UI.

wiedld:
3 places need to all be compatible with the import statement:
- [x] LSP composition mod
    * https://github.com/influxdata/flux-lsp/pull/596
- [ ] LSP server applyEdit construction
    * TODO (confirmed does require update)
- [ ] UI composition block detection (e.g. styling, and divergence)
    * draft PR in progress

wiedld:
The UI portion is merged in. The LSP portion is blocked on other work. Going to unassign myself for now, since have to wait until the dependent LSP changes are up.

PR #6088, opened 2022-10-16
feat(6063): suppport import statements with composition

wiedld:
Part of #6063 

## Bug:
* onChange() detect:
  * occurs **after** change is applied.
  * the change object's positional information (range lines), are based on the previous editor value **before the change** was applied
* solution:
  * to really see if change occurred within the composition block, we need to compare to the snapshot editor text from **before** the change.

 ## Vid:

https://user-images.githubusercontent.com/10232835/195953925-6fc436bf-8d5e-40c7-a6d4-cd28dc082716.mov

   

## Checklist
- [x] ~Feature flagged, if applicable~ Already behind flag.

### DIFF
--- a/cypress/e2e/shared/fluxQueryBuilder.test.ts
+++ b/cypress/e2e/shared/fluxQueryBuilder.test.ts
@@ -491,6 +491,18 @@ describe('Script Builder', () => {
             'disabled'
           )
 
+          cy.log(
+            'does not diverge, when adding import statement outside of block'
+          )
+          cy.getByTestID('flux-toolbar-search--input')
+            .should('exist')
+            .type('schema')
+          cy.getByTestID('flux--fieldsAsCols').should('exist').click()
+          cy.getByTestID('flux-sync--toggle').should(
+            'not.have.class',
+            'disabled'
+          )
+
           cy.log('does diverge, within block')
           cy.getByTestID('flux-editor').monacoType(
             '{enter}{upArrow}{upArrow}{upArrow} // make diverge'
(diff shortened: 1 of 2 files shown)


cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -491,9 +491,27 @@ describe('Script Builder', () => {
...
+          cy.getByTestID('flux-editor').contains('import')
+          cy.getByTestID('flux-sync--toggle').should(
+            'not.have.class',
+            'disabled'
+          )
+
+          /// FIXME: Does not work yet, since the LSP response for applyEdit starts at line 0
+          // cy.log(
+          //   'does not diverge, when further modifying block (with imports at top)'
+          // )
```
appletreeisyellow, line 510: Just curious, is this intended for another PR? If the test doesn't work yet, is there a bug?

Or it is done already? I tested on my local machine, it looks good

https://user-images.githubusercontent.com/14298407/196198408-09e08c87-1357-4853-b8a0-4b63f81a0fb7.mov

wiedld: It needs to be another PR. The remaining work (referenced in that `cy.log()`) is that if we add an import statement, then use the schemaBrowser again --> it will recreate the composition block at line 1.


https://user-images.githubusercontent.com/10232835/196234967-eac172d9-ab89-4f13-8a3a-b2c5eec7eb4d.mov




Update for that needs to be in the lsp, since it always returns an `applyEdit` message to the monaco-editor with the replacement line starting at the first line.

appletreeisyellow (approved):

cypress/e2e/shared/fluxQueryBuilder.test.ts
wiedld: Your question brings up an excellent point. This next bit of work (in the flux-lsp) should probably merge before this one does. Because the current prod (before this PR) just disables the sync once we use imports. So we probably want to leave that behavior until the lsp update goes in too.

PR #6108, opened 2022-10-18
feat: add createNewOrg and fetchClusters to API layer

wdoconnell:
Helps with #5932, #5894

Adds UI API layer support for the quartz endpoints for creating orgs (`POST` /orgs) and getting the list of available clusters in which an org can be created (`GET` /clusters).

Matching openAPI specs: [create orgs](https://github.com/influxdata/openapi/blob/master/src/unity/paths/orgs.yml), [get clusters](https://github.com/influxdata/openapi/blob/master/src/unity/paths/clusters.yml).

Create Orgs Response
--
In quartz-remocal, we get a `401` as expected, as the default free account cannot create more than one org.
Screen Shot 2022-10-14 at 5 11 27 PM

Get Clusters Response
--

### Checklist

### DIFF
--- a/src/identity/apis/org.ts
+++ b/src/identity/apis/org.ts
@@ -2,9 +2,12 @@
 import {
   getAccounts,
   getAccountsOrgs,
+  getClusters,
   getOrg,
+  postOrg,
   putAccountsOrgsDefault,
   Organization,
+  OrganizationCreateRequest,
 } from 'src/client/unityRoutes'
 
 // Types
@@ -12,6 +15,7 @@ import {RemoteDataState} from 'src/types'
 import {
   GenericError,
   NotFoundError,
+  OrgNameConflictError,
   ServerError,
   UnauthorizedError,
   UnprocessableEntityError,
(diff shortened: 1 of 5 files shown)


TCL735 (dismissed):

subirjolly (approved):

src/identity/apis/org.ts
```diff
@@ -34,6 +38,55 @@ export type QuartzOrganizations = {
...
+  if (response.status === 500) {
+    throw new ServerError(response.data.message)
+  }
+
+  // Status code in response when successfully creating an org is 201.
+  return response.data
+}
+
+// fetch the list of clusters in which an org can be created
+export const fetchClusterList = async () => {
```
hoorayimhelping, line 76: this method is called `fetchClusterList` but below we have `getDefaultAccountDefaultOrg`, but below that, we also have `fetchOrbsByAccountID`. can we make `fetch` and `get` verbiage consistent here? or point out what is different about them?

wdoconnell: I'll make an edit to change this to `fetch` throughout, which is probably clearer:
https://stackoverflow.com/questions/2141818/method-names-for-getting-data

> Example: [fetch](https://www.google.com/search?q=define%20fetch) implies that some entity needs to go and get something that is remote and bring it back. Dogs fetch a stick, and [retrieve](https://www.google.com/search?q=define%20get) is a synonym for fetch with the added semantic that you may have had possession of the thing prior as well. [get](https://www.google.com/search?q=define%20get) is a synonym for [obtain](https://www.google.com/search?q=define%20obtain) as well which implies that you have sole possession of something and no one else can acquire it simultaneously.

TCL735 (approved):

Issue #6109, opened 2022-10-18
Bug Duty: Go through the stats page on CircleCI insights and fix the most frequently flaking tests


ALuckette:
Go through the stats page on CircleCI insights and fix the most frequently flaking tests

wdoconnell:
Resolved. CI passing consistently and tests re-added here. https://github.com/influxdata/ui/issues/6562

PR #6111, opened 2022-10-18
fix(about): target parent element to access content

blegesse-git:
Part of #6109 

Pr fixes a flaky test  in shared folder -  `about` e2e test. 
Flakiness is caused by failure to find text content `Cloud Provider` within DOM element. 

Fix: target direct parent element of text content.
Screen Shot 2022-10-18 at 10 19 04 AM

### Checklist

### DIFF
--- a/src/organizations/components/OrgProfileTab/index.tsx
+++ b/src/organizations/components/OrgProfileTab/index.tsx
@@ -80,7 +80,7 @@ const OrgProfileTab: FC = () => {
             justifyContent={JustifyContent.SpaceBetween}
             stretchToFitWidth={true}
             style={orgProviderExists ? {width: '85%'} : {width: '48%'}}
-            testID="org-profile--labeled_data"
+            testID="org-profile--labeled-data"
           >
             {orgProviderExists && (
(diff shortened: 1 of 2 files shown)


src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -80,6 +80,7 @@ const OrgProfileTab: FC = () => {
             justifyContent={JustifyContent.SpaceBetween}
             stretchToFitWidth={true}
             style={orgProviderExists ? {width: '85%'} : {width: '48%'}}
+            testID="org-profile--labeled_data"
```
hoorayimhelping, line 83: For consistency sake, I think this should be `org-profile--labeled-data` (dash, instead of underscore)

hoorayimhelping (commented):
Hate to do this, but I think the underscore in the test id should be a dash

src/organizations/components/OrgProfileTab/index.tsx
blegesse-git: you're totally right.

PR #6113, opened 2022-10-18
test(operator): provide orgid to hit NotFound page

### LINKED ISSUE #6109: Bug Duty: Go through the stats page on CircleCI insights and fix the most frequently flaking tests

blegesse-git:
Closes #6109 

https://user-images.githubusercontent.com/66275100/196504286-754c6408-b301-4abc-8883-51716d38857d.mov

There is a useEffect hook inside `SetOrg.tsx` that redirects user to `/no-orgs` page if `orgID` from url doesn't match any orgs that exist. 
When Operator test is run in circle ci, and it tries to access `/operator`, this useEffect hook intervenes and re-routes page to `/no-orgs` since the url at that instance doesn't have any orgs or orgID. 

Pr tries to fix this by providing an org and org id in the url, so it can properly get routed to `NotFound` page. 

### Checklist

### DIFF
--- a/cypress/e2e/cloud/operator.test.ts
+++ b/cypress/e2e/cloud/operator.test.ts
@@ -1,3 +1,5 @@
+import {Organization} from '../../../src/types'
+
 describe('Operator Page', () => {
   beforeEach(() =>
     cy.flush().then(() =>
@@ -239,12 +241,14 @@ describe('Operator Page should not be accessible for non-operator users', () => 
   beforeEach(() =>
     cy.flush().then(() =>
       cy.signin().then(() => {
-        cy.get('@org').then(() => {
-          cy.getByTestID('home-page--header').should('be.visible')
-          cy.quartzProvision({
-            isOperator: false,
-          }).then(() => {
-            cy.visit(`/operator`)
+        cy.get('@org').then(({id}: Organization) => {
+          cy.fixture('routes').then(({orgs}) => {
+            cy.getByTestID('home-page--header').should('be.visible')
+            cy.quartzProvision({
+              isOperator: false,
+            }).then(() => {
+              cy.visit(`${orgs}/${id}/operator`)
+            })
           })
         })
       })


TCL735 (approved):

cypress/e2e/cloud/operator.test.ts
```diff
@@ -239,12 +241,14 @@ describe('Operator Page should not be accessible for non-operator users', () =>
...
-          }).then(() => {
-            cy.visit(`/operator`)
+        cy.get('@org').then(({id}: Organization) => {
+          cy.fixture('routes').then(({orgs}) => {
+            cy.getByTestID('home-page--header').should('be.visible')
+            cy.quartzProvision({
+              isOperator: false,
+            }).then(() => {
+              cy.visit(`${orgs}/${id}/operator`)
+            })
```
wdoconnell, line 251: What's the rationale behind changing the url to be visited here - was that also incorrect?  I just want to make sure we're not triggering a 404 for the wrong reason - namely, that this page never exists.

wdoconnell (commented):
I might be misunderstanding: but isn't the `not-found` intentional?

cypress/e2e/cloud/operator.test.ts
wdoconnell: Screen Shot 2022-10-18 at 2 09 59 PM

What I mean is - I think the url `/operator` - as opposed to `${orgs}/${id}/operator` - is the correct URL we're trying to hit, according to getOrganizations.

blegesse-git: Technically, I didn't change the url. We are still visiting `/operator`. The only thing added is we're now providing `orgs` and `orgID` in the link. The goal of the test is to make sure users without proper permission can't access the operator's page. When they do, they will get a 404 page. While the logic in the test is correct, I think what was causing the failures was this code: 
Screen Shot 2022-10-18 at 1 14 04 PM
Instead of routing user's without operator permission to a 404 page, line 112 returns true and re-routes user to `no-orgs`. Test then fails because it's expecting a 404 page

wdoconnell: I could be totally missing something here - and apologies if I am - but isn't this change making the test visit a different URL?
Screen Shot 2022-10-18 at 2 20 24 PM

blegesse-git: If you run the test locally, you'll see there is no flaky behavior. I think that's because locally, when visiting /operator, the url contains both org and orgID. But for a reason I don't quite understand yet, circle ci url merely contains /operaor with no org and orgID.

blegesse-git:
Flakiness of test was due to quartz-mock not providing a default account. ([PR](https://github.com/influxdata/quartz-mock/pull/38) addressing this issue). UI code has logic that re-routes user to `/no-orgs` page when it's unable to find a default account. The quartz-mock pr mentioned fixes this issue by making sure request to `/api/accounts` always returns a default account. I was able to confirm this solution works by intercepting the GET request to `/api/accounts` and sending back a response with `isDefault: true`, which fixed the flakiness seen in `operator` test.

Issue #6121, opened 2022-10-19
Multi Org Phase 2: Add /allowances/orgs/create to the UI


wdoconnell:
In [openAPI #573](https://github.com/influxdata/openapi/pull/573), AIM specced out a new endpoint that provides the UI with information about whether or not the user can create more orgs, and whether the user has account upgrades available. This needs to be added to quartz-mock and the UI API layer before it can be used.

NOTE: this endpoint does not yet exist in prod - [AIM is currently creating the actual endpoint](https://github.com/influxdata/quartz/issues/6725). So any components relying on a response from this endpoint should not be released until that work is complete.

PR #6122, opened 2022-10-19
feat: update openAPI SHA and add fetchOrgCreationAllowance to UI

wdoconnell:
Helps with #6121 

Updates the openAPI SHA to pull in the [spec for getAllowancesOrgsCreate](https://github.com/influxdata/openapi/commit/f7cc87286abffcea007df69e74b4162f9f1790b3), and adds it to the UI API layer.

Not used, and AIM hasn't added this endpoint to quartz yet. It will be mocked in quartz-mock so we can proceed with UI development until the endpoint is ready.

### Checklist

- [X] Feature flagged, if applicable - NO - NOT YET USED

### DIFF
--- a/src/identity/apis/org.ts
+++ b/src/identity/apis/org.ts
@@ -2,6 +2,7 @@
 import {
   getAccounts,
   getAccountsOrgs,
+  getAllowancesOrgsCreate,
   getClusters,
   getOrg,
   postOrg,
@@ -115,6 +116,19 @@ export const fetchDefaultAccountDefaultOrg = async (): Promise<
     }
     throw new GenericError('No default account found')
   }
+}
+
+// fetch data regarding whether the user can create new orgs, and associated upgrade options.
+export const fetchOrgCreationAllowance = async () => {
+  const response = await getAllowancesOrgsCreate({})
+
+  if (response.status !== 200) {
+    throw new GenericError(
+      'Failed to determine whether this user can create a new organization.'
+    )
+  }
+
+  return response.data
 }
 
 // fetch the list of organizations associated with a given account ID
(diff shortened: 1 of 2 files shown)


src/identity/apis/org.ts
```diff
@@ -117,6 +118,19 @@ export const fetchDefaultAccountDefaultOrg = async (): Promise<
...
+
+  if (response.status !== 200) {
+    throw new GenericError(
+      'Failed to determine whether this user can create a new organization.'
+    )
+  }
+
+  return response.data
+}
+
```
wdoconnell, line 133: Calling this an `allowance` based on some discussion with Randy about the right nomenclature. Calling it a 'permission' is a little misleading because it's not a roles-and-permissions issue, nor is it really an entitlement. "Am I allowed to do X" is a simpler and easier way to characterize this.

Note that what it returns is an object with two `boolean`s - `allowed` and `upgradesAvailable`.

src/identity/apis/org.ts
```diff
@@ -117,6 +118,19 @@ export const fetchDefaultAccountDefaultOrg = async (): Promise<
...
 
+// fetch data regarding whether the user can create new orgs, and associated upgrade options.
+export const fetchOrgCreationAllowance = async () => {
+  const response = await getAllowancesOrgsCreate({})
+
+  if (response.status !== 200) {
+    throw new GenericError(
+      'Failed to determine whether this user can create a new organization.'
+    )
+  }
```
wdoconnell, line 129: I'll check with Randy on whether there should be any other error codes when the endpoint is created. Right now it's just a `200`, but I would assume we should ultimately be able to get at least a `401` if the user isn't logged in, or a `500` if there is some unforeseen server issue.

TCL735 (approved):

PR #6124, opened 2022-10-19
feat: allow create org for multiorg phase 2

### LINKED ISSUE #5899: Multi Org Phase 2 - add new modal and functionality to create organization
Mock: https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1967%3A20534

Add new modal according to the figma mockups to create an organization.

Details:
* Perform duplicate org name check using organizations list in identity.
* Update openapi sha to use the Post Organization route.

Pending Questions for @juliajames12 :
- If there's a conflict in the name of the organization when creating one, do we still display a Toast with Retry or should we display an inline error under the org name textbox?
- Do we allow `Cancel` while creating an organization?
- Do we straight-away go to the new organization page or should we just display the Toast with `Switch to Organization` link as displayed in the mockup?

subirjolly:
Closes #5899 

## 2. SIMPLE INPUT VALIDATION
2 simplevalidation

## 3. DUPLICATE ORGANIZATION CHECK
3 duplicateorgcheck-local

## 4. NO PROVIDERS FETCHED
4 NoProviders

## 5. THREE PROVIDERS AVAILABLE
5 ThreeProviders

## 6. TWO PROVIDERS AVAILABLE
6 TwoProviders

## 7. SINGLE PROVIDER AVAILABLE
7 SingleProvider

## 8. 422 ERROR CODE AS NAME CONFLICT(DUPLICATE ORG)
8 422ConsideredAsNameConflict

## 9. ERROR FETCHING CLUSTERS
9 ErrorFetchingClusters

## 10. CREATE ORG - 401 ERROR
10 CreateOrg401Error

## 11. CREATE ORG - 409 ERROR
11 CreateOrg409Error

## 12. CREATE ORG - 422 ERROR
12 CreateOrg422Error

## 13. CREATE ORG - 500 ERROR
13 CreateOrg500Error

### Checklist


src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.tsx
```diff
@@ -0,0 +1,86 @@
+// Libraries
+import React, {FC, useCallback, useContext, useMemo} from 'react'
+import classnames from 'classnames'
+import {FlexBox, FlexDirection} from '@influxdata/clockface'
+
+// Style
+import './ClusterBox.scss'
+
+// Components
+import {CreateOrgContext, ProviderID} from './CreateOrganizationContext'
```
wdoconnell, line 10: I would consider alphabetizing these

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+interface OwnProps {
+  showLogoWithText: Boolean
+}
+
+type Props = OwnProps & Cluster
+
+// TODO(Subir): Please suggest a better name
+export const ClusterBox: FC<Props> = ({
+  providerId,
```
wdoconnell, line 40: I think this name is fine.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.tsx
```diff
@@ -0,0 +1,86 @@
...
+  providerName,
+  showLogoWithText,
+}) => {
+  const {
+    clusters,
```
wdoconnell, line 45: Would alphabetize

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.tsx
```diff
@@ -0,0 +1,86 @@
...
+import {Cluster} from 'src/client/unityRoutes'
+import {AWSLogo} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/AWSLogo'
+import {AzureLogo} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/AzureLogo'
+import {GCPLogo} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/GCPLogo'
+import {AzureLogoWithText} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/AzureLogoWithText'
+import {GCPLogoWithText} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/GCPLogoWithText'
+import {AWSLogoColor} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/AWSLogoColor'
+
+const LOGOS = {
+  AWS: (showLogoWithText: Boolean) =>
```
wdoconnell, line 20: I don't think this object needs to be allcaps

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.tsx
```diff
@@ -0,0 +1,86 @@
...
+      return
+    }
+    changeCurrentProvider(providerId as ProviderID)
+    changeCurrentRegion(clusters?.[providerId]?.[0]?.regionId)
+  }, [
+    clusters,
+    currentProvider,
+    providerId,
+    changeCurrentProvider,
+    changeCurrentRegion,
```
wdoconnell, line 70: would alphabetize

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+  createContext,
+  ReactChild,
+  useState,
+  useEffect,
+  useCallback,
+} from 'react'
+import {Cluster} from 'src/client/unityRoutes'
+import {createNewOrg, fetchClusterList} from 'src/identity/apis/org'
+import {RemoteDataState} from 'src/types'
+
```
wdoconnell, line 13: Would organize the headings here - separate apis from types, libraries, etc.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+  changeOrgName: (_: string) => void
+  clusters: ClusterMap
+  currentProvider: ProviderID
+  changeCurrentProvider: (_: ProviderID) => void
+  currentRegion: string
+  changeCurrentRegion: (_: string) => void
+  createOrgLoadingStatus: RemoteDataState
+  changeCreateOrgLoadingStatus: (_: RemoteDataState) => void
+  createOrg: () => Promise<void>
+}
```
wdoconnell, line 35: Would alphabetize

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+  const changeCreateOrgLoadingStatus = useCallback(
+    (status: RemoteDataState) => {
+      setCreateOrgLoadingStatus(status)
+    },
+    [setCreateOrgLoadingStatus]
+  )
+
+  useEffect(() => {
+    fetchClusterList().then(clusterData => {
+      // setClusters(r?.data)
```
wdoconnell, line 257: We need a catch here to handle errors, and to remove the comments.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+//     isRegionAvailable: true,
+//     priority: 99,
+//     providerId: 'GCP',
+//     providerName: 'Google Cloud',
+//     regionId: 'us-west1',
+//     regionName: 'Oregon',
+//   },
+// ]
+// const FAKE_CLUSTERS: Cluster[] = [...FAKE_AWS, ...FAKE_AZURE, ...FAKE_GCP]
+
```
wdoconnell, line 198: Maybe this is an oafish criticism of mine, but do we need a context here?  This is one overlay - couldn't we simply handle the states we need in the parent component?  It doesn't seem like we have enough components that prop drilling is going to be an issue.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+          (a, b) => a.priority - b.priority
+        )
+      })
+      setClusters(clusterMap)
+      if (!!providers.length) {
+        setCurrentProvider(providers[0] as ProviderID)
+        setCurrentRegion(clusterMap[providers[0]][0].regionId)
+      }
+    })
+  }, [])
```
wdoconnell, line 282: It looks like there are a lot of dependencies here - isn't this going to throw new useEffect warnings?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
wdoconnell: If we do need access to a persistent unified state, we might also used Redux here (which would give us the flexibility to access the state in other parts of the app, and would keep the cluster data alongside the other data retrieved from quartz).

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+      // clusterData = FAKE_CLUSTERS
+
+      const clusterMap = clusterData.reduce((acc, curr) => {
+        if (!acc?.[curr.providerId]) {
+          acc[curr.providerId] = []
+        }
+        acc[curr.providerId].push(curr)
+
+        return acc
+      }, {} as ClusterMap)
```
wdoconnell, line 269: Is the `as` here obscuring a typing problem?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+    (region: string) => {
+      setCurrentRegion(region)
+    },
+    [setCurrentRegion]
+  )
+
+  const createOrg = useCallback((): Promise<void> => {
+    if (createOrgLoadingStatus === RemoteDataState.Loading) {
+      return new Promise(resolve => resolve(null))
+    }
```
wdoconnell, line 230: What's the purpose of this `if` check for the remote data state being `Loading`?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+      orgName,
+      provider: currentProvider,
+      region: currentRegion,
+    })
+      .then(() => {
+        setCreateOrgLoadingStatus(RemoteDataState.Done)
+      })
+      .catch(e => {
+        setCreateOrgLoadingStatus(RemoteDataState.Error)
+
```
wdoconnell, line 243: If we get into an error state, we might want to consider sending a honeybadger error so that we can see if users start hitting this issue.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+      // r
+      // let clusterData = r.data
+      // clusterData = FAKE_CLUSTERS
+
+      const clusterMap = clusterData.reduce((acc, curr) => {
+        if (!acc?.[curr.providerId]) {
+          acc[curr.providerId] = []
+        }
+        acc[curr.providerId].push(curr)
+
```
wdoconnell, line 267: I would change these to more descriptive variables so it's clear what `acc` and `curr` represent. Also, since we're just building a list of clusters, can't we just do this with a loop? I'm finding it difficult to understand what this is doing.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -0,0 +1,128 @@
...
+    useContext(CreateOrgContext)
+  const dispatch = useDispatch()
+  const [errorMessage, setErrorMessage] = useState(DEFAULT_ERROR_MESSAGE)
+  const [showError, setShowError] = useState(false)
+
+  const handleOrgNameChange = useCallback(
+    e => {
+      changeOrgName(e.target.value)
+      setShowError(false)
+      setErrorMessage(DEFAULT_ERROR_MESSAGE)
```
wdoconnell, line 46: Why are we setting `showError` to `false` here?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -0,0 +1,128 @@
...
+    },
+    [changeOrgName]
+  )
+
+  const handleCreateOrg = useCallback(() => {
+    if (orgName.trim() === '') {
+      return
+    }
```
wdoconnell, line 54: Do we want to give feedback to the user instead of just `return`ing if they try to enter an org with no name?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/InputWithLabel.tsx
```diff
@@ -0,0 +1,101 @@
...
+  label,
+  required,
+  onChange,
+  description,
+  value,
+  errorMessage = DEFAULT_ERROR_MESSAGE,
+  showError = false,
+  ...args
+}) => {
+  const [isDirty, setIsDirty] = useState(false)
```
wdoconnell, line 38: What does `isDirty` mean?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/InputWithLabel.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const [inputVal, setInputVal] = useState(value)
+  const [hasError, setHasError] = useState(showError)
+
+  const handleChange = useCallback(
+    e => {
+      const val = e.target.value
+      if (val !== inputVal) {
```
wdoconnell, line 45: I would change val to something more descriptive so that we know what `val` versus `inputVal` are.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.tsx
subirjolly: i planned on using `BillingProvider` as type for keys. But there was a lot of `toLowerCase` and back and forth in typecasting. So, I kept it the way we receive `providerId` in response.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
subirjolly: moved it. will be available in the next commit

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
subirjolly: No. It's not using any value(s) from the context/state

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
subirjolly: as soon as we change org name, I think the error should go away.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
subirjolly: updated. should be available in the next commit.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/InputWithLabel.tsx
subirjolly: per our discussion, changing the name of the variable

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/InputWithLabel.tsx
subirjolly: Changed to `newValue`. Feel free to suggest a better name.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+interface OwnProps {
+  showLogoWithText: Boolean
+}
+
+type Props = OwnProps & Cluster
+
+// TODO(Subir): Please suggest a better name
+export const ClusterBox: FC<Props> = ({
```
wdoconnell, line 39: Name of the component looks good to me, so okay to remove the comment.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.tsx
```diff
@@ -0,0 +1,86 @@
...
+// Components
+import {CreateOrgContext, ProviderID} from './CreateOrganizationContext'
+import {Cluster} from 'src/client/unityRoutes'
+import {AWSLogo} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/AWSLogo'
+import {AzureLogo} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/AzureLogo'
+import {GCPLogo} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/GCPLogo'
+import {AzureLogoWithText} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/AzureLogoWithText'
+import {GCPLogoWithText} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/GCPLogoWithText'
+import {AWSLogoColor} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderLogos/AWSLogoColor'
+
```
wdoconnell, line 18: On balance I would alphabetize the component imports so that they're easier to find, but I see the logic behind how you've organized them here.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
wdoconnell: I would still consider adding honeybadger reporting here - and would suggest we do it at the component level (rather than in the context).

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,166 @@
...
+      const providers = Object.keys(clusterMap)
+      providers.forEach(providerId => {
+        clusterMap[providerId] = clusterMap[providerId].sort(
+          (a, b) => a.priority - b.priority
+        )
+      })
+      setClusters(clusterMap)
+      if (!!providers.length) {
+        setCurrentProvider(providers[0] as ProviderID)
+        setCurrentRegion(clusterMap[providers[0]][0].regionId)
```
wdoconnell, line 136: Looks like we have a forced type in the `as` here.

wdoconnell: How many properties deep is this `clusterMap`?  Since `sort()` is in-place, I think we may need to spread or deep-clone to avoid inadvertently mutating the array before we hit `setClusters()`.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
wdoconnell: Did we resolve this one - is `setErrorMessage` what was added?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderChooser.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <FlexBox
+      alignItems={AlignItems.FlexStart}
+      justifyContent={JustifyContent.Center}
+      direction={FlexDirection.Column}
+      margin={ComponentSize.Large}
+      stretchToFitWidth={true}
+      className="provider-and-region-chooser"
+    >
```
wdoconnell, line 31: Super minor, but would move `className` up in the prop list.

wdoconnell (changes requested):
This is looking great, but I had a few minor final comments. Also, I noticed two nits in the videos:

(1) Should we keep the modal margins consistent with and without the error message? In the first video, when the error message is displayed, the margins in the modal expand noticeably to accommodate the error.

(2) I wondered whether we should tweak the CSS so that the provider logos are always centered. I think it's a design choice, but that looks nicer to me than having the 1-3 logos left-aligned.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
subirjolly: that's correct. I added
```
setShowError(true)
setErrorMessage(DEFAULT_ERROR_MESSAGE)
```

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
subirjolly: yes, `as` is a need here as I mentioned in the `AWS`, `Azure` comment. 

`setClusters` is a one time operation. This been said, it's an object with `ProviderID` as keys and array of cluster objects as its value. We don't need to do a spread here.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderChooser.tsx
subirjolly: absolutely

subirjolly:
> 

Absolutely valid points. I was thinking about the error message point but didn't change it to maintain consistency. but I think breaking the consistency here would be better.

As for the other point, @juliajames12, please suggest.

juliajames12:
@subirjolly Here are my thoughts based on Bill's feedback:
(1) Should we keep the modal margins consistent with and without the error message? In the first video, when the error message is displayed, the margins in the modal expand noticeably to accommodate the error. 
**I agree that the error message should not make the modal resize. The spacing should accommodate for the error message.**

(2) I wondered whether we should tweak the CSS so that the provider logos are always centered. I think it's a design choice, but that looks nicer to me than having the 1-3 logos left-aligned.
**Is it possible to have the 2 or 3 logo options (within the boxes) expand the full width of the modal (obviously not including the margins). For the 1 logo option, I still like it left aligned vs. having it floating in the center**

TCL735:
@subirjolly One thing I noticed for 2. Simple Input Validation and 3. Duplicate Organization Check is that, while the error message is great, the input box outline is still blue. Ideally, the input box outline should turn orange which indicates it is invalid.

subirjolly:
> 

Correct. There are two ways to display form element error in clockface.
1. `Form.ValidationElement`
2. `Form.Element` with `errorMessage` as a prop

The first one lets you provide a validationFunc but it's not very flexible as to when you should/can display the error message. The 2nd element is very flexible as to when you want to display. The first element provides the red border but the latter element doesn't. This been said, I have fixed my component to display the red border for the input box.

image

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.scss
```diff
@@ -0,0 +1,53 @@
...
+  cursor: pointer;
+  max-width: 350px;
+
+  &.selected {
+    border-color: $c-laser;
+    > .cluster-box-title {
+      background-color: $c-laser;
+    }
+  }
+
```
ChitlangeSahas, line 33: Do you think this `SelectableCard` component will take care of all this custom styling? 

https://influxdata.github.io/clockface/?path=/story/components-cards-selectablecard--selectablecard

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -0,0 +1,204 @@
...
+      changeOrgName(e.target.value)
+      setShowError(false)
+      setInputError(DEFAULT_ERROR_MESSAGE)
+    },
+    [changeOrgName]
+  )
+
+  useEffect(() => {
+    if (createOrgLoadingStatus === RemoteDataState.Error) {
+      if (error === 'NAME_CONFLICT') {
```
ChitlangeSahas, line 79: I wonder if the `NAME_CONFLICT` string should be a constant.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -0,0 +1,204 @@
...
+          setError('UNAUTHORIZED')
+        } else if (e instanceof ServerError) {
+          setError('SERVER_ERROR')
+        } else if (e instanceof UnprocessableEntityError) {
+          // Even though there can be different meanings/reasons for this issue.
+          //  We are deliberately choosing to display this error(Status Code: 422)
+          //  as Duplicate Org Name issue.
+          setError('NAME_CONFLICT')
+        } else {
+          setError('GENERIC_ERROR')
```
ChitlangeSahas, line 113: Same comment as whether this should be a constant.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -0,0 +1,204 @@
...
+            icon={IconFont.AlertTriangle}
+          >
+            {errorMessage}
+          </Alert>
+        )}
+        <FlexBox
+          direction={FlexDirection.Column}
+          className="create-org-overlay-body"
+        >
+          <InputWithLabel
```
ChitlangeSahas, line 164: What does InputWithLabel do? 

We have `Form` and it's `Form.ValidationElement` that handles the displaying of errors, perhaps that will do here?

subirjolly: I started off with the ValidationElement but it requires a `validationFunction` and does not provide the flexibility to display an error based on your requirement/need. If you know that component better, then I'd be happy to change it back.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ClusterBox.scss
subirjolly: didn't know about this component. will update it! thank you!

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/ProviderChooser.tsx
```diff
@@ -0,0 +1,73 @@
...
+      direction={FlexDirection.Column}
+      justifyContent={JustifyContent.Center}
+      margin={ComponentSize.Large}
+      stretchToFitWidth={true}
+    >
+      <InputLabel className="prc-title">Provider & Region</InputLabel>
+      <InputLabel className="prc-description" size={ComponentSize.Small}>
+        Tell us where you would like to store the time series data for this
+        organization.
+      </InputLabel>
```
hoorayimhelping, line 37: I'm not sure this is a proper use of `InputLabel`. According to the [storyboard](https://influxdata.github.io/clockface/?path=/story/components-inputs-base--inputlabel)

>InputLabel is designed to visually and functionally label an adjacent input element.

An HTML label is meant to label a form element (like an `input` or `checkbox`) and is usually used in the context of a form. I think semantically, it might be more appropriate to use a `Heading` element (so like `h3` or `h4`)

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/RegionDropdown.tsx
```diff
@@ -0,0 +1,67 @@
...
+    useContext(CreateOrgContext)
+  const handleRegionChange = useCallback(
+    region => {
+      changeCurrentRegion(region)
+    },
+    [changeCurrentRegion]
+  )
+  const regions = useMemo(() => {
+    return clusters?.[currentProvider]
+  }, [clusters, currentProvider])
```
hoorayimhelping, line 21: For most cases, this is a simple `O(1)` lookup on hashmap. Memoizing it will probably make it less performant, since the memoized lookup now has to do a shallow check of `clusters` to see if it's changed before using the memoized function. So it checks the entire object and compares keys instead of just accessing a single key. I think the `useMemo` should probably get axed here

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/RegionDropdown.tsx
```diff
@@ -0,0 +1,67 @@
...
+          {regions?.map((c: Cluster) => (
+            <Dropdown.Item
+              key={c.regionId}
+              id={c.regionId}
+              value={c.regionId}
+              onClick={handleRegionChange}
+              testID={`variable-type-dropdown-${c.regionId}`}
+              selected={currentRegion === c.regionId}
+            >
+              {c.regionName}
```
hoorayimhelping, line 60:
```suggestion
          {regions?.map((cluster: Cluster) => (
            <Dropdown.Item
              key={cluster.regionId}
              id={cluster.regionId}
              value={cluster.regionId}
              onClick={handleRegionChange}
              testID={`variable-type-dropdown-${cluster.regionId}`}
              selected={currentRegion === cluster.regionId}
            >
              {cluster.regionName}
```

I'd prefer this use variable names instead of letters, so if we ever need to search for cluster usage, we can search for `cluster` instead of `c\.`

hoorayimhelping (changes requested):
What do you think about using Clockface where possible? Is that something that should go in this PR or in a follow up one?

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/RegionDropdown.tsx
subirjolly: you are absolutely correct and will remove it.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/RegionDropdown.tsx
subirjolly: 👍

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -0,0 +1,204 @@
...
+const errorStyle = {
+  margin: '20px 0',
+}
+
+const errorMap = {
+  SERVER_ERROR: 'Organization creation failed.',
+  NAME_CONFLICT: NAME_CONFLICT_MESSAGE,
+  UNAUTHORIZED: 'Not authorized to create organization.',
+  CLUSTERS_FETCH_ERROR: 'Error fetching cluster information.',
+  GENERIC_ERROR: 'Organization creation failed.',
```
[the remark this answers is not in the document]
subirjolly: would this work? `Error creating organization. Please refresh and try again in some time.`

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,195 @@
...
+        currentProvider,
+        error,
+        orgName,
+        setError,
+      }}
+    >
+      {children}
+    </CreateOrgContext.Provider>
+  )
+}
```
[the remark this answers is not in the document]
subirjolly: Bill and I talked about this and decided to go with it. I am happy to talk about it. Please let me know.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -0,0 +1,209 @@
...
+        onClose()
+        dispatch(notify(orgCreateSuccess()))
+      })
+      .catch(e => {
+        if (e instanceof UnauthorizedError) {
+          setError(CreateOrgError.Unauthorized)
+        } else if (e instanceof ServerError) {
+          setError(CreateOrgError.ServerError)
+        } else if (e instanceof UnprocessableEntityError) {
+          // Even though there can be different meanings/reasons for this issue.
```
wdoconnell, line 112: I would modify this comment to be clear on what the 'issue' is and why we're choosing to interpret the error this way. Here's an example:

```
/*
Quartz responds with a 422 status code if the user-selected org name is a duplicate of an 
org name in the current account. It may also respond with a 422 for other reasons. 

However, the UI will interpret the 422 as an org name duplicate for purposes of user feedback, 
since a 422 generated for other reasons is very unlikely, and still means that the user cannot 
create an organization using the current input values anyway.' 
*/
```

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,195 @@
...
+export const CreateOrgContext = createContext<CreateOrgContextType>(
+  DEFAULT_CREATE_ORG_CONTEXT
+)
+
+interface ClusterMap {
+  [providerId: string]: Cluster[]
+}
+
+export const CreateOrgProvider: FC<Props> = ({children}) => {
+  const quartzOrganizations = useSelector(selectQuartzOrgs)
```
wdoconnell, line 74: Instead of using this selector, you can use the more specific selector `selectQuartzOrgsContents`, which doesn't include the `.status` property.

Separately - I think we might need a `useEffect` here that fetches the list of organizations IF it's not currently loaded. In the future, we may be dynamically dispatching the thunk that fetches the user's organizations only after they click the 'switch' button in global header, in which case it'll be possible to get to this modal before the full org list has been fetched for the first time.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/MenuItem.tsx
```diff
@@ -1,21 +1,28 @@
...
 
 export const CreateOrganizationMenuItem: FC = () => {
+  const dispatch = useDispatch()
   const handleCreateOrg = () => {
-    // TODO: Enable this when #5899 is worked on
-    // event('globalheader_createOrg_menuItem_clicked')
-    // TODO: Remove this comment when you are working on https://github.com/influxdata/ui/issues/5899
-    //       Add create organization functionality with modal, error handling and redirect to new organization
+    if (isFlagEnabled('createOrgTempFlag')) {
+      dispatch(
```
wdoconnell, line 20: What is this `createOrgTempFlag` doing?

wdoconnell (changes requested):
This is looking good! But I have a few more minor proposed changes now that I'm reviewing after the other edits made in the last couple days.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/RegionDropdown.tsx
```diff
@@ -0,0 +1,67 @@
+import React, {FC, useCallback, useContext, useMemo} from 'react'
+import {ComponentStatus, Dropdown} from '@influxdata/clockface'
+
+// Components
+import {Cluster} from 'src/client/unityRoutes'
+import {CreateOrgContext} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext'
+
+const SINGLE_PROVIDER_WIDTH = 375
+
+export const RegionDropdown: FC = () => {
```
ChitlangeSahas, line 10: I was curious about the need for this dropdown as it's own component?

subirjolly: i like to keep the parent container with less code and only the components or gluing code. this been said, i am happy to move it up but i feel that'd clutter the parent a little more as it's already big enough.

subirjolly: let me know your thoughts

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
```diff
@@ -0,0 +1,195 @@
...
+export enum CreateOrgError {
+  NameConflict = 'name-conflict',
+  Unauthorized = 'unauthorized',
+  ServerError = 'server-error',
+  ClustersFetchError = 'clusters-fetch-error',
+  GenericError = 'generic-error',
+}
+
+export interface CreateOrgContextType {
+  changeCreateOrgLoadingStatus: (_: RemoteDataState) => void
```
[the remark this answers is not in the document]
subirjolly: Should we do this instead? remove the default functions from DEFAULT_CONTEXT?

image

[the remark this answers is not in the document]
subirjolly: 2nd option 

image

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationContext.tsx
subirjolly: nice!

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/RegionDropdown.tsx
ChitlangeSahas: I don't know what convention we want to follow, I definitely hear your point that it will bloat up the parent file. I'll let @hoorayimhelping make a call. 

The only reason I bring this up was in the past we had a few split components that when we needed to add any new functionality, a lot of prop drilling was needed.

subirjolly: as discussed in code review, i am going to handle this in ffaf

wdoconnell:
Closing this (with gratitude for the original work done on it by Subir), as this was completed after further modification and refactoring by #6327.

Issue #6007, opened 2022-10-06
Prefer named exports over default exports


hoorayimhelping:
### Background

We're doing away with default exports in favor of named exports. For some background on the issue, see the following blog posts:

- https://www.codingzeal.com/posts/developer-dark-arts-default-exports
- https://betterprogramming.pub/understanding-the-difference-between-named-and-default-exports-in-react-2d253ca9fc22

### Process
**This should happen piecemeal, rather than all-at-once.** It will be too much work, too big a change, and too risky to change all our default exports at the same time. Instead, we should clean up exports as we see them.

The analogy I (bucky) used is: you can go to the beach to clean it and pick up trash, or you can go to the beach to hang out, play in the ocean, have a good time, and then on the way back to the car, grab a couple of pieces of trash because every little bit helps. What we should do is akin to picking up a couple pieces of trash on the way back to the car.

If you're modifying a file and you notice it has a default export, change it to a named export and change files that import that module as part of your larger change. If you're creating a new file, go with named exports from the start.

Obviously, we expect you to use your judgement here. If you're changing one word in a file that has a default export, but there are 30 files that import from it, consider making a follow-up PR to change that default export. If you're refactoring a component that has a default export and is imported in two places, it's probably a good idea to just make the change as part of that larger PR (maybe as a separate commit)

PR #6127, opened 2022-10-19
feat: add edit and delete functionality

### LINKED ISSUE #6120: New Data Explorer: Add a Delete Script button and functionality
Implement a button within the Open Script dialog that allows users to delete scripts from their list per the below design:
https://www.figma.com/file/0qAntPk5LVangAHguWT74X/Query-Experience-Project?node-id=2366%3A108394

This solution will be an interim solution so that Delete capability will be available until we can implement an overall redesign of the header and header buttons as proposed in this design:
https://www.figma.com/file/0qAntPk5LVangAHguWT74X/Query-Experience-Project

asalem1:
Closes #6120

This is a hack to meet a deadline and introduce a new feature for explicit renaming and deleting in the same overlay. Not proud of creating an even more overly complicated API but such is life. There's an issue to track this clean-up where we eventually want to clean this up and have a different flow, especially when we get to saving and accessing other resources. Until then, this is it:

delete-edit

### DIFF
--- a/src/dataExplorer/components/DeleteScript.tsx
+++ b/src/dataExplorer/components/DeleteScript.tsx
@@ -8,6 +8,9 @@ import {useSelector, useDispatch} from 'react-redux'
 import {getOrg} from 'src/organizations/selectors'
 import {notify} from 'src/shared/actions/notifications'
 import {deleteScriptFail} from 'src/shared/copy/notifications/categories/scripts'
+import {ResultsContext} from 'src/dataExplorer/components/ResultsContext'
+import {RemoteDataState} from 'src/types'
+import {QueryContext} from 'src/shared/contexts/query'
 
 let deleteScript
 
@@ -22,6 +25,8 @@ interface Props {
 
 const DeleteScript: FC<Props> = ({onBack, onClose}) => {
   const {resource} = useContext(PersistanceContext)
+  const {cancel} = useContext(QueryContext)
+  const {setStatus, setResult} = useContext(ResultsContext)
   const history = useHistory()
   const dispatch = useDispatch()
   const org = useSelector(getOrg)
@@ -35,6 +40,9 @@ const DeleteScript: FC<Props> = ({onBack, onClose}) => {
           throw new Error(resp.data.message)
         }
 
+        setStatus(RemoteDataState.NotStarted)
+        setResult(null)
+        cancel()
         history.replace(`/orgs/${org.id}/data-explorer/from/script/`)
         onClose()
       } catch (error) {
--- a/src/dataExplorer/components/SaveAsScript.tsx
+++ b/src/dataExplorer/components/SaveAsScript.tsx
@@ -37,7 +37,7 @@ interface Props {
 const SaveAsScript: FC<Props> = ({onClose, setOverlayType, type}) => {
   const dispatch = useDispatch()
   const history = useHistory()
-  const {hasChanged, resource, setResource, clearCompositionSelection, save} =
+  const {hasChanged, resource, setResource, save} =
     useContext(PersistanceContext)
   const {cancel} = useContext(QueryContext)
   const {setStatus, setResult} = useContext(ResultsContext)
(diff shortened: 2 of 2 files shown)


src/dataExplorer/components/DeleteScript.tsx
```diff
@@ -0,0 +1,72 @@
...
+if (CLOUD) {
+  deleteScript = require('src/client/scriptsRoutes').deleteScript
+}
+
+interface Props {
+  onBack: () => void
+  onClose: () => void
+}
+
+const OpenScript: FC<Props> = ({onBack, onClose}) => {
```
appletreeisyellow, line 23:
```suggestion
const DeleteScript: FC<Props> = ({onBack, onClose}) => {
```
Should it be `DeleteScript` according to the file name?

src/dataExplorer/components/DeleteScript.tsx
```diff
@@ -0,0 +1,72 @@
...
+          onClick={handleDeleteScript}
+          text="delete"
+          icon={IconFont.Trash_New}
+        />
+      </Overlay.Footer>
+    </Overlay.Container>
+  )
+}
+
+export default OpenScript
```
appletreeisyellow, line 72:
```suggestion
export {DeleteScript}
```
Same as above. And also it's a good time to use named exports, related to #6007

src/dataExplorer/components/DeleteScript.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const handleDeleteScript = async () => {
+    if (resource.data?.id) {
+      try {
+        const resp = await deleteScript({scriptID: resource.data.id})
+
+        if (resp.status !== 204) {
+          throw new Error(resp.data.message)
+        }
+
+        history.replace(`/orgs/${org.id}/data-explorer/from/script/`)
```
appletreeisyellow, line 38: Just curious, what's the purpose of this line? When I test on my local, I saw the URL changed very quickly from `.../data-explorer/` to `.../data-explorer/from/script/` then back to `.../data-explorer/` after a script is deleted

src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -161,6 +165,20 @@ const FluxQueryBuilder: FC = () => {
                     icon={IconFont.SaveOutline}
                   />
                 )}
+                {isFlagEnabled('saveAsScript') && (
+                  <Button
+                    className="flux-query-builder__action-button"
+                    onClick={() => setOverlayType(OverlayType.EDIT)}
+                    status={
+                      // TODO(ariel): check to see if it's saved
```
appletreeisyellow, line 173: Are the TODOs in this patch already done or for future?

appletreeisyellow (changes requested):
Thanks for putting this up so quickly! Two small change request and I have some questions below. Overall it looks good!

appletreeisyellow (dismissed):

src/dataExplorer/components/DeleteScript.tsx
wiedld: Is this loading another script resource? And that trickles into any of the state objects (session store, contexts, etc)?

wiedld (approved):
Thanks for rushing this out!

Have a non-blocking request, to have @juliajames12 tell us if we should be using the Delete button/icon style per the figma design (for the Edit Script overlay).

src/dataExplorer/components/DeleteScript.tsx
wiedld: Actually, we need to do more clearing of state. See the example `clear()` method in SaveAsScript component.

Otherwise, things stick around after delete. Such as the results:

https://user-images.githubusercontent.com/10232835/196870011-999135e1-f35a-4d61-bf81-0d1ca0bedf20.mov

wiedld (changes requested):
please clear additional state. See the `clear()` method in SaveAsScript for reference.

appletreeisyellow (approved):

wiedld (approved):
thanks!

src/dataExplorer/components/DeleteScript.tsx
juliajames12: Hey! Sorry for not seeing this earlier. I would like it implemented how I designed it because I think the filled-in button over-emphasizes it in the modal. Not a blocker, but if we could update it eventually to match, that'd be great! Thanks for catching that @wiedld. I can talk to the design team to get this incorporated into our design system so it's not just a one-off.

juliajames12:
Hey @asalem1. When @wiedld ask for my feedback as a "non-blocking request" and I responded with "Not a blocker, but if we could update it eventually to match, that'd be great!" that means it's not a blocker. I'm confused what you're waiting on from me to complete this work. Is it something else other than the delete button? If so, please let me know. Also you can always slack message me if you feel blocked. I'll be quicker to respond that way.

PR #6128, opened 2022-10-20
feat: add multi-query support for Adaptive Zoom

### LINKED ISSUE #5956: Adaptive Zoom - handle mutliple queries simultaneously
In the old Data Explorer's query builder, it is possible to build multiple queries. Hitting the submit button with multiple queries refreshes the data on all queries.

However, using adaptive zoom will only zoom in on the first query, with the zoomed-in view showing only 1 query and all other queries "lost" until the user zooms out back to the original view.

Adaptive Zoom should handle multiple queries.

### LINKED ISSUE #6134: Adaptive Zoom - usage page crashes when adaptive zoom is ON
On the `/usage` page, the graphs for the usage stats allows for zooming, just like any other line graph. However, when adaptive zoom is ON, a zoom-in will crash this page, due to the empty `queries` given to the usage graphs' `View` component.

Adaptive Zoom relies on having the Flux query (or queries) readily available, and a zoom-in action by the user triggers a re-query of the database using the same Flux queries with a smaller time range. Since usage stats are propagated through an api layer, the Flux queries (if any) are not directly available to the UI or the user. Therefore, Adaptive Zoom does not work here since there are no Flux queries to work with.

Adaptive Zoom would require either to have access to the usage stats' underlying Flux queries or additional capabilities to be built out using the usage api to do the zoom-in.

## For now, we are simply disabling adaptive zoom on the Usage page.

TCL735:
Closes #5956  
Closes #6134  

- Adds multi-query support for Adaptive Zoom in Data Explorer and Dashboard Cells.
- Disables Adaptive Zoom whenever queries cannot be found, such as on the `/usage` page

https://user-images.githubusercontent.com/10736577/196830215-10b9c534-c055-41be-9929-3e8f4a796047.mp4

### DIFF
--- a/src/visualization/utils/useVisDomainSettings.ts
+++ b/src/visualization/utils/useVisDomainSettings.ts
@@ -271,10 +271,10 @@ export const useZoomRequeryXDomainSettings = (args: ZoomRequeryArgs) => {
   }, [domain]) // eslint-disable-line react-hooks/exhaustive-deps
 
   // sync preZoomDomain and domain:
-  //   if it is the value axis (not time)
-  //   only when not zoomed in
-  //     - if they are different, or
-  //     - if it is the time axis and the time range has changed
+  //   if it is the value axis (not time), or
+  //   if it is the time axis only when not zoomed in, and
+  //      - they are different, or
+  //      - the time range has changed
   useEffect(() => {
     if (
       !timeRange ||
@@ -449,10 +449,10 @@ export const useZoomRequeryYDomainSettings = (args: ZoomRequeryArgs) => {
   }, [domain]) // eslint-disable-line react-hooks/exhaustive-deps
 
   // sync preZoomDomain and domain:
-  //   if it is the value axis (not time)
-  //   only when not zoomed in
-  //     - if they are different, or
-  //     - if it is the time axis and the time range has changed
+  //   if it is the value axis (not time), or
+  //   if it is the time axis only when not zoomed in, and
+  //      - they are different, or
+  //      - the time range has changed
   useEffect(() => {
     if (
       !timeRange ||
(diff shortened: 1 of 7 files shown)


src/visualization/utils/useVisDomainSettings.ts
```diff
@@ -211,7 +214,7 @@ export const useZoomRequeryXDomainSettings = (args: ZoomRequeryArgs) => {
 
   const [windowPeriod, setWindowPeriod] = useState<number>(
     normalizeWindowPeriodForZoomRequery(
-      getWindowPeriodFromVariables(query, variables),
+      getWindowPeriodFromVariables(queries[activeQueryIndex], variables),
```
blegesse-git, line 217: `queries[activeQueryIndex]` is returning undefined when selecting time range in `/usage` page, causing the graph to crash. I wonder why that is...🤔

TCL735: ah good catch. i'll check it out.

TCL735: This was an excellent find. So good, I have to mention it again! A bug was found that i didn't even realize existed.

I have pushed up a change to disable Adaptive Zoom whenever queries cannot be found, such as in the Usage page where the data comes from an api rather than a query.

Thanks again @blegesse-git !

blegesse-git: so glad I could be of help!!

blegesse-git (approved):

blegesse-git (approved):

TCL735:
I inadvertently removed the value axis (y-axis) syncing in this recent PR: https://github.com/influxdata/ui/pull/6048/files

Adding it back now. Here is an updated video showing it working

https://user-images.githubusercontent.com/10736577/197592016-a16511b7-0eb6-4415-bb8b-bbfbb2b9e0ed.mp4

blegesse-git (approved):

blegesse-git (approved):

PR #6137, opened 2022-10-21
chore(Identity): refactor fetchDefaultAccountDefaultOrg

### LINKED ISSUE #6136: Refactor: fetchDefaultAccountDefaultOrg logic to return an account if no default account exists

blegesse-git:
Closes #6136 

### Checklist

### DIFF
--- a/src/identity/apis/org.ts
+++ b/src/identity/apis/org.ts
@@ -113,7 +113,7 @@ export const fetchDefaultAccountDefaultOrg = async (): Promise<
 
     return defaultQuartzOrg
   }
-  throw new GenericError('no accounts found')
+  throw new GenericError('No accounts found')
 }
 
 // fetch data regarding whether the user can create new orgs, and associated upgrade options.
--- a/src/identity/apis/org.ts
+++ b/src/identity/apis/org.ts
@@ -104,17 +104,15 @@ export const fetchDefaultAccountDefaultOrg = async (): Promise<
   const {data} = response
 
   if (Array.isArray(data) && data.length) {
-    const defaultAccount = data.find(account => account.isDefault)
 
-    // fetch default org
-    if (defaultAccount) {
-      const quartzOrg = await fetchOrgsByAccountID(defaultAccount.id)
-      const defaultQuartzOrg =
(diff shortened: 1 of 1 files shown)


src/identity/apis/org.ts
```diff
@@ -101,21 +100,20 @@ export const fetchDefaultAccountDefaultOrg = async (): Promise<
   if (response.status === 500) {
     throw new ServerError(response.data.message)
   }
-  const {data} = response
+  const {data: accounts} = response
```
hoorayimhelping, line 103: :+1: `accounts` is a much better name than `data`.

```ts
const accounts = response.data
```
is an alternative to this destructuring syntax which has less syntax (fewer things like `{` and `:`) and might be easier to follow

hoorayimhelping (dismissed):
this looks good to me, but I'll let Bill weigh in on it since he knows more about what is needed for this

src/identity/apis/org.ts
```diff
@@ -101,21 +100,20 @@ export const fetchDefaultAccountDefaultOrg = async (): Promise<
...
 
-  if (Array.isArray(data) && data.length) {
-    const defaultAccount = data.find(account => account.isDefault)
+  if (Array.isArray(accounts) && accounts.length) {
+    // if no default Account, reference the 0-indexed account
+    const defaultAccount =
+      accounts.find(account => account.isDefault) || accounts[0]
```
wdoconnell, line 108: Behavior is probably the same either way in this case, but might be better to use nullish coalescing (`??`) over or (`||`) here, since the only behavior we're trying to guard against is `.find` returning `undefined`

wdoconnell (approved):

wdoconnell:
@blegesse-git Changes all look good to me - could you please test this with a short vid as well? I believe this function is only used for deep link routing.

src/identity/apis/org.ts
blegesse-git: agreed! I think my way might be unnecessarily fancy. I like your straight-forward approach

wdoconnell (approved):

PR #6145, opened 2022-10-21
fix(universalLogin): create a local copy of flag before signout clears it

ChitlangeSahas:
part of https://github.com/influxdata/idpe/issues/15477

When we tried testing the **logout** portion of the universal login feature, i.e. logout should use `quartz` endpoint and not idpe's, we noticed that the code under the flag `universalLogin` inside `Logout.tsx` was never hit.

After some digging this was the issue in summary:

User clicks logout -> flags cleared -> code looks for the flag value and sees it's false (because it was cleared), so it never runs.

the solution is to save a local copy of the flag on component mount, and then use that instead, so we don't have to worry about flags being cleared.

### Checklist

### DIFF
--- a/src/Logout.tsx
+++ b/src/Logout.tsx
@@ -21,12 +21,8 @@ const Logout: FC = () => {
   const history = useHistory()
   const dispatch = useDispatch()
 
-  let universalLoginFlagValue = false
   useEffect(() => {
-    universalLoginFlagValue = isFlagEnabled('universalLogin')
-  }, [])
-
-  useEffect(() => {
+    const universalLoginFlagValue = isFlagEnabled('universalLogin') ?? false
     const handleSignOut = async () => {
       if (CLOUD) {
         console.warn('universalLoginFlagValue: ', universalLoginFlagValue)
--- a/src/Logout.tsx
+++ b/src/Logout.tsx
@@ -21,10 +21,16 @@ const Logout: FC = () => {
   const history = useHistory()
   const dispatch = useDispatch()
 
+  let universalLoginFlagValue = false
+  useEffect(() => {
+    universalLoginFlagValue = isFlagEnabled('universalLogin')
+  }, [])
+
   useEffect(() => {
     const handleSignOut = async () => {
       if (CLOUD) {
-        if (isFlagEnabled('universalLogin')) {
+        console.warn('universalLoginFlagValue: ', universalLoginFlagValue)
+        if (universalLoginFlagValue) {
           fetch('/api/env/quartz-login-url')
             .then(async response => {
               const quartzUrl = await response.text()


src/Logout.tsx
```diff
@@ -21,10 +21,17 @@ const Logout: FC = () => {
   const history = useHistory()
   const dispatch = useDispatch()
 
+  let universalLoginFlagValue = false
+  useEffect(() => {
+    universalLoginFlagValue = isFlagEnabled('universalLogin')
+  }, [])
```
ChitlangeSahas, line 27: only run once on mount. mimic `componentDidMount`

src/Logout.tsx
```diff
@@ -21,10 +21,17 @@ const Logout: FC = () => {
...
+
+
   useEffect(() => {
     const handleSignOut = async () => {
       if (CLOUD) {
-        if (isFlagEnabled('universalLogin')) {
+        console.warn('universalLoginFlagValue: ', universalLoginFlagValue)
+        if (universalLoginFlagValue) {
```
ChitlangeSahas, line 34: use the saved flag value

src/Logout.tsx
```diff
@@ -21,10 +21,17 @@ const Logout: FC = () => {
...
+  useEffect(() => {
+    universalLoginFlagValue = isFlagEnabled('universalLogin')
+  }, [])
+
+
   useEffect(() => {
     const handleSignOut = async () => {
       if (CLOUD) {
-        if (isFlagEnabled('universalLogin')) {
+        console.warn('universalLoginFlagValue: ', universalLoginFlagValue)
```
ChitlangeSahas, line 33: just for sanity test. i will clean the warns once we decide to remove the flag.

ChitlangeSahas: This is hard to test locally, because of weird interactions with dex.

wdoconnell:
FYI, I'm reviewing this now.

hoorayimhelping:
>User clicks logout -> flags cleared -> code looks for the flag value and sees it's false (because it was cleared), so it never runs.

It looks to me like the flow is: user clicks logout -> `handleSignout` function is created -> `dispatch(reset())` is called -> `handleSignout` is called -> during this call, flag is checked but it has just been reset

Would it be simpler to  call `dispatch(reset())` after checking the state of the flag?

ChitlangeSahas:
> > User clicks logout -> flags cleared -> code looks for the flag value and sees it's false (because it was cleared), so it never runs.
> 
> It looks to me like the flow is: user clicks logout -> `handleSignout` function is created -> `dispatch(reset())` is called -> `handleSignout` is called -> during this call, flag is checked but it has just been reset
> 
> Would it be simpler to call `dispatch(reset())` after checking the state of the flag?

@hoorayimhelping : Me and bill were thinking the same, on friday, the only issue is the handleSignout will redirect to quartz codebase, and will no longer continue to execute the ui code. (logout happens upon window.replace to a quartz url)

this is why we thought it may not work.

hoorayimhelping:
>the only issue is the handleSignout will redirect to quartz codebase, and will no longer continue to execute the ui code. (logout happens upon window.replace to a quartz url)

If it's no longer executing UI code, why does it matter if `reset()` is called?

wdoconnell:
> > the only issue is the handleSignout will redirect to quartz codebase, and will no longer continue to execute the ui code. (logout happens upon window.replace to a quartz url)
> 
> If it's no longer executing UI code, why does it matter if `reset()` is called?

The `reset()` is needed for a fix that was applied here:
https://github.com/influxdata/ui/commit/ae9f199f6d9783bede66f573619f493b474d053f
https://github.com/influxdata/ui/pull/176

I don't believe the OSS login flow has changed since then in a way that would permit removing the reset.

wdoconnell:
FYI @hoorayimhelping - Sahas is going to move the conditional into the same useEffect hook.

hoorayimhelping:
@wdoconnell 

>The `reset()` is needed for a fix that was applied here:
>https://github.com/influxdata/ui/commit/ae9f199f6d9783bede66f573619f493b474d053f
>https://github.com/influxdata/ui/pull/176

Are you sure? That change added

```ts
dispatch({type: 'USER_LOGGED_OUT'})
```

```ts
dispatch(reset())
```

was already there. `reset()` clears the feature flag state and is what is applicable to this PR.

wdoconnell:
I'm not familiar with how to access PRs from this time period - looks like another repo? https://github.com/influxdata/ui/commit/43ba458e8cf533725108470ce768b6c300803679

hoorayimhelping:
https://github.com/influxdata/influxdb/pull/17926

wdoconnell:
(For context - that's the commit that adds the feature flag `reset`. But not linked to a PR we can use to ascertain the reason for the change.)

wdoconnell:
Let me ask Ari, since I think he reviewed that PR. It's a little hard to ascertain what the background behind adding the `reset()` is. This is a while back but maybe he'll remember. 😄

wdoconnell (approved):

PR #6157, opened 2022-10-25
feat(6106): pgsql support in monaco-editor

### LINKED ISSUE #6106: SQL Editing: Monaco Editor Integration
The Monaco editor component needs to be able to understand the model to be sql. It shouldn't try and send the file to the flux lsp. Right now, the component hard codes flux, and doesn't have availability for other document types. Models should keep track of their own file type, using [the optional language parameter in createModel](https://microsoft.github.io/monaco-editor/api/modules/monaco.editor.html#createModel).

wiedld:
Closes #6106 

## Done:
* Provide the monaco-editor with pgsql specific:
  * syntax tokenizer
  * theme/coloring of that syntax
  * submit hotkey
* Decided to provide separate root editor because:
  * we already provide editor components per language (e.g. toml, markdown, python, flux, etc)
  * will need a separate LSP connection anyways.  

## Theme:

@juliajames12 -- I am guessing these styles will need updated colors. Took a guess based upon colors used in the flux editor.

Screen Shot 2022-10-24 at 7 36 28 PM

## Checklist
- [x] Feature flagged, if applicable. `uiSqlSupport`

### DIFF
--- a/src/languageSupport/languages/sql/monaco.sql.syntax.ts
+++ b/src/languageSupport/languages/sql/monaco.sql.syntax.ts
@@ -1,0 +1,23 @@
+import * as pgsql from 'monaco-editor/esm/vs/basic-languages/pgsql/pgsql'
+
+import {MonacoBasicLanguage} from 'src/types'
+
+const LANGID = 'pgsql'
+
+function addSyntax() {
+  monaco.languages.register({
+    id: LANGID,
+  })
+  monaco.languages.setMonarchTokensProvider(
+    LANGID,
+    (pgsql as MonacoBasicLanguage).language
+  )
+  monaco.languages.setLanguageConfiguration(
+    LANGID,
+    (pgsql as MonacoBasicLanguage).conf
+  )
+}
+
+addSyntax()
+
+export default LANGID
--- a/src/languageSupport/languages/sql/monaco.sql.syntax.ts
+++ b/src/languageSupport/languages/sql/monaco.sql.syntax.ts
@@ -20,4 +20,4 @@ function addSyntax() {
 
 addSyntax()
 
-export default LANGID
+export {LANGID}
--- a/src/languageSupport/languages/sql/monaco.sql.theme.ts
+++ b/src/languageSupport/languages/sql/monaco.sql.theme.ts
@@ -1,0 +1,74 @@
+import {MonacoType} from 'src/types'
+import {InfluxColors} from '@influxdata/clockface'
+
+const THEME_NAME = 'sqlTheme'
+
+function addTheme(monaco: MonacoType) {
+  // Using monaco-editor pgsql tokenizer.
+  // Example tokens: https://github.com/microsoft/monaco-editor/blob/136ce723f73b8bd284565c0b7d6d851b52161015/src/basic-languages/pgsql/pgsql.test.ts
+  monaco.editor.defineTheme(THEME_NAME, {
+    base: 'vs-dark',
+    inherit: false,
+    rules: [
+      {
+        token: 'delimiter.square.sql',
+        foreground: '#6BDFFF',
+      },
+      {
+        token: 'delimiter.parenthesis.sql',
+        foreground: '#6BDFFF',
+      },
+      {
+        token: 'delimiter.sql',
+        foreground: '#6BDFFF',
+      },
+      {
+        token: 'identifier.sql',
+        foreground: '#7CE490',
+      },
+      {
+        token: 'comment.quote.sql',
+        foreground: '#676978',
+      },
+      {
+        token: 'comment.sql',
+        foreground: '#676978',
+      },
+      {
+        token: 'keyword.sql',
+        foreground: '#9394FF',
+      },
+      {
+        token: 'operator.sql',
+        foreground: '#FFB6A0',
+      },
+      {
+        token: 'number.sql',
+        foreground: '#6BDFFF',
+      },
+      {
+        token: 'string.sql',
+        foreground: '#7CE490',
+      },
+      {
+        token: 'identifier.quote.sql',
+        foreground: '#7CE490',
+      },
+    ],
+    colors: {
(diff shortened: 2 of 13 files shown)


src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -144,7 +141,7 @@ const FluxEditorMonaco: FC<Props> = ({
       <ErrorBoundary>
         <div className={wrapperClassName} data-testid="flux-editor">
           <MonacoEditor
-            language={language}
+            language={FLUXLANGID}
```
appletreeisyellow, line 144: Why not using the enum `LanguageType.FLUX`? It's the same string `flux` as `FLUXLANGID`. Since now we have two supported languages and it will grow more in the future, I think it's better to use the same source of truth

src/languageSupport/languages/sql/monaco.sql.syntax.ts
```diff
@@ -0,0 +1,23 @@
+import * as pgsql from 'monaco-editor/esm/vs/basic-languages/pgsql/pgsql'
+
+import {MonacoBasicLanguage} from 'src/types'
+
+const LANGID = 'pgsql'
```
appletreeisyellow, line 5: Is it required in Monaco editor to have language id to be `pgsql`? I tried to look up Monaco editor docs but couldn't find any. The reason I asked is because there is an enum `LanguageType.SQL` that is set to be `sql`. I'd like to see if it is possible to combine these two constants in one way or another

appletreeisyellow (changes requested):
Overall great work! I commented a couple places about code pattern related to named exports and asked some questions related to constants

src/languageSupport/languages/sql/monaco.sql.syntax.ts
```diff
@@ -0,0 +1,23 @@
...
+    LANGID,
+    (pgsql as MonacoBasicLanguage).language
+  )
+  monaco.languages.setLanguageConfiguration(
+    LANGID,
+    (pgsql as MonacoBasicLanguage).conf
+  )
+}
+
+addSyntax()
```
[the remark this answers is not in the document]
wiedld: I was following the existing pattern, rather than updating. In the existing pattern we loading syntax for a given language whenever the langid for that language is used. This means that when we import the sql `LANGID` into the SqlMonacoEDitor,  the root called `addSyntax()` will load the language features into the monaco-editor.

(Note: this is the existing pattern for all of our languages. I have not re-done this pattern yet.)

This also answers the next question, of why the `FLUXLANGID` is used for the FluxMonacoEditor component. It's so that flux the syntax is loaded into monaco-editor.

src/shared/components/FluxMonacoEditor.tsx
wiedld: To trigger the syntax loading into the monaco-editor.

src/languageSupport/languages/sql/monaco.sql.syntax.ts
wiedld: monaco-editor provides a series of language syntax out of the box. We use the `MonacoWebpackPlugin` to determine which languages to load -- specifically, those languages where monaco-editor also provides a baked-in LSP.

```
new MonacoWebpackPlugin({
      languages: ['json', 'markdown'],
      filename: `${STATIC_DIRECTORY}[name].worker.[contenthash].js`,
      globalAPI: true,
    })
```

For other languages, specifically those found under the[ basic-languages directory here](https://github.com/microsoft/monaco-editor/tree/main/src/basic-languages), there is no baked-in LSP available. But there are syntaxes available, which we load and use with our own custom color theme.

In that^^ list of basic syntaxes, monaco-editor uses the language id `pgsql`. Since our backend dialect is postgres sql, I choose that sql dialect for the syntax.

src/shared/components/SqlMonacoEditor.tsx
```diff
@@ -0,0 +1,114 @@
...
+          />
+          <div className="monaco-editor__language">{SQLLANGID}</div>
+        </div>
+      </ErrorBoundary>
+    ),
+    [onChangeScript, script]
+  )
+}
+
+export default SqlEditorMonaco
```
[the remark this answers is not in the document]
wiedld: done

src/languageSupport/languages/sql/monaco.sql.theme.ts
```diff
@@ -0,0 +1,74 @@
...
+      'editorCursor.foreground': '#ffffff',
+      'editorActiveLineNumber.foreground': '#bec2cc',
+      'minimap.background': InfluxColors.Grey15,
+    },
+  })
+}
+
+addTheme(monaco)
+
+export default THEME_NAME
```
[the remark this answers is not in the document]
wiedld: done

src/languageSupport/languages/sql/monaco.sql.syntax.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default LANGID
```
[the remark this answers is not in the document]
wiedld: done

src/shared/components/FluxMonacoEditor.tsx
appletreeisyellow: Got it. I was confused with resource language type in [resources/index.ts](https://github.com/influxdata/ui/blob/master/src/dataExplorer/components/resources/index.ts#L6) with `FLUXLANGID` 😅

wiedld: Totally agreed. It's definitely not ideal. Which is why we should eventually change how/why we load the language syntax (for all languages), to make it more explicit.

src/shared/components/SqlMonacoEditor.tsx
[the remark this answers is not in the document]
wiedld: done

appletreeisyellow (approved):

PR #6161, opened 2022-10-25
chore(maps): make maps use uiproxyd routes instead of mapsd routes

### LINKED ISSUE #6160: Remove mapsd service
See https://github.com/influxdata/ui/issues/4304

Currently uiproxyd has a route to get a map token, but it's not in use because the code was never cut over to use the new route. Change the code to use `uiproxyd`s route for getting the map token rather than `mapsd`

hoorayimhelping:
Closes #6160

Connects maps visualizations to `uiproxyd`s version of the `getMapToken` route, rather than `mapsd`s version.

Works locally:
Screen Shot 2022-10-25 at 11 08 23 AM

You can test the route in cloud environments by going to your cloud instance and logging in, then appending `/api/v2private/uiproxy/mapToken` to the end of the url. You should get a maps token back.

### Checklist

### DIFF
--- a/src/visualization/types/Map/view.tsx
+++ b/src/visualization/types/Map/view.tsx
@@ -11,15 +11,11 @@ import {VisualizationProps} from 'src/visualization'
 import {getGeoCoordinates} from 'src/shared/utils/vis'
 import {event} from 'src/cloud/utils/reporting'
 import {CLOUD} from 'src/shared/constants'
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 let getMapToken = null
 
 if (CLOUD) {
-  getMapToken = require('src/client/mapsdRoutes').getMapToken
-  if (isFlagEnabled('uiproxyd')) {
-    getMapToken = require('src/client/uiproxydRoutes').getMapToken
-  }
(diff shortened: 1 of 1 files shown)


wdoconnell:
Is the `uiproxyd` flag also going away - or only going away for this visualization?

ChitlangeSahas (approved):
Just curious if this will be considered a dynamic import?

hoorayimhelping:
>Is the `uiproxyd` flag also going away - or only going away for this visualization?

@wdoconnell the flag is already gone. this instance of it was missed

>Just curious if this will be considered a dynamic import?

@ChitlangeSahas I don't think this is technically a dynamic import, because it's using `require` instead of `import()`. However, if I changed it to something like:

```ts
import('src/client/uiproxydRoutes').then(module => {
  getMapToken = module.getUiproxyMapToken
})
```
it would be dynamic import. I think I'm actually going to do this instead, since to make the codebase ESM compliant, we need to only import via `import` and not require.

hoorayimhelping:
>I think I'm actually going to do this instead, since to make the codebase ESM compliant, we need to only import via import and not require.

update: I tried this, and lint failed. We can't have nice things.

ChitlangeSahas (approved):

PR #6162, opened 2022-10-25
feat: create a unique search parameter to differentiate between the two data explorers

### LINKED ISSUE #6022: Create unique URL for new script editor to differentiate from the old data explorer
Daniel and I would like to gather heatmap data of the new script editor using VWO, but in order to do so we need the new script editor to have its own unique URL. 

I don't have a strong preference but we could call it /fluxscripteditor

asalem1:
Closes #6022

- Adds a new unique search parameter so that we can differentiate user flows between the data explorer and the new script editor
- Adds guardrails for the ui SQL selection process for users that are part of an IOx org

### DIFF
--- a/src/dataExplorer/components/FluxQueryBuilder.tsx
+++ b/src/dataExplorer/components/FluxQueryBuilder.tsx
@@ -255,14 +255,12 @@ const FluxQueryBuilder: FC = () => {
             >
               <ResultsPane />
             </DraggableResizer.Panel>
-            <>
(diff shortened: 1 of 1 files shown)


wiedld (approved):
LGTM.

Thanks for the change for the resizer panel. It may be worthwhile to add a clockface feature request to be able to denote initial collapsed state on the Draggable Resizer. That way we can set the right hand panel to be collapsed.

Issue #5819, opened 2022-09-19
Remove dynamic import of libraries


kristinarobinson:
There are a few places where import are generated dynamically based on folder location. To improve code readability and code tracing, convert these to standard library imports.

wiedld:
Related issue ticket:
https://github.com/influxdata/ui/issues/5687

TCL735:
As noted in the other issue, the self-registering pattern is also happening in new Data Explorer. See https://github.com/influxdata/ui/blob/master/src/dataExplorer/components/resources/index.ts#L31-L45

PR #6164, opened 2022-10-26
refactor: remove dynamic import of visualizations

TCL735:
Part of #5819 

As part of the undertaking to remove all dynamic imports, we start with the visualizations.

- removes the `require.context` loading of visualizations
- removes `export default` from visualizations and associated files
- updates naming to be more consistent with historical naming (even where historical naming may be inconsistent, let's at least be consistent with the inconsistency 🤣 )


hoorayimhelping (dismissed):
I love this so much! Great work!

I included Paul and Denise since they're also not a fan of this pattern and I wanted them to be aware of it

src/visualization/types/Band/index.tsx
```diff
@@ -1,15 +1,13 @@
...
-export default register => {
-  register({
-    type: 'band',
-    name: 'Band',
-    graphic: icon,
-    initial: properties,
-    component: view,
-    options,
-  })
+export const view = {
```
hoorayimhelping, line 6: what happens if we did something like:

```ts
export const view implements Visualization = {
```

does that work? I think this might be a missing piece of this puzzle

hoorayimhelping: >This works but the `view` becomes a JavaScript `class` (not to be confused with a React class component). I'll push up the changes.

@TCL735 what do you think of this change? Is it worth it? Too much overhead to just `implement` an interface?

TCL735: It's fine to have JavaScript classes. The ones we are implementing are small and very straightforward. The benefit we get is an easy-to-follow type for future reference in making new visualizations.

hoorayimhelping (approved):

rockstar (commented):
This looks great. It increases readability and that diffstat is just 😍.

Now I think it's important to have to invite [Chesterton to have a seat and talk about this fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence), because I think we should at least be clear here...

We're removing this pattern, but are we removing it because we have an actual _need_ to remove it? Is it getting in the way of delivering code _right now_? Or could we say this is shaving a yak? This change touches a lot of files, and could potentially cause a lot of spurious conflicts that might get in the way of other, more valuable changes. Undoing complexity that isn't really in the way right now may be more disruptive than just leaving it alone for now.

I realize that there's a feeling that there wasn't enough friction when these changes were first introduced, and I'm not meaning to introduce more friction than should be necessary now. I do think we should be very clear about why this fence is in the field, what it provides, and why we think we'll be better off without the fence.

TCL735:
@rockstar i needed to do it to stay sane. Does that count? 😀

hoorayimhelping:
@rockstar @TCL735 

>Now I think it's important to have to invite [Chesterton to have a seat and talk about this fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence), because I think we should at least be clear here...

Fair. But in this case, we were both around when this fence was erected. The problems this pattern was intending to solve are: making use of visualization code more consistent and easier to work with, making adding new visualizations more consistent, and demonstrating how clever the author is at implementing novel patterns.

To be fair, the original code _did_ solve those problems, but it did it in a profoundly confusing way, and right at a time when we stopped adding new visualizations. This new PR solves the problems of making visualization code more consistent to create and use, but it does it in a much more conventional and understandable way (it doesn't rely on implicitly defined directory structure and filenames, but instead uses the explicit mechanism of a TS interface).

>We're removing this pattern, but are we removing it because we have an actual need to remove it? Is it getting in the way of delivering code right now?

I definitely understand and agree with the point you're making about making, but I want to make sure we're not utterly utilitarian when determining what to work on. Sometimes bringing some oroder to the chaos is reason enough to do work, even if there might be more business value elsewhere. Sometimes the act of taking control of something that has been out of control is empowering and that is reason enough. This was an effort of a couple of hours including handling review comments, and this can provide a template to base future work of removing this pattern off of.

PR #6167, opened 2022-10-26
fix(clientLibrary): shorten js client library name

### LINKED ISSUE #6119: Homepage - JavaScript/Node.js icon and text look out of alignment compared to the other icons...
## Issue
- The JavaScript/Node.js icon and text look out of alignment compared to the other icons...

## Screenshot
Image

blegesse-git:
Closes #6119 

Updates client library name from `JavaScript/Node.js` to just `Node.js` in Homepage and Load Data page. 

Homepage 
Screen Shot 2022-10-26 at 2 00 21 PM

Load Data Page
Screen Shot 2022-10-26 at 2 01 31 PM

### Checklist


hoorayimhelping:
Do we need to update the url and deeplinks? This might be a question that Amy needs to answer.

The url for this feature is still `new-user-setup/nodejs`: https://github.com/influxdata/ui/blob/master/src/shared/containers/SetOrg.tsx#L302

And the deeplink is `/me/setup-nodejs`: https://github.com/influxdata/ui/blob/master/src/utils/deepLinks.ts#L28

Maybe we can keep the existing deeplink for backwards compatibility and add a new one?

blegesse-git:
> Do we need to update the url and deeplinks? This might be a question that Amy needs to answer.
> 
> The url for this feature is still `new-user-setup/nodejs`: https://github.com/influxdata/ui/blob/master/src/shared/containers/SetOrg.tsx#L302
> 
> And the deeplink is `/me/setup-nodejs`: https://github.com/influxdata/ui/blob/master/src/utils/deepLinks.ts#L28
> 
> Maybe we can keep the existing deeplink for backwards compatibility and add a new one?

Or - instead of updating the name to `JavaScript` we can do `Node.js`. That way, the naming will remain consistent with component name (NodejsWizard), deeplink route, icon, etc.

ALuckette:
I'm leaning towards Beth's idea of naming it "node.js" ... since that's what it is. I'm consulting with Marketing to see if this will be a problem. 

In terms of deep linking...will naming it "node.js" cause any issues?

hoorayimhelping:
>In terms of deep linking...will naming it "node.js" cause any issues?

@ALuckette I can't think of any, it's already called nodejs, so it will probably cause less confusion

ALuckette:
Ok, that's good to know. I'll let you now what I hear from marketing.

ALuckette:
Hello,

After checking with @mhall119, there is no harm in naming the client library "node.js" 

@hoorayimhelping and @blegesse-git move this forward by naming it "node.js" 

Thanks!

hoorayimhelping (dismissed):

Issue #4815, opened 2022-06-14
Resolve linter warnings re: react-hooks/exhaustive-deps


wdoconnell:
## About the bug

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. Run eslint on current UI build

**Expected behavior:**
No warnings from linter.

**Actual behavior:**
Over 100 warnings of the form shown below.

**Visual Proof:**
`157:6  warning  React Hook useMemo has missing dependencies: 'data' and 'update'. Either include them or remove the dependency array                react-hooks/exhaustive-deps`

## About your environment

**Environment info:**
N/A (any build, or CI-CD)

**Config:**
N/A

**Logs:**
N/A

**Performance:**
Generate profiles with the following commands for bugs related to performance, locking, out of memory (OOM), etc.

```sh
# Commands should be run when the bug is actively.
# Note: This command will run for at least 30 seconds.
curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all?cpu=true"
curl -o vars.txt "http://localhost:8086/debug/vars"
iostat -xd 1 30 > iostat.txt
# Attach the `profiles.tar.gz`, `vars.txt`, and `iostat.txt` output files.
```

kristinarobinson:
Break this down into smaller chunks by feature.

wdoconnell:
Updated the description here to avoid suggesting that `// eslint-disable-line exhaustive-deps` could be a solution, which it really shouldn't be unless it's absolutely required, or the code is truly too complex or risky to refactor.

Here's [a helpful discussion](https://github.com/facebook/react/issues/14920#issuecomment-471070149) of why it's usually a bad idea, and recommendations about how to fix the useEffect hooks without disabling the linter warning.

wdoconnell:
Closing based on priority.

PR #6170, opened 2022-10-26
chore(Authorizations): Resolve linter warnings in tokens

### LINKED ISSUE #6131: Resolve linter warnings: Tokens
Part of #4815

blegesse-git:
Closes #6131 

breaking down #4815 by feature - pr cleans up linter warnings inside `Authorizations` component. 

https://user-images.githubusercontent.com/66275100/199047667-e92af9b7-4ec6-496d-a4e0-ceb5adc1ab0b.mov

### Checklist


src/authorizations/components/CustomApiTokenOverlay.tsx
```diff
@@ -77,25 +77,34 @@ const CustomApiTokenOverlay: FC<Props> = props => {
...
 
+  const {
+    allResources,
+    bucketPermissions,
+    getBuckets,
+    getTelegrafs,
+    remoteDataState,
+    telegrafPermissions,
+  } = props
+
```
blegesse-git, line 88: clean up by destructuring props.

src/authorizations/components/CustomApiTokenOverlay.tsx
```diff
@@ -77,25 +77,34 @@ const CustomApiTokenOverlay: FC<Props> = props => {
...
   useEffect(() => {
-    props.getBuckets()
-    props.getTelegrafs()
-  }, [])
+    getBuckets()
+    getTelegrafs()
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
```
wdoconnell, line 93: I don't think we need to disable the warning here - is the dependency it's tracking `getBuckets` and `getTelegrafs`? Those shouldn't change, so I think they are probably safe to include in the array.

src/authorizations/components/CustomApiTokenOverlay.tsx
```diff
@@ -106,10 +115,10 @@ const CustomApiTokenOverlay: FC<Props> = props => {
     // BUT, code inside the hook won't run until remoteDataState is 'Done'.
     // Only then will props.bucketPermissions.sublevelPermissions will have value.
     // Consequently, we update the permissions state.
-  }, [props.remoteDataState])
+  }, [remoteDataState]) // eslint-disable-line react-hooks/exhaustive-deps
```
wdoconnell, line 119: Same here - what dependencies does the useEffect expect other than `remoteDataState`?  If we can safely include those dependencies, we should likely do so here.

src/authorizations/components/EditTokenOverlay.tsx
```diff
@@ -61,7 +61,7 @@ const EditTokenOverlay: FC<Props> = props => {
     if (_.isEmpty(permissions)) {
       formatPermissions()
     }
-  }, [])
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
```
wdoconnell, line 65: Same here - if the dependency we're waiting on is `formatPermissions()`, it's likely safe to include in the array since it shouldn't change. It is definitely worth re-testing the page after adding the dependencies though, just to make sure it doesn't produce any unexpected behavior.

wdoconnell (changes requested):
I had a couple of thoughts here on dependencies being able to be included in the `useEffect` hooks 👍

src/authorizations/components/CustomApiTokenOverlay.tsx
blegesse-git: aside from `remoteDataState`, it expects `allResources, bucketPermissions, and telegrafPermissions`. Adding the rest will cause the hook to run more times than necessary. Only dependency we're concerned with here is `remoteDataState`. Only when its state changes, we want the code inside the hook to run.

src/authorizations/components/CustomApiTokenOverlay.tsx
blegesse-git: You are correct, those two are actions and shouldn't change. Reason I left them out was to show the effect doesn't need to re-run as it doesn't depend on any values. But since adding them to the array doesn't change the logic, I'll remove the eslint comment.

src/authorizations/components/CustomApiTokenOverlay.tsx
wdoconnell: Elsewhere, a possible solution is to include an `if` check that will only run the logic when the remote data state is `NotStarted` (or whatever it should be when the logic is run). That way, the logic of the hook runs only when we want it to, but we can also include everything in the dependency array.

blegesse-git: so what you're suggesting (`if` check containing code that will only run depending on state of remoteDataState) is exactly what the code is currently doing. `if (!isEmpty(bucketPermissions.sublevelPermissions))` will only return true when `remoteDataState` is `Done`. In other words, code inside the if statement won't execute until all necessary data is fetched. But still, including the rest of the values in the dependency array means useEffect will run one more time than necessary. Check out the video: first one is with `remoteDataState` as the only thing in the dependency array. Second video shows console when useEffect has all values in it's array : 

https://user-images.githubusercontent.com/66275100/198349558-38e3a2bd-bf5d-4bc3-b65f-c1f2076ad84c.mov

https://user-images.githubusercontent.com/66275100/198349675-363d090d-1808-4fbb-9a13-19ef5b317636.mov

src/authorizations/components/EditTokenOverlay.tsx
blegesse-git: here, `permissions` and `formatPermissions` is what eslint is expecting in the dependency array. To add formatPermissions to the array, I'll have to restructure the code and move the useEffect hook after `formatPermissions` function is declared. Doing that will make the dependencies of useEffect Hook change on every render. To fix this issue, we can move the function inside the useEffect callback. or wrap the definition of 'formatPermissions' in its own useCallback() Hook. I tried the second option wrapping it in its own useCallback hook and it worked fine, but... is all of this necessary? @hoorayimhelping do you have opinions on this?

hoorayimhelping: Since `formatPermissions` uses props, I think we'll want it to be part of the dependency array, since when the props change, the outcome of the function is liable to change as well. Seems like doing the `useCallback` treatment is the best idea here

blegesse-git: okay, after reading [this](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies), I see why it's necessary @wdoconnell. and thank you Bucky for chiming in.

src/authorizations/components/CustomApiTokenOverlay.tsx
wdoconnell: In other words - could you try seeing what happens if you change the `if` check to directly check whether the remotedatastate is `Done` before running the logic (instead of looking for sublevelPermissions?)? If so, I think the logic will only run once, because the interim changes in permissions all happen before the data state is finally set to `Done`

wdoconnell:
Thanks for handling this @blegesse-git, especially since this component is a bit of a monster.

I'm going to blame myself here for sewing confusion in the original ticket https://github.com/influxdata/ui/issues/4815 by suggesting that some of the warnings should be disabled. 🙃 

I've removed that aside from the ticket. You found an article on this too, but here's an explanation from the React discussion threads where folks go through lots of examples of ways to fix problematic dependency arrays. https://github.com/facebook/react/issues/14920#issuecomment-471070149

src/authorizations/components/CustomApiTokenOverlay.tsx
blegesse-git: yes, we could also change the if statement to directly check `props.remoteDataState === RemoteDataState.Done` but we our dependency array dilemma will still exist. Actually, doing so will cause the useEffect hook to run even more times. check it out: 

https://user-images.githubusercontent.com/66275100/198364241-bb64a9c1-0f84-4cfa-b176-f68787f61a1b.mov

Here is the code:

src/authorizations/components/CustomApiTokenOverlay.tsx
```diff
@@ -77,39 +77,54 @@ const CustomApiTokenOverlay: FC<Props> = props => {
...
+    telegrafPermissions,
+  } = props
+
   useEffect(() => {
-    props.getBuckets()
-    props.getTelegrafs()
-  }, [])
+    getBuckets()
+    getTelegrafs()
+  }, [getBuckets, getTelegrafs])
```
wdoconnell, line 93: Since these methods are being added to the dependency array - should we memoize the function with `useCallback`?

src/authorizations/components/EditTokenOverlay.tsx
```diff
@@ -99,7 +93,13 @@ const EditTokenOverlay: FC<Props> = props => {
     }
 
     setPermissions(newPerms)
-  }
+  }, [props])
+
```
wdoconnell, line 97: I think adding `props` as a dependency in a useEffect will create a new warning (and creates loop risk) - so I would consider destructuring this first, before adding it to the array. [Adding props as dependency in a useEffect](https://stackoverflow.com/questions/72757679/react-adding-props-something-as-a-dependency-of-useeffect)

src/authorizations/components/EditTokenOverlay.tsx
```diff
@@ -99,7 +93,13 @@ const EditTokenOverlay: FC<Props> = props => {
...
+  useEffect(() => {
+    if (_.isEmpty(permissions)) {
+      formatPermissions()
+    }
```
wdoconnell, line 101: Can we import `isEmpty` directly here? I think the existing page is importing the entire lodash library (_), which we don't need.

wdoconnell (changes requested):
Looking good! But I had a couple final comments/proposed changes.

src/authorizations/components/CustomApiTokenOverlay.tsx
wdoconnell: I'm wrong here - there's nothing to memoize because this is being imported directly from another file (not passed from another component). The reason it's destructured from props is due to mdtp.

wdoconnell (approved):
Looks good - thanks for accommodating my proposals. 😸

PR #6180, opened 2022-10-27
fix: pin to node image with node16

genehynson:
`node:lts` is now using node18 which has a requirement for OpenSSL v3 which requires Webpack 5. 

https://stackoverflow.com/questions/69665222/node-js-17-0-1-gatsby-error-digital-envelope-routinesunsupported-err-os

Until we can upgrade to Webpack 5, we will need to stay on a node16 image.

### Checklist

### DIFF
--- a/docker/Dockerfile.chronograf.prod
+++ b/docker/Dockerfile.chronograf.prod
@@ -1,5 +1,5 @@
 # from node lts alpine
-FROM node:lts-alpine AS repo
+FROM node:lts-alpine3.14 AS repo
 
 # env vars to configure the system
 
--- a/docker/Dockerfile.chronograf
+++ b/docker/Dockerfile.chronograf
@@ -1,4 +1,4 @@
-FROM node:lts AS base
+FROM node:lts-gallium AS base
 
 RUN curl --compressed -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.22.4
 


TCL735 (approved):

docker/Dockerfile.chronograf.prod
```diff
@@ -1,5 +1,5 @@
 # from node lts alpine
-FROM node:lts-alpine AS repo
+FROM node:lts-alpine3.14 AS repo
```
hoorayimhelping, line 2: I may be reading [this release](https://alpinelinux.org/posts/Alpine-3.14.0-released.html) wrong, but it looks like this will put us on Node 14? 

Do we want [`alpine3.15`](https://alpinelinux.org/posts/Alpine-3.15.0-released.html) instead?

genehynson: according to the docker tag itself, `alpine3.15` is on node v18 🤷 
3.15: https://hub.docker.com/layers/library/node/lts-alpine3.15/images/sha256-7b47aea5e936bb7d848fdc29ef3d5afb8835fa074eadcf082680b5f0cfb8f00b?context=explore

3.14: https://hub.docker.com/layers/library/node/lts-alpine3.14/images/sha256-e3c033748d2c218ab9a12d17c6ad9e25756be6517a45f33b175d418fa8b5582d?context=explore

hoorayimhelping (approved):

PR #6194, opened 2022-10-28
fix(buckets): clean up eslint warnings in buckets

blegesse-git:
Part of #4815 

breaking down 4815 by feature - pr cleans up linter warnings inside Buckets component.

### Checklist


src/buckets/components/BucketList.tsx
```diff
@@ -63,23 +63,27 @@ const BucketList: FC<Props> = ({
...
-    history.replace(`${url.pathname}?page=${page}`, null)
-    setCurrentPage(page)
-  }, [])
+  const paginate = useCallback(
+    (page: number) => {
+      const url = new URL(location.href)
+      history.replace(`${url.pathname}?page=${page}`, null)
+      setCurrentPage(page)
+    },
+    [history]
```
blegesse-git, line 72: Because `history` is used in the callback body, eslint wants it to be added in the dependency array. Two possible solutions here; disable eslint or add it to the array. I don't see issues with adding it since it won't be causing unnecessary re-renders.

PR #6198, opened 2022-10-28
refactor: remove require.context from endpoints

TCL735:
Part of #5819   

Removes the self-registering pattern that uses `require.context` for Flows Notification Endpoints

### DIFF
--- a/src/flows/pipes/Notification/endpoints.ts
+++ b/src/flows/pipes/Notification/endpoints.ts
@@ -5,7 +5,29 @@ export const TEST_NOTIFICATION = 'This is a test notification'
 export interface TypeLookup {
   [key: string]: EndpointTypeRegistration
 }
-export const ENDPOINT_DEFINITIONS: TypeLookup = {}
+
+import {Endpoint as AWS} from './endpoints/AWS'
+import {Endpoint as HTTP} from './endpoints/HTTP'
+import {Endpoint as Mailgun} from './endpoints/Mailgun'
+import {Endpoint as Mailjet} from './endpoints/Mailjet'
+import {Endpoint as PagerDuty} from './endpoints/PagerDuty'
+import {Endpoint as SendGrid} from './endpoints/SendGrid'
+import {Endpoint as Slack} from './endpoints/Slack'
+import {Endpoint as Telegram} from './endpoints/Telegram'
+import {Endpoint as Zenoss} from './endpoints/Zenoss'
+
+export const ENDPOINT_DEFINITIONS: TypeLookup = {
+  aws: new AWS(),
+  http: new HTTP(),
+  mailgun: new Mailgun(),
+  mailjet: new Mailjet(),
+  pagerduty: new PagerDuty(),
+  sendgrid: new SendGrid(),
+  slack: new Slack(),
+  telegram: new Telegram(),
+  zenoss: new Zenoss(),
+}
+
 export const ENDPOINT_ORDER = [
   'slack',
   'http',
@@ -22,25 +44,3 @@ export const ENDPOINT_ORDER = [
   'sensu',
   'zenoss',
 ]
-
-// NOTE: this loads in all the modules under the current directory
-// to make it easier to add new types
-const context = require.context(
-  './endpoints',
-  true,
-  /^\.\/([^\/])+\/index\.(ts|tsx)$/
-)
-context.keys().forEach(key => {
-  const module = context(key)
-  module.default((definition: EndpointTypeRegistration) => {
-    if (ENDPOINT_DEFINITIONS.hasOwnProperty(definition.type)) {
-      throw new Error(
-        `Endpoint of type [${definition.type}] has already been registered`
-      )
-    }
-
-    ENDPOINT_DEFINITIONS[definition.type] = {
-      ...definition,
-    }
-  })
-})
--- a/src/flows/pipes/Notification/endpoints/AWS/index.ts
+++ b/src/flows/pipes/Notification/endpoints/AWS/index.ts
@@ -1,43 +1,46 @@
+import {AWSReadOnly} from './readOnly'
+import {AWS} from './view'
+import {EndpointTypeRegistration} from 'src/types'
 import {TEST_NOTIFICATION} from 'src/flows/pipes/Notification/endpoints'
-import View from './view'
-import ReadOnly from './readOnly'
 
-export default register => {
-  register({
-    type: 'aws',
-    name: 'AWS SES Email',
-    data: {
-      url: 'https://email.your-aws-region.amazonaws.com/sendemail/v2/email/outbound-emails',
-      authAlgo: '',
-      accessKey: '',
-      credScope: '',
-      signedHeaders: '',
-      calcSignature: '',
-      email: '',
-    },
-    component: View,
-    readOnlyComponent: ReadOnly,
-    generateImports: () =>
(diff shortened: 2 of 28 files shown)


genehynson (commented):
hey sorry - can you explain what benefit we're gaining from this large refactor? I really liked this registration pattern because it allowed for new notification endpoints to be introduced and removed by simply dropping or removing a single file. This change seems to introduce higher coupling which I think we would be against?

TCL735:
> hey sorry - can you explain what benefit we're gaining from this large refactor? I really liked this registration pattern because it allowed for new notification endpoints to be introduced and removed by simply dropping or removing a single file. This change seems to introduce higher coupling which I think we would be against?

This seems to come up on every PR related to this story. I will have to defer to @kristinarobinson and @hoorayimhelping on this. This story was something that was planned, in addition to being a topic discussed internally with various stakeholders.

I am here to execute the story. It is not a passion project or necessarily something that i was personally advocating. However, I do agree with the general premise that we should remove dynamic loading where possible.

TCL735:
> Kind of confused by this change since this part of the codebase is within the scope of the QX team

True. But this doesn't change the behavior. See above regarding why this being done. TLDR: it was planned, and we had buy-in.

genehynson:
> This seems to come up on every PR related to this story. I will have to defer to @kristinarobinson and @hoorayimhelping on this. This story was something that was planned, in addition to being a topic discussed internally with various stakeholders.
> 
> I am here to execute the story. It is not a passion project or necessarily something that i was personally advocating. However, I do agree with the general premise that we should remove dynamic loading where possible.

Can you link those discussions? I was tagged to review this PR so I'm just trying to gain enough context to provide a valuable review. And I feel like the context around the benefit of this refactor is missing from the attached issue.

TCL735:
> > This seems to come up on every PR related to this story. I will have to defer to @kristinarobinson and @hoorayimhelping on this. This story was something that was planned, in addition to being a topic discussed internally with various stakeholders.
> > I am here to execute the story. It is not a passion project or necessarily something that i was personally advocating. However, I do agree with the general premise that we should remove dynamic loading where possible.
> 
> Can you link those discussions? I was tagged to review this PR so I'm just trying to gain enough context to provide a valuable review. And I feel like the context around the benefit of this refactor is missing from the attached issue.

I can't because I wasn't part of them. I am here to execute the story.

TCL735:
> > > Kind of confused by this change since this part of the codebase is within the scope of the QX team
> > 
> > 
> > True. But this doesn't change the behavior. See above regarding why this is being done. TLDR: it was planned, and we had buy-in.
> 
> Sure, but again, it seems out of scope. Just as if I were to rename visualizations internally on the FirstMile components, despited discussions that may have occurred outside the scope of the team owning those

I hear your concern. You are definitely code owners here.

But I just want to gently point out that your analogy is not quite right. If there were discussions on renaming the visualizations, then sure, anyone could execute this work. Just like there were discussions on removing this pattern.

It sounds like you were not part of those discussions, is what I am getting. Correct me if that's inaccurate. I wasn't part of those discussions either. I was given a story to execute. I am open to further discussing the necessity of these changes. I'm glad people are chiming in! Keep the comments coming.

We can take our time to resolve this. 👍

TCL735:
> > But I just want to gently point out that your analogy is not quite right. If there were discussions on renaming the visualizations, then sure, anyone could execute this work. Just like there were discussions on removing this pattern.
> 
> Historically that hasn't been the case though right? Discussions around parsing improvements that had been discussed were met with discussions on code ownership, despite being prioritized on sprints and having product buy-in. This seems like a purely engineering-based decision that introduces a new pattern that hasn't necessarily been communicated towards all UI teams.

We are going off on a tangent regarding the parsing work... but that was executed. And by the team that were not code owners.

Is being "met with discussions" a bad thing? Sorry. Just trying to address your concerns.

hoorayimhelping:
@asalem1 @genehynson @TCL735

I'm sorry that I'm coming in late. I was in a block of meetings while this discussion was taking place.

From what I've read of Gene and Ari's responses, I can see two main issues with this PR. One is that this work seems to be coming out of nowhere and not related to any sprint or product goals. The other is that the reasoning behind why this work is even valuable is unclear.

Like Timmy said, this was work decided by the UI Governance Board to be valuable as part of our architectural clean up. I don't want to get into a discussion about the merits of that decision on this PR. If there are concerns with it, I, or any other member of that group, am more than happy to address them separately. If you have concerns, you can reach out to me, Paul Hummer, Hannah, Kristina, or Randy. Having said that, the reason we made the decision is because we agreed this pattern is overly complicated for what it does, it's opaque if you don't already have experience with it, and it causes problem for static analysis tools.

Ari, I think I understand your frustration here. We essentially did to your team what we asked you not to do to our team. That's not really fair. I mentioned that to Timmy - I told him I thought this could have been communicated better than it was. I suggested that if we were going to live up to our own expectations of other teams, we should have written up a ticket describing the work, then had a discussion with the QX team on if and how it should get worked on and who should do it. This was also brought up in our retro. I said if we expect other teams to behave in a certain way, then we had better make sure that we behave that way first, otherwise we're just hypocrites. This is something we, the Dumplings team, will work on.

I think it's clear that the UI Governance Board needs to do a better job communicating what work our teams should be doing and why that work is important. I don't want that group to come off as some ivory tower cabal of bigwigs who make decisions in secret and drop them on the UI teams, and it sounds like it's gone too far in that direction. I'll make sure this communication issue gets addressed there.

In the meantime, let's hold off on worrying about this PR. I'll talk with Kristina and Timmy about this whole thing next week when she's feeling better and we can decide where to go from there.

One last thing: We're all on the same team, gentlemen, and we're all working towards the same goal.

genehynson:
> Like Timmy said, this was work decided by the UI Governance Board to be valuable as part of our architectural clean up. I don't want to get into a discussion about the merits of that decision on this PR. If there are concerns with it, I, or any other member of that group, am more than happy to address them separately. If you have concerns, you can reach out to me, Paul Hummer, Hannah, Kristina, or Randy. Having said that, the reason we made the decision is because we agreed this pattern is overly complicated for what it does, it's opaque if you don't already have experience with it, and it causes problem for static analysis tools.

Thanks for this explanation! I noticed that the list of individuals you listed here doesn't include any of the original authors or advocates of this design pattern. I think it would be worth have a discussion on this specific topic with Ari and/or myself on the benefits of this pattern since we introduced it intentionally. I'd just like the opportunity to explain the rationale for the pattern before its ripped away and deemed unmaintainable. 

> I think it's clear that the UI Governance Board needs to do a better job communicating what work our teams should be doing and why that work is important. I don't want that group to come off as some ivory tower cabal of bigwigs who make decisions in secret and drop them on the UI teams, and it sounds like it's gone too far in that direction. I'll make sure this communication issue gets addressed there.

+1, it was unclear from the issue attached _why_ the refactor is occurring. As a tagged reviewer, that context is critical to performing a valuable review. I would love to see some documentation produced on the best practices & design patterns desired by the governance board and some context on _why_. Otherwise it does feel like a small group of people on a hill making decisions as you described. 

> In the meantime, let's hold off on worrying about this PR. I'll talk with Kristina and Timmy about this whole thing next week when she's feeling better and we can decide where to go from there.

Sounds good, thanks! I appreciate the response!

TCL735:
per @kristinarobinson we are not doing this, because the entire encompassing feature may end up getting removed which would be a lot faster.

Closing.

PR #6199, opened 2022-10-28
chore: remove saveAsScript feature flag

asalem1:
This is routine removal of the saveAsScript feature flag

### DIFF
--- a/cypress/e2e/shared/editor.test.ts
+++ b/cypress/e2e/shared/editor.test.ts
@@ -107,7 +107,7 @@ describe('Editor+LSP communication', () => {
           if (!$toggle.hasClass('active')) {
             // hasClass is a jQuery function
             $toggle.click()
-            cy.getByTestID('flux-query-builder--menu').contains('Clear')
+            cy.getByTestID('flux-query-builder--menu').contains('New Script')
           }
         })
       })


cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -303,7 +301,6 @@ describe('Script Builder', () => {
       loginWithFlags({
         schemaComposition: false,
         newDataExplorer: true,
-        saveAsScript: false,
```
appletreeisyellow, line 306: Seems like the test fails at line 309 because there is no longer a "Clear" button

src/dataExplorer/components/FluxQueryBuilder.tsx
```diff
@@ -191,36 +161,30 @@ const FluxQueryBuilder: FC = () => {
...
+                  className="flux-query-builder__action-button"
+                  onClick={() => setOverlayType(OverlayType.SAVE)}
+                  status={
+                    hasChanged
+                      ? ComponentStatus.Default
+                      : ComponentStatus.Disabled
+                  }
+                  text="Save"
+                  testID="flux-query-builder--save-script"
+                  icon={IconFont.SaveOutline}
```
appletreeisyellow, line 185:
```suggestion
                  icon={IconFont.Save}
```
`SaveOutline` works fine in remocal but it's mapping pie chart in production and in tools. #6179 has more details about this issue. Let's use the solid icon `Save` for now until clockface fixes the icon file

appletreeisyellow (approved):

PR #6221, opened 2022-11-01
fix: display non-empty window period and update only when adaptive zoom is on

### LINKED ISSUE #6220: Bug: Adaptive Zoom - window period sometimes appears on cells with no value & updates incorrectly
On dashboard cells, sometimes `window period:` is displayed without any value. This is confusing, especially when the same cell has a visualization in it. Additionally, when the feature flag is off, the window period updates to a progressively smaller value when zooming even when no query has been made to get more granular data.

Screen Shot 2022-11-01 at 8 46 40 AM

## Proposal
- When the underlying string for the window period is empty, the `window period:` should be suppressed. Only show it for non-empty strings.
- When the adaptive zoom feature is off, the window period should not update.

TCL735:
Closes #6220  

When window period has no value (the underlying state variable is an empty string), suppress the display:

https://user-images.githubusercontent.com/10736577/199318411-5177280b-70ce-4a50-87ce-2657a99fccf8.mp4

When the adaptive zoom feature is OFF, do not update the window period:

https://user-images.githubusercontent.com/10736577/199318513-2adddedc-bb06-488e-9b65-770cf8981734.mp4


hoorayimhelping (approved):

TCL735:
Oops, i also needed to check for the user turning off the feature.

hoorayimhelping (approved):

TCL735:
I moved the adaptive zoom e2e tests to Cloud only because the OSS tests are sending in `adaptiveZoomHide` as `true` in the user's customization settings for the line graph.

wdoconnell (approved):

PR #6230, opened 2022-11-02
chore: makes the UI more usable for error states, and more seemless for loading states

asalem1:
This closes an issue where the API for name searches weren't working with the API due to a postgres sql interpolation error. This is a temporary resolution to a permanent problem that is being addressed concurrently


src/dataExplorer/components/OpenScript.tsx
```diff
@@ -42,7 +41,7 @@ const OpenScript: FC<Props> = ({onCancel, onClose}) => {
   const history = useHistory()
   const org = useSelector(getOrg)
 
-  const handleGetScripts = useCallback(async (name: string = '') => {
```
appletreeisyellow, line 45: Not a big deal - should the `name` in line 48 also be removed since it is not declared anywhere? When I test it on my local, the API call works fine

appletreeisyellow (dismissed):
LGTM! Just one comment about a remnant param

appletreeisyellow (approved):

PR #6246, opened 2022-11-03
feat: add a copy to clipboard mechanism for the invokable url

### LINKED ISSUE #6141: Invokable Script URL Display
## User Story

<p></p>

As an invokable script script author, I want a way to know the right parameter to use use in my application for calling the invokable script so that I can update the code in my application to use it\.

<p></p>

## Acceptance Criteria

<p></p>

- When viewing\/editing script properties, the invokable script URL for the script being edited will be shown

- The URL will not be shown if a script has not been saved yet \(untitled scripts will not be eligible\)

- A copy to clipboard option will be available to copy the URL so the user can paste it into their code

- The URL will include the users region, the scripts api path and the script ID to formulate the complete path \(minus parameters\)

<p></p>

## Notes

<p></p>

- Option for "making" the script an invokable script \(via checkbox or something\) to be discussed \- but may require back\-end changes to enforce \- so it may not be an option at this time, even though it is desirable

- \<region\>\/api\/v2\/scripts\/\<scriptID\>

<p></p>

<p></p>

asalem1:
Closes #6141

This PR adds some conditional logic to the SaveOverlay to allow for users to copy to clipboard their specific invokable script URL. This PR also updates the renaming logic in the DataExplorerPage & the SaveOverlay. Whereas previously, renaming the script would persist the name change in local storage, it is now making an explicit save to the API based on user input when the makes those changes. In the case of the SaveOverlay, those changes are stored locally until the user explicitly saves / updates their script so that those changes aren't persisted to the database or localstorage until the user actually saves them. This fixes a bug where the previous implementation was storing the temporary changes in local storage and was not persisting the data in the database, leading to a confusing UX 

invokable-url

### DIFF
--- a/src/dataExplorer/components/DataExplorerPage.tsx
+++ b/src/dataExplorer/components/DataExplorerPage.tsx
@@ -44,7 +44,7 @@ import {SCRIPT_EDITOR_PARAMS} from 'src/dataExplorer/components/resources'
 
 const DataExplorerPageHeader: FC = () => {
   const {fluxQueryBuilder, setFluxQueryBuilder} = useContext(AppSettingContext)
-  const {resource, setResource} = useContext(PersistanceContext)
+  const {resource, save} = useContext(PersistanceContext)
   const history = useHistory()
 
   const toggleSlider = () => {
@@ -60,13 +60,8 @@ const DataExplorerPageHeader: FC = () => {
   }
 
   const handleRename = (name: string) => {
-    setResource({
-      ...resource,
-      data: {
-        ...resource.data,
-        name,
-      },
-    })
+    resource.data.name = name
+    save(resource?.language)
   }
 
   const showNewExplorer = fluxQueryBuilder && isFlagEnabled('newDataExplorer')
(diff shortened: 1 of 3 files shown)


src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -60,32 +60,23 @@ const DataExplorerPageHeader: FC = () => {
...
   const handleRename = (name: string) => {
-    setResource({
-      ...resource,
-      data: {
-        ...resource.data,
-        name,
-      },
-    })
+    resource.data.name = name
+    save(resource?.language)
```
rockstar, line 64: This feels like an awkward api. `save` takes one argument and it's the language if the resource? Is there a better API, or a better name than `save`?

rockstar (approved):
Thanks for getting to this. I had one tiny question about the patch itself.

This looks like two disparate changes. I'm curious why they're in the same patch. What am I missing that makes them related?

src/dataExplorer/components/DataExplorerPage.tsx
rockstar: > they are two separate things.

In this case, it's not a huge deal, but we shouldn't make a habit out of merging disparate changes. I realize our CI can be a pain. Working around that pain only makes the pain stay around.

> that's a byproduct of the patchwork design we have for all this work.

Cool. Let's make sure that issue is created and scheduled for next sprint before queuing this for merge.

PR #6249, opened 2022-11-03
feat(tasks): create new task scheduler page

### LINKED ISSUE #6185: New Tasks Scheduler - Select a Script - Text and Dropdown Behavior
## ACs

- User Flow: Select the task tab from navigation > Display the task list page > click + create task > all users should then see the NEW Task scheduler. (NO Toggle for feature)
- The new tasks UI will have a header text of "Create Task" since that's the core user action for that page.
- The first main section of this page will be "Select a Script" as shown in Figma. 
- There should be a dropdown underneath the text. There should be additional text inside the dropdown before the user clicks it that says "Select a Script*" with a red asterisk next to it indicating that it's a mandatory step. See Figma. 
- If the user clicks on the "Select a Script" dropdown, there should be a search bar inside the dropdown with grey text indicating to the user that they can "Search Scripts". 
- The dropdown should also display the option for users to "+ Create Script in Editor". 
- Underneath the dropdown display the text "You can create or edit Scripts in the Script Editor" ...Script Editor should be hyperlinked and send the user to the NEW Data Explorer script editor (auto-toggle it on). 
- If the user clicks on "+ Create Script in Editor, they should be navigated to the new query experience data explorer page. Deep link the data explorer text to navigate the user to the NEW query experience script editor. If the user has previously toggled off the new script editor, automatically toggle it on when you navigate the user to the NEW query experience script editor. 
-  The dropdown should display a scrollable list of all the saved scripts that belong to that organization. (Filterable list via a predictive text search bar, NOT typeahead dropdown)
- If the user chose a script that has parameters, then display the parameters value entry fields as sub-steps to the right of the page. Details for this behavior are outlined in another ticket/user story. 
- If the user chose a script that does not have parameters, then do not display any parameter value fields to the right of the page. 
- Underneath the "select a script" dropdown should be an "Edit Script" button. This button should remain disabled until the user selects a script. 
- If the user selects a script and clicks the "Edit Script" button, the system should navigate the user to the NEW data explorer script editor (auto-toggle it on for the user). The script that the user selected should display in the script editor in the new data explorer script editor. 

## Implementation Details
- Contact @gavincabbage if any help is needed leveraging the Invokable Scripts APIs
- Figma: https://www.figma.com/file/r91KxurPL3VQ7JMqc9Ldqo/New-Tasks-Experience?node-id=1669%3A33743
- Use this to retrieve a list of scripts: /api/v2/scripts
- No need to parse the flux, you can pull the params from /api/v2/scripts/{scriptName}/params 
- See Invokable Scripts API here: https://docs.influxdata.com/influxdb/cloud/api/#operation/PostScripts
- Request is out to the QX team to enable a way for users to navigate back to the tasks page from the new QX page: https://github.com/influxdata/ui/issues/6173

blegesse-git:
Closes #6185 

Behind feature flag `tasksUiEnhancements`
New create task form.  No functionality yet; Script selection and Schedule Task feature coming soon. 
[Figma](https://www.figma.com/file/r91KxurPL3VQ7JMqc9Ldqo/New-Tasks-Experience?node-id=1655%3A28429)
Screen Shot 2022-11-03 at 8 32 15 PM

Screen Shot 2022-11-03 at 7 27 43 PM

https://user-images.githubusercontent.com/66275100/199859727-31143864-77af-4ddc-9a36-80176ff94bdc.mov


src/tasks/components/NewTaskScheduler/ScriptSelector.tsx
```diff
@@ -0,0 +1,127 @@
...
+            </Dropdown.Button>
+          )}
+          menu={onCollapse => (
+            <Dropdown.Menu onCollapse={onCollapse}>
+              <Input
+                icon={IconFont.Search_New}
+                size={ComponentSize.Small}
+                value={searchTerm}
+                placeholder="Search Scripts"
+                onChange={evt => handleSearchTerm(evt.target.value)}
```
hoorayimhelping, line 110: I think we want to change how this system works. Right now, the task scheduler fetches the scripts and passes them down to the scripts selector, which couples the ScriptSelector to the task scheduler. Or more generically, it forces whatever component uses the ScriptSelector to fetch scripts and re-implement searching/filter on them. The way this event handler works currently, it'll re-fetch all the scripts every time there's a keystroke (after some debouncing time), after fetching all scripts up to `limit` on component load.

I think what this system should do instead is that this component should fetch all the scripts once, then when the user types a script name in the search box, this should filter the names of the scripts it already fetched. This will be much more efficient in network usage (it'll only make a single request, rather than up to `n` requests where `n` is number of keystrokes) and it should be much faster, since it'll filter an in-memory list, rather than having to make a trip across the newtork.

Take a look at this [blog post](http://highscalability.com/blog/2011/1/26/google-pro-tip-use-back-of-the-envelope-calculations-to-choo.html) to get an idea of the scale of network operations. If we assume this data isn't in the CPU cache but is in main memory, and that we're getting worst-performance when accessing it from memory, the differences in filtering in memory vs making a network request look like this:

```
Read 1 MB sequentially from memory 250,000 ns
Read 1 MB sequentially from network 10,000,000 ns
````

That's two orders of magnitude slower to fetch something across the network.

File: src/tasks/components/NewTaskScheduler/ScriptSelector.tsx [hunk 2]
```diff
@@ -0,0 +1,127 @@
...
+        You can create or edit Scripts in the{' '}
+        <Link to={`/orgs/${orgID}/data-explorer?fluxScriptEditor`}>
+          Script Editor.
+        </Link>
+      </p>
+    </div>
+  )
+}
+
+export default ScriptSelector
```
hoorayimhelping, line 127: we're doing away with default exports. [Here's](https://www.codingzeal.com/posts/developer-dark-arts-default-exports) a decent explanation why.

This should end up looking something like:

```ts
export const ScriptSelector: FC<Props> = ({
```

src/tasks/components/NewTaskScheduler/TaskScheduler.scss
```diff
@@ -0,0 +1,46 @@
+.cf-form--footer {
+  position: relative;
+  margin-top: 20%;
+  justify-content: end !important;
```
hoorayimhelping, line 4: do we need these `!important` in these files? is it possible to get this style using a more specific selector?

src/tasks/components/NewTaskScheduler/TaskScheduler.tsx
```diff
@@ -0,0 +1,181 @@
...
+  IconFont,
+  Input,
+  Page,
+  SelectGroup,
+} from '@influxdata/clockface'
+import TaskSchedulerFormField from 'src/tasks/components/NewTaskScheduler/TaskSchedulerFormField'
+import ScriptSelector from 'src/tasks/components/NewTaskScheduler/ScriptSelector'
+
+// Utils
+import {debouncer} from 'src/dataExplorer/shared/utils'
```
hoorayimhelping, line 21: I think we should use lodash's `debounce`, rather than the home rolled one, since it's maintained and more familiar to most people.

src/tasks/components/NewTaskScheduler/TaskScheduler.tsx
```diff
@@ -0,0 +1,181 @@
...
+import {SafeBlankLink} from 'src/utils/SafeBlankLink'
+
+// Types
+import {RemoteDataState, TaskOptions, TaskSchedule} from 'src/types'
+
+import 'src/tasks/components/NewTaskScheduler/TaskScheduler.scss'
+
+const getScripts = require('src/client/scriptsRoutes').getScripts
```
hoorayimhelping, line 29: why the `require`? can't this be `import`ed like

```ts
import {getScripts} from 'src/client/scriptsRoutes'
```

?

src/tasks/components/NewTaskScheduler/TaskScheduler.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const [loading, setLoading] = useState(RemoteDataState.NotStarted)
+  const [scripts, setScripts] = useState([])
+  const [searchTerm, setSearchTerm] = useState('')
+  const [selectedScript, setSelectedScript] = useState<any>({})
+  const [taskName, setTaskName] = useState('')
+
+  const handleGetScripts = useCallback(async (name: string = '') => {
+    try {
+      setLoading(RemoteDataState.Loading)
+      const resp = await getScripts({query: {limit: 250, name}})
```
hoorayimhelping, line 51: we're moving away from making calls to the routes layer directly in components. This logic should get moved to an api layer, similar to how the identity stuff works: https://github.com/influxdata/ui/blob/master/src/identity/apis/account.ts, https://github.com/influxdata/ui/blob/master/src/identity/apis, and also a thunk layer, since it's dispatching actions to the redux state, like here: https://github.com/influxdata/ui/blob/49952435971518aedcd14c3b1b5a3c92713190cd/src/identity/actions/thunks/index.ts#L61

Making a request across the network and then translating the response into something the application can use should be the responsibility of the api layer. The application (like redux, and components) shouldn't be aware of response status codes. That's for the api layer to handle. Notice how the api files in `identity` translate error codes (`401`, `500`, `404` into named error types `UnauthorizedError`, `ServerError`, `NotFound`). This decouples the application code from the code to handle network responses.

If you need help with this, you can pair with me or Bill to understand the reasoning.

src/tasks/components/NewTaskScheduler/TaskScheduler.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+      if (resp.status !== 200) {
+        throw new Error(resp.data.message)
+      }
+
+      setScripts(resp.data.scripts)
+      setLoading(RemoteDataState.Done)
+    } catch (error) {
+      setLoading(RemoteDataState.Error)
+      console.error({error})
```
hoorayimhelping, line 61: 🙅 `console.error` does nothing for us. It spits out an error in the user's console, and provides no information to them. This should be replaced (in the thunk layer) with either a honeybadger call and/or notification that there was an error fetching scripts. We shouldn't add any `console.error` calls under the majority of circumstances anymore

src/tasks/components/NewTaskScheduler/TaskScheduler.tsx
```diff
@@ -0,0 +1,181 @@
...
+        <Grid>
+          <Grid.Row>
+            <Grid.Column widthXS={Columns.Five}>
+              <p>
+                A task is a scheduled script that takes a stream of input data,
+                modifies or analyzes it, then writes the modified data back to
+                influxDB or perform other actions.{' '}
+                <SafeBlankLink href="https://docs.influxdata.com/influxdb/cloud/process-data/get-started/">
+                  Learn More.
+                </SafeBlankLink>{' '}
```
hoorayimhelping, line 95: 👍

src/tasks/components/NewTaskScheduler/TaskScheduler.tsx
```diff
@@ -0,0 +1,181 @@
...
+import {RemoteDataState, TaskOptions, TaskSchedule} from 'src/types'
+
+import 'src/tasks/components/NewTaskScheduler/TaskScheduler.scss'
+
+const getScripts = require('src/client/scriptsRoutes').getScripts
+
+interface Props {
+  taskOptions: TaskOptions
+  onChangeScheduleType: (schedule: TaskSchedule) => void
+  onChangeInput: (e: ChangeEvent<HTMLInputElement>) => void
```
hoorayimhelping, line 34:
```suggestion
  onChangeInput: (event: ChangeEvent<HTMLInputElement>) => void
```

src/tasks/components/NewTaskScheduler/TaskScheduler.tsx
```diff
@@ -0,0 +1,181 @@
...
+              </Form.Footer>
+            </Grid.Column>
+          </Grid.Row>
+        </Grid>
+      </Form>
+    </Page.Contents>
+  )
+}
+
+export default TaskScheduler
```
hoorayimhelping, line 181: same bit here about default exports

hoorayimhelping (changes requested):
I have a few comments here. Let me know if any of them are unclear or you need help on them and want to pair up

blegesse-git:
Pr needs to be split into smaller prs.

PR #6257, opened 2022-11-04
chore: consolidate imports to flux-lsp-browser

rockstar:
There is currently a shim for both `parse` and `format_from_js_file` for
using those functions in a node environment. However, not all call to
`parse` or `format_from_js_file` use that shim. This patch fixes that.

_Why is this needed?_, you may ask yourself. Well, it's not, but it certainlys
 makes it easier toidentify places that need to be ported over to the
(in progress) async versions of both of those functions. We'll know we
got all the places it's used when we can delete `parse` and
`format_from_js_file` and nothing breaks.

### Checklist

### DIFF
--- a/src/languageSupport/languages/flux/parser/index.ts
+++ b/src/languageSupport/languages/flux/parser/index.ts
@@ -8,7 +8,7 @@ import {
         a node environment (this is only for handling tests). If a test requires a specific AST result
         then you will need to mock that out in the test.
 */
-export const parse = (script): File => {
+export const parse = (script: string): File => {
   if (window) {
     return flux_parse(script)
   } else {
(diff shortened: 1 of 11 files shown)


src/languageSupport/languages/flux/parser/index.ts
```diff
@@ -27,10 +27,10 @@ export const parse = (script): File => {
   }
 }
 
-export const format_from_js_file = (script): string => {
+export const format_from_js_file = (script: File): string => {
   if (window) {
     return flux_format_from_js_file(script)
   } else {
-    return script
+    return ''
```
wiedld, line 34: Check with @genehynson on this change.

rockstar: I'm unsure this has ever been used properly. The change happened because adding type information makes in clear that we were _saying_ we'd return a string, but in this case, we're returning a AST object. I don't think this ever _actually_ affected anything, the tests pass, and it's close to the raw AST we're already returning.

genehynson: I did some digging on how this function is used and I think this change is fine. It's probably safer anyways to return `''` here instead of the AST obj that we provide as `script` ✅

genehynson (approved):

PR #6275, opened 2022-11-07
fix(6240): have Sidebar container always render

### LINKED ISSUE #6240: Visualization Options not Available when editing SQL Script
When editing a SQL script, the graphic/visualization customization options are not available.  This is because the flux function panel has been removed from the SQL view; but that area also was used for showing the visualization options.  We need to find a way to expand that panel and show the visualization customization options and collapse it back when the user is finished.

wiedld:
Closes influxdata/ui#6240 

Sidebar container is used for other launched menus (such as graph customize). Have the sidebar container always exist, and only remove flux-specific sidebar menus when in SQL mode.

https://user-images.githubusercontent.com/10232835/200433616-3fbdd867-ace2-4627-a040-c0c933c12492.mov

### Checklist
- [x] ~Feature flagged, if applicable~ already done.

### DIFF
--- a/src/dataExplorer/components/FluxQueryBuilder.tsx
+++ b/src/dataExplorer/components/FluxQueryBuilder.tsx
@@ -220,10 +220,7 @@ const FluxQueryBuilder: FC = () => {
               <ResultsPane />
             </DraggableResizer.Panel>
             <DraggableResizer.Panel isCollapsible={true}>
-              {isFlagEnabled('uiSqlSupport') &&
-              resource?.language === LanguageType.SQL ? null : (
-                <Sidebar />
-              )}
+              <Sidebar />
             </DraggableResizer.Panel>
           </DraggableResizer>
         </FlexBox>
--- a/src/dataExplorer/components/Sidebar.tsx
+++ b/src/dataExplorer/components/Sidebar.tsx
@@ -16,9 +16,11 @@ import {
 // Contexts
 import {SidebarContext} from 'src/dataExplorer/context/sidebar'
 import {EditorContext} from 'src/shared/contexts/editor'
+import {PersistanceContext} from 'src/dataExplorer/context/persistance'
 
 // Types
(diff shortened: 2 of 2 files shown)


rockstar (approved):
I recall there be some funky stuff around feature flags in tests, but it looks like we're removing a feature flag, so maybe it doesn't apply. Is it possible to add some tests here, to ensure that we don't lose the sidebar some unforeseen reason in the future? I'm not quite sure what's appropriate to ask for, since I don't know what different levels of tests we have, and whether those are reasonable.

Approved, assuming the answers are not too much of an issue.

wiedld:
> Is it possible to add some tests here, to ensure that we don't lose the sidebar some unforeseen reason in the future? 

The SQL editing (with the sidebar still present), is only available in iox. Therefore we would need to be deploying remocal-iox in our CI tests suites, in order to write a test for this use case.

It's possible...but we would need to have a separate ticket, with crossteam coordination with deployments, in order to make this work. One could imagine this ticket being: (a) spin up and run (in parallel) all chrome and firefox tests on remocal-iox, as well as tsm-remocal (current), and (b) ensure all remocal-iox tests are passing, and (c) make a simple approach (in the cypress tests) for running certain iox-only tests (a.k.a. these SQL tests).

rockstar:
Hm, yeah, that's a good point. Sounds like our CI and deployments are lagging behind a bit. Can you see if there's an issue open about this, and if not, file it? Then just file an issue about this missing test and link to that issue. I think that will suffice, for now.

rockstar:
To be clear, I think those issues need to be created before we merge this. They're dependent tasks, and in order for us to consider our part of this "done," we need to make sure we leave the bookmarks for the outstanding tasks.

wiedld:
I was already making the ticket when I hit merge.
https://github.com/influxdata/ui/issues/6281
https://github.com/influxdata/ui/issues/6282

PR #6291, opened 2022-11-08
chore: remove `newQueryBuilder` in new script editor

### LINKED ISSUE #6201: Remnants from Previous Query is shown in Schema Browser when doing a new Query
When selecting a measurement that has multiple tag keys and values and then switching to a measurement that has only 1 tag key, I am observing that it is showing in the list tag keys from my previous query, even though they are not applicable to the measurement I have switched to (somehow the tag keys are being cached, maybe?)

See video..  I can invite you to my org to reproduce if you can't reproduce on your own.

appletreeisyellow:
Closes #6201 

Since the feature flag `newQueryBuilder` was originally created for the query builder in the old Data Explorer and Notebooks, there is no need to check for this flag in new Script Editor. So the main purpose of this PR is to remove the logic check in if statements for `isFlagEnabled('newQueryBuilder')`. I also refactored the code to remove un-used variables.

### Checklist

~~- [ ] Feature flagged, if applicable~~

### DIFF
--- a/src/dataExplorer/components/Schema.tsx
+++ b/src/dataExplorer/components/Schema.tsx
@@ -73,7 +73,7 @@ const Schema: FC = () => {
       <FieldsProvider scope={scope}>
         <TagsProvider scope={scope}>
           <FluxQueryBuilderProvider>
-            <BucketProvider scope={scope} omitSampleData>
+            <BucketProvider scope={scope}>
               <div className="scroll--container">
                 <DapperScrollbars>
                   <div className="schema-browser" data-testid="schema-browser">
--- a/src/dataExplorer/components/Schema.tsx
+++ b/src/dataExplorer/components/Schema.tsx
@@ -73,7 +73,7 @@ const Schema: FC = () => {
       <FieldsProvider scope={scope}>
         <TagsProvider scope={scope}>
           <FluxQueryBuilderProvider>
-            <BucketProvider scope={scope}>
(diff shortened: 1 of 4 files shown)


appletreeisyellow:
These are good questions! 

> do we want to follow this up with any lessons learned? 

When I wrote the query texts in the new Script Editor, I referenced them from the Notebook where the feature flag `newQueryBuilder` was there. One lesson I learned is to always check and understand the usage and scope of a feature flag before applying it to a new piece of code, especially in a different scope of work (Notebooks vs new Script Editor). In this case, the flag `newQueryBuilder` didn't need to be there in the first place. 

> Any takeaways on whether we want this flag in our codebase at all given what we know now?

I'd say it depends. My understanding is that there are two different use cases of the feature flag `newQueryBuilder`. 1) it controls the UI of the query builder (force measurement) in the old Data Explorer and the Notebooks (requested from #4253 & #4288), 2) it determines what flux query text to send to the backend, e.g. [notebook QueryBuilder context](https://github.com/influxdata/ui/blob/master/src/flows/pipes/QueryBuilder/context.tsx#L280).

For the first use case, it depends on how product feels about removing the feature flag. For the second use case, I think the feature flag can be removed in the if statements since the [flux schema package](https://docs.influxdata.com/flux/v0.x/stdlib/influxdata/influxdb/schema/) is designed for getting the InfluxDB data schema and we can use this package regardless the `newQueryBuilder` flag.

Issue #6239, opened 2022-11-02
Spike: Can we shrink our monaco-editor?


rockstar:
![image](https://user-images.githubusercontent.com/60306/199589201-4e4cb131-fe1b-47ac-bcfa-6f72b532eca8.png)

`monaco-editor` accounts for almost 1/3 of our webpack bundle size. Maybe that's just the state of affairs, but it'd be nice if we could investigate the `features` attribute of `MonacoWebpackPlugin` and see if we can turn off some features and deliver a smaller bundle as a result.

wiedld:
Agreed with overall premise. To investigate using the `features` part of the webpack plugin, to reduce payload size. We were using this approach before, but removed it since there was not a significant impact on page load time. But it's always worth a re-visit.

.

## Previous code, using `features`
* removed code in [this PR](https://github.com/influxdata/ui/pull/4293), while also bumping all the major editor libs.
* there was [increased bundle size](https://github.com/influxdata/ui/pull/4293#issuecomment-1086540589), but [no significant increase in page load time](https://github.com/influxdata/ui/pull/4293#issuecomment-1092291553).

.

## Current (as of 2022/11/09) re-done bundle analysis:
Analysis done using [webpack query analyzer](https://webpack.js.org/api/cli/#analyzing-bundle).

### All bundles:
All bundles are not served for all UI subapplications. So this represents all bundles across the app:
<img width="1200" alt="Screen Shot 2022-11-09 at 12 55 57 PM" src="https://user-images.githubusercontent.com/10232835/200941075-8ff20673-ae74-41dd-8e6e-4341fd85e08a.png">

Usually one chooses the entry point for this analysis, but the webpack bundler analyzer could not parse that. So instead, I attempted an ad hoc analysis below.

### Monaco-editor bundles:
Gzipped size (what's transmitted over the wire):
<img width="350" alt="Screen Shot 2022-11-09 at 12 59 24 PM" src="https://user-images.githubusercontent.com/10232835/200941527-8b4fb520-7619-488c-bbcb-13008e5d1963.png">


**Total:** (833.74+107.54+72.83)/4780 = **21%**

Note that the json.worker is only sent when we are using a json editor. But that doesn't really move the needle.

rockstar:
We've been talking very seriously about abandoning the chunking, mostly because it causes many more problems than it solves for us. It's good analysis, to see it through the lens of chunks, but it's probably not the be-all-end-all for us.

PR #6319, opened 2022-11-16
build: upgrade webpack to version 5

### LINKED ISSUE #2526: Webpack upgrade to version 5
Webpack 5 has a new Rule.parser.parse that allows a custom parsing function to turn any file into JSON or other formats. This is needed for the Telegraf Refresh epic that is ongoing, in order to serve up the .conf files and use them as needed--when the user selects a plugin.

- Upgrade webpack to version 5
- Ensure nothing is broken

hoorayimhelping:
Closes #2526

index

### Updates webpack to version 5.
- Removes building a vendor bundle in dev environments (our vendor config doesn't work under webpack 5)
  - Removes the call to it that ran before the prod bundle as this has no affect on prod bundles
- Makes use of new asset modules (replaces `file-loader`, and `raw-loader` with new webpack built in loaders)
- Upgrades plugin versions
- Updates configs to conform to Webpack 5 config shape

### Checklist


package.json
```diff
@@ -110,11 +110,11 @@
...
     "@typescript-eslint/eslint-plugin": "^5.35.1",
     "@typescript-eslint/parser": "^5.41.0",
-    "babel-loader": "^8.2.5",
-    "clean-webpack-plugin": "^3.0.0",
+    "babel-loader": "^9.1.0",
+    "clean-webpack-plugin": "^4.0.0",
     "clipboardy": "^2.3.0",
     "cross-env": "^7.0.3",
-    "css-loader": "^5.2.7",
+    "css-loader": "^6.7.2",
```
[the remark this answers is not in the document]
hoorayimhelping: Yeah, I think so. Good suggestion!

package.json
```diff
@@ -16,14 +16,14 @@
     "src": "./src"
   },
   "scripts": {
-    "start": "yarn install && yarn generate && cross-env TS_NODE_PROJECT=\"webpack.tsconfig.json\" && yarn run build:vendor && yarn run start:dev",
-    "start:cloud": "yarn install && yarn generate && cross-env TS_NODE_PROJECT=\"webpack.tsconfig.json\" && yarn run build:vendor && yarn run start:dev-cloud",
```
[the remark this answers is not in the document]
wiedld: Ah, wait. May have answered my own question. Webpack docs suggest that the `optimization.splitChunks.cacheGroups` will bundle vendors separately as the default. But I believe we only use splitChunks in the prod build, not the dev build.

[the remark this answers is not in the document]
hoorayimhelping: >It looks like the `webpack.vendor.ts` is never consumed anymore. Is this correct? So we are undoing the separate bundling as per [this example in the docs](https://webpack.js.org/concepts/entry-points/#separate-app-and-vendor-entries)?

@wiedld This is correct. I had trouble getting the vendor configuration to run without erroring out, so I tried removing it to see if it worked. From what I read, vendor bundles are a dev-level optimization, since on a production build, there isn't any reloading happening, so optimizing our code from our vendors' code isn't as important.

I'll look into using `splitChunks` to achieve something similar, thanks.

webpack.dev.ts
```diff
@@ -23,7 +23,7 @@ module.exports = mergeDev(commonDev, {
     poll: 1000,
   },
   devServer: {
-    hot: true,
+    hot: false,
```
wiedld, line 26: 😱 ??

hoorayimhelping: Before I say anything, this isn't necessarily going to stay. I was just experimenting with settings.

Anyway, .I was reading the docs, and this bit from the document on [Hot Module Replacement](https://webpack.js.org/concepts/hot-module-replacement/) caught my eye (emphasis mine):

>Hot Module Replacement (HMR) exchanges, adds, or removes [modules](https://webpack.js.org/concepts/modules/) while an application is running, **without a full reload.** This can significantly speed up development in a few ways...

As long as I've been developing our application, it's done full reloads, not module replacement. This got me thinking - what if our dev config is misconfigured or not working as intended? As far as I can tell, changing `hot` to `false` does not change the behavior of the application in development mode when it comes to changing a piece of code.

webpack.common.ts
```diff
@@ -17,12 +16,31 @@ const {
 const MONACO_DIR = path.resolve(__dirname, './node_modules/monaco-editor')
 
 module.exports = {
+  mode: 'development',
```
hoorayimhelping, line 19: all configs that use `common` as their base other than `production` are run in a dev environment. `production` config overrides the `mode`.

webpack.common.ts
```diff
@@ -37,18 +55,14 @@ module.exports = {
         'node_modules/monaco-languageclient/lib/vscode-compatibility'
       ),
     },
-    extensions: ['.tsx', '.ts', '.js', '.wasm'],
+    extensions: ['.tsx', '.ts', '.js', '.wasm', '...'],
```
hoorayimhelping, line 58: this `'...'`, addition was recommended by a webpack maintainer in a test repo as a best practice that automatically includes the default values for `extensions`

webpack.vendor.ts
```diff
@@ -86,13 +91,13 @@ module.exports = {
     }),
   ],
   stats: {
-    colors: true,
+    assetsSort: '!size',
     children: false,
+    colors: true,
+    errorDetails: true,
+    excludeAssets: [/\.(hot-update|woff|eot|ttf|svg|ico|png|wasm)/],
```
hoorayimhelping, line 98: alphabetization

.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           command: |
             set +e
             TESTFILES=$(circleci tests glob "src/**/*.test.ts*" | circleci tests split --split-by=timings)
-            npx yarn test:circleci $TESTFILES
+            npx yarn test:circleci $TESTFILES --forceExit --detectOpenHandles
       - store_test_results:
```
wdoconnell, line 28: What's the thinking behind this change?

Per jest docs, there's a performance penalty associated with enabling this:
https://jestjs.io/docs/cli#--detectopenhandles

package.json
```diff
@@ -38,7 +38,7 @@
     "test:e2e:all": "yarn test:e2e:clean && yarn test:e2e; yarn test:e2e:report;",
     "test:e2e:kind": "CYPRESS_dexUrl=http://dex-twodotoh.a.influxcloud.dev.local CYPRESS_baseUrl=https://twodotoh.a.influxcloud.dev.local CYPRESS_password=password cypress open --browser chrome --config chromeWebSecurity=false",
     "test:e2e:remocal": "CYPRESS_dexUrl=https://dex-${NS}.remocal.influxdev.co CYPRESS_baseUrl=https://${NS}.remocal.influxdev.co CYPRESS_password=password cypress open --browser chrome --config chromeWebSecurity=false",
-    "test:circleci": "yarn run test:ci --maxWorkers=2",
+    "test:circleci": "yarn run test:ci --runInBand",
```
wdoconnell, line 41: Similar question to above - why do we want tests running serially here? https://jestjs.io/docs/cli#--runinband

wdoconnell (commented):
LGTM, but had two questions about some flag changes related to jest and circleCI.

package.json
abshierjoel, line 41: Is there a reason we'd want these to run in serially instead?

abshierjoel (dismissed):
LGTM 🚀 
I just had one question about the circleci tests.

package.json
wiedld: From what I can see, the unit test times went from [<3min to 4-6minutes](https://app.circleci.com/pipelines/github/influxdata/ui?branch=bucky_upgrade_webpack) the moment that [the commit went in](https://github.com/influxdata/ui/pull/6319/commits/081062727a3708d70b3dd6f5ee105ce39fcdf7a4) to change these flags. This is not a long time, especially given how long monitor-ci takes.

However, I'm not sure if we need this change.

I would note that when this change/commit went in, the CI unit test began passing instead of the [previous failure](https://app.circleci.com/pipelines/github/influxdata/ui/18960/workflows/06663776-00be-4648-9a75-6be64c2c6e86/jobs/105184). However, the previous failure was in only the variables unit test -- which is now skipped by another commit pushed at the same time. Therefore, I think it may be worthwhile to see if we really need this change?

webpack.prod.ts
```diff
@@ -31,17 +31,13 @@ module.exports = mergeWebpack(commonWebpack, {
...
+    minimize: true,
     minimizer: [
       new TerserJSPlugin({
-        cache: true,
         parallel: 2,
-        sourceMap: true,
       }),
       new OptimizeCSSAssetsPlugin({}),
     ],
-    splitChunks: {
```
wiedld, line 42: Double checking. The webpack merge is a deep merge, correct? Because the configuration of `chunks: 'all'` is a [good thing](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunkschunks) we want to make sure stays in.

wiedld (commented):
Thank you for removing the vendor config!

Do you mind sanity checking the bundle analyzer? A baseline comparison could be t[his analysis done <1 month ago](https://github.com/influxdata/ui/issues/6239#issuecomment-1309569345).  (Don't care about the editor bundles per se. Just in comparison to the screenshot of all bundles. TY!)

package.json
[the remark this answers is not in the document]
wiedld: thanks for removing the vendor config!

package.json
hoorayimhelping: @abshierjoel it turns out that our jest tests were running out of memory. The general way to fix it on the internet is to set `maxWorkers` <= the number of CPUs your circle instance has and go from there. Since setting `maxWorkers` to 2 was causing problems, I dropped the number down to `1`, which is an alias for `runInBand`.

I talked with Gene about this. This isn't necessarily the best way to do it - there are tradeoffs. Lowering the number of CPUs will make the tests run for longer which will cost more money, but we'll still be on a relatively cheap CPU plan. The alternative is to increase the number of CPUs for this circleci job which will also cost more money. If this ends up taking too long or being too expensive, we can increase the resources available to our circle instance.

webpack.prod.ts
hoorayimhelping: @wiedld thanks for the review! I'm almost certain that the prod config will inherit `splitChunks: 'all'` from the common config, but I will double check it.

I'll also look at the bundle size analysis. Good suggestions, love where your head is.

hoorayimhelping:
@wiedld Using the analyzer plugin, I generated these against the production config:

Wepback 5's bundle size is smaller by about 30% (27.39 MB vs 38.7 MB).
The largest bundle under Webpack 5 is smaller by about 45% (6.75 MB vs 10.39 MB).

### Webpack 4


### Webpack 5

wiedld:
> Wepback 5's bundle size is smaller by about 30% (27.39 MB vs 38.7 MB). The largest bundle under Webpack 5 is smaller by about 45% (6.75 MB vs 10.39 MB).

TY for running the analyzer!

The only part which makes me a bit worried is the `@influxdata/flux-lsp-browser/flux-lsp-browser_bg.wasm` which is shown in the upper left webpack4 bundle. That's the LSP used in both the LSP server worker, and on the main thread as a utils library for parsing variables from editor text.

There is a big change in the wasm size:

BEFORE (master branch pulled today, prod build)
Screen Shot 2022-12-02 at 6 16 36 PM

AFTER (latest on this branch, prod build)
Screen Shot 2022-12-02 at 6 13 40 PM

Not sure what's up with the drastic difference in wasm size. Sanity checked by also running the analyzer on the dev build for this branch (which we know is passing CI tests) -- and it also shows this giant reduction is wasm size. So...is ok? 🤷🏼

wiedld:
> (Since it's now in bytes, I'm guessing it's no longer including the full wasm in the analysis.)

yeah. Inspecting a wasm bundle gives a different size. Says 12.37 KB upon inspection. So I think the analyzer is failing on the wasm, under webpack 5.

package.json
hoorayimhelping: @wdoconnell @wiedld copied from this comment: https://github.com/influxdata/ui/pull/6319/files#r1037614697

it turns out that our jest tests were running out of memory. The general way the internet recommends fixing it is to set `maxWorkers` <= the number of CPUs your circle instance has and go from there. Since setting `maxWorkers` to 2 was causing problems, I dropped the number down to `1`, which is an alias for `runInBand`.

I talked with Gene about this. This isn't necessarily the best way to do it - there are tradeoffs. Lowering the number of CPUs will make the tests run for longer which will cost more money, but we'll still be on a relatively cheap CPU plan. The alternative is to increase the number of CPUs for this circleci job which will also cost more money. If this ends up taking too long or being too expensive, we can increase the resources available to our circle instance.

hoorayimhelping:
@wiedld I couldn't find any information on the why our wasm size is different now. I did change the loader for `.wasm` files to use the new asset pipeline instead of `file-loader`, [here](https://github.com/influxdata/ui/pull/6319/files#diff-f9fd5ce5245988264dcc04849d2b7cac87426cd9b69031df477fac881fedce90R75) but I'm not sure why that would change the bundle size, since it's the loader side of things.

src/flows/index.ts
```diff
@@ -27,8 +27,8 @@ const context = require.context(
   true,
   /^\.\/([^\/])+\/index\.(ts|tsx)$/
 )
-context.keys().forEach(key => {
-  const module = context(key)
+context.keys().forEach(async key => {
+  const module = await context(key)
   module.default((definition: TypeRegistration) => {
```
wdoconnell, line 32: What was the need for this change driven by?

src/flows/pipes/Notification/endpoints.ts
```diff
@@ -30,8 +30,8 @@ const context = require.context(
   true,
   /^\.\/([^\/])+\/index\.(ts|tsx)$/
 )
-context.keys().forEach(key => {
-  const module = context(key)
+context.keys().forEach(async key => {
+  const module = await context(key)
   module.default((definition: EndpointTypeRegistration) => {
```
wdoconnell, line 35: Same question here.

webpack.common.ts
```diff
@@ -37,18 +48,14 @@ module.exports = {
...
-    fs: 'empty',
     global: true,
-    crypto: 'empty',
-    tls: 'empty',
-    net: 'empty',
-    process: true,
-    module: false,
-    clearImmediate: false,
-    setImmediate: true,
   },
```
wdoconnell, line 59: What's the rationale for removing these properties from the node object?

wdoconnell (commented):
LGTM, but added a couple of questions.

src/flows/index.ts
hoorayimhelping: @wdoconnell there's no documentation around this, but it looks like webpack 5 changed how [`require.context`](https://webpack.js.org/guides/dependency-management/) works so that it's asynchronous.

More snarkily: the change was driven by the app not running and pointing to these lines with an error that said something to the effect of `default` doesn't exist on a promise.

webpack.common.ts
hoorayimhelping: @wdoconnell these existing failed the build because the `node` object no longer accepts these keys except for `Buffer`, `process`, (which is false by default), and `global`. Webpack 5 changed how the default polyfills it includes work. This may be helpful: https://gist.github.com/ef4/d2cf5672a93cf241fd47c020b9b3066a

src/flows/index.ts
wdoconnell: 👍 🚀

wdoconnell (dismissed):
Looks good. Thanks for the responses on the webpack config. 👍  🚀

wiedld (dismissed):
Thanks for tackling this challenge.

abshierjoel (approved):

PR #6324, opened 2022-11-17
feat(Identity): use account allowances state to allow org creation

### LINKED ISSUE #6224: Adjust 'Create New Organization' button to use /allowances/orgs/create API once ready
The visibility of the 'create new organization button' in the global header, which is behind the `createDeleteOrgs` flag, turns on account type, pending completion of a backend API.

Once the /allowances/orgs/create endpoint is ready, that endpoint rather than the account type should be used to determine whether the user can create organizations, and whether the Upgrade button should be shown.

See quartz PR [#6726](https://github.com/influxdata/quartz/issues/6726)

blegesse-git:
Closes #6224 

Pr adds org creation allowance property to the Identity state in redux and uses `state.identity.allowances.orgCreation` to determine whether the user can create organizations.
Pr also refactors code to replace use of `accountType` with  `state.identity.allowances.orgCreation` to determine whether the Upgrade button should be shown in the global header. 

behind `createDeleteOrgs` feature flag
https://user-images.githubusercontent.com/66275100/202481864-032d0b98-a36f-44fe-9485-eb45ec23821d.mov

### Checklist

### DIFF
--- a/src/identity/allowances/actions/creators/index.ts
+++ b/src/identity/allowances/actions/creators/index.ts
@@ -1,0 +1,21 @@
+// Types
+import {RemoteDataState} from 'src/types'
+
+export const SET_ALLOWANCES = 'SET_ALLOWANCES'
+export const SET_ALLOWANCES_LOADING_STATUS = 'SET_ALLOWANCES_LOADING_STATUS'
+
+export type Actions =
+  | ReturnType<typeof setAllowances>
+  | ReturnType<typeof setAllowancesLoadingStatus>
+
+export const setAllowances = allowances =>
+  ({
+    type: SET_ALLOWANCES,
+    allowances: allowances,
+  } as const)
+
+export const setAllowancesLoadingStatus = (loadingStatus: RemoteDataState) =>
+  ({
+    type: SET_ALLOWANCES_LOADING_STATUS,
+    loadingStatus: loadingStatus,
+  } as const)
--- a/src/identity/allowances/actions/creators/index.ts
+++ b/src/identity/allowances/actions/creators/index.ts
@@ -1,21 +1,24 @@
 // Types
 import {RemoteDataState} from 'src/types'
 
-export const SET_ALLOWANCES = 'SET_ALLOWANCES'
-export const SET_ALLOWANCES_LOADING_STATUS = 'SET_ALLOWANCES_LOADING_STATUS'
+export const SET_ORG_CREATION_ALLOWANCES = 'SET_ORG_CREATION_ALLOWANCES'
+export const SET_ORG_CREATION_ALLOWANCES_STATUS =
+  'SET_ORG_CREATION_ALLOWANCES_STATUS'
 
 export type Actions =
-  | ReturnType<typeof setAllowances>
-  | ReturnType<typeof setAllowancesLoadingStatus>
+  | ReturnType<typeof setOrgCreationAllowances>
+  | ReturnType<typeof setOrgCreationAllowancesStatus>
 
-export const setAllowances = allowances =>
+export const setOrgCreationAllowances = allowances =>
   ({
-    type: SET_ALLOWANCES,
+    type: SET_ORG_CREATION_ALLOWANCES,
     allowances: allowances,
   } as const)
 
-export const setAllowancesLoadingStatus = (loadingStatus: RemoteDataState) =>
+export const setOrgCreationAllowancesStatus = (
+  loadingStatus: RemoteDataState
+) =>
   ({
-    type: SET_ALLOWANCES_LOADING_STATUS,
+    type: SET_ORG_CREATION_ALLOWANCES_STATUS,
     loadingStatus: loadingStatus,
   } as const)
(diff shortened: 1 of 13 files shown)


src/identity/allowances/actions/creators/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const SET_ORG_CREATION_ALLOWANCES = 'SET_ORG_CREATION_ALLOWANCES'
+export const SET_ORG_CREATION_ALLOWANCES_STATUS =
+  'SET_ORG_CREATION_ALLOWANCES_STATUS'
+
+export type Actions =
+  | ReturnType<typeof setOrgCreationAllowances>
+  | ReturnType<typeof setOrgCreationAllowancesStatus>
+
+export const setOrgCreationAllowances = allowances =>
```
wdoconnell, line 12: Can we add typing for the allowance here?  Should be able to import from unityRoutes or the UI API layer.

src/identity/allowances/actions/thunks/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    try {
+      dispatch(setOrgCreationAllowancesStatus(RemoteDataState.Loading))
+
+      const allowances = await fetchOrgCreationAllowance()
+
+      dispatch(setOrgCreationAllowances(allowances))
+
+      dispatch(setOrgCreationAllowancesStatus(RemoteDataState.Done))
+    } catch (error) {
+      reportErrorThroughHoneyBadger(error, {
```
wdoconnell, line 33: Once we hear back from AIM, let's see if we can report only errors that aren't a `401` status code (we can skip error messages that are an artifact of the user being logged out).

src/identity/apis/org.ts
```diff
@@ -132,17 +138,18 @@ export const fetchDefaultAccountDefaultOrg = async (): Promise<
...
+export const fetchOrgCreationAllowance =
+  async (): Promise<OrgCreationAllowances> => {
+    const response = await getAllowancesOrgsCreate({})
 
-  if (response.status !== 200) {
-    throw new GenericError(
-      'Failed to determine whether this user can create a new organization.'
-    )
-  }
+    if (response.status !== 200) {
```
wdoconnell, line 145: Do you know what's going on with the linter here? I don't see a new code block but it's indenting it more.

src/identity/selectors/index.ts
```diff
@@ -64,3 +64,13 @@ export const selectQuartzOrgsStatus = (
 ): AppState['identity']['quartzOrganizations']['status'] => {
   return state.identity.quartzOrganizations.status
 }
+
+export const selectOrgCreationAllowances = (
+  state: AppState
+): AppState['identity']['allowances']['orgCreation'] => {
+  return state.identity.allowances.orgCreation
+}
```
wdoconnell, line 72: I might consider having separate selectors for the two properties. I know that sounds duplicative, but it may help with re-renders not to have the selected property be the entire object.

wdoconnell (changes requested):
Looking great!  Couple a minor edits here.

src/identity/allowances/actions/thunks/index.ts
blegesse-git: yes, I didn't forget about our discussion on filtering out 401s from being sent to honeybadger. are they working on adding 401 status to the api response?

src/identity/apis/org.ts
blegesse-git: line 142 `: Promise<OrgCreationAllowances>` is the only addition here. The indentations are most likely caused by prettier

src/identity/components/GlobalHeader/OrgDropdown.tsx
```diff
@@ -1,6 +1,7 @@
 // Libraries
-import React, {FC} from 'react'
-import {IconFont} from '@influxdata/clockface'
+import React, {FC, useEffect} from 'react'
```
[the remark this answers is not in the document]
blegesse-git: good catch. I'll move imports from clockface down to the _Components_ section

src/identity/selectors/index.ts
blegesse-git: agreed. Since `loadingStatus` is listed as a property of the `orgCreation` object, every time it's value is updated from `NotStarted`-> `Loading`-> to `Done`, it will force the component to re-render. I didn't take that into consideration.

src/identity/allowances/actions/thunks/index.ts
```diff
@@ -0,0 +1,38 @@
...
+        name: 'Failed to fetch org creation Allowances',
+        context: {state: getState()},
```
wdoconnell, line 35: Also - this is minor, but I don't think we need all of state here (when I've viewed that on honeybadger it makes it hard to debug). I would just grab so much of state as we need - probably the information from the `identity` object.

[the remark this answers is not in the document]
blegesse-git: no problem! i'll fix that

wdoconnell (approved):
Thanks for quick work on the comments - looks good to me. 👍

PR #6335, opened 2022-11-21
feat: on creating last org in account, user gets an upgrade notification

### LINKED ISSUE #6267: feat: display organization quota limit reached notification after creating org
After you create an organization using the new modal, we need to check whether the quota limit has been reached or not yet. If the organization limit hasn't reached then just let the org creation flow work normally. If the quota limit has reached, then we need to display a notification to show the same. See images below. This notification needs to be dismissable manually and shouldn't just go away automatically. We do not need to persist the dismissed status in settings or anywhere. If the page is refreshed after seeing this notification, then we don't need to display it again.

## Notification on Page
image

## Notification
image

To check whether we have reached quota limit, please use `fetchOrgCreationAllowance`.

wdoconnell:
Closes #6267 

This PR adds a notification and associated business logic so that, after a user creates a new org, the allowances endpoint is re-queried. If the response indicates no more orgs can be created in the account, it sends the user a notification indicating that they have hit their organization quota. Design is per [this figma].(https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1961%3A18041&t=5XorbJgkM9q43DnN-0).

https://user-images.githubusercontent.com/91283923/203148438-4e562358-7126-44de-8845-b8267eab80ed.mov

### Checklist

### DIFF
--- a/src/identity/allowances/actions/thunks/index.ts
+++ b/src/identity/allowances/actions/thunks/index.ts
@@ -31,10 +31,13 @@ export const getOrgCreationAllowancesThunk =
       dispatch(setOrgCreationAllowance(allowances))
 
       dispatch(setOrgCreationAllowanceStatus(RemoteDataState.Done))
+
+      return allowances.allowed
     } catch (error) {
       reportErrorThroughHoneyBadger(error, {
         name: 'Failed to fetch org creation allowance',
         context: {state: getState().identity},
       })
+      return false
     }
   }
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrgInput.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrgInput.tsx
@@ -1,4 +1,4 @@
-import React, {FC, useState} from 'react'
+import React, {FC} from 'react'
 import {
   ComponentSize,
   ComponentStatus,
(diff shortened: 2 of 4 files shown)


src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrgInput.tsx
```diff
@@ -25,12 +25,7 @@ export const CreateOrgInput: FC<Props> = ({
   orgName,
   validationMsg,
 }) => {
-  const [userTouchedFormInput, setUserTouchedFormInput] = useState(false)
-
   const handleInputChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
-    if (!userTouchedFormInput) {
-      setUserTouchedFormInput(true)
-    }
     handleValidateOrgName(evt.target.value)
```
wdoconnell, line 29: This state is a relic of a prior refactor and is no longer required.

src/identity/allowances/actions/thunks/index.ts
```diff
@@ -31,10 +31,13 @@ export const getOrgCreationAllowancesThunk =
...
       dispatch(setOrgCreationAllowanceStatus(RemoteDataState.Done))
+
+      return allowances.allowed
     } catch (error) {
       reportErrorThroughHoneyBadger(error, {
         name: 'Failed to fetch org creation allowance',
         context: {state: getState().identity},
       })
+      return false
     }
```
wdoconnell, line 42: Return values are necessary to make this invocation of `dispatch` chainable.

abshierjoel (approved):

PR #6337, opened 2022-11-21
fix: only one /accounts call on app load

### LINKED ISSUE #6336: App makes three calls to /accounts endpoint on initial load
On initial load, the app makes three calls to the /accounts endpoint (not one).

## About the bug

**Steps to reproduce:**

1. Log in
2. Open network tab
3. Filter for calls to /accounts

**Expected behavior:**
One call to /accounts endpoint on initial load.

**Actual behavior:**
Three calls to /accounts endpoint on initial load.

**Visual Proof:**

https://user-images.githubusercontent.com/91283923/203166891-277e43bb-4497-4333-9637-41e072f2d62d.mov

wdoconnell:
Closes #6336 

On initial load, the app hits the /accounts endpoint three times. This PR reduces it to one call.

Root cause is a `useEffect` hook that has been hanging out in the user account context. The hook calls a function (`handleGetAccounts`) that fetches from `/accounts` on initial app load, or when there is a change in the value of the user's `defaultAccountId` or `activeAccountId`. Since `handleGetAccounts` itself sets the `defaultAccountId` and `activeAccountId` based on the array the endpoint returns, we end up with three API calls: (1) when the context loads; (2) when the context sets the default account id based on the response; (3) when the context sets the active account id based on the response.

There's no need for the `defaultAccountId` dependency because the user's new default org can be changed cheaply without the need for any api call by simply setting the value of the accounts array to a clone of the accounts array with the new default org setting. There's no need for the `activeAccountId` dependency for that reason, but also because the `activeAccount` never changes until the app reloads (active account changes are handled by quartz redirect).

Ultimately (and we have tickets for this), it'd be cleaner for Redux to be used here, not context. But for now, this will reduce loadtimes and number of calls to quartz without the need for a broader refactor of UI code, given more pressing needs.

Before
--

https://user-images.githubusercontent.com/91283923/203167318-30b0e473-faab-4495-92f7-b55147499203.mov

After
--

https://user-images.githubusercontent.com/91283923/203167325-dbdd16c3-6a1b-4d9f-b767-96579d7c8f01.mov

### Checklist

### DIFF
--- a/src/identity/components/userprofile/UserDefaults.tsx
+++ b/src/identity/components/userprofile/UserDefaults.tsx
@@ -44,7 +44,8 @@ import 'src/identity/components/userprofile/UserProfile.scss'
 export const UserDefaults: FC = () => {
   const dispatch = useDispatch()
 
-  const {userAccounts, handleSetDefaultAccount} = useContext(UserAccountContext)
+  const {defaultAccountId, userAccounts, handleSetDefaultAccount} =
+    useContext(UserAccountContext)
   const quartzOrganizations = useSelector(selectQuartzOrgs)
 
   const identity = useSelector(selectQuartzIdentity)
@@ -58,8 +59,10 @@ export const UserDefaults: FC = () => {
 
   const defaultAccount = useMemo(
     () =>
-      accounts ? accounts.find(el => el.isDefault === true) : emptyAccount,
-    [accounts]
+      accounts
+        ? accounts.find(acct => acct.id === defaultAccountId)
+        : emptyAccount,
+    [accounts, defaultAccountId]
   )
   const defaultOrg = useMemo(
     () => (orgs ? orgs.find(el => el.isDefault === true) : emptyOrg),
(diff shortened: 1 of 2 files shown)


src/identity/components/userprofile/UserDefaults.tsx
```diff
@@ -58,8 +59,10 @@ export const UserDefaults: FC = () => {
 
   const defaultAccount = useMemo(
     () =>
-      accounts ? accounts.find(el => el.isDefault === true) : emptyAccount,
-    [accounts]
+      accounts
+        ? accounts.find(acct => acct.id === defaultAccountId)
+        : emptyAccount,
+    [accounts, defaultAccountId]
   )
```
wdoconnell, line 66: Before, this component didn't need to check for updates in the `defaultAccountId` loaded from context because the context forced a new api call to `/accounts`. Now, no call to `/accounts` is needed here either - we can simply retrieve the accounts array from context when the default account id changes.

alejom99 (approved):

PR #6342, opened 2022-11-22
chore: make org list tab use allowances state

### LINKED ISSUE #6338: Make org list page use new allowance thunk
Organization creation allowances are fetched once and stored in redux state, since they're now used in three locations in the app. Make the org list page rely on the redux state using the new allowance selectors.

wdoconnell:
Closes #6338 

Multiple components in the UI now require access to the data returned from the /orgs/allowances/create endpoint. Since this data is now fetched on application load and stored in redux state, this PR makes the organization list tab rely on the state stored in redux so that no new API call is required.

In the edge case where the data isn't loaded in redux state yet, a thunk is dispatched.

Before
--

https://user-images.githubusercontent.com/91283923/203386402-be7ee284-7515-48ed-ac55-8fdff9349788.mov

After
--

https://user-images.githubusercontent.com/91283923/203386407-dc6f40e9-6fde-482c-8d15-30421bba21a8.mov

### Checklist

### DIFF
--- a/src/cloud/notifications/account-settings-notifications.ts
+++ b/src/cloud/notifications/account-settings-notifications.ts
@@ -1,7 +1,0 @@
-import {Notification} from 'src/types'
-import {defaultErrorNotification} from 'src/shared/copy/notifications'
-
-export const fetchOrgAllowanceFailed = (): Notification => ({
-  ...defaultErrorNotification,
-  message: `Cannot determine whether more organizations can be created in this account.`,
-})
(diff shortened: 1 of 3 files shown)


src/cloud/notifications/account-settings-notifications.ts
```diff
@@ -1,7 +0,0 @@
-import {Notification} from 'src/types'
-import {defaultErrorNotification} from 'src/shared/copy/notifications'
-
```
wdoconnell, line 3: This toast notification isn't needed anymore. Dispatching a notification to the user was less desirable (since the user can't do anything with a bad /orgs/allowances/create response), and we now have separate error handling (honeybadger) in the thunk.

hansenreid (approved):
🔥

PR #6343, opened 2022-11-22
chore: filter inactive orgs from global header and user profile

### LINKED ISSUE #5935: Multi Org Phase 2 - Filter deactivated organizations from organization dropdown
AIM will modify the /quartz/accounts/:accountID/orgs endpoint that returns the orgs in the user's current account to also return orgs currently in the process of being deleted, with a property that identifies the deletion state.

Current logic in global header lists _all_ orgs returned from this endpoint when rendering the list of orgs in the user's current account. This will need to be adjusted when the endpoint changes so that the header doesn't display - and the user can't try to switch to - a deleted org.

wdoconnell:
Closes #5935 

The list of all orgs in an account now includes a `state` property to indicate whether each org is `provisioned`, `pending`, or `suspended`. 
Screen Shot 2022-11-28 at 10 49 59 AM

This PR adds a new selector for provisioned orgs, so that only those organizations are shown in the global app header, and on the profile page for changing default orgs. 

Edit: added unit tests to ensure the `selectQuartzActiveOrg` selector is working as expected.
Screen Shot 2022-11-28 at 6 28 22 PM

Unit tests may not be used by CI until [this issue](https://github.com/influxdata/ui/issues/6346) is resolved.

Before
--

https://user-images.githubusercontent.com/91283923/204320736-f96d2e43-d479-47db-bf78-f35bb04c99fc.mov

After
--

https://user-images.githubusercontent.com/91283923/204321084-464266e1-7450-4342-bba9-b69cd0796cad.mov

Quartz-Remocal Testing
--

https://user-images.githubusercontent.com/91283923/204565529-7144eca5-bdd5-42df-811c-cb05fda3037a.mov

### Checklist

### DIFF
--- a/src/identity/components/userprofile/UserDefaults.tsx
+++ b/src/identity/components/userprofile/UserDefaults.tsx
@@ -12,7 +12,10 @@ import {
 
 // Selectors and Context
 import {UserAccountContext} from 'src/accounts/context/userAccount'
-import {selectQuartzIdentity, selectQuartzOrgs} from 'src/identity/selectors'
+import {
+  selectQuartzActiveOrgs,
+  selectQuartzIdentity,
+} from 'src/identity/selectors'
 
 // Thunks
 import {updateDefaultOrgThunk} from 'src/identity/quartzOrganizations/actions/thunks'
@@ -46,7 +49,7 @@ export const UserDefaults: FC = () => {
 
   const {defaultAccountId, handleSetDefaultAccount, userAccounts} =
     useContext(UserAccountContext)
-  const quartzOrganizations = useSelector(selectQuartzOrgs)
+  const orgs = useSelector(selectQuartzActiveOrgs)
 
   const identity = useSelector(selectQuartzIdentity)
   const loggedInAccount = identity.currentIdentity.account
@@ -54,8 +57,7 @@ export const UserDefaults: FC = () => {
   const accounts = userAccounts
   const numAccounts = userAccounts ? userAccounts.length : 0
 
-  const orgs = quartzOrganizations.orgs
-  const numOrgs = quartzOrganizations.orgs ? quartzOrganizations.orgs.length : 0
+  const numOrgs = orgs ? orgs.length : 0
 
   const defaultAccount = useMemo(
     () =>
@@ -74,11 +76,11 @@ export const UserDefaults: FC = () => {
 
   useEffect(() => {
     setSelectedAccount(defaultAccount)
-  }, [accounts, defaultAccount])
+  }, [defaultAccount])
 
   useEffect(() => {
     setSelectedOrg(defaultOrg)
-  }, [orgs, defaultOrg])
+  }, [defaultOrg])
 
   const userPickedNewAccount =
     defaultAccount?.id !== selectedAccount?.id && selectedAccount !== null
@@ -141,7 +143,6 @@ export const UserDefaults: FC = () => {
             setSelectedAccount={setSelectedAccount}
           />
         )}
-
         {numOrgs > 1 && (
           <DefaultOrgForm
             accounts={accounts}
--- a/src/identity/selectors/index.test.ts
+++ b/src/identity/selectors/index.test.ts
@@ -14,7 +14,10 @@ import {cloneDeep} from 'lodash'
 
 const mockState = {
   identity: {
-    quartzOrganizations: {orgs: [emptyOrg], status: RemoteDataState.Done},
+    quartzOrganizations: {
+      orgs: cloneDeep(mockOrgData),
+      status: RemoteDataState.Done,
+    },
   },
 } as AppState
 
@@ -23,7 +26,6 @@ describe('selectQuartzActiveOrgs selector', () => {
 
   beforeEach(() => {
     localState = cloneDeep(mockState)
-    localState.identity.quartzOrganizations.orgs = cloneDeep(mockOrgData)
   })
 
   it('returns an array of length one with the default empty org', () => {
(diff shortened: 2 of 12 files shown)


src/identity/components/userprofile/UserDefaults.tsx
```diff
@@ -46,16 +49,15 @@ export const UserDefaults: FC = () => {
 
   const {defaultAccountId, handleSetDefaultAccount, userAccounts} =
     useContext(UserAccountContext)
-  const quartzOrganizations = useSelector(selectQuartzOrgs)
+  const orgs = useSelector(selectQuartzActiveOrgs)
```
wdoconnell, line 53: Users should only be able to set 'active' orgs as their default org in the current account.

src/identity/components/userprofile/UserDefaults.tsx
```diff
@@ -74,11 +76,11 @@ export const UserDefaults: FC = () => {
 
   useEffect(() => {
     setSelectedAccount(defaultAccount)
-  }, [accounts, defaultAccount])
+  }, [defaultAccount])
 
   useEffect(() => {
     setSelectedOrg(defaultOrg)
-  }, [orgs, defaultOrg])
+  }, [defaultOrg])
```
wdoconnell, line 84: This dependency is not necessary (and causes the selected dropdown item to toggle back to the default if there is an error and a re-render).

src/identity/selectors/index.ts
```diff
@@ -52,6 +48,18 @@ export const selectOrgCreationAvailableUpgrade = (
   return state.identity.allowances.orgCreation.availableUpgrade
 }
 
+export const selectQuartzActiveOrgs = (
+  state: AppState
+): AppState['identity']['quartzOrganizations']['orgs'] => {
+  return state.identity.quartzOrganizations.orgs.filter(
+    org => org.state !== 'suspended' && org.state !== 'pending'
+  )
+}
```
wdoconnell, line 57: Orgs can only be `active`, `suspended`, or `pending`, so this selector could be written `org ==> org.state === 'active'`. 

The negative was adopted here to ensure that if more states are added for orgs that _should_ be shown to the user, we're being overinclusive rather than underinclusive.

hoorayimhelping: So if we add another state that is supposed to restrict the account, it might get through this check? Like if we add `delinquent` and forget to update this selector, then deliquent accounts will show up in the header?

hoorayimhelping (approved):
I had question about adding another state that should be restricted.

I also think calling the account standing `state` is a really vague term and might end up being a mistake. `state` is so vague it can apply to anything, and it's an overloaded term that might mean `entire state tree`, `entire application state` or `status of something` (which it means in this case). I think adding a descriptor is a good idea going forward, even if it's just on the frontend side rather than the api side. An example might be something like `accountStanding` or `approvalStatus`

src/identity/selectors/index.ts
wdoconnell: Correct. On balance, let's switch this back to the positive - `org => org.state === provisioned`. openAPI will identify any changes in the options for org state, and you're right that we could just as easily want to add a state that restricts access here.

wdoconnell:
> I had question about adding another state that should be restricted.
> 
> I also think calling the account standing `state` is a really vague term and might end up being a mistake. `state` is so vague it can apply to anything, and it's an overloaded term that might mean `entire state tree`, `entire application state` or `status of something` (which it means in this case). I think adding a descriptor is a good idea going forward, even if it's just on the frontend side rather than the api side. An example might be something like `accountStanding` or `approvalStatus`

I'm a little leery of doing this on the frontend only. I think `provisioned`, `suspended`, and `pending` as values makes it clear what state we're referring to (it's the 'state' within the object literal for a specific org, so I don't think we'd confuse that with the application state or entire state tree). If we change the name on the frontend, we'd also be mapping through the response from quartz after the API call and changing none of the properties of any objects in the array _except_ for the state key, which seems unnecessary and might cause some confusion.

abshierjoel (approved):
Looks great! 🚀

hoorayimhelping:
> think `provisioned`, `suspended`, and `pending` as values makes it clear what state we're referring to (it's the 'state' within the object literal for a specific org, so I don't think we'd confuse that with the application state or entire state tree).

Just saw this comment. Yes, those values, when they're present (and their being present is the key) make it pretty clear what `state` means. The issue I'm thinking of is what `state` means when inspecting the structure of the redux state without any of those values present (like say looking at the code without the application running). The term `state` gives no clue about what the value it holds represents, without actually seeing that value. It's been my experience that this makes reasoning about the application and the code much harder if you're not running the code

src/identity/selectors/index.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    localState.identity.quartzOrganizations.orgs[1].state = 'pending'
+
+    const activeOrgs = selectQuartzActiveOrgs(localState)
+
+    expect(activeOrgs.length).toBe(3)
+    expect(activeOrgs.map(org => org.name)).toEqual(
+      expect.arrayContaining(['Test GmbH 3', 'Test Inc. 4', 'Test SA 5'])
+    )
+  })
+})
```
hoorayimhelping, line 74: Tests!

hoorayimhelping (approved):
I still think `state` could be named better. Whether you change it here or not, consider this in the future.

wdoconnell:
> > think `provisioned`, `suspended`, and `pending` as values makes it clear what state we're referring to (it's the 'state' within the object literal for a specific org, so I don't think we'd confuse that with the application state or entire state tree).
> 
> Just saw this comment. Yes, those values, when they're present (and their being present is the key) make it pretty clear what `state` means. The issue I'm thinking of is what `state` means when inspecting the structure of the redux state without any of those values present (like say looking at the code without the application running). The term `state` gives no clue about what the value it holds represents, without actually seeing that value. It's been my experience that this makes reasoning about the application and the code much harder if you're not running the code. To me this seems to create a coupling between the variable name and what the system fills it in, which is a bit unnecessary in my opinion.

@hoorayimhelping How about about `provisioningStatus` as a name? Ultimately the intent of the property is to validate whether the org is currently provisioned, whether it's not yet provisioned and therefore unusable (pending) or whether it _was_ provisioned but has since been deactivated. I wouldn't say `accountStanding` because the status of one of the orgs generally doesn't have implications for the account.

hoorayimhelping:
>@hoorayimhelping How about about provisioningStatus as a name?

@wdoconnell I think that indicates pretty clearly what the variable represents!

hoorayimhelping (approved):

Issue #6334, opened 2022-11-21
EPIC: Flux composition, integration of yieldless


wiedld:
After the `|> yield(name: "_editor_composition")` was removed as part of [this work](https://github.com/influxdata/flux-lsp/issues/583), we need to handle integration with the UI. This includes, but is not limited to, the follow use cases:
* getting the basic composition working
* With removal of the yield...how can we visually indicate to the user, where is the current composition block being edited?
* how handle flux "sync" and "divergence"?
     * is the concept of "divergence detected by the UI" still hold?
* how often does the LSP fail to update the composition block?
     *  if frequent, how handle these errors? 

## Implementation:
This is the current implementation, in progress:
- [x] FIRST TASK: determine if the executeCommands sent the LSP, results in the predicted output code in the UI.
     * https://github.com/influxdata/flux-lsp/issues/615
     * https://github.com/influxdata/ui/issues/6341
- [x] SECOND TASK: which server-side events (LSP spec) to use? how get UI listening and reacting to these events?
     * confirmed. Found a way. PoC is [this commit here](https://github.com/influxdata/ui/commit/c6597932850b39cf3244d0fae9453ec988763e52), with a detailed commit message on all approaches tried.
     * Server-initiated showMessage ActionItems, intercepted in the UI:
         *  ![Screen Shot 2022-12-01 at 6 30 12 PM](https://user-images.githubusercontent.com/10232835/205201448-b7eb0e6a-133c-4582-93f3-e9a4e33ee0f8.png)

- [x] THIRD TASK: using the LSP triggered events, do the following:
     * provide UX to indicate location of the composition code block
     * give UX to state when composition is dropped.
          * can we still re-init a new composition, after the last is dropped? 
     * how often does the LSP fail to update the composition block?
          *  if frequent, how handle these errors?

wiedld:
FIRST TASK: outcome.
* edits basic block
* preserves user changes to the `|> range()` and transformations added after the block.
* when selecting new measurement --> is the same block.
* when selecting new bucket --> inits a new composition.

https://user-images.githubusercontent.com/10232835/203623726-24a17c23-7600-436b-bccd-8e8e1c72e1f9.mov


Only thing not resolved; is part of THIRD TASK:
* sometimes there is ambiguity with the current composition block, and another subquery (a.k.a. a previous composition).
  * when this occurs, the composition is dropped in the server.
  * once that occurs --> the LSP server stops responding.
[...]

PR #6344, opened 2022-11-23
feat(6341): get basic composition working with yieldless

### LINKED ISSUE #6341: Composition: update executeCommand to work with yieldless paradigm.
In a previous debugging effort, we [jointly made a decision](https://github.com/influxdata/ui/issues/5622) to have the composition block always re-initialize fully on each schema change. This required changes in [the UI](https://github.com/influxdata/ui/pull/5630) and [the LSP](https://github.com/influxdata/flux-lsp/pull/575) to make this functional.

Under the new yieldless paradigm, each change in composition in made incrementally. Which means finding the block again, and making the incremental change. Therefore, we cannot be re-init'ing on each change.

Additionally, the new LSP implementation assumes a new composition is being created which each new bucket selected. We need to carry that decision into the UI, and only create a new composition upon bucket selection. This decision means that the measurement selection cannot be another `executeCommand/initialize`. (Or else the composition block gets doubled.) Therefore, we need to have a `executeCommand/setMeasurement` or other equivalent command.

.

## Tasks to change in the UI:
- [x] don't re-init the entire composition block, on each change. Do incremental executeCommands instead.
- [x] use `executeCommand/setMeasurement`

.

## Current Bug/Result in the UI:

The result is this behavior in the UI, when everything uses the `executeCommand/initialize`.

#### Build a composition
https://user-images.githubusercontent.com/10232835/203146234-3f059644-55cb-402a-afba-0b26a8cb1c29.mov

#### old and new block, always occurs
https://user-images.githubusercontent.com/10232835/203146251-7f7792e2-c83a-434f-85d5-0c948a70ea39.mov

wiedld:
Part of https://github.com/influxdata/ui/issues/6334
Closes https://github.com/influxdata/ui/issues/6341

Note: this is part of a chain of PRs.
`staging/yieldless-composition`  <-- 1st PR <-- 2nd PR <-- 3rd PR
The final staging branch will be the merge into master.

#### PR Chain:
1. this PR.
2. https://github.com/influxdata/ui/pull/6374
3. https://github.com/influxdata/ui/pull/6397 where tests finally pass!

## DONE:

Closes #6341  (`FIRST TASK` on the epic, UI side changes).
That^^ issue ticket describes in detail, on why these changes were required.

## Visual Evidence:
Video of current status after UI and flux-lsp `FIRST TASK` changes:

https://user-images.githubusercontent.com/10232835/203626079-946b8ed6-2c4a-47a3-8106-8384306d265b.mov

## Checklist
- [x] Feature flagged, if applicable. Is behind `schemaComposition` flag, which I believe is on in prod.

### DIFF
--- a/src/languageSupport/languages/flux/lsp/connection.ts
+++ b/src/languageSupport/languages/flux/lsp/connection.ts
@@ -13,7 +13,6 @@ import {
   DEFAULT_FLUX_EDITOR_TEXT,
   CompositionSelection,
 } from 'src/dataExplorer/context/persistance'
-import {CompositionInitParams} from 'src/languageSupport/languages/flux/lsp/utils'
 import {comments} from 'src/languageSupport/languages/flux/monaco.flux.hotkeys'
 
 // LSP methods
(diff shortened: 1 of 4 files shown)


src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -480,18 +389,15 @@ class LspConnectionManager {
...
+      return // do not re-init schema session on reload
     }
 
     const {toAdd, toRemove} = this._diffSchemaChange(schema, previousState)
     if (Object.keys(toAdd).length || Object.keys(toRemove).length) {
-      this._updateLsp(toAdd, toRemove)
+      // since this._diffSchemaChange() can set the model
+      // we need the executeCommand to be issued after the model update
+      const delay = toAdd.bucket ? 1500 : 0
+      setTimeout(() => this._updateLsp(toAdd, toRemove), delay)
```
nathanielc, line 400: Can you explain why we need these timeout delays? I was not able to figure it out by following the chain of issues. Maybe a call would be good for this?

wiedld: 1. When we start a brand new script, there is [default text populated](https://github.com/influxdata/ui/blob/74b2910b4c94fd4886446732893dc4a11ba79552/src/dataExplorer/components/resources/types/script/index.ts#L26) in the editor.
2. When inserting a new composition for the first time:
   * if there is only the default text present, t[hen remove the default text](https://github.com/influxdata/ui/blob/9c4fb1fc5900cc9c898868b6cd15b75a3f4ba58f/src/languageSupport/languages/flux/lsp/connection.ts#L323-L326).
3. Since the resetting of the model text is indirect (a.k.a. `this._model.setValue('')`), we do introduce a race condition.
    * `this._model.setValue('')` --> monaco-editor updates --> monaco-editor syncs with LSP by sending over the new File.
    * `this._updateLsp()` --> we send to the LSP an `executeCommand` --> LSP makes update using it's current copy of File
4. when dev'ing locally, I occasionally saw this race condition come into play. 

I could:
(a) make this delay occur only when we are removing the default text, if you would like. 
(b) Or never remove the default text; in which case it will remain at the end of the newly constructed composition (and the user will need to delete it manually).

wiedld: I feel like the underlying issue is that we communicate with the LSP in 2 ways:
1. the monaco-editor client messages the LSP
2. the UI itself directly injects messages into the communication stream, bypassing the monaco-editor client.

The reason why we are using the approach # 2 above, is because we could not get the monaco-editor to issue the messages we needed. So we introduced the race condition with the bypass.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -276,119 +275,42 @@ class LspConnectionManager {
...
+      this.inject(ExecuteCommand.CompositionInit, {bucket: toAdd.bucket?.name})
+    }
+
+    if (toAdd.measurement) {
+      this.inject(ExecuteCommand.CompositionSetMeasurement, {
+        value: toAdd.measurement,
+      })
     }
 
     if (toRemove.fields?.length) {
```
appletreeisyellow, line 292:
```suggestion
    if (toRemove?.fields?.length) {
```
Should there by a `?` after `toRemove`? Since it is set to `null` by default on line 280. Same for the the rest of the `toRemove`

appletreeisyellow (approved):
I like the simpler version of `_updateLsp()`! I have one question below. Otherwise, looks good to me!

src/languageSupport/languages/flux/lsp/connection.ts
hoorayimhelping: Thanks for this explanation. This is pretty a pretty frustrating set of circumstances 🤷‍♂️

hoorayimhelping (approved):

src/languageSupport/languages/flux/lsp/connection.ts
wiedld: TY! done.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -276,119 +275,42 @@ class LspConnectionManager {
...
-    const msg = this._bufferComposition[this._o % 100]
-    if (!!msg) {
-      this.inject(...msg)
-      this._o++
-    }
-    return
-  }
-
-  _addUpdatesToBuffer(
+  _updateLsp(
```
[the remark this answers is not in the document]
wiedld: I'm showing my age. lol.

That's an older convention, but there is [a newer one enforcable by js](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields). Down for a followup PR to update these...shall we?

PR #6349, opened 2022-11-29
chore: add ability to use Iox org in tests

genehynson:
Allows for testing with iox from cypress tests.

Passing `useIox: true` in `cy.signin` is one way to run a single test against Iox. Setting env var `CYPRESS_useIox=true` will run all the tests against Iox.

This PR does not change existing testing behavior, it just allows for customization in future PRs.

### Checklist

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -29,8 +29,10 @@ Cypress.on('uncaught:exception', (err, _) => {
   )
 })
 
-export const signin = (): Cypress.Chainable<Cypress.Response<any>> => {
-  return cy.setupUser().then((response: any) => {
+export const signin = (
+  useIox: boolean = false
+): Cypress.Chainable<Cypress.Response<any>> => {
+  return cy.setupUser(useIox).then((response: any) => {
     wrapDefaultUser()
       .then(() => wrapDefaultPassword())
       .then(() => {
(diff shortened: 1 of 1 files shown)


cypress/support/commands.ts
```diff
@@ -592,13 +594,21 @@ export const createToken = (
...
   const defaultUser = Cypress.env('defaultUser')
-  const userParam = defaultUser ? `?user=${defaultUser}` : ''
+  let params = ''
+  if (useIox) {
+    params = defaultUser
+      ? `?user=${defaultUser}&orgSuffix=ioxlighthouse`
+      : '?orgSuffix=ioxlighthouse'
+  } else {
+    params = defaultUser ? `?user=${defaultUser}` : ''
+  }
```
hoorayimhelping, line 607: not a huge deal, just a suggestion that using `URLSearchParams` can simplify the logic around needing to figure out if a `?` or `&` is needed for a query param, if you're into that sort of thing:

```ts
const params = new URLSearchParams()
if (defaultUser) {
  params.set('user', defaultUser)
}
if (useIox) {
  params.set('orgSuffix', 'ioxlighthouse')
}

// ...
    url: `/debug/provision?${params.toString()}`
```

hoorayimhelping (approved):

cypress/support/commands.ts
genehynson: i like it

cypress/support/commands.ts
```diff
@@ -592,13 +594,20 @@ export const createToken = (
...
+    params.set('user', defaultUser)
+  }
+  if (useIox) {
+    params.set('orgSuffix', 'ioxlighthouse')
+  }
   return cy
     .request({
       method: 'GET',
-      url: `/debug/provision${userParam}`,
+      url: `/debug/provision${params.toString()}`,
```
hoorayimhelping, line 610: `params.toString()` doesn't include the `?` identifier, so you'll need to add it here (but it should be fine if `params.toString()` is an empty string, since an empty query string with `?` shouldn't hurt anything)

```ts
`/debug/provision?${params.toString()}`
```

genehynson: ah nice catch

hoorayimhelping (approved):

PR #6350, opened 2022-11-29
feat: redirect to orgslist, offer switch-org option on creating new org

### LINKED ISSUE #5205: Multi Org Phase 2 - PAYG & Annual - Create an Organization
**ACs**

- For PAYG and Annual Users, display a "Create Organization" option in the global header organization dropdown. 
- If marketplace user, then only display that user's originally chosen cloud provider in the create org modal
- There should be a "+" sign icon to the left of the "Create Organization" option in the org dropdown
- For Free users, display an "Add More Organizations" upgrade button (since free users cannot create orgs)
- If the user clicks the "Create Organization" option in the org dropdown on the global header, then redirect users to the AIM "create an organization" page and displays the page content inside a modal. Modal should match the design in Figma: https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1967%3A20534      
- If the user tries to create the same org name [repeating an existing org name] within the same account...we should throw an error message. "You already have an org with this name, choose a unique name."
- The first org created in an account should be the user's 'default' org; unless the user manually changed their default org.
- Once the new org is created, redirect the user to the same app page they're currently on, but in their new org. 
- Once the new org is created, display a toast-like message stating that the new org was successfully created. 

**Implementation**

-

wdoconnell:
Closes #5205 

Adds logic to the 'Create New Organization' overlay, so that, on successfully creating a new org, the user is directed to the 'Organization List' page. The displayed popup now gives the user the option to switch to that org (if provisioned).

Creating a New Org in the Account
--

https://user-images.githubusercontent.com/91283923/204661309-bb638204-0695-4740-b497-ebba13146f5b.mov

Creating the Last Available Org in the Org Quota
--

https://user-images.githubusercontent.com/91283923/204661347-b6a2ff6b-297d-4004-9b40-95a361b45728.mov

### Checklist

- [X] Feature flagged, if applicable - YES - `createDeleteOrgs`


src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/SwitchToNewOrgButton.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+const linkStyle = {
+  color: 'black',
+  textDecoration: 'underline',
+  cursor: 'pointer',
+}
+
+export const SwitchToNewOrgButton = (url: string): JSX.Element => {
+  const handleSwitchToNewOrg = () => {
```
hoorayimhelping, line 10: Is this handler necessary? can this be done by putting an `href` on the link? Should the href go on the anchor and an `event.preventDefault()` go here for usability (if an anchor tag has an `href` and the anchor has focus, hitting `enter` will follow the link

hoorayimhelping (approved):
nice work! I'm assuming that `create org` button in the dropdown has logic built in to not be clickable / not display if the account is at its org limit?

wdoconnell:
> nice work! I'm assuming that `create org` button in the dropdown has logic built in to not be clickable / not display if the account is at its org limit?

Before that, actually - the user is not able to open the overlay if the account is at its org limit. Whether the user has access to the overlay depends on whether their org-creation allowance is `true` (the allowance is stored in redux, and it's loaded (a) when the app loads and (b) every time you create a new org).

wdoconnell:
> > nice work! I'm assuming that `create org` button in the dropdown has logic built in to not be clickable / not display if the account is at its org limit?
> 
> Before that, actually - the user is not able to open the overlay if the account is at its org limit. Whether the user has access to the overlay depends on whether their org-creation allowance is `true` (the allowance is stored in redux, and it's loaded (a) when the app loads and (b) every time you create a new org).

Also - if you _do_ somehow end up in the overlay (e.g. two users racing to create an org) the button will still be turned off by a useEffect hook (`if (!orgCreationAllowed)`).

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/SwitchToNewOrgButton.tsx
wdoconnell: It sure isn't 🙃 . Removed it and converted it to a standard `href`.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -84,13 +92,33 @@ import {getOrgCreationAllowancesThunk} from 'src/identity/allowances/actions/thu
 import {generateProviderMap} from 'src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/utils/generateProviderMap'
 import {SafeBlankLink} from 'src/utils/SafeBlankLink'
 
+const switchToNewOrgButtonStyle = {
+  color: 'black',
+  textDecoration: 'underline',
+  cursor: 'pointer',
+}
+
```
wdoconnell, line 100: Since all this is doing now is creating a function called by the notification overlay that returns a link, I've moved it into the same component as the org creation overlay. No need to create a new file for that.

hoorayimhelping (approved):

PR #6353, opened 2022-11-30
feat: add marketo overlay to orgs list and global header

### LINKED ISSUE #6206: Create in-app marketo modal for upgrading from PAYG to Contract
New create-org design entails a modal using the marketo form when a user is upgrading from PAYG to Contract. Can likely use the shell of the[ create-org overlay modal.](https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1961%3A18041) 

Coordinate with sales and/or support based on input from @ALuckette and @gunnaraasen.

wdoconnell:
Closes #6206 

Adds the 'marketo' form used to contact support if the user elects to upgrade their account to add more organizations. 

When the `createDeleteOrgs` FF is on and the user has hit their org quota in the current account, the 'Upgrade to Create More Organizations' ad appears both in the global header, and also on the banner that displays on the orgslist page. Clicking that ad pulls up this overlay.

Since the data being submitted to associate the user with the sales inquiry already exists in the client redux state, we don't need to actually _display_ the marketo form to the user. For simplicity (and to avoid inserting the inconsistent marketo stylings into the app), this PR instead hides the form itself, and uses the Marketo Forms API to submit the form with our stock clockface components.

After I can sync up with growth/analytics, I'll add the one field the user should input (a comments field for any information on what they're requesting), and ensure that we're actually sending the form of id that is needed to auto-associate the inquiry with the user's account.

Screen Shot 2022-12-01 at 3 27 33 PM

Contact Form Submission
--
https://user-images.githubusercontent.com/91283923/205151046-cb6c67e5-a3a8-477d-a2d1-076cc76e83d5.mov

Backup Methods for Contacting Sales
--
If Marketo Script Fails to Load 
--
https://user-images.githubusercontent.com/91283923/205151205-cba3e2b5-7637-4193-bfa4-e62240772607.mov

If Marketo Form Fails to Load
--
https://user-images.githubusercontent.com/91283923/205151365-aceee309-53bd-45f7-8a00-d9e36f56ea1c.mov

If Marketo Form Fails to Submit
--
https://user-images.githubusercontent.com/91283923/205151547-f15ca757-b14b-4a48-8a42-322b86de0328.mov

### Checklist

### DIFF
--- a/src/identity/components/MarketoAccountUpgradeOverlay.scss
+++ b/src/identity/components/MarketoAccountUpgradeOverlay.scss
@@ -1,6 +1,6 @@
 @import '@influxdata/clockface/dist/variables.scss';
 
-#mktoForm_2826 {
+.marketo-account-upgrade--form {
   display: none;
 }
 
--- a/src/identity/components/MarketoAccountUpgradeOverlay.scss
+++ b/src/identity/components/MarketoAccountUpgradeOverlay.scss
@@ -1,0 +1,44 @@
+@import '@influxdata/clockface/dist/variables.scss';
+
+#mktoForm_2826 {
+  display: none;
+
+  // button {
+  //   opacity: 0;
+  // }
+
+  .mktoLabel {
+    height: 0px;
+    font-weight: 500;
+    color: white;
+    width: 150px !important;
+  }
+
+  .mktoFormRow {
+    width: 300px;
+  }
+
+  .mktoField {
+    border: 0px;
+    font-weight: 400;
+    font-family: 'Proxima Nova', Helvetica, Arial, Tahoma, Verdana, sans-serif;
+    transition: background-color 200ms cubic-bezier(0.18, 0.16, 0.2, 1),
+      border-color 200ms cubic-bezier(0.18, 0.16, 0.2, 1),
+      color 200ms cubic-bezier(0.18, 0.16, 0.2, 1),
+      box-shadow 200ms cubic-bezier(0.18, 0.16, 0.2, 1),
+      color 200ms cubic-bezier(0.18, 0.16, 0.2, 1);
+    outline: none;
+    appearance: none;
+    border-radius: 2px;
+    background-color: #07070e;
+    color: #fff;
+    border: 2px solid #333346;
+  }
+}
+
+.upgrade-to-contract-overlay--body {
+  p {
+    font-size: $cf-text-base-1;
+    margin-block-start: 0px;
+  }
+}
--- a/src/identity/components/MarketoAccountUpgradeOverlay.scss
+++ b/src/identity/components/MarketoAccountUpgradeOverlay.scss
@@ -2,41 +2,9 @@
 
 #mktoForm_2826 {
   display: none;
-
-  // button {
-  //   opacity: 0;
-  // }
-
-  .mktoLabel {
-    height: 0px;
-    font-weight: 500;
-    color: white;
-    width: 150px !important;
-  }
-
-  .mktoFormRow {
-    width: 300px;
-  }
-
-  .mktoField {
-    border: 0px;
-    font-weight: 400;
-    font-family: 'Proxima Nova', Helvetica, Arial, Tahoma, Verdana, sans-serif;
-    transition: background-color 200ms cubic-bezier(0.18, 0.16, 0.2, 1),
-      border-color 200ms cubic-bezier(0.18, 0.16, 0.2, 1),
-      color 200ms cubic-bezier(0.18, 0.16, 0.2, 1),
-      box-shadow 200ms cubic-bezier(0.18, 0.16, 0.2, 1),
-      color 200ms cubic-bezier(0.18, 0.16, 0.2, 1);
-    outline: none;
-    appearance: none;
-    border-radius: 2px;
-    background-color: #07070e;
-    color: #fff;
-    border: 2px solid #333346;
-  }
 }
 
-.upgrade-to-contract-overlay--body {
+.marketo-upgrade-account-overlay--body {
   p {
     font-size: $cf-text-base-1;
     margin-block-start: 0px;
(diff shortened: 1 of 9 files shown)


src/identity/components/MarketoAccountUpgradeOverlay.scss
```diff
@@ -0,0 +1,12 @@
+@import '@influxdata/clockface/dist/variables.scss';
+
+#mktoForm_2826 {
+  display: none;
+}
+
```
wdoconnell, line 6: The marketo form styling isn't pretty and is inconsistent with the app. To resolve that, we use our own submit button, and hide the marketo form itself.

src/identity/components/MarketoAccountUpgradeOverlay.tsx
```diff
@@ -0,0 +1,227 @@
...
+const MARKETO_SERVER_INSTANCE = '//get.influxdata.com'
+const MARKETO_SUBSCRIPTION_ID = '972-GDU-533'
+const MARKETO_FORM_ID = 2826
+
+// Error Messaging
+enum MarketoError {
+  FormLoadingError = 'failed to load marketo account upgrade form',
+  FormSubmitError = 'failed to submit marketo account upgrade form',
+  ScriptFetchError = 'failed to fetch marketo account upgrade script',
+  ScriptRuntimeError = 'failed to run marketo account upgrade script',
```
hoorayimhelping, line 62: It seems like these are used as error codes, right? I was initially confused beacuse I thought these were error messages shown to the user, rather than used as error codes to determine which toast to show - I think the error names, e.g. `FormLoadingError` and `FormSubmitError` are descriptive enough for these purposes, and it might actually be slightly more confusing to treat these like error messages shown to the user, rather than just making them a code like 0, 1, 2, 3, especially since `handleError` uses the actually error that is caught.

src/identity/components/MarketoAccountUpgradeOverlay.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const loadMarketoForm = useCallback(() => {
+    try {
+      if (!formHasLoaded) {
+        // API: https://developers.marketo.com/javascript-api/forms/api-reference/
+        window.MktoForms2?.loadForm(
+          MARKETO_SERVER_INSTANCE,
+          MARKETO_SUBSCRIPTION_ID,
+          MARKETO_FORM_ID,
+          () => {
+            // Marketo fields should remain hidden from user (CSS display: false)
```
hoorayimhelping, line 133:
```ts
CSS display: false
```

shouldn't this be `display: none` ? Just want to make sure the actually CSS is doing `display: none;` since `display: false` doesn't real

src/identity/components/MarketoAccountUpgradeOverlay.tsx
```diff
@@ -0,0 +1,227 @@
...
+    }
+  }, [accountId, formHasLoaded, handleError, user])
+
+  useEffect(() => {
+    try {
+      if (!scriptHasLoaded) {
+        const marketoScript = document.createElement('script')
+        marketoScript.id = 'marketoScript'
+        marketoScript.src = `${MARKETO_SERVER_INSTANCE}/js/forms2/js/forms2.min.js`
+        marketoScript.onload = loadMarketoForm
```
hoorayimhelping, line 169: do we need to add

```ts
marketoScript.async = true
```
here?

src/shared/copy/notifications/categories/accounts-users-orgs.ts
```diff
@@ -74,6 +74,27 @@ export const invitationWithdrawnSuccessful = (): Notification => ({
...
+  ...defaultErrorNotification,
+  message: 'Unable to load in-app contact form. ',
+  buttonElement,
+})
+
+export const marketoFormSubmitFailure = (
+  buttonElement: NotificationButtonElement
+): Notification => ({
+  ...defaultErrorNotification,
+  message: 'Submission failed. ',
```
hoorayimhelping, line 89: These messages a little bit mechanical and cold. It might be useful to let the user know that it's probably not their fault and how they can go about fixing the problem, even if it's a page refresh

hoorayimhelping (approved):
LGTM! A couple of minor suggestions I'll leave to you to decide

src/identity/components/MarketoAccountUpgradeOverlay.tsx
wdoconnell: Yup, that comment should reflect what the scss file actually has and is actually doing, i.e., `display: none`. 👍

src/identity/components/MarketoAccountUpgradeOverlay.tsx
wdoconnell: Yes, I'll add that - you're right that we want the script load to be non-blocking.

src/shared/copy/notifications/categories/accounts-users-orgs.ts
wdoconnell: I'll humanize these a bit, instead of using my 🤖 voice. 😉

src/identity/components/MarketoAccountUpgradeOverlay.tsx
wdoconnell: So, these are serving two purposes: (1) they determine which toast is sent to the user; and (2) they provide a descriptive message that goes to honeybadger (so that we can see if users are hitting a persistent issue with the form). 

Assigning the messages to an enum lets a single error handler deal with both questions - which toast am I showing the user, which is more limited - there are only two toasts - but also, which more detailed error message am I sending to honeybadger.

src/shared/copy/notifications/categories/accounts-users-orgs.ts
wdoconnell: (FYI - this message is also the first half of the toast only. The second half - `buttonElement` - is the link that directs them to the backup contact form.)

src/identity/components/MarketoAccountUpgradeOverlay.tsx
wdoconnell: I see your point though - this is a long message for something that's just being used as a honeybadger error name. let me fix this.

hoorayimhelping (approved):

PR #6365, opened 2022-12-05
fix: fix shared/explorer test with overwrite, rather than append request header in downloadhelper

wdoconnell:
We're seeing CI test failures in remocal-ui-test-firefox-1-tsm-shared/explorer that seem tied to recent changes to the csv downloader and explorer test. There's a service worker that appends a new content-type header. Proper behavior is to set rather than append the content-type header. 

The cred goes to @hoorayimhelping for tracking this down.

### Checklist


wiedld (approved):
Thanks for catching this one.

PR #6366, opened 2022-12-05
feat: add email and comment fields to marketo account upgrade form

wdoconnell:
Connects #6206 

This PR implements the UI side of changes by sales to the Marketo form to add user email and a comments field. This data is needed to associate submissions of the form with a specific user/account combination.

https://user-images.githubusercontent.com/91283923/205763653-5e63fae6-4da4-4d29-a8e6-5cdd451cfafd.mov

### Checklist

### DIFF
--- a/src/identity/components/MarketoAccountUpgradeOverlay.tsx
+++ b/src/identity/components/MarketoAccountUpgradeOverlay.tsx
@@ -5,6 +5,8 @@ import {
   Button,
   ComponentColor,
   ComponentStatus,
+  Input,
+  InputType,
   Overlay,
 } from '@influxdata/clockface'
 
@@ -62,6 +64,11 @@ enum MarketoError {
   ScriptRuntimeError = 'ScriptRuntimeError',
 }
 
+interface MarketoFormElement extends Element {
+  readOnly: boolean
+  disabled: boolean
+}
+
 // If marketo isn't working, still need the user to have some means of contacting sales.
 const SalesFormLink = (): JSX.Element => {
   return (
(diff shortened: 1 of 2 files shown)


src/identity/components/MarketoAccountUpgradeOverlay.tsx
```diff
@@ -145,10 +168,17 @@ export const MarketoAccountUpgradeOverlay: FC = () => {
...
+                .querySelectorAll('input, textarea, button')
+                .forEach((input: MarketoFormElement) => {
+                  if (
+                    input.className.includes('mktoField') ||
+                    input.className.includes('mktoButton')
+                  ) {
+                    input.readOnly = true
+                    input.disabled = true
+                  }
+                })
```
wdoconnell, line 181: The marketo form itself is hidden via CSS, since no user input is required other than in the (optional) comments field that lives in our UI. However, to ensure someone can't unhide the form in the dev console and change the inputs, I've set `readonly` and `disabled` true on each input in the marketo form. 

Setting these properties doesn't prevent the form from reflecting or being submitted with the correct data.

eatondustin1: Disabling the inputs won't actually protect from anything since the user could just unhide _and_ enable the fields, right? Since this is all front end, I'm not sure there's actually a good way to stop the user from changing the values, especially when the request goes out since they could just edit the values from the browser before it sends using something like the [Tamper Chrome extension](https://chrome.google.com/webstore/detail/tamper-dev/mdemppnhjflbejfbnlddahjbpdbeejnn).

The only way to really validate the input is to send it to our backend for us to validate, but I'm not sure we really need to avoid cases where the user puts in a different email, since that just means that we would follow up with someone who didn't request contact, which isn't a huge deal. The only issue I could see here would be spamming the request, in which case we should probably implement some kind of rate limiting, though even that may be impossible from the front end since the user could just modify the code in the browser.

eatondustin1 (approved):
Looks good 🚢

src/identity/components/MarketoAccountUpgradeOverlay.tsx
wdoconnell: Thanks! To close the loop on this: this is a super great point. I agree that we don't have a good way to _stop_ the user from changing the values with 100% certainty on the frontend (and since we don't control Marketo's script, we can't change or validate it on the backend). 

My inclination is that it's still worth having the form be protected on the frontend by at least hiding it via `display:none` and also setting `readonly` and `disabled` because it will resolve at least for practical purposes any ability to change the form inputs. A user could theoretically change them with the right extension, opening devtools, and playing with our CSS. But any user who gets to this stage is unlikely to want to do that, in an account where they are already paying for a full PAYG or Contract subscription with a number of orgs, and where the only point of the overlay is to streamline the point of contact with sales. Definitely possible, but I think even the worst case scenario if the form inputs were changed is that we wouldn't end up with a record that is easily identifiable as wrong (since the email / accountId combination likely won't match up in Marketo in the case of a user's manual edit).

PR #6369, opened 2022-12-06
fix(EAR-3809): Start service worker prior to page load.

wiedld:
May close https://github.com/influxdata/EAR/issues/3809.

The service worker is running in prod, but not intercepting fetch requests. Per mdn service work lifecycle docs, service worker should be registered prior to page load -- to ensure is properly set.

## Debugging evidence, from prod:
**Check in prod (new tools), after Bill's fix was promoted.**

Bug is still occurring:

Screen Shot 2022-12-06 at 11 08 29 AM

Can see service work is loaded and listening to events. (On inspect, confirmed in proper worker.)

Screen Shot 2022-12-06 at 11 08 52 AM

Outgoing csv download network calls (prior to service worker intercept), look the same in remocal (where no bug) versus in prod (where is bug). Only substantial difference the http protocol being used.

in_remocal

in_prod

## Attempted fix
Was not able to find a smoking gun on what is causing the service worker to not intercept requests in prod. However, did find documents stating that if the service work is running (and listening) but not intercepting -- then make sure the service worker is running prior to page load. (See mdn docs for [service worker life cycle](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#basic_architecture).) So this is an experiment, which will either (a) have no impact, or (b) fix the issue.

## Checklist

### DIFF
--- a/src/shared/contexts/app.tsx
+++ b/src/shared/contexts/app.tsx
@@ -1,4 +1,4 @@
-import React, {FC, useCallback} from 'react'
+import React, {FC, useCallback, useEffect} from 'react'
 import {useSelector, useDispatch} from 'react-redux'
 
 import {
@@ -10,6 +10,7 @@ import {
   disablePresentationMode,
   setFlowsCTA as setFlowsCTAAction,
   setSubscriptionsCertificateInterest as setSubscriptionsCertificateInterestAction,
+  setWorkerRegistration,
 } from 'src/shared/actions/app'
 import {
   timeZone as timeZoneFromState,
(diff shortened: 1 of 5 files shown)


src/shared/contexts/app.tsx
```diff
@@ -27,6 +28,17 @@ import {presentationMode as presentationModeCopy} from 'src/shared/copy/notifica
 import {AppState, TimeZone, Theme, NavBarState, FlowsCTA} from 'src/types'
 import {event} from 'src/cloud/utils/reporting'
 
+// Worker -- load prior to page load event, in order to intercept fetch in http/2.
+// see service worker life cycle. https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import downloadWorker from 'worker-plugin/loader!../workers/downloadHelper'
```
appletreeisyellow, line 35: Just curious, why use `worker-plugin/loader!...` instead of the absolute file path `src/workers/downloadHelper`?

jest.config.js
```diff
@@ -19,6 +19,8 @@ module.exports = {
     'react-dnd$': 'react-dnd/dist/cjs',
     'react-dnd-html5-backend$': 'react-dnd-html5-backend/dist/cjs',
     'dnd-core$': 'dnd-core/dist/cjs',
+    'worker-plugin/loader!../workers/downloadHelper':
+      '<rootDir>/src/shared/workers/downloadHelper',
```
wiedld, line 23: required to run jest unit tests, otherwise it does not resolve the import path.

hoorayimhelping (approved):

eatondustin1 (approved):
LGTM!

appletreeisyellow (approved):
Just one question below!

src/shared/contexts/app.tsx
wiedld: This is required by the webpack build (note: this is different from jest) in order to load the service worker using the worker-plugin loader (under webpack 4).

Issue #5305, opened 2022-08-05
Schema Composition:  design middleware layer between LSP and monaco-editor client.


wiedld:
## Problem

The monaco-editor and the LSP have a direct communication channel, which the UI does not currently listen into. What this means is that any LSP response/event sent to the monaco-editor client, must result in a monaco-editor action which provides the behavior we want. If we want any UI action (in response to the LSP), outside of the monaco-editor....then we currently cannot do so.

### Example use case: Composition event fails
1. user starts a schema composition
2. user asks for an LSP schema event
    * e.g. user adds a field
3. the UI (not the monaco-editor client) then:
    * triggers a json-rpc message to the LSP
    * adds the `field X` badge to the UI
4. the LSP returns some form of failure event
    * monaco-editor client gets the response.
    * UI itself (outside the editor) does not get the response.
6. monaco-editor can take actions, but those are limited.
    * e.g. `window/showMessage` only surfaces in the browser console. Not to the user.
7. if we have the UI listening to this traffic in some way:
    * remove the `field X` badge. Because the addition of the field filter was a failure.
    * also can surface to the user a notification, for any other failure case

### Other use cases:
* with the async nature of events, we can have a race condition with user-triggered events.
   * client requests, server responds, client applies changes:
        * each executeCommand (a.k.a. request for composition change), gets an applyEdit response.
        * response is an atomic operation to be performed
        * client performs the change
   * it takes time for the client to apply the change.
        * During this time, if another executeCommand is issued, the applyEdit response will be based on previous/old state. Before the last applyEdit was performed.
   * solution needs:
       * something listening to request/response
       * something which queues our executeCommand requests, and only sends the next request after the previous applyEdit completes.
       * something which updates the UI is the server response is a failure. 

### Current setup
* monaco-editor and LSP has a communication channel, via the monaco-languageclient
* the ways that we can interface with this channel, does not fulfill our needs.
  * see here: https://github.com/influxdata/ui/pull/5289

## Proposed solution

Build our own middleware between the workers, and the monaco-languageclient channel. This middleware will be on the level of the worker post/send with the buffer streams, which is lower down (used by) to the monaco-languageclient. In this way, we can be unconstrained in the middleware abilities, rather than relying on the interfaces provided by monaco-languageclient channel.

### Proposed implementation layer:
We already have that layer of abstract for the fallback worker. Likely will do a similar pattern around the worker and fallback worker.
[...]

PR #6374, opened 2022-12-07
feat(6334): UI responds at acts upon LSP-initiated commands, as a result of new yieldless paradigm.

### LINKED ISSUE #5717: Schema composition: build middleware, or equivalent solution.
Implementation of the design done here: https://github.com/influxdata/ui/issues/5305

### LINKED ISSUE #5925: Create UI to enable the lsp to present the user with messages
The [LSP specification describes a "show message" notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage) that can be used by the LSP server to present messages to the user on the client side. @juliajames12 has done some work already to present these messages to the user (mostly errors, last time I saw the design). We should listen for these `window/showMessage` notifications and hook them into our UI.

DOD:
  - when the LSP fires a `window/showMessage` notification, that message is presented to the user

### LINKED ISSUE #5926: Connect to the log trace lsp notifications and present them in the browser console
While doing some quick browsing in the LSP spec for #5925, I realized that [we could be using `window/logMessage` to present information in the browser console](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_logMessage), which would be super powerful for debugging. Once we have addressed #5925, all the plumbing should be in place to do something like the following:

```javascript
import {LogMessageNotification, MessageType} from 'vscode-languageserver-protocol';

connection.onNotification(LogMessageNotification, (rawMessage: string) => {
    const message = JSON.parse(rawMessage);
    if (message.type == MessageType.Error) {
        console.error(message.message)
    }
});
```

This API is just a guess, but a lot of this would be sorted out by solving #5925 first. [See this file for more information](https://github.com/microsoft/vscode-languageserver-node/blob/2e889fbffaccee0d054dc8af1feb356b27dc2873/protocol/src/common/protocol.ts)

DOD:
  - When the LSP sends an error message through to the client, it is visible in the browser console.

wiedld:
Part of https://github.com/influxdata/ui/issues/6334

Note: this is part of a chain of PRs.
`staging/yieldless-composition`  <-- 1st PR <-- 2nd PR <-- 3rd PR
The final staging branch will be the merge into master.

#### PR Chain:
1. https://github.com/influxdata/ui/pull/6344
2. this PR.
3. https://github.com/influxdata/ui/pull/6397 where tests finally pass!

## DONE:

Closes #5926 — trace/intercept lsp requests from the server
Closes #5717, #5305 — build a middleware, or equivalent solution, to prevent composition from getting out of sync with ui
Closes #5925 — surface to user the errors

Each commit gives a rough outline of the item done. ("Rough" because broke up the commits after the fact.). A high level summary is:
* break out the lsp-related types into a submodule. Divided into client-initiated commands (`executeCommand`) versus LspServer-inititiated commands.
* don't use default export for LspConnectionManager
* remove concept of composition divergence, since it no longer applies.
  * e.g. the UI-side no longer detects if a user-action occurs --> which results in divergence
  * instead, the LSP itself decides if it can find the composition, or drop it
* remove any reference/dependent use of the `|> yield()`
* separate out the different pathways (and what to message the LSP) for:
   * page reload (or first load)
   * change to composition
   * new script creation. (In which case the LSP drops the composition indirectly, based on a monaco-client request)
* introduce equivalent replacement (based on minimal needs) for the middleware.
  * this closes a bunch of issues.
  * also messages to the user, when the LspServer has given the current composition state:
     * if fields, tags, etc is not added --> will update the session
     * if composition is dropped --> will (1) turn off flux sync, (2) toast the user, and (3) remove composition styles
     * for each change to editor text --> will update the composition styles (e.g. when lines added within the block)
* debug tagValues     
  * the session store needs to be the src of truth
  * this assume requires the different tagValue store structures, to be mapped to each other.
  * QA'ed locally --> if LSP server does not accept an added tagValue, the middleware listener will correctly  update the UI session.

## Visual evidence:
Basic rules:
* if cannot recognize a previous composition --> start a new one.
  * this most often occurs when a new bucket is selected.
* if the LSP has dropped the composition for any reason --> toast, and turn off flux sync --> prompt user to resync
* tries to reattached composition on page reload  

https://user-images.githubusercontent.com/10232835/207445854-3381245d-de28-4e98-b30a-7d3e4c0bc8b2.mov

## Checklist

### DIFF
--- a/src/languageSupport/languages/flux/lsp/connection.ts
+++ b/src/languageSupport/languages/flux/lsp/connection.ts
@@ -13,7 +13,6 @@ import {
   DEFAULT_FLUX_EDITOR_TEXT,
   CompositionSelection,
 } from 'src/dataExplorer/context/persistance'
-import {comments} from 'src/languageSupport/languages/flux/monaco.flux.hotkeys'
 
 // LSP methods
 import {
@@ -35,20 +34,10 @@ import {
 // Utils
 import {event} from 'src/cloud/utils/reporting'
 
-// hardcoded in LSP
-const COMPOSITION_YIELD = '_editor_composition'
-
-const findLastIndex = (arr, fn) =>
-  (arr
-    .map((val, i) => [i, val])
-    .filter(([i, val]) => fn(val, i, arr))
-    .pop() || [-1])[0]
-
-class LspConnectionManager {
+export class ConnectionManager {
   private _worker: Worker
   private _editor: EditorType
   private _model: MonacoTypes.editor.IModel
-  private _snapshot: MonacoTypes.editor.ITextSnapshot
   private _preludeModel: MonacoTypes.editor.IModel
   private _variables: Variable[] = []
   private _compositionStyle: string[] = []
@@ -58,9 +47,8 @@ class LspConnectionManager {
   private _callbackSetSession: (
     schema: RecursivePartial<CompositionSelection>
   ) => void = () => null
-
-  // only add handlers on first page load.
-  private _compositionHandlersSet = false
+  private _dispatcher = _ => {}
+  private _first_load = true
 
   constructor(worker: Worker) {
     this._worker = worker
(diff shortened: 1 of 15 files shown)


src/languageSupport/languages/flux/lsp/types/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  didChange = 'textDocument/didChange',
+  didOpen = 'textDocument/didOpen',
+  didSave = 'textDocument/didSave',
+  Highlight = 'textDocument/documentHighlight',
+  Symbol = 'textDocument/documentSymbol',
+  FoldingRange = 'textDocument/foldingRange',
+  Hover = 'textDocument/hover',
+  References = 'textDocument/references',
+  Rename = 'textDocument/rename',
+  ExecuteCommand = 'workspace/executeCommand',
```
hoorayimhelping, line 29: What's the intent of having `executeCommand` and `showMessage`, but keeping this block in `index`?

src/languageSupport/languages/flux/lsp/types/showMessage.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export enum ActionItemCommand {
+  CompositionRange = 'CompositionRange',
+  CompositionState = 'CompositionState',
+}
+
+export enum LspClientRequestPriority {
+  ERROR = 1,
+  INFO = 3,
+}
```
hoorayimhelping, line 50: Completely unimportant to this overall PR, but I'm curious about `ERROR` being `1` and `INFO` being `3` with no key for `2`.

hoorayimhelping (approved):
I don't have anything that is blocking, just some random comments and friendly suggestions. I don't think I'm willing to push super hard on any of these though.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -256,31 +170,32 @@ class LspConnectionManager {
...
     this._compositionStyle = this._editor.deltaDecorations(
       this._compositionStyle,
       removeAllStyles
         ? []
-        : this._compositionSyncStyle(
-            compositionBlock?.startLine,
-            compositionBlock?.endLine,
-            schema.composition.synced
-          )
+        : this._compositionSyncStyle(range?.start.line, range.end.line)
```
appletreeisyellow, line 180: Just curious why `range?.start` but `range.end` without a `?`

My understanding is that line 174 already checked if range is null, then line 179 `[]` will be used, so it's good to use `range.start.line` without the question mark

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -126,112 +138,14 @@ class LspConnectionManager {
...
-          composition: {synced: false, diverged: true},
-        })
-      }
-    })
-  }
-
-  _compositionSyncStyle(startLine: number, endLine: number, synced: boolean) {
-    const classNamePrefix = synced
-      ? 'composition-sync--on'
-      : 'composition-sync--off'
```
appletreeisyellow, line 234: Seems like `composition-sync--off` and the related classes `composition-sync--off--first` and `composition-sync--off--last` can be removed in scss as well

appletreeisyellow (approved):
Thanks for splitting the work into each commit. The commit messages are helpful for me to review and understand the scope of each commit! The middleware is finally here! 🎉

I left two small comments below

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -103,6 +102,19 @@ class LspConnectionManager {
     )
   }
 
+  subscribeToConnection(connection: MonacoLanguageClient) {
+    /// class: https://github.com/microsoft/vscode-languageserver-node/blob/f97bb73dbfb920af4bc8c13ecdcdc16359cdeda6/client/src/browser/main.ts#L13
+    /// extended from class: https://github.com/microsoft/vscode-languageserver-node/blob/d7b0ef6eab79f31f514f6559b6950326b170a691/client/src/common/client.ts#L431
```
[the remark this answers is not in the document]
wiedld: Commenting convention the team started using under Paul, which is imported from rust (technically rust doc convention). Will switch back to js convention.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -256,31 +170,32 @@ class LspConnectionManager {
...
   }
 
-  _setEditorBlockStyle(schema: CompositionSelection = this._session) {
-    const compositionBlock = this._getCompositionBlockLines(
-      this._model.getValue()
-    )
-
-    const removeAllStyles = !compositionBlock || schema.composition.diverged
+  _setEditorBlockStyle(range: LspRange | null) {
+    const removeAllStyles = range == null
```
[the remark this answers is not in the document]
wiedld: All about reducing mental load. Done.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -256,31 +170,32 @@ class LspConnectionManager {
...
-          )
+        : this._compositionSyncStyle(range?.start.line, range.end.line)
     )
   }
 
+  _isNewScript(
+    schema: CompositionSelection,
+    previousState: CompositionSelection
+  ): boolean {
+    return previousState.bucket != null && schema.bucket == null
```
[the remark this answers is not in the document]
wiedld: Purposefully playing a little loose here, since the source of this `bucket` is plenty of upstream, interconnected code which manages the session state. So I want to handle cases where we go from "user see nothing" (undefined or null), to user sees a bucket.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -311,18 +226,47 @@ class LspConnectionManager {
...
+    }
+  }
+
   _diffSchemaChange(
     schema: CompositionSelection,
     previousState: CompositionSelection
   ) {
     const toAdd: Partial<CompositionSelection> = {}
     const toRemove: Partial<CompositionSelection> = {}
+    let delay = false
```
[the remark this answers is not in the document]
wiedld: Like this use of verbage. Updated.

src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -349,69 +293,99 @@ class LspConnectionManager {
...
+      schema,
+      previousState
+    )
 
-    if (!this._compositionHandlersSet) {
-      this._initCompositionHandlers()
-      return // do not re-init schema session on reload
+    if (this._first_load) {
+      this._first_load = false
+      setTimeout(() => this._initLspComposition(toAdd), 3000) // LSP server must come online
```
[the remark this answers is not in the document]
wiedld: Excellent idea. Done.

src/languageSupport/languages/flux/lsp/types/index.ts
wiedld: Made some changes, and added in code-docs, to make this more clear. Basically:

```
/**
 * "Methods" == per each request sent to the LSP, there is an associated method.
 * 
 * Special methods:
 *    * the executeCommand method:
 *        * we construct this message request in the UI code (not the monaco-editor) --> send to the LSP
 *        * specific executeCommand structure (per spec) is typed in './executeCommand'
 *    * the showMessage method:
 *        * the LSP sends the UI this message request --> UI (not monaco-editor) performs downstream actions.
 *        * specific showMessage structure (per spec) is typed in './showMessage
 */

export * from './executeCommand'
export * from './methods'
export * from './showMessage'
```

src/languageSupport/languages/flux/lsp/types/showMessage.ts
wiedld: `LspClientRequestPriority` is from the spec, with those values set. We currently only use the levels 1 and 3. I'll add 2 just for completeness.

src/languageSupport/languages/flux/lsp/connection.ts
wiedld: Thanks for catching, fixed!

src/languageSupport/languages/flux/lsp/connection.ts
wiedld: Good call. Removed.

Issue #5477, opened 2022-08-19
Multi Org Phase 2 - Delete Org


Luckette:
For multi-org, once users can create new organizations, we anticipate the lack of ability to delete organizations will lead to many new support cases and EARs to clean up organizations that were created unintentionally. Therefore, we want to include org deletion in the scope of the phase 2 work.

[Corresponding quartz epic](https://github.com/influxdata/quartz/issues/6843)

[Delete organization design (Figma)](https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1944%3A14887)

Key highlights:
- Update the "delete organization" section for PAYG users to match the delete org button that already exists for annual users. 
- Clicking on the button brings up a pop-up modal with information on the deletion, a confirmation checkbox, and a button to trigger org deletion. See modal design here: https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1944%3A14887
- Clicking the confirmation button will make an org delete request to quartz.
- After clicking the confirmation button, the button will display a spinner until the org delete request returns with success or failure. Display spinner from clockface. 
- If the request fails, the error is displayed to the user. see message here: https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1944%3A14557
- If successful, switch the user to their default organization (or new default org) and provide a notification that the organization has been deleted.
- In the organization list, the deleted organization should be shown as "Deletion in progress" with a tooltip that the organization can be recovered by reaching out to support.
- After seven days, the organization moves to a "deleted" state and the org is deleted in IDPE and will no longer show up in the organization list.

Implementation:

- [ ] TBD

Luckette:
@gunnaraasen This is the UI ticket for the delete org experience. We don't have concrete requirements or designs yet.

hoorayimhelping:
Blocked on backend API work

wdoconnell:
Delete button needs to function as follows.



![Image](https://user-images.githubusercontent.com/91283923/204342793-c8726ff1-f0dd-46fe-9216-350568e4c19c.png)

PR #6379, opened 2022-12-07
feat: add overlay for deleting PAYG and Contract orgs

wdoconnell:
WIP while prepping screenshots.

Helps with #5477 

### Checklist

### DIFF
--- a/src/identity/selectors/index.ts
+++ b/src/identity/selectors/index.ts
@@ -2,10 +2,22 @@
 import {AppState} from 'src/types'
 import {RemoteDataState} from '@influxdata/clockface'
 
+export const selectCurrentAccount = (
+  state: AppState
+): AppState['identity']['currentIdentity']['account'] => {
+  return state.identity.currentIdentity.account
+}
+
 export const selectCurrentAccountId = (
   state: AppState
 ): AppState['identity']['currentIdentity']['account']['id'] => {
   return state.identity.currentIdentity.account.id
+}
+
+export const selectCurrentAccountName = (
+  state: AppState
+): AppState['identity']['currentIdentity']['account']['name'] => {
+  return state.identity.currentIdentity.account.name
 }
 
 export const selectCurrentAccountType = (
@@ -20,10 +32,20 @@ export const selectCurrentIdentity = (
   return state.identity.currentIdentity
 }
 
+export const selectCurrentOrg = (
+  state: AppState
+): AppState['identity']['currentIdentity']['org'] => {
+  return state.identity.currentIdentity.org
+}
+
 export const selectCurrentOrgId = (
   state: AppState
 ): AppState['identity']['currentIdentity']['org']['id'] => {
   return state.identity.currentIdentity.org.id
+}
+
+export const selectDefaultOrgId = (state: AppState): string => {
+  return state.identity.quartzOrganizations.orgs.find(org => org.isDefault).id
 }
 
 export const selectOperatorRole = (
--- a/src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
+++ b/src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
@@ -1,0 +1,187 @@
+// Libraries
+import React, {FC, useContext, useState} from 'react'
+import {useDispatch, useSelector} from 'react-redux'
+import {
+  Alert,
+  AlignItems,
+  Button,
+  ComponentColor,
+  ComponentSize,
+  ComponentStatus,
+  FlexBox,
+  FlexDirection,
+  IconFont,
+  Input,
+  InputLabel,
+  InputType,
+  Overlay,
+} from '@influxdata/clockface'
+
+// Selectors
+import {
+  selectCurrentAccount,
+  selectCurrentOrg,
+  selectUser,
+} from 'src/identity/selectors'
+
+// API
+import {deleteOrganization} from 'src/identity/apis/org'
+
+// Constants
+import {CLOUD_URL} from 'src/shared/constants'
+
+// Notifications
+import {deleteOrgFailed} from 'src/shared/copy/notifications'
+import {notify} from 'src/shared/actions/notifications'
+
+// Utils
+import {OverlayContext} from 'src/overlays/components/OverlayController'
+import {reportErrorThroughHoneyBadger} from 'src/shared/utils/errors'
+import {setToLocalStorage} from 'src/localStorage'
+
+// Styles
+import './DeletePaidOrgOverlay.scss'
+
+// This is required to avoid unnecessary console errors thrown by <Input /> when type=radio.
+const nullOnChange = () => null
+
+export const DeletePaidOrgOverlay: FC = () => {
+  const account = useSelector(selectCurrentAccount)
+  const org = useSelector(selectCurrentOrg)
+  const user = useSelector(selectUser)
+  const dispatch = useDispatch()
+
+  const {onClose} = useContext(OverlayContext)
+
+  const [deleteButtonStatus, setDeleteButtonStatus] = useState(
+    ComponentStatus.Disabled
+  )
+  const [userAcceptedTerms, setUserAcceptedTerms] = useState(false)
+
+  const toggleAcceptedTerms = () => {
+    const currentAcceptanceStatus = !userAcceptedTerms
+    setUserAcceptedTerms(currentAcceptanceStatus)
+
+    setDeleteButtonStatus(
+      currentAcceptanceStatus
+        ? ComponentStatus.Default
+        : ComponentStatus.Disabled
+    )
+  }
+
+  const handleDeleteOrg = async () => {
+    try {
+      setDeleteButtonStatus(ComponentStatus.Loading)
+
+      await deleteOrganization(org.id)
+
+      setDeleteButtonStatus(ComponentStatus.Default)
+      setToLocalStorage('justDeletedOrg', org.name)
+      window.location.href = `${CLOUD_URL}`
+
+      onClose()
+    } catch (err) {
+      dispatch(notify(deleteOrgFailed(org.name)))
+      reportErrorThroughHoneyBadger(err, {
+        name: 'Org deletion failed',
+        context: {
+          user,
+          account,
+          org,
+        },
+      })
+      setDeleteButtonStatus(ComponentStatus.Disabled)
+      setUserAcceptedTerms(false)
+    }
+  }
+
+  return (
+    <Overlay.Container
+      className="org-delete-overlay"
+      maxWidth={600}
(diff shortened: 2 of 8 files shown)


hansenreid (dismissed):
Looks good to my naive react eyes. Might be worth getting a second opinion, especially around the local storage as you mentioned during standup.

src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
```diff
@@ -0,0 +1,191 @@
...
+        ? ComponentStatus.Default
+        : ComponentStatus.Disabled
+    )
+  }
+
+  const handleDeleteOrg = async () => {
+    try {
+      setDeleteButtonStatus(ComponentStatus.Loading)
+
+      await deleteOrganization(org.id)
```
hoorayimhelping, line 76: does this need to be `await`ed? can some things be done immediately afterward? It seems like `setDeleteButtonStatus()` and `onClose()` can happen immediately. From a usability standpoint, it might be better to make the button disabled as soon as the request is made.

It might be worth changing this into a `.then().catch()` pattern, like:

```ts
deleteOrganization(org.id).then(() => {
  // Adding (and on reload, removing) this property from local storage
  // is a work-around so that it's possible for the user to be notified of org deletion
  // AFTER app reload. See src/components/notifications/Notifications.tsx,
  // which detects the property, and if found, sends a notification then deletes it.
  setToLocalStorage('justDeletedOrg', org.name)
  window.location.href = `${CLOUD_URL}`
}).catch(err => {
  dispatch(notify(deleteOrgFailed(org.name)))
    name: 'Org deletion failed',
    context: {
      user,
      account,
      org,
    },
  })
  setDeleteButtonStatus(ComponentStatus.Disabled)
  setUserAcceptedTerms(false)
})

setDeleteButtonStatus(ComponentStatus.Default)
setDeleteButtonStatus()
```

that will let `setDeleteButtonStatus` and `setDeleteButtonStatus` get called immediately, then the `setToLocalStorage` and `window.location.href` happen after it succeeds

src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
```diff
@@ -0,0 +1,191 @@
...
+            <b>suspended immediately</b>.
+          </li>
+          <li>
+            Your final billing statement will be calculated for any usage and
+            storage incurred prior to the deletion of your organization and your
+            credit card will be charged for the amount.
+          </li>
+          <li>
+            Before continuing, you are responsible for exporting any data or
+            content - including dashboards, tasks and variables - from the user
```
hoorayimhelping, line 135:
```suggestion
            content - including dashboards, tasks, and variables - from the user
```
I think we want the Oxford comma here to distinguish tasks and variables as separate entities

src/organizations/components/OrgProfileTab/DeletePanel.tsx
```diff
@@ -75,11 +84,32 @@ const DeletePanel: FC = () => {
...
                 icon={IconFont.Trash_New}
-                onClick={handleDeleteClick}
+                onClick={handleDeleteFreeAccount}
               />
             </FlexBox.Child>
           </>
         )}
+        {CLOUD &&
+          isFlagEnabled('createDeleteOrgs') &&
+          (account.type === 'pay_as_you_go' || account.type === 'contract') && (
```
hoorayimhelping, line 94: might be worth abstracting this expression out into a variable

src/shared/components/notifications/Notifications.tsx
```diff
@@ -26,8 +34,22 @@ const matchGradientToColor = (style: NotificationStyle): Gradients => {
 
 const Notifications: FC = () => {
   const notifications = useSelector(getNotifications)
+  const accountName = useSelector(selectCurrentAccountName)
+  const userJustDeletedAnOrg = Boolean(getFromLocalStorage('justDeletedOrg'))
```
hoorayimhelping, line 38: Will this run reliably run every time? Is `Notifications.tsx` guaranteed to re-render after the delete operation has been carried out?

Is this a way to get around the lack of any kind of event bus and the delete operation being highly asynchronous?

hoorayimhelping (approved):
Generally LGTM. Just a couple of general questions

src/identity/selectors/index.ts
```diff
@@ -2,12 +2,24 @@
...
 export const selectCurrentAccountId = (
   state: AppState
 ): AppState['identity']['currentIdentity']['account']['id'] => {
   return state.identity.currentIdentity.account.id
 }
 
+export const selectCurrentAccountName = (
+  state: AppState
+): AppState['identity']['currentIdentity']['account']['name'] => {
+  return state.identity.currentIdentity.account.name
```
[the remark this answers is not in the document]
wdoconnell: Good point - I'll remove. Generally, I've tried to get away from selectors that return an entire object rather than a primitive value. However, in this case, the entire account object will never change, because the current account remains identical until app reload.

src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
wdoconnell: 💯 Sure do.

src/organizations/components/OrgProfileTab/DeletePanel.tsx
wdoconnell: 👍

src/shared/components/notifications/Notifications.tsx
wdoconnell: As far as I can tell in remocal testing, this is reliable. I will need to do additional validating in prod behind the feature flag to ensure it functions the same way. 

Notifications.tsx is guaranteed to re-render not due to the delete operation, but due to the quartz redirect that causes an organization switch. That redirect reloads the app, and Notifications along with it. 

To flesh out the thinking here (and provide some alternatives that weren't chosen): 

 Since that switch is a reload, we lose our application state, including redux and context, so dispatching a notification before the reload is a no-go (moreover, closing the overlay triggers a re-render, so the user would see a flash of the notification early, then a reload - have tested this). We're left with either storing the trigger for the reload on the client side but __outside_ of the application state (e.g._ LocalStorage) or server side (send a different value from quartz when a user is redirected after deleting an org). I've opted for the former here since it's simpler than a backend change, and because the notification behaves as expected in my testing (appears once on app reload, then never again until another org is deleted).

If there are lingering concerns: another alternative would be to have the switch occur after a timeout: the notification displays _before_ the org is deleted, then the user gets switched into the default org after a few seconds. Probably more annoying, but more reliable since we're not storing anything in the browser.

src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
gunnaraasen: Users cannot take any actions on a deleted org so we need to redirect the user to an existing org as soon as the delete request returns a successful response. My vote is to keep the modal open until the request completes. In addition to showing the button as loading, we should also prevent the modal from being dismissed until the request completes (and re-enable dismissal if there's an error returned).

wdoconnell: 👍

src/accounts/CancellationTerms.tsx
```diff
@@ -28,7 +28,7 @@ const CancellationTerms: FC<Props> = ({hasAgreedToTerms, onAgreedToTerms}) => (
       </li>
       <li>
         Before continuing, you are responsible for exporting any data or content
-        (including dashboards, tasks and variables) you wish to keep. (edited)
+        (including dashboards, tasks, and variables) you wish to keep. (edited)
       </li>
```
wdoconnell, line 32: We were missing oxford commas for the same copy throughout, so I confirmed it.

src/shared/components/notifications/Notifications.tsx
hoorayimhelping: Cool, thanks for the explanation, and for doing the work required to write the explantion.

src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
hoorayimhelping: Thanks for the direction Gunnar

hoorayimhelping (approved):

gunnaraasen:
This is looking really good. I have a couple small tweaks.

- Can we change the end of the error message to say "Please reach out to support@influxdata.com" instead of "Try again later"?
- When we redirect to the default org after deletion, can we put the user on the orgs list page instead of the homepage? That way they'll also be able to see the org as deleted on the org list in addition to the notification.

wdoconnell:
@gunnaraasen I've implemented the above (minus a portion that involves the deeplinking ticket, which we're handling separately and which requires confirmation on some aim work). 

But here's another thought, with thanks for some pairing on this with @blegesse-git :

Instead of sending the notification that the org was deleted _after_ the user is redirected into the new org (which requires the gross localStorage workaround, and is less testable, since we can't test behavior that occurs after a quartz redirect), what if we keep the overlay open and leave the button state as 'loading' until the user has a couple seconds to read the notification, then we redirect into the new org? Would look like this.

https://user-images.githubusercontent.com/91283923/206761053-ffe31fa0-741e-493c-9179-3d22027f63ac.mov

I tend to like it better - it's not a long wait, it's clearer which org the user is in when the notification pops up, and the user is still locked from trying to close the window until the redirect is over.

wdoconnell:
FYI - this is good to go with a new implementation, but I'll hold off on merging until Gunnar weighs in. I can revert the last commit if he prefers the other one.

gunnaraasen:
I think both options are roughly equivalent in terms of displaying the information to the user. If showing the notification a few seconds before redirect is more maintainable, then let's go that route.

src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
```diff
@@ -0,0 +1,218 @@
...
+    }, 5000)
+
+    deleteOrganization(org.id)
+      .then(() => {
+        clearTimeout(responseTimer)
+        dispatch(notify(deleteOrgSuccess(org.name, account.name)))
+        setTimeout(() => {
+          onClose()
+          window.location.href = CLOUD_URL
+        }, 4000)
```
hoorayimhelping, line 111: I think we should consider storing the value of these [magic numbers](https://stackoverflow.com/a/47902/92446 into variables.

hoorayimhelping (approved):
lgtm, except for the magic numbers

blegesse-git (approved):

PR #6373, opened 2022-12-07
fix: skip flaky csv download test for explorer.

wiedld:
Hotfix current CI blockage.

## Observations:
* explorer tests have begun timing out without failure:
   * https://app.circleci.com/pipelines/github/influxdata/monitor-ci/18092/workflows/0543693d-be04-4647-a709-4a149de97935/jobs/1285612
   * https://app.circleci.com/pipelines/github/influxdata/monitor-ci/18087/workflows/e5af1ea5-70f5-4e22-b742-c6dca3c49091/jobs/1285232
   * https://app.circleci.com/pipelines/github/influxdata/monitor-ci/18085/workflows/681f11a6-6128-4703-ad8f-272aff1389bf/jobs/1284833
   * https://app.circleci.com/pipelines/github/influxdata/monitor-ci/18076/workflows/7433d3a2-fac9-46f3-a264-f4dd1575552f/jobs/1283822
* Not sure of the root cause. Test the below scenarios in 20x test runs:
   * Attempting various explicit timeout controls at any waiting points...no dice.
   * Tried with latest service worker fixes...no dice.
   * Tried a few other low hanging fruit...no dice.
   * Therefore => Skipping tests.

## Checklist
- [x] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [x] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [x] Documentation updated or issue created (provide link to issue/PR)
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [x] Feature flagged, if applicable

PR #6355, opened 2022-12-02
chore: document in code the tech debt behind the LSP worker buffer.

wiedld:
Since this topic keeps coming up in conversation, decided to document in code. For our future selves.
The current time and ordering dependencies of how the LSP server boots up.

### Checklist
- [x] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [x] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [x] Documentation updated or issue created (provide link to issue/PR)
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [x] Feature flagged, if applicable

PR #6399, opened 2022-12-14
chore: replace csv download using service worker in query experience

### LINKED ISSUE #6354: Qx Builder: use new csv download approach (full dataset bypassing browser memory).
Is basically the solution from [here](https://github.com/influxdata/ui/pull/6062).
* reuse same server worker
* feel free to extract DRY common code, if it's not gnarly.
* basic UI rule:
  * when no service worker --> don't even have the download button
  * if have worker (so can download) --> have button, but with optional disabled state (e.g. if no query written yet).
  *  That's^^ the general rule in the old data explorer. Feel free to suggest/improve on this UX decision.

appletreeisyellow:
Closes #6354 

This PR replaces the csv download approach in the query experience with service worker. The reason why we decide to use this new approach is documented in https://github.com/influxdata/ui/pull/6355.

The main work was covered in https://github.com/influxdata/ui/pull/6062, https://github.com/influxdata/ui/pull/6369, and https://github.com/influxdata/ui/pull/6377 for the old data explorer, and it's working in production. This PR is a follow-up for the query experience. 

The tests for csv download is skipped with the same reason as https://github.com/influxdata/ui/pull/6373

### Checklist


src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -120,9 +115,10 @@ const ResultsPane: FC = () => {
     selection,
     resource,
   } = useContext(PersistanceContext)
+  const orgID = useSelector(getOrg).id
```
hoorayimhelping, line 118: This is a potentially dangerous expression, since [`getOrg`](https://github.com/influxdata/ui/blob/master/src/organizations/selectors/index.ts#L22) can return `null`. This seems very unlikely, since if there's no org in state, it probably shouldn't ever be able to get to this point and there are probably larger issues going on. But I still wanted to call it out, just in case.

src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -142,29 +138,43 @@ const ResultsPane: FC = () => {
...
-      {
-        rawBlob: true,
-        language,
-        bucket: selection.bucket,
-      }
-    )
-    setCancelId(promise.id)
+  const downloadByServiceWorker = () => {
+    try {
+      event('runQuery', {context: 'query experience'})
```
hoorayimhelping, line 143: should we differentiate this event from a regular query execution? Technically it's running a query, but it's running a query for a CSV download. I think we can either do something like

```ts
event('runQuery.downloadCSV', {context: 'query experience'})
```
or something like
```ts
event('runQuery', {context: 'query experience - download CSV'})
```

src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -313,4 +304,4 @@ const ResultsPane: FC = () => {
   )
 }
 
-export default ResultsPane
+export {ResultsPane}
```
hoorayimhelping, line 307: thank you thank you!

hoorayimhelping (approved):
LGTM! A couple of minor suggestions, but they're not hard requirements.

I just want to say: thank you for taking the time to write such a fantastic PR description. It explains what's happening and gives context as well as relevant prior work. All of that made my job reviewing this code much easier. Outstanding job!

src/dataExplorer/components/ResultsPane.tsx
appletreeisyellow: Good call!

hoorayimhelping (dismissed):

src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -262,29 +272,10 @@ const ResultsPane: FC = () => {
...
-                    cancel(csvDownloadCancelID)
-                    setCancelId(null)
-                  }}
-                  color={ComponentColor.Danger}
-                />
-              )}
+              <CSVExportButton
+                disabled={submitButtonDisabled}
+                download={downloadByServiceWorker}
+              />
```
wiedld, line 278: Nice reuse!

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -182,7 +182,7 @@ describe('Script Builder', () => {
       cy.wait('@query -15m')
     })
 
-    describe('data completeness', () => {
+    describe.skip('data completeness', () => {
```
wiedld, line 185: This will fail because:
(1)  the download now has a dynamically downloaded csv name
   * so we need to i[ntercept the request response](https://github.com/influxdata/ui/blob/master/cypress/e2e/shared/explorer.test.ts#L634-L646), instead of presuming an exact name.

(2) since a form submit, we need to mock the page reload.
   * cypress will fail tests unless this happens.
   * it's done [this way](https://github.com/influxdata/ui/blob/master/cypress/e2e/shared/explorer.test.ts#L621-L623), (as pulled directly from cypress docs)

wiedld (changes requested):
Excellent work. Just asking that we try to update the test, rather than skipping.

It was skipped in the explorer.test due to firefox flaky. But since no firefox --> hopefullly we can have these csv download tests functional!

cypress/e2e/shared/fluxQueryBuilder.test.ts
appletreeisyellow: Thanks for looking into this! I updated the test

cypress/e2e/shared/fluxQueryBuilder.test.ts
```diff
@@ -275,32 +290,32 @@ describe('Script Builder', () => {
...
-        cy.getByTestID('data-explorer--csv-download').should('be.disabled')
-        selectMeasurement('ndbc_1table')
-        cy.getByTestID('flux-editor').contains(
-          `|> filter(fn: (r) => r._measurement == "ndbc_1table")`
-        )
+        cy.getByTestID('csv-download-button').should('be.disabled')
+        cy.getByTestID('flux-editor').monacoType(`{selectall}{enter}
+          from(bucket: "defbuck4") |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
+          |> filter(fn: (r) => r._measurement == "ndbc_1table")
+        `)
```
appletreeisyellow, line 306: Instead of using flux lsp composition, I just type the query into the flux-editor since this test case is mainly testing for the csv download, instead of flux lsp composition

wiedld: good call.

wiedld (approved):
Excellent updates. Thanks for getting this test working!

PR #6409, opened 2022-12-16
chore: allow free org with non-default allowances to delete org

### LINKED ISSUE #6396: make 'delete' button for free accounts permit org deletion if there is more than one org in the account (proof of concept)
Free users only have the option to delete their entire account. Normally this is fine, as free accounts have a one-org cap by default.

For proof of concept trials where a customer is granted additional orgs in their free account, grant the ability to delete an org in the account (not just the entire account). This should be flagged behind `createDeleteOrgs`.

wdoconnell:
Closes #6396 

For proof of concept use cases we want the ability to give a user more than one org in a free account - and to be able to delete those orgs. This PR adjusts the logic for the 'delete' buttons on the org settings page.

Old Behavior:
- In a free account, your only "delete" option is to delete the free account. 
- If the `createDeleteOrgs` FF is on, a payg or contract account can delete the currently active org.

New Behavior:
- If you're in a free account with _only one org_, same behavior as before (only option is to delete the free account).
- If the `createDeleteOrgs` FF is on (and you're not in a free account with only one org), you can delete the current org.

### Checklist

- [X] Feature flagged, if applicable - `createDeleteOrgs` (partial)

### DIFF
--- a/cypress/e2e/cloud/about.test.ts
+++ b/cypress/e2e/cloud/about.test.ts
@@ -46,7 +46,13 @@ describe.skip('About Page for free users with only 1 user', () => {
 })
 
 describe('About Page for free users with multiple users', () => {
-  beforeEach(() =>
+  beforeEach(() => {
+    cy.intercept('GET', 'api/v2/quartz/identity', req => {
+      req.continue(res => {
+        res.body.user.orgCount = 1
+        res.send(res.body)
+      })
+    })
     cy.flush().then(() =>
       cy.signin().then(() => {
         cy.get('@org').then(({id}: Organization) => {
@@ -60,7 +66,7 @@ describe('About Page for free users with multiple users', () => {
         })
       })
     )
-  )
+  })
   it('should display the warning and allow users to navigate to the users page when trying to delete when the user has multiple users', () => {
     cy.getByTestID('delete-org--button').should('exist').click()
 
--- a/src/organizations/components/OrgProfileTab/DeletePanel.tsx
+++ b/src/organizations/components/OrgProfileTab/DeletePanel.tsx
@@ -67,15 +67,16 @@ const DeletePanel: FC = () => {
     handleDeleteFreeAccount = handleShowWarning
   }
 
-  const showDeleteButtonInFreeAccount = account.type === 'free'
-  const showDeleteButtonInPaidAccount =
-    isFlagEnabled('createDeleteOrgs') &&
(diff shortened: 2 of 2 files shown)


wdoconnell:
Best reviewed with `?w=1`

src/organizations/components/OrgProfileTab/DeletePanel.tsx
```diff
@@ -67,15 +67,16 @@ const DeletePanel: FC = () => {
...
   }
 
-  const showDeleteButtonInFreeAccount = account.type === 'free'
-  const showDeleteButtonInPaidAccount =
-    isFlagEnabled('createDeleteOrgs') &&
-    (account.type === 'pay_as_you_go' || account.type === 'contract')
+  const showDeleteFreeAccountButton =
+    CLOUD && account.type === 'free' && user.orgCount === 1
+
+  const showDeleteOrgButton =
```
hoorayimhelping, line 73: These boolean variables are easy to confuse with functions. `showDeleteFreeAccountButton()` and `showDeleteOrgButton()` are pretty natural. I think it might be less confusing to name them something like `shouldShowDeleteFreeAccountButton` and `shouldShowDeleteOrgButton`

hoorayimhelping (approved):

abshierjoel (approved):

src/organizations/components/OrgProfileTab/DeletePanel.tsx
wdoconnell: Good point - I'll rename them. 👍 

I need to make a modification to the test for this page too, to cover the intended case - free account with one org in it.

cypress/e2e/cloud/about.test.ts
```diff
@@ -46,7 +46,13 @@ describe.skip('About Page for free users with only 1 user', () => {
 })
 
 describe('About Page for free users with multiple users', () => {
-  beforeEach(() =>
+  beforeEach(() => {
+    cy.intercept('GET', 'api/v2/quartz/identity', req => {
+      req.continue(res => {
+        res.body.user.orgCount = 1
+        res.send(res.body)
```
wdoconnell, line 53: The default state in quartz mock is more than one org, so we have to modify it here to ensure the test for a 'default' free account (which has only one org) works correctly.

hoorayimhelping (approved):

PR #6410, opened 2022-12-16
feat(notebooks): add actions, thunks, reducer, selector for checking if org has notebooks

hoorayimhelping:
Connects: https://github.com/influxdata/idpe/issues/16723

- Creates the redux machinery to store a list of notebooks in redux state for simple checking if an org has notebooks.
- Calls the thunk in `GetOrganizations` after `org.id` is retreived. This makes a request across the wire and stores notebooks in redux state
- Nothing is currently done with the notebooks in state, that will come in a future PR

This doesn't use the existing flows context in order to avoid wrapping most of the appliction in the flows provider, which seems an odd thing to do when deprecating a feature. This also avoids potentially re-rendering of the entire application (since `GetOrganizations` is a component that is at the root of nearly every other component in the application) when a single thing in flows is updated.

### Checklist

### DIFF
--- a/src/flows/actions/flowsThunks.ts
+++ b/src/flows/actions/flowsThunks.ts
@@ -1,0 +1,22 @@
+import {Dispatch} from 'redux'
+
+import {Notebook} from 'src/client/notebooksRoutes'
+
+import {getNotebooks as fetchNotebooks} from 'src/client/notebooksRoutes'
+import {setNotebooks} from 'src/flows/actions/flowsActions'
+
+export const getNotebooks =
+  (orgID: string) => async (dispatch: Dispatch<any>) => {
+    try {
+      const response = await fetchNotebooks({query: {orgID}})
+
+      if (response.status !== 200) {
+        throw new Error()
+      }
+
+      dispatch(setNotebooks(response.data.flows))
+    } catch {
+      const emptyNotebooks = [] as Notebook[]
+      dispatch(setNotebooks(emptyNotebooks))
+    }
+  }
--- a/src/shared/containers/GetOrganizations.tsx
+++ b/src/shared/containers/GetOrganizations.tsx
@@ -39,6 +39,7 @@ import {RemoteDataState} from 'src/types'
 
 // Thunks
 import {getQuartzIdentityThunk} from 'src/identity/actions/thunks'
+import {getNotebooks} from 'src/flows/actions/flowsThunks'
 
 const GetOrganizations: FunctionComponent = () => {
   const {status: orgLoadingStatus} = useSelector(getAllOrgs)
@@ -49,7 +50,6 @@ const GetOrganizations: FunctionComponent = () => {
 
   const dispatch = useDispatch()
 
-  // This doesn't require another API call.
   useEffect(() => {
     if (orgLoadingStatus === RemoteDataState.NotStarted) {
       dispatch(getOrganizations())
@@ -91,6 +91,12 @@ const GetOrganizations: FunctionComponent = () => {
       })
     }
   }, [account.type, account.isUpgradeable])
+
+  useEffect(() => {
+    if (CLOUD && org?.id) {
+      dispatch(getNotebooks(org.id))
+    }
+  }, [org.id])
 
   return (
     <PageSpinner loading={orgLoadingStatus}>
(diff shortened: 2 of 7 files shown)


src/flows/actions/flowsThunks.ts
```diff
@@ -0,0 +1,22 @@
...
+import {setNotebooks} from 'src/flows/actions/flowsActions'
+
+export const getNotebooks =
+  (orgID: string) => async (dispatch: Dispatch<any>) => {
+    try {
+      const response = await fetchNotebooks({query: {orgID}})
+
+      if (response.status !== 200) {
+        throw new Error()
+      }
```
hoorayimhelping, line 15: I know this an example of "do as I say, not as I do," with respect to creating an API layer that manages the communciation between the API and the application. My thinking here is that this is code that is being written so that it can be disappeared and that we don't need to build this with long-term maintenance and consistency in mind

src/shared/containers/GetOrganizations.tsx
```diff
@@ -92,6 +92,12 @@ const GetOrganizations: FunctionComponent = () => {
     }
   }, [account.type, account.isUpgradeable])
```
wdoconnell, line 94: Do we need a redux state for whether this API call has happened? I _suspect_ this is only going to make one API because this hook is going to run only when the component mounts and when org id is populated (and org id won't change until the app reload managed by quartz when switching orgs). But it might be safer to have another condition be that the status of have fetched the notebook array is also `RemoteDataState.NotStarted`.

Might be worth testing to see how many times this API call gets made in remocal. If it's more than once, I think we need a redux state for tracking the status of the API call.

hoorayimhelping: @wdoconnell I only saw it get called once, since `org.id` hasn't changed. I didn't test it against switching orgs, but I imagine it'll get called again in that instance.

wdoconnell: Gotcha. As long as it's only running once, that works.

wdoconnell (approved):

hoorayimhelping:
I'm going to merge this on Monday, since this got approved mid-afternoon on a Friday

src/types/resources.ts
```diff
@@ -90,6 +93,7 @@ export interface ResourceState {
   [ResourceType.Dashboards]: DashboardsState
   [ResourceType.Labels]: NormalizedState<Label>
   [ResourceType.Members]: NormalizedState<Member>
+  [ResourceType.Notebooks]: NotebooksState
```
wiedld, line 96: I'm confused on what we are doing on a higher level. Is the goal to:
(a) move the flows state management model into redux?
     or
(b) do existence checks on the presence of notebooks, to determine if the user has that feature depreciated or not?

Because if it's the latter...I'm not sure we want to populate resources for tracking of feature depreciation. It's a lot of stuff for a simple boolean response, which may be better consolidated in a `state.depreciation.<map>` with the boolean calculated on the single fetch (on load of org user).

Does this make sense?

wiedld (commented):
Missing context. Please clarify!

src/types/resources.ts
hoorayimhelping: @wiedld it's the latter. We're planning on deprecating notebooks, and the first part of that is hiding notebooks depending on rules listed in the [linked ticket.](https://github.com/influxdata/idpe/issues/16723)

My thinking here is that the response that comes back is small - it's essentially a date, an id, and a name. On prod, I fetched 4 notebooks and the total size was 1.37 kB, see attached screenshot. Since it's so small and trivial to store in memory, I figured that it might be better to just store the notebooks themselves in case they might be useful for the deprecation of this feature (maybe we need to include a nag that says "We're deprecating notebooks soon, and you have 13 notebooks. Please follow this link to start transferring them" or something like that).

If I was going to do it right, I'd create a new backend route that just returns a count of notebooks, but doing that would require making changes in at least two repos and I want to move quickly on this, since we have other unknown work to do like making our visualizations work with SQL without a lot of work.

Screen Shot 2022-12-19 at 10 02 27 AM

hoorayimhelping: Due to our time differences and me wanting to finish this up, I'm going to move forward with this. If you have strong opinions against it, we can follow up with a tweak to how this is done

wiedld: No worries. Glad you moved ahead and I didn't block it!

PR #6415, opened 2022-12-18
feat: add ui eventing for multi-org phase 2

### LINKED ISSUE #5707: Multi Org Phase 2 - UI Eventing
Add eventing for Phase 2 features. 

## UI Eventing Dictionary
https://influxdb.atlassian.net/wiki/spaces/IPO/pages/2530574337/UI+Eventing+Dictionary

blegesse-git:
Closes #5707 

Pr adds UI eventing for multi org phase 2 features 
All events introduced in this pr are documented [here](https://influxdb.atlassian.net/wiki/spaces/IPO/pages/2530574337/UI+Eventing+Dictionary). 

Events fire...
- [x] when new org is created
- [x] when new org failes to create 
- [x] when user clicks on `create organization` button of the header nav
- [x] when user switches to newly created org by clicking on `switch org` pop up
- [x] when user successfully deletes an org
- [x] when user submits Marketo account upgrade form
- [x] when user clicks on the contact sales link after Marketo upgrade account form fails to submit
- [x] when user clicks on the `upgrade` button found in the header nav drop down

### Checklist

### DIFF
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
@@ -260,7 +260,7 @@ export const CreateOrganizationOverlay: FC = () => {
             dispatch(notify(quartzOrgQuotaReached()))
           }
 
-          event(CreateOrgOverlayEvent.OrgCreationSuccess, multiOrgTag, {
+          event(CreateOrgOverlayEvent.OrgCreated, multiOrgTag, {
             newOrgID: newOrg.id,
             newOrgName: newOrg.name,
             oldOrgID: currentOrgId,
@@ -281,11 +281,6 @@ export const CreateOrganizationOverlay: FC = () => {
       } else {
         setNetworkErrorMsg(OrgOverlayNetworkError.DefaultError)
       }
-
-      event(CreateOrgOverlayEvent.OrgCreationFail, multiOrgTag, {
-        oldOrgID: currentOrgId,
-      })
-
       setCreateOrgButtonStatus(ComponentStatus.Error)
     }
   }
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
@@ -105,24 +105,13 @@ const switchToNewOrgButtonStyle = {
   cursor: 'pointer',
 }
 
-const SwitchToNewOrgButton = (
-  url: string,
-  newOrg: CreatedOrg,
-  currentOrgId
-): JSX.Element => {
-  const handleEventing = () => {
-    event(CreateOrgOverlayEvent.SwitchToNewOrg, multiOrgTag, {
-      currentOrgId,
-      createdOrgId: newOrg.id,
-      createdOrgName: newOrg.name,
-    })
-  }
+const SwitchToNewOrgButton = (url: string, handleClick: () => void): JSX.Element => {
   return (
     <a
       href={url}
       data-testid="go-to-new-org--link"
       style={switchToNewOrgButtonStyle}
-      onClick={handleEventing}
+      onClick={handleClick}
     >
       Switch to new org.
     </a>
@@ -224,13 +213,12 @@ export const CreateOrganizationOverlay: FC = () => {
     retrieveClusters()
   }, [])
 
-  // Event Handlers
-  const createNewOrgPopup = (newOrg: CreatedOrg) => {
+  const createNewOrgPopup = (newOrg: CreatedOrg, handleEventing: () => void) => {
     const newOrgUrl = `${CLOUD_URL}/orgs/${newOrg.id}`
 
     const switchToOrgLink =
       newOrg.provisioningStatus === 'provisioned'
-        ? () => SwitchToNewOrgButton(newOrgUrl, newOrg, currentOrgId)
+        ? () => SwitchToNewOrgButton(newOrgUrl, handleEventing)
         : null
 
     dispatch(notify(quartzOrgCreateSuccess(newOrg.name, switchToOrgLink)))
@@ -252,16 +240,24 @@ export const CreateOrganizationOverlay: FC = () => {
         orgCreationAllowed => {
           dispatch(getQuartzOrganizationsThunk(currentAccountId))
 
-          createNewOrgPopup(newOrg)
+          // Event Handlers
+          const handleSwitchToNewOrgEvent = () => {
+            event(CreateOrgOverlayEvent.SwitchToNewOrg, multiOrgTag, {
+              newOrgID: newOrg.id,
+              newOrgName: newOrg.name,
+              oldOrgID: currentOrgId,
+            })
+          }
+          createNewOrgPopup(newOrg, handleSwitchToNewOrgEvent)
 
           if (!orgCreationAllowed) {
             dispatch(notify(quartzOrgQuotaReached()))
           }
 
           event(CreateOrgOverlayEvent.OrgCreationSuccess, multiOrgTag, {
-            createdOrgId: newOrg.id,
-            createdOrgName: newOrg.name,
-            currentOrgId,
+            newOrgID: newOrg.id,
+            newOrgName: newOrg.name,
+            oldOrgID: currentOrgId,
           })
 
           history.push(`/orgs/${currentOrgId}/accounts/orglist`)
@@ -281,7 +277,7 @@ export const CreateOrganizationOverlay: FC = () => {
       }
 
       event(CreateOrgOverlayEvent.OrgCreationFail, multiOrgTag, {
-        currentOrgId,
+        oldOrgID: currentOrgId,
       })
 
       setCreateOrgButtonStatus(ComponentStatus.Error)
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
@@ -110,7 +110,6 @@ const SwitchToNewOrgButton = (
   newOrg: CreatedOrg,
   currentOrgId
 ): JSX.Element => {
-
   const handleEventing = () => {
     event(CreateOrgOverlayEvent.SwitchToNewOrg, multiOrgTag, {
       currentOrgId,
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
@@ -218,13 +218,13 @@ export const CreateOrganizationOverlay: FC = () => {
 
   const createNewOrgPopup = (
     newOrg: CreatedOrg,
-    handleEventing: () => void
+    handleSwitchToNewOrgEvent: () => void
   ) => {
     const newOrgUrl = `${CLOUD_URL}/orgs/${newOrg.id}`
 
     const switchToOrgLink =
       newOrg.provisioningStatus === 'provisioned'
-        ? () => SwitchToNewOrgButton(newOrgUrl, handleEventing)
+        ? () => SwitchToNewOrgButton(newOrgUrl, handleSwitchToNewOrgEvent)
         : null
 
     dispatch(notify(quartzOrgCreateSuccess(newOrg.name, switchToOrgLink)))
--- a/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
+++ b/src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
@@ -105,7 +105,10 @@ const switchToNewOrgButtonStyle = {
   cursor: 'pointer',
 }
 
-const SwitchToNewOrgButton = (url: string, handleClick: () => void): JSX.Element => {
+const SwitchToNewOrgButton = (
+  url: string,
+  handleClick: () => void
+): JSX.Element => {
   return (
     <a
       href={url}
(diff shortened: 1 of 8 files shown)


src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -92,18 +92,37 @@ import {getOrgCreationAllowancesThunk} from 'src/identity/allowances/actions/thu
...
   color: 'black',
   textDecoration: 'underline',
   cursor: 'pointer',
 }
 
-const SwitchToNewOrgButton = (url: string): JSX.Element => {
+const SwitchToNewOrgButton = (
+  url: string,
+  newOrg: CreatedOrg,
+  currentOrgId
```
hoorayimhelping, line 111: Let's talk about method design as it relates to method dependencies and interfaces.

Currently / before this change, the `SwitchToNewOrgButton` takes in a `url`, which makes sense - this component's main purpose is to render a button that navigates to a link, so being able to set the destination url of that link is completely reasonable.

The change we want is to be able to fire an event when the button is clicked. We have two options here - we can pass in all the things that the event needs to be able to fire correctly, which is the route you're currently taking. The event needs the current org id and the org that is being switched to. If we pass those in to this component, suddenly this component now has many responsibilities - it needs to be aware of not just the current and new org, but these concepts in general. Whereas this component used to be kind of dumb and only needed a url, it now needs system-level concepts - orgs.

This will work, but it leaks information about the system into a really dumb component, which causes two issues: it makes the component much more difficult to reuse (it can only be reused if the thing using it has access to current and new orgs), and it makes the intent of the component a little muddled.

An alternative to passing in current and new org is to pass in a callback function that is called when the button is clicked. So rather than pass in `newOrg` and `currentOrgId` to `SwitchToNewOrgButton`, if this code instead passed in `handleEventing`, which was constructed outside of this method, then `handleEventing` would abstract its dependencies on `newOrg` and `currentOrgId` away from the `SwitchToNewOrgButton`. This sidesteps the two issues mentioned above - now only `handleEventing` needs to worry about new and currente orgs, and `SwitchToNewOrgButton` is much more reusable, since a callback to be called `onClick` is much more generic, and can be a no-op.

Here's something along the lines that I was thinking:

```ts
const SwitchToNewOrgButton = (
  url: string,
  handleClick? = () => {}
) => {
  return (
    <a
      href={url}
      data-testid="go-to-new-org--link"
      style={switchToNewOrgButtonStyle}
      onClick={handleClick}
}
```

then when it's called:

```ts
const handleEventing = () => {
  event(CreateOrgOverlayEvent.SwitchToNewOrg, multiOrgTag, {
    currentOrgId,
    createdOrgId: newOrg.id,
    createdOrgName: newOrg.name,
  })
}
const switchToOrgLink =
  newOrg.provisioningStatus === 'provisioned'
    ? () => SwitchToNewOrgButton(newOrgUrl, handleEventing)
    : null
```

Now in the grand scheme of things, this button probably won't be reused very much, since it's defined in a very narrow context. But these principles are applicable to the design of methods, classes, components, etc.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -239,6 +258,12 @@ export const CreateOrganizationOverlay: FC = () => {
             dispatch(notify(quartzOrgQuotaReached()))
           }
 
+          event(CreateOrgOverlayEvent.OrgCreationSuccess, multiOrgTag, {
+            createdOrgId: newOrg.id,
+            createdOrgName: newOrg.name,
+            currentOrgId,
+          })
```
hoorayimhelping, line 265: This method being called again and being essentially the same as `handleEventing` is another clue that maybe the abraction of that method is in the wrong place. If you take my suggestion above, this code can be changed to just `handleEventing()` (or whatever the name of that method becomes). Which to me is an indicator that maybe the abstractions should be moved around.

In my experience, the wrong abstraction leads to hard abstractions to create, while the right abstraction makes things a lot easier - if we can move the abstraction around and in doing so remove some duplicated code while making the method a little simpler, that is usually a good indicator that we're on the right track. This is a hueristic I use when trying to design software systems.

hoorayimhelping (commented):
Some thoughts on method design and abstraction. Let me know what you think

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
blegesse-git: your reasoning makes perfect sense. It's a simple change but it does a good job clearing up the blurriness of "labour division" that currently exists. Great suggestion. I'll update accordingly

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -92,18 +92,37 @@ import {getOrgCreationAllowancesThunk} from 'src/identity/allowances/actions/thu
...
+  newOrg: CreatedOrg,
+  currentOrgId
+): JSX.Element => {
+  const handleEventing = () => {
+    event(CreateOrgOverlayEvent.SwitchToNewOrg, multiOrgTag, {
+      currentOrgId,
+      createdOrgId: newOrg.id,
+      createdOrgName: newOrg.name,
+    })
+  }
```
wdoconnell, line 119: I thought about the names for these keys a bit more: I think we should mimic the names of the keys that are used for switching orgs in the other location that occurs in the app (the Org Dropdown in the header). That way, when analytics is trying to run the SQL query for these events, they can use the same column name for both searches

eg

`SELECT OLD_ORG_ID FROM [Table] WHERE OLD_ORG_ID= . . . `

Instead of having to call it `CURRENT_ORG_ID` in one SQL query and `OLD_ORG_ID` in another.

Here are the key names we're already using for switch org in OrgDropdown.tsx:

```
  const switchOrg = (org: TypeAheadMenuItem) => {
    event(HeaderNavEvent.OrgSwitch, multiOrgTag, {
      oldOrgID: activeOrg.id,
      oldOrgName: activeOrg.name,
      newOrgID: org.id,
      newOrgName: org.name,
    })
    window.location.href = `${CLOUD_URL}/orgs/${org.id}`
  }
```

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
wdoconnell: This sounds good to me - but here I would implement the same comment I mentioned above re: we should have the same names for the keys of the object literal sent by the event - 'old' versus 'current' for org id/name - whether the event is coming from the Create Organization overlay or elsewhere in the app. That'll make for less confusion on the analytics side having to tailor SQL queries differently for different events.

I would use the same key names as the old events (rather than changing the old events' key names) to avoid disturbing the existing data that already exists re: usage of the switch-org components.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
```diff
@@ -254,6 +279,11 @@ export const CreateOrganizationOverlay: FC = () => {
       } else {
         setNetworkErrorMsg(OrgOverlayNetworkError.DefaultError)
       }
+
+      event(CreateOrgOverlayEvent.OrgCreationFail, multiOrgTag, {
+        currentOrgId,
+      })
+
```
wdoconnell, line 286: same comment re 'current' vs. 'old'.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/MenuItem.tsx
```diff
@@ -1,20 +1,40 @@
...
+  const user = useSelector(selectUser)
+  const accountId = useSelector(selectCurrentAccountId)
+  const accountType = useSelector(selectCurrentAccountType)
+
   const handleCreateOrg = () => {
+    event(HeaderNavEvent.CreateOrgClick, multiOrgTag, {
+      userId: user.id,
+      accountId,
+      accountType,
+    })
```
wdoconnell, line 37: For this event and the next one, I might add `accountName` here, which will be easier for analytics if they're browsing the data later. That way they don't need to join SQL tables to see 'oh, this account refers to this company'.

src/identity/components/GlobalHeader/OrgDropdown.tsx
```diff
@@ -56,10 +59,18 @@ export const OrgDropdown: FC<Props> = ({activeOrg, orgsList}) => {
...
+  const accountType = useSelector(selectCurrentAccountType)
 
   const dispatch = useDispatch()
 
   const openMarketoOverlay = () => {
+    event(HeaderNavEvent.UpgradeButtonClick, multiOrgTag, {
+      userId: user.id,
+      accountId,
+      accountType,
+    })
```
wdoconnell, line 73: Same comment here re: account name.

src/identity/components/MarketoAccountUpgradeOverlay.tsx
```diff
@@ -70,13 +81,26 @@ interface MarketoFormElement extends Element {
 }
 
 // If marketo isn't working, still need the user to have some means of contacting sales.
-const SalesFormLink = (): JSX.Element => {
+const SalesFormLink = () => {
+  const user = useSelector(selectUser)
+  const accountId = useSelector(selectCurrentAccountId)
+  const accountType = useSelector(selectCurrentAccountType)
+
```
wdoconnell, line 88: Nice.

src/identity/components/MarketoAccountUpgradeOverlay.tsx
```diff
@@ -137,9 +161,19 @@ export const MarketoAccountUpgradeOverlay: FC = () => {
...
+      })
       onClose()
     } catch (err) {
       handleError(MarketoError.FormSubmitError, err)
+      event(AccountUpgradeOverlay.MarketoAccountUpgradeFail, multiOrgTag, {
+        userId: user.id,
+        accountId,
+        accountType,
+      })
     }
```
wdoconnell, line 177: I actually don't think we need the data in the `catch` evented - an upgrade failing isn't an event we can generate user behavior data from, so much as an indicator of a problem in prod that we'd need to catch and resolve. The existing handler sends that data to honeybadger, which I think is all we need to do here.

If you remove the event from the catch, I think we could just send the success event to analytics as `MarketoAccountUpgrade`, which I suspect is the data that would be useful - how many users are upgrading accounts; when are they doing that; etc.

src/identity/components/OrganizationListTab/OrgBannerPanel.tsx
```diff
@@ -28,8 +35,14 @@ export const OrgBannerPanel: FC<OrgAllowance> = ({
 }) => {
   const dispatch = useDispatch()
   const history = useHistory()
+  const accountId = useSelector(selectCurrentAccountId)
 
   const handleUpgradeAccount = () => {
+    event(OrgListEvent.UpgradeAccount, multiOrgTag, {
+      availableUpgrade,
+      accountId,
+    })
```
wdoconnell, line 44: Same comment here re: adding accountName.

src/identity/events/multiOrgEvents.ts
```diff
@@ -12,6 +12,30 @@ export enum HeaderNavEvent {
...
   UserLogoutClick = 'headerNav.userAvatarLogOut.clicked',
+
+  CreateOrgClick = 'headerNav.createOrg.clicked',
+  UpgradeButtonClick = 'headerNav.upgradeButton.clicked',
+}
+
+export enum CreateOrgOverlayEvent {
+  OrgCreationSuccess = 'headerNav.createOrgOverlay.creationSuccess',
+  OrgCreationFail = 'headerNav.createOrgOverlay.creationFail',
+  SwitchToNewOrg = 'headerNav.createOrg.switchedToNewOrg',
```
wdoconnell, line 23: I think this one needs to be `headerNav.createOrgOverlay.switchedToNewOrg`

src/organizations/components/OrgProfileTab/DeletePaidOrgOverlay.tsx
```diff
@@ -115,6 +125,12 @@ export const DeletePaidOrgOverlay: FC = () => {
       .catch(err => {
         clearTimeout(responseTimer)
         dispatch(notify(deleteOrgFailed(SupportLink, org.name)))
+        event(DeleteOrgOverlay.DeleteOrgFail, multiOrgTag, {
+          userId: org.id,
+          orgName: org.name,
+          accountType: account.type,
+          accountId: account.id,
+        })
```
wdoconnell, line 133: Same comment here - I'm not sure we need to event org deletion failure, as we're already catching it through honeybadger, and the use of this information isn't analytics - how many org deletions failed - but rather, identifying a problem with the UI that needs to be resolved.

For the same reason we don't need to call the event an OrgDeleteSuccess - it can just be [eventName].OrgDeleted, or similar.

wdoconnell (changes requested):
Looking good, and thank you for handling! 

I had a couple suggestions throughout about the keys in the events we're sending to analytics; and suspect there are two events we should remove.

src/identity/events/multiOrgEvents.ts
wdoconnell: I might also consider alphabetizing the enums in this file, now that it's growing larger.

src/identity/components/GlobalHeader/GlobalHeaderDropdown/CreateOrganization/CreateOrganizationOverlay.tsx
blegesse-git: since this is a case where no new org has been created, doesn't it make more sense to name it `current` ? Calling it old implies there is a new org created. Personally, I think currentOrgID or just orgID is a more accurate and descriptive name than `oldOrgID`. What do you think?

src/identity/components/MarketoAccountUpgradeOverlay.tsx
blegesse-git: ah that's a fair point. I'll remove it

hoorayimhelping (dismissed):

wdoconnell (approved):
Awesome work, thanks for accommodating my nits!

PR #6425, opened 2022-12-20
feat(aim): Hide Cardinality for iOX Orgs

abshierjoel:
Removes the cardinality limit from the UI for iOX-enabled organizations. This will require a new SHA pending influxdata/openapi#625.

image

### Checklist

### DIFF
--- a/src/operator/OrgOverlay.tsx
+++ b/src/operator/OrgOverlay.tsx
@@ -57,7 +57,7 @@ const OrgOverlay: FC = () => {
   const history = useHistory()
   const canReactivateOrg =
     hasWritePermissions && organization?.state === 'suspended'
-  const isIOx = organization?.storageType == 'iox'
+  const isIOx = organization?.storageType.toLowerCase() === 'iox'
   const canSeeCardinalityLimits = !isIOx
 
   useEffect(() => {
--- a/src/operator/OrgOverlay.tsx
+++ b/src/operator/OrgOverlay.tsx
@@ -57,7 +57,8 @@ const OrgOverlay: FC = () => {
   const history = useHistory()
   const canReactivateOrg =
     hasWritePermissions && organization?.state === 'suspended'
-  const canSeeCardinalityLimits = organization?.storageType == 'tsm'
+  const isIOx = organization?.storageType == 'iox'
+  const canSeeCardinalityLimits = !isIOx
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/operator/OrgOverlay.tsx
```diff
@@ -57,6 +57,7 @@ const OrgOverlay: FC = () => {
   const history = useHistory()
   const canReactivateOrg =
     hasWritePermissions && organization?.state === 'suspended'
+  const canSeeCardinalityLimits = organization?.storageType == 'tsm'
```
hoorayimhelping, line 60: I think we might want to invert this conditional so that we greedily show cardinality limits. From what I understand, every org _should_ have a `storageType` associated with it, but there might have been some orgs that fell through the cracks, and if they are tsm orgs without a `storageType`, this won't show cardinality limits properly. Probably a longshot edge case, but worth considering.

Example of what I'm rambling about:
```ts
import {isOrgIOx} from 'src/organizations/selectors'
// ...

const isIOx = useSelector(isOrgIOx)
const canSeeCardinalityLimits = !isIOx
```

hoorayimhelping (approved):
LGTM, but I did provide a suggestion. Let me know if you have any strong opinions on it!

src/operator/OrgOverlay.tsx
```diff
@@ -57,6 +57,8 @@ const OrgOverlay: FC = () => {
   const history = useHistory()
   const canReactivateOrg =
     hasWritePermissions && organization?.state === 'suspended'
+  const isIOx = organization?.storageType == 'iox'
```
hoorayimhelping, line 60: for safety, let's convert `storageType` to lowercase before doing the comparison

```ts
organization?.storageType.toLowerCase() === 'iox'
```

note the `===` vs `==`, which forces the check to use [strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality) which won't do any type-casting. Not a huge deal when comparing strings, but it's more about the fact comparing using normal equality is often done explicitly to loosen the type checks. So in this case, it'll still do a string comparison just fine; the issue is that doing a non-strict check might make someone stop and see if there's a special reason not to do strict equality checking.

hoorayimhelping (approved):
>10 lines of code = 10 issues.

>500 lines of code = "looks fine."

>Code reviews.

Issue #6426, opened 2022-12-20
Notify user that they cannot delete last org in an account


wdoconnell:
Quartz will prevent the user from deleting the last organization in an account. 

On the UI side, we need to notify multi-org users that they won't be able to go through the 'delete org' prompts when an org is the last org in the account.

wdoconnell:
Completed by #6427

PR #6427, opened 2022-12-20
feat: update org profile to use new 'isSuspendable' logic

### LINKED ISSUE #6419: Eliminate caching of current org data from org/settings tab
The org/settings tab uses the data from /api/v2/quartz/orgs/:orgId to update the user's identity/org redux state if the data required by that tab is not presently available in redux. The data is _not_ reloaded if it has been fetched from this endpoint already during the current session, and is therefore already in redux state.

This behavior isn't desirable prospectively for multi-org - we want the most up-to-date information about the user's current org, including whether it's `suspendable` or not - so navigating to and from this tab should force another request to the api/v2/quartz/orgs/:orgId endpoint. The trigger for making this request should live in local state, not redux going forward.

wdoconnell:
Closes #6419 #6426

This PR prevents users from deleting the last organization in any account with multiple orgs. Instead, on clicking 'delete', such a user receives a popup that they can't delete the last org in an account, instead of proceeding to the overlay that asks the user to accept certain terms before deleting an org.

To facilitate this, the org settings page will now issue an api call checking the user's `orgDetails` _every_ time that tab is loaded - not just the first time it's loaded in the current session. This is to minimize the chance of unexpected behavior if a second user is deleting orgs in the account during the first user's session.

Org is Not Suspendable (Updated)
--
Now gives the user a clickable link to create a new org (or upgrade the account, if, for example, they're only able to see one org in a 3-org account) if that's the reason they can't delete the current one.

https://user-images.githubusercontent.com/91283923/208964055-b7aa667c-5e51-418a-b8ff-6788b89dcd04.mov

https://user-images.githubusercontent.com/91283923/208964066-13126943-c9bd-48c1-8485-14498d8c4863.mov

Org is Suspendable
--

https://user-images.githubusercontent.com/91283923/208756800-689dcff1-90f1-491a-b399-a1ab8d7e7249.mov

Bad Response from API (Don't Show Delete Panel)
--

https://user-images.githubusercontent.com/91283923/208758272-641554ee-3108-4c22-bcaa-462375f9a5ab.mov

### Checklist

- [X] Feature flagged, if applicable - partially behind `createDeleteOrgs`


src/identity/actions/thunks/index.ts
```diff
@@ -81,12 +84,18 @@ export const getCurrentOrgDetailsThunk =
...
+      const state = getState()
+      const identity = selectCurrentIdentity(state)
+
       reportErrorThroughHoneyBadger(err, {
         name: 'Failed to fetch /quartz/orgs/:orgId',
-        context: {state: getState()},
+        context: {identity},
       })
+      throw new Error(err)
     }
```
wdoconnell, line 100: Modifies this thunk to throw an error to the UI that can be caught to deliver a notification.

Reduced the amount of data we're sending on honeybadger errors because it's oversize and unnecessary. The only information we really need to debug an api failure here is the user's identity.

src/identity/selectors/index.ts
```diff
@@ -78,11 +84,6 @@ export const selectQuartzIdentityStatus = (state: AppState): RemoteDataState =>
 export const selectQuartzBillingStatus = (state: AppState): RemoteDataState =>
   state.identity.currentIdentity.loadingStatus.billingStatus
 
-export const selectQuartzOrgDetailsStatus = (
-  state: AppState
-): RemoteDataState =>
-  state.identity.currentIdentity.loadingStatus.orgDetailsStatus
-
 export const selectQuartzOrgs = (
```
wdoconnell, line 87: We don't need this redux status selector anymore - we are using local state to determine when an API call should be made, because we want it to be run every time the org settings tab is loaded, even if the user navigates back later in the same session.

src/organizations/components/OrgProfileTab/DeletePanel.tsx
```diff
@@ -1,44 +1,81 @@
 // Libraries
-import React, {FC, useContext} from 'react'
```
wdoconnell, line 2: This file was getting too large and confusing. Instead, I refactored it into three components:

- DeletePanel: either shows the Delete Account button in a free account with one org; or Delete Org in any account with multiple orgs.
- Delete Account button
- Delete Org button

src/organizations/components/OrgProfileTab/SuspendPaidOrgOverlay.tsx
```diff
@@ -68,7 +68,7 @@ const SupportLink = (): JSX.Element => {
   )
 }
```
wdoconnell, line 70: Better to call this the org 'suspension' overlay, as the user can't immediately delete orgs through this.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -1,64 +1,74 @@
...
-  const orgDetailsStatus = useSelector(selectQuartzOrgDetailsStatus)
-
+  const quartzOrg = useSelector(selectCurrentOrg)
   const dispatch = useDispatch()
 
-  const identityOrgId = quartzOrg.id
+  // Data about the user's organization is intentionally re-fetched when this component mounts again.
+  const [orgDetailsStatus, setOrgDetailsStatus] = useState(
+    RemoteDataState.NotStarted
+  )
```
wdoconnell, line 50: We now use local state to determine whether we make the API call to get the user's org details, not redux.

The data itself still lives in redux. In theory we could tear down portions of the reducer and put a few of the `org` properties in local state in this component now. However, that's confusing (because half of the org properties will live in redux, and half won't), and wasteful because there will likely be other pages where we need to use this data later.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -72,33 +82,25 @@ const OrgProfileTab: FC = () => {
...
-            {orgProviderExists && (
-              <LabeledData label="Cloud Provider" src={quartzOrg.provider} />
-            )}
-            {quartzOrg.regionCode && (
-              <LabeledData label="Region" src={quartzOrg.regionCode} />
-            )}
-            {quartzOrg.regionName && (
-              <LabeledData label="Location" src={quartzOrg.regionName} />
-            )}
+            <LabeledData label="Cloud Provider" src={quartzOrg.provider} />
```
wdoconnell, line 95: All of these checks are extra verbiage. These properties are required, so if the API call comes back we don't need to do individualized guards.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -1,64 +1,74 @@
...
     if (orgDetailsStatus === RemoteDataState.NotStarted) {
-      dispatch(getCurrentOrgDetailsThunk(identityOrgId))
+      Promise.resolve(dispatch(getCurrentOrgDetailsThunk(quartzOrg.id)))
+        .then(() => {
+          setOrgDetailsStatus(RemoteDataState.Done)
+        })
+        .catch(() => {
+          setOrgDetailsStatus(RemoteDataState.Error)
+          dispatch(notify(orgDetailsFetchError()))
+        })
```
hoorayimhelping, line 67: I'm a little confused about this usage of `Promise.resolve`. Typically `Promise.resolve` is used to convert an existing value into a promise that is then returned from a method that needs to be async. What's the intent here? Is it to handle an error thrown from `getCurrentOrgDetailsThunk` ?

I think `dispatch` is a synchronous function, so can you do somethign like

```ts
try {
  dispatch(getCurrentOrgDetailsThunk(quartzOrg.id))
  setOrgDetailsStatus(RemoteDataState.Done)
} catch {
  setOrgDetailsStatus(RemoteDataState.Error)
  dispatch(notify(orgDetailsFetchError()))
}
```

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -1,64 +1,74 @@
...
     }
-  }, [dispatch, orgDetailsStatus, identityOrgId])
-
-  const hasIdentityData =
-    quartzOrg.provider || quartzOrg.regionCode || quartzOrg.regionName
+  }, [dispatch, orgDetailsStatus, quartzOrg.id])
 
-  const orgProviderExists = !!quartzOrg.provider
+  const fetchedOrgDetails = orgDetailsStatus === RemoteDataState.Done
```
hoorayimhelping, line 71:
```suggestion
  const hasFetchedOrgDetails = orgDetailsStatus === RemoteDataState.Done
```

to avoid ambiguity between "[this component has] fetched and org details" and "[these are the previously] fetched org details"

hoorayimhelping (commented):
LGTM, just a couple of comments

src/organizations/components/OrgProfileTab/DeletePanel.tsx
```diff
@@ -61,64 +98,90 @@ const DeletePanel: FC = () => {
...
+        <>
+          {!orgCanBeSuspended && (
+            <Popover
+              contents={onHide => (
+                <div>
+                  <>
+                    <Popover.DismissButton onClick={onHide} />
+                    <b>"{org.name}"</b> cannot be deleted because it is the last
+                    organization in this account. <br />
+                    Before deleting it, another organization must be created.
```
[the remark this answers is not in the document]
wdoconnell: Nice - I like it.

src/organizations/components/OrgProfileTab/index.tsx
wdoconnell: The issue here is that, because dispatch is synchronous but the thunk is asynchronous, that code will never hit the catch block whether there is an error or not (which is obnoxious). 😠 

On testing, the way this works is: the thunk is dispatched, `setOrgDetailsStatus` is set to `RemoteDataState.Done`, we exit the `try/catch` block, and then the thunk (at a later point, whenever we get a bad response from the API) will throw an uncaught error. So, we either need a promise chain or an await, and Typescript doesn't accept `await` as valid since `dispatch` doesn't return a promise.

I agree with you it's better to use a `try/catch` here. So here's a solution that hews a little closer to your suggestion - we just need to add an `await`, and also a `Promise.resolve` to convert the return value from the dispatch into a function.

```
  useEffect(() => {
    if (!CLOUD) {
      return
    }

    const retrieveOrgDetails = async () => {
      if (orgDetailsStatus === RemoteDataState.NotStarted) {
        try {
          await Promise.resolve(
            dispatch(getCurrentOrgDetailsThunk(quartzOrg.id))
          )
          setOrgDetailsStatus(RemoteDataState.Done)
        } catch (err) {
          setOrgDetailsStatus(RemoteDataState.Error)
          dispatch(notify(orgDetailsFetchError()))
        }
      }
    }

    retrieveOrgDetails()
  }, [dispatch, orgDetailsStatus, quartzOrg.id])
```

src/organizations/components/OrgProfileTab/index.tsx
wdoconnell: 👍

src/organizations/components/OrgProfileTab/DeletePanel.tsx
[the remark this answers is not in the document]
wdoconnell: Tested this, and this is a dope addition. Great suggestion.

wdoconnell:
Rebased against current master.

wdoconnell:
@hoorayimhelping : fyi, the 'new' part, which includes the link in the popup to direct the user to create a new org (or upgrade their account), begins at this function, which is referenced in the `Popup`: https://github.com/influxdata/ui/blob/e9d17d99050b38ef928925d168998e61bd1663d3/src/organizations/components/OrgProfileTab/DeletePanel.tsx#L141

hoorayimhelping (approved):

PR #6428, opened 2022-12-20
fix: removes the ability to remove user from account when running in …

djcarpe:
@gunnaraasen 

@wdoconnell  @eatondustin1 and I just brainstormed on this - we realized this button won't make sense anymore once multi-org creation is enabled because removing a user from an org won't remove them from the entire account. So this will automatically hide the button once multi-org creation is on. Steps will be:

1. This PR - hide button that will be confusing post multi org
2. At some point, we should consider moving this button over to the org/settings page
3. We need to add a button to account/settings page at some point, but that will require backend changes to make removing a user from an account possible

There is no remove from account endpoint for this logic yet and want to remove the ability from accidentally removing a user from a single org in an account.

### Checklist

### DIFF
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -103,6 +103,7 @@ const AccountAboutPage: FC = () => {
   }
 
   const showDeactivateAccountSection = isFlagEnabled('freeAccountCancellation')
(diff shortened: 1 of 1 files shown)


eatondustin1 (approved):
:shipit:

wdoconnell (approved):

PR #6437, opened 2022-12-28
feat(OrgProfileTab): add Leave Org button in Org settings page

### LINKED ISSUE #6434: Convert 'Leave account' button on the account settings page to a 'leave org' button on the org settings page
Pre-multi org, the 'Leave Account' button on the account/settings page would remove the user from the current org (because of the 1:1 relationship between orgs and accounts). 

That button now belongs more appropriately on the _org_ settings page, since it's possible to leave one org in account but not leave others.

For now, the button is hidden when multiOrg is on, but more permanently should just be migrated over to the org settings page.

blegesse-git:
Closes #6434 

Pr introduces change in Orgs settings page by adding `Leave Organization` button (behind `createDeleteOrgs` feature flag. 
When flag is off, The button is displayed only inside Account's settings page. 

Orgs Settings page when `createDeleteOrgs` feature flag is on (updated image: Leave org comes before Delete)

Screen Shot 2022-12-30 at 12 10 01 PM

Orgs Settings page when `createDeleteOrgs` feature flag is off

https://user-images.githubusercontent.com/66275100/210096412-1f8a19fc-d5a2-43f1-9ef9-9f2d725e93dd.mov

Accounts Settings page when `createDeleteOrgs` feature flag is on 

https://user-images.githubusercontent.com/66275100/210096505-749efe39-b850-4a24-bbe4-eb1131d73884.mov

Accounts Settings page when `createDeleteOrgs` feature flag is off

https://user-images.githubusercontent.com/66275100/210096584-415dc095-173f-40af-95cf-84856206d10d.mov

### Checklist

### DIFF
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -1,13 +1,9 @@
 // Libraries
 import React, {ChangeEvent, FC, useContext, useEffect, useState} from 'react'
-import {useSelector} from 'react-redux'
 
 import {
   Button,
-  ButtonShape,
-  ComponentColor,
   ComponentSize,
-  ConfirmationButton,
   FlexBox,
   FlexDirection,
   Input,
@@ -15,9 +11,6 @@ import {
   Overlay,
   Page,
 } from '@influxdata/clockface'
-
-import {getMe} from 'src/me/selectors'
-import {UsersContext, UsersProvider} from 'src/users/context/users'
 
 // Utils
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
@@ -31,19 +24,10 @@ import CancelServiceProvider from 'src/billing/components/PayAsYouGo/CancelServi
 
 // Styles
 import './AccountPageStyles.scss'
-import {CLOUD_URL} from 'src/shared/constants'
-
-const leaveBtnStyle = {
-  width: 'auto',
-  marginTop: 32,
-  paddingLeft: '8px',
-  paddingRight: '8px',
-}
 
 const AccountAboutPage: FC = () => {
   const {userAccounts, handleRenameActiveAccount} =
     useContext(UserAccountContext)
-  const {users, handleRemoveUser} = useContext(UsersContext)
 
   const [isDeactivateAccountVisible, setDeactivateAccountVisible] =
     useState(false)
@@ -60,30 +44,6 @@ const AccountAboutPage: FC = () => {
   const updateAcctName = (evt: ChangeEvent<HTMLInputElement>) => {
     setActiveAcctName(evt.target.value)
   }
-
-  const currentUserId = useSelector(getMe)?.id
-
-  const handleRemove = () => {
-    handleRemoveUser(currentUserId)
-    window.location.href = CLOUD_URL
-  }
-
-  const allowSelfRemoval = users.length > 1
-
-  const leaveAcctBtn = (
-    <ConfirmationButton
-      confirmationLabel="This action will remove yourself from accessing this organization"
-      confirmationButtonText="Leave Account"
-      titleText="Leave Account"
-      text="Leave Account"
-      confirmationButtonColor={ComponentColor.Danger}
-      color={ComponentColor.Default}
-      shape={ButtonShape.Square}
-      onConfirm={handleRemove}
-      testID="delete-user"
-      style={leaveBtnStyle}
-    />
-  )
 
   const inputStyle = {width: 250}
   const labelStyle = {marginBottom: 8, maxWidth: '500px'}
@@ -103,7 +63,6 @@ const AccountAboutPage: FC = () => {
   }
 
   const showDeactivateAccountSection = isFlagEnabled('freeAccountCancellation')
-  const showLeaveAcctBtn = !isFlagEnabled('createDeleteOrgs')
 
   return (
     <AccountTabContainer activeTab="settings">
@@ -131,7 +90,6 @@ const AccountAboutPage: FC = () => {
             text="Save"
           />
         </FlexBox>
-        {allowSelfRemoval && showLeaveAcctBtn && leaveAcctBtn}
         {showDeactivateAccountSection && (
           <>
             <hr style={dividerStyle} />
@@ -169,9 +127,7 @@ const AccountPage: FC = () => {
   return (
     <Page titleTag={pageTitleSuffixer(['Account Settings Page'])}>
       <AccountHeader testID="account-page--header" />
-      <UsersProvider>
-        <AccountAboutPage />
-      </UsersProvider>
+      <AccountAboutPage />
     </Page>
   )
 }
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -69,7 +69,7 @@ const AccountAboutPage: FC = () => {
   }
 
   const showDeactivateAccountSection = isFlagEnabled('freeAccountCancellation')
-  const showLeaveOrgtButton = !isFlagEnabled('createDeleteOrgs')
+  const showLeaveOrgButton = !isFlagEnabled('createDeleteOrgs')
   const allowSelfRemoval = users.length > 1
 
   return (
@@ -98,7 +98,7 @@ const AccountAboutPage: FC = () => {
             text="Save"
           />
         </FlexBox>
-        {allowSelfRemoval && showLeaveOrgtButton && (
+        {allowSelfRemoval && showLeaveOrgButton && (
           <>
             <hr style={dividerStyle} />
             <LeaveOrgButton />
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -11,6 +11,10 @@ import {
   Overlay,
   Page,
 } from '@influxdata/clockface'
+import {LeaveOrgButton} from 'src/organizations/components/OrgProfileTab/LeaveOrg'
+
+// Context
+import {UsersContext, UsersProvider} from 'src/users/context/users'
 
 // Utils
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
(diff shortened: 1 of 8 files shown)


src/organizations/components/OrgProfileTab/style.scss
```diff
@@ -19,8 +19,15 @@
...
   flex-shrink: 0;
+  flex-wrap: wrap;
 }
 
 .org-profile-tab--copyableText {
   min-width: 500px;
 }
+
+.org-profile-tab--leaveOrgButton {
+  width: auto !important;
```
blegesse-git, line 30: needed to override cf style

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -132,6 +136,44 @@ const OrgProfileTab: FC = () => {
...
+      window.location.href = CLOUD_URL
+    }
+
+    return (
+      <>
+        {allowSelfRemoval && (
+          <FlexBox.Child>
+            <h4>Leave Organization</h4>
+            <p className="org-profile-tab--heading org-profile-tab--deleteHeading">
+              Leave the <b>{org.name}</b> organization.
```
blegesse-git, line 156: thoughts on text under the `Leave Org` title?

src/accounts/AccountPage.tsx
```diff
@@ -103,7 +63,6 @@ const AccountAboutPage: FC = () => {
   }
 
   const showDeactivateAccountSection = isFlagEnabled('freeAccountCancellation')
-  const showLeaveAcctBtn = !isFlagEnabled('createDeleteOrgs')
```
blegesse-git, line 106: TODO: check if this feature needs to remain behind a feature flag

src/organizations/components/OrgProfileTab/index.tsx
wdoconnell: Sounds good to me.

src/accounts/AccountPage.tsx
blegesse-git: if (`createDeleteOrgs`) Leave Account becomes Leave Org and gets moved to Orgs setting page 
if (!`createDeleteOrgs`) Leave Account remains in Accounts settings page

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -132,6 +136,44 @@ const OrgProfileTab: FC = () => {
...
+  const LeaveOrgButton = () => {
+    const currentUserId = me.id
+    const {users, handleRemoveUser} = useContext(UsersContext)
+
+    const allowSelfRemoval = users.length > 1
+
+    const handleRemove = () => {
+      handleRemoveUser(currentUserId)
+      window.location.href = CLOUD_URL
+    }
```
hoorayimhelping, line 148: I know this is copied over from the other file, so don't approach this as a critisism but more as a learning opportunity or chance at gaining perspective. Anyway this method name is backwards and kind of poorly thought out, when you think about it.

Events are handled, whereas actions are performed. The application isn't handling the removal of the user. The application is handling the click event that is generated by the browser when the user clicks the button that says `Leave Organization`.

The action that is performed when then the user clicks `Leave Organization` is to remove the user. The method in the context name is also kind of dumb, because it shows the author's metnal model that has a the tight coupling of the event and the action performed. `removeUser` is a much better name beause it describes what is happening simply - the user is removed. `removeUser` is appropriate to use in response to an event (in this case clicking on the `Leave Organization` button), or if a user needs to be removed by the system as part of some other process that doesn't involve user interaction. Handling the removal of the user sounds like there's an event that is fired by the system called `RemoveUserEvent` and we need to handle it.

The times when it's appropriate to use a method named `handleSomething` are typically when handling an event is a complex process that involves multiple steps or multiple responsibilities. For example, if this component needed to call the method to remove the user and then also call a method to send an email, it might be more appropriate to call this handler `handleClick`, since it's a more generic handler. But in general, I think it's still probably better to encapsulate all the logic to remove a user into a method called `removeUser`, since part of removing the user would also involve sending an email.

It's a small thing from a naming perspective, but a huge thing from an overall system design perspective.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -132,6 +136,44 @@ const OrgProfileTab: FC = () => {
     </FlexBox.Child>
   )
 
+  const LeaveOrgButton = () => {
+    const currentUserId = me.id
+    const {users, handleRemoveUser} = useContext(UsersContext)
+
+    const allowSelfRemoval = users.length > 1
```
hoorayimhelping, line 143: this is nice :+1:

hoorayimhelping (approved):
LGTM! Obviously since there's just some shuffling of code around, I felt compelled to leave you a 400 word essay about software design ;)

src/organizations/components/OrgProfileTab/LeaveOrg.tsx
```diff
@@ -0,0 +1,60 @@
...
+          confirmationButtonText="Leave Organization"
+          titleText="Leave Organization"
+          text="Leave Organization"
+          confirmationButtonColor={ComponentColor.Danger}
+          color={ComponentColor.Default}
+          shape={ButtonShape.Square}
+          onConfirm={handleRemove}
+          testID="delete-user"
+          icon={IconFont.Logout}
+        />
```
wdoconnell, line 56: I know these were in here before, but might want to reorder these props.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -149,11 +156,18 @@ const OrgProfileTab: FC = () => {
       </FlexBox>
 
       {CLOUD && orgDetailsLoaded && (
-        <FlexBox.Child className="org-profile-tab--section">
+        <FlexBox
+          direction={FlexDirection.Row}
+          stretchToFitWidth={true}
+          className="org-profile-tab--section"
```
wdoconnell, line 162: Same here - might reorder the props by alpha.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -149,11 +156,18 @@ const OrgProfileTab: FC = () => {
...
+        >
           <UsersProvider>
-            <DeletePanel />
+            <>
+              <DeletePanel />
+              {allowSelfRemoval && showLeaveOrgButton && <LeaveOrgButton />}
+            </>
```
wdoconnell, line 168: I think I would put the "Leave" org button above delete, rather than putting it below Delete, since "Deleting" the org is a lot more final than leaving it.

src/organizations/components/OrgProfileTab/LeaveOrg.tsx
```diff
@@ -0,0 +1,60 @@
...
+// Constants
+import {CLOUD_URL} from 'src/shared/constants'
+
+// Styles
+import 'src/organizations/components/OrgProfileTab/style.scss'
+
+export const LeaveOrgButton: FC = () => {
+  const org = useSelector(getOrg)
+  const currentUserId = useSelector(getMe)?.id
+  const {handleRemoveUser} = useContext(UsersContext)
```
wdoconnell, line 30: Instead of using the idpe selectors here, you can use the identity selectors here, because this component only lives in CLOUD. One day it'd be nice to get rid of the idpe ones in cloud, and the fewer we have to remove at a later date, the better.

for org you can use `selectCurrentOrg`, and for the user id you can use `selectUser` (and destructure the `id` property).

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -73,6 +78,8 @@ const OrgProfileTab: FC = () => {
     retrieveOrgDetails()
   }, [dispatch, orgDetailsStatus, quartzOrg.id])
 
+  const allowSelfRemoval = users.length > 1
+  const showLeaveOrgButton = isFlagEnabled('createDeleteOrgs')
```
wdoconnell, line 82: I spent some time thinking about whether this check - `allowSelfRemoval` - is sufficient for multi-org accounts. I think it is.

We have additional checks like `isSuspendable` to prevent whether a user can _delete_ the last org in the account. That's necessary because we don't want an orphaned account without any orgs in it.

I don't think we have that problem here as long as we check if `users.length > 1` - if a user tries to remove themselves from an org, even if it's the last org, that check will require the presence of another user. So, barring unusual circumstances like two users removing themselves simultaneously, it shouldn't end up being possible for all users to get removed from the last org.

wdoconnell (commented):
Looks good! A couple minor comments like moving the button up. 👍

src/organizations/components/OrgProfileTab/index.tsx
blegesse-git: > The times when it's appropriate to use a method named `handleSomething` are typically when handling an event is a complex process that involves multiple steps or multiple responsibilities. 

I agree with your assessment, and I appreciate you taking time to explain your points in details. But playing devil's advocate here - I think that's probably why the person who wrote this function named it `handleRemove`. The function is doing two things here. It's invoking the `handleRemoveUser` context by passing it the `currentUserId` and then its redirecting user to `CLOUD_URL`. Since its executing two things here, I'd assume it would make sense to take the `handleSomething` naming convention, no?

src/organizations/components/OrgProfileTab/index.tsx
blegesse-git: good call

wdoconnell:
Ignore the name inconsistency on one of the videos - this is due to an issue with quartz-mock, which I'll resolve separately in this ticket.

https://github.com/influxdata/ui/issues/6444

src/organizations/components/OrgProfileTab/index.tsx
blegesse-git: disregard my comment above. You're saying the function that gets called by the `onConfirm` prop should be named `handleRemoveUser` and inside it, the context should be called `removeUser`. I misunderstood

wdoconnell (approved):

PR #6446, opened 2023-01-03
fix: keep flux in sync when select a new measurement with existing fields and tag values selection

### LINKED ISSUE #6429: Composition: upon measurement selection failure, the LSP can go out of sync with the UI
## About the bug
1. create composition query in Qx builder
2. select bucket, measurement, fields and/or tagValues
3. modify the measurement
4. on occasion, measurement will not update

## Expected behavior:
* schema browser should never go out of sync with the text
   1. if set measurement fails --> 
   2. the LSP server should msg the UI --> 
   3. the UI state (a.k.a. schema browser state) should then update to reflect (e.g. no measurement added)
* confirm each step of this^^ process is working.
* also check for memoization in UI state.

## Video (in prod):

https://user-images.githubusercontent.com/10232835/208775783-fb2973f5-df41-4024-96c2-16db9a99958f.mov

appletreeisyellow:
Closes #6429 

This PR fixes a bug in flux composition that upon a new measurement selection on an existing flux composition, the LSP went out of sync with the UI. 

## Before

https://user-images.githubusercontent.com/14298407/210408168-83e08512-acdc-4871-a90a-e83a33ebfc28.mov

## After

https://user-images.githubusercontent.com/14298407/210430599-63b18454-2fef-4aad-bac8-a6e29bf9d7ed.mov

### Checklist

### DIFF
--- a/src/languageSupport/languages/flux/lsp/connection.ts
+++ b/src/languageSupport/languages/flux/lsp/connection.ts
@@ -397,7 +397,7 @@ export class ConnectionManager {
         if (shouldRemoveDefaultMsg) {
           this._removeDefaultAndUpdateLsp(() => this._initLspComposition(toAdd))
         } else {
-          this._initLspComposition(toAdd)
+          this._updateLsp(toAdd, toRemove)
         }
         break
       case 'false|false|true':
--- a/src/languageSupport/languages/flux/lsp/connection.ts
+++ b/src/languageSupport/languages/flux/lsp/connection.ts
@@ -393,11 +393,17 @@ export class ConnectionManager {
         this._first_load = false
         break
       case 'false|true|true':
-        // re-sync just turned on.
+        // 1) re-sync just turned on or
+        // 2) measurement selection is changed when fields
+        //    and/or tag values were previously selected
+        if (!toAdd.bucket) {
+          // bucket is needed for initialization
+          toAdd.bucket = schema.bucket
+        }
         if (shouldRemoveDefaultMsg) {
           this._removeDefaultAndUpdateLsp(() => this._initLspComposition(toAdd))
         } else {
-          this._updateLsp(toAdd, toRemove)
+          this._initLspComposition(toAdd)
         }
         break
       case 'false|false|true':


src/languageSupport/languages/flux/lsp/connection.ts
```diff
@@ -397,7 +397,7 @@ export class ConnectionManager {
         if (shouldRemoveDefaultMsg) {
           this._removeDefaultAndUpdateLsp(() => this._initLspComposition(toAdd))
         } else {
-          this._initLspComposition(toAdd)
```
wiedld, line 400: This code (switch stmt) works like this:
1. determine if `hasMultipleItemsToSync`
2. if we do --> then we want to re-init the entire composition block.
  * this handles the race condition which can occur, on the LSP side, because the LSP is now stateful
  * This use case is then we have (1) turned off sync, (2) made lots of schema selector changes, (3) turned back on sync
3. to re-init entire composiiton --> `this._initLspComposition()` --> send LSP a `fluxComposition/initialize` request

The proposed change here is a step backwards. Re-introduces the race condition, without handling the root cause. I would recommend: (1) figuring out what LSP messages are sent during the bug, (2) confirm we are sending the right messages, then (3) either fix so we have the right messages (if bug is UI side), or fix the LSP response (if the bug is LSP side).

wiedld (changes requested):

src/languageSupport/languages/flux/lsp/connection.ts
appletreeisyellow: In the case of updating the existing measurement selection if the fields and/or tag values are selected, the switch statement will hit the case that `hasMultipleItemsToSync` is true. The bug happens here with missing information. 

See example below:

The previous UI session storage state is 
```
bucket: {id: "123", name: "sample_data", ...}
fields: ["humidity"]
measurement: "airSensors"
tagValues: [{key: "sensor_id", value: "TLM0101"}]
```

After a new measurement is selected, the current UI session storage state is
```
bucket: {id: "123", name: "sample_data", ...}
fields: []
measurement: "explosion"
tagValues: []
```
Note that fields and tag values are set to empty arrays

`_diffSchemaChange()` will generate
```
toAdd: {measurement: 'explosion'}
toRemove:
  fields: ['humidity']
  tagValues: [
    {key: 'sensor_id', value: 'TLM0101'}
  ]
shouldRemoveDefaultMsg: false
```
 
Then the code will hit the switch statement case `false|true|true` where `hasMultipleItemsToSync` is true. Since the `toAdd` doesn't have a bucket and `_initLspComposition()` only send request to LSP when there is a bucket ([source code](https://github.com/influxdata/ui/blob/2b43c7ddc48fe0ac6c9110483fc6ca8476af5d63/src/languageSupport/languages/flux/lsp/connection.ts#L261-L280)), this is where the bug happens. 

There are many different ways to fix this bug, but I'm hesitating to add special cases in `_diffSchemaChange()` or `_initLspComposition()` in order to keep the functionality pure. I'd rather re-write the statements within the `case 'false|true|true'` to make sure `toAdd` is correctly set before passing it to `_initLspComposition()`. What do you think?

wiedld: Excellent sleuthing!

That feels like the right fix, if we prioritize not sending the LSP too many requests at once (since we may get a race condition resulting in a dropped composition).

This may mean the LSP decides to insert an entirely new composition, while keeping the old one too, when the measurement is changed. But that's an ok behavior imo. Since the user can delete the old composition themselves -- and we never lose sync.

wiedld (approved):
Excellent debugging. TY!

PR #6448, opened 2023-01-03
chore: add deeplink for multiorg phase 2

### LINKED ISSUE #5244: Multi Org Phase 2 - Add / Edit Deep Links where appropriate

wdoconnell:
Closes #5244 

Multiorg phase 2 added an "organization list" tab to the Accounts page so that users can browse the statuses of all organizations in the current account. This adds a deeplink to that page.

This also adds an intercept to ensure the link actually works properly - since e2e tests run against the mock, we need to intercept the real IDPE org ID and make the mock use that ID.

### Checklist

### DIFF
--- a/src/utils/deepLinks.ts
+++ b/src/utils/deepLinks.ts
@@ -19,6 +19,7 @@ export const buildDeepLinkingMap = (orgId: string) => ({
   '/me/nodejsclient': `/orgs/${orgId}/load-data/client-libraries/javascript-node`,
   [`/me/${PROJECT_NAME_PLURAL.toLowerCase()}`]: `/orgs/${orgId}/${PROJECT_NAME_PLURAL.toLowerCase()}`,
   '/me/notebooks': `/orgs/${orgId}/${PROJECT_NAME_PLURAL.toLowerCase()}`,
+  '/me/orglist': `/orgs/${orgId}/accounts/orglist`,
   '/me/profile': `/orgs/${orgId}/user/profile`,
   '/me/pythonclient': `/orgs/${orgId}/load-data/client-libraries/python`,
   '/me/secrets': `/orgs/${orgId}/settings/secrets`,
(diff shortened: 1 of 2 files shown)


src/utils/deepLinks.ts
```diff
@@ -19,6 +19,7 @@ export const buildDeepLinkingMap = (orgId: string) => ({
   '/me/nodejsclient': `/orgs/${orgId}/load-data/client-libraries/javascript-node`,
   [`/me/${PROJECT_NAME_PLURAL.toLowerCase()}`]: `/orgs/${orgId}/${PROJECT_NAME_PLURAL.toLowerCase()}`,
   '/me/notebooks': `/orgs/${orgId}/${PROJECT_NAME_PLURAL.toLowerCase()}`,
+  '/me/orglist': `/orgs/${orgId}/accounts/orglist`,
```
hoorayimhelping, line 22: `orglist` just sits weird with me. I get why it's called that, given we have an `/orgs/$orgid` endpoint, but maybe it should be called `/me/orgs`?

hoorayimhelping (approved):
Don't forget to update https://github.com/influxdata/docs.influxdata.io/blob/main/content/development/guides/deeplinks.md

src/utils/deepLinks.ts
wdoconnell: I think `/orgs` works, but I have a slight preference for keeping it as something distinct - like `orgslist` - to avoid confusion on whether something is a deep link or not (naming them the same means we'll have a url "/me/orgs" and also "/orgs").

wdoconnell:
> Don't forget to update `deeplinks.md` in the docs.influxdata.io repo

@hoorayimhelping FYI - https://github.com/influxdata/docs.influxdata.io/pull/956

src/utils/deepLinks.ts
hoorayimhelping: yeah, I figured that was what was behind it.

Issue #6451, opened 2023-01-04
New datepicker (used in iox Qx builder) does not utilize timezones.


wiedld:
When I use the datePicker in the Qx Builder, on iox, I want to make sure the local vs UTC timezone is properly utilized.

This is about checking for the timezone usage:
* in persistence context
* in creation of flux query variables --> for the payload sent to the backend
* in the creation of SQL composition query

PR #6452, opened 2023-01-04
feat(6313): SQL composition

### LINKED ISSUE #6183: Using InfluxData schemas with SQL querying. (a.k.a. special identifiers needed)
## Scope of problem:
This addresses an issue concerning the querying of influxdata schemas in iox using SQL. Cross-team discussion will hopefully help: (1) scope the problem, (2) decide what is the reasonable amount of change, which  translates into (3) where those changes will occur (a.k.a. in UI, or in flux, or elsewhere).

## Problem:
tl;dr -- influxdata schema rules on WRITE, sql schema rules on READ.

We are going to be taking an influxdata schema, writing/migrating it into iox, and then querying with SQL. This means that users who are querying using sql, need to make sure to use special identifiers. This includes SQL in all the various interfaces, including the use of SQL within `iox.sql()`.

Specifically:
* in IOX only.
* once migrating influxdata schemas onto iox:
  * influxdata schemas have case sensitivity, and allow whitespace.
  * e.g. measurement names `airSensors` and `quarry blast` 
* SQL querying is case insensitive and does not accept whitespace, unless using special identifiers
  * so it needs to be double quoted
  * e.g. `select * from \"airSensors\"` 
* currently, queryd is not surfacing the errors.
  * this means that if we don't use the special identifiers, we get an obtuse 500 error.
  *   https://github.com/influxdata/idpe/issues/16129

## Re-creating the issue:

Made an iox lighthouse account, then populated my bucket using our sample_data on S3 (same task we run in remocal local dev). I got these case-sensitive and whitespace-containing identifiers:

Screen Shot 2022-10-27 at 2 07 38 PM

Then on querying, the user just needs to remember to use special identifiers. (Or perhaps should we have better error codes that state that?)

Screen Shot 2022-10-27 at 2 19 01 PM

## Scope of this ticket:

Is to address whether we need to: (1) message this need for special identifiers, to the users, and/or (2) do anything automatic escaping within flux's `iox.sql()`, and/or (3) add this messaging/hints in the error responses.

### LINKED ISSUE #6313: SQL Select Composition
## User Story

<p></p>

As a user, I would like the selections that I make in the Schema Browser to be used to formulate my basic SQL Select Statement so that I can easily start queries\.

<p></p>

## Acceptance Criteria

<p></p>

- An "Add" or "Inject" type function \(needs Daniel Design\) will be available in the schema browser after bucket, measurement, and at least one field or tag has been selected\.

- Upon initiating the function, the SQL select statement will be formulated and added to the SQL editing window

- It is NOT expected that there will be any synchronization capability; the function will always just be an add \(the user will have to manually delete when needed\)

<p></p>

## Notes

<p></p>

- While this may seem like just a "nice\-to\-have" feature, I am advocating for it for multiple reasons:

    - If we do not do this, then we will be forced to do some other re\-design work as the schema browser interaction is a confusing user experience with SQL in its current state \(remember, this was only a work\-in\-progress feature\)\.  Somehow we have to represent to the user that bucket selection is required, but that all other selections have no consequence, even though they are somewhat prompted to enter them\.  We can re\-design the schema browser to make this more apparent and solve this issue \- but it is a trade\-off of providing this feature as that will also require development time

    - Users will most likely know SQL, so yes, they can build a select statement\.  But users may have experience with date and time columns, they may not always have experience with timestamps\.  Getting a time range on their queries will be important and they may not know how to formulate the right "where" clause for the timestamp\.  Especially relative time\.  By showing them this, at least once, we get them on the right track and help prevent non\-time bound queries that would have a high cost to us

    - SQL is going to be the most performant query type in the IOx world; I don't want users to pick the flux editor solely because it helps them graphically build their query; so we really want them to want to use the SQL editor

    - We want users to be able to build queries quickly and iterate quickly \(largely for their benefit, but also it is a PAYG metric\)

    - We will get this feature request almost immediately if we don't have it\.  It doesn't mean we will have to do it, but if we are choosing other work that could be avoided if we do this, then we will already have this feature when the time comes  

- Sample Implementation for reference for representational purposes only:

<p></p>

<p></p>

```
# This is a function to build a single SQL Select Statement based on schema browser selections
def build_sql_statement(measurement, field_selections, tag_selection, timeindex):
    # Measurement - pass as a string
    # field_selections - pass as a list of strings
    # tag_selection - pass as a tagkey, tagvalue (only 1 tag key/value pair supported)
    # timeindex - pass as an integer containing index value for time_options:
    time_options = ["time > now() - interval \'5 minutes\'",
                    "time > now() - interval \'15 minutes\'",
                    "time > now() - interval \'30 minutes\'",
                    "time > now() - interval \'1 hour\'",
                    "time > now() - interval \'4 hours\'",
                    "time > now() - interval \'8 hours\'",
                    "time > now() - interval \'12 hours\'",
                    "time > now() - interval \'1 day\'",
                    "time > now() - interval \'2 days\'",
                    "time > now() - interval \'7 days\'",
                    "time > now() - interval \'14 days\'",
                    "time > now() - interval \'30 days\'",
                    "time > now() - interval \'60 days\'"]
    # Initialize select statement
    if len(field_selections) > 0:
        sql_statement = "select time, "
    else:
        sql_statement = "select * "
    # add field selections to select statement
    for index, item in enumerate(field_selections):
        if index != len(field_selections) - 1:
            sql_statement = sql_statement + item + ", "
        else:
            sql_statement = sql_statement + item + " "
    # add table/measurement selection to select statement
    sql_statement = sql_statement + "from " + measurement + " "
    # add tag selections to select statement
    if len(tag_selection) > 0:
        sql_statement = sql_statement + "where " + tag_selection[0] + " = " + tag_selection[1]
    # add time selection to select statement
    if 'where' not in sql_statement:
        sql_statement = sql_statement + "where "
    else:
        sql_statement = sql_statement + " and "
    sql_statement = sql_statement + time_options[timeindex]
    #print(sql_statement)
    return(sql_statement)

# testing and demonstration

# Example #1:  Passing measurement, 2 field selections, and 1 tag value with time option 1 (15 minutes)
returned_sql = build_sql_statement("climate",["temperature, humidity"],["location","\'Outside\'"],1)
print(returned_sql)
# Example #2: Passing measurement, no field selections, and 1 tag value with time option 1 (15 minutes)
returned_sql = build_sql_statement("climate","",["location","\'Outside\'"],1)
print(returned_sql)
# Example #3: Passing measurement, 1 field selection, and 1 tag value with time option 5 (8 hours)
returned_sql = build_sql_statement("climate",["temperature"],["location","\'Outside\'"],5)
print(returned_sql)
# Example #4: Passing measurement, 1 field selection, no tag key/values, and time option 5 (8 hours)
returned_sql = build_sql_statement("climate",["temperature"],"",5)
print(returned_sql)
```

<p></p>

wiedld:
Closes #6313 
Closes #6183 

## Done:
* make language agnostic parent connector to handle shared functionality
  * we rather frequently tweak the UX for the basic composition features (e.g. sync). 
  * Have this be shared "composition" code btwn SQL composition and Flux composition.
* Make SQL-specific connector.
  * this handles the SQL composition.
  * should be extensible for future(?) ideas of using a SQL LSP. (if we do so.)
* Handle our different timeRange types. 

## Future PRs:
* rename the "flux sync" toggle to be language agnostic. (Since used for SQL too.)
   * renamed in the UI (user facing) text only. 
   * Did not update the classnames and test data-ids.  
* after the remocal-iox e2es are enabled => make tests for SQL composition.

## Vid evidence:
Current functionality.

https://user-images.githubusercontent.com/10232835/210681111-e0c2bf51-78da-4edb-9256-a84c71c44fe0.mov

## Limitations of MVP implementation.
Note that on page reload (while sync'ed), it does not reattached the composition. Because no LSP, no parsing, is not that smart...so it will always recreate a duplicate (new) composition at the top.

https://user-images.githubusercontent.com/10232835/210681348-3818a63d-ec59-4e48-90ea-f01ceca15969.mov

 

## Checklist


src/shared/utils/sqlInterval.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        msg: 'for timestamps without upper bounds',
+        input: {
+          lower: '2023-01-03 00:00Z',
+          upper: null,
+          type: 'custom',
+        },
+        expected: `time >= timestamp '2023-01-03T00:00:00.000Z' AND time <= now()`,
+      },
+      // TODO -- this is what the new datePicker places in persistence context. No timezone. Will Fail.
+      // {
```
wiedld, line 62: Placeholder test, for [this ticket](https://github.com/influxdata/ui/issues/6451).

src/shared/utils/sqlInterval.ts
```diff
@@ -0,0 +1,88 @@
...
+          interval = interval.replace(init, replacement)
+        })
+      } catch (e) {
+        throw new Error(`Cannot handle nanosecond: ${e}`)
+      }
+    }
+
+    // our flight sql is not supporting microseconds. Convert.
+    toReplace = interval.match(microsecondDuration)
+    if (!!toReplace || toReplace?.length > 0) {
```
wiedld, line 49: Evidence of error:
Screen Shot 2023-01-04 at 2 37 03 PM

src/languageSupport/languages/sql/connection.ts
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  _editorChangeIsWithinComposition(change) {
+    if (!this._compositionRange) {
+      return
+    }
+    const {
+      start: {line: startLine},
+      end: {line: endLine},
+    } = this._compositionRange
```
appletreeisyellow, line 52:
```suggestion
    const {
      start: {line: startLine},
      end: {line: endLine},
    } as LspRange = this._compositionRange
```

It would be easier for future code maintainer to understand the shape of the object using`LspRange` or a defined type
https://github.com/influxdata/ui/blob/e4c590c99c1784a82e0f5184bd94dcd9b1db3b94/src/languageSupport/languages/flux/lsp/types/showMessage.ts#L53-L61

src/languageSupport/languages/sql/connection.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    const changeInBlock =
+      change.range.startLineNumber >= startLine &&
+      change.range.endLineNumber <= endLine
+
+    const isDeletion = change.text == ''
+    let deletionFromBlock = false
```
appletreeisyellow, line 59: Just a suggestion as in the other PR about boolean names (unless you decided you're not into that, in which case, ignore this)

`hasChangeInBlock` 
`hasDeletionFromBlock`

src/languageSupport/languages/sql/connection.ts
```diff
@@ -0,0 +1,240 @@
...
+      )
+      if (shouldEndSync) {
+        // use setTimeout to remove race condition
+        // have changes propogation first to SqlEditorMonaco.onChange()
+        setTimeout(() => {
+          this._callbackSetSession({
+            composition: {synced: false},
+          })
+          this._dispatcher(notify(compositionEnded()))
+        }, 0)
```
appletreeisyellow, line 86: Interesting! Good to know

src/languageSupport/languages/sql/connection.ts
```diff
@@ -0,0 +1,240 @@
...
+        lines += 4
+        break
+      default:
+        composition = composition.concat(['WHERE', timeRangeExpr])
+        lines += 2
+        break
+    }
+    return {
+      composition: composition.join('\n'),
+      lines,
```
appletreeisyellow, line 155: Instead of counting `lines`, why not just use `composition.length`?

src/languageSupport/languages/sql/connection.ts
```diff
@@ -0,0 +1,240 @@
...
+    const {composition, lines, lenLastLine} = this._buildComposition(
+      toAdd,
+      toRemove
+    )
+
+    // replacement Range
+    const startLineNumber = this._compositionRange?.start?.line ?? 1
+    const endLineNumber = this._compositionRange?.end?.line ?? 1
+    const addNewLine = startLineNumber == 1 && endLineNumber == 1
+    const endColumn = addNewLine ? 1 : Infinity
```
appletreeisyellow, line 173: Same suggestion as above for boolean names

`shouldAddNewLine`
`shouldEndColumn`

src/languageSupport/languages/sql/connection.ts
```diff
@@ -0,0 +1,240 @@
...
+        },
+      },
+    ])
+
+    // updated composition's new Range
+    this._setEditorBlockStyle(
+      {
+        start: {line: startLineNumber, column: 1},
+        end: {line: startLineNumber + lines - 1, column: lenLastLine},
+      },
```
appletreeisyellow, line 192: Same suggestion above for `LspRange`

appletreeisyellow (changes requested):
Overall really great work! 👍 I like how you pulled the common usage from flux lsp into `AgnosticConnectionManager` and laid infrastructure for future languages!

The comments I left including bugs I found, questions, and suggestions that you can take or ignore

src/languageSupport/languages/agnostic/connection.ts
```diff
@@ -0,0 +1,183 @@
...
+// Types
+import {EditorType, RecursivePartial} from 'src/types'
+import {LspRange as FluxLspRange} from 'src/languageSupport/languages/flux/lsp/types'
+
+// Utils
+import {notify} from 'src/shared/actions/notifications'
+import {oldSession} from 'src/shared/copy/notifications'
+
+// May as well re-use this well formed type.
+export type LspRange = FluxLspRange
```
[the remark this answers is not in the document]
wiedld: Because the `LspRange` should be language agnostic. Since the datePicker is no longer specifically for use in flux, and here is being used in SQL.

Instead of doing the work (in this PR) to rename all the classnames, type names, etc affliated with the datePicker to be not `flux--<rest>` --- we are instead just mapping the `FluxLspRange` type to the agnostic type. 

In a future cleanup PR, we can choose to make the datePicker not longer refer to `flux--` anywhere.

src/languageSupport/languages/sql/connection.ts
```diff
@@ -0,0 +1,240 @@
...
+import {rangeToInterval as buildTimeRange} from 'src/shared/utils/sqlInterval'
+import {notify} from 'src/shared/actions/notifications'
+import {compositionEnded} from 'src/shared/copy/notifications'
+
+const castType = (value, _typ = null) => {
+  // Outstanding tech debt.
+  // For the TagValue selector (Qx builder), flux composition, and rest of codebase.
+  // It only supports tagValues of type string.
+  return `'${value}'`
+}
```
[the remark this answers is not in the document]
wiedld: Technically, we should be looking at the type of the tag value and converting the SQL query to use that type.
`WHERE stringTag = 'quotes-add-string'` versus `WHERE numberTag = 42`.

I believe the influxdb data model enables any enum type as a tagValue. But our current frontend code only supports a string tagValue, [as per here](https://github.com/influxdata/ui/blob/master/src/dataExplorer/context/tags.tsx#L118).

src/languageSupport/languages/sql/connection.ts
wiedld: the class prop itself is already typed as `protected _compositionRange: LspRange`

src/languageSupport/languages/sql/connection.ts
wiedld: Good idea on the `shouldAddNewLine`, as it is a bool.
The `endColumn` is a number, representing the `endColumn` used in the monaco-editor range object (a few lines below).

src/languageSupport/languages/sql/connection.ts
wiedld: done

src/languageSupport/languages/sql/connection.ts
wiedld: done

src/shared/components/SqlMonacoEditor.tsx
```diff
@@ -38,11 +50,31 @@ const SqlEditorMonaco: FC<Props> = ({
...
+        sessionStore.range
+      )
+    }
+  }, [
+    connection.current,
+    sessionStore?.selection,
+    sessionStore?.selection.composition || null,
+    sessionStore?.setSelection,
+    sessionStore.range,
+  ])
```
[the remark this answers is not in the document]
wiedld: Nice catch.
The `sessionStore.range` should absolutely have the existence check too in the memoization.

The choice of using the existence check in the function call, is not required. Since this function is called when the sessionStore changes from undefined into existing. But it doesn't hurt to add it. Doing so.

src/languageSupport/languages/sql/connection.ts
wiedld: That's more of a style choice. As the code is written now in `_buildComposition()`, the `lines` count does match the length of the composition array (prior to join). I was just being pedantic about making sure future us/devs are aware of how the line counting works.

src/languageSupport/languages/agnostic/connection.ts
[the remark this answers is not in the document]
wiedld: done

src/shared/utils/sqlInterval.ts
```diff
@@ -0,0 +1,88 @@
...
+ *     https://www.postgresql.org/docs/15/datatype-datetime.html#DATATYPE-INTERVAL-INPUT
+ */
+export const rangeToInterval = (range: TimeRange) => {
+  switch (range.type) {
+    case 'selectable-duration':
+      return range.sql
+    case 'custom':
+    case 'duration':
+      const upper = convertToInterval(range.lower)
+      const lower = convertToInterval(range.upper)
```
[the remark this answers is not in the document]
wiedld: Not sure if we want to use the try/catch, since it requires us to not error. Which translate to returning a default behavior (such as the default timeRange). I feel like instead we want the user to know that they entered in an invalid time. What do you think?

src/shared/utils/sqlInterval.ts
```diff
@@ -0,0 +1,88 @@
...
+const durationExpr = /^\-(([0-9]+)(y|mo|w|d|h|ms|s|m|us|µs|ns))+$/g
+const singleDuration = /(([0-9]+)(y|mo|w|d|h|ms|s|m|us|µs|ns))/g
+const nanosecondDuration = /(([0-9]+)(ns))/g
+const microsecondDuration = /(([0-9,\.]+)(\W?)(microsecond))/g
+
+const convertToInterval = (upperOrLower: string) => {
+  if (upperOrLower == null || upperOrLower == 'now()') {
+    return 'now()'
+  }
+  if (!!upperOrLower.match(durationExpr)) {
```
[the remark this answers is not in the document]
wiedld: Nice catch!
Have updated the sqlInterval to handle future durations.

src/shared/utils/sqlInterval.ts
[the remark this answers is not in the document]
appletreeisyellow: Oh, I was asking the case what if `convertToInterval()` throw an error like this https://github.com/influxdata/ui/blob/6c01e6ed975a5aadc9dfe41e8d867630e7abe0d8/src/shared/utils/sqlInterval.ts#L44
should it be handled under `case 'duration':`? Something like:

```
case 'duration':
  try {
      const upper = convertToInterval(range.lower)
      const lower = convertToInterval(range.upper)
      return `time >= ${upper} AND time <= ${lower}`
  } catch (e) {
      ....
      return DEFAULT_TIME_RANGE.sql
  }
```

appletreeisyellow (approved):
Excellent work in putting the SQL composition together! 🙌 I really like the idea of having an agnostic conntection.ts

I replied for an edge case of try/catch in one of the comment, but I don't think that's a blocker

src/shared/utils/sqlInterval.ts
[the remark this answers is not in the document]
wiedld: Exactly.
But if return it as the default tiemRange --> we are silently doing things, and not telling the user why. So the user case to notice (on their own) that the inserted timeRange does not match.

Alternatively, we could check to see if this throws at a higher try/catch level, where we can toast to the user. Such as within the SqlConnector. We would then also want to reset the persistence context session back to the default timeRange -- but this persistence context only exists for the QxBuilder. And our connectors are right now designed to work inside this persistence context (e.g. the QxBuilder), as well as outside the persistence context (e.g. the other raw editors based on timeMachine use the Flux LspConnector for LSP actions outside of composition).

So that's where I'm not sure whether or not it's worthwhile to enter more code complexity. Especially since the `session.range` can only come from the datePicker, which already does not let an invalid duration (or date) be entered. This means that if we hit the current/existing throw clause -- it must be an error in our code.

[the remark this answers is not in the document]
wiedld: For the specific throwing for nanoseconds -- our code already converts to millisecond. So that throw clause is hit if (and only if) we have a bug in our replacement code.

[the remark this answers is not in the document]
appletreeisyellow: Ah, I got what you mean by "since it requires us to not error"! Yeah, it makes sense. Let's just keep it as is

Issue #6447, opened 2023-01-03
Use is_valid_flux() in ui, instead of the improper use of parse().


wiedld:
The contract of `parse()` versus `is_valid_flux()` is [documented in the LSP tests](https://github.com/influxdata/flux-lsp/blob/fc06b38e3d889974f72e9d133817eafda9c901dc/integration/lsp.test.ts#L168-L190).

We are currently, in the UI, using parse() incorrectly. We assume that any time there is a successful parse, it's therefore a vald flux query. This is incorrect.

appletreeisyellow:
Here are the places that uses `parse()`. I'm planning to replace them in separate PRs:

- [ ] Old Data Explorer: src/timeMachine/actions/queries.ts
- [ ] Variables:
    - src/variables/utils/getWindowVars.ts
    - src/variables/utils/astim.ts
- [ ] Shared: src/shared/contexts/query/preprocessing.ts
- [ ] Write Data: src/writeData/components/ClientCodeQueryHelper.tsx
- [ ] Language support: src/languageSupport/languages/flux/parser/index.ts
- [ ] Tasks: src/tasks/actions/thunks.ts
- [x] ~~Notebooks~~ (Notebooks is going to be deprecated, so not updating it)

wiedld:
Looks like we have a bunch of unit test fixtures which don't use valid flux.
* [hydrate vars](https://github.com/influxdata/ui/blob/master/src/variables/utils/hydrateVars.test.ts) unit tests.
* [exportVariables](https://github.com/influxdata/ui/blob/master/src/variables/utils/exportVariables.test.ts) unit tests.
* [filterUnusedVars](https://github.com/influxdata/ui/blob/master/src/shared/utils/filterUnusedVars.test.ts) unit tests.
* And a [single test](https://github.com/influxdata/ui/blob/master/src/timeMachine/components/SubmitQueryButton.test.tsx#L72) for the submit query button.

For the monitor-ci e2e tests, there was a bad remocal-deploy build (which is why everything failed) -- so restarted that workflow from the beginning. But there are [many oss-e2e failures](https://app.circleci.com/pipelines/github/influxdata/monitor-ci/18627/workflows/9ead852d-0cb2-4f42-9c64-3a118e851c70/jobs/1319255), and I suspect many are also from invalid flux.

Now the question is whether or not it's worthwhile to fix all of these test fixtures.

[...]

PR #6455, opened 2023-01-05
chore: remove `getOrgIDFromBuckets`

### LINKED ISSUE #6363: Remove getOrgIDFromBuckets
Per [this comment](https://github.com/influxdata/ui/pull/6062#discussion_r1037499748), we can remove some or all uses of `getOrgIDFromBuckets` in favor of `getOrg(state).id`.

appletreeisyellow:
Closes #6363 

This PR removes the legacy function `getOrgIDFromBuckets`. See the original issue for the reason. 

This PR also removes one of the `parse()` item related to #6447 

### Checklist

### DIFF
--- a/src/shared/components/TimeSeries.tsx
+++ b/src/shared/components/TimeSeries.tsx
@@ -205,6 +205,7 @@ class TimeSeries extends Component<Props, State> {
       variables,
     } = this.props
     const queries = this.props.queries.filter(({text}) => !!text.trim())
+    const orgID = this.props.match.params.orgID
 
     if (!queries.length) {
       this.setState(defaultState())
@@ -239,8 +240,6 @@ class TimeSeries extends Component<Props, State> {
 
       // Issue new queries
       this.pendingResults = queries.map(({text}) => {
-        const orgID = this.props.match.params.orgID
-
         const windowVars = getWindowVarsFromVariables(text, variables)
         const extern = buildUsedVarsOption(text, variables, windowVars)
 
--- a/src/shared/components/TimeSeries.tsx
+++ b/src/shared/components/TimeSeries.tsx
@@ -20,7 +20,6 @@ import {getRangeVariable} from 'src/variables/utils/getTimeRangeVars'
 import {getWindowVarsFromVariables} from 'src/variables/utils/getWindowVars'
 import {buildUsedVarsOption} from 'src/variables/utils/buildVarsOption'
 import 'intersection-observer'
-import {getOrgIDFromBuckets} from 'src/timeMachine/actions/queries'
 import {hashCode} from 'src/shared/apis/queryCache'
 import {RunQueryPromiseMutex} from 'src/shared/apis/singleQuery'
 import {parseASTIM} from 'src/variables/utils/astim'
@@ -199,7 +198,6 @@ class TimeSeries extends Component<Props, State> {
 
   private reload = async () => {
     const {
-      buckets,
       check,
       isCurrentPageDashboard,
       notify,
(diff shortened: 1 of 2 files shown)


src/shared/components/TimeSeries.tsx
```diff
@@ -241,8 +239,7 @@ class TimeSeries extends Component<Props, State> {
 
       // Issue new queries
       this.pendingResults = queries.map(({text}) => {
-        const orgID =
-          getOrgIDFromBuckets(text, buckets) || this.props.match.params.orgID
+        const orgID = this.props.match.params.orgID
```
hoorayimhelping, line 242: Minor tweak, but I think this line can be pulled up to the top of the `reload` function, rather than being called on every iteration of the loop, since it doesn't need to use `text` anymore.

That's how the code was structured before this change was implemented. See [this line](https://github.com/influxdata/ui/commit/45d1b2d30a5f0481a5db56bb9a31e3d07cbf2e3b#diff-43db5ef9584cae7f3e2a4eceae743ecc2dda44e9b26fc68f467dffbd62bf637cL158).

If you don't know how I found that line, I did it like this: I opened `TimeSeries.tsx` in the github UI in a separate window, then clicked on on the icon that appears and clicking `view git blame for line`, then followed that line through history to see when it was initially introduced.

src/timeMachine/actions/queries.ts
```diff
@@ -319,13 +295,12 @@ export const executeQueries =
...
       await dispatch(hydrateVariables())
       const allVariables = getAllVariables(state)
-      const allBuckets = getAll<Bucket>(state, ResourceType.Buckets)
 
       const startTime = window.performance.now()
       const startDate = Date.now()
 
       const pendingResults = queries.map(({text}) => {
-        const orgID = getOrgIDFromBuckets(text, allBuckets) || getOrg(state).id
+        const orgID = getOrg(state).id
```
hoorayimhelping, line 303: Same story here about calling `getOrg(state).id` outside of the queries loop. See [this line.](https://github.com/influxdata/ui/commit/45d1b2d30a5f0481a5db56bb9a31e3d07cbf2e3b#diff-0f78a1f05484420b38348fd715c2179c9f471f3ae36ba25dec94d60f5529d1edL139)

src/shared/components/TimeSeries.tsx
appletreeisyellow: Yeah, it makes sense to move the `orgID` up to the top. Thanks for the `view git blame for line` instruction! It will be helpful for future development!

hoorayimhelping (approved):

src/timeMachine/actions/queries.ts
```diff
@@ -87,26 +84,6 @@ export const setQueryResults = (
 
 let pendingCheckStatuses: CancelBox<StatusRow[][]> = null
 
-export const getOrgIDFromBuckets = (
```
wiedld, line 90: Sanity checking. This function [was put in](https://github.com/influxdata/ui/commit/45d1b2d30a5f0481a5db56bb9a31e3d07cbf2e3b) to handle the use case of querying buckets from "other" orgs. What does that mean @hoorayimhelping ? Does this use case still exist?

wiedld (commented):
Love the red diff. Just want to double check that the use case no longer exists.

src/timeMachine/actions/queries.ts
appletreeisyellow: Barbara [commented](https://github.com/influxdata/ui/issues/6363#issuecomment-1338625874) in the original issue that the "other" orgs is an organization that was created to receive sample data and this organization is no longer exist

wiedld: Sweet. TY!

wiedld (approved):

PR #6458, opened 2023-01-05
chore: refresh org creation allowance after free-checkout

### LINKED ISSUE #6457: Refresh org creation allowance after checkout to upgrade to payg
With `createDeleteOrgs` on, if a user upgrades to PAYG, they still see the "Upgrade" button in the global header after checking out, rather than seeing "Create Organization". At that point they should still have two orgs in their PAYG quota.

Root cause is that the org creation allowance wasn't being refreshed after checkout (only at login, or after creating a new org, or visiting the org list tab). We need another call to the thunk when a user successfully checks out.

Credit to @eatondustin1 for finding this one.

wdoconnell:
Closes #6457 

Calls the thunk for refreshing the organization creation allowance after a user checks out. 

When checking out from free to PAYG, the user's org creation allowance must be updated so that they are immediately able to create additional orgs, up to the PAYG default (3), and increase from the Free default (1).

This is feature flagged because the organization creation allowance is not used unless the `createDeleteOrgs` FF is on.

### Checklist

### DIFF
--- a/src/checkout/context/checkout.tsx
+++ b/src/checkout/context/checkout.tsx
@@ -28,11 +28,15 @@ import {getErrorMessage} from 'src/utils/api'
 import {event} from 'src/cloud/utils/reporting'
 
 // Thunks
+import {getOrgCreationAllowancesThunk} from 'src/identity/allowances/actions/thunks'
 import {getQuartzIdentityThunk} from 'src/identity/actions/thunks'
 
 // Selectors
 import {selectCurrentIdentity} from 'src/identity/selectors'
 import {shouldGetCredit250Experience} from 'src/me/selectors'
+
+// Utils
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 export type Props = {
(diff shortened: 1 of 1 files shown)


src/checkout/context/checkout.tsx
```diff
@@ -311,6 +315,10 @@ export const CheckoutProvider: FC<Props> = React.memo(({children}) => {
         dispatch(notify(submitError()))
       } finally {
         setIsSubmitting(false)
+        // Refresh whether user is allowed to create new orgs after upgrading to PAYG.
+        if (isFlagEnabled('createDeleteOrgs')) {
```
hoorayimhelping, line 319: what's a `delete org` and why would I create one?

smirk_emoji

hoorayimhelping (approved):

src/checkout/context/checkout.tsx
wdoconnell: `isFlagEnabled(sassEmoji)`

PR #6459, opened 2023-01-05
feat: 'remove users first' block on deleting orgs in payg/contract accounts

### LINKED ISSUE #6450: Eliminate 'remove users first' block on deleting orgs in payg/contract accounts
@mavarius has reviewed the delete-org flow for paid org deletion in multi-org phase 2 and would like the following changes:
 

- [x] If other users exist in a PAYG/Contract account, don't block the user from removing the org until they remove the other users.

- [x] No popup should appear when the user clicks the 'delete org' button for PAYG/Contract

- [x] The 'delete org' modal for PAYG/contract instead just warns of the number of users in the org that will be deleted before proceeding.

Figma adjustments illustrated here:
[Phase 2 delete org modal adjustments](https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=2962%3A18294&t=FdXkr7wzAM394BWj-1)

FYI @gunnaraasen

blegesse-git:
Closes #6450 

Currently in the UI, if a non free account user tries to delete an org with multiple users associated with it, the deletion process gets blocked by a notification warning user to remove other users first before deleting the org. 
Pr removes this block (notification) and instead display the delete org overlay that allows user to delete org. The number of users associated with the org will now be shown inside the overlay, and when clicked, user will be redirected to the members page as shown in the demo below. 

Screen Shot 2023-01-09 at 4 36 58 PM

### Checklist

### DIFF
--- a/src/organizations/components/OrgProfileTab/DeletePanel.tsx
+++ b/src/organizations/components/OrgProfileTab/DeletePanel.tsx
@@ -11,7 +11,6 @@ import {
   PopoverInteraction,
   PopoverPosition,
 } from '@influxdata/clockface'
-import {OrgUsersLink} from 'src/shared/components/notifications/NotificationButtons'
 import PageSpinner from 'src/perf/components/PageSpinner'
 
 // Contexts
@@ -25,17 +24,10 @@ import {
   selectOrgSuspendable,
 } from 'src/identity/selectors'
 
-// Notifications
-import {deleteOrgWarning} from 'src/shared/copy/notifications'
-import {notify} from 'src/shared/actions/notifications'
-
 // Constants
 import {CLOUD} from 'src/shared/constants'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {dismissOverlay, showOverlay} from 'src/overlays/actions/overlays'
-
-// Types
-import {NotificationButtonElement} from 'src/types'
 
 // Styles
 import 'src/organizations/components/OrgProfileTab/style.scss'
@@ -84,21 +76,13 @@ const DeleteOrgButton: FC = () => {
     hidePopup()
   }
 
-  const shouldShowUsersWarning = users.length > 1
-
   const handleSuspendOrg = () => {
     if (orgCanBeSuspended) {
-      if (shouldShowUsersWarning) {
-        const buttonElement: NotificationButtonElement = onDismiss =>
-          OrgUsersLink(`/orgs/${org.id}/members`, onDismiss)
-        dispatch(notify(deleteOrgWarning(buttonElement)))
-      } else {
-        dispatch(
-          showOverlay('suspend-org-in-paid-account', null, () =>
-            dispatch(dismissOverlay())
-          )
+      dispatch(
+        showOverlay('suspend-org-in-paid-account', {users: users.length}, () =>
+          dispatch(dismissOverlay())
         )
-      }
+      )
     }
   }
 
--- a/src/organizations/components/OrgProfileTab/DeletePanel.tsx
+++ b/src/organizations/components/OrgProfileTab/DeletePanel.tsx
@@ -20,6 +20,7 @@ import {UsersContext} from 'src/users/context/users'
 // Selectors
 import {getOrg} from 'src/organizations/selectors'
 import {
+  selectCurrentAccount,
   selectCurrentIdentity,
   selectOrgCreationAllowance,
   selectOrgSuspendable,
@@ -65,7 +66,7 @@ const DeleteOrgButton: FC = () => {
   const orgCanBeSuspended = useSelector(selectOrgSuspendable)
   const canCreateOrgs = useSelector(selectOrgCreationAllowance)
   const {users} = useContext(UsersContext)
-  const {account} = useSelector(selectCurrentIdentity)
+  const account = useSelector(selectCurrentAccount)
 
   const popoverRef = useRef()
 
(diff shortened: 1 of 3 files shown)


src/organizations/components/OrgProfileTab/DeletePanel.tsx
```diff
@@ -65,6 +65,7 @@ const DeleteOrgButton: FC = () => {
   const orgCanBeSuspended = useSelector(selectOrgSuspendable)
   const canCreateOrgs = useSelector(selectOrgCreationAllowance)
   const {users} = useContext(UsersContext)
+  const {account} = useSelector(selectCurrentIdentity)
```
wdoconnell, line 69: You can use a `selectCurrentAccount` selector here, to be more specific. There is potential for a re-render because `identity.org` gets more data loaded into it on one or two pages.

src/organizations/components/OrgProfileTab/DeletePanel.tsx
```diff
@@ -94,8 +96,10 @@ const DeleteOrgButton: FC = () => {
         dispatch(notify(deleteOrgWarning(buttonElement)))
       } else {
         dispatch(
-          showOverlay('suspend-org-in-paid-account', null, () =>
-            dispatch(dismissOverlay())
+          showOverlay(
+            'suspend-org-in-paid-account',
+            {users: users.length},
+            () => dispatch(dismissOverlay())
```
wdoconnell, line 102: Let's call this something like `userCount`, since it's just a number instead of a user object.

wdoconnell, line 102: I would handle the situation where `users` isn't set here or in the child component. `users` is coming from an API call that's made in the context, so I believe the page can load before users comes back (and before the user tries to open the overlay).

One way would be: 
`{users: users.length || 0}`

src/organizations/components/OrgProfileTab/SuspendPaidOrgOverlay.tsx
```diff
@@ -158,6 +160,16 @@ export const SuspendPaidOrgOverlay: FC = () => {
         >
           You will be able to recover this organization's data for up to 7 days
           by contacting support. It will be unrecoverable afterwards.
+          {params.users > 0 && (
+            <>
```
wdoconnell, line 164: Same here: I might add a guard `params?.users` so that we don't get a `ReferenceError` if the users param is not passed to the overlay controller.

wdoconnell, line 164: Is it `params?.users > 0` or `params?.users > 1`? If there is only one user on the org (the one removing it), I don't think we need to warn the user.

src/organizations/components/OrgProfileTab/DeletePanel.tsx
```diff
@@ -84,7 +85,8 @@ const DeleteOrgButton: FC = () => {
     hidePopup()
   }
 
-  const shouldShowUsersWarning = users.length > 1
+  // only show warning for free accounts with multiple users
+  const shouldShowUsersWarning = account.type === 'free' && users.length > 1
```
wdoconnell, line 90: The meaning of this term has changed because there's more than one warning now - you get a warning of one type with multiple users in a free account, and a warning once you make it into the overlay with multiple users in a PAYG or Contract account.

How about `freeAccountWithOtherUsers`?

src/organizations/components/OrgProfileTab/SuspendPaidOrgOverlay.tsx
```diff
@@ -158,6 +160,16 @@ export const SuspendPaidOrgOverlay: FC = () => {
...
+              <br />
+              <br />
+              <Link to={`/orgs/${org.id}/members`} onClick={onClickCancel}>
+                {params.users} users in this organization
+              </Link>{' '}
+              will lose access to all organization data immediately.
+            </>
+          )}
```
wdoconnell, line 172: In the video, we see `0 users will lose access to all organization data immediately`. How is that happening?  Isn't the warning only shown when `params.users > 0`?

wdoconnell (changes requested):
Looking good, but a couple of minor comments here. I think we should probably only show the warning if there's more than one user (similar to what happens with free accounts).

src/organizations/components/OrgProfileTab/SuspendPaidOrgOverlay.tsx
blegesse-git: ohh I think I assumed users.length === 0 meant org had no other users aside from the user who's currently trying to delete it. If that's not the case, yes, we should definitely update the code to say `params?.users > 1`. I'll get that fixed

src/organizations/components/OrgProfileTab/SuspendPaidOrgOverlay.tsx
blegesse-git: yup you're correct. Ignore the video for now. I'll have to add a new video that reflects the most recent commit.

src/organizations/components/OrgProfileTab/SuspendPaidOrgOverlay.tsx
blegesse-git: I added a fallback value for userCount `userCount: users.length || 0` so we are sure to get a params value in this component

wdoconnell (approved):

wdoconnell (approved):

PR #6460, opened 2023-01-05
feat: add storage type to org profile page

wdoconnell:
Closes [AIM #7058](https://github.com/influxdata/quartz/issues/7058)

Adds display on the org profile page to disclose whether the current org is an iox or a tsm org. Adds associated cloud2 e2e test.

Screen Shot 2023-01-05 at 6 18 42 PM

### Checklist

### DIFF
--- a/src/organizations/components/OrgProfileTab/index.tsx
+++ b/src/organizations/components/OrgProfileTab/index.tsx
@@ -47,8 +47,9 @@ const OrgProfileTab: FC = () => {
   const org = useSelector(getOrg)
   const quartzOrg = useSelector(selectCurrentOrg)
   const storageType = org?.defaultStorageType
-  const dispatch = useDispatch()
   const {users} = useContext(UsersContext)
+
+  const dispatch = useDispatch()
 
   // Data about the user's organization is intentionally re-fetched when this component mounts again.
   const [orgDetailsStatus, setOrgDetailsStatus] = useState(
@@ -89,7 +90,7 @@ const OrgProfileTab: FC = () => {
     tsm: 'TSM',
   }
 
-  const capitalizedStorageType =
+  const formattedStorageType =
     storageTypeMap[storageType.toLowerCase()] || storageType
 
   const OrgProfile = () => (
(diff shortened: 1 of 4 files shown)


src/organizations/selectors/index.ts
```diff
@@ -4,12 +4,8 @@ import {get} from 'lodash'
 // Types
 import {AppState, Organization} from 'src/types'
 
-interface OrganizationWithStorageType extends Organization {
-  defaultStorageType?: string
-}
-
 export const isOrgIOx = (state: AppState): boolean => {
-  const org: OrganizationWithStorageType = getOrg(state)
+  const org: Organization = getOrg(state)
```
wdoconnell, line 9: This type is now part of the openapi spec, so we don't need this interface.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -81,6 +81,7 @@ const OrgProfileTab: FC = () => {
   const allowSelfRemoval = users.length > 1
   const showLeaveOrgButton = isFlagEnabled('createDeleteOrgs')
   const hasFetchedOrgDetails = orgDetailsStatus === RemoteDataState.Done
+  const hasFetchedStorageType = Boolean(org?.defaultStorageType)
```
wdoconnell, line 85: I though about using the `isOrgIOX` selector here and having the storage type set to something like: `isOrgIox ? 'IOx' : 'TSM'`

Decided against it because the boolean selector is lossy (we're losing data on what the endpoint actually sends), and trying to convert it back into a string will create headaches if we have the UI glossing over what the endpoint sends, especially if we change the name of the storage type or add more in the future.

alejom99 (approved):

hoorayimhelping (dismissed):

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -81,6 +83,14 @@ const OrgProfileTab: FC = () => {
...
   const showLeaveOrgButton = isFlagEnabled('createDeleteOrgs')
   const hasFetchedOrgDetails = orgDetailsStatus === RemoteDataState.Done
+  const hasFetchedStorageType = Boolean(storageType)
+
+  const storageTypeMap = {
+    iox: 'IOx',
+    tsm: 'TSM',
+  }
+
+  const formattedStorageType = storageTypeMap[storageType] || storageType
```
wdoconnell, line 94: I removed the `toLowerCase()` here. It isn't necessary since the endpoint is already sending back a lowercase response. It also causes an error in OSS, where storageType isn't returned (can't run the `toLowerCase` method on an undefined, and not worth overcomplicating this by adding in additional logic to check whether the type is `string` or `undefined`).

blegesse-git (approved):

PR #6467, opened 2023-01-09
chore: use `is_valid_flux()` to check flux scripts before saving them

### LINKED ISSUE #5169: Thought experiment: What if `flux-lsp-*/{parse|format_from_js_file}` went away?

### LINKED ISSUE #6447: Use is_valid_flux() in ui, instead of the improper use of parse().

appletreeisyellow:
Close #6447
Close #5169

This PR is still working in progress.

A try out to use `is_valid_flux()` in the places directly and indirectly use `parse()`. 

### Checklist


appletreeisyellow:
Work being done in https://github.com/influxdata/ui/pull/6534 instead

PR #6468, opened 2023-01-09
fix: update zuora url on credit card form

wdoconnell:
We're seeing `502` errors this morning in downloading the zuora credit card form script in certain regions. Per Zuora, they made a change to the Zuora sandbox environment that is causing the error. They've reverted the change, so it should be working now. However, the underlying issue is that we're pulling our Zuora script from their 'API Sandbox' not the copy in prod. 

The zuora-min.js script is identical in both cases (I've confirmed by `diff`ing them locally). But we shouldn't be retrieving this checkout script from the sandbox url.


wdoconnell:
Holding off on merging this until tomorrow a.m. since the change on the Zuora side has resolved this for now.

eatondustin1 (approved):
Nice!

PR #6470, opened 2023-01-10
chore(Account Page): remove deactivate button

### LINKED ISSUE #6453: Refactor(AccountPage): remove Deactivate Account Button from Account settings page
Background - `free account cancellation` epic was created with a goal to replace free account deletion. But according to this [issue](https://github.com/influxdata/quartz/issues/6340), `freeAccountCancellation` feature has been abandoned since deletion of free account means automatic deletion (free account data wont be retained after deletion).  
Currently, there is a `Deactivate Account` button inside Account Page behind `freeAccountCancellation` feature flag that utilizes the `/api/v2/quartz/cancel,` endpoint. Remove this code as it is no longer relevant.  

Screen Shot 2023-01-04 at 6 39 15 PM

blegesse-git:
Closes #6453 

Pr removes code behind `freeAccountCancellation` feature flag which was originally meant to allow free tier users to cancel their accounts using the `Deactivate Button` inside Account Page. But since that's no longer a possibility, pr removes this feature from codebase. 

Upcoming pr will remove the feature flag from idpe and config cat

Before: Account Page when feature flag is on/off
https://user-images.githubusercontent.com/66275100/211583812-47088170-aba4-4472-9e90-31903b844be1.mov

After: Account Page when feature flag is on/off
https://user-images.githubusercontent.com/66275100/211584260-0e3ed0b7-dedb-449d-ae48-781aac4ddb82.mov

### Checklist


src/accounts/AccountPage.tsx
```diff
@@ -34,9 +33,6 @@ import AccountTabContainer from 'src/accounts/AccountTabContainer'
 import AccountHeader from 'src/accounts/AccountHeader'
 import {DeleteFreeAccountProvider} from 'src/accounts/context/DeleteFreeAccountContext'
 
-import CancellationOverlay from './CancellationOverlay'
-import CancelServiceProvider from 'src/billing/components/PayAsYouGo/CancelServiceContext'
```
wdoconnell, line 38: This is called CancellationOverlay, but there are actually a couple overlapping 'Cancellation' overlays in the code, one of which is for PAYG. This one would better be called FreeAccountCancellationOverlay, but we're deleting it.

We're safe to remove this particular overlay, because it's only used in the context of free account cancellation from what I can tell (I tested it by eliminating the default export and seeing what relied on that).

src/accounts/CancellationOverlay.tsx
```diff
@@ -1,162 +0,0 @@
-// Libraries
-import React, {FC, useContext, useMemo, useState} from 'react'
-import {useDispatch, useSelector} from 'react-redux'
```
wdoconnell, line 3: Same comment as earlier: re - okay to remove this particular file.

wdoconnell (approved):
Great job with the red diff, and disentangling these confusingly named components between free and payg. 👍

PR #6478, opened 2023-01-11
chore: no cardinality on orglimits page if org is iox; simplify limits

wdoconnell:
Helps with [AIM 7046](https://github.com/influxdata/quartz/issues/7046)

Removes cardinality references from the org limits page when an org is an iox org. Adds corresponding test.

This component needed a refactor because it was too complicated, and the variable names obscured what it was doing. It renders 9 limit cards. The reason it was confusing is that the data from the api comes in three different formats, so different render logic is needed for the buckets limits, rate limits, and other limits. 

This goes partway to simplifying it for the next person. The better fix would be to change the data shape, but that would require some heavier lifting in other components.

Not an IOx Org
--
Screen Shot 2023-01-12 at 10 42 00 AM

An IOx Org
--
Screen Shot 2023-01-12 at 10 51 00 AM

### Checklist

### DIFF
--- a/src/billing/components/Free/OrgLimits.tsx
+++ b/src/billing/components/Free/OrgLimits.tsx
@@ -1,78 +1,118 @@
+// Libraries
 import React, {FC} from 'react'
 import {useSelector} from 'react-redux'
-import {Grid, Columns} from '@influxdata/clockface'
+import {Columns, Grid} from '@influxdata/clockface'
+
+// Components
 import OrgLimitStat from 'src/billing/components/Free/OrgLimitStat'
+
+// Selectors
+import {isOrgIOx} from 'src/organizations/selectors'
 
 // Types
 import {AppState} from 'src/types'
 import {Limit} from 'src/cloud/actions/limits'
 import {LimitsState} from 'src/cloud/reducers/limits'
 
-const limits = (orgLimits: LimitsState | {}) =>
+enum LimitNames {
+  Buckets = 'buckets',
+  Endpoints = 'endpoints',
+  MaxRetentionSeconds = 'maxRetentionSeconds',
+  OrgID = 'orgID',
+  Rate = 'rate',
+  Status = 'status',
+}
+
+interface LimitCardProps {
+  limitName: string
+  limitValue: Limit
+}
+
+const removeUnrenderedProps = (orgLimits: LimitsState | {}) =>
   Object.entries(orgLimits).filter(
     ([limitName]) =>
-      limitName !== 'orgID' &&
-      limitName !== 'status' &&
-      limitName !== 'endpoints'
+      limitName !== LimitNames.OrgID &&
+      limitName !== LimitNames.Status &&
+      limitName !== LimitNames.Endpoints
   )
 
-const OrgLimits: FC = () => {
+const LimitCard: FC<LimitCardProps> = ({limitName, limitValue}) => {
+  return (
+    <Grid.Column
+      key={limitName}
+      widthXS={Columns.Twelve}
+      widthSM={Columns.Six}
+      widthMD={Columns.Four}
+      widthLG={Columns.Three}
+    >
+      <OrgLimitStat name={limitName} value={limitValue.maxAllowed} />
+    </Grid.Column>
+  )
+}
+
+export const OrgLimits: FC = () => {
   const orgLimits = useSelector((state: AppState) => state.cloud?.limits ?? {})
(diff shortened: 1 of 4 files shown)


src/billing/components/Free/OrgLimits.tsx
```diff
@@ -1,78 +1,87 @@
...
+          const hiddenLimits = ['queryTime']
+          if (orgUsesIOx) {
+            hiddenLimits.push('cardinality')
+          }
+
+          return Object.keys(limitData)
+            .filter(sublimitType => !hiddenLimits.includes(sublimitType))
+            .map(sublimitType => {
               return (
-                <Grid.Column
```
wdoconnell, line 30: This exact grid code was repeated 3x, so I've moved it into a component.

src/billing/components/Free/OrgLimits.tsx
```diff
@@ -1,78 +1,110 @@
...
   return (
     <Grid>
-      {limits(orgLimits).flatMap(([name, value]: [string, Limit]) => {
-        if (name === 'buckets') {
-          return Object.entries(value)
-            .filter(([n]) => n !== 'limitStatus')
-            .map(([n, v]: [string, number]) => {
+      {removeUnrenderedProps(orgLimits).map(limit => {
+        const limitName = limit[0]
+        const limitValue = limit[1]
```
hoorayimhelping, line 52: alternatively, the new meta is:

```ts
const [limitName, limitValue] = limit
```

but who cares. It's a small stylistic thing. It's just that the kids love their destructuring, fr fr no cap

src/billing/components/Free/OrgLimits.tsx
```diff
@@ -1,78 +1,110 @@
...
+
+        // The "Rate" category includes multiple object literal 'limits', each with its own 'limitStatus.'
+        if (limitName === 'rate') {
```
hoorayimhelping, line 55: only real piece of feedback I have is maybe `limitName` should be based off a type or an enum, rather than a string. Do any parts of `openapi` define these?

hoorayimhelping (dismissed):

src/billing/components/Free/OrgLimits.tsx
wdoconnell: Nah, you're right I should destructure this, it's more readable.

src/billing/components/Free/OrgLimits.tsx
wdoconnell: There doesn't appear to be a satisfying way to do this, unfortunately: `openapi` does define an interface `Limit`, but not an enum. The slick solution would be converting the keys of the _interface_ in openapi into an enum. I can't find a way to do it in the Typescript docs though, and [I think this is why: ](https://stackoverflow.com/questions/72359790/typescript-generate-enum-from-an-interface-keys-with-index-number)

> actually now that I think about it, I think it is not possible to convert an interface to an enum at compile since enums are a runtime thing and without external tools this might be impossible. The other way around might work though. Is enum to interface something that would fit your needs? – 
> [Tobias S.](https://stackoverflow.com/users/8613630/tobias-s)
> 
>  [May 24, 2022 at 8:54](https://stackoverflow.com/questions/72359790/typescript-generate-enum-from-an-interface-keys-with-index-number#comment127832528_72359790)
> 
> I don't see how that would work. Types only exist at compile time, but you need the values of the enum at runtime. – 
> [Robby Cornelissen](https://stackoverflow.com/users/3558960/robby-cornelissen)
>  [May 24, 2022 at 8:55](https://stackoverflow.com/questions/72359790/typescript-generate-enum-from-an-interface-keys-with-index-number#comment127832534_72359790) 
> Not really possible. The order is problematic to begin with (as you mention it's important)

wdoconnell: I'll write out an enum anyway - that's likely to be less buggy here so that someone doesn't fatfinger one of the names and cause one of the limits to be missing.

wdoconnell:
Done - I've retested to confirm the enums are identical to the strings, and we get the correct number of limits on the page still.

hoorayimhelping (dismissed):
Thanks for taking the time to clean this ole boy up!

wdoconnell:
I'll hold off on merging this until Tuesday am - I had a `skip CI` in the commits so it wasn't running in the merge queue.

hoorayimhelping (approved):

PR #6484, opened 2023-01-13
feat(6483): Separate queries and results, for graph versus table.

### LINKED ISSUE #6483: QxBuilder: separate queries for table vs graph.
Task in epic #6482.

Subtasks:
- [x] Make the query for the table, be different than the query for the graph.
- [x] Store data separately.

wiedld:
Closes #6483 

Should be no evident, visible feature change. This PR is all about decoupling the table versus graph. Additionally, it sets up the code pattern to have the graph subquerying & data results subscribe to viewOption changes, which is the crux of the next few subtasks (follow up PRs on the Epic).

## Done, per commit:
* Moved file. `ResultsContext` is a context, not a component.
* Separate the `ResultsViewContext` from `ResultsContext`.
* Disentangle the `TableResults` view from the `GraphResults` view.
   * this is so that the `TableResults` can use different data than the `GraphResults`.
* Make a new `ChildResultsContext`, which subscribes to changes from the parent `ResultsContext`.
   * this subscription is such that if the Results will change (e.g. query text is changed and query is re-issued), the child results will update.
* Add `ResultsViewContext.viewOptions`.
  * the `ChildResultsContext` also subscribes to these viewOptions.
  * As the user selects a new option (e.g. add tagKey to grouping, or select field for smoothing), then a subquery is triggered in `ChildResultContext`. 

## Checklist


src/dataExplorer/components/Results.tsx
```diff
@@ -144,58 +124,157 @@ const Results: FC = () => {
...
+    [setStatus, setResult]
+  )
+
+  // TODO: make component, for QxBuilder-specific graph subquery options
+  const subQueryOptions = null
+
+  return (
+    <>
+      {subQueryOptions}
+      {/* <SubQueryOptions update={setViewOptions} /> */}
```
wiedld, line 190: This `SubQueryOptions` component, or a functional equivalent, will be done in followup PRs. a.k.a. other Epic subtasks.

src/dataExplorer/context/results/childResults.tsx
```diff
@@ -0,0 +1,125 @@
...
+    setResult(resultFromParent)
+  }, [resultFromParent])
+
+  useEffect(() => {
+    if (resource?.language !== LanguageType.SQL) {
+      return
+    }
+
+    // TODO: tranform functions, based on viewOptions change
+    const text = transformSmoothing(queryText)
```
wiedld, line 79: TODO in followup PRs (epic subtasks).

src/dataExplorer/context/resultsView.tsx
```diff
@@ -0,0 +1,63 @@
...
+  setView: (view: View) => void
+  setViewOptions: (viewOptions: Object) => void
+}
+
+const DEFAULT_STATE: ResultsViewContextType = {
+  view: {
+    state: 'table',
+    properties: SUPPORTED_VISUALIZATIONS['xy'].initial,
+  },
+  viewOptions: {}, // TODO -- set default options. Such as all tagKeys for `|> group()`
```
wiedld, line 27: TODO in followup PRs (other Epic subtasks).

src/dataExplorer/context/results/childResults.tsx
```diff
@@ -0,0 +1,129 @@
...
+    setResult(resultFromParent)
+  }, [resultFromParent])
+
+  useEffect(() => {
+    if (resource?.language !== LanguageType.SQL) {
+      return
+    }
+
+    // TODO: tranform functions, based on viewOptions change
+    const sqlQueryModifiers = modifiersToApply(viewOptions)
```
wiedld, line 82: TODO in followup PRs (other Epic subtasks).

src/dataExplorer/components/Results.tsx
hoorayimhelping: Since this is still production code, I'd say either remove the commented lines, or create a placeholder `SubQueryOptions` component that doesn't have any functionality in it similar to how you're doing `subQueryOptions = null`

src/dataExplorer/components/Results.tsx
```diff
@@ -144,58 +124,157 @@ const Results: FC = () => {
...
+  const [search, setSearch] = useState('')
+  const {status} = useContext(ResultsContext)
+  const {view, setView} = useContext(ResultsViewContext)
+
+  let resultView
+
+  if (status === RemoteDataState.NotStarted) {
+    resultView = <EmptyResults />
+  } else {
+    if (view.state === 'table') {
```
hoorayimhelping, line 256: This is a good candidate to make into a TypeScript [enum](https://www.typescriptlang.org/docs/handbook/enums.html) so we don't have to do raw string comparisons.

src/dataExplorer/components/Results.tsx
```diff
@@ -144,58 +124,157 @@ const Results: FC = () => {
...
+        state: 'table',
+        properties: SUPPORTED_VISUALIZATIONS['simple-table'].initial,
+      })
+    }
+  }
+
+  const Header =
     view.state === 'table' ? (
       <>
         <div style={{width: '300px'}}>
```
hoorayimhelping, line 280: For a minor re-rendering optimization, this can be pulled out into a variable defined outside the `GraphHeader` component. As it stands, this will cause unnecessary re-renders, because JavaScript treats the object literal `{width: '300px'}` as not the same object when doing comparisons on re-renders because each re-render creates a new object which doesn't have the same memory address as the object created on the previous re-render.

This can be solved by storing this in a variable outside of the component it's in and used in place of the object literal:

```ts
const headerStyle = {width: '300px'}
const GraphHeader: FC = () => {
...

      <div style={headerStyle}
```

The only reason I even bring up such a minor optimization is that this code is on the render path of graphs, which can have a tendency to get janky and chuggy.

src/dataExplorer/components/Results.tsx
```diff
@@ -212,52 +291,16 @@ const Results: FC = () => {
...
-      })
-    }
-  }
+    ) : (
+      <GraphHeader />
+    )
 
   return (
     <div className="data-explorer-results">
       <FlexBox direction={FlexDirection.Column} style={{height: '100%'}}>
```
hoorayimhelping, line 300: same bit here about inline styles in object literals

src/dataExplorer/context/results/childResults.tsx
```diff
@@ -0,0 +1,129 @@
...
+        vars: rangeToParam(range),
+      },
+      {
+        language: resource?.language ?? LanguageType.SQL,
+        bucket: selection.bucket,
+        sqlQueryModifiers,
+      }
+    )
+      .then(r => {
+        setResult(r)
```
hoorayimhelping, line 101: I'd prefer if this was `result` rather than `r`. Even though it's fairly clear what `r` represents, I think there's little downside to just calling it `result`. It is certainly easier to search `result` than `r` if you're ever globally searching for the pattern `\.then\(result`

src/dataExplorer/context/results/childResults.tsx
```diff
@@ -0,0 +1,129 @@
...
+        setStatus(RemoteDataState.Done)
+      })
+      .catch(e => {
```
hoorayimhelping, line 104: This should definitely be `error` or `err`. `e` is quite overloaded in JavaScript. It can represent `error` or `event` and as the most common letter in the English language, searching for `e` in an entire codebase is impossible.

(I know this code isn't currently running, but better to change this now so we don't have to worry about it in the future).

hoorayimhelping (commented):
This works for me. Have a few minor pieces of feedback.

The concern about the deeply nested pyramid of doom in `FluxQueryBuilder.tsx` still exists in my mind. I'm not sure if I have any constructive feedback around that beyond that though.

src/dataExplorer/components/Results.tsx
wiedld: yaz. I like where your head's at.

src/dataExplorer/components/Results.tsx
wiedld: I learned something. Yes!

src/dataExplorer/components/Results.tsx
```diff
@@ -71,33 +72,12 @@ const EmptyResults: FC = () => {
...
-  const [search, setSearch] = useState('')
+const TableResults: FC<{search: string}> = ({search}) => {
+  const {result, status} = useContext(ResultsContext)
   const {range} = useContext(PersistanceContext)
-  const {result, status, view, setView} = useContext(ResultsContext)
-  const {launch} = useContext(SidebarContext)
+
   const res = useMemo(() => {
-    if (view.state === 'graph' || !search.trim() || !result?.parsed) {
+    if (!search.trim() || !result?.parsed) {
```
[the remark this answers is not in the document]
wiedld: Agreed. Making the change.

hoorayimhelping (approved):

PR #6487, opened 2023-01-13
chore: remove cardinality option from cancel service context if org type is iox

wdoconnell:
Helps with [AIM 7046](https://github.com/influxdata/quartz/issues/7046)

When canceling service in PAYG, the user is presented with options to identify why they're leaving. This PR drops the cardinality 'reason' if the org is iox (since cardinality wouldn't be a reason).

The remaining changes reflect some tidying up and readability improvements.

Not IOx Org
--
Screen Shot 2023-01-17 at 5 38 30 PM

IOx Org
--
Screen Shot 2023-01-17 at 6 06 35 PM

### Checklist

### DIFF
--- a/cypress/e2e/cloud/billing.test.ts
+++ b/cypress/e2e/cloud/billing.test.ts
@@ -15,26 +15,23 @@ describe('Billing Page Free Users', () => {
 
   beforeEach(() => {
     cy.flush().then(() => {
-      cy.signin()
-    })
-  })
-
-  const setupFreeTest = () => {
-    cy.get('@org').then(({id}: Organization) => {
-      cy.quartzProvision({
-        accountType: 'free',
-      }).then(() => {
-        cy.visit(`/orgs/${id}/billing`)
-        cy.getByTestID('billing-page--header').should('be.visible')
+      cy.signin().then(() => {
+        cy.get('@org').then(({id}: Organization) => {
+          cy.quartzProvision({
+            accountType: 'free',
+          }).then(() => {
+            cy.visit(`/orgs/${id}/billing`)
+            cy.getByTestID('billing-page--header').should('be.visible')
+          })
+        })
       })
     })
-  }
+  })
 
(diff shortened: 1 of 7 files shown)


src/billing/components/PayAsYouGo/CancelServiceContext.tsx
```diff
@@ -24,6 +24,12 @@ const RedirectLocations = {
 }
 const DEFAULT_REDIRECT_LOCATION = '/mkt_cancel'
 
+const getDefaultCancelationReason = () => {
+  return Object.keys(CancelationReasons).filter(
+    reason => CancelationReasons[reason] === CancelationReasons.NONE
+  )[0]
+}
+
```
wdoconnell, line 32: This (which is a little gross) is to resolve the issue with line 48. Enums are being used to track the user's selected 'cancelation reason', and 'reason' represents the current _key_ of the enum reflecting that cancelation reason. Having that remain a string like 'NO_OPTION' is a big no-no imho, because someone changing the enum won't know to search for that string.

The above ^ is marginally better, because even though the logic of what it's doing is stupid (find me the name of the key that corresponds to `CancelationReasons.NONE`, it isn't expensive in a way we care about, and at least means that changes to the enum names will work correctly without having to run more manual searches.

cypress/e2e/cloud/billing.test.ts
```diff
@@ -238,4 +234,32 @@ describe('Billing Page PAYG Users', () => {
     // TLDR; we double confirm here, this is by design. The overlay changes to reflect a new state so this isn't an error in the test
     cy.getByTestID('cancel-service-confirmation--button').click()
   })
+
+  it('does not display high cardinality as a cancelation reason in iox orgs', () => {
+    cy.isIoxOrg().then(isIOx => {
+      cy.skipOn(!isIOx)
+    })
```
wiedld, line 241: woot!

Issue #6482, opened 2023-01-12
EPIC: implement SQL data viz for QxBuilder


wiedld:
Reasoning behind this EPIC is [documented in this spike](https://github.com/influxdata/ui/issues/6085#issuecomment-1380974158).

In brief, we are splitting by the table query vs the graph query. Then we apply "magic" to make the graphs work. 

## Tasks
Each checkbox (tasks) can be done independently. a.k.a. is a minimum merge-able unit, without a feature flag.
- [x] make the query for the table, be different than the query for the graph.
   - https://github.com/influxdata/ui/issues/6483
- [x] if data is truncated for graph query --> display message in graph view
- [x] when choosing graph --> auto-open the Customize options in right hand panel.
   * for SQL and flux
- [x] for graph view, add option for data smoothing:
   * for SQL
   * view:
       * Dropdown: nothing, fieldA, fieldB, fieldC
       * ? description of what it does
   * state:
      * persist graph options
      * dataSmoothing: null | fieldName
- [x] Handling tagKeys:
   - https://github.com/influxdata/ui/issues/6496
   - checkboxes. (any_of)
   - by default, append `|> group([<all-tagKeys>])` to each iox.sql() query
       * add button to "show ad hoc query sent"
   - Add grouping checkboxes in Customize graph options:
       * toggle on/off which tags to include in the groupings
- [x] Handling fields:
   -  radio button. (one_of)
   -  By default set a field to view
   -  field --> `valColumn` in `polyline.rdp()`
   -  onFieldSelect() → place in persistence context
- Handling axis defaults:
   - [x] ~X axis default: `_time` or `time`~ Already done. 
   - [x] Y axis default (order to preference).
- [x] affordance → see "my query sent for the graph"
- [x] fix discovered bug:
   -  https://github.com/influxdata/ui/issues/6503
- [ ] All the tests!

## Example with polyline (data smoothing):
Outcome produces same `_time` and `_value`.
Most of the graphs just work. (Or at least, "work" to the same level as it does in flux.)

https://user-images.githubusercontent.com/10232835/212179155-b2abaa02-3ef9-4f6b-a343-f030ed730e87.mov

## Example without polyline:
Need to choose the default option (e.g. Y column). We will be programmatically choosing the initial default.
Fewer of the graphs will work, without the user doing additionally efforts on their own -- such as modifying the graph options and/or the query.
(Note: the default is polyline being on.)

[...]

PR #6502, opened 2023-01-18
feat(6496): SQL graphing viewOptions: groupby tag keys

### LINKED ISSUE #6496: Add groupby tagKeys to the SQL editor.
Subtask of Epic #6482 

Handling tagKeys:
* checkboxes. (any_of)
* by default, append `|> group([<all-tagKeys>])` to each iox.sql() query
     * add button to "show ad hoc query sent"
* Add grouping checkboxes in Customize graph options:
     * toggle on/off which tags to include in the groupings

### LINKED ISSUE #6503: Graph customize options do not update with changes in schema selection.
In the qx builder.
The graph viz options (e.g. which Y-axis is chosen) do not update as new schema is selected. This is the bug.

(In comparison, the query options located just above -- do change with schema selection.)

https://user-images.githubusercontent.com/10232835/213073834-a29bff2e-df21-433d-a46a-c946e1329cdc.mov

wiedld:
Part of #6482 
Closes #6496 
Closes #6503

Add ability to groupby tagKeys (a.k.a. into dataSeries) for SQL queries being graphed.

## Qx contract:
* how many queries (network calls) are run:
   * when I change by query viewOptions => re-run the query
   * if I change queryText => have no viewOptions => only 1 query with run (see network tab in vid)
   * if I change queryText => have viewOptions => 2 queries will run (one for table, one for graph)
* how are the groupby options chosen?
   * when produce results => use the table column headers for the groupby options
       * **THIS IS THE SAME CONTRACT** as displaying the table data. You have to re-run the query to produce results. 
   * set the default groupby, to be the tagKeys
        * once we choose the bucket & measurement => we can then know the tagKeys for the default

## Viz evidence: How many queries?

https://user-images.githubusercontent.com/10232835/213072458-f8997199-aa56-4bdd-8f45-2f909c467104.mov

## Viz Evidence: How groupby options chosen?

https://user-images.githubusercontent.com/10232835/213300460-51ccc49c-13e6-4b1c-8710-3aae7f03acd8.mov

## Checklist

### DIFF
--- a/src/dataExplorer/components/Results.tsx
+++ b/src/dataExplorer/components/Results.tsx
@@ -1,10 +1,10 @@
 import React, {
   FC,
-  useState,
-  useContext,
-  useMemo,
   useCallback,
+  useContext,
   useEffect,
+  useMemo,
+  useState,
 } from 'react'
 import {
   FlexBox,
--- a/src/dataExplorer/components/Results.tsx
+++ b/src/dataExplorer/components/Results.tsx
@@ -192,6 +192,7 @@ const WrappedOptions: FC = () => {
   const {view, setView, selectViewOptions, viewOptions, selectedViewOptions} =
     useContext(ResultsViewContext)
   const {resource} = useContext(PersistanceContext)
+  const dataExists = !!result?.parsed
 
   const updateChildResults = useCallback(
     update => {
@@ -206,8 +207,13 @@ const WrappedOptions: FC = () => {
     [setStatus, setResult]
   )
 
+  if (!dataExists) {
+    return null
+  }
+
   const subQueryOptions =
-    resource?.language === LanguageType.SQL ? (
+    resource?.language === LanguageType.SQL &&
+    view.state == ViewStateType.Graph ? (
       <SqlViewOptions
         selectViewOptions={selectViewOptions}
         allViewOptions={viewOptions}
@@ -231,7 +237,7 @@ const GraphHeader: FC = () => {
   const {view, setView, viewOptions} = useContext(ResultsViewContext)
   const {result} = useContext(ResultsContext)
   const {result: subQueryResult} = useContext(ChildResultsContext)
-  const {launch} = useContext(SidebarContext)
+  const {launch, clear: closeSidebar} = useContext(SidebarContext)
 
   const dataExists = !!result?.parsed
 
@@ -243,6 +249,11 @@ const GraphHeader: FC = () => {
       launcher()
     }
   }, [viewOptions])
+  useEffect(() => {
+    if (!dataExists) {
+      closeSidebar()
+    }
+  }, [dataExists])
 
   const updateType = viewType => {
     setView({
--- a/src/dataExplorer/components/Results.tsx
+++ b/src/dataExplorer/components/Results.tsx
@@ -9,7 +9,17 @@ import {
   ComponentColor,
 } from '@influxdata/clockface'
 
-import {RemoteDataState, SimpleTableViewProperties} from 'src/types'
+// Components
+import {SearchWidget} from 'src/shared/components/search_widget/SearchWidget'
+import {
+  View,
+  ViewTypeDropdown,
+  ViewOptions,
+  SUPPORTED_VISUALIZATIONS,
+} from 'src/visualization'
+import {SqlViewOptions} from 'src/dataExplorer/components/SqlViewOptions'
+
+// Contexts
 import {ResultsContext} from 'src/dataExplorer/context/results'
 import {
   ResultsViewContext,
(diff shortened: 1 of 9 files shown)


src/dataExplorer/components/Results.tsx
```diff
@@ -200,14 +228,21 @@ const WrappedOptions: FC = () => {
...
+  const dataExists = !!result?.parsed
+
   const launcher = () => {
     launch(<WrappedOptions />)
   }
+  useEffect(() => {
+    if (dataExists) {
+      launcher()
+    }
+  }, [viewOptions])
```
wiedld, line 245: This causes the sidebar (graph customize) to autolaunch when the graph view is open.

src/dataExplorer/components/Results.tsx
```diff
@@ -1,4 +1,11 @@
-import React, {FC, useState, useContext, useMemo, useCallback} from 'react'
+import React, {
+  FC,
+  useState,
+  useContext,
+  useMemo,
+  useCallback,
+  useEffect,
```
hoorayimhelping, line 7: Minor feedback that goes along way, but absent any other organizing scheme, I prefer to alphabetize these lists. It's one of those small annoyances at write time that have an outsized impact on ease of reading comprehension

src/dataExplorer/context/resultsView.tsx
```diff
@@ -47,19 +65,67 @@ export const ResultsViewProvider: FC = ({children}) => {
...
 
-  const [viewOptions, setViewOptions] = useSessionStorage(
+  // what can be chosen (e.g. the list of all options)
+  const [viewOptionsAll, saveViewOptionsAll] = useSessionStorage(
+    'dataExplorer.resultsOptions.all',
+    DEFAULT_VIEW_OPTIONS
+  )
+  const setViewOptionsAll = useCallback(
+    (updatedOptions: Partial<ViewOptions>) => {
+      saveViewOptionsAll({...viewOptionsAll, ...updatedOptions})
```
hoorayimhelping, line 75: `setViewsOptionAll` and `saveViewOptionsAll` are really similarly named, and it takes a little bit of digging to understand the difference between the two. I can't think of any conventions around using set vs save - they're pretty interchangeable, so to avoid confusion, it might be worth calling this method `saveViewOptionsAllMemoized` to know that the only difference is this is memoized on `viewOptionsAll`

src/dataExplorer/context/resultsView.tsx
```diff
@@ -47,19 +65,67 @@ export const ResultsViewProvider: FC = ({children}) => {
...
+    },
+    [viewOptionsAll]
+  )
+
+  // default options (a.k.a. based on schema)
+  const [defaultViewOptions, saveDefaultViewOptions] = useSessionStorage(
+    'dataExplorer.resultsOptions.default',
+    DEFAULT_VIEW_OPTIONS
+  )
+  const setDefaultViewOptions = useCallback(
```
hoorayimhelping, line 85: if you choose to name `setViewOptionsAll` as something like `saveViewOptionsAllMemoized`, this should also be updated to follow the same convention

src/dataExplorer/context/resultsView.tsx
```diff
@@ -47,19 +65,67 @@ export const ResultsViewProvider: FC = ({children}) => {
...
+  )
+
+  useEffect(() => {
+    // if parent query is re-run => decide what to reset in subquery viewOptions
+
+    // reset groupby
+    const excludeFromGroupby = NOT_PERMITTED_GROUPBY
+    const groupby = Object.keys(
+      resultFromParent?.parsed?.table?.columns || {}
+    ).filter(g => !excludeFromGroupby.includes(g))
```
hoorayimhelping, line 111: What does `g` represent here? Shouldn't this be `c`, since this is an array of column keys? In fact, `c` is a bit ambiguous as well. I think these single variables should be given full names to describe what it is these `filter` functions are doing

src/dataExplorer/context/resultsView.tsx
```diff
@@ -47,19 +65,67 @@ export const ResultsViewProvider: FC = ({children}) => {
...
+    setViewOptionsAll({groupby})
+    const defaultsWhichExist = defaultViewOptions.groupby.filter(g =>
+      groupby.includes(g)
+    )
```
hoorayimhelping, line 115: Same bit here, `g` should be `o`, but it's probably much better to call `g` `viewOption` or `option` instead of `o`

hoorayimhelping (approved):
A couple of minor things, but all-in-all it looks good

src/dataExplorer/components/Results.tsx
wiedld: I'll do it for this one example here. But note that there are linter rules which can enforce, and auto-fix for us.

src/dataExplorer/context/resultsView.tsx
wiedld: I'm hesitant to use the word `memoized`, as in the definition is to remember -- and the ALL options is reset rather frequently (with every parent query run). Depending on the user flow, they could be memoizing the ALL (versus the DEFAULT) more or less often.

Will muse on another naming convention.

src/dataExplorer/context/resultsView.tsx
wiedld: Here it's definitely the columnName. Can make that explicit.

src/dataExplorer/context/resultsView.tsx
wiedld: Yeah, exactly. It's not necessarily a tagKey. Just any default groupingOption.

src/dataExplorer/context/resultsView.tsx
wiedld: I was going for `set` == "add to setting", and `save` == "save to local persistance". 🤷🏼

src/dataExplorer/context/resultsView.tsx
wiedld: See if [this last commit added](https://github.com/influxdata/ui/pull/6502/commits/eff939c4937828baa12cc723736fb5a09fed6b36), helps to clarify.

wiedld: Also, here is some manual QA on the behavior for `clear()`.
(Note: we really need to put in tests sometime soon-ish.)


https://user-images.githubusercontent.com/10232835/213325137-697cda14-a337-46ba-8105-eb518176b235.mov

hoorayimhelping: Cool, works for me! Thanks for thinking about this critically!

hoorayimhelping (approved):
Shipit!

PR #6284, opened 2022-11-08
perf: reduce jank in graph options sliders

hoorayimhelping:
Closes #6283

### The problem
Jankiness was caused by updating the redux store every time an input in the options panel changed. In text boxes: every keystroke. In sliders: every time the slider was moved. We'd dispatch an action, which would update the redux store, which would cause re-renders which would make the UI appear sluggish.

https://user-images.githubusercontent.com/146112/200597573-90d80871-86c2-4f90-94e5-f7dea7621421.mov

### The fix
Store these changes in local component state, then debounce the call to update redux state so that the visualization will be updated after the input stops.

https://user-images.githubusercontent.com/146112/200598326-7e465d5e-b387-4cde-9f07-c418a017c614.mov

### Tested in:

- [x] Existing dashboard cell
- [x] New dashboard cell
- [x] Data explorer Classic™
- [x] New notebook
- [x] Existing notebook
- [x] New query builder experience

### Checklist

Authors and Reviewer(s), please verify the following:

- [x] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [x] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [x] Documentation updated or issue created (provide link to issue/PR)
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [ ] Feature flagged, if applicable

PR #6507, opened 2023-01-19
feat(6482): Message separately if graph subquery hits limit

wiedld:
Part of #6482 

The table view tab shows the raw output of the SQL query.
The graph view tab shows the graph subquery, based on QueryModifiers chosen on the upper right sidebar.

In this PR, we add the ability to adjust the amount of data retained when using the data smoothing. Since the graph subquery is separate from the raw table query, we also implemented a separate message when the graph subquery had truncated data.

The net result:
The user can slide down to retain fewer data points (in the smoothing) --> resulting in the entire timespace being retained for all dataseries. (Versus the raw table data still being truncated.)

## DONE:
* add slider to data smoothing, such that can retain more vs less data points
* surface separate msg when data is truncated, based on the graph subquery
    * this is different than the table truncation message. 
Such that the user can see when adding more/less data points --> results in truncation of the graph subquery.

## Visual proof -- have separate truncation messages:
Table view shows full QueryStat (including table & row counts).
Graph view shows only the "truncated message".

https://user-images.githubusercontent.com/10232835/213330216-f1c87424-0729-4a22-ba57-e17785dea83a.mov

## Visual proof -- the slider adjusts the number of point retained, which impacts whether or not the truncation limit is hit:
(Note: this was done with artificially lowered data limits.)

https://user-images.githubusercontent.com/10232835/214391534-4698961e-69f2-47e4-a0e6-04790fceee06.mov

## Checklist

### DIFF
--- a/src/dataExplorer/components/SqlViewOptions.tsx
+++ b/src/dataExplorer/components/SqlViewOptions.tsx
@@ -41,6 +41,11 @@ export const SqlViewOptions: FC<SqlViewOptionsT> = ({
     }
   }
 
+  const handleSmoothingRangeSlider = event =>
+    selectViewOptions({
+      smoothing: {percentageRetained: parseInt(event.target.value)},
+    })
+
   const groupbyTooltipContents = (
     <div>
       <span>Select the GROUPBY used for the graph subquery.</span>
@@ -132,11 +137,7 @@ export const SqlViewOptions: FC<SqlViewOptionsT> = ({
               min={5}
               max={100}
               step={5}
-              onChange={event =>
-                selectViewOptions({
-                  smoothing: {percentageRetained: parseInt(event.target.value)},
-                })
-              }
+              onChange={handleSmoothingRangeSlider}
               labelPrefix="retained "
               labelSuffix="%"
             />
--- a/src/dataExplorer/components/SqlViewOptions.tsx
+++ b/src/dataExplorer/components/SqlViewOptions.tsx
@@ -5,6 +5,7 @@ import {
   ComponentStatus,
   FlexBox,
   Grid,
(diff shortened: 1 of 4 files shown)


src/dataExplorer/components/SqlViewOptions.tsx
```diff
@@ -126,6 +127,19 @@ export const SqlViewOptions: FC<SqlViewOptionsT> = ({
...
+              min={5}
+              max={100}
+              step={5}
+              onChange={event =>
+                selectViewOptions({
+                  smoothing: {percentageRetained: parseInt(event.target.value)},
+                })
+              }
+              labelPrefix="retained "
+              labelSuffix="%"
```
wiedld, line 141: Labels were added **after** the vids were made.
Screen Shot 2023-01-24 at 11 42 46 AM

hoorayimhelping: If this behaves like any of the other sliders in graph options, I think you might want to consider setting the value locally and then calling `selectViewOptions` after some amount of time, similarly to how it's done [here](https://github.com/influxdata/ui/pull/6284/files#diff-6bf175e958e1b62be5e473e5061ea15d322054a8a0e7e6ab18610b85c9cc6c48R163). I found that updating remote state that the graph depends on causes the experience to be incredibly janky.

hoorayimhelping (approved):

src/dataExplorer/components/SqlViewOptions.tsx
wiedld: Good call. Definitely improves it.

hoorayimhelping (approved):

Issue #5207, opened 2022-07-28
Transfer Tokens/Alerts From Removed User


Luckette:
**ACs**

- User can click the trash can icon per member (row) on the org members list to remove the teammate
- Upon clicking the trash can, display a modal (see Figma: https://www.figma.com/file/fvVSKwnepmRWvzFlIxCWCe/Multi-org?node-id=1412%3A10037 )
- Top of the modal should display a warning message stating that "When you remove a teammate from an organization, you need to transfer their tokens and alerts to another teammate so scripts are uninterrupted."
- Next, allow the user to select a user to transfer tokens and alerts to (Typeahead dropdown of all active users in the org) 
- The "Remove Teammate" button on the modal should be disabled until they've chosen a teammate to transfer their tokens and alerts to. 
- Display a toast-like notification when a user is removed from org stating "‘user@company.com’ successfully removed from [Organization Name]. API tokens and alerts transferred to ‘user@company.com’        
- Match colors, icons, etc. that are in the Figma.

wdoconnell:
Converting the context menus to a modal will be completed by UI #5206, but the tokens/alerts transfer is blocked and cannot be completed without additional quartz work.

There is an [idpe endpoint](https://github.com/influxdata/openapi/blob/master/src/cloud/schemas/ProvisionUserDeleteRequest.yml) to handle transfer of tokens and alerts. After testing in remocal and discussion with @gavincabbage, my understanding is that this is implemented in a way that could be called by quartz using an operator token, but cannot be called by Cloud2 UI users.

PR #6509, opened 2023-01-19
chore: add new idpe transfer property to deleteOrgsUser

### LINKED ISSUE #6473: Replace use of old 'remove user' endpoint with new quartz user eviction api
The UI uses `deleteOrgsUser` (api/v2/quartz/orgs/:orgId/users/:userId) to remove users from an organization. Quartz has a new user eviction API that (optionally) transfers the removed user's idpe objects (eg tasks, alerts) to a new user. Replace all references to `deleteOrgsUser` with the new quartz user eviction API.

This is the [ticket](https://github.com/influxdata/quartz/issues/7094) for tracking progress of exposing new user eviction API in quartz.

wdoconnell:
Closes #6473 

The quartz user eviction API now allows transfer of a removed user's IDPE objects (tasks, alerts) to a new user. We should be using that in the UI now. Fortunately, we don't need to modify the endpoint we're using, as the transfer-option has been added as an optional parameter to the existing API call - we just need the openapi spec updated. Thanks @abshierjoel 

This unblocks [AIM 5207](https://github.com/influxdata/ui/issues/5207)

### Checklist


abshierjoel (approved):

hoorayimhelping (approved):
probably the greatest single diff i've ever seen. the way you changed the SHA in this PR really moved me 😏

PR #6512, opened 2023-01-19
chore: wrap users provider around org profile tab

### LINKED ISSUE #6511: Self-Removal from Org on Account Page But Not Org Tab
With multi-org enabled, the "Leave Org" button (where a user removes _themselves_ from an org, which is allowed only if there are other users) has been moved from the Account Page (which didn't make sense) to the Org Profile Page (which is more logical). On navigating to the org profile page, however, I can't currently self-remove.

The reason is that the Org Page isn't wrapped in the `UsersProvider` like the Account Tab is. So when the check looks for whether `users > 1`, it evaluates to `false` because the `users` property is `0` without being wrapped in the Provider.

We need to wrap that provider around the Org Tab.

## About the bug

**Steps to reproduce:**

1. With multiorg on, navigate to org profile page
2. No "leave org" allowing the current user to leave the org

**Expected behavior:**
"Leave org" button should be present

**Visual Proof:**

Screen Shot 2023-01-19 at 5 17 12 PM

wdoconnell:
Closes #6511

Resolves issue with org profile tab where "leave org" button doesn't appear to allow user to leave org with other users in it.

The UsersProvider needs to be moved up a level, from within the Org Profile Tab to above the Org Profile Tab to ensure it's possible for the "leave org" button to display.

Before
--
Screen Shot 2023-01-19 at 5 22 39 PM

After
--
Screen Shot 2023-01-19 at 5 22 56 PM

### Checklist

### DIFF
--- a/src/organizations/components/OrgProfileTab/index.tsx
+++ b/src/organizations/components/OrgProfileTab/index.tsx
@@ -20,7 +20,7 @@ import {orgDetailsFetchError} from 'src/shared/copy/notifications'
 import {notify} from 'src/shared/actions/notifications'
 
 // Providers
-import {UsersContext, UsersProvider} from 'src/users/context/users'
+import {UsersContext} from 'src/users/context/users'
 
 // Selectors
 import {getMe} from 'src/me/selectors'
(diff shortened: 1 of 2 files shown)


eatondustin1 (dismissed):
LGTM!

wiedld (approved):
I trust those whom have walked before. 😆

PR #6526, opened 2023-01-25
chore(Tasks Deprecation): conditionally show Tasks in the UI

### LINKED ISSUE #6518: Hide Tasks Feature from any New Accounts
## User Story

<p></p>

As an organization with a new SQL optimized storage engine, I want to prohibit any sign\-ups after Jan 31 from seeing the tasks menu item so that they don't start using Flux when the DB is optimized for SQL\.  I want to feature flag control this in case I need to make any exceptions\.

<p></p>

## Acceptance Criteria

<p></p>

- A feature flag will be created to control whether or not the Tasks feature is shown in the main menu of the UI

- If the sign\-up date for the user is after Jan 31st and the feature flag is false, then the tasks menu item will not be shown

- If the feature flag is true, or the sign\-up date is before Jan 31st, then the tasks menu item will be shown

- These changes will only apply to IOx Enabled Orgs\.  For non\-IOx orgs, this hiding should not occur\.

**Flow chart**   
Screen Shot 2023-01-26 at 12 19 34 PM

<p></p>

<p></p>

blegesse-git:
Closes #6518 

Behind feature flag `hideTasks` 
Pr hides Tasks from `CLOUD` users whose orgs that are IOx enabled and are created after 1/31/23 
For those orgs that meet this requirement, they will no longer see Tasks in Navigation bar, can no longer use a link to route to Tasks, or save their queries as Tasks. 

**Orgs created _before_ 1/31/23 with `hideTasks` feature flag off:**
https://user-images.githubusercontent.com/66275100/214604026-9b81364b-a0d3-4586-8c2c-bc0ea625e730.mov
**Orgs created _after_ 1/31/23 with `hideTasks` feature flag on:**
https://user-images.githubusercontent.com/66275100/214604116-43f8fff0-65a9-4ed5-9986-b8ab933f00de.mov

fl

### Checklist


eatondustin1 (approved):
LGTM!

src/tasks/selectors/tasksSelectors.test.ts
```diff
@@ -0,0 +1,69 @@
...
+}))
+
+import {selectOrgCreationDate} from 'src/organizations/selectors'
+
+describe('selecting whether tasks should be shown', () => {
+  let fakeState
+  beforeEach(() => {
+    jest.resetAllMocks()
+    jest.resetModules()
+
```
wdoconnell, line 24: What's the upside of calling `resetModules()` here?

src/tasks/selectors/tasksSelectors.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  })
+  it('does not show tasks when the org creation date is the exact same as the IOx cutoff date', () => {
+    jest
+      .mocked(selectOrgCreationDate)
+      .mockImplementationOnce(() => '2023-01-31T00:00:00Z')
+    expect(selectShouldShowTasks(fakeState)).toBeFalsy()
+  })
+
+  it('shows tasks regadless of org creation date in OSS', () => {
+    jest.mock('src/shared/constants/index', () => ({
```
wdoconnell, line 56: Minor typo above ^, but looks good, since I understand we're not deprecating this outside of the cloud context.

src/tasks/selectors/tasksSelectors.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  beforeEach(() => {
+    jest.resetAllMocks()
+    jest.resetModules()
+
+    fakeState = {
+      resources: {
+        orgs: {
+          org: {
+            createdAt: '2023-02-01T00:00:00:000Z',
+          },
```
wdoconnell, line 30: Do we need this `createdAt` value, since we're mocking the return value each time anyway?

wdoconnell (commented):
Couple of minor comments, but looks good.

blegesse-git:
Since we're on a time crunch, I'll be merging this as is. I'll be creating a separate issue that will address comments posted by @wdoconnell.

src/tasks/selectors/tasksSelectors.test.ts
blegesse-git: yikes

hoorayimhelping:
Do we need an affordance in the check to see if they have active tasks?

blegesse-git:
> Do we need an affordance in the check to see if they have active tasks?

According to Gary we don't need to check whether they have tasks or not or factor that into this logic. Can you confirm @garylfowler ?

hoorayimhelping (approved):

PR #6534, opened 2023-01-27
feat(6447): validate flux before persistance

### LINKED ISSUE #6447: Use is_valid_flux() in ui, instead of the improper use of parse().

wiedld:
Closes #6447 

Whenever we can manually create a flux query, we should validate that it's correct before persistence.

## Done:
* housekeeping:
   * renamed directory `parser` as `lspUtils`, since that's what it is.
   * make the `RESOURCES` object be statically typed, and not dynamically loaded. 
* validate flux prior to persistence in:
   * scripts
   * tasks
 

## Not Done:
* doesn't use raw flux editor, so not validating:
   * checks
   * notificationRules
* Not doing yet, future PR?
   * dashboards, since we:
      * would need more code to handle variables in the extern
      * May not be worth it. Dashboards is not run as frequently, since is a pull (a.k.a. dashboard loading) instead of a continuously running task/check/notificationRule.
      * If we do validate in the future, we would want to do it:
         * only on updates to dashboard cells
         * since we ourselves (in new tools) have broken queries. So don't stop the entire dashboard.
   * notebooks

## Checklist


src/tasks/actions/thunks.ts
```diff
@@ -399,6 +403,9 @@ export const cloneTask = (task: Task) => async (dispatch: Dispatch<Action>) => {
     })
 
     const fluxWithNewName = format_from_js_file(ast)
+    if (!isValidFlux(fluxWithNewName)) {
+      throw new Error('Invalid flux script. Please check your query text.')
```
wiedld, line 407: This one is not tested, since it's cloning an existing task which has invalid flux.
Since we can no longer create tasks which are invalid, we cannot create an invalid task to the be cloned.

Instead, this is something the users will encounter in the wild => when they attempt to clone invalid flux tasks.

src/tasks/containers/TaskPage.tsx
```diff
@@ -144,6 +147,11 @@ class TaskPage extends PureComponent<Props> {
     // currently we delete that part of the script
     script = script.replace(new RegExp('option\\s+task\\s+=\\s+{(.|\\s)*}'), '')
 
+    if (!isValidFlux(script)) {
+      this.props.invalidFlux()
+      return
```
wiedld, line 152: Doing the check here in the TasksPage, allows us to keep the "Create Task" open on failure. Therefore, the user doesn't "lose their work" as they try creating a new task.

src/dataExplorer/components/resources/types/script/index.ts
```diff
@@ -16,8 +16,8 @@ const {getScript, patchScript, postScript} = CLOUD
       postScript: false,
     }
 
-export default function script(register) {
-  register({
+export default () => {
```
appletreeisyellow, line 19:
```suggestion
export const scriptResourceRegistration = () => {
```

Suggest to use named exports to help with https://github.com/influxdata/ui/issues/6007

src/dataExplorer/components/resources/index.ts
```diff
@@ -28,18 +29,6 @@ interface Resources {
...
-      )
-    }
-
-    RESOURCES[def.type] = {
-      ...def,
-    }
-  })
-})
+export const RESOURCES: Resources = {
+  scripts: scriptResourceRegistration(),
```
appletreeisyellow, line 33: This is much easier to understand! Love it ❤️

appletreeisyellow (approved):
Thank you for chasing this down! Just one suggestion on named export, but not a blocker

PR #6535, opened 2023-01-27
fix(NewDatePicker): utilize local vs. UTC time zone

### LINKED ISSUE #6451: New datepicker (used in iox Qx builder) does not utilize timezones.

appletreeisyellow:
Closes #6451 

This PR persists the custom time range in UTC time zone with ISO format for the new date picker. Then the input form converts and displays the time according to the time zone that's selected by users. 

## Before

https://user-images.githubusercontent.com/14298407/215828719-73c89995-53d8-43a2-9765-9545052bde6f.mov

## After

https://user-images.githubusercontent.com/14298407/215828298-fac53073-8a0f-4383-995d-189c01fa0fe3.mov

### Checklist

### DIFF
--- a/src/shared/components/dateRangePicker/NewDatePicker.tsx
+++ b/src/shared/components/dateRangePicker/NewDatePicker.tsx
@@ -1,4 +1,5 @@
 import React, {FC, useState, useContext, useEffect, useCallback} from 'react'
+import moment from 'moment'
 import {
   AlignItems,
   Button,
@@ -21,6 +22,7 @@ import {PersistanceContext} from 'src/dataExplorer/context/persistance'
 
 // Utils
 import {getTimeRangeLabel} from 'src/shared/utils/duration'
+import {isISODate} from 'src/shared/utils/dateTimeUtils'
 import {SELECTABLE_TIME_RANGES} from 'src/shared/constants/timeRanges'
 import {useSelector} from 'react-redux'
 import {getTimeZone} from 'src/dashboards/selectors'
@@ -30,6 +32,15 @@ import {TimeRange} from 'src/types'
 
 const NBSP = '\u00a0\u00a0'
 const MAX_WIDTH_FOR_CUSTOM_TIMES = 325
+
+// if the input date is in ISO format: 'YYYY-MM-DDTHH:mm:ss.sssZ',
+// then convert it to 'YYYY-MM-DD HH:mm'
+const convertToDisplayFormat = (date: string): string => {
+  if (!isISODate(date)) {
+    return date
+  }
+  return moment(date).format('YYYY-MM-DD HH:mm')
+}
 
 interface Props {
   onCollapse: () => void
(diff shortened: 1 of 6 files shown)


src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -120,14 +146,6 @@ const DatePickerMenu: FC<Props> = ({onCollapse, timeRange, timeRangeLabel}) => {
...
   const handleSelectDate = (dates): void => {
     const [start, end] = dates
-    if (timeZone === 'UTC') {
-      if (start) {
-        start.setMinutes(start.getMinutes() + start.getTimezoneOffset())
-      }
-      if (end) {
-        end.setMinutes(end.getMinutes() + end.getTimezoneOffset())
-      }
-    }
```
appletreeisyellow, line 130: No need to adjust time with the time zone offset since the new date picker set the time from 00:00 to 23:59 regardless of the time zone. See the code here https://github.com/influxdata/ui/blob/c5ebfaf890fa29c0ccc37f9d220d35abda686f83/src/shared/components/dateRangePicker/NewDatePicker.tsx#L135-L164

src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -296,7 +326,7 @@ const DatePickerMenu: FC<Props> = ({onCollapse, timeRange, timeRangeLabel}) => {
                     : ComponentStatus.Error
                 }
                 type={InputType.Text}
-                value={inputStartDate ?? ''}
+                value={convertToDisplayFormat(inputStartDate, timeZone) ?? ''}
```
appletreeisyellow, line 329: The persisted time is always in UTC time zone, but users might have different setting with the time zone, either in "Local" or in "UTC". `convertToDisplayFormat()` convert the time to display according to the time zone

eatondustin1 (dismissed):
Looks good!. The only other place I know we use the date range picker is in the operator UI (Operator UI -> Free or Payg Account -> Convert to Contract in upper right), so it might be worth verifying that functionality still works, since I had quite a bit of trouble getting it to work in the first place. Happy to help test that if you want, but this looks good either way.

appletreeisyellow:
> The only other place I know we use the date range picker is in the operator UI (Operator UI -> Free or Payg Account -> Convert to Contract in upper right), so it might be worth verifying that functionality still works, since I had quite a bit of trouble getting it to work in the first place. Happy to help test that if you want, but this looks good either way.

@eatondustin1 

This PR updates the new date picker component that's only used in the new data explorer. My guess is that the date range picker in the operator UI is an old one. But I'm happy to double check. 

Can you show me how to set up and navigate to the operator UI page? I tried on my remocal to visit `/operator` and it shows page not found. 

Screen Shot 2023-02-02 at 12 06 13 PM

appletreeisyellow:
@eatondustin1 

We don't need to test operator for this PR then. 

Good to know the higher level of testing an operator. I'll reach out to you if any work is scheduled for the old date picker!

wiedld (commented):
Looks good.

If you have time, can you add a test? Just because we are doing string manipulation to handle dates -- and that's prone to future breaking (us breaking ourselves). A single test which checks UTC --> local, and then local --> UTC is good.

appletreeisyellow:
Convert to draft since I'm adding more e2e tests

appletreeisyellow:
Merge this PR after https://github.com/influxdata/ui/pull/6610 is merged since there might be conflicts

wiedld (approved):

PR #6540, opened 2023-01-28
fix(6537): script crud should not be enabled for non-cloud. Scripts e2es cleanup.

### LINKED ISSUE #6537: Update ScriptBuilder to not have script crud features for oss.
## About the bug

The OSS openapi is [currently locked on Dec 2021.](https://github.com/influxdata/influxdb/blob/f78f9eda9ca4cbe1b0e22265235b8b8739b91640/scripts/fetch-swagger.sh#L13) This means that the scripts crud endpoints are not supported -- which reflects a product decision to make scripts (as a crud object) be a cloud-only concept.

Currently in the Script Builder ui, we have this weird state. Most of the SaveAsScript features are available for the oss users, except for the missing single "Save" button. The result are things like this:
Screen Shot 2023-01-27 at 10 40 20 AM

We really should have the SaveAsScript be hidden at the root level, in the toolbar. Since there is no opening of existing scripts, editing of script metadata (e.g. the description) is pointless, there is not an invokable url, etc.

\

wiedld:
Closes #6537 

## Done:
* for bug #6537 
    * Turn off toolbar items related to scripts crud, when not on cloud 
    * add data-Ids to the various UI elements related to scripts crud
* e2es:
    * create a new scripts crud test, for cloud-only
    * show the crud elements are not present in oss.
    * modify scripts test fixtures (a.k.a. cypress commands) to handle additional use cases (i.e. oss, already saved script)

## Test on iox:

Screen Shot 2023-02-08 at 9 07 49 PM

## Checklist

### DIFF
--- a/cypress/e2e/oss/scriptsCrud.test.ts
+++ b/cypress/e2e/oss/scriptsCrud.test.ts
@@ -1,0 +1,17 @@
+describe('Script Builder -- scripts crud on cloud', () => {
+  before(() => {
+    cy.flush().then(() => cy.signin())
+  })
+
+  beforeEach(() => {
+    cy.scriptsLoginWithFlags({}).then(() => {
+      cy.switchToDataExplorer('new')
+    })
+  })
+
+  it('should not have the scripts crud options', () => {
+    cy.getByTestID('script-query-builder--open-script').should('not.exist')
+    cy.getByTestID('script-query-builder--save-script').should('not.exist')
+    cy.getByTestID('script-query-builder--edit-script').should('not.exist')
+  })
+})
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -765,6 +765,14 @@ export const createNotebook = (
 }
 
 export const newScriptWithoutLanguageSelection = () => {
+  const CLOUD = Cypress.env('dexUrl') === 'OSS' ? false : true
+  if (!CLOUD) {
+    return cy
+      .getByTestID('script-query-builder--new-script')
+      .should('be.visible')
+      .click()
+  }
+
   cy.getByTestID('script-query-builder--save-script').then($saveButton => {
     cy.getByTestID('page-title')
       .invoke('text')
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -766,16 +766,26 @@ export const createNotebook = (
 
 export const newScriptWithoutLanguageSelection = () => {
   cy.getByTestID('script-query-builder--save-script').then($saveButton => {
-    if (!$saveButton.is(':disabled')) {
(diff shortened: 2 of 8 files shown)


wiedld:
We have a weird failure in cypress only (not when using in the UI).
I looked at the cypress stdout, and it returns a 200 success from the DELETE endpoint. So we need to figure out why.

 
Screen Shot 2023-01-31 at 5 29 32 PM

cypress/support/commands.ts
```diff
@@ -920,7 +920,8 @@ export const scriptsLoginWithFlags = (flags): Cypress.Chainable<any> => {
         })
         .then(() =>
           cy.reload().then(() => {
-            cy.getByTestID('page-title').contains('Data Explorer')
+            cy.getByTestID('tree-nav')
+            cy.getByTestID('data-explorer-page').should('exist')
```
wiedld, line 924: The page-title can change, when using a saved script. Therefore, did the proper test-id check for this fixture.

cypress/support/commands.ts
```diff
@@ -886,7 +904,7 @@ export const selectScriptFieldOrTag = (name: string, beActive: boolean) => {
 }
 
 export const switchToDataExplorer = (typ: 'new' | 'old') => {
-  cy.getByTestID('page-title').contains('Data Explorer')
+  cy.getByTestID('data-explorer-page').should('exist')
```
appletreeisyellow, line 907: Why do you use `exist` instead of `be.visible`?

cypress/e2e/oss/scriptsCrud.test.ts
```diff
@@ -0,0 +1,17 @@
...
+    cy.scriptsLoginWithFlags({}).then(() => {
+      cy.switchToDataExplorer('new')
+    })
+  })
+
+  it('should not have the scripts crud options', () => {
+    cy.getByTestID('script-query-builder--open-script').should('not.exist')
+    cy.getByTestID('script-query-builder--save-script').should('not.exist')
+    cy.getByTestID('script-query-builder--edit-script').should('not.exist')
+  })
```
appletreeisyellow, line 16: Nice adding!

eatondustin1 (approved):
LGTM

appletreeisyellow (approved):
Looks good to me! Just a quick question below

cypress/support/commands.ts
wiedld: Good question. We could done either way. It's a bigger deal to use the `is.visible` for clickable elements like buttons etc --> to make sure the user can interact with the element.

Here, we are just looking for the page loading having finished (future proofing, for whatever cypress test calls this fixture). So we could use either visible, or exist.

PR #6543, opened 2023-01-30
fix: Settings page sends user to settings/labels in new iox orgs

wdoconnell:
In local testing, if I simulate the conditions of being a new IOx org, clicking the settings link gives me a 404. The reason is that the link is still hardcoded to send the user to variables by default, which won't exist for new IOx orgs.

Solution is to send the user to labels by default if it's a new IOx org.

### Checklist

### DIFF
--- a/src/shared/containers/SetOrg.tsx
+++ b/src/shared/containers/SetOrg.tsx
@@ -313,7 +313,7 @@ const SetOrg: FC = () => {
           <Route
             exact
             path={`${orgPath}/${SETTINGS}`}
-            component={VariablesIndex}
+            component={isNewIOxOrg ? LabelsIndex : VariablesIndex}
           />
           {/* Users - route has multiple paths to ensure backwards compatibility while https://github.com/influxdata/ui/issues/5396 is being worked on*/}
           {CLOUD && (
--- a/src/shared/containers/SetOrg.tsx
+++ b/src/shared/containers/SetOrg.tsx
@@ -313,7 +313,7 @@ const SetOrg: FC = () => {
           <Route
             exact
             path={`${orgPath}/${SETTINGS}`}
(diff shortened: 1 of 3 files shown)


src/shared/containers/SetOrg.tsx
```diff
@@ -313,7 +313,7 @@ const SetOrg: FC = () => {
           <Route
             exact
             path={`${orgPath}/${SETTINGS}`}
-            component={VariablesIndex}
+            component={isNewIOxOrg ? LabelsIndex : VariablesIndex}
```
hoorayimhelping, line 316: Instead of conditionally rendering variables or labels, (which will persist after the conditions that require the conditional are gone) why don't we just make Labels the default page under settings? That way it'll just work, whether the org has variables or not

wdoconnell: We could do that - it'll just mean it changes the UI experience for users who aren't new IOx.

@garylfowler @gunnaraasen Does it matter whether the "Labels" page or the "Variables" page renders by default when users visit Settings? If we switch from Variables to "Labels" we won't need a different routing behavior for the "Settings" page depending on whether the user is in a New IOx org or not.

garylfowler: I don't think it matters.

wdoconnell: Unless they disagree, I'm inclined to leave it as-is for now, just because Labels isn't the first tab. Reordering the tabs out of alpha order seems like a little yak-shave-y for just fixing the default page.

hoorayimhelping (approved):

hoorayimhelping (approved):

blegesse-git (approved):
LGTM

PR #6545, opened 2023-01-30
chore: conditionally show old Data Explorer 

### LINKED ISSUE #6522: Hide Legacy Data Explorer Feature from any New Accounts
## User Story

<p></p>

As an organization with a new SQL optimized storage engine, I want to prohibit any sign\-ups after Jan 31 from seeing the old Data Explorer which uses Flux, so that they don't start using Flux when the DB is optimized for SQL\.  I want to feature flag control this in case I need to make any exceptions\.

<p></p>

## Acceptance Criteria

<p></p>

- A feature flag will be created to control whether or not the "new script editor" toggle is shown in Data Explorer in the UI

- If the sign\-up date for the user is after Jan 31st and the feature flag is false, then the when clicking on the Data Explorer, only the new script editor will be shown, and there will be no toggle option; also \- when clicking on "New Script" the user will only be able to select SQL \(the flux option should no longer be available\)

- If the feature flag is true, or the sign\-up date is before Jan 31st, then the toggle option to flip back and forth between old and new will still be shown

- These changes will only apply to IOx Enabled Orgs\.  For non\-IOx orgs, this hiding should not occur\.

<p></p>

blegesse-git:
Closes #6522 

For CLOUD IOx enabled users who's org is created after 1/31/23, the following items are hidden 

-  the `switch to old Data Explorer` toggle in the top right corner of the UI
- the old Data Explorer page 
- The flux functions inside the New Query Builder 
- the `flux` option in the New Script drop down button inside New Query Builder

When `showOldDataExplorerInNewIOx` feature flag is flipped on, new IOx users will be able to access both old Data Explorer and New Query Builder just like TSM users

Logic flow chart:
Screen Shot 2023-01-30 at 3 09 26 PM

UI Demo: new IOx org when feature flag is on and off
https://user-images.githubusercontent.com/66275100/215594720-bf33947f-009e-46da-bdf4-9a0d9c7ae016.mov
UI Demo: non-IOx org when feature flag is on and off
https://user-images.githubusercontent.com/66275100/215604792-01f5ec40-d0c4-4a54-ace7-01445de8968c.mov

### DIFF
--- a/src/dataExplorer/components/DataExplorerPage.tsx
+++ b/src/dataExplorer/components/DataExplorerPage.tsx
@@ -52,7 +52,7 @@ const DataExplorerPageHeader: FC = () => {
     useContext(AppSettingContext)
   const {resource, save} = useContext(PersistanceContext)
   const isNewIOxOrg = useSelector(selectIsNewIOxOrg)
-  const showDataExplorerToggle =
+  const shouldShowDataExplorerToggle =
     isFlagEnabled('newDataExplorer') &&
     (!isNewIOxOrg || isFlagEnabled('showOldDataExplorerInNewIOx'))
 
@@ -100,7 +100,7 @@ const DataExplorerPageHeader: FC = () => {
     >
       {pageTitle}
       <FlexBox margin={ComponentSize.Large}>
-        {showDataExplorerToggle && (
+        {shouldShowDataExplorerToggle && (
           <FlexBox margin={ComponentSize.Medium}>
             <InputLabel>Switch to old Data Explorer</InputLabel>
             <SlideToggle
@@ -123,13 +123,13 @@ const DataExplorerPage: FC = () => {
   const isNewIOxOrg =
     useSelector(selectIsNewIOxOrg) &&
     !isFlagEnabled('showOldDataExplorerInNewIOx')
-  const showNotebooks = useSelector(selectShouldShowNotebooks)
-  const showNewExplorer =
+  const shouldShowNotebooks = useSelector(selectShouldShowNotebooks)
+  const shouldShowNewExplorer =
     (scriptQueryBuilder && isFlagEnabled('newDataExplorer')) || isNewIOxOrg
 
-  const showSaveAsButton =
+  const shouldShowSaveAsButton =
     !isNewIOxOrg ||
-    showNotebooks ||
+    shouldShowNotebooks ||
     isFlagEnabled('showTasksInNewIOx') ||
     isFlagEnabled('showDashboardsInNewIOx') ||
     isFlagEnabled('showVariablesInNewIOx')
@@ -201,7 +201,7 @@ const DataExplorerPage: FC = () => {
             </div>
           </FeatureFlag>
         )}
-        {!showNewExplorer && (
+        {!shouldShowNewExplorer && (
           <Page.ControlBar fullWidth={true}>
             <Page.ControlBarLeft>
               <ViewTypeDropdown />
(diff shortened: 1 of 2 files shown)


src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -51,6 +51,11 @@ const DataExplorerPageHeader: FC = () => {
   const {scriptQueryBuilder, setScriptQueryBuilder} =
     useContext(AppSettingContext)
   const {resource, save} = useContext(PersistanceContext)
+  const isNewIOxOrg = useSelector(selectIsNewIOxOrg)
+  const showDataExplorerToggle =
+    isFlagEnabled('newDataExplorer') &&
+    (!isNewIOxOrg || isFlagEnabled('showOldDataExplorerInNewIOx'))
```
hoorayimhelping, line 57: 👍  pretty easy to follow

src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -51,6 +51,11 @@ const DataExplorerPageHeader: FC = () => {
   const {scriptQueryBuilder, setScriptQueryBuilder} =
     useContext(AppSettingContext)
   const {resource, save} = useContext(PersistanceContext)
+  const isNewIOxOrg = useSelector(selectIsNewIOxOrg)
+  const showDataExplorerToggle =
```
hoorayimhelping, line 55: Out of context, it might be easy to think `showDataExplorerToggle` is a method that shows the Data Explorer toggle - it reads like a verb, which is why I typically recommend a conjugated verb like `should` a la `shouldShowDataExplorerToggle`.

src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -114,10 +119,13 @@ const DataExplorerPage: FC = () => {
...
     useContext(AppSettingContext)
   useLoadTimeReporting('DataExplorerPage load start')
-  const showNewExplorer = scriptQueryBuilder && isFlagEnabled('newDataExplorer')
   const history = useHistory()
-  const isNewIOxOrg = useSelector(selectIsNewIOxOrg)
+  const isNewIOxOrg =
+    useSelector(selectIsNewIOxOrg) &&
+    !isFlagEnabled('showOldDataExplorerInNewIOx')
   const showNotebooks = useSelector(selectShouldShowNotebooks)
+  const showNewExplorer =
```
hoorayimhelping, line 127: This is another example where it would be easy to confuse this boolean expression with a method name

src/dataExplorer/components/ScriptQueryBuilder.tsx
```diff
@@ -105,6 +110,9 @@ const ScriptQueryBuilder: FC = () => {
     }
   }, [handleClear, hasChanged])
 
+  const filterOutFluxInIOx = option =>
```
hoorayimhelping, line 113: To hammer on this point some more, this _is_ an action/verb where `filterOutFluxInIOx` is a method, rather than a boolean value. Without any context into what's happening, the variable name can give a clue to what the intent of this code is

hoorayimhelping (commented):
Looks good, let me know what you think of my comments about boolean variable names

src/dataExplorer/components/DataExplorerPage.tsx
blegesse-git: good point, I'll update the rest of variables names in this files to follow the same naming convention

wdoconnell (approved):

wdoconnell:
Thank you for handling the crazy boolean logic required for this. 🥳

PR #6557, opened 2023-01-31
fix: reattached existing composition on page reload or nav event

wiedld:
## New UI asap request for SQL composiiton
They want the SQL composition to reattached on page reload and navigation. This was a limitation already noted when the original work was done >1 month ago -- but we suddenly have an asap issue that it's no longer acceptable.

https://user-images.githubusercontent.com/10232835/215886827-902800c2-1ca2-4ff1-91d5-10a0eabcff1e.mov

## The risk

This is NOT language aware, so we have all kinds of fragile edges cases which can occur. Such as this error which I already encountered:
Screen Shot 2023-01-31 at 1 34 35 PM

The page was reload while updates (to remove the `alert` filter) were still being applied. So it was not aware that it was only re-attaching a partial composition. This is because we don't have language-awareness which can parse the SQL and determine that the remaining editor text was actually part of a composition (and not something a user just typed into the text editor).

So I know we have lots of complaints today to make the composition re-attach. But there be dragons, when we have no language-awareness.

## Checklist

### DIFF
--- a/src/languageSupport/languages/sql/connection.ts
+++ b/src/languageSupport/languages/sql/connection.ts
@@ -36,6 +36,9 @@ const groupedTagValues = (
     {}
   )
 
+// tech debt: because product requirement to re-attach composition on page reload/nav
+const TRACKED_COMPOSITION_START_LINE = 1
+
 export class ConnectionManager extends AgnosticConnectionManager {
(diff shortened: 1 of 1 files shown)


wiedld:
Agreed we are not doing this. Closing.

PR #6559, opened 2023-01-31
feat(ui): update main doc link

mavarius:
Changes the documentation link in the help bar and removes the docs search bar.

The current search bar is only searching the old docs site and not the IOx docs.

### Checklist

### DIFF
--- a/src/pageLayout/containers/MainNavigation.tsx
+++ b/src/pageLayout/containers/MainNavigation.tsx
@@ -275,7 +275,7 @@ export const MainNavigation: FC = () => {
   }
 
   const docslink = isNewIOxOrg
-    ? 'http://docs.influxdata.com/influxdb/cloud-iox/'
+    ? 'https://docs.influxdata.com/influxdb/cloud-iox/'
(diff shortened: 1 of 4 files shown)


src/pageLayout/containers/MainNavigation.tsx
```diff
@@ -274,6 +275,10 @@ export const MainNavigation: FC = () => {
     return null
   }
```
wdoconnell, line 277: For context for others - the reason we need both of these selectors in `MainNavigation` is:

(1) `isNewIOxOrg` is used for feature deprecation in new IOx orgs (orgs created on or after 1/31). Old IOx users can still see certain features, but going forward _new_ IOx users cannot see features like tasks, alerts, old data explorer, etc.
(2) `isIOxOrg` is used to identify _any_ IOx org, and is therefore used to identify the corresponding docs link. That ensures that 'old' iox users (e.g. lighthouse program) will still be pointed to the right docs.

wdoconnell (approved):

Issue #6562, opened 2023-02-01
Get e2es working on iox


wiedld:
Follow up to the excellent repair job (to get our iox e2es running again). Here are all the [skipped tests](https://github.com/influxdata/ui/pull/6561).

Please update below as you grab a test to enable.
Please screenshot a passing `yarn test:e2e:remocal:iox` for the test fixed!

## Fixes done/in-progress:
- [x] cypress/e2e/shared/editor.test.ts  :hamster: DLW  
- [x] cypress/e2e/cloud/helpBar.test.ts  📋 BL
- [x] cypress/e2e/cloud/communityTemplates.test.ts    BL
- [x] cypress/e2e/cloud/dashboardsView.test.ts :hamster: DLW
- [x] cypress/e2e/shared/adaptiveZoom.test.ts :hamster: DLW
- [x] cypress/e2e/shared/annotations.band.test.ts :hamster: DLW
- [x] cypress/e2e/shared/annotations.range.test.ts :hamster: DLW
- [x] cypress/e2e/shared/annotations.singlestat.test.ts :hamster: DLW
- [x] cypress/e2e/shared/annotations.xy.test.ts :hamster: DLW
- [x] cypress/e2e/shared/checks.test.ts  📋 BL
- [x] cypress/e2e/shared/dashboardsIndex.test.ts :hamster: DLW
- [x] cypress/e2e/shared/dashboardsRefresh.test.ts :hamster: DLW
- [x] cypress/e2e/shared/dashboardsVariables.test.ts :hamster: DLW
- [x] cypress/e2e/shared/dashboardsView.test.ts 📋 CY + :hamster: DLW
- [x] cypress/e2e/shared/legends.test.ts - BOCO
- [x] cypress/e2e/shared/loadDataSources.test.ts 📋 BOCO
- [x] cypress/e2e/shared/nav.test.ts - :hamster: DLW
- [x] cypress/e2e/shared/notificationEndpoints.test.ts  📋 BL
- [x] cypress/e2e/shared/notificationRules.test.ts 📋  BOCO
- [x] cypress/e2e/shared/queryBuilder.test.ts 📋 CY
- [x] cypress/e2e/shared/scriptQueryBuilder.flux.test.ts 📋 CY
- [x] cypress/e2e/shared/scriptQueryBuilder.results.test.ts 📋 CY
- [x] cypress/e2e/shared/scriptQueryBuilder.sql.test.ts 📋 CY
[...]

PR #6567, opened 2023-02-01
fix(6562): dashboard tests passing on iox

wiedld:
part of #6562 

## Reasoning:
* e2es goals:
    * prevent regressions of any existing code (by internal or external users)
    * to ensure e2e compatibility with the full stack.
       * any exceptions added are about the backend/storage
           * e.g. only tsm supports explicit bucket schema --> use cy.skipOn() as needed
           * e.g. only iox supports SQL --> use cy.skipOn() as needed
* enablement flags are selectively on:
    * which flags are required, explicitly indicates the feature coupling.
    * we could also use this to set ordering of feature deletion (when we eventually get there).
    * e.g. `showDashboardsInNewIOx` + `showVariablesInNewIOx` => dashboards have to be removed, before we can remove variables.

## Evidence of tests passing on iox:

Screenshots from after the rebase onto the proper cypress env vars.

Screen Shot 2023-02-01 at 4 47 32 PM
Screen Shot 2023-02-01 at 4 42 54 PM

Screen Shot 2023-02-01 at 5 15 50 PM
Screen Shot 2023-02-01 at 5 11 18 PM
Screen Shot 2023-02-01 at 4 59 29 PM

## Checklist

### DIFF
--- a/cypress/e2e/shared/dashboardsView.test.ts
+++ b/cypress/e2e/shared/dashboardsView.test.ts
@@ -6,19 +6,22 @@ const isTSMOrg = !isIOxOrg
 
 describe('Dashboard - TSM and pre marty release IOx', () => {
   beforeEach(() => {
-    cy.flush()
-    cy.signin()
-    cy.setFeatureFlags({
-      showDashboardsInNewIOx: true,
-      showVariablesInNewIOx: true,
-    }).then(() => {
-      cy.fixture('routes').then(({orgs}) => {
-        cy.get<Organization>('@org').then(({id: orgID}: Organization) => {
-          cy.visit(`${orgs}/${orgID}/dashboards-list`)
-          cy.getByTestID('tree-nav')
-        })
-      })
-    })
+    cy.flush().then(() =>
+      cy.signin().then(() =>
+        cy
+          .setFeatureFlags({
+            showDashboardsInNewIOx: true,
+          })
+          .then(() =>
+            cy.fixture('routes').then(({orgs}) => {
+              cy.get<Organization>('@org').then(({id: orgID}: Organization) => {
+                cy.visit(`${orgs}/${orgID}/dashboards-list`)
+                cy.getByTestID('tree-nav')
+              })
+            })
+          )
+      )
+    )
   })
 
   it("can edit a dashboard's name", () => {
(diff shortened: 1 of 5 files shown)


appletreeisyellow (dismissed):

wdoconnell (commented):
Looks good - but do we need the variables flag on as well for all features? We're only flipping `showDashboardsInIOx` for a couple of these tests, it looks like.

I would also add a test along the lines of "Dashboards doesn't display if it's an IOx org."

appletreeisyellow:
@wdoconnell I was working on `shared/dashboardsView.test.ts` and I put up a PR here https://github.com/influxdata/ui/pull/6573, which includes an additional test for "Dashboards doesn't display if it's an IOx org" case. So it's covered!

wdoconnell:
Thanks @appletreeisyellow

wiedld:
Had to rebase a merge conflict on 1 of the tests. (Thanks for the updates @appletreeisyellow !)

Re-ran the iox test locally, just to make sure ok
Screen Shot 2023-02-02 at 3 05 43 PM
Screen Shot 2023-02-02 at 3 13 20 PM

cypress/e2e/shared/dashboardsView.test.ts
```diff
@@ -6,19 +6,22 @@ const isTSMOrg = !isIOxOrg
...
-          cy.visit(`${orgs}/${orgID}/dashboards-list`)
-          cy.getByTestID('tree-nav')
-        })
-      })
-    })
+    cy.flush().then(() =>
+      cy.signin().then(() =>
+        cy
+          .setFeatureFlags({
+            showDashboardsInNewIOx: true,
```
wiedld, line 13: This specific test does not need variables in order to pass.

wiedld:
> Looks good - but do we need the variables flag on as well for all features? We're only flipping `showDashboardsInIOx` for a couple of these tests, it looks like.
> 
> I would also add a test along the lines of "Dashboards doesn't display if it's an IOx org." And that test will be skipped if the org type is not IOx.

The tests here pass iox without variables being on. So I was thinking along the lines of keeping the minimum flags needed. But I can change that too.

appletreeisyellow (approved):

PR #6569, opened 2023-02-02
feat(6562): marty release test

wiedld:
part of #6562 

Since the rest of the e2es are testing UI compatibility with the backend, we also need a specific test for deprecations of each cloud release.

## Evidence of tests passing on iox:
Screen Shot 2023-02-02 at 4 48 07 PM

## Checklist

### DIFF
--- a/cypress/e2e/cloud/releases.test.ts
+++ b/cypress/e2e/cloud/releases.test.ts
@@ -1,5 +1,7 @@
-import {IOX_SWITCHOVER_CREATION_DATE} from 'src/shared/constants'
 import {Organization} from '../../../src/types'
+
+// webpack bundling error if try importing --> due to imports in the constants file
+const IOX_SWITCHOVER_CREATION_DATE = '2023-01-31T00:00:00Z'
 
 describe('Deprecations per cloud release', () => {
   before(() => {
--- a/cypress/e2e/cloud/releases.test.ts
+++ b/cypress/e2e/cloud/releases.test.ts
@@ -1,0 +1,146 @@
+import {IOX_SWITCHOVER_CREATION_DATE} from 'src/shared/constants'
+import {Organization} from '../../../src/types'
+
+describe('Deprecations per cloud release', () => {
+  before(() => {
+    cy.flush().then(() =>
+      cy.signin().then(() =>
+        cy.isIoxOrg().then(isIox => {
+          cy.skipOn(!isIox)
+          cy.visit('/')
+          cy.getByTestID('home-page--header').should('be.visible')
+        })
+      )
+    )
+  })
+
+  describe('Marty-release', () => {
+    describe('has specific navigation options', () => {
+      beforeEach(() => {
+        cy.signinWithoutUserReprovision()
+      })
+
+      it('depreciated features do not exist', () => {
+        cy.getByTestID('nav-item-flows').should('not.exist')
+        cy.getByTestID('nav-item-alerting').should('not.exist')
+        cy.getByTestID('nav-item-dashboards').should('not.exist')
+        cy.getByTestID('nav-item-tasks').should('not.exist')
+      })
+
+      describe('supported features are still navigable', () => {
+        ;[
+          {navId: 'nav-item-data-explorer', title: 'Data Explorer'},
+          {navId: 'nav-item-load-data', title: 'Load Data'},
+          {navId: 'nav-item-settings', title: 'Settings'},
+        ].forEach(({navId, title}) => {
+          it(`for ${title} page`, () => {
+            cy.getByTestID(navId).should('be.visible').click({force: true})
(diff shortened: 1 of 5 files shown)


src/usage/context/usage.tsx
```diff
@@ -226,15 +226,17 @@ export const UsageProvider: FC<Props> = React.memo(({children}) => {
...
+          const amount: number = result
+            .reduce((a: number, b) => a + parseFloat(b), 0)
+            .toFixed(2)
+          setCreditUsage({
+            amount,
+            status: RemoteDataState.Done,
+          })
         })
-      })
+        .catch(err => console.error(err))
```
wiedld, line 239: Handling uncaught promise exception.

Screen Shot 2023-02-01 at 9 52 45 PM

cypress/e2e/cloud/releases.test.ts
```diff
@@ -0,0 +1,146 @@
...
+          }`, () => {
+            cy.getByTestID(navId).should('be.visible').click({force: true})
+            cy.getByTestID(navSubId).should(supported ? 'exist' : 'not.exist')
+            cy.getByTestID(tabId).should(supported ? 'be.visible' : 'not.exist')
+          })
+        })
+      })
+    })
+
+    // TODO: add any more marty-specific features
```
wiedld, line 105: Do we have any marty-specific features, which do not exist for the other/older iox orgs?

cypress/e2e/cloud/releases.test.ts
```diff
@@ -0,0 +1,148 @@
+import {Organization} from '../../../src/types'
+
+// webpack bundling error if try importing --> due to imports in the constants file
+const IOX_SWITCHOVER_CREATION_DATE = '2023-01-31T00:00:00Z'
```
wiedld, line 4: FYI -- why not importing this constant.
Screen Shot 2023-02-01 at 10 37 25 PM

abshierjoel (dismissed):

appletreeisyellow:
Should we also test the behavior of the old and new data explorer here? I know it is implicitly covered in scriptQueryBuilder tests. Adding it here makes the logic for showing the old or new data explorer clearer.

Here are different cases:

* If in a TSM org or in an IOx org created **before** the Marty release, users have the toggle to switch between the old and the new data explorer
* If in an IOx org created **after** the Marty release, users landed on the new data explorer page and the toggle should not exist

What do you think?

wiedld:
> Should we also test the behavior of the old and new data explorer here? I know it is implicitly covered in scriptQueryBuilder tests. Adding it here makes the logic for showing the old or new data explorer clearer.
> 
> Here are different cases:
> 
>     * If in a TSM org or in an IOx org created **before** the Marty release, users have the toggle to switch between the old and the new data explorer
> 
>     * If in an IOx org created **after** the Marty release, users landed on the new data explorer page and the toggle should not exist
> 
> 
> What do you think?

Exactly right! 

Additionally, I moved the dashboard marty-release navigation/404 test into here. Because that way we can glance at a single e2e test file, in order to see the contract per each cloud release. 

This single `release.test` means that we do not have to manually check each test for the product-level deprecation decisions. All of those decisions (which are subject to change, and staged depreciations) are located in this single file here.

Whereas the backend integration issues (e.g. what tsm versus iox can handle) still utilize cy.skipOn() in each individual tests. Does these different use cases make sense?

wdoconnell:
> > Should we also test the behavior of the old and new data explorer here? I know it is implicitly covered in scriptQueryBuilder tests. Adding it here makes the logic for showing the old or new data explorer clearer.
> 
> > What do you think?
> 
> Exactly right!
> 
> Additionally, I moved the dashboard marty-release `shared/dashboardsView` test into here. Because that way we can glance at a single e2e test file, in order to see the contract per each cloud release.
> 
> This single `release.test` means that we do not have to manually check each test for the product-level deprecation decisions. All of those decisions (which are subject to change, and staged depreciations) are located in this single file here.
> 
> Whereas the backend integration issues (e.g. what tsm versus iox can handle) still utilize cy.skipOn() in each individual tests. Does this distinction make sense?

My thinking here is explained a bit more in my slack message, but my preference is not to do blanket "release" tests. Those will proliferate every time we deprecate a feature, and I worry about splitting tests between a release-organized file and a feature-organized file that we'll need to hop between.

appletreeisyellow:
I can see the benefit of this PR and it's very smart! I mainly explained my thoughts and concerns on [Slack](https://influxdata.slack.com/archives/CLHATC50E/p1675441240996239?thread_ts=1675377540.256969&cid=CLHATC50E). Same as Bill, my preference is to keep the iox test in each feature test file.

wiedld:
Discussed. We are good to merge.

appletreeisyellow (approved):

PR #6579, opened 2023-02-03
fix: update load data sources test for iox

wdoconnell:
Part of #6562 

Updates the load data sources test to be compatible with both the TSM and IOx pipeline.

There was a 'write data' test in here that had been skipped even before the IOx release. Unskipped and updated it so that it works with the combination of the load data page and new data explorer.

### Checklist

### DIFF
--- a/cypress/e2e/shared/loadDataSources.test.ts
+++ b/cypress/e2e/shared/loadDataSources.test.ts
@@ -1,7 +1,6 @@
 import {Organization} from '../../src/types'
 
 const isIOxOrg = Boolean(Cypress.env('useIox'))
-const isTSMOrg = !isIOxOrg
 
 // Library to use when testing client library details view.
 // As of 2/2023, IOx only has one client library - python, which has a wizard.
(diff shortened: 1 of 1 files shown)


cypress/e2e/shared/loadDataSources.test.ts
```diff
@@ -152,32 +173,49 @@ describe.skip('Load Data Sources', () => {
...
+    cy.skipOn(isTSMOrg)
+    cy.flush()
+    cy.signin()
+    cy.setFeatureFlags({
+      newDataExplorer: true,
+    })
+    cy.get('@org').then(({id}: Organization) =>
+      cy.fixture('routes').then(({orgs}) => {
+        cy.visit(`${orgs}/${id}/load-data/sources`)
+        cy.getByTestID('tree-nav')
```
blegesse-git, line 217: what is this testing exactly? Do we even need an IOx test for this particular test? The only thing different in load sources page for iox orgs is the number of client libraries, correct? If testing is necessary, maybe we can add a line that checks if the correct client library shows up?

blegesse-git (commented):
just one little comment, everything else looks good

cypress/e2e/shared/loadDataSources.test.ts
wdoconnell: This was old cruft from another test - you're right that this just needs to be removed. Good catch!

blegesse-git (approved):

appletreeisyellow (approved):

PR #6591, opened 2023-02-06
fix: update legends test for iox

wdoconnell:
Part of #6562 

This PR:
1. Updates the legends tests to work with IOx.
2. Unskips tests that weren't being run due to Cypress firefox compatibility issues (we only test Chrome now).
3. Fixes a bug where a new iox user with access to old data explorer but none of the other features could still see the 'Save As' button, even if there was nowhere to save the query.

### Checklist

### DIFF
--- a/cypress/e2e/shared/legends.test.ts
+++ b/cypress/e2e/shared/legends.test.ts
@@ -13,11 +13,26 @@ const VIS_TYPES = [
   'table',
 ]
 
-describe.skip('Legends', () => {
+const CLOUD = Cypress.env('dexUrl') === 'OSS' ? false : true
+
+const dataExplorerFeatureFlags = {
+  showDashboardsInNewIOx: true,
+  showOldDataExplorerInNewIOx: true,
+  showTasksInNewIOx: true,
+}
+
+// Absence of this feature (legends - old data explorer) in IOx is tested
+// in the Data Explorer tests.
+describe('Legends', () => {
   describe('in the Data Explorer', () => {
     beforeEach(() => {
       cy.flush()
       cy.signin()
+
+      if (CLOUD) {
+        cy.setFeatureFlags(dataExplorerFeatureFlags)
+      }
+
       cy.get('@org').then(({id}: Organization) => {
         cy.createMapVariable(id)
         cy.fixture('routes').then(({orgs, explorer}) => {
@@ -348,8 +363,7 @@ describe.skip('Legends', () => {
         )
       })
 
-      // Skip for now because Firefox does not run the test correctly with a newly created cell with query and view options included
-      it.skip('saves to a dashboard as a cell with the static legend options open and with the query pre-submitted', () => {
+      it('saves to a dashboard as a cell with the static legend options open and with the query pre-submitted', () => {
         const cellName = 'anti-crash test pre-submitted data explorer'
         cy.writeData(points(100))
 
@@ -411,10 +425,19 @@ describe.skip('Legends', () => {
     })
   })
 
+  const dashboardsFeatureFlags = {
+    showDashboardsInNewIOx: true,
+  }
+
   describe('in Dashboards', () => {
     beforeEach(() => {
       cy.flush()
       cy.signin()
+
+      if (CLOUD) {
+        cy.setFeatureFlags(dashboardsFeatureFlags)
+      }
+
       cy.fixture('routes').then(({orgs}) => {
         cy.get('@org').then(({id}: Organization) => {
           cy.visit(`${orgs}/${id}/dashboards-list`)
@@ -475,7 +498,7 @@ describe.skip('Legends', () => {
             .click()
             .type(cellName + '{enter}')
 
-          // Without submitting the query, save it to a dashboard
+          cy.get('.veo-contents').click() // click out of inline editor
           cy.getByTestID('save-cell--button').click()
           cy.get('.cell--name').should('have.text', cellName)
           cy.getByTestID('giraffe-legend-table').should('not.exist')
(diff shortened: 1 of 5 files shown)


cypress/e2e/shared/legends.test.ts
```diff
@@ -540,7 +549,9 @@ describe.skip('Legends', () => {
           cy.getByTestID('page-header').click().type(cellName)
 
           // Save it to a dashboard
+          cy.get('.veo-contents').click() // click out of inline editor
           cy.getByTestID('save-cell--button').click()
```
wdoconnell, line 553: This is needed because the save-cell button isn't properly selected by Cypress unless you click out of the inline editor. This is the same fix we've employed on other tests here.

src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -137,7 +137,7 @@ const DataExplorerPage: FC = () => {
     (scriptQueryBuilder && isFlagEnabled('newDataExplorer')) || isNewIOxOrg
 
   const shouldShowSaveAsButton =
-    !isNewIOxOrg ||
+    !useSelector(selectIsNewIOxOrg) ||
     shouldShowNotebooks ||
```
wdoconnell, line 141: This edit is a minor bug fix that requires some explanation because it's pretty far down the logic tree of what can happen within the matrix of feature overrides.

`shouldShowSaveAsButton` controls whether the "Save As" button on data explorer displays. This variable evaluates to `true` and permits the display of the "Save As" button in data explorer if the user is _not_ in an IOx org; or, if the user _is_ in an IOx org, but still has access to any one of the following: notebooks, tasks, dashboards, or variables. That can happen if the user is in the IOx lighthouse program, or has a feature override flag enabled. 

The logic wasn't right here for a specific edge case that appeared during the fixed e2e tests:
1. You're in an IOx org and
2. `showOldDataExplorerInNewIOx` is on and
3. All other overrides are off

In that case, `!isNewIOxOrg` incorrectly evaluates to `true`, since the override flag for data explorer is _not_ off (`!isFlagEnabled('showOldDataExplorerInNewIOx')`). This shows the "Save As" button even though the resulting overlay wouldn't offer you any place to save the query. The "Save As" button shouldn't care whether you have the `showOldDataExplorerInNewIOx` flag on or off. The "Save As" button should only display if you're in a TSM/IOx lighthouse org; or one of the tasks/notebooks/dashboards/variables overrides is on.

src/flows/selectors/flowsSelectors.ts
```diff
@@ -11,6 +11,10 @@ export const selectNotebooks = (state: AppState): Notebook[] => {
 }
 
 export const selectShouldShowNotebooks = (state: AppState): boolean => {
+  if (isFlagEnabled('showNotebooksForCI')) {
+    return true
+  }
+
```
wdoconnell, line 17: TLDR version: If `showNotebooksForCI` is on, this selector should always return true.

--
Less TLDR version: 

This selector was producing issues in this e2e test because, when trying to force the display of notebooks for CI testing in Cloud, we would get this result:
1. `if (!CLOUD)` is `false` - skipped.
2. `if (!wasCreatedBeforeIOxCutoff & !isFlagEnabled('showNotebooksForCI'))` is false (the flag is on) - skipped.
3. Selector returns whether the number of notebooks in state is `> 0`.

3. is always false if the test hasn't set up a notebook yet. So we need to move the logic that forces the display of notebooks based on the `showNotebooksForCI` flag up, so that Cypress can navigate to the notebooks page to create whatever notebook it needs for the test.

appletreeisyellow (approved):
Looks good to me! The comments help a lot with why the changes were made!

src/dataExplorer/components/DataExplorerPage.tsx
wiedld: We are getting some weird behaviors, because we have conditionals (including flags and iox org checking) at multiple levels in our code. In the scripts tests, we ran into a conflict between the `isFlagEnabled('showOldDataExplorerInNewIOx')` and the AppSettings global state of `scriptQueryBuilder`.

We are merging in a new fixture with [this PR ](https://github.com/influxdata/ui/pull/6593#discussion_r1099621263)(hopefully merged today), which will have a new `cy.mockIsCloud2Org()` which returns an older createdAt date for the org.

This means that any deprecated features based on org createdAt date, will be turned back on if we use this in the beforeEach(). This includes features deprecated on iox (e.g. marty release), and also in the future could be used on tsm if we deprecate features prior to the migration to iox backend for those folks.

wiedld: People now have the choice of using the flags (e.g. `showDashboardsInNewIOx`) as needed for some tests. But for features that have nested conditionals throughout shared logic (a.k.a. old and new data explorer), using this `cy.mockIsCloud2Org()`, followed by `cy.switchToDataExplorer('old')`may be easier.

src/flows/selectors/flowsSelectors.ts
wiedld: Or we could mock a notebook creation. See the `release.test`. 
https://github.com/influxdata/ui/blob/master/cypress/e2e/cloud/releases.test.ts#L165-L171

wdoconnell: 👍

wdoconnell:
Removed the proposed changes to the notebooks selector after discussion with @wiedld. We actually don't need them for this specific PR, as it turns out.

appletreeisyellow (approved):

wdoconnell:
Need a quick re-review to test a fix - in merge queue, not working with OSS, likely due to the cloud-only FFs being switched on.

appletreeisyellow (dismissed):

blegesse-git (approved):

wdoconnell:
Rebasing this because tests have been updated (tasks has been split into two tests) since the creation of this PR.

wdoconnell:
This one required further edits because it was failing in merge queue, and the test was so long that was leaving no artifact.

I will merge this when I'm back online on Thursday.

appletreeisyellow (dismissed):

appletreeisyellow (approved):

PR #6593, opened 2023-02-07
fix(6562): extract common cypress methods, and fix script builder iox tests

wiedld:
Part of #6562 

## Newly unskipped tests, on iox:

Screen Shot 2023-02-07 at 6 47 15 PM

Screen Shot 2023-02-07 at 6 41 24 PM

Screen Shot 2023-02-07 at 7 41 20 PM

## Modified tests, still passing iox:

Screen Shot 2023-02-07 at 6 36 54 PM

Screen Shot 2023-02-07 at 7 13 58 PM

Screen Shot 2023-02-07 at 7 05 42 PM

## Checklist

### DIFF
--- a/cypress/e2e/cloud/explorer.test.ts
+++ b/cypress/e2e/cloud/explorer.test.ts
@@ -15,21 +15,28 @@ function getTimeMachineText() {
 }
 
 describe('DataExplorer', () => {
+  before(() => {
+    cy.flush().then(() => cy.signin())
+  })
+
   beforeEach(() => {
-    cy.flush().then(() =>
-      cy.signin().then(() => {
-        cy.setFeatureFlags({
-          showOldDataExplorerInNewIOx: true,
-        }).then(() =>
-          cy.get('@org').then(({id}: Organization) => {
-            cy.createMapVariable(id)
-            cy.fixture('routes').then(({orgs, explorer}) => {
-              cy.visit(`${orgs}/${id}${explorer}`)
-              cy.getByTestID('tree-nav').should('be.visible')
-            })
+    cy.mockIsCloud2Org().then(() =>
+      cy.signinWithoutUserReprovision().then(() =>
+        cy
+          .setFeatureFlags({
+            newDataExplorer: true,
           })
-        )
-      })
+          .then(() =>
+            cy.get('@org').then(({id}: Organization) => {
+              return cy.fixture('routes').then(({orgs, explorer}) => {
+                cy.visit(`${orgs}/${id}${explorer}`)
+                cy.getByTestID('tree-nav').should('be.visible')
+                cy.switchToDataExplorer('old')
+                cy.getByTestID('switch-to-script-editor').should('exist')
+              })
+            })
+          )
+      )
     )
   })
 
--- a/cypress/e2e/cloud/explorer.test.ts
+++ b/cypress/e2e/cloud/explorer.test.ts
@@ -26,6 +26,7 @@ describe('DataExplorer', () => {
             cy.fixture('routes').then(({orgs, explorer}) => {
               cy.visit(`${orgs}/${id}${explorer}`)
               cy.getByTestID('tree-nav').should('be.visible')
+              cy.switchToDataExplorer('old')
             })
           })
         )
--- a/cypress/e2e/cloud/explorer.test.ts
+++ b/cypress/e2e/cloud/explorer.test.ts
@@ -26,7 +26,6 @@ describe('DataExplorer', () => {
             cy.fixture('routes').then(({orgs, explorer}) => {
               cy.visit(`${orgs}/${id}${explorer}`)
               cy.getByTestID('tree-nav').should('be.visible')
-              cy.switchToDataExplorer('old')
             })
           })
         )
--- a/cypress/e2e/shared/explorer.queries.graphs.test.ts
+++ b/cypress/e2e/shared/explorer.queries.graphs.test.ts
@@ -23,6 +23,7 @@ describe('writing queries and making graphs using Data Explorer', () => {
                 route = `${orgs}/${id}${explorer}`
                 cy.visit(route)
                 cy.getByTestID('tree-nav').should('be.visible')
+                cy.switchToDataExplorer('old')
               })
             })
           })
--- a/cypress/e2e/shared/explorer.queries.graphs.test.ts
+++ b/cypress/e2e/shared/explorer.queries.graphs.test.ts
@@ -23,7 +23,6 @@ describe('writing queries and making graphs using Data Explorer', () => {
                 route = `${orgs}/${id}${explorer}`
                 cy.visit(route)
                 cy.getByTestID('tree-nav').should('be.visible')
-                cy.switchToDataExplorer('old')
               })
             })
           })
--- a/cypress/e2e/shared/explorerVisualizations.test.ts
+++ b/cypress/e2e/shared/explorerVisualizations.test.ts
@@ -34,7 +34,6 @@ describe('visualizations', () => {
               })
             })
             cy.writeData(points(NUM_POINTS))
-            cy.switchToDataExplorer('old')
             cy.getByTestID('time-machine--bottom')
           })
       )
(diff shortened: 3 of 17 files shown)


cypress/e2e/shared/explorer.queries.graphs.test.ts
```diff
@@ -23,6 +23,7 @@ describe('writing queries and making graphs using Data Explorer', () => {
                 route = `${orgs}/${id}${explorer}`
                 cy.visit(route)
                 cy.getByTestID('tree-nav').should('be.visible')
+                cy.switchToDataExplorer('old')
```
wiedld, line 26: confirmed. passes iox e2e.

cypress/e2e/shared/explorerVisualizations.test.ts
```diff
@@ -34,6 +34,7 @@ describe('visualizations', () => {
               })
             })
             cy.writeData(points(NUM_POINTS))
+            cy.switchToDataExplorer('old')
```
wiedld, line 37: confirmed. works in iox e2e.

cypress/e2e/shared/loadDataSources.test.ts
```diff
@@ -20,6 +20,7 @@ describe('Load Data Sources', () => {
       cy.fixture('routes').then(({orgs}) => {
         cy.visit(`${orgs}/${id}/load-data/sources`)
         cy.getByTestID('tree-nav')
+        cy.switchToDataExplorer('old')
```
wiedld, line 23: confirmed. works in iox e2e.

cypress/e2e/cloud/explorer.test.ts
```diff
@@ -26,6 +26,7 @@ describe('DataExplorer', () => {
             cy.fixture('routes').then(({orgs, explorer}) => {
               cy.visit(`${orgs}/${id}${explorer}`)
               cy.getByTestID('tree-nav').should('be.visible')
+              cy.switchToDataExplorer('old')
```
wiedld, line 29: confirmed. works in iox e2e.

cypress/e2e/shared/explorerVisualizations.test.ts
```diff
@@ -31,6 +31,8 @@ describe('visualizations', () => {
               cy.createMapVariable(id)
               cy.fixture('routes').then(({orgs, explorer}) => {
                 cy.visit(`${orgs}/${id}${explorer}`)
+                cy.getByTestID('tree-nav')
+                cy.switchToDataExplorer('old')
```
wiedld, line 35: confirmed. works in iox e2e.

src/shared/reducers/app.ts
```diff
@@ -33,7 +33,7 @@ const initialState: AppState = {
     showTemplateControlBar: false,
     timeZone: 'Local',
     navBarState: 'collapsed',
-    scriptQueryBuilder: false,
+    scriptQueryBuilder: true,
```
wiedld, line 36: This should be the default now. Was never done properly -- uncovered while testing.

cypress/e2e/shared/simpleTable.test.ts
```diff
@@ -16,6 +16,7 @@ describe('simple table interactions', () => {
           cy.visit(`${orgs}/${orgID}${explorer}`)
         })
         cy.getByTestID('tree-nav')
+        cy.switchToDataExplorer('old')
```
wiedld, line 19: confirmed. works in iox e2e.

cypress/e2e/shared/queryBuilder.test.ts
```diff
@@ -40,6 +40,7 @@ describe('The Query Builder', () => {
       cy.get('@org').then((org: Organization) => {
         cy.visit(`orgs/${org.id}/data-explorer`)
         cy.getByTestID('tree-nav')
+        cy.switchToDataExplorer('old')
```
wiedld, line 43: confirmed. works in iox e2e.

wiedld:
Have tests failing on CI, which match the fluxdocs being missing again.

cypress/e2e/shared/loadDataSources.test.ts
```diff
@@ -175,8 +174,6 @@ describe('Load Data Sources', () => {
     // use new data explorer, since old data explorer is being deprecated
     cy.getByTestID('nav-item-data-explorer').click({force: true})
 
-    cy.getByTestID('script-query-builder-toggle').click()
```
wiedld, line 178: Minor adjustments in this test, since the new data explorer is now the default.

src/shared/reducers/app.ts
appletreeisyellow: Setting this to `true` as default will cause the [flux function injection failure ](https://app.circleci.com/pipelines/github/influxdata/monitor-ci/19356/workflows/70d227d8-1656-4553-8f02-147925199b95/jobs/1373051/tests#failed-test-0) in old data explorer since the old style will be replaced by the new style. 

https://github.com/influxdata/ui/blob/43c27ee951dc258f762a04bd9cb2d0b113d3c97c/src/shared/components/FilterList/InjectionOption.tsx#L99-L108

For the TSM org users and pre-marty release IOx org users, the old data explorer is still the default option unless they turn off the toggle "Switch to the old Data Explorer". 

I think we should set this value back to `false` for now, then turn back on when the old data explorer is completely gone

cypress/e2e/shared/loadDataSources.test.ts
appletreeisyellow: Adding it back since we still want to set the default value as `scriptQueryBuilder: false` in `src/shared/reducers/app.ts`

src/shared/reducers/app.ts
wiedld: Worked with @appletreeisyellow , and we are instead going to do a proper fix:
* leave `scriptQueryBuilder: true` since it's the proper, intended behavior
* Make sure we don't use the feature flag `showOldDataExplorerInNewIOx`, since it forces the view state out of alignment with this `AppSettings.scriptQueryBuilder`
* instead, make use of a new fixture `cy.mockIsCloud2Org()`, followed by `cy.switchToDataExplorer()`

cypress/e2e/shared/loadDataSources.test.ts
wiedld: We undid this changed, based on the convo documented above.

cypress/support/commands.ts
```diff
@@ -234,6 +234,24 @@ export const isIoxOrg = (): Cypress.Chainable<boolean> => {
   })
 }
 
+const IOX_SWITCHOVER_CREATION_DATE = '2023-01-31T00:00:00Z'
+
+export const mockIsCloud2Org = (): Cypress.Chainable<null> => {
+  const beforeCutoff = new Date(
+    new Date(IOX_SWITCHOVER_CREATION_DATE) - 10000
+  ).toISOString()
```
wiedld, line 242: This mocking produces an older tsm or iox account. Bescially, our full supported feature set prior to our next (cloud 3? 2.5?) release iterations.

Probably could do with better namings. Right now, this only impacts deprecations with older versus newer iox accounts. Eventually could also apply to tsm accounts if we do staged deprecations prior to migration to iox.  Either way, I thought "Cloud2" was a good name. Maybe. 🤷🏼‍♀️

appletreeisyellow (commented):
Seems like the local iox tests for these tests are not included:
* `loadDataSources.test.ts`
* `scriptQueryBuilder.results.test.ts`
 * `oss/explorer.test.ts` 

So I tried on my local machine and I got `loadDataSources.test.ts` passing, but `oss/explorer.test.ts` and `scriptQueryBuilder.results.test.ts` failed. Very wired that they all passed on CI 🤔

Screen Shot 2023-02-08 at 10 47 07 AM

Screen Shot 2023-02-08 at 11 41 29 AM

Screen Shot 2023-02-08 at 12 30 01 PM


Screen Shot 2023-02-08 at 11 39 49 AM

Screen Shot 2023-02-08 at 12 34 19 PM

Screen Shot 2023-02-08 at 12 34 32 PM

appletreeisyellow (approved):
The local test failure was due to my local environment. OSS tests are supposed to fail when test locally because oss doesn't have login. The `scriptQueryBuilder.results` tends timeout when run for too long locally, but they pass individually. 

So the above comment is not a problem! Other than that, the fix looks good to me!

PR #6610, opened 2023-02-14
fix(3921): new DatePicker should use proper typing, and not do manual date math.

wiedld:
Closes https://github.com/influxdata/EAR/issues/3921

## Bug:
* existing code was creating a Date type, and under certain circumstances was over-writing this date type with a string type. Then passing to state.
* existing code was doing manual math to adjust dates, and did not properly handle wrap cases (e.g. last day of month). All of this shenanigans was to avoid using a datetime lib to format a date.
* existing code had wrong when it was a base-0 integer for a date part, versus a base 1 (e.g. human reading). Month is base 0, date is base 1 (24th of month is 24th of month).
   * this last one produced the actual bug reported in the EAR. 

## New difference in functionality:
If you double click on a date, it sets the start/end of the same day. Just at beginning and end hours of the day.

### BROKEN:

https://user-images.githubusercontent.com/10232835/218835043-d0d5420c-b67c-4c82-aa2a-5e424b520bec.mov

### NOW FIXED:

https://user-images.githubusercontent.com/10232835/218835079-a45fbb49-5d46-400e-9365-4ac0a792a5ed.mov

https://user-images.githubusercontent.com/10232835/218835099-b8862a10-96a0-43b3-a317-20736a0c6bcd.mov

## Checklist

### DIFF
--- a/src/shared/components/dateRangePicker/NewDatePicker.tsx
+++ b/src/shared/components/dateRangePicker/NewDatePicker.tsx
@@ -1,4 +1,5 @@
 import React, {FC, useState, useContext, useEffect, useCallback} from 'react'
+const moment = require('moment/min/moment.min.js')
 import {
   AlignItems,
   Button,
@@ -30,6 +31,7 @@ import {TimeRange} from 'src/types'
 
 const NBSP = '\u00a0\u00a0'
 const MAX_WIDTH_FOR_CUSTOM_TIMES = 325
+const FORMAT = 'YYYY-MM-DD HH:mm'
 
 interface Props {
   onCollapse: () => void
@@ -118,8 +120,17 @@ const DatePickerMenu: FC<Props> = ({onCollapse, timeRange, timeRangeLabel}) => {
     setInputEndDate(value)
   }
 
-  const handleSelectDate = (dates): void => {
+  const handleSelectDate = (dates: [Date, Date]): void => {
     const [start, end] = dates
+    // end should be EOD
+    end && end.setMinutes(59)
+    end && end.setHours(23)
+
+    // clone mutable objects
+    const inputStart = new Date(start)
+    const inputEnd = new Date(end)
+
+    // set local state
     if (timeZone === 'UTC') {
       if (start) {
         start.setMinutes(start.getMinutes() + start.getTimezoneOffset())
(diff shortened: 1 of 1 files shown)


src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -129,38 +140,16 @@ const DatePickerMenu: FC<Props> = ({onCollapse, timeRange, timeRangeLabel}) => {
...
+      startInput = moment(inputStart).format(FORMAT)
     }
-    if (endInput instanceof Date) {
-      const endDate = new Date(endInput)
-      // this sets the time to the end of the selected end day
-      endDate.setMinutes(endDate.getMinutes() - 1)
-      const dateParts = [
-        endDate.getFullYear(),
-        endDate.getMonth() + 1,
-        endDate.getDate() + 1,
```
wiedld, line 156: wrong. 24th of month is 24th of month.

src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -129,38 +140,16 @@ const DatePickerMenu: FC<Props> = ({onCollapse, timeRange, timeRangeLabel}) => {
       }
     }
     setDateRange([start, end])
-    let startInput = start
-    let endInput = end
```
wiedld, line 133: we should not be setting this to a Date, and latter on overwriting with a string.

src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -129,38 +140,16 @@ const DatePickerMenu: FC<Props> = ({onCollapse, timeRange, timeRangeLabel}) => {
...
-        endDate.getFullYear(),
-        endDate.getMonth() + 1,
-        endDate.getDate() + 1,
-      ]
-        .map(pad)
-        .join('-')
-      const timeParts = [endDate.getHours(), endDate.getMinutes()]
-        .map(pad)
-        .join(':')
-      endInput = `${dateParts} ${timeParts}`
```
wiedld, line 163: All this to avoid using moment lib. Such that our own code will need to be debugged for all the same problems. 🤦🏼‍♀️

src/shared/components/dateRangePicker/NewDatePicker.tsx
```diff
@@ -1,4 +1,5 @@
 import React, {FC, useState, useContext, useEffect, useCallback} from 'react'
+const moment = require('moment/min/moment.min.js')
```
appletreeisyellow, line 2: Is this what you mean with "strip down the bundle size" by using `require` instead of `import`?

appletreeisyellow (approved):
Nice fix and the reasonings! Just a quick question

src/shared/components/dateRangePicker/NewDatePicker.tsx
wiedld: It's the pathing to the minimal js file, to be loaded. Otherwise we get moment bundled with all the different locales.

There are several different approaches in how to reduce a moment bundle size. This is a very easy one.

wiedld:
Confirmed failed on different, flaking tests.
Trying the merge queue again.

PR #6612, opened 2023-02-14
chore: add new limits to Operator UI org overlay

### LINKED ISSUE #6495: Add new limit controls to operator UI
We have added [new controls to the Cloud limits structure](https://github.com/influxdata/idpe/blob/d1e1a27351d06cc8726903873b85031413c057eb/limits.go#L39-L40):

- One new field with two subfields to control an org's unconditional and idle write timeouts for Flux queries
- One new field with one subfield which controls whether an org can use the Stacks feature

We would like to be able to adjust these on the fly via the operator UI.

blegesse-git:
Closes #6495 

Pr updates openapi `sha` to generate updated unity routes and adds two fields to the Limits section of the Org detail overlay in the Operator UI. 

Screen Shot 2023-02-14 at 3 42 52 PM

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "cy": "CYPRESS_dexUrl=https://$INGRESS_HOST:$PORT_HTTPS CYPRESS_baseUrl=http://localhost:9999 cypress open",
     "cy:dev": "source ../monitor-ci/.env && CYPRESS_dexUrl=CLOUD CYPRESS_baseUrl=https://$INGRESS_HOST:$PORT_HTTPS cypress open --config testFiles='{cloud,shared}/**/*.*'",
     "cy:dev-oss": "source ../monitor-ci/.env && CYPRESS_dexUrl=OSS CYPRESS_baseUrl=https://$INGRESS_HOST:$PORT_HTTPS cypress open --config testFiles='{oss,shared}/**/*.*'",
-    "generate": "export SHA=689356a1c8e3dbd4e6ff6783cf81cbb720f8e64a && export REMOTE=https://raw.githubusercontent.com/influxdata/openapi/${SHA}/ && yarn generate-meta",
+    "generate": "export SHA=db14de0d91be00e75eb946d768e88d771b20a2a9 && export REMOTE=https://raw.githubusercontent.com/influxdata/openapi/${SHA}/ && yarn generate-meta",
     "generate-local": "export REMOTE=../openapi/ && yarn generate-meta",
     "generate-local-cloud": "export REMOTE=../openapi/ && yarn generate-meta-cloud",
     "generate-meta": "if [ -z \"${CLOUD_URL}\" ]; then yarn generate-meta-oss; else yarn generate-meta-cloud; fi",
--- a/src/operator/OrgOverlay.tsx
+++ b/src/operator/OrgOverlay.tsx
@@ -277,7 +277,7 @@ const OrgOverlay: FC = () => {
                       />
                     </Grid.Column>
                   </Grid.Row>
-                  {limits.timeout && (
+                  {limits?.timeout && (
                     <Grid.Row>
                       <Grid.Column widthMD={Columns.Four}>
                         <Form.Label label="Unconditional Timeout (seconds)" />
@@ -299,7 +299,7 @@ const OrgOverlay: FC = () => {
                       </Grid.Column>
                     </Grid.Row>
                   )}
-                  {limits.stack && (
+                  {limits?.stack && (
                     <Grid.Row>
                       <Grid.Column widthMD={Columns.Four}>
                         <Form.Label label="Stack" />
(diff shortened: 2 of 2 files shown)


src/operator/OrgOverlay.tsx
```diff
@@ -277,6 +277,41 @@ const OrgOverlay: FC = () => {
                       />
                     </Grid.Column>
                   </Grid.Row>
+                  {limits.timeout && (
```
wdoconnell: Best here would be `Boolean(limits.timeout)`. Probably doesn't matter for this value though, since I don't think the `timeout` object will be anything but a valid object or `null`. Anyway, here's the case that would occur in, since `logical OR` will return the first non falsy value:

`0 && (<Grid.Row>)` - expect to render `<Grid.Row>`, but instill will render `0`, since `0` is not falsy

src/operator/OrgOverlay.tsx
```diff
@@ -277,6 +277,41 @@ const OrgOverlay: FC = () => {
...
+                  )}
+                  {limits?.stack && (
+                    <Grid.Row>
+                      <Grid.Column widthMD={Columns.Four}>
+                        <Form.Label label="Stack" />
+                        <LimitsField
+                          type={InputType.Text}
+                          name="stack.enabled"
+                          limits={limits}
+                          onChangeLimits={setLimits}
```
wdoconnell, line 310: I'm having a little trouble understanding how `setLimits` is working here. It looks like the trace is:

1. In `LimitsField`, a change in the value triggers the `onChange` function, which
2. sets the value of `newLimits` to `set(name, newValue, limits)`. That doesn't make sense to me - I would think the syntax here should be `set(limits, name, newValue)` https://lodash.com/docs/4.17.15#set
3. Run `onChangeLimits(newLimits)`, which is the `setLimits` function in the operator context.
4. That appears to refresh the current limits with `handleGetLimits`.
5. When the user hits `Submit Changes` in `OrgOverlay.tsx`, it triggers the `updateLimits` function, which does the following:
6. `const backendLImits = fromDisplayLimits(limits)` and then
7. `handleUpdateLimits(orgID, backendLimits)`
8. Which runs `putOperatorOrgsLimits({orgId: id, data: updatedLimits})

TLDR, I'm a little confused why `set` is even working here if it's in the wrong order, so I want to make sure we're not missing something on how the org limit update occurs.

package.json
```diff
@@ -53,7 +53,7 @@
     "cy": "CYPRESS_dexUrl=https://$INGRESS_HOST:$PORT_HTTPS CYPRESS_baseUrl=http://localhost:9999 cypress open",
     "cy:dev": "source ../monitor-ci/.env && CYPRESS_dexUrl=CLOUD CYPRESS_baseUrl=https://$INGRESS_HOST:$PORT_HTTPS cypress open --config testFiles='{cloud,shared}/**/*.*'",
     "cy:dev-oss": "source ../monitor-ci/.env && CYPRESS_dexUrl=OSS CYPRESS_baseUrl=https://$INGRESS_HOST:$PORT_HTTPS cypress open --config testFiles='{oss,shared}/**/*.*'",
-    "generate": "export SHA=689356a1c8e3dbd4e6ff6783cf81cbb720f8e64a && export REMOTE=https://raw.githubusercontent.com/influxdata/openapi/${SHA}/ && yarn generate-meta",
+    "generate": "export SHA=db14de0d91be00e75eb946d768e88d771b20a2a9 && export REMOTE=https://raw.githubusercontent.com/influxdata/openapi/${SHA}/ && yarn generate-meta",
```
wdoconnell, line 56: One thing I noticed in testing this locally: each of the three options for the `query` limits is a specific number, `0`, -`-1`, or anything else. But in the openapi spec that's generated, they're just 'any number'. So we should fix that so that it's `0` or `-1` or `number`.

wdoconnell:
Looks good, but would like to validate how `set` is working with these limits to make sure it's operating right.

src/operator/OrgOverlay.tsx
blegesse-git: Your trace is correct. And the syntax for lodash's set is `object, path, value`. But if you take a look at the LodashSet interface declared in the node modules, the syntax is `<T extends object>(path: lodash.PropertyPath, value: any, object: T): T;` - which is how the LimitsField component is using it. We're passing the path we want to modify first, then the value we want to change it to, followed by the object itself.  This is precisely how we want to update timeout and stack limits as well.

package.json
blegesse-git: Screen Shot 2023-02-15 at 9 00 01 AM
For timeout limits, the subfields should be of types `RestrictedLimit` `Unlimited` and `Limit`, which map to the numbers you mentioned. Check out the screenshot.

wdoconnell: Correct, but check what those map to - it's just a `number`.

src/operator/OrgOverlay.tsx
wdoconnell: 👍

wdoconnell (approved):
Looks good. Thanks for looking into this.

wdoconnell (approved):

PR #6625, opened 2023-02-21
chore: update account page to use quartz account deletion logic

blegesse-git:
Closes #[7108](https://github.com/influxdata/quartz/issues/7108)

Similar to the Operator UI Account Deletion logic, Pr updates Account page to use data returned from Quartz /account/:id end point, `isDeletable`, to determine whether user can delete an account. This piece of data will be stored in a context provider which will be available in the AccountPage component for consumption. 
Prior to this pr, only criteria used to determine account deletion ability in Account settings page is account type (free vs non-free - if free, allow deletion). Now, requirement for account deletion will be as follows: 
- The account must be free
- The account must have 1 or 0 orgs
- The account must have 1 or 0 users
- If there is a user in the account, the user must not belong to any other account (because we actually delete the user when we delete the account)

Account Details response coming from `/api/v2/quartz/accounts/accountID` endpoint 
Screen Shot 2023-02-21 at 12 25 03 PM

Account Settings Page Delete button displayed using isDeletable property
Screen Shot 2023-02-21 at 12 26 59 PM

Network calls when navigating to and from app homepage and Account Settings Page:
https://user-images.githubusercontent.com/66275100/220715553-816f5b79-5bd2-412f-bf1e-1d7b20ea1295.mov

### Checklist

### DIFF
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -33,12 +33,20 @@ import AccountTabContainer from 'src/accounts/AccountTabContainer'
 import AccountHeader from 'src/accounts/AccountHeader'
 import {DeleteFreeAccountProvider} from 'src/accounts/context/DeleteFreeAccountContext'
 
+// Types
+import {RemoteDataState} from 'src/types'
+
 // Styles
 import './AccountPageStyles.scss'
 
 const AccountAboutPage: FC = () => {
-  const {accountDetails, userAccounts, handleRenameActiveAccount} =
-    useContext(UserAccountContext)
+  const {
+    accountDetails,
+    accountDetailsStatus,
+    userAccounts,
+    handleGetAccountDetails,
+    handleRenameActiveAccount,
+  } = useContext(UserAccountContext)
 
   const {users} = useContext(UsersContext)
   const activeAccount =
@@ -49,6 +57,12 @@ const AccountAboutPage: FC = () => {
   useEffect(() => {
     setActiveAcctName(activeAccount?.name)
   }, [activeAccount])
+
+  useEffect(() => {
+    if (accountDetailsStatus === RemoteDataState.NotStarted) {
+      handleGetAccountDetails()
+    }
+  }, [accountDetailsStatus, handleGetAccountDetails])
 
   const updateAcctName = (evt: ChangeEvent<HTMLInputElement>) => {
     setActiveAcctName(evt.target.value)
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -43,9 +43,9 @@ const AccountAboutPage: FC = () => {
   const {
     accountDetails,
     accountDetailsStatus,
-    userAccounts,
     handleGetAccountDetails,
     handleRenameActiveAccount,
+    userAccounts,
   } = useContext(UserAccountContext)
 
   const {users} = useContext(UsersContext)
@@ -72,7 +72,9 @@ const AccountAboutPage: FC = () => {
     handleRenameActiveAccount(activeAccount.id, activeAcctName)
   }
 
-  const shouldShowDeleteFreeAccountButton = CLOUD && accountDetails?.isDeletable
+  const shouldShowDeleteFreeAccountButton = Boolean(
+    CLOUD && accountDetails?.isDeletable
+  )
 
   const showLeaveOrgButton = !isFlagEnabled('createDeleteOrgs')
   const allowSelfRemoval = users.length > 1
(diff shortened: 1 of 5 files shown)


wdoconnell:
For further context here: as explained in Dustin's issue https://github.com/influxdata/quartz/issues/7108, this logic is already implemented in quartz, so a user can't delete an account unless it's free; doesn't have more than one org in it; and doesn't have more than one user in it. 

What this PR is doing is ensuring that the UI doesn't allow the user to get to the 'Delete free account' button until those requirements are met.

src/accounts/AccountPage.tsx
```diff
@@ -33,15 +33,22 @@ import AccountTabContainer from 'src/accounts/AccountTabContainer'
...
 // Styles
 import './AccountPageStyles.scss'
 
 const AccountAboutPage: FC = () => {
-  const {userAccounts, handleRenameActiveAccount} =
-    useContext(UserAccountContext)
+  const {
+    accountDetails,
+    accountDetailsStatus,
+    userAccounts,
```
wdoconnell, line 46: alpha issue here ^

src/accounts/context/userAccount.tsx
```diff
@@ -65,6 +78,11 @@ export const UserAccountProvider: FC<Props> = React.memo(({children}) => {
   const [userAccounts, setUserAccounts] = useState<UserAccount[]>(null)
   const [defaultAccountId, setDefaultAccountId] = useState<number>(null)
   const [activeAccountId, setActiveAccountId] = useState<number>(null)
+  const [accountDetails, setAccountDetails] = useState<Account>(null)
+  const [accountDetailsStatus, setAccountDetailsStatus] = useState(
+    RemoteDataState.NotStarted
+  )
+  const {account} = useSelector(selectCurrentIdentity)
```
wdoconnell, line 85: Is there a `selectCurrentAccount` selector?

src/accounts/AccountPage.tsx
```diff
@@ -60,7 +71,8 @@ const AccountAboutPage: FC = () => {
   const onRenameAccountBtnClick = () => {
     handleRenameActiveAccount(activeAccount.id, activeAcctName)
   }
-  const shouldShowDeleteFreeAccountButton = CLOUD && account.type === 'free'
+
+  const shouldShowDeleteFreeAccountButton = CLOUD && accountDetails?.isDeletable
```
wdoconnell, line 75: For clarity, and to make sure we don't get weird behavior if the `isDeletable` property values change, I think this should be

`Boolean(CLOUD && accountDetails?.isDeletable)`

src/accounts/AccountPage.tsx
```diff
@@ -51,7 +58,11 @@ const AccountAboutPage: FC = () => {
     setActiveAcctName(activeAccount?.name)
   }, [activeAccount])
 
-  const {account} = useSelector(selectCurrentIdentity)
+  useEffect(() => {
+    if (accountDetailsStatus === RemoteDataState.NotStarted) {
+      handleGetAccountDetails()
+    }
+  }, [accountDetailsStatus, handleGetAccountDetails])
```
wdoconnell, line 65: Great 👍

wdoconnell (commented):
Looks good - minor changes. Would you also mind posting a video of the network calls on loading the app, just to demonstrate that we don't always call this endpoint (only when hitting the account/settings page)?

wdoconnell:
FYI for visibility, @gunnaraasen - https://github.com/influxdata/ui/pull/6625#issuecomment-1439214456

src/accounts/AccountPage.tsx
blegesse-git: good catch

src/accounts/AccountPage.tsx
blegesse-git: yup, I think that will serve as a nice safeguard

src/accounts/context/userAccount.tsx
blegesse-git: yes indeed. I'll switch it out

src/accounts/context/userAccount.tsx
```diff
@@ -38,24 +45,30 @@ interface SetDefaultAccountOptions {
...
+  accountDetailsStatus: RemoteDataState.NotStarted,
   activeAccountId: -1,
+  defaultAccountId: -1,
+  handleGetAccountDetails: () => {},
   handleGetAccounts: () => {},
-  handleSetDefaultAccount: () => {},
   handleRenameActiveAccount: () => {},
+  handleSetDefaultAccount: () => {},
+  userAccounts: [],
 }
```
wdoconnell, line 72: Thanks for the extra alpha 😃

wdoconnell (dismissed):
Awesome work - thank you for accommodating my comments.

wdoconnell (approved):

PR #6643, opened 2023-03-03
chore: remove multiorg phase2 feat flag

wdoconnell:
Part of #5243

Removes the feature flag associated with multiorg phase two (creating and deleting orgs). After this PR, the flag still needs to be removed from IDPE, and then configcat.

### Checklist

### DIFF
--- a/src/accounts/AccountPage.tsx
+++ b/src/accounts/AccountPage.tsx
@@ -12,10 +12,9 @@ import {
   InputType,
   Page,
 } from '@influxdata/clockface'
-import {LeaveOrgButton} from 'src/organizations/components/OrgProfileTab/LeaveOrg'
 
 // Context
-import {UsersContext, UsersProvider} from 'src/users/context/users'
+import {UsersProvider} from 'src/users/context/users'
 import {DeleteFreeAccountButton} from 'src/accounts/DeleteFreeAccount'
 import {DeleteFreeAccountOverlay} from 'src/accounts/DeleteFreeAccountOverlay'
 
@@ -26,7 +25,6 @@ import {selectCurrentIdentity} from 'src/identity/selectors'
 import {CLOUD} from 'src/shared/constants'
 
 // Utils
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
 import {UserAccountContext} from 'src/accounts/context/userAccount'
 import AccountTabContainer from 'src/accounts/AccountTabContainer'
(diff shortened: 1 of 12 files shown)


src/accounts/AccountPage.tsx
```diff
@@ -105,12 +99,6 @@ const AccountAboutPage: FC = () => {
             text="Save"
           />
         </FlexBox>
-        {allowSelfRemoval && showLeaveOrgButton && (
-          <>
-            <hr className="account-settings--divider" />
-            <LeaveOrgButton />
-          </>
-        )}
         {shouldShowDeleteFreeAccountButton && <DeleteFreeAccountButton />}
```
wdoconnell, line 102: The "Leave Org" button was only here for backwards compatibility for situations where the feature flag was turned off. It now lives on the org settings page.

src/accounts/AccountTabs.tsx
```diff
@@ -35,7 +37,7 @@ const AccountTabs: FC<Props> = ({activeTab}) => {
     },
     {
       id: 'organizations',
-      enabled: isFlagEnabled('createDeleteOrgs'),
+      enabled: CLOUD,
       link: `/orgs/${orgID}/accounts/orglist`,
```
wdoconnell, line 41: The organization tab cannot be accessed from OSS, where there is no multi-org or access to quartz APIs.

src/identity/components/GlobalHeader/AccountDropdown.tsx
```diff
@@ -47,31 +44,25 @@ export const AccountDropdown: FC<Props> = ({
...
-          enabled: true,
-        },
-      ].filter(menuOption => menuOption.enabled),
-    [activeOrg.id, isCreateDeleteFlagOn]
+    () => [
+      {
+        name: 'Settings',
+        iconFont: IconFont.CogSolid_New,
+        href: `/orgs/${activeOrg.id}/accounts/settings`,
+      },
```
wdoconnell, line 53: There's no need for the 'enabled' distinction here, as all of these tabs are accessible without condition. The global header only exists in cloud, so there's no need for a CLOUD guard either.

src/organizations/components/OrgProfileTab/index.tsx
```diff
@@ -175,7 +171,7 @@ const OrgProfileTab: FC = () => {
           stretchToFitWidth={true}
         >
           <>
-            {allowSelfRemoval && showLeaveOrgButton && <LeaveOrgButton />}
```
wdoconnell, line 178: This is already subordinate to a `CLOUD` check, so we don't need another one.

PR #6651, opened 2023-03-13
feat: add operator token to onboarding wizard

jeffreyssmith2nd:
Closes https://github.com/influxdata/influxdb/issues/24082

After updating the UI in OSS, we lost the ability to see the operator token after creation. This causes a lot of headache and confusion for users.

I have simply added the token to the final page in the onboarding wizard.

image

### Checklist

### DIFF
--- a/src/onboarding/actions/index.ts
+++ b/src/onboarding/actions/index.ts
@@ -20,6 +20,7 @@ export type Action =
   | SetStepStatus
   | SetOrganizationID
   | SetBucketID
+  | SetToken
 
 interface SetSetupParams {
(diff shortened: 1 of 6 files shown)


wdoconnell (commented):
Looks good generally, and thanks for doing this. But do we need a cloud check here? 

From what I've seen in the code, this component is shared by OSS and Cloud, and if so, I don't think we want to dispatch the `SET_OPERATOR_TOKEN` action outside of OSS.

jeffreyssmith2nd:
> But do we need a cloud check here?

Good catch, I initially assumed it was OSS only. Updated with Cloud check.

PR #6679, opened 2023-04-26
feat(ui): Add Blogs to Homepage

mavarius:
This PR brings a feed of the latest blog posts into the UI. The blogs are listed under the "What's New" section under the announcements. If the blogs fail to load, only the announcements are shown.

This is part of an effort to present users with pertinent information to help them make use of the platform. The devrels are writing good blog content about how to use the new database and much of that content isn't seen by current users and isn't in our documentation.

image

### Checklist

### DIFF
--- a/src/me/components/BlogFeed.tsx
+++ b/src/me/components/BlogFeed.tsx
@@ -1,0 +1,145 @@
+// Libraries
+import React, {FC, useState, useEffect} from 'react'
+import moment from 'moment'
+
+// Components
+import {
+  Heading,
+  HeadingElement,
+  IconFont,
+  RemoteDataState,
+  SpinnerContainer,
+  TechnoSpinner,
+} from '@influxdata/clockface'
+import AnnouncementBlock from 'src/me/components/AnnouncementBlock'
+import {SafeBlankLink} from 'src/utils/SafeBlankLink'
+
+// Utils
+import {event} from 'src/cloud/utils/reporting'
+
+// Styles
+import 'src/me/components/BlogFeed.scss'
+
+// Expected structure of blog JSON data
+interface Author {
+  name: string
+  url: string
+  avatar?: string
+}
+
+interface BlogItem {
+  url: string
+  title: string
+  content_text: string
+  date_published: string
+  date_modified: string
+  authors?: Author[]
+  author: Author
+  tags: string[]
+  attachments: {
+    url: string
+    mime_type: string
+    title: string
+    size_in_bytes: number
+    duration_in_seconds: number
+  }[]
+}
+
+interface BlogFeedData {
+  version: string
+  title: string
+  home_page_url: string
+  feed_url: string
+  description: string
+  items: BlogItem[]
+  language: string
+}
+
+export const BlogFeed: FC = () => {
+  const [jsonData, setJsonData] = useState<BlogFeedData | null>(null)
+  const [error, setError] = useState<string | null>(null)
+
+  useEffect(() => {
+    const abortController = new AbortController()
+
+    const fetchData = () => {
+      fetch('https://www.influxdata.com/blog/feed.json', {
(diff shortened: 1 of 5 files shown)


src/me/components/Resources.tsx
```diff
@@ -59,6 +60,7 @@ const ResourceLists: FC = () => {
             ctaText="Learn more"
             title="New time-series engine for InfluxDB"
           />
+          <BlogFeed />
```
wdoconnell, line 63: If we feature flag this with a temp flag, we can test first to see if we're running into a CORS issue before folks generally see it in prod.

src/me/components/Resources.tsx
```diff
@@ -15,8 +15,9 @@ import {
   Heading,
   HeadingElement,
 } from '@influxdata/clockface'
-import AnnouncementCenter from './AnnouncementCenter'
```
wdoconnell, line 18: While you're in the file, would you mind changing the component name ("Resources") to something a little more descriptive re what it's doing? And that also clarifies it's a cloud component?

Since everything above it in the hierarchy is potentially an OSS component, you have to browse through the full history to figure out that this is actually a cloud-only component, but only because of the `CLOUD ?` check in HomePageContents.tsx and HomePageContentsTsm.tsx.

src/App.tsx --> SetOrg --> route for `HomePageContainer` --> HomepageContents / HomepageContentsTSM --> CLOUD check confirms this component isn't used in OSS.

src/me/components/Resources.tsx
```diff
@@ -59,6 +60,7 @@ const ResourceLists: FC = () => {
             ctaText="Learn more"
             title="New time-series engine for InfluxDB"
           />
+          <BlogFeed />
         </AnnouncementCenter>
       ) : (
         <DocSearchWidget />
```
wdoconnell, line 66: Also, would you mind getting rid of the default export? 😄

wdoconnell (commented):
This looks good, but had a comment re: potentially flagging this to debug any CORS issue, and renaming this component for clarity.

src/me/components/Resources.tsx
appletreeisyellow: +1 about having a temporary feature flag to test in prod

mavarius: done

src/me/components/Resources.tsx
mavarius: done

src/me/components/BlogFeed.tsx
```diff
@@ -0,0 +1,145 @@
...
+            />
+          ))}
+        <div className="announcement-block--more-blogs">
+          <SafeBlankLink
+            href="https://www.influxdata.com/blog/"
+            onClick={handleMoreBlogsClick}
+          >
+            <div className="cf-button cf-button-xs cf-button-tertiary">
+              <span className="cf-button-label">more blogs</span>
+            </div>
```
appletreeisyellow, line 139: A question for my future UI development: does `cf-button-*` class names defined in clockface? Is [this](https://github.com/influxdata/clockface/blob/e6fd47e31584936a70bed188b9600e6298e2f211/src/Components/Button/Button.scss#L9) where it is defined? I hope I can reuse some of them and not repeating myself defining them again in UI.

appletreeisyellow (approved):
Great work! I like you put into the consideration of cancelling ongoing fetch requests when component unmount!

src/me/components/BlogFeed.tsx
mavarius: Ya, if you ever want to just copy the styles of a component, like in this case where I wanted a link to look like a button, you can just copy the class names from it.

wdoconnell (approved):
Thanks for accommodating my comments 👍

PR #6693, opened 2023-05-23
feat(influxql): add time range into query composition

### LINKED ISSUE #6653: Add date picker timestamp in InfluxQL query composition

appletreeisyellow:
Closes #6653

This PR adds the time range from date picker into the InfluxQL query composition. I also add unit tests to cover the utility function that converts UI date picker time range into InfluxQL time syntax.

## After

https://github.com/influxdata/ui/assets/14298407/38b1fa36-55b7-4fe0-b8f1-444739b64b0c

All the work is behind the feature flag `influxqlUI`.

To test on local remocal:

1. create DBRP mappings in your org ([instruction](https://github.com/influxdata/idpe/issues/17205)), and
2. also make sure to turn on the following flags in your browser:
    * `showOldDataExplorerInNewIOx`
    * `schemaComposition`

### Checklist

### DIFF
--- a/src/languageSupport/languages/influxql/connection.ts
+++ b/src/languageSupport/languages/influxql/connection.ts
@@ -15,7 +15,7 @@ import {LspRange} from 'src/languageSupport/languages/agnostic/types'
 
 // Utils
 import {DEFAULT_TIME_RANGE} from 'src/shared/constants/timeRanges'
-import {rangeToInterval as buildTimeRange} from 'src/shared/utils/sqlInterval'
+import {rangeToSQLInterval as buildTimeRange} from 'src/shared/utils/rangeToInterval'
 import {notify} from 'src/shared/actions/notifications'
 import {compositionEnded} from 'src/shared/copy/notifications'
 import {groupedTagValues} from 'src/languageSupport/languages/agnostic/utils'
--- a/src/languageSupport/languages/influxql/connection.ts
+++ b/src/languageSupport/languages/influxql/connection.ts
@@ -15,7 +15,7 @@ import {LspRange} from 'src/languageSupport/languages/agnostic/types'
 
 // Utils
 import {DEFAULT_TIME_RANGE} from 'src/shared/constants/timeRanges'
-import {rangeToSQLInterval as buildTimeRange} from 'src/shared/utils/rangeToInterval'
+import {rangeToInfluxQLInterval as buildTimeRange} from 'src/shared/utils/rangeToInterval'
 import {notify} from 'src/shared/actions/notifications'
 import {compositionEnded} from 'src/shared/copy/notifications'
 import {groupedTagValues} from 'src/languageSupport/languages/agnostic/utils'
--- a/src/languageSupport/languages/influxql/connection.ts
+++ b/src/languageSupport/languages/influxql/connection.ts
@@ -15,9 +15,11 @@ import {LspRange} from 'src/languageSupport/languages/agnostic/types'
 
 // Utils
 import {DEFAULT_TIME_RANGE} from 'src/shared/constants/timeRanges'
+import {rangeToInterval as buildTimeRange} from 'src/shared/utils/sqlInterval'
 import {notify} from 'src/shared/actions/notifications'
 import {compositionEnded} from 'src/shared/copy/notifications'
 import {groupedTagValues} from 'src/languageSupport/languages/agnostic/utils'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 export class ConnectionManager extends AgnosticConnectionManager {
   private _timeRange: TimeRange = DEFAULT_TIME_RANGE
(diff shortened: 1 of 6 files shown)


src/languageSupport/languages/influxql/connection.ts
```diff
@@ -53,19 +55,18 @@ export class ConnectionManager extends AgnosticConnectionManager {
       groupedTagValues(this._session.tagValues)
     )
       .map(([key, values]) =>
-        values.map((value: string) => `"${key}" = '${value}'`).join(' AND ')
+        values.map((value: string) => `"${key}" = '${value}'`).join(' OR ')
```
appletreeisyellow, line 58: This was a bug fix. Changing `AND` to `OR` for same tag key with different tag values.

src/languageSupport/languages/influxql/connection.ts
```diff
@@ -178,7 +179,7 @@ export class ConnectionManager extends AgnosticConnectionManager {
       sessionCb,
       dispatch
     )
-    if (!shouldContinue) {
+    if (!shouldContinue || !isFlagEnabled('schemaComposition')) {
```
appletreeisyellow, line 182: Adding the feature flag `schemaComposition` to control the query composition effect. If the feature flag is off, users should explorer the schema list in the left sidebar without getting the auto query composition in the text editor.

src/shared/utils/rangeToInterval.test.ts
```diff
@@ -0,0 +1,159 @@
+import {
+  rangeToSQLInterval,
+  rangeToInfluxQLInterval,
+} from 'src/shared/utils/rangeToInterval'
+import {SELECTABLE_TIME_RANGES} from 'src/shared/constants/timeRanges'
+import {DurationTimeRange} from 'src/types'
+
+describe('rangeToSQLInterval:', () => {
```
appletreeisyellow, line 8: The tests under `rangeToSQLInterval` are copied from the old file `src/shared/utils/sqlInterval.test.ts` which is deleted/renamed in this PR.

src/shared/utils/rangeToInterval.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    ]
+    toTest.forEach(({msg, input, expected}) => {
+      it(msg, () => {
+        expect(rangeToSQLInterval(input as DurationTimeRange)).toEqual(expected)
+      })
+    })
+  })
+})
+
+describe('rangeToInfluxQLInterval:', () => {
```
appletreeisyellow, line 78: The following tests are newly added in this PR

wdoconnell:
@appletreeisyellow Do you need a review on this, or is it waiting for @wiedld ?

appletreeisyellow:
@wdoconnell it would be good if you can review it since Denise focuses on write team now. Thank you!

wdoconnell:
You got it, thanks for clarifying - I will give this a review after my next meeting. 👍

src/languageSupport/languages/influxql/connection.ts
wdoconnell: What is this a bug fix for - is there an issue for this?

src/shared/utils/rangeToInterval.ts
```diff
@@ -67,23 +67,65 @@ const convertToInterval = (upperOrLower: string) => {
...
+ *     https://docs.influxdata.com/influxdb/cloud/query-data/influxql/explore-data/time-and-timezone/#time-syntax
+ */
+const convertToInfluxQLTimeSyntax = (upperOrLower: string): string => {
+  if (upperOrLower === null || upperOrLower === 'now()') {
+    return 'now()'
+  }
+
+  // relative time
+  if (!!upperOrLower.match(durationRegExp)) {
+    const durationIntoPast = upperOrLower.charAt(0) === '-'
```
wdoconnell, line 101: This is more of an 'fyi', but we tried for a while to replace `!!` operators with the `Boolean()` function just because it's clearer. But do whatever works / is more common in the context of these files which may be different. 😅

wdoconnell (commented):
Looks good - I just had one question here about what bug is being fixed by the `AND` to `OR` conversion, since unless I'm mistaken, it looks like that part is not flagged.

src/shared/utils/rangeToInterval.ts
appletreeisyellow: Good to know! Thanks, I think `Boolean()` is clearer. Updated!

src/languageSupport/languages/influxql/connection.ts
appletreeisyellow: No, I didn't file an issue because I found this bug when I went through this PR. The bug is not exposed to anyone because InfluxQL is behind the feature flag `influxqlUI` and the flag is off currently. 

The bug was, if using `AND` to join tag values with the same tag key, the query will result in zero row. For example,

```sql
SELECT "air_temp_degc"
FROM database_name.retention_policy_name."ndbc"
WHERE time >= now() - 1h
  AND ("station_id" = '13001' AND "station_id" = '13002' AND "station_type" = 'fixed')
```
The above query will result a zero row.

This fix is to change the above query to the following:

```sql
SELECT "air_temp_degc"
FROM database_name.retention_policy_name."ndbc"
WHERE time >= now() - 1h
  AND ("station_id" = '13001' OR "station_id" = '13002' AND "station_type" = 'fixed')
```

Sorry, it is confusing that this bug fix is not related to this PR 😅 I can move this bug fix to a separate PR if it is clearer

wdoconnell: No, that makes sense! Please keep in the same PR - I don't want to make this more burdensome than it needs to be. Just wanted to make sure there wasn't something else we could close down here.

wdoconnell:
@appletreeisyellow The failure in that test rerun is caused by a recent update to quartz-mock that just went in. I worked on it with @alejom99  so I can push up a fix real quick.
https://github.com/influxdata/quartz-mock/pull/51

wdoconnell:
fyi @appletreeisyellow  - https://github.com/influxdata/ui/pull/6698

wdoconnell:
@appletreeisyellow FYI, Alejandro is fixing it here - https://github.com/influxdata/ui/pull/6693

appletreeisyellow:
@wdoconnell Thank you for sorting the CI pipeline out and keeping me updated!

wdoconnell (approved):
Looks good - thank you for your patience as we sorted out the CI issue.

PR #6699, opened 2023-05-24
feat(operator): Allow operators to cancel accounts

alejom99:
This change allows a user to cancel a direct payg account through the operator page.

### Checklist

### DIFF
--- a/src/operator/account/AccountViewHeader.tsx
+++ b/src/operator/account/AccountViewHeader.tsx
@@ -22,6 +22,8 @@ const AccountViewHeader: FC = () => {
     account,
     setConvertToContractOverlayVisible,
     convertToContractOverlayVisible,
+    setCancelOverlayVisible,
+    cancelOverlayVisible,
     setDeleteOverlayVisible,
     deleteOverlayVisible,
   } = useContext(AccountContext)
@@ -59,19 +61,24 @@ const AccountViewHeader: FC = () => {
           Convert to Contract
         </ButtonBase>
       )}
-      {hasWritePermissions && (
+      {hasWritePermissions && account?.deletable && (
         <ButtonBase
           color={ComponentColor.Danger}
           shape={ButtonShape.Default}
           onClick={() => setDeleteOverlayVisible(!deleteOverlayVisible)}
-          status={
-            account?.deletable
-              ? ComponentStatus.Default
-              : ComponentStatus.Disabled
-          }
           testID="account-delete--button"
         >
           Delete Account
+        </ButtonBase>
+      )}
+      {hasWritePermissions && account?.cancellable && (
+        <ButtonBase
+          color={ComponentColor.Danger}
+          shape={ButtonShape.Default}
+          onClick={() => setCancelOverlayVisible(!cancelOverlayVisible)}
+          testID="account-cancel--button"
+        >
+          Cancel Account
         </ButtonBase>
       )}
     </FlexBox>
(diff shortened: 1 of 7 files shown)


src/operator/account/AccountViewHeader.tsx
```diff
@@ -59,21 +61,26 @@ const AccountViewHeader: FC = () => {
...
           testID="account-delete--button"
         >
           Delete Account
         </ButtonBase>
       )}
+      {hasWritePermissions && account?.cancellable && (
+        <ButtonBase
+          color={ComponentColor.Danger}
+          shape={ButtonShape.Default}
+          onClick={() => setCancelOverlayVisible(!cancelOverlayVisible)}
```
wdoconnell, line 78: Same here- do we need the bang operator when we can just set `setCancelOverlayVisible` to true, since the button shows the overlay rather than toggling it?

src/operator/account/CancelAccountOverlay.tsx
```diff
@@ -0,0 +1,75 @@
+import React, {FC, useContext} from 'react'
+import {
```
wdoconnell, line 2: Normally we would alpha-order these for ease of identifying the components, but probably not worth re-running CI for that

src/operator/account/AccountViewHeader.tsx
```diff
@@ -59,21 +61,26 @@ const AccountViewHeader: FC = () => {
           Convert to Contract
         </ButtonBase>
       )}
-      {hasWritePermissions && (
+      {hasWritePermissions && account?.deletable && (
         <ButtonBase
           color={ComponentColor.Danger}
           shape={ButtonShape.Default}
           onClick={() => setDeleteOverlayVisible(!deleteOverlayVisible)}
-          status={
```
wdoconnell, line 67: Isn't this a one-way operation - if you click the button, the visibility state always sets to `true` (instead of being toggled)?

src/operator/account/CancelAccountOverlay.tsx
```diff
@@ -0,0 +1,75 @@
...
+    cancelOverlayVisible,
+  } = useContext(AccountContext)
+
+  const cancelAccount = () => {
+    if (account?.cancellable) {
+      try {
+        handleDeleteAccount()
+      } catch (e) {
+        setCancelOverlayVisible(false)
+      }
```
wdoconnell, line 27: Doesn't need to happen in this PR, but I think this may confuse folks since there's no notification anything went wrong - if deleteAccount fails, the overlay will just close. 

We may want a follow-on PR adding a pop-up notification indicating that the account couldn't be deleted (there is a `dispatch(notify)` function for that in the UI).

src/operator/account/CancelAccountOverlay.tsx
```diff
@@ -0,0 +1,75 @@
...
+      )}
+      testID="cancel-overlay"
+      transitionDuration={0}
+    >
+      <Overlay.Container maxWidth={600}>
+        <Overlay.Header
+          title="Cancel Account"
+          style={{color: '#FFFFFF'}}
+          onDismiss={() => setCancelOverlayVisible(!cancelOverlayVisible)}
+        />
```
wdoconnell, line 50: Same comment here - I think dismissing should just set the cancel overlay to false rather than flipping a toggle.

wdoconnell (dismissed):
Looks good, but one suggestion for a follow-on PR to handle error state from canceling an account.

wdoconnell (approved):

PR #6709, opened 2023-06-01
feat(ui): add grafana announcement

mavarius:
Adds an announcement for the Grafana Flight SQL plugin and highlights the option in the resource list.

image

### Checklist

### DIFF
--- a/src/homepageExperience/components/OptionAccordion/OptionAccordionElement.tsx
+++ b/src/homepageExperience/components/OptionAccordion/OptionAccordionElement.tsx
@@ -5,7 +5,7 @@ import React, {FC} from 'react'
 import {Accordion, Heading, HeadingElement} from '@influxdata/clockface'
 
 interface OwnProps {
-  elementTitle: string
+  elementTitle: string | JSX.Element
   elementDescription?: string
   bodyContent?: () => JSX.Element
   cta?: () => JSX.Element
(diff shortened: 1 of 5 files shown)


src/me/components/CloudWidgets.tsx
```diff
@@ -46,26 +45,23 @@ export const CloudWidgets: FC = () => {
             body={
               <>
                 <p>
```
wdoconnell, line 47: We're also intentionally removing this info re: cloud dedicated, right? (since it's a past event?)

src/homepageExperience/components/OptionAccordion/OptionAccordionElement.tsx
```diff
@@ -5,7 +5,7 @@ import React, {FC} from 'react'
 import {Accordion, Heading, HeadingElement} from '@influxdata/clockface'
 
 interface OwnProps {
-  elementTitle: string
+  elementTitle: string | JSX.Element
   elementDescription?: string
```
wdoconnell, line 9: LGTM: since `OptionAccordionElement` renders `{elementTitle}` and doesn't rely on it having a specific type, this works swell.

src/me/components/CloudWidgets.tsx
wdoconnell: 👍

PR #6710, opened 2023-06-02
feat(ui): remove support links

### LINKED ISSUE #6688: Remove "Contact Support" and "Report a bug" links for non-contract accounts
Because of a change in service, only annual account customers, or customers that are specifically paying for support will be authorized for tech support services by default.

Because of this, we need to remove the "Contact Support" link for any non annual account.  And we should remove the "Report a bug" link in the Cloud2 UI since that is no longer being monitored.

This is from EPIC:  [https://github.com/influxdata/quartz/issues/7257]

Assigning to @mavarius (thank you) for completion.

mavarius:
Closes #6688 

This PR removes the "contact support" for non-contract users. It also removes the "report a bug" link for all users.
image

### Checklist


src/support/components/ConfirmationOverlay.tsx
```diff
@@ -80,12 +80,14 @@ const ConfirmationOverlay: FC<OwnProps> = () => {
       <Overlay.Header
         testID="confirmation-overlay-header"
         title={
-          params.type === 'PAYG' ? 'Contact Support' : 'Feedback & Questions'
+          params.type === 'contract'
+            ? 'Contact Support'
+            : 'Feedback & Questions'
         }
```
wdoconnell, line 86: I see that you can still end up in this overlay in other account types from FeedbackQuestionsOverlay. What path does the user need to follow to end up in FeedbackQuestionsOverlay? I don't see it mentioned as a reference anywhere other than the overlay controller.

src/support/components/ContactSupportOverlay.tsx
```diff
@@ -229,19 +229,17 @@ const PayGSupportOverlay: FC<OwnProps> = () => {
...
-          testID="payg-contact-support--submit"
+          testID="contact-support--submit"
           status={submitButtonStatus}
           onClick={handleSubmit}
         />
       </Overlay.Footer>
     </Overlay.Container>
   )
 }
-
```
wdoconnell, line 246: Yay 🥳

wdoconnell (commented):
Looks good to me, just a couple minor questions.

src/me/components/DocSearchWidget.tsx
```diff
@@ -40,11 +40,6 @@ const supportLinks = [
     title: 'Write a Task',
     eventName: 'WriteTask',
   },
-  {
-    link: 'https://github.com/influxdata/ui/issues/new',
-    title: 'Report a bug',
-    eventName: 'GitBugs',
-  },
   {
```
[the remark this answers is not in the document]
mavarius: No one is monitoring this anymore so it was decided to remove it entirely.

src/support/components/ConfirmationOverlay.tsx
mavarius: Hm, good point, doesn't look like there's another way to get to this. I think it may be a holdover from when the form had the option to submit feedback without having a severity level assigned. I'll remove it.

mavarius: updated

wdoconnell: Taking a look in a moment

wdoconnell:
Re-running CI since it looks like this flaked on an unrelated explorer test.

wdoconnell (approved):
Looks good - thanks for making those adjustments.

Issue #6649, opened 2023-03-10
Refactor: replace persistance with persistence 


appletreeisyellow:
The typo was introduced early on when `src/dataExplorer/context/persistance.tsx` file was created. It should be an easy refactor

Here are the typos:

https://github.com/search?q=repo%3Ainfluxdata%2Fui%20PersistanceContext&type=code
https://github.com/search?q=repo%3Ainfluxdata%2Fui+PersistanceProvider&type=code
https://github.com/search?q=repo%3Ainfluxdata%2Fui%20src%2FdataExplorer%2Fcontext%2Fpersistance&type=code

PR #6713, opened 2023-06-05
feat(influxql): integrate InfluxQL '/query'

### LINKED ISSUE #6632: Integrate with InfluxQL query path to execute query

appletreeisyellow:
Closes #6632 

This PR integrates InfluxQL v1 API `/query` into UI to query data using InfluxQL

## After

https://github.com/influxdata/ui/assets/14298407/c9c15449-7750-4a1e-ad2a-e933f3628805

All the work is behind the feature flag `influxqlUI`.

To test on local remocal:

1. create DBRP mappings in your org ([instruction](https://github.com/influxdata/idpe/issues/17205)), and
2. also make sure to turn on the following flags in your browser:
    * `showOldDataExplorerInNewIOx`
    * `schemaComposition`

### Checklist


src/dataExplorer/components/Results.tsx
```diff
@@ -34,15 +34,16 @@ import {PersistanceContext} from 'src/dataExplorer/context/persistance'
...
 
 // Utils
 import {bytesFormatter} from 'src/shared/copy/notifications'
 
 import './Results.scss'
-import {LanguageType} from './resources'
 
 const QueryStat: FC = () => {
   const {result} = useContext(ResultsContext)
+  const {resource} = useContext(PersistanceContext)
```
wdoconnell, line 47: Super minor (and could go in the next PR), but think this should be `PersistenceContext`.

src/dataExplorer/components/Results.tsx
```diff
@@ -142,7 +145,11 @@ const TableResults: FC<{search: string}> = ({search}) => {
 
   return (
     <div
-      className="data-explorer-results--view"
```
wdoconnell, line 145: Defer to you, but could probably use the `classNames` library to make this logic a little simpler.

src/shared/contexts/query/context.tsx
```diff
@@ -98,16 +103,49 @@ export const QueryContext =
...
+): {
+  url: string
+  body: RequestBody | null
+  headers: {
+    'Content-Type': string
+    'Accept-Encoding': string
+    Accept?: string
+  }
+} => {
   const mechanism = options?.overrideMechanism ?? OverrideMechanism.Extern
```
wdoconnell, line 119: Will these changes (above) affect all queries, not just influxQL queries?

src/shared/utils/addAnnotationToCSV.ts
```diff
@@ -0,0 +1,30 @@
+// Annotation row format:
+// https://docs.influxdata.com/influxdb/cloud/reference/syntax/annotated-csv
+export const addAnnotationToCSV = (csv: string): string => {
+  if (!csv) {
+    return csv
+  }
```
wdoconnell, line 6: What is the point of this check to determine if `csv` is falsy? Are we implicitly trying to check for an empty string `''`, or some other value? And if so, why are we returning the falsy value, instead of a fixed value (like `''`)?

wdoconnell (commented):
Looks good, but I had a couple of questions regarding portions that I thought might have an impact outside of the currently flagged logic.

src/dataExplorer/components/Results.tsx
appletreeisyellow: You got it! It's tracked in https://github.com/influxdata/ui/issues/6649 and I should have sometime this week to refactor it

src/shared/contexts/query/context.tsx
appletreeisyellow: I manually tested Flux and SQL for the above change and they won't affect other queries. And the CI tests all passed, so the changes are fine 👍 Thanks for asking!

src/dataExplorer/components/Results.tsx
appletreeisyellow: Good to know about this library! Updated

src/shared/utils/addAnnotationToCSV.ts
appletreeisyellow: Good point! Updated

src/dataExplorer/components/Results.tsx
wdoconnell: Thanks!

wdoconnell (approved):
Looks good. Thanks for accommodating my questions and comments. 👍

PR #6722, opened 2023-06-07
fix(sql): disable the run button until a measurement is selected

### LINKED ISSUE #6601: User Confusion when no table/measurement is chosen
If a user utilizes the SQL/Script Editor and selects only a bucket without a measurement and hits run (without manually modifying the SQL statement), the user will get back a long complicated error message, like:

executing query: failed to create Flight record reader: arrow/flight: could not create flight reader: arrow/ipc: could not read schema from stream: arrow/ipc: could not read message schema: rpc error: code = InvalidArgument desc = Error while planning query: Schema error: No field named 'time'.

We need to come up with a way to make this a better user experience.  Some thoughts:

Could we supercede this from happening by disabling the run button until a from exists?  This may not be possible since we aren't examining the SQL statement, but then could we do it when the user hits "run" (check to make sure it has a "from" specified) before we execute the query and inform the user they need to add a from clause in their statement?

Does it make sense to enhance the lower levels to report a more friendly message (vs. making the UI smarter) to address this?

appletreeisyellow:
Closes #6601

I came across the origin issue and found that it is a low hanging fruit. 

This PR disables the run button for SQL script editor until a measurement is selected when the composition sync is on. A message will show up when users hover over the disabled run button. 

## Before

https://github.com/influxdata/ui/assets/14298407/0ec88005-74ec-4999-8eb2-6e80feb5e98b

## After

Screen Shot 2023-06-07 at 11 07 55 AM

https://github.com/influxdata/ui/assets/14298407/9c03dffc-db70-411f-9ee1-e88673a57b74

### Checklist

### DIFF
--- a/src/dataExplorer/components/ResultsPane.tsx
+++ b/src/dataExplorer/components/ResultsPane.tsx
@@ -96,6 +96,14 @@ const ResultsPane: FC = () => {
   } else if (language === LanguageType.SQL && !selection.bucket) {
     submitButtonDisabled = true
     disabledTitleText = 'Select a bucket before running script'
+  } else if (
+    language === LanguageType.SQL &&
+    selection.composition.synced && // using composition
+    selection.bucket &&
+    !Boolean(selection.measurement)
+  ) {
+    submitButtonDisabled = true
+    disabledTitleText = 'Select a measurement before running script'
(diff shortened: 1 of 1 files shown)


wdoconnell (approved):
Nice find. 👍

PR #6748, opened 2023-06-28
feat(operators): Add Migrate Organization Tool

abshierjoel:
Allows operators to migrate all of an account's organizations or a specific organization to a different account.

image

### Checklist

### DIFF
--- a/src/operator/MigrateOrg.tsx
+++ b/src/operator/MigrateOrg.tsx
@@ -1,0 +1,83 @@
+import {
+  AlignItems,
+  Button,
+  ButtonType,
+  ComponentColor,
+  ComponentSize,
+  ComponentStatus,
+  FlexBox,
+  FlexDirection,
+  Form,
+  Input,
+} from '@influxdata/clockface'
+import React, {ChangeEvent, FC, useContext, useState} from 'react'
+import {OverlayContext} from 'src/operator/context/overlay'
+import {OperatorAccount, getOperatorAccount} from 'src/client/unityRoutes'
+import MigrateOrgOverlay from './MigrateOrgOverlay'
+
+const MigrateOrg: FC = () => {
+  const {organization, setMigrateOverlayVisible} = useContext(OverlayContext)
+  const [toAccountId, setToAccountId] = useState('')
+  const [toAccount, setToAccount] = useState<OperatorAccount>(null)
+
+  const changeToAccountId = (event: ChangeEvent<HTMLInputElement>) => {
+    setToAccountId(event.target.value)
+  }
+
+  const submit = async () => {
+    if (toAccountId === '') {
+      return
+    }
+
+    try {
+      const resp = await getOperatorAccount({accountId: toAccountId})
+      if (resp.status !== 200) {
+        return
+      }
+      setToAccount(resp.data)
+      setMigrateOverlayVisible(true)
+    } catch (error) {
+      console.error(error)
+    }
+  }
+
+  const canSubmit = toAccountId !== ''
+
+  return organization?.state !== 'provisioned' ? (
+    <></>
+  ) : (
+    <>
+      <MigrateOrgOverlay toAccount={toAccount} />
+      <h4 data-testid="migrate-org--title">Migrate Org to another Account</h4>
+      <Form onSubmit={submit}>
+        <FlexBox
+          direction={FlexDirection.Row}
+          alignItems={AlignItems.FlexStart}
+          margin={ComponentSize.Large}
+        >
+          <Input
+            placeholder="Account ID to migrate resources to"
+            inputStyle={{width: '400px'}}
+            style={{width: 'auto'}}
+            value={toAccountId}
+            onChange={changeToAccountId}
+            testID="accounts-migrate--account-id"
+          />
+          <Button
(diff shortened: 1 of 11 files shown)


src/operator/account/MigrateOrgsOverlay.tsx
```diff
@@ -0,0 +1,105 @@
...
+    account,
+    organizations,
+    migrateStatus,
+    handleMigrateOrgs,
+    setMigrateOverlayVisible,
+    migrateOverlayVisible,
+  } = useContext(AccountContext)
+
+  const migrateAccountOrgs = () => {
+    if (account?.type !== 'cancelled') {
```
wdoconnell, line 31: Is this just an extra guard in case there's a race condition, like the account being canceled while the overlay is being opened - since we only render an empty fragment instead of the `MigrateOrgsOverlay` if the account is canceled?

Since this `try / catch` block is only hit if the account type isn't canceled, might still want a notification here so that support doesn't presume the migration works if they hit the migration button and nothing happens.

src/operator/context/account.tsx
```diff
@@ -168,17 +180,43 @@ export const AccountProvider: FC<Props> = React.memo(({children}) => {
...
+          fromAccountId: accountID,
+          toAccountId: toAccountId,
+        })
+
+        if (resp.status !== 200) {
+          throw new Error(resp.data.message)
+        }
+
+        setMigrateOverlayVisible(false)
+        setMigrateStatus(RemoteDataState.Done)
```
wdoconnell, line 197: I might reverse the order of lns `196` and `197` here, since `migrateStatus` is used in the migration overlay. By the time `migrateStatus` is `Done` I suspect the overlay will be removed from the DOM.

src/operator/MigrateOrgOverlay.tsx
```diff
@@ -0,0 +1,92 @@
...
+          >
+            I understand, migrate the organization
+          </ButtonBase>
+        </Overlay.Footer>
+      </Overlay.Container>
+    </Overlay>
+  )
+}
+
+export default MigrateOrgOverlay
```
wdoconnell, line 92: Same comment here re: default exports. Let's make this a named import so that it's easier to trace and tree shake.

src/operator/account/MigrateOrgs.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const [toAccountId, setToAccountId] = useState('')
+  const [toAccount, setToAccount] = useState<OperatorAccount>(null)
+
+  const changeToAccountId = (event: ChangeEvent<HTMLInputElement>) => {
+    setToAccountId(event.target.value)
+  }
+
+  const submit = async () => {
+    if (toAccountId === '') {
+      return
```
wdoconnell, line 29: Do we need a notification here - like "please enter an account to migrate to"?

src/operator/account/MigrateOrgs.tsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    try {
+      const resp = await getOperatorAccount({accountId: toAccountId})
+      if (resp.status !== 200) {
+        return
```
wdoconnell, line 35: Same - could we add a notification letting the user know there was an error retrieving the `ToAccount`?

wdoconnell (commented):
Looks great - couple minor comments here, mainly around adding some notifications for error states for when support is using the migration overlay.

src/operator/MigrateOrg.tsx
```diff
@@ -0,0 +1,83 @@
...
+            status={
+              canSubmit ? ComponentStatus.Default : ComponentStatus.Disabled
+            }
+          />
+        </FlexBox>
+      </Form>
+    </>
+  )
+}
+
```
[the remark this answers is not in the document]
abshierjoel: Absolutely!

src/operator/MigrateOrgOverlay.tsx
```diff
@@ -0,0 +1,92 @@
...
+} from '@influxdata/clockface'
+import {OperatorAccount} from 'src/client/unityRoutes'
+import {OverlayContext} from './context/overlay'
+
+interface Props {
+  toAccount: OperatorAccount
+}
+
+const MigrateOrgOverlay: FC<Props> = ({toAccount}) => {
+  const {
```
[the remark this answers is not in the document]
abshierjoel: 🔤 🔤 🔤

src/operator/context/account.tsx
abshierjoel: That's a good point. Changing it.

src/operator/account/MigrateOrgs.tsx
abshierjoel: Yeah, that's a good point. This is a pretty blind failure. It seems like I can pull use the `useDispatch()` hook with `notify` to put up an error. Does that make sense?

Also, remocal is failing me... so I'm pushing up these changes to run in CI 😅 🤞

wdoconnell (dismissed):
Looks good - thanks for the commits to cover my comments 😄

wdoconnell (approved):

PR #6750, opened 2023-06-29
chore: update vwo smartcode

mavarius:
Addresses https://github.com/influxdata/idpe/issues/16543

This PR updates the version of the VWO smartcode to v1.5

### Checklist

### DIFF
--- a/src/utils/vwo.ts
+++ b/src/utils/vwo.ts
@@ -8,14 +8,17 @@ export const executeVWO = () => {
     window._vwo_code ||
     (function () {
       var account_id = 634566,
+        version = 1.5,
         settings_tolerance = 2000,
         library_tolerance = 2500,
         use_existing_jquery = false,
         is_spa = 1,
         hide_element = '',
+        hide_element_style = '',
         /* DO NOT EDIT BELOW THIS LINE */
         f = false,
         d = document,
+        vwoCodeEl = d.querySelector('#vwoCode'),
         code = {
           use_existing_jquery: function () {
             return use_existing_jquery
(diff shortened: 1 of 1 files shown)


wdoconnell:
@mavarius Is this needed to take advantage of new VWO features?

mavarius:
> @mavarius Is this needed to take advantage of new VWO features?

No, this is to address the security vulnerability outlined in https://github.com/influxdata/idpe/issues/16543.

wdoconnell:
@mavarius Thanks - has VWO updated the code on their end to ensure it isn't using the deprecated jquery version?

wdoconnell:
The critical thing here is ensuring that our patch to resolve the "black screen of death" issue still works - namely that in `src/App.tsx`, we are still removing the correct style as a fallback in the event that the VWO script fails like it did a couple months ago. It looks like to be same style name (`_vis_opt_path_hides`), which is good. I don't love having to use third-party code like this that is already difficult to understand, but I get why we need to given the IDPE issue.

If we run into issues with the upgraded script, the first line of defense is to flip the flag `vwoAbTesting` off in ConfigCat to troubleshoot any issues with the scritp.

mavarius:
@wdoconnell I noticed that these two rules were removed in the previous implementation, so I removed them here again. Maybe they were part of what caused the black screen issue last time.

```
hide_element='body',
hide_element_style = 'opacity:0 !important;filter:alpha(opacity=0) !important;background:none !important',
```

wdoconnell (approved):

PR #6769, opened 2023-07-19
Variable name fix in python tutorial

spicydonkey:
This PR replaces `write_client` to `client`, because subsequent tutorials refer to the object as such.

The single change in varname is in the "code snippet" of the python tutorial InitializeClient at `src/homepageExperience/components/steps/python/InitializeClient.tsx`.

This is consistent with all code snippets in the subsequent examples.

### DIFF
--- a/src/homepageExperience/components/steps/python/InitializeClient.tsx
+++ b/src/homepageExperience/components/steps/python/InitializeClient.tsx
@@ -25,7 +25,7 @@ token = os.environ.get("INFLUXDB_TOKEN")
 org = "${org.name}"
 url = "${url}"
 
-write_client = influxdb_client.InfluxDBClient(url=url, token=token, org=org)
+client = influxdb_client.InfluxDBClient(url=url, token=token, org=org)
 `
 
   return (


appletreeisyellow (commented):
@spicydonkey Thanks for contributing. 

`write_client` is a local variable name that is used to initialize the write client. This variable is used in the next step of the tutorial (see code below), so replacing it to `client` would break the set up flow. 

https://github.com/influxdata/ui/blob/73f0cde8ff90726f74b01a5680c95b06e18c37f2/src/homepageExperience/components/steps/python/WriteDataSql.tsx#L55 

I think it's better to keep the variable as `write_client` to differentiate it from the `query_client` that is used in the later steps of the tutorial (see code below):

https://github.com/influxdata/ui/blob/73f0cde8ff90726f74b01a5680c95b06e18c37f2/src/homepageExperience/components/steps/python/ExecuteQuerySql.tsx#L35-L43

JakovGlavac:
I was just about to open this issue.

PR #6782, opened 2023-08-15
feat(sql): add `/api/v2private/query` for SQL query

appletreeisyellow:
Closes influxdata/idpe#17953

This PR is the start of the UI work for [this epic issue](https://github.com/influxdata/idpe/issues/17711). 

This PR adds the path `/api/v2private/query` for SQL queries. When the feature flag is on and the script language is SQL, the query will send to this path instead of being wrapped in `iox.sql()` and sending to `/api/v2/query` in Flux.

To test locally, deploy the remocal with the following flag:
```
```

Then you will see SQL queries will send to `/api/v2private/query` in DevTool Network tab

### Checklist

### DIFF
--- a/src/shared/contexts/query/context.tsx
+++ b/src/shared/contexts/query/context.tsx
@@ -260,7 +260,7 @@ export const QueryProvider: FC = ({children}) => {
     const promise = fetch(url, {
       method: 'POST',
       headers,
-      body: JSON.stringify(body),
+      body: typeof body === 'string' ? body : JSON.stringify(body),
       signal: controller.signal,
     })
       .then((res: Response): Promise<RunQueryResult> => {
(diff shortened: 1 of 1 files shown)


src/shared/contexts/query/context.tsx
```diff
@@ -222,7 +234,7 @@ const handleQueryResponse = (response, cb) => {
     } catch {
       return {
         type: 'UNKNOWN_ERROR',
-        message: 'Failed to execute Flux query',
+        message: 'Failed to execute query',
```
appletreeisyellow, line 237: Make the error message more generic so that it can apply to Flux, SQL, InfluxQL

src/shared/contexts/query/context.tsx
```diff
@@ -248,7 +260,7 @@ export const QueryProvider: FC = ({children}) => {
     const promise = fetch(url, {
       method: 'POST',
       headers,
-      body: JSON.stringify(body),
+      body: typeof body === 'string' ? body : JSON.stringify(body),
```
appletreeisyellow, line 263: When the body is already a string, no need to call `JSON.stringify()` again. Otherwise, `/api/v2private/query` will error

wdoconnell (approved):
Looks good. 👍

PR #6790, opened 2023-08-22
chore(epic-idpe-17711): use SQL HTTP API to get field list

appletreeisyellow:
Wait for https://github.com/influxdata/ui/pull/6783 to merge

Part of https://github.com/influxdata/idpe/issues/17981

This PR uses the SQL HTTP API to query fields

# To test locally

    ```
    ```

# Before and After

https://github.com/influxdata/ui/assets/14298407/a7e67caa-8ffa-4959-9af5-026d6b885c30

### Checklist


src/dataExplorer/context/fields.tsx
```diff
@@ -68,6 +77,37 @@ export const FieldsProvider: FC<Prop> = ({children, scope}) => {
...
+           AND column_name ILIKE '%${sanitized}%'
+      LIMIT ${DEFAULT_LIMIT}
+      `
+      try {
+        const resp = await queryAPI(queryTextSQL, scope, {
+          language: LanguageType.SQL,
+          bucket,
+        } as QueryOptions)
+        const values = (resp.parsed.table?.columns?.column_name?.data ??
+          []) as string[]
```
wdoconnell, line 100: Is the reason we don't get strong typing here (and have to type cast using `as`) that we can't predict the type of `resp.parsed.table.columns.column_name.data`?

appletreeisyellow: @wdoconnell We actually can predict the type of `resp.parsed.table.columns.column_name.data` which is `string[] | number[] | boolean[]`. Since field keys are strings, so we cast the data to `string[]` specifically

PR #6791, opened 2023-08-22
chore(epic-idpe-17711): use SQL HTTP API to get tag keys

appletreeisyellow:
Part of https://github.com/influxdata/idpe/issues/17981

This PR uses the SQL HTTP API to query tag keys

# To test locally

    ```
    ```

# Before and After

https://github.com/influxdata/ui/assets/14298407/ab1b4930-3049-4f05-9c42-70bde7b98a03

### Checklist


src/dataExplorer/context/tags.tsx
```diff
@@ -82,6 +91,53 @@ export const TagsProvider: FC<Prop> = ({children, scope}) => {
...
+        LIMIT ${DEFAULT_LIMIT}
+      `
+      const newTags: Tags = {}
+      try {
+        const resp = await queryAPI(queryTextSQL, scope, {
+          language: LanguageType.SQL,
+          bucket,
+        } as QueryOptions)
+        const keys = (resp.parsed.table.columns?.column_name?.data ??
+          []) as string[]
```
wdoconnell, line 114: Some comment here as in other PRs re: whether the type casting is avoidable.

src/dataExplorer/context/tags.tsx
```diff
@@ -82,6 +91,53 @@ export const TagsProvider: FC<Prop> = ({children, scope}) => {
...
+        keys.map(key => {
+          newTags[key] = []
+        })
+
+        // Initialize status for each key
+        const tagValueStatuses = {} as Hash<RemoteDataState>
+        keys.map(key => {
+          tagValueStatuses[key] = RemoteDataState.NotStarted
+        })
+
```
wdoconnell, line 126: Could we avoid a second map but initializing both the tags and statuses in a single iteration?

const tagValueStatuses = {}

`keys.map(key => {
   newTags[key] = []
   tagValueStatuses[key] = RemoteDataState.NotStarted
})

src/dataExplorer/context/tags.tsx
```diff
@@ -158,6 +214,41 @@ export const TagsProvider: FC<Prop> = ({children, scope}) => {
...
+        LIMIT ${DEFAULT_LIMIT}
+      `
+      try {
+        const resp = await queryAPI(queryTextSQL, scope, {
+          language: LanguageType.SQL, // use SQL to get measurement list
+          bucket,
+        } as QueryOptions)
+        const values = (Object.values(resp.parsed.table?.columns ?? [])[0]
+          ?.data ?? []) as string[]
+
```
wdoconnell, line 231: Same comment here re: the type cast.

wdoconnell (approved):
Looks good - small comment re: we might be able to avoid one of the `map`s here.

src/dataExplorer/context/tags.tsx
appletreeisyellow: Thanks for mentioning it!`resp.parsed.table.columns?.column_name?.data` has type in ` number[] | boolean[] | string[]`. The response data is tag keys and tag keys are strings, so casting the data to strings specifically here. For the tag values, it makes sense to not casting into `string[]` since tag values can be `number` or `boolean`

src/dataExplorer/context/tags.tsx
appletreeisyellow: Of course yes! Updated in https://github.com/influxdata/ui/pull/6791/commits/26ec5315ae39fc461447cea6b182a42953fef6c0

src/dataExplorer/context/tags.tsx
appletreeisyellow: Good point! It makes sense to avoid casting the tag values since the original data type is `string[] | boolean[] | number[]`. Updated in https://github.com/influxdata/ui/pull/6791/commits/689f58c8c69ec61a85ac543f5a8a897cad69f3ee

appletreeisyellow:
@wdoconnell Thanks for the good questions! I updated the code

wdoconnell (dismissed):
Looks good - since all of this is flagged, is there a story or other ticket around additional testing that needs to be done before this is released generally?

PR #6792, opened 2023-08-25
chore(epic-idpe-17711): update csv download to path `/api/v2private/query`

appletreeisyellow:
Closes https://github.com/influxdata/idpe/issues/18001

This PR routes the CSV download for SQL through the HTTP path `/api/v2private/query`. In addition, a service worker for scope `/api/v2private/query` is also added to make the feature complete.

The backend work (https://github.com/influxdata/idpe/pull/18054) is already merged

# To test locally

    ```
    ```

# After

Customers won't notice any difference before and after the change

https://github.com/influxdata/ui/assets/14298407/fc01d6fc-5c63-4702-8f07-c2a694a2388c

### Checklist

### DIFF
--- a/src/dataExplorer/components/ResultsPane.tsx
+++ b/src/dataExplorer/components/ResultsPane.tsx
@@ -18,7 +18,7 @@ import {useSelector, useDispatch} from 'react-redux'
 
 // Contexts
 import {ResultsContext} from 'src/dataExplorer/context/results'
-import {QueryContext, QueryOptions} from 'src/shared/contexts/query'
+import {QueryContext} from 'src/shared/contexts/query'
 import {
   PersistenceContext,
   DEFAULT_FLUX_EDITOR_TEXT,
--- a/src/dataExplorer/components/ResultsPane.tsx
+++ b/src/dataExplorer/components/ResultsPane.tsx
@@ -133,10 +133,14 @@ const ResultsPane: FC = () => {
             // the `\n` character is interpreted by the browser as a line break
             // so replacing the `\n` char with empty space to keep query in correct syntax
             inputValue = text.replace(/\n/g, ' ')
+            // Intentionally keep the `break` inside the if statement;
+            // Otherwide, users cannot download csv for SQL queries when
+            // the feature flag is not enabled
             break
           }
-        // if the flag is not enabled, fall back
-        // to use the following Flux method
+        // If the flag is not enabled, each SQL query actually will be wrapped
+        // inside a Flux query `iox.sql()`, and the if statement will be false,
+        // so we want the logic to fall to the next case LanguageType.FLUX
         case LanguageType.FLUX:
           url = `${API_BASE_PATH}api/v2/query?${new URLSearchParams({orgID})}`
           const query =
--- a/src/dataExplorer/components/ResultsPane.tsx
+++ b/src/dataExplorer/components/ResultsPane.tsx
@@ -18,7 +18,7 @@ import {useSelector, useDispatch} from 'react-redux'
 
 // Contexts
 import {ResultsContext} from 'src/dataExplorer/context/results'
-import {QueryContext} from 'src/shared/contexts/query'
+import {QueryContext, QueryOptions} from 'src/shared/contexts/query'
 import {
   PersistenceContext,
   DEFAULT_FLUX_EDITOR_TEXT,
@@ -50,6 +50,7 @@ import {
   updateWindowPeriod,
 } from 'src/shared/contexts/query/preprocessing'
 import {rangeToParam} from 'src/dataExplorer/shared/utils'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Constants
 import {TIME_RANGE_START, TIME_RANGE_STOP} from 'src/variables/constants'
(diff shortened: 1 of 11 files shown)


src/shared/components/CSVExportButton.tsx
```diff
@@ -59,7 +64,27 @@ class CSVExportButton extends PureComponent<Props, State> {
...
+          console.error(
+            'Feature not available, because ServiceWorker registration failed: ',
+            err
+          )
+        }
+      )
+    }
+
+    if (isFlagEnabled('v2privateQueryUI') && props.workerRegistrationSQL) {
+      ;(props.workerRegistrationSQL as Promise<ServiceWorkerRegistration>).then(
```
wdoconnell, line 84: I'm not very familiar with the ;() syntax here. Would you be able to explain what it's doing?

src/shared/contexts/app.tsx
```diff
@@ -159,6 +169,10 @@ export const AppSettingProvider: FC = ({children}) => {
     dispatch(setWorkerRegistrationInfluxQL(workerRegistrationInfluxQL))
   }, [workerRegistrationInfluxQL])
 
+  useEffect(() => {
+    dispatch(setWorkerRegistrationSQL(workerRegistrationSQL))
+  }, [workerRegistrationSQL])
+
```
wdoconnell, line 175: Does this need to be registered multiple times, or is once on app load if sufficient? If the latter, it might make more sense to add an `if remoteDataState != RemoteDataState.Done {}` style check to ensure this useEffect hook is only run once.

wdoconnell (commented):
Looks good, but had two questions on this one.

src/shared/components/CSVExportButton.tsx
appletreeisyellow: I also had this question before! It's automatically added by prettier when I save the file. It's a wired case and here is an explanation: https://github.com/influxdata/ui/pull/5270#discussion_r936974286

src/shared/contexts/app.tsx
appletreeisyellow: Good point! Like you mentioned in the latter point, this service worker is registered once before the app load. I like the idea of ensuring the code to run `dispatch(setWorkerRegistrationSQL(workerRegistrationSQL))` once 😃

Since this context provider component runs before the page load, there is no data yet, so we cannot check on `remoteDataState` at this moment. 

One thing I can think of is to register the service worker when the component is just mounted, which only runs once. Something like this:

```js
useEffect(() => {
    dispatch(setWorkerRegistration(workerRegistration))
    dispatch(setWorkerRegistrationInfluxQL(workerRegistrationInfluxQL))
    dispatch(setWorkerRegistrationSQL(workerRegistrationSQL))
  }, [])
```

```js
// useEffect hook with the empty array is equivalent to componentDidMount() and only run once
useEffect(() => {
  ...
}, [])
```
Where `workerRegistration` and `workerRegistrationInfluxQL` are service workers for Flux and InfluxQL. Since each service worker can only serve on HTTP path, that's why we need three different service workers to support `/api/v2private/query` (SQL), `/query` (InfluxQL), and `/api/v2/query` (Flux) separately

What do you think?

wdoconnell: I don't usually love this approach:

```
useEffect(() => {
    dispatch(setWorkerRegistration(workerRegistration))
    dispatch(setWorkerRegistrationInfluxQL(workerRegistrationInfluxQL))
    dispatch(setWorkerRegistrationSQL(workerRegistrationSQL))
  }, [])
```

for two reasons. For one, it's somewhat anti-pattern in React because the linter will probably warn you that the variables (like dispatch or setWorkerRegistration...) need to be included in the dependency array. The way you did it previously is completely right IMHO, as long as there's not a risk that the equality check fails for `workerRegistrationSQL` and causes extra re-renders. 

I think a way to test this would be -  keep the useEffect hook like this, but test it (for example, temporarily add a `console.log`) to make sure the hook isn't being called multiple times on app load. If it isn't, I actually wouldn't change anything.

```
useEffect(() => {
    dispatch(setWorkerRegistrationSQL(workerRegistrationSQL))
  }, [workerRegistrationSQL])
```

src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -126,8 +127,17 @@ const ResultsPane: FC = () => {
...
 
       switch (language) {
-        case LanguageType.FLUX:
         case LanguageType.SQL:
+          if (isFlagEnabled('v2privateQueryUI')) {
+            url = `${API_BASE_PATH}api/v2private/query?database=${selection.bucket?.name}`
+            // the `\n` character is interpreted by the browser as a line break
+            // so replacing the `\n` char with empty space to keep query in correct syntax
+            inputValue = text.replace(/\n/g, ' ')
+            break
```
wdoconnell, line 136: I think this `break` needs to be outside the `if` block. If the language type is SQL (the `case`) but the underlying `if` block with the feature flag isn't satisfied, javascript will miss the `break` and continue to the remaining case statements (and most importantly, to the `default` block that throws an error) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#breaking_and_fall-through

wdoconnell (changes requested):
Looks good, but I would change the `case`/`if` block to avoid a situation where an error is thrown when the feature flag is turned off and we are using `LanguageType.SQL`. Also, based on your recommendation I think the original approach to this `useEffect` hook should be right, as long as you're able to validate it's not causing re-renders. Thanks!

src/shared/contexts/app.tsx
appletreeisyellow: I tested using a `console.log()` in the useEffect hook and confirmed that the useEffect hook only run once:

```
useEffect(() => { 
  dispatch(setWorkerRegistrationSQL(workerRegistrationSQL))
}, [workerRegistrationSQL])
```

Then I'll keep the original code. Thanks for the brainstorming 😄🧠

src/dataExplorer/components/ResultsPane.tsx
appletreeisyellow: The break is intentionally put inside the `if` statement. When the feature flag is not satisfied, then SQL falls to the next case which is the Flux case. 

When the feature flag `v2privateQueryUI` is off,  each SQL query actually is wrapped inside a Flux query `iox.sql()`. 

Sorry it is kind of confusing 😅 I'll add a comment to make it more clear

appletreeisyellow: Updated comments in https://github.com/influxdata/ui/pull/6792/commits/b384bf99036faa6dde6290706e0b43170e59b527

wdoconnell: Got it. Thanks for explaining the logic and updating this comment 👍

wdoconnell (approved):

PR #6808, opened 2023-09-25
chore: remove feature flag `v2privateQueryUI`

appletreeisyellow:
Part of https://github.com/influxdata/idpe/issues/18016

`v2privateQueryUI` has been on in production for two weeks and it's working without issues. This PR removes the feature flag `v2privateQueryUI` in UI.

### Checklist

### DIFF
--- a/src/dataExplorer/components/ResultsPane.tsx
+++ b/src/dataExplorer/components/ResultsPane.tsx
@@ -45,12 +45,8 @@ import {event} from 'src/cloud/utils/reporting'
 import {notify} from 'src/shared/actions/notifications'
 import {getWindowPeriodVariableFromVariables} from 'src/variables/utils/getWindowVars'
 import {csvDownloadFailure} from 'src/shared/copy/notifications'
-import {
-  sqlAsFlux,
-  updateWindowPeriod,
-} from 'src/shared/contexts/query/preprocessing'
+import {updateWindowPeriod} from 'src/shared/contexts/query/preprocessing'
 import {rangeToParam} from 'src/dataExplorer/shared/utils'
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Constants
 import {TIME_RANGE_START, TIME_RANGE_STOP} from 'src/variables/constants'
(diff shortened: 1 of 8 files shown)


src/dataExplorer/components/ResultsPane.tsx
```diff
@@ -128,25 +124,14 @@ const ResultsPane: FC = () => {
...
-        // so we want the logic to fall to the next case LanguageType.FLUX
+          url = `${API_BASE_PATH}api/v2private/query?database=${selection.bucket?.name}`
+          // the `\n` character is interpreted by the browser as a line break
+          // so replacing the `\n` char with empty space to keep query in correct syntax
+          inputValue = text.replace(/\n/g, ' ')
+          break
         case LanguageType.FLUX:
           url = `${API_BASE_PATH}api/v2/query?${new URLSearchParams({orgID})}`
-          const query =
-            language == LanguageType.SQL
```
wdoconnell, line 147: Will this cause a different behavior when the flag is removed, since we are no longer checking for whether language is `LanguageType.SQL`? Or was `language == LanguageType.SQL` only when the feature flag was present?

wdoconnell (commented):
Looks good, but first I had one question about the removal of the `sqlAsFlux` function, and whether that is still needed at all.

src/dataExplorer/components/ResultsPane.tsx
appletreeisyellow: Good question! Since we are removing the feature flag, so `language == LanguageType.SQL` will always be false under the switch statement `case LanguageType.FLUX`, because there is a `break` ([code](https://github.com/influxdata/ui/pull/6808/files#diff-e221caeec2ec90c4cb59ba36d3dda5fd8990b74997169cf2c79c0172c8571cacR131)) under `case LanguageType.SQL` which separate SQL from Flux. Then there is no need to have `language == LanguageType.SQL` and `sqlAsFlux(text, selection.bucket)`

appletreeisyellow:
@wdoconnell 

`sqlAsFlux` is still used in one place -- for graphs (see details [here](https://github.com/influxdata/idpe/issues/17986#issuecomment-1703346936) and [here](https://github.com/influxdata/idpe/issues/17986#issue-1837337481)), so that's why I didn't remove `sqlAsFlux`

After the removal of the feature flag `v2privateQueryUI `, query code is straight forward:
* when users select SQL language, query will be sent to `/api/v2private/query`
* when users select Flux language, query will be sent to `/api/v2/query`

(except for graphs, regardless what language a user select, there is always one extra API call to `/api/v2/query` to get Flux data format in order to draw graphs)

wdoconnell (approved):

Issue #6800, opened 2023-09-05
Remove feature flag `influxqlUI`


appletreeisyellow:
Now the feature flag has been on and in production without issue for two weeks, we can remove the flag in the following places

```[tasklist]
- [x] UI https://github.com/influxdata/ui/pull/6809
- [x] idpe https://github.com/influxdata/idpe/pull/18199
- [x] ConfigCat
```

PR #6809, opened 2023-09-25
chore: remove feature flag `influxqlUI`

appletreeisyellow:
Part of https://github.com/influxdata/ui/issues/6800

The feature flag `influxqlUI` has been on in production for two weeks without any issues, so removing it in this PR

### Checklist

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -820,6 +820,12 @@ const setScriptToLanguage = (
   defaultEditorText: string
 ) => {
   return cy.isIoxOrg().then(isIox => {
+    if (lang === 'influxql') {
+      // give cypress some time to set up
+      // database and retetion policy (DBRP) mappings
+      cy.wait(1000)
+    }
+
     // influxql works on both IOx and TSM
     if (isIox || lang === 'influxql') {
       cy.getByTestID('script-query-builder--new-script')
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -820,12 +820,6 @@ const setScriptToLanguage = (
   defaultEditorText: string
 ) => {
   return cy.isIoxOrg().then(isIox => {
-    if (lang === 'influxql') {
-      // give cypress some time to turn on the feature flag `influxqlUI`
-      // this block can be removed after this feature flag is removed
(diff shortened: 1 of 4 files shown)


cypress/support/commands.ts
```diff
@@ -821,8 +821,8 @@ const setScriptToLanguage = (
 ) => {
   return cy.isIoxOrg().then(isIox => {
     if (lang === 'influxql') {
-      // give cypress some time to turn on the feature flag `influxqlUI`
-      // this block can be removed after this feature flag is removed
+      // give cypress some time to set up
+      // database and retetion policy (DBRP) mappings
```
wdoconnell, line 825: minor nit - retention policy

src/dataExplorer/components/SaveAsScript.tsx
```diff
@@ -111,7 +112,10 @@ const SaveAsScript: FC<Props> = ({language, onClose, setOverlayType, type}) => {
     setResult(null)
     clearViewOptions()
 
-    if (isIoxOrg || isFlagEnabled('influxqlUI')) {
+    // InfluxQL works on both IOx and TSM, so if an account has
```
wdoconnell, line 115: This will also work in OSS, correct? The only different thing I could think of doing differently here is if we need a `CLOUD` check.

wdoconnell (dismissed):

cypress/support/commands.ts
appletreeisyellow: Nice catch! Fixed in https://github.com/influxdata/ui/pull/6809/commits/dbd487c18a083ae691c2a33d99ed0e693d667207

wdoconnell (approved):

src/dataExplorer/components/SaveAsScript.tsx
appletreeisyellow: InfluxQL will not show up in OSS because of [this line](https://github.com/influxdata/ui/pull/6809/files#diff-705df60af76f336059d091ad2442f9d08f2f28ce2ffa6a5c8b947bd84ceef4b0R199), so OSS users will never need to save an InfluxQL script

PR #6821, opened 2023-10-18
feat: Add click through announcement feature

mavarius:
This PR adds a new feature to Serverless, the click-through announcement. It is used for important announcements that users must see and includes event tracking for interacting with the announcement component.

The first announcement is for the PAYG pricing increase. There is also a banner on the billing page regarding this announcement. Both of these are limited to direct signup PAYG users.

image

image

### Checklist

### DIFF
--- a/src/billing/components/PayAsYouGo/PayAsYouGo.tsx
+++ b/src/billing/components/PayAsYouGo/PayAsYouGo.tsx
@@ -1,4 +1,8 @@
+// Libraries
 import React, {FC} from 'react'
+import {useSelector} from 'react-redux'
+
+// Components
 import {
   AlignItems,
   ComponentSize,
@@ -7,8 +11,6 @@ import {
   Panel,
   ResourceList,
 } from '@influxdata/clockface'
-
-// Components
 import PlanTypePanel from 'src/billing/components/PayAsYouGo/PlanTypePanel'
 import PaymentPanel from 'src/billing/components/PaymentInfo/PaymentPanel'
 import BillingContactInfo from 'src/billing/components/BillingContactInfo'
@@ -17,35 +19,45 @@ import CancellationPanel from 'src/billing/components/PayAsYouGo/CancellationPan
 import NotificationPanel from 'src/billing/components/PayAsYouGo/NotificationPanel'
 import InvoiceLoadingWrapper from 'src/billing/components/AssetLoading/InvoiceWrapper'
 import BillingInfoWrapper from 'src/billing/components/AssetLoading/BillingInfoWrapper'
+import {PricingAlert} from 'src/billing/components/PayAsYouGo/PricingAlert'
 
-const BillingPayAsYouGo: FC = () => (
-  <FlexBox
-    direction={FlexDirection.Column}
-    alignItems={AlignItems.Stretch}
-    margin={ComponentSize.Small}
-  >
-    <BillingInfoWrapper>
-      <>
-        <PlanTypePanel />
-        <Panel>
-          <Panel.Header testID="past-invoices--header">
-            <h4>Past Invoices</h4>
-          </Panel.Header>
-          <Panel.Body>
-            <ResourceList>
-              <InvoiceLoadingWrapper>
-                <InvoiceHistory />
-              </InvoiceLoadingWrapper>
-            </ResourceList>
-          </Panel.Body>
-        </Panel>
-        <PaymentPanel />
-        <BillingContactInfo />
-      </>
-    </BillingInfoWrapper>
-    <NotificationPanel />
-    <CancellationPanel />
-  </FlexBox>
-)
+// Utils
+import {selectCurrentIdentity} from 'src/identity/selectors'
+
+const BillingPayAsYouGo: FC = () => {
+  const {account} = useSelector(selectCurrentIdentity)
+  const directSignup = account.billingProvider === 'zuora'
+
+  return (
+    <FlexBox
+      direction={FlexDirection.Column}
+      alignItems={AlignItems.Stretch}
+      margin={ComponentSize.Small}
+    >
+      <BillingInfoWrapper>
+        <>
+          {directSignup && <PricingAlert />}
+          <PlanTypePanel />
+          <Panel>
+            <Panel.Header testID="past-invoices--header">
+              <h4>Past Invoices</h4>
+            </Panel.Header>
+            <Panel.Body>
+              <ResourceList>
+                <InvoiceLoadingWrapper>
+                  <InvoiceHistory />
+                </InvoiceLoadingWrapper>
+              </ResourceList>
+            </Panel.Body>
+          </Panel>
+          <PaymentPanel />
+          <BillingContactInfo />
+        </>
+      </BillingInfoWrapper>
+      <NotificationPanel />
+      <CancellationPanel />
+    </FlexBox>
+  )
+}
 
 export default BillingPayAsYouGo
--- a/src/billing/components/PayAsYouGo/PayAsYouGo.tsx
+++ b/src/billing/components/PayAsYouGo/PayAsYouGo.tsx
@@ -26,7 +26,7 @@ import {selectCurrentIdentity} from 'src/identity/selectors'
 
 const BillingPayAsYouGo: FC = () => {
   const {account} = useSelector(selectCurrentIdentity)
-  const directSignup = account.billingProvider === 'zuora'
+  const isdDirectSignup = account.billingProvider === 'zuora'
 
   return (
     <FlexBox
@@ -36,7 +36,7 @@ const BillingPayAsYouGo: FC = () => {
     >
       <BillingInfoWrapper>
         <>
-          {directSignup && <PricingAlert />}
+          {isdDirectSignup && <PricingAlert />}
           <PlanTypePanel />
           <Panel>
             <Panel.Header testID="past-invoices--header">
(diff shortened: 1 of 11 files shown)


src/billing/components/PayAsYouGo/PayAsYouGo.tsx
```diff
@@ -17,35 +19,45 @@ import CancellationPanel from 'src/billing/components/PayAsYouGo/CancellationPan
...
 import InvoiceLoadingWrapper from 'src/billing/components/AssetLoading/InvoiceWrapper'
 import BillingInfoWrapper from 'src/billing/components/AssetLoading/BillingInfoWrapper'
+import {PricingAlert} from 'src/billing/components/PayAsYouGo/PricingAlert'
+
+// Utils
+import {selectCurrentIdentity} from 'src/identity/selectors'
+
+const BillingPayAsYouGo: FC = () => {
+  const {account} = useSelector(selectCurrentIdentity)
+  const directSignup = account.billingProvider === 'zuora'
```
appletreeisyellow, line 29:
```suggestion
  const isDirectSignup = account.billingProvider === 'zuora'
```

Just a suggestion on naming. Since the data type is boolean, adding an `is` makes it easier to read later in the code. Readers don't need to go up to where this variable is defined and then think about what type it is. Just a suggestion, not a blocker. The current code is also good 👍

src/homepageExperience/ClickThroughAnnouncementHandler.tsx
```diff
@@ -0,0 +1,67 @@
...
+    }))
+    dismissOverlay()
+  }
+
+  useEffect(() => {
+    // Current Announcement: PAYG Pricing Increase
+    // Audience: Cloud, Pay As You Go, Direct Signups
+    const paygAccount = account.type === 'pay_as_you_go'
+    const directSignup = account.billingProvider === 'zuora'
+    const audience = CLOUD && paygAccount && directSignup
```
appletreeisyellow, line 49:
```suggestion
    const isPaygAccount = account.type === 'pay_as_you_go'
    const isDirectSignup = account.billingProvider === 'zuora'
    const isAudience = CLOUD && isPaygAccount && isDirectSignup
```

Same reason as above. Just a suggestion, not a blocker

src/homepageExperience/ClickThroughAnnouncementHandler.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const dispatch = useDispatch()
+  const {account} = useSelector(selectCurrentIdentity)
+
+  const [announcementState, setAnnouncementState] = useLocalStorageState(
+    'clickThroughAnnouncement',
+    {
+      announcementID: '',
+      display: true,
+    }
+  )
```
appletreeisyellow, line 25: By using `useLocalStorageState`, users who already acknowledged the click through announcement will not see the announcement again. However, users will see the announcement again if they do the following:
* clear their browser cache
* use another browser
* use a private browser

Is it right?

appletreeisyellow (approved):
Looks good to me! I have one question and one non-blocking suggestion

src/homepageExperience/ClickThroughAnnouncementHandler.tsx
mavarius: Yes, this is correct and is acceptable behavior for these types of announcements.

src/billing/components/PayAsYouGo/PayAsYouGo.tsx
mavarius: changed

src/homepageExperience/ClickThroughAnnouncementHandler.tsx
mavarius: changed

src/billing/components/PayAsYouGo/PayAsYouGo.tsx
```diff
@@ -17,35 +19,45 @@ import CancellationPanel from 'src/billing/components/PayAsYouGo/CancellationPan
...
 import InvoiceLoadingWrapper from 'src/billing/components/AssetLoading/InvoiceWrapper'
 import BillingInfoWrapper from 'src/billing/components/AssetLoading/BillingInfoWrapper'
+import {PricingAlert} from 'src/billing/components/PayAsYouGo/PricingAlert'
+
+// Utils
+import {selectCurrentIdentity} from 'src/identity/selectors'
+
+const BillingPayAsYouGo: FC = () => {
+  const {account} = useSelector(selectCurrentIdentity)
+  const isdDirectSignup = account.billingProvider === 'zuora'
```
wdoconnell, line 29: minor, but should be `isDirectSignup`

wdoconnell, line 29: `billingProvider` is not fetched until accessing the billing page, so you may want to add an explicit check here accounting for if the property is absent

src/billing/components/PayAsYouGo/PayAsYouGo.tsx
```diff
@@ -17,35 +19,45 @@ import CancellationPanel from 'src/billing/components/PayAsYouGo/CancellationPan
...
-)
+  return (
+    <FlexBox
+      direction={FlexDirection.Column}
+      alignItems={AlignItems.Stretch}
+      margin={ComponentSize.Small}
+    >
+      <BillingInfoWrapper>
+        <>
+          {isdDirectSignup && <PricingAlert />}
```
wdoconnell, line 39: Same comment re: var name

src/billing/components/PayAsYouGo/PricingAlert.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <Alert icon={IconFont.AlertTriangle} color={ComponentColor.Primary}>
+      <FlexBox margin={ComponentSize.Medium}>
+        <FlexBox.Child grow={1} shrink={0}>
+          Starting on <b>December 1, 2023</b> there will be an increase in to
+          your usage-based pricing. Please feel free to{' '}
+          <SafeBlankLink
+            href={`mailto:payg-price-change@influxdata.com?
+            &subject=PAYG%20price%20change%20inquiry
+            &body=User%20ID:%20${user.email}%0D%0AOrg%20ID:%20${org.id}%0D%0A%0D%0APlease%20describe%20your%20inquiry%20here.`}
+            onClick={handleContactUsClick}
```
wdoconnell, line 47: This `href` is pretty hard to read - could we use [encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) so that the url is human-readable in the code?

src/homepageExperience/ClickThroughAnnouncementHandler.tsx
```diff
@@ -0,0 +1,67 @@
...
+    dismissOverlay()
+  }
+
+  useEffect(() => {
+    // Current Announcement: PAYG Pricing Increase
+    // Audience: Cloud, Pay As You Go, Direct Signups
+    const isPaygAccount = account.type === 'pay_as_you_go'
+    const isDirectSignup = account.billingProvider === 'zuora'
+    const isTargetAudience = CLOUD && isPaygAccount && isDirectSignup
+
```
wdoconnell, line 50: I think it may be best to return out of the component immediately if not `CLOUD`, as this is a homepage/firstmile component that can run in OSS. In OSS, the account.type and account.billingProvider properties don't exist, so ideally we would return before checking those to avoid potential errors.

src/me/components/ClickThroughAnnouncementOverlay.tsx
```diff
@@ -0,0 +1,101 @@
...
+          </ul>
+          <p>
+            This increase to your total monthly charges will depend on the
+            specific distribution of your workload across the billable vectors.
+            Most customers will experience an increase between 14% - 24%.
+          </p>
+          <p>
+            Please feel free to{' '}
+            <SafeBlankLink
+              href={`mailto:payg-price-change@influxdata.com?
```
wdoconnell, line 74: Same suggestion re: encoding the URI instead of including the %20s in the code

wdoconnell (commented):
Looks good, but had a few suggestions that may be worth considering to avoid potential conflicts between serverless and oss.

src/billing/components/PayAsYouGo/PayAsYouGo.tsx
mavarius: lol, good catch

src/billing/components/PayAsYouGo/PricingAlert.tsx
mavarius: Made human readable & encoded.

src/me/components/ClickThroughAnnouncementOverlay.tsx
mavarius: Made human readable & encoded.

src/homepageExperience/ClickThroughAnnouncementHandler.tsx
mavarius: moved & fetched

src/billing/components/PayAsYouGo/PricingAlert.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const PricingAlert: React.FC = () => {
+  useEffect(() => {
+    event(`pricingAnnouncementBanner.displayed`)
+  }, [])
+
+  const org = useSelector(selectCurrentOrg)
+  const user = useSelector(selectUser)
+
+  const encodedSubect = encodeURI('PAYG Pricing Increase')
```
appletreeisyellow, line 29:
```suggestion
  const encodedSubject = encodeURI('PAYG Pricing Increase')
```

appletreeisyellow (commented):
LGTM! Seems like the ci tests failure is a legit one. After you fix that, it should be good to go

appletreeisyellow (approved):

PR #6657, opened 2023-03-15
feat: make InfluxQL available to TSM orgs

appletreeisyellow:
Closes #6652

This PR allows TSM users to use InfluxQL in the script editor if they already have one or more DBRP mappings

All the work is behind the feature flag `influxqlUI`. 

To test on local remocal:
1) create DBRP mappings in your org ([instruction](https://github.com/influxdata/idpe/issues/17205)), and 
2) also make sure to turn on the following flags in your browser:
* `showOldDataExplorerInNewIOx`
* `newDataExplorer`
* `schemaComposition`

## Before

Screen Shot 2023-03-15 at 10 24 41 AM

## After

Screen Shot 2023-03-15 at 10 26 05 AM

https://user-images.githubusercontent.com/14298407/225366054-bd894b5c-a603-4307-991d-c8e26edaa58c.mov

### Checklist

Authors and Reviewer(s), please verify the following:

- [x] A PR description, regardless of the triviality of this change, that communicates the value of this PR
- [x] [Well-formatted conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) that provide context into the change
- [x] Documentation updated or issue created (provide link to issue/PR)
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [x] Feature flagged, if applicable

PR #6875, opened 2024-03-21
feat: add ioxQuery limits

mhilton:
Add support for viewing and modifying the new IOx Query limits.

Closes #

### Checklist


appletreeisyellow:
Is it possible to add a before and after screenshot in the description? Something like this https://github.com/influxdata/ui/pull/6657

src/operator/OrgOverlay.tsx
```diff
@@ -339,6 +339,28 @@ export const OrgOverlay: FC = () => {
                       </Grid.Column>
                     </Grid.Row>
                   )}
+                  {limits?.ioxQuery && (
```
appletreeisyellow, line 342: Seems like [ioxQuery](https://github.com/influxdata/openapi/blob/3380e8851d26ca33131a3e183d7147b42fdf2a09/contracts/priv/unity.yml#L2738) is defined under the type [OrgLimits](https://github.com/influxdata/openapi/blob/35ecbc2e2890e77756d0be0f801b9e503a8ffd44/contracts/priv/quartz-oem.yml#L471), but `limits` is a type of [OperatorOrgLimits](https://github.com/influxdata/openapi/blob/3380e8851d26ca33131a3e183d7147b42fdf2a09/contracts/priv/unity.yml#L1890)

wdoconnell: I think this is now resolved after https://github.com/influxdata/openapi/pull/649

wdoconnell:
> Is it possible to add a before and after screenshot in the description? Something like this #6657

@mhilton and I discussed this: Given the current implementation for mocking these limits, the addition of this limit on the operator page will not reflect meaningful data in local dev. I will add more details in the 🧵 , but staging is the better place to revalidate this.

wdoconnell (approved):

PR #6879, opened 2024-03-28
fix: downgrade rudderstack so debug is not pulled in

david-rusnak:
Closes https://github.com/influxdata/idpe/issues/18591

Downgrades rudderstack to not pull in debug `0.7.4`. Hopefully this version is ok, in the future we will probably remove rudderstck completely since it has been deprecated.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,7 +215,7 @@
     "redux-thunk": "^2.3.0",
     "remark-external-links": "8.0.0",
     "reselect": "^4.1.6",
-    "rudder-sdk-js": "2.48.2",
+    "rudder-sdk-js": "2.40.5",
     "s2-geometry": "^1.2.10",
     "use-local-storage-state": "^9.0.2",
     "uuid": "^9.0.0",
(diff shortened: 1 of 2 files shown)


wdoconnell (commented):
@david-rusnak It looks like they may have put in a hotfix in 2.46 - would you be able to check? https://github.com/rudderlabs/rudder-sdk-js/pull/1674

The PR links to a url including the line "error when upgrading rudder sdk js to the latest version"

If that still includes the `debug` package, let's stick with 2.40.5 as you propose here.

david-rusnak:
@wdoconnell Yeah just double checked, still brings in `0.7.4`.

wdoconnell (approved):
Thanks @david-rusnak. In that case let's use this version of the PR as is, since upgrading would bring in the problematic dependency.

Looks good from my perspective, but please wait for @jdstrand's approval as well. Thanks!

jdstrand (commented):
This looks ok to me in general, but I thought the rudderstack was updated recently to fix a different issue with CI tests? Something about it or a dependency that an earlier rudderstack used wasn't installable?

jdstrand:
> This looks ok to me in general, but I thought the rudderstack was updated recently to fix a different issue with CI tests? Something about it or a dependency that an earlier rudderstack used wasn't installable?

That said, this did pass tests, so I guess it's not a problem.

jdstrand (approved):

david-rusnak:
@jdstrand 

Yeah `2.14` had something that broke our CI, but then upgrading it to latest broke our linting because it added new types that weren't compatible - so I had to find a version that still worked for our version of typescript

PR #6889, opened 2024-05-15
fix(ui): remove mentions of TSM-specific limits for Serverless users

mavarius:
This PR removes the mention of TSM-specific limits for Serverless users.
image

### Checklist

### DIFF
--- a/src/billing/components/Free/OrgLimits.tsx
+++ b/src/billing/components/Free/OrgLimits.tsx
@@ -104,6 +104,15 @@ export const OrgLimits: FC = () => {
           )
         }
 
+        // The 'dashboards', 'tasks', 'checks', and 'rules' limits are not applicable to orgs using IOx.
+        if (orgUsesIOx) {
+          const hiddenLimits = ['dashboards', 'tasks', 'checks', 'rules']
+
+          if (hiddenLimits.includes(limitName)) {
+            return null
+          }
+        }
+
         // By default, any 'limit' is a single object literal containing one limitStatus.
         return (
           <LimitCard
--- a/src/billing/components/Free/PAYGConversion.tsx
+++ b/src/billing/components/Free/PAYGConversion.tsx
@@ -67,15 +67,17 @@ export const Credit250PAYGConversion: FC = () => {
           <ul className="credit-250-conversion-panel--benefits">
             <li>Unlimited buckets to store your data</li>
             <li>Unlimited storage retention</li>
-            <li>Unlimited dashboards</li>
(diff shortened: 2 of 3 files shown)


src/billing/components/Free/OrgLimits.tsx
```diff
@@ -104,6 +104,15 @@ export const OrgLimits: FC = () => {
           )
         }
 
+        // The 'dashboards', 'tasks', 'checks', and 'rules' limits are not applicable to orgs using IOx.
```
wdoconnell, line 107: Can we put this check at the beginning of the `return`/`render`, since it looks like it supersedes the other rules about which limits should be displayed?

src/billing/components/Free/PAYGConversion.tsx
```diff
@@ -67,15 +67,17 @@ export const Credit250PAYGConversion: FC = () => {
...
-            <li>Unlimited tasks</li>
-            <li>Unlimited alert checks and notification rules</li>
-            <li>HTTP and PagerDuty notifications</li>
             {!orgUsesIOx && (
-              <li>
-                Up to {Intl.NumberFormat().format(CARDINALITY_LIMIT)} series
-                cardinality
-              </li>
+              <>
+                <li>Unlimited dashboards</li>
```
wdoconnell, line 72: Does the absence of limits (no task limit, no alert limit, no pagerduty limit) only apply to TSM?

src/billing/components/Free/PAYGConversion.tsx
```diff
@@ -121,15 +123,18 @@ export const PAYGConversion: FC = () => {
                     <ul className="conversion-panel--benefits">
                       <li>Unlimited buckets to store your data</li>
                       <li>Unlimited storage retention</li>
-                      <li>Unlimited dashboards</li>
-                      <li>Unlimited tasks</li>
-                      <li>Unlimited alert checks and notification rules</li>
-                      <li>HTTP and PagerDuty notifications</li>
                       {!orgUsesIOx && (
```
wdoconnell, line 126: Same question as above.

wdoconnell (commented):
Couple of questions - main one of note was which of TSM versus IOx the "unlimited tasks, unlimited alert checks" messaging applies to.

src/billing/components/Free/PAYGConversion.tsx
mavarius: Only TSM should get this message, I can see how that isn't very clear. I can restate it as a ternary on `orgUsesIOx`.

Issue #1980, opened 2021-07-12
Enable variables to return table streams 


Anaisdg:
**Proposal:**

**Current behavior:**
<img width="896" alt="Screen Shot 2021-07-12 at 12 08 54 PM" src="https://user-images.githubusercontent.com/30506042/125328654-2e3f2380-e30a-11eb-8f65-8d8d8ef44da0.png">
<img width="726" alt="Screen Shot 2021-07-12 at 12 11 01 PM" src="https://user-images.githubusercontent.com/30506042/125328715-3f883000-e30a-11eb-8e34-701384a591a2.png">
Yields error 
` type error @2:4-2:11: expected [A] but found string (argument tables)`

**Desired behavior:**
The ability to call variables directly. 

**Alternatives considered:**
Describe other solutions or features you considered.

**Use case:**
Enables code reuse and streamlines flux development. 

Related to influxdata/flux#3872

kristinarobinson:
@Anaisdg - This cannot be supported using the existing architecture and how variables are defined.  I talked to @nathanielc, and he said that when they build the "flux module system" then this type of functionality could be implemented.  Closing as a UI issue.  Feel free to open a similar issue in IDPE and assign to the flux team.  I don't have an ETA for this new functionality.

PR #6904, opened 2024-06-26
chore: Replace `honeybadger-js` with `honeybadger-io/js`

abshierjoel:
Part of [#1980](https://github.com/influxdata/quartz/issues/7619)

This is based on honeybadger's migration guide here: https://docs.honeybadger.io/lib/javascript/support/upgrading-to-v3/

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,6 +162,7 @@
   "dependencies": {
     "@codingame/monaco-jsonrpc": "^0.3.1",
     "@docsearch/react": "^3.0.0-alpha.37",
+    "@honeybadger-io/js": "^6.9.3",
     "@influxdata/clockface": "^6.3.11",
     "@influxdata/flux-lsp-browser": "0.8.40",
     "@influxdata/giraffe": "^2.38.1",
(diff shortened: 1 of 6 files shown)


wdoconnell (approved):

wdoconnell:
Taking a look now 👍

wdoconnell:
This generally looks correct but I need to spend a little more time with it - the case that applies to us is somewhat different from the guide in that we're going from 1.x of the client-side package (instructions are for 2.x of client side package or 1.x of node package).

wdoconnell:
Finished reviewing - this looks correct. The guide isn't specific to honeybadger-v1.x upgrades but for our purposes the changes required look to be the same.

Only other thing I'd note from the [upgrade instructions](https://docs.honeybadger.io/lib/javascript/support/upgrading-to-v3/):
`Previously deprecated snake case config options such as api_key, project_root, etc. are no longer supported. Use apiKey, projectRoot instead.`

Importantly, this only affects the config options themselves. The arguments passed to the honeybadger v1 api (see scripts/upload_sourcemap.sh) are [unchanged](https://docs.honeybadger.io/api/reporting-source-maps), so "api_key" is there still "api_key" rather than `apiKey`.

I'll monitor this after it merges to make sure we don't lose honeybadger repoting.

PR #6910, opened 2024-07-22
fix: add guidance re usage query count

wdoconnell:
Clarifies that the query count on the C2 usage page reflects flux queries only.

### Checklist

### DIFF
--- a/cypress/e2e/cloud/usage.test.ts
+++ b/cypress/e2e/cloud/usage.test.ts
@@ -2,7 +2,7 @@ import {Organization} from '../../../src/types'
 
 const statHeaders = [
   'Data In (MB)',
-  'Query Count',
+  'Query Count (Flux Only) ?',
   'Storage (GB-hr)',
   'Data Out (GB)',
 ]
--- a/src/usage/UsageSingleStat.tsx
+++ b/src/usage/UsageSingleStat.tsx
@@ -1,10 +1,18 @@
 // Libraries
 import React, {FC, useContext} from 'react'
-import {Panel, ComponentSize, InfluxColors} from '@influxdata/clockface'
+import {
+  Panel,
+  ComponentSize,
+  InfluxColors,
+  QuestionMarkTooltip,
+} from '@influxdata/clockface'
 
 // Components
 import {View} from 'src/visualization'
 import {UsageContext} from 'src/usage/context/usage'
+
+// Constants
+const QUERY_COUNT_VECTOR_NAME = 'Query Count'
 
 // Types
 import {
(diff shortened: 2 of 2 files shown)


src/usage/UsageSingleStat.tsx
```diff
@@ -39,6 +44,10 @@ const UsageSingleStat: FC<Props> = ({
     decimalPlaces: {isEnforced: false, digits: 0},
   }
 
+  // Adjusts table properties to warn user that only flux queries are included in the Query Count.
+  const isQueryCount: Boolean = usageVector.name === VectorName.QueryCount
+  const vectorName = isQueryCount ? 'Query Count (Flux Only)' : usageVector.name
```
wdoconnell, line 49: This overrides the default backend-provided vector name.

src/usage/context/usage.tsx
```diff
@@ -41,6 +41,13 @@ const DEFAULT_USAGE: Usage = {
   status: RemoteDataState.NotStarted,
 }
 
+export enum VectorName {
+  QueryCount = 'Query Count',
+  Storage = 'Storage',
+  DataIn = 'Data In',
+  DataOut = 'Data Out',
+}
```
eatondustin1, line 49: If we're only using `'Query Count'` today, I think we're better off either inlining the `'Query Count'` string in `/src/usage/UsageSingleStat.tsx` or exporting a `const QueryCount = 'Query Count'` here instead. Otherwise, we're maintaining the other vector names without any confirmation that they're correct, which could lead to someone mistakenly relying on their accuracy in the future.

eatondustin1 (dismissed):
Looks good, but I did have one suggestion to mitigate future issues.

src/usage/context/usage.tsx
wdoconnell: That's a good point -- this was really anticipating future use since I wasn't happy with the fact that we don't identify the actual string values anywhere in the UI. But there is no need for any of the enum values other than `QueryCount` now.

wdoconnell:
Amended commit addresses @eatondustin1 's comment above ^

cypress/e2e/cloud/usage.test.ts
```diff
@@ -2,7 +2,7 @@ import {Organization} from '../../../src/types'
 
 const statHeaders = [
   'Data In (MB)',
-  'Query Count',
+  'Query Count (Flux Only) ?',
```
eatondustin1, line 5: I just noticed the `?` here - doesn't the tooltip element cover that?

wdoconnell: This is a change I identified after running the ui integration tests, which failed without the change. This is part of a test that designates what text to look for in the header. `QuestionMarkToolip` is a Clockface component that, among other things, renders and formats a `?` as text wherever the component is rendered. As a result, the test needed to be updated to address the fact that the text within the header now includes that ?.

eatondustin1: Ah, that makes sense, thanks for the explanation!

eatondustin1 (approved):

PR #6911, opened 2024-07-25
fix: correct tests to be handled by faster ci runner

wdoconnell:
Part of internal AUX team issue. 

Best reviewed with ?w=1 for whitespace changes

This PR resolves several issues with UI e2e tests that cause the tests to fail when running on a different CI runner. Primary issue is that existing runners generated some built-in delay was effectively adding a `.wait()` or `should()` where not present in the code. On the faster runner this is no longer the case.

1. Using `cy.wait`s to handle cases where a `click` did not wait for a corresponding DOM change. _Once this is tested with the new runner, I would prefer to convert `wait`s to asserting on the presence of specific DOM elements if time allows._
2. Using `.first()` to handle cases where later renders generated more than one DOM element matching on the selector when we are only looking for the first one.
3. Cleaned up cases where `cy.reload()` caused a page reload without setting feature flags.

### Checklist

### DIFF
--- a/cypress/e2e/shared/explorer.queries.graphs.test.ts
+++ b/cypress/e2e/shared/explorer.queries.graphs.test.ts
@@ -1,32 +1,34 @@
 import {Organization} from '../../../src/types'
 
+const featureFlags = {
+  showOldDataExplorerInNewIOx: true,
+  showTasksInNewIOx: true,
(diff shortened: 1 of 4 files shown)


hansenreid (approved):

wdoconnell:
Failed status checks were network errors unable to reach a pod.

PR #6927, opened 2024-08-20
chore: improve error notifications in annotations and notebooks

philjb:
I noticed that annotations, dashboards, labels, and notebooks sometimes display the api error when requests fail and sometimes don't- they are not consistent about it. I'd like to have it be consistent to be clear about errors.

I did not improve all error messages: variables, secrets, tasks, etc could also be improved but this is progress!

This needs to be ported to the OSS release branch.

Examples of the error notification popups:

Screenshot 2024-08-23 at 4 08 34 PM
Screenshot 2024-08-23 at 4 07 11 PM

### Checklist


blegesse-w (approved):

src/annotations/components/annotationForm/AnnotationForm.tsx
```diff
@@ -140,7 +140,7 @@ export const AnnotationForm: FC<Props> = (props: Props) => {
     }
 
     try {
-      dispatch(deleteAnnotations(editedAnnotation))
+      await dispatch(deleteAnnotations(editedAnnotation))
```
philjb, line 143: @blegesse-w, @wdoconnell  - perhaps you can help me here? I don't know react/redux so I'm just guessing. 

What I am observing is that when the api call within `deleteAnnotations` fails - it throws an Error if the http code is >= 300 - it is _not_ caught at line 149 - so this delete annotation always appears to be successful in the ui even if the api call fails. 

I guess that the issue is that the dispatch is async so this code advances to the notify success before the error arrives, but this isn't an async context so I added

```
const handleDelete = async (): Promise<void> => {
```

but that still doesn't catch the error thrown. 

Any idea?

I have noticed that if editing an annotation or creating an annotation fails at the api, those errors _are_ caught in their `catch` blocks but I can't see what is different in those code paths and this code path.

philjb: You can see here that the delete and put api calls are structured identically to my eye and these are where api errors are thrown

https://github.com/influxdata/ui/blob/f46e47de2d7ece95e2ad91beded0c4529a9c34e7/src/annotations/api/index.ts#L77-L101


The put/edit errors are caught tho and the delete one isn't.

philjb: This edit annotation seems to catch thrown api errors so i was trying to make the delete code look the same

https://github.com/influxdata/ui/blob/f46e47de2d7ece95e2ad91beded0c4529a9c34e7/src/annotations/components/EditAnnotationOverlay.tsx#L41-L57

blegesse-w: Hey @philjb thanks for tagging me in. I can't say what the problem is for sure but from a quick glance at the code, it does look like the issue might be with how the deleteAnnotations API call is handling errors. From what I can see, even though you're using await in your try...catch block, the error isn't getting caught as expected. This could be happening if the deleteAnnotations function isn't throwing the error properly when the API call fails.
Here is what i suggest. Double-check that deleteAnnotations is an asynchronous action creator that throws an error when the API responds with a status code of 300 or higher. This way, the error will propagate correctly up to your try...catch block.

blegesse-w: Also it's probably worth checking that the redux thunk is not swallowing the error before it reaches your catch block.

blegesse-w: @philjb No worries, Philip! I’ll take a closer look at this tomorrow with Bill, and we’ll figure out what's going on. I know the whole async/Redux world can be a bit tricky so I'll make sure to loop you in as I go

blegesse-w: I did some debugging on this locally and ended up reaching the same solution you implemented—adding `await` before the `dispatch(deleteAnnotations(editedAnnotation))` call. I tested the changes on my end, and everything is working smoothly now.

Nice job identifying the fix! We’re on the same page, so I think we can move forward with this.

Checkout the UI behavior after I forced an error in the `deleteAnnotations` thunk to simulate a failed API request:
https://github.com/user-attachments/assets/fa7a2925-73e1-4b2f-9be0-187fb943b674
error propagation is working and the UI displays the correct message on failure.

philjb: I think it was an issue with my local testing. I thought `yarn build` would rebuild everything but it apparently doesn't so i'm cleaning the `build/` dir now and then rebuilding. I was on the right track- sorry i didn't figure it out entirely myself. thank you for your help!! very validating.

blegesse-w (approved):
PR LGTM. But I do see a few unrelated failed monitor ci tests. You might need to push an empty commit so the tests can run on a clean pipeline.

philjb:
> PR LGTM. But I do see a few unrelated failed monitor ci tests. You might need to push an empty commit so the tests can run on a clean pipeline.

Thanks Beth!

These are the test failing that i see
```
2024-08-28T20:23:53.210Z        WARN      - job remocal-tsm-ui-test-chrome-1-tsm-shared/annotations.band failed (status: failed)
2024-08-28T20:23:53.210Z        WARN      - job remocal-tsm-ui-test-chrome-1-tsm-shared/tasks-searching failed (status: failed)
```

wdoconnell:
@philjb UI feature branches are clearing our tests properly now. You should be able to rebase against the current master branch and force-push to successfully re-run. Thanks for your patience.

philjb:
master merged in!

blegesse-w (approved):
Thank you for doing this!

Issue #6999, opened 2025-01-27
[notebooks]: if the autosave api returns an error, the ui still shows that the notebook saved, when it didn't


philjb:
## About the bug

The Notebooks page includes an autosave feature which saves the notebook on any edit. The ui says "autosaved"

![Image](https://github.com/user-attachments/assets/d362ba50-2b19-4e2e-8273-1b4a817651f8)

The ui for an edit makes a request to `PATCH /api/v2private/notebooks/0e56555f4d866000 HTTP/1.1` for example (0e56555f4d866000 is the notebook id) with the updated notebook as the patch data.

Like any request, this request can fail. If it does fail the UI doesn't indicate it, so it _looks like_ the notebook was saved when it wasn't.

For example, notebooks are saved to the sqlite database. If sqlite is configured to be read only, the api will return a response like 

```
{
	"code": "internal error",
	"message": "attempt to write a readonly database"
}
```

**Steps to reproduce:**
List the minimal actions needed to reproduce the behavior.

1. Launch influxd with a rw sqlite db like (note the `rwc`)

```
./influxd --sqlite-path "file:/path/to/.influxdbv2/influxd.sqlite?mode=rwc"
```

3. create a notebook and see that it is saved
4. shutdown influxd and ...
5. re-launch influxd with a read only sqlite db file by using this config option (note the `ro`)

```
./influxd --sqlite-path "file:/path/to/.influxdbv2/influxd.sqlite?mode=ro"
```
 
4. navigate to notebooks page, and make an edit
5. See that the ui doesn't reflect the error but you can see the error in the browser console network logs

**Expected behavior:**

Errors are shown in toaster popups

**Actual behavior:**
No error is shown and it looks like a notebook was saved when it wasn't.

**Visual Proof:**

## Requested Behavior

Show a pop error similar to this one that says "Could not autosave notebook: <actual internal error>" - the actual error will likely be "attempt to write a readonly database".

![Image](https://github.com/user-attachments/assets/8b213ac0-f8b5-41cc-9b92-19668a7b82b5)

## About your environment

**Environment info:**

OSS 2.7.x

PR #7009, opened 2025-02-14
feat(ui): show error when update notebook fail

NguyenHoangSon96:
Closes https://github.com/influxdata/ui/issues/6999

UI will show error notification when the auto-update notebook call fails:

Screenshot 2025-02-14 at 17 23 11

### Checklist


bednar (changes requested):
The source branch should be `main`; the `OSS-2.7` branch is used for releases.

NguyenHoangSon96:
> The source branch should be `master`; the `OSS-2.7` branch is used for releases.

Hi Jakub
I changed the source branch to the master
Please check again

src/flows/context/api.tsx
```diff
@@ -40,14 +46,18 @@ export const pooledUpdateAPI = (flow: PatchNotebookParams) => {
...
+      )
 
       stagedFlow = DEFAULT_API_FLOW
     }, REPORT_MAX_WAIT)
   }
 
   reportDecayTimeout = setTimeout(() => {
-    updateAPI(stagedFlow)
+    updateAPI(stagedFlow).catch(err =>
+      dispatch(notify(notebookEditFail(err.message)))
```
wdoconnell, line 59: Ideally we would not need to pass `dispatch` to this function, but would just throw an error so that the `dispatch` can be handled from the React component. However, I understand why you needed to do that here (since `updateAPI` is an async function subordinated to a synchronous function), so there is not a good way for us to throw the error further up the callstack without a meaningful restructure. I think this is fine for this case.

wdoconnell: I tested in local dev, and this does not seem to give rise to any rendering issues.

src/shared/copy/notifications/categories/notebooks.ts
```diff
@@ -41,6 +41,11 @@ export const notebookDeleteSuccess = (): Notification => ({
   message: 'Notebook was deleted successfully',
 })
 
+export const notebookEditFail = (message: string): Notification => ({
```
wdoconnell, line 44: This looks fine to me. My only observation from testing is that the default error notification time here is quite long (the toast remains for 10 seconds). However, since that is the standard elsewhere, I don't see a reason to diverge from it.

wdoconnell (commented):
@NguyenHoangSon96 This looks good to me based on my testing, with the caveat that I would (1) add the `dispatch` `useCallback` dependency I mentioned and (2) that you will need to rebase agains the current `master` branch, as there were some structural changes and a test update that will need to be incorporated. Thanks!

src/shared/copy/notifications/categories/notebooks.ts
NguyenHoangSon96: I agree ☝️

NguyenHoangSon96:
Hi @wdoconnell. Thank you for the review 😇
I fixed all things you mentioned but I have messup the rebase, I'm not very familiar with using rebase
So I created new branch with all the fix
Please check this branch [New branch](https://github.com/influxdata/ui/pull/7021), sorry for the inconvenience 🥹

wdoconnell:
> Hi @wdoconnell. Thank you for the review 😇 I fixed all things you mentioned but I have messup the rebase, I'm not very familiar with using rebase So I created new branch with all the fix Please check this branch [New branch](https://github.com/influxdata/ui/pull/7021), sorry for the inconvenience 🥹

No worries. In the future all you need to do is pull latest `master`, then `git rebase -i` your feature branch against `master`. In the interim you can close this PR, since we will address it in the other branch.

PR #7016, opened 2025-02-25
fix: update retention rule for CLOUD and OSS

### LINKED ISSUE #1607: Changing retention to "forever" in UI does not work
__Steps to reproduce:__

1. Open the InfluxDB Web GUI
2. Create a bucket with limited retention time, e.g. 30 days
3. Edit the bucket and set Delete Data to "Never"

__Expected behavior:__
The retention of the bucket is changed to "forever"

__Actual behavior:__
The retention of the bucket is unchanged

__Environment info:__

* System info: Linux 5.4.0-1043-azure x86_64
* InfluxDB version: InfluxDB 2.0.6 (git: 4db98b4c9a) build_date: 2021-04-29T16:48:12Z
* Other relevant environment details: Containerized on Azure Kubernetes Services (Cluster version 1.19.9, Nodepool version 1.18.14)

NguyenHoangSon96:
Closes #1607

Fix the bug when updating the retention period of buckets:
-  Before fix: set a retention period, eg. 1 hour, then change to never -> the retention period is still 1 hour
-  After fix: set a retention period, eg. 1 hour, then change to never -> the retention period changes to never

The recorded screen is the result when update in OSS:

https://github.com/user-attachments/assets/88c03b7a-06fb-4f68-a66b-62bddf8217b8

The image is the result  when update in CLOUD:

Note: I have to mock the variable CLOUD because I don't know how to run the CLOUD in my local. In the captured screen the retentionRules in the payload is not exists when send to the backend like ui team suggested
Screenshot 2025-02-25 at 15 14 05

### Checklist

### DIFF
--- a/src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
+++ b/src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
@@ -1,13 +1,13 @@
 // Libraries
 import React, {
-  ChangeEvent,
-  FormEvent,
   FunctionComponent,
-  useCallback,
   useEffect,
   useState,
+  ChangeEvent,
+  FormEvent,
+  useCallback,
 } from 'react'
-import {useHistory, useParams} from 'react-router-dom'
+import {useParams, useHistory} from 'react-router-dom'
 import {useDispatch, useSelector} from 'react-redux'
 import {get} from 'lodash'
 
--- a/src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
+++ b/src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
@@ -154,7 +154,7 @@ const UpdateBucketOverlay: FunctionComponent = () => {
       if (CLOUD) {
         setBucketDraft({
           ...bucketDraft,
-          retentionRules: undefined
+          retentionRules: undefined,
         })
         return
       }
@@ -163,7 +163,6 @@ const UpdateBucketOverlay: FunctionComponent = () => {
         ...bucketDraft,
         retentionRules: [{everySeconds: 0}],
       })
-
     }
   }
 
--- a/src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
+++ b/src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
@@ -1,13 +1,13 @@
 // Libraries
 import React, {
-  FunctionComponent,
-  useEffect,
-  useState,
   ChangeEvent,
   FormEvent,
+  FunctionComponent,
   useCallback,
+  useEffect,
+  useState,
 } from 'react'
-import {useParams, useHistory} from 'react-router-dom'
+import {useHistory, useParams} from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
```diff
@@ -151,9 +151,17 @@ const UpdateBucketOverlay: FunctionComponent = () => {
...
+          ...bucketDraft,
+          retentionRules: undefined,
+        })
+        return
+      }
+
       setBucketDraft({
         ...bucketDraft,
-        retentionRules: [],
+        retentionRules: [{everySeconds: 0}],
```
wdoconnell, line 164: I think we can make this a one-line change (and omit the multiple `setBucketDraft` invocations) if we just use a ternary operator here. Would also be safer imho as it's clear from the line itself that there is no change in Cloud environments to the existing behavior.

>      setBucketDraft({
>         ...bucketDraft,
>         retentionRules: CLOUD ? [] : [{everySeconds: 0}],
>       })
>

wdoconnell (commented):
Thanks for doing this: I have one suggestion about how to convert this to a ternary. Otherwise, it looks good to me. Let me confer with the folks working on the OSS backend to ensure the OSS-specific change is correct.

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
wdoconnell: Note also that the above change is difference in that, rather than passing an array, it would pass `undefined`. Let's stick to the existing behavior of an empty array so as to avoid an unintentional change in behavior in C2.

NguyenHoangSon96:
Hi @wdoconnell. Thank for the review 
I fixed all things you mentioned
Please check again

wdoconnell (commented):
This looks good now from a UI perspective. 

Please wait on a review from folks who are closer to the OSS for a backend approval. 

FYI @davidby-influx: this is the PR handed off from @philjb in Slack.

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
```diff
@@ -1,13 +1,13 @@
 // Libraries
 import React, {
-  FunctionComponent,
-  useEffect,
-  useState,
   ChangeEvent,
   FormEvent,
+  FunctionComponent,
   useCallback,
+  useEffect,
```
devanbenz, line 7: I assume this is a linter change?

devanbenz (approved):
LGTM just a comment about the linter changes

src/buckets/components/createBucketForm/UpdateBucketOverlay.tsx
NguyenHoangSon96: Actually, It from my IDE runs optimize import
I have changed it back

devanbenz (approved):

karel-rehor (approved):
Makes sense

blegesse-w (approved):
looks good!

PR #7060, opened 2025-07-25
chore(deps): bump form-data 4.0.0 to 4.0.4

jdstrand:

PR #7061, opened 2025-07-29
fix: pin expect and pretty-format

wdoconnell:
Helps with https://github.com/influxdata/ui/pull/7060

CI is currently failing due to an incompatibility between the Node version used in CI and updated versions of the packages in this PR. Immediate way of resolving this is for now is to pin these specific package resolutions.

Separately we will need to update the Node versions in the CI images, which requires a few steps but will be a more sustainable long-term fix.

### Checklist

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -13,6 +13,7 @@
     "www.googletagmanager.com",
     "www.honeybadger.io"
   ],
+  "video": false,
   "viewportWidth": 1400,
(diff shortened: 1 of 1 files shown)


logan-bobo (dismissed):

jdstrand (dismissed):
Assuming tests pass, LGTM. Thanks!

david-rusnak (approved):

PR #7087, opened 2025-10-22
fix: invalid config for vsphere

NguyenHoangSon96:
Closes [26912](https://github.com/influxdata/influxdb/issues/26912)

- My steps to fix:
1. Check out the source [telegraf](https://github.com/influxdata/telegraf) to my local.
2. Run `go run ./cmd/telegraf config > etc/telegraf.conf` command to generate `etc/telegraf.conf` file.
3. Tweak the `src/writeData/utils/updateTelegrafPlugins.mjs` so it will get contents from my local `influxdata/telegra/etc/telegraf.conf` instead of the remote repository `/influxdata/telegraf`.
4. Run `telegraf-plugins:update` command to generate new contents for `md` and `conf` files.
5. Create this PR.

**This is just a very primitive implementation of the updateTelegrafPlugins.mjs to make the script work**
[updateTelegrafPlugins.txt](https://github.com/user-attachments/files/23097791/updateTelegrafPlugins.txt)

- Some thoughts:
1. I don't know how these things should work exactly; this is just my assumption. If my fixes are not correct, then I will just manually remove the `-` at the beginning of the `.conf` file.
2. If this is correct, then we should update the `src/writeData/utils/README.md` guidance and the `src/writeData/utils/updateTelegrafPlugins.mjs` script file to get contents from `etc/telegraf.conf` in the local repository, not the remote.

- Before the fix, there is a `-` at the beginning of the file.

https://github.com/user-attachments/assets/b2266961-e5a1-4e55-9eb0-bfdb064c5ad6

- After I modify `src/writeData/utils/updateTelegrafPlugins.mjs` and run the command `telegraf-plugins:update`, the `-` at the beginning of the file disappears.

https://github.com/user-attachments/assets/dcb2fd5a-4563-4e6c-935c-b2f8d8073d4b

### Checklist

### DIFF
--- a/src/writeData/components/telegrafInputPluginsConfigurationText/vsphere.conf
+++ b/src/writeData/components/telegrafInputPluginsConfigurationText/vsphere.conf
@@ -1,4 +1,4 @@
--# Read metrics from one or many vCenters
+# Read metrics from one or many vCenters
 [[inputs.vsphere]]
   ## List of vCenter URLs to be monitored. These three lines must be uncommented
   ## and edited for the plugin to work.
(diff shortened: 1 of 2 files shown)


philjb:
> * I edited the `updateTelegrafPlugins.mjs` a little bit so it only modifies the ".conf" and ".md" of the vsphere file.
> * But I don't know, should we need to generate contents for the rest of the plugins ".conf" and ".md" files? Looks like they are all out of sync compared to  "influxdata/telegraf/${version}/etc/telegraf.conf". Maybe we should do it in another PR 🤷

I would regenerate all of them.  You can do that in a separate pr if you like.

I will approve this pr since it is only touching statically served files and I assume it imported them as they are in the telegraf repo.

philjb:
> I may be missing some context here. Based on the original issue, do we need changes other than just removing the dash? This looks like more changes than that.

Yes and no -- we could just remove the dash but the dash is there because these configs are programmatically generated from the telegraf repo and at the time this conf was generated, the dash was in the telegraf config (incorrectly). 

The other changes in this pr bring this telegraf plugin conf up to date with the latest conf in the telegraf repo. 

It's been a long time since these were sync'd. Gunnar used to do them.

